PR #866, opened 2015-06-23
True infinite scroll

theefer:
Replaces the current "load more when you reach the bottom" with a lazy view that you can scroll anywhere to. Images get loaded around the viewport automatically as you move around. (Images far from the viewport are `display: none`'d in the DOM for performance.)

Resizing the window and opening the side-panel should make cells adjust accordingly.

Clicking on an image and then navigating Back should bring the search view back at the same scroll position. (The `<- search` link currently doesn't do that.)

*This is ready to be reviewed but needs a little more work, user testing and analysis before it can be rolled out*

Sorry no video yet, haven't found a good workflow to capture GIFs, but happy to demo to people or put it on TEST tomorrow.

:heart: RxJS

## Performance

I did some basic Timeline performance analysis to iron out the main performance bottlenecks but there is still plenty of space for improvement in making it smoother and faster I think. In particular:
- ngAnimate seems to get in the way and slow things down. I've tried to disable it locally in my directive but none of the tricks mentioned in the docs worked :-(
- More tuning of the observable caching, debounce and delays to reduce work.

Also worth doing some more heap dump to check there is no leak.

I thought of releasing image data for images far from the viewport too, but that might conflict with an active selection so I'll leave that for later.

## Before release

- [x] Try out with a few users (incl. Jonny?) to see their reaction, get their feedback and confirm this is better
- [x] Check the scrollbar sensitivity in TEST and whether the current 5000 result limit is enough/right
- [ ] Check for slowdown on editorial Mac's, any memory leaks

## After release

- [x] Show “loading” placeholders as a UX trick for simulating faster load than it actually is
- [ ] Fix Back button and `<- search` link to also load search results at the same scroll position
- [ ] More cleanup of integration with search results controller, abstract directive into its own standalone package?
- [ ] Optimise unsubscription of listeners/observables when disposing of a large set of results
- [ ] Avoid doing two API calls (one to fetch the total number of images, then another one to get the initial data)
- [ ] Bug when images are shown, hidden and then revealed again after more than the image expiration time. Are cache-control headers wrong?


theefer:
Merged master in, addressed some more issues and added some more perf improvements.

theefer:
It's on TEST again, does seem smoother subjectively.

kenoir:
@theefer when you scroll out of the set of loaded images you are left with a blank area meaning that the sense of moving through a set of images gets lost.

foo2
 
Compare: 

foo

**I'm not suggesting adding a stripy background** - just to compare the sense of movement during a scroll.

paperboyo:
I agree with @kenoir. Maybe the most sensible would be to draw just empty background rectangles, like:

empty rect

theefer:
@kenoir Yes I've got a separate branch with placeholders added, I'm just trying to improve the performance without the added complexity of placeholders first (put otherwise: if adding placeholders makes everything more sluggish and buggy, is their addition worth their while for an initial release?). It's looking unlikely that I am going to responsibly ship this before going away, so might pick up when I'm back next week.

alastairjardine:
My sense and advice would be to hold off shipping until you have performance _and_ background context. It's really hard to keep sense of space and place with out some type of affordance.

paperboyo:
Thinking about it some more... While I like my empty rectangles, they could hint at the actual _place_ you’ve scrolled into, by e.g. displaying just a date/time, similarly to how GDocs are indicating page number for longer docs (unless, obviously, it will impact negatively on performance so much that it will start to be a resolution to the problem it created :laughing:):

place indicator for scroll

theefer:
@paperboyo Yes that's also been mentioned, more of a nice to have for the future though.

theefer:
Pushed [my WIP branch](https://github.com/guardian/grid/compare/sc-infinite-scroll...sc-infinite-scroll+placeholders-bis?expand=1) that has placeholders to TEST, if people want to have a play to compare.

cc @alastairjardine

jamesgorrie:
:+1: Tested locally too.
As  note Traceur is out of date, but we can update that in another PR.

theefer:
Had to introduce an extra div to get the positioning right after merging master in. :cry:

All looking good now, I just want to do a little bit of performance testing against PROD as the rather bad API latency in PROD (in general) combined with this may push the API over the limit of acceptable latencies.

kahuna/public/js/components/gu-lazy-table/gu-lazy-table-cell.js
```diff
@@ -0,0 +1,38 @@
+import angular from 'angular';
+
```
jamesgorrie: Noticed you haven't imported your helpers here.

jamesgorrie: Or in a few places below.

theefer: Too many layers of modules. Added now.

PR #899, opened 2015-07-02
Use correct cluster name in thrall and media-api conf.

akash1810:
Cluster name matches the one defined in the [ES yaml](https://github.com/guardian/grid/blob/master/elasticsearch/elasticsearch.yml#L6).

The logs for thrall and media-api no longer report `node null not part of the cluster Cluster [media-api], ignoring`

(Not entirely sure how these services were working when referencing a cluster that was different from elasticsearch?!?!)

### DIFF
--- a/media-api/conf/application.conf
+++ b/media-api/conf/application.conf
@@ -3,7 +3,7 @@ application.langs="en"
 session.httpOnly=false
 session.secure=true
 
-es.cluster = "media-api"
+es.cluster = "media-service"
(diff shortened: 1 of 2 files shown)


theefer:
Any updates on this one?

akash1810:
The changes are fine locally, however I don't fully understand how; I've read that using the ES REST API doesn't care about the cluster name. Closing until I find out more.

PR #901, opened 2015-07-03
Updating thrall health check end-point

akash1810:
Updating thrall health check end-point to report unhealthy if:

- Not processed a heartbeat message from SQS since last health check.
- Not processed a heartbeat message in `heartRate` minutes.

Where `heartRate` is the value of `sqs.heatbeat.message.min.frequency` in `thrall.properties` (5 minutes ATM).

Before release:
- [ ] [Apply CF update](https://github.com/guardian/grid-infra/pull/97) to send heartbeats.

### DIFF
--- a/thrall/app/controllers/HealthCheck.scala
+++ b/thrall/app/controllers/HealthCheck.scala
@@ -1,17 +1,32 @@
 package controllers
 
-import play.api.mvc.{Action, Controller}
+import play.api.mvc.{Result, Action, Controller}
 import play.api.libs.concurrent.Execution.Implicits._
-import lib.{MessageConsumer, ElasticSearch}
+import lib.{Config, MessageConsumer, ElasticSearch}
 import com.gu.mediaservice.syntax._
 
 object HealthCheck extends Controller {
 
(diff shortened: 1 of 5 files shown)


thrall/app/controllers/HealthCheck.scala
```diff
@@ -1,17 +1,35 @@
...
-      .map(_ => Ok("OK"))
+      .map(_ => Ok("ES is healthy"))
   }
 
+  def sqsHealth = {
+    val beatsSinceLastCheck = MessageConsumer.heartbeats.getAndSet(0)
+    val lastHeartBeat = MessageConsumer.lastHeartBeat.get
+
+    if (beatsSinceLastCheck == 0)
+      ServiceUnavailable("Heart has not beat since last health check")
```
theefer: What is this check for? Does that mean GETing the healthcheck endpoint results in a state mutation on the server?

akash1810: > Does that mean GETing the healthcheck endpoint results in a state mutation on the server?

Yes. The counter gets reset on a GET.

PR #910, opened 2015-07-08
Autocomplete credit on Image view

kenoir:
Add the autocomplete credit field to the individual image view.

foo


paperboyo:
> order of what elastic decides is closest to the search

Does it allow for exceptions? I’m thinking of people not realising that the correct credit should read “The [Guardian/Observer]” and not “[Guardian/Observer]”. If these would be added as exceptions to the autocomplete, it could suggest only the correct content of credit. Minor point, though!

PR #914, opened 2015-07-08
Flatten Staff Photographer config

kenoir:
This change is to make it easier to write a staff photographer parser for usage rights.

The config is moved closed to a "flat representation" as might be loaded from an external source, in order that it can be more flexibly manipulated for various uses.


theefer:
:+1:

PR #916, opened 2015-07-09
Replacing "upload" with an icon.

akash1810:
Results in a (slight) reduction in the break-point from 1044px to 991px.

Before
screen shot 2015-07-09 at 08 59 44

After: at > 1200px wide
screen shot 2015-07-09 at 10 38 12

After: at <=1200px wide
screen shot 2015-07-09 at 08 59 30

### DIFF
--- a/kahuna/public/stylesheets/main.css
+++ b/kahuna/public/stylesheets/main.css
@@ -165,7 +165,7 @@ button {
     box-sizing: border-box;
     color: white;
     display: block;
-    padding: 5px 20px !important;
+    padding: 5px 20px;
     font-size: 1.4rem;
     position: relative;
 }
@@ -308,7 +308,7 @@ textarea.ng-invalid {
    ========================================================================== */
 
 .top-bar-item {
-    margin-left: 20px;
+    margin-left: 10px;
     vertical-align: middle;
     line-height: 28px;
     display: inline-block;
@@ -518,6 +518,15 @@ input.search-query__input {
     display: none;
 }
 
+.file-uploader__select-files {
+    padding: 5px 10px;
+}
+
+@media screen and (max-width: 1200px) {
+    .file-uploader__select-files--label {
+        display: none;
+    }
+}
 
 /* ==========================================================================
    Results
--- a/kahuna/public/js/upload/file-uploader.html
+++ b/kahuna/public/js/upload/file-uploader.html
@@ -9,6 +9,9 @@
                multiple
                ui:file ui:file-change="fileUploader.uploadFiles" />
 
-        <button class="file-uploader__select-files button">upload</button>
+        <button class="file-uploader__select-files button">
+            <i class="material-icons">file_upload</i>
+            <span class="file-uploader__select-files--label">upload</span>
+        </button>
     </div>
 </form>


theefer:
Would it be clearer if it was icon+text when there is space, and just icon when there isn't? That way users can learn about the meaning of the button and the icon on wider screens and associated it to just the icon on smaller screens?

FWIW people already seem confused by the upload button, mostly on the upload screen, although I don't think this is making it worse.

kahuna/public/stylesheets/main.css
```diff
@@ -518,6 +518,9 @@ input.search-query__input {
     display: none;
 }
 
+.file-uploader__select-files.button {
+    padding: 5px 10px !important; /* !important to override .button's !important */
```
theefer: Why does the button have `!important` again?

akash1810:
+1 for responsive.

kahuna/public/stylesheets/main.css
akash1810: Not entirely sure. Git blame shows it was introduced [here](https://github.com/guardian/grid/commit/5c145fbe71a8999321f2e72482edabdfae552e77#diff-f7f0cf9f6b1bcaecc2be857deae091b2R67), but that's a reverting commit - will remove the rule and see how it changes things rather than going on a git history stack trace...

theefer: :+1:

kahuna/public/stylesheets/main.css
```diff
@@ -518,6 +518,15 @@ input.search-query__input {
     display: none;
 }
 
+.file-uploader__select-files {
+    padding: 5px 10px;
+}
+
+@media screen and (max-width: 1200px) {
```
akash1810: Whilst we've not got a pre-processor, should there be a separate file for responsive rules so we wouldn't have to repeat this query? i.e

```css
@media screen and (max-width: 1200px) {
    body: {
        color: hotpink;
   }

/* more rules that use media queries */

   .file-uploader__select-files--label {
     display: none; 
   }
}
```

vs.

```css
@media screen and (max-width: 1200px) {
    body: {
        color: hotpink;
   }
  }

/* more rules that may or may not use media queries */

@media screen and (max-width: 1200px) {
   .file-uploader__select-files--label {
     display: none; 
   }
}

```

jamesgorrie:
I would argue that this is more confusing - especially since we'll be introducing a whole bunch of people to it relatively soon. I wouldn't assume everyone knows what that icon is.

This doesn't solve the overall problem that on the upload screen it looks like a button to complete your action.

I think the language would be more important - i was playing around with "add images" or something like that (although I still don't think this makes a huge difference between "upload").

jamesgorrie:
Playing with it, using a `+` and "add" makes a bit of sense as it seems slightly less of a completion word.
I would be completely wrong though.

Whatever we go with - could we stick tracking on it to see if people are using the button at all?

addimages

akash1810:
> This doesn't solve the overall problem that on the upload screen it looks like a button to complete your action.

This is more around reducing the min-width of the top bar.

However, the YouTube solution may work here - the "upload" button is a link to `/upload` and on the uploads page, there's a bigger "upload files" button.

youtube-upload-flow

alastairjardine:
Isn't all this tinkering, rather than solving the actual problem that the uploading flow is really confusing and needs work?

kahuna/public/stylesheets/main.css
kenoir: Perhaps simply organising the contents of the main.css file in the 1st way you describe would suffice?

akash1810:
> rather than solving the actual problem that the uploading flow is really confusing and needs work

Agreed. This PR isn't addressing this issue though, its just making the break point smaller.

jamesgorrie:
:+1: to the shrinking - it helps a bit.

kahuna/public/stylesheets/main.css
jamesgorrie: I wouldn't say so - it goes against our component based organisation.
If I were working on the upload button I want to open upload.css and edit everything in there.

akash1810: It would, but the more rules the more you'll have to remember what scope (wrong term) you're in when looking at a rule - a separate file would make that clearer. No strong opinions either way and probably premature optimisation at this point.

theefer:
:+1: to icon and collapsing on smaller breakpoint.

This isn't solving any upload problems; this just makes the search bar break slightly less on narrow screens.

alastairjardine:
@akash1810 @theefer Cool – but let's chat next time and consider it a bit more holistically.

PR #917, opened 2015-07-09
Fix jspm version

kenoir:
The current version causes an error when attempting to install via npm.


theefer:
Odd, sorry about that! :+1:

PR #920, opened 2015-07-09
Usage rights category logging

jamesgorrie:
As we have indexed some images with `{ "usageRights": { "supplier": "Gonzo" } }` it was failing to parse them.

Reindex will fix this.


kenoir:
:+1:

PR #921, opened 2015-07-09
No cost for NoRights

jamesgorrie:
This is so we fallback to any other calculations we have.
Fixes bug in master at the moment.


kenoir:
:+1:

PR #919, opened 2015-07-09
Usage rights properties

jamesgorrie:
Creating forms somewhat dynamically.
Seems a little like overkill - but a lot simpler than having all the logic on the client.

I want to look into reflection similar to the Play JSON stuff or perhaps a lib, but I can't find one that could do the nice stuff like the optionsMap I have here.

usage-rights-staff-phot

PR #922, opened 2015-07-09
Usage rights on search page

kenoir:
This PR makes usage rights editable on the search page along with other metadata.

crop

Depends: https://github.com/guardian/grid/pull/919

**TODO:**
- [x] Make publication into radio-button style rather than select
- [x] Group photographer options 
- [x] Order dropdowns alphabetically for usage-rights panel
- [x] Fix updating search result status on batch save


kenoir:
Here's a (non-functional) prototype of what the usage rights might look like stuffed into the metadata editing panel on the search page:

foo

What do you think @alastairjardine @theefer?

theefer:
Would it make sense to always display the current image category, and the UI would be an expanded version allowing to edit it?

From the picture editor POV (one with edit rights), it seems that we'd usually be in one of the following cases:
* The category has been auto-detected correctly (agency, photographer, etc), the image is free and I don't have to do anything. I still have the option to edit in case it's been detected wrong, but that should be rare.
* No category has been auto-detected, the image is red and I want to make it free/restricted.

Note that the POV of a user *without* edit rights may be slightly different, but they may still want to be able to see at a glance that it's a PR image, a staff photographer pic, etc?

PS: It's stuck a in really weird place (between byline and credit), but I assume that's just a temporary hack.

theefer:
Would be nice for the UX to be somewhat unified with that of the upload form, eventually.

kenoir:
> Would it make sense to always display the current image category, and the UI would be an expanded version allowing to edit it? 

Yup.

> It's stuck a in really weird place (between byline and credit), but I assume that's just a temporary hack.

What about right at the top, underneath where the restrictions appear it they exist?

theefer:
> What about right at the top, underneath where the restrictions appear it they exist?

Given that it is driving the restrictions/pay banners, that makes sense to me.

kenoir:
Progress so far:
foo

theefer:
Nice start!

kenoir:
Getting there. Save is working and the UI is a bit clearer.

Need to sync changes back to view.

foo3

kenoir:
Batch updating of usage rights not reflected in search and selection:

foo3

kahuna/public/js/usage-rights/usage-rights-editor.js
```diff
@@ -0,0 +1,158 @@
...
+]);
+
+usageRightsEditor.controller(
+    'UsageRightsEditorCtrl',
+    ['$q', '$scope', '$window', '$timeout', 'editsService', 'editsApi',
+    function($q, $scope, $window, $timeout, editsService, editsApi) {
+
+    var ctrl = this;
+
+    ctrl.$q = $q;
```
kenoir: I'm pretty sure this isn't the best way to do this. Advice appreciated.

kahuna/public/js/usage-rights/usage-rights-editor.js
```diff
@@ -0,0 +1,158 @@
...
+
+    function save(data) {
+        ctrl.saving = true;
+        var a = ctrl.images.map((image) => {
+            return editsService.
+                update(getResource(image), data, image).
+                then(resource => {
+
+                    // Cost is a calulated field so we must retrieve it
+                    return image.get().then((newImage) => {
```
kenoir: We have to retrieve the whole image here as there's no simple way to extract the cost information. 

As cost is a generated field we can't rely on the current mechanism of applying the edit client side after the change is successful (as the cost will not be updated in the client as it is on the server).

kahuna/public/js/components/gr-panel/gr-panel.js
```diff
@@ -53,15 +58,35 @@ grPanel.controller('GrPanel', [
...
+            ctrl.selectedCosts = selection.getCost();
+            ctrl.selectedUsageRights = selection.getUsageRights();
+        };
+
         $scope.$watch(() => selection.getMetadata(), onValChange(newMetadata => {
             ctrl.rawMetadata = newMetadata;
             ctrl.metadata = selection.getDisplayMetadata();
+            ctrl.images = Array.from(ctrl.selectedImages);
+
+            $rootScope.$broadcast('usage-rights:update-images', ctrl.images);
```
kenoir: Is `$broadcast` an acceptable thing to do here? Don't want to end up with event spaghetti.

jamesgorrie:
Bugs:
- [ ] Upload button bounces when clicked
- [ ] Calendar is a bit weird

kahuna/public/js/components/gr-panel/gr-panel.js
```diff
@@ -53,24 +59,47 @@ grPanel.controller('GrPanel', [
             });
         };
 
+        ctrl.updateUsageRights = () => {
```
kenoir: I'm really struggling to work out how to sensibly propagate the image changes back to the view. After we've saved we need to replace the image model in the search view and in the panel, and I don't know how to get a handle on those objects.

jamesgorrie:
On TEST.

theefer:
Would we be comfortable rolling this out even before all old images have been reingested (i.e. the editor will not be showing the correct category for all those images)?

akash1810:
Do we need to compare `usageRights.restrictions` too? For example:

Image A has:
```json
"usageRights": {
  "category": "social-media",
  "restrictions": "foo"
}
```

Image B has:
```json
"usageRights": {
  "category": "social-media",
  "restrictions": "bar"
}
```

At the moment, we're checking if the category matches, so if I select both Image A and Image B, the ure form doesn't warn that the restrictions vary, instead the field is blank.

jamesgorrie:
@akash1810 I personally think that adds way to much complexity for very little gain.

These fields are also controlled on the server so it would have to be generic enough to deal with all fields e.g. photographer.

My preferred model would be the 500px / Flickr model which is you have matching || multiple. For multiple, you make it clear that you want to edit all at one go and do it.

jamesgorrie:
@theefer there is [some code](imageArr()) here to deal with the model before ingestion that we can remove after. It will work irrespective of the model being correct / old.

kahuna/public/js/edits/service.js
```diff
@@ -105,8 +105,12 @@ service.factory('editsService',
 
         return resource.put({ data }).then(edit =>
             getSynced(originalImage, newImage => matches(edit, newImage))).
-            then(edit => {
+            then(({ edit, image }) => {
+
                 runWatcher(resource, 'update-end');
+
+                $rootScope.$emit('image-updated', image, originalImage);
```
jamesgorrie: This is a bit of an oddity of us having - now - two ways of emitting events.
@kenoir and I think it would be worth taking a breath and trying to untangle some of this mess as it's not only creating multiple paths to the same ends, there are couplings that are making things hard to move.

I personally would like to move this logic to the image service, which would expose a metadata object (similar to the [`usageRights`](https://github.com/guardian/grid/blob/master/kahuna/public/js/image/service.js#L10)) which would work in a silo with the UI . The image service would use the [edits service events](https://github.com/guardian/grid/pull/922/files#diff-81f57b8057236574f9266cb5d4cb54e3R23) to know it's been updated and update itself.

jamesgorrie: @theefer could we discus the cost / value benefit here?

theefer:
A few minor comments, none particularly blocking:

Should we progressively reveal the Collection field only for some agencies? I can't remember which ones we fill it in for (if any) besides Getty, but it may just be confusing for others?

image

Yet another "saving" visual indicator?

image

Cancel/Save buttons seemingly in opposite position in rights vs metadata editor. It seems more natural to me that Save is to the right of Cancel (like on rights).

image

image

I find the concept of the "bin" to remove rights rather confusing. I forget why we can't just have a "None" category for all, simplifying the choice of completion actions?

image

If the rights editor dialog is open, I would expect that changing the selection (e.g. selecting one more image) closes it? In particular it doesn't seem to always get correctly updated when toggling between same/mixed categories.

Should we provide more detail than the mere category name? e.g. "Agency: Getty Images", "Staff Photographer: Tom Jenkins", etc. Else the only way to review the data (e.g. for correctness) is to open the editor.

image

Similarly, there doesn't seem to be a way to see the restriction (or "mixed") of your selection.

image

kenoir:
@akash1810 
> At the moment, we're checking if the category matches, so if I select both Image A and Image B, the form doesn't warn that the restrictions vary, instead the field is blank.

This is intentional. We provide a visual indicator that you have multiple varying restrictions within the same category and other fields are blank as we leave it to the user to fill in what is appropriate.

kenoir:
> Similarly, there doesn't seem to be a way to see the restriction (or "mixed") of your selection. 

Restrictions can be seen in the

jamesgorrie:
__@theefer re collections:__
We do have certain collections, but I have been asked by Jonny (multiple times) and Jo (once or twice) to be able to add in information that isn't in our list and won't affect usage. This is better I think as we could potentially use this data later.

It would be nice to autocomplete on collections of suppliers we know about... :bicyclist:

jamesgorrie:
__re: deletion__:
I've moved the buttons around.
We can't set `NoRights`until we get rid of the old cost model.
This is adding the ability to reset - which if we need to bench, we can for now.

undo-usage-rights

jamesgorrie:
__re another saving indicator__ we don't have one for this little save button, so hopefully we can just adopt this one moving forward?

jamesgorrie:
> In particular it doesn't seem to always get correctly updated when toggling between same/mixed categories.

Could you give an example as I can't replicate?

jamesgorrie:
> If the rights editor dialog is open, I would expect that changing the selection (e.g. selecting one more image) closes it? 

I wouldn't think so as you would be able to see the changes your selection has on the box (if it's working correctly, which it looks like it is)

jamesgorrie:
> Should we provide more detail than the mere category name?

This would be nice, but the way we're diffing the usage rights at the moment is on the category only.
I am working on a branch that works on a more intelligent way of surfacing data to the panel - but if you think this is worth it I will add it.

jamesgorrie:
> In particular it doesn't seem to always get correctly updated when toggling between same/mixed categories.

 I potentially read this wrong - when there are multi-rights, we have a warning saying that you're going to override this on all of them. I've made this a little cleaner with [this warning](https://github.com/guardian/grid/commit/97cc9afb6e931613c33bbbb358a5a26d233a7a35#diff-64c53e166971d3124ace68ce226bc331R30):

multi-usage-rughts-warn

theefer:
> > In particular it doesn't seem to always get correctly updated when toggling between same/mixed categories.
> Could you give an example as I can't replicate?

I think it may have been a bug with the selection instead, as it said 2 images selected when I could only see one.

theefer:
> This would be nice, but the way we're diffing the usage rights at the moment is on the category only.
I am working on a branch that works on a more intelligent way of surfacing data to the panel - but if you think this is worth it I will add it.

Not needed for a first pass, let's queue it up.

theefer:
On the following:

img

What does the "Left arrow" do? To me it should revert any change and close the dialog, but that's already what the X does? And how do you remove an existing usage right?

kenoir:
@theefer ic_undo_black_24dp_2x removes any overrides.

theefer:
The arrow is a bit counter intuitive to me, I think the bin seemed like a better icon for that purpose. YMMV.

theefer:
(Left arrow = Undo)

kenoir:
Or maybe this one: ic_settings_backup_restore_black_24dp_2x
?

akash1810:
> The arrow is a bit counter intuitive to me

Might be better to include some copy, e.g. `Revert to original`?

alastairjardine:
Or perhaps just the text "undo". We're introducing YABS (Yet Another Button Style), which I would strongly advice against doing. 

This whole piece needs a UX polish. What is the best way to make that happen? (UX & usability is part of Q2 cc. @blishen )

theefer:
Is it reverting to what was previously there before you opened the panel (I thought that's what X did?) or is it clearing any usage right? I thought the latter.

alastairjardine:
And why can't we just talk about this face to face rather than on PRs?

theefer:
@akash1810 and @jamesgorrie are not at their desk at the moment. Providing general feedback here made it visible to all (including you!).

kenoir:
@alastairjardine discussions in PRs provide visibility and justification for code changes. 

Icons should all come from here: https://www.google.com/design/icons/

kenoir:
@alastairjardine we'll hold release until we've talked it through.

theefer:
After chat, OK to remove Delete/Revert everywhere (incl upload).

jamesgorrie:
:camel: anyone have a status on this?
I think @kenoir and I are happy with it?

kenoir:
**TODO:**

- [ ] Make publication into radio-button style rather than select
- [ ] Group photographer options 
- [ ] Order dropdowns alphabetically for usage-rights panel
- [ ] Investigate exposing multiple categories when selected
- [ ] Fix updating search result status on batch save

kenoir:
IT IS COMPLETE :ship: :ship: :shipit: :cake:

jamesgorrie:
:bike::house: IT.

theefer:
Put it on TEST for last review?

kenoir:
Will put this in test and try and :shipit: tomorrow.

PR #923, opened 2015-07-09
add click tracking

jamesgorrie:
We should be tracking flows - but this is always a simple sense check on whether things on the system are being used.


theefer:
:+1: 

Would be lovely to add more throughout soon.

jamesgorrie:
Yip - this was just to get the tracker in.
I'll be having a day of it soon.

PR #924, opened 2015-07-09
add tracking to file uploader

jamesgorrie:
* added tracking
* made it more in our component style
* removed redundant event hijacking directive as it was silly :facepunch:.

The tracking is of `Upload action` with `Drag enter`, `Button click`, and `Drop` which then has properties of `Files`, `Witness` or `Invalid`.

### DIFF
--- a/kahuna/public/js/analytics/track.js
+++ b/kahuna/public/js/analytics/track.js
@@ -103,13 +103,13 @@ track.run(['$rootScope', '$window', 'mixpanelToken', 'mixpanel', 'trackingServic
 
 }]);
 
-track.directive('grTrackClick', ['$parse', 'track', function($parse, track) {
+track.directive('grTrackClick', ['track', function(track) {
 
     return {
         restrict: 'A',
         link: function(scope, element, attrs) {
             const name = attrs.grTrackClick;
-            const data = $parse(attrs.grTrackClickData)(scope);
+            const data = scope.$eval(attrs.grTrackClickData);
 
             element.on('click', () => track.action(name, data));
         }
--- a/kahuna/public/js/analytics/track.js
+++ b/kahuna/public/js/analytics/track.js
@@ -103,13 +103,17 @@ track.run(['$rootScope', '$window', 'mixpanelToken', 'mixpanel', 'trackingServic
 
 }]);
 
-track.directive('grTrackClick', ['track', function(track) {
+track.directive('grTrackClick', ['$parse', 'track', function($parse, track) {
 
     return {
         restrict: 'A',
         link: function(scope, element, attrs) {
             const name = attrs.grTrackClick;
-            element.on('click', () => track.action(name));
+            const data = $parse(attrs.grTrackClickData)(scope);
+
+            element.on('click', () => {
+                track.action(name, data);
+            });
         }
     };
 
--- a/kahuna/public/js/analytics/track.js
+++ b/kahuna/public/js/analytics/track.js
@@ -111,9 +111,7 @@ track.directive('grTrackClick', ['$parse', 'track', function($parse, track) {
             const name = attrs.grTrackClick;
             const data = $parse(attrs.grTrackClickData)(scope);
 
-            element.on('click', () => {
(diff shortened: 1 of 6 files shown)


kahuna/public/js/upload/file-uploader.html
```diff
@@ -1,17 +1,14 @@
...
+           multiple
+           gr:file-change="ctrl.uploadFiles" />
 
-        <button class="file-uploader__select-files button">
-            <i class="material-icons">file_upload</i>
-            <span class="file-uploader__select-files--label">upload</span>
-        </button>
-    </div>
+    <button class="file-uploader__select-files button"
+            gr:track-click="Upload button">
```
theefer: Are we not already tracking image uploads in other ways? Is this trying to capture something else?

Could we have track-click send which page (e.g. search vs uploads etc) the event was triggered from?

jamesgorrie: Yea, it's not to see number of uploads - but rather if the button is being used.
We could then put tracking on the dragndrop bit and compare.

Mixpanel adds the URL automatically.

theefer: Ah OK, yes would be interesting to measure use of Button (on search vs upload) and drag and drop (on search vs upload).

When adding dnd tracking, can we please have a differentiator between files vs Witness URLs?

jamesgorrie: What value would that add? Can't we just search for Witness images in the Grid and we know how many were done?

jamesgorrie: Either way - that's been added.
Seems to make sense to release it all at the same time.

theefer: Nicer to be able to compare data at the same place if it's easy to do.

jamesgorrie:
Added a few things to make this more fully featured.

kahuna/public/js/analytics/track.js
```diff
@@ -103,13 +103,15 @@ track.run(['$rootScope', '$window', 'mixpanelToken', 'mixpanel', 'trackingServic
...
 
-track.directive('grTrackClick', ['track', function(track) {
+track.directive('grTrackClick', ['$parse', 'track', function($parse, track) {
 
     return {
         restrict: 'A',
         link: function(scope, element, attrs) {
             const name = attrs.grTrackClick;
-            element.on('click', () => track.action(name));
+            const data = $parse(attrs.grTrackClickData)(scope);
```
theefer: Out of curiosity, is there a difference between that and `scope.$eval(attrs.grTrackClickData)`?

kahuna/public/js/directives/gr-file-change.js
```diff
@@ -0,0 +1,15 @@
...
+
+export var fileChange = angular.module('gr.fileChange', []);
+
+fileChange.directive('grFileChange', ['$parse', function($parse) {
+    return {
+        restrict: 'A',
+        link: function(scope, element, attrs) {
+            element.on('change', function() {
+                const onchange = $parse(attrs.grFileChange)(scope);
+                onchange(Array.from(element[0].files));
```
theefer: One pattern I've noticed in other Angular libs is rather than passing "function pointers" to directives, they expose data to expressions and evaluate the expressions.

e.g. instead of 
``` html
<input gr-file-change="doSomething">
```

do

``` html
<input gr-file-change="doSomething($files)">
```

by invoking

``` js
scope.$eval(attrs.grFileChange, {$files: Array.from(element[0].files)})
```

Thoughts?

theefer:
:+1:

kahuna/public/js/directives/gr-file-change.js
jamesgorrie: I prefer this - we've done it else where. I wrote this when I first did Angular - apologies for the gumpf.

It makes it more explicit in the view, thus controller, what you should be expecting.

Changed.

theefer: Cool

Issue #902, opened 2015-07-03
Empty POSTs to /labels when uploading.


akash1810:
This lovely error has been seen a few times in the last few days, causing a few alarms too.

```
2015-07-03 09:38:53,822 - [ERROR] - from play in play-akka.actor.default-dispatcher-55788
Cannot invoke the action, eventually got an error: com.amazonaws.AmazonServiceException: ExpressionAttributeValues contains invalid value: One or more parameter values were invalid: An string set  may not be empty for key :value (Service: AmazonDynamoDBv2; Status Code: 400; Error Code: ValidationException; Request ID: 4E39J2LRLHJFVCPULHBPNI0IEJVV4KQNSO5AEMVJF66Q9ASUAAJG)

2015-07-03 09:38:53,823 - [ERROR] - from application in play-akka.actor.default-dispatcher-55788


! @6mk69n2cn - Internal server error, for (POST) [/metadata/524e7ba22ba5eaa06cf81380dbd73bd60a77c88b/labels] ->

play.api.Application$$anon$1: Execution exception[[AmazonServiceException: ExpressionAttributeValues contains invalid value: One or more parameter values were invalid: An string set  may not be empty for key :value (Service: AmazonDynamoDBv2; Status Code: 400; Error Code: ValidationException; Request ID: 4E39J2LRLHJFVCPULHBPNI0IEJVV4KQNSO5AEMVJF66Q9ASUAAJG)]]
	at play.api.Application$class.handleError(Application.scala:296) ~[app.jar:0.1]
	at play.api.DefaultApplication.handleError(Application.scala:402) [app.jar:0.1]
	at play.core.server.netty.PlayDefaultUpstreamHandler$$anonfun$3$$anonfun$applyOrElse$4.apply(PlayDefaultUpstreamHandler.scala:320) [app.jar:0.1]
	at play.core.server.netty.PlayDefaultUpstreamHandler$$anonfun$3$$anonfun$applyOrElse$4.apply(PlayDefaultUpstreamHandler.scala:320) [app.jar:0.1]
	at scala.Option.map(Option.scala:146) [app.jar:0.1]
	at play.core.server.netty.PlayDefaultUpstreamHandler$$anonfun$3.applyOrElse(PlayDefaultUpstreamHandler.scala:320) [app.jar:0.1]
	at play.core.server.netty.PlayDefaultUpstreamHandler$$anonfun$3.applyOrElse(PlayDefaultUpstreamHandler.scala:316) [app.jar:0.1]
[...]

PR #925, opened 2015-07-09
Use ctrl syntax and explicit filtering of empty labels.

akash1810:
Helps address #902 from the client-side.

### DIFF
--- a/kahuna/public/js/services/label.js
+++ b/kahuna/public/js/services/label.js
@@ -13,16 +13,20 @@ labelService.factory('labelService', ['$rootScope', '$q', function ($rootScope, 
                 .then(newLabels => {
                     image.data.userMetadata.data.labels = newLabels;
                     $rootScope.$emit('image-updated', image, image);
+                    return image;
                 });
         }
     }
 
     function add (image, labels) {
+        labels = labels.filter(label => label && label.trim().length > 0);
+
         return image.data.userMetadata.data.labels
             .post({data: labels})
             .then(newLabels => {
                 image.data.userMetadata.data.labels = newLabels;
                 $rootScope.$emit('image-updated', image, image);
+                return image;
             });
     }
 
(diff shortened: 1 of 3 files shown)


kahuna/public/js/services/label.js
```diff
@@ -13,16 +13,20 @@ labelService.factory('labelService', ['$rootScope', '$q', function ($rootScope,
                 .then(newLabels => {
                     image.data.userMetadata.data.labels = newLabels;
                     $rootScope.$emit('image-updated', image, image);
+                    return image;
                 });
         }
     }
 
     function add (image, labels) {
+        labels = labels.filter(label => label.trim().length > 0);
```
akash1810: Filter out all empty strings (don't trust anything...)

jamesgorrie:
Does this make sense - what if I want to apply none to all?

kahuna/public/js/services/label.js
jamesgorrie: You could filter on `!` which would remove `undefined`s etc

akash1810:
You mean remove all labels? Why would you want to here? When you upload, there are no labels. So presumably, if you've applied a label(s) you did so on purpose?

jamesgorrie:
> did so on purpose

Or hit apply to all by accident. Every action should have some form of redo - apply none to all would be the antithesis of apply to all.

akash1810:
> apply none to all would be the antithesis of apply to all.

I was planning on replicating the behaviour of the batch selection labeller:

screen shot 2015-07-10 at 10 47 53

So you could apply/remove a label on a per label basis. Thoughts?

theefer:
Trying to unify the batch systems in the future makes sense.

jamesgorrie:
Could we move your label stuff over instead of this?

akash1810:
Just checked - apply none to all doesn't work atm...

> Could we move your label stuff over instead of this?

Yep - will confirm the UX with @alastairjardine beforehand though.

Will change this PR to be just about the filter.

theefer:
:+1:

PR #926, opened 2015-07-10
Uploader from image panel

jamesgorrie:
There seems to be quite a few niggles with "my uploads" especially around our sticky states.
I'll look into it a little more.

This makes way more sense too as there is feedback on who is the uploader, whereas the URL would be the only indicator of that before.

Sorry for the icon change too.
Screengrab to prove it works.
crop-ico


theefer:
:+1: let Jonny know!

PR #929, opened 2015-07-10
Update thrall health check.

akash1810:
Thrall is deemed healthy if it has processed a message at least `sqs.message.min.frequency` minutes before now.

This is better than relying on just the receipt of a heartbeat message as it decreases the chances of scenario where one instance is consuming all the heartbeat messages.


theefer:
:+1: but beware of late Friday releases :)

Issue #672, opened 2015-04-26
[UI] Minor date picker pickles


paperboyo:
Hello,

These are super-minor things (maybe except no. 1), because it’s Sunday evening:
1. [see below, or don't] ~~Hmm. Not only Sunday was first day of a week (fixed now), also days seem to start and end an hour earlier wherever the picker comes from :wink: - here is an example: “04-Apr-2015 to 19-Apr-2015” reads `since=2015-04-03T23:00:00.000Z&until=2015-04-19T22:59:59.999Z` in the URL~~
2. Because of different number of days in different months, the layout is jumpy, like so:
   ![grid picker jump](https://cloud.githubusercontent.com/assets/6032869/7339353/73168aba-ec63-11e4-91dc-2c4f88ac7acb.gif)
3. Picker’s months and days use `user agent stylesheet` font definition resulting in e.g. Arial on a PC instead of Grid’s exclusive Guardian Agate Sans.
4. Tooltips and change of cursors over abbreviations of days of the week seem excessive :-)
5. I still think the inside of the picker should use darker tones like the rest of the UI. All the colours should be specified and used consistently throughout Grid.
6. Now, this may seem even more absurd of an issue and is of even less practical importance, but I kind of like how most of Grid survives not only zooming in/out, but also text zooming in/out. This is how I personally quickly judge the quality of layouts on the web. I may have some ideas on why that may matter in the future, but I also think that if layout survives both these zooms, its less prone to breaking and more maintainable itself (I may be totally wrong here, though). Take a look - text zooming in:
   ![grid zooming picker](https://cloud.githubusercontent.com/assets/6032869/7339465/23601022-ec68-11e4-812c-8bc419912a8c.gif)
7. I think, consistently with how there is no border around ▸advanced next to search (which is, effectively, a dropdown), there shouldn’t be one around date picker button. Or they both should have it (not). Also, the arrow is on the left of ▸advanced, but on the right of e.g. anytime▸ (here I would go for the right, but mostly I’m for consistency - which should not suggest I’m a centrist, though!).

[...]

PR #932, opened 2015-07-14
Improve top bar usability

kenoir:
Image page:
image

Search page:
search

Crop page:
crop


kahuna/public/js/components/gr-top-bar/gr-top-bar.js
```diff
@@ -22,8 +22,7 @@ topBar.directive('grTopBarNav', [function() {
         transclude: true,
         // Annoying to have to hardcode root route here, but only
         // way I found to clear $stateParams from uiRouter...
-        template: `<a href="/search" class="home-link">Home</a>
-                   <ng:transclude></ng:transclude>`
+        template: `<ng:transclude></ng:transclude>`
```
theefer: Why remove the global link to home?

theefer: (And clue of what system you are in)

theefer:
Crop dimensions aren't interactive so shouldn't change background on hover?

image


Archive button area goes light grey on hover but the whole area isn't clickable.

image

theefer:
Grid logo should be transparent to accommodate background colour change

image

theefer:
Separator height issue on search screen:

image

theefer:
Weird Upload button hover:

image

theefer:
Upload button on upload screen:

image

kahuna/public/js/components/gr-top-bar/gr-top-bar.js
kenoir: Moved to the html template rather than in js.

See `kahuna/public/js/search/view.html:3`

theefer:
Same issue with crop button on crop screen:

image

theefer:
Apart from these :+1:

kahuna/public/js/components/gr-top-bar/gr-top-bar.js
theefer: It's meant to be on every screen though, not just on the search page.

kenoir:
> Crop dimensions aren't interactive so shouldn't change background on hover?

Fixed.

> Archive button area goes light grey on hover but the whole area isn't clickable.

Won't fix here. @alastairjardine is going to take a look.

kenoir:
> Separator height issue on search screen
> Weird Upload button hover
> Upload button on upload screen
> Same issue with crop button on crop screen

Should all be fixed.

Clickable area issue remains.

kahuna/public/js/crop/view.html
```diff
@@ -39,7 +39,7 @@
         </div>
 
         <div class="top-bar-item">
-            <button class="button"
+            <button class="button crop__start-crop"
```
akash1810: pedantic point: `crop__state-crop` doesn't reflect all states of the button - the text of the button becomes `cropping` when a crop is in process.

alastairjardine:
grid-logo-32

Transparent backgrounded icon

kenoir:
@alastairjardine top bar looked particular weird at shorter widths due to the lines, made it a bit more responsiv
e.

foo3

kenoir:
> Grid logo should be transparent to accommodate background colour change
Fixed

Also fixed logo appearing correctly on all pages.

kenoir:
Suggest we fix the archiver button click width in another PR.

@akash1810 review?

kahuna/public/stylesheets/main.css
```diff
@@ -418,6 +440,61 @@ textarea.ng-invalid {
...
+        display: none;
+    }
+}
+@media screen and (max-width: 700px) {
+    input.search-query__input {
+        width: 170px
+    }
+
+}
+@media screen and (max-width: 1054px) {
```
akash1810: How are these widths calculated?

kenoir: They are made up by me looking at where the UI breaks in my browser.

akash1810:
Looks like we hide the `advanced` drop down at 614px - given its how to discover the search syntax, should we hide it at all?

akash1810:
Loving the additional filter list. Should we update the search syntax to make the state of the checkboxes clear when they're hidden? For example, `is:free` `is:archived`.

kenoir:
@akash1810 
> Looks like we hide the advanced drop down at 614px - given its how to discover the search syntax, should we hide it at all?

Not sure, it's a compromise between UI space and functionality. I'd argue that viewing the search syntax is not essential at smaller widths.

> Loving the additional filter list. Should we update the search syntax to make the state of the checkboxes clear when they're hidden? For example, is:free is:archived.

Yes, I think the search box should probably represent the current filter state and flagging up that a filter is on at smaller widths is probably useful. I'm not sure it needs to be done in this PR as we're not losing functionality at full width, just making compromises at smaller widths that currently have a broken UI.

kahuna/public/stylesheets/main.css
akash1810: The crop page needs updating too?

screen shot 2015-07-15 at 10 52 19

akash1810: > They are made up by me looking at where the UI breaks in my browser.

Should we use mixpanel stats to find the most common size?

paperboyo:
> I think the search box should probably represent the current filter state

That is an interesting and important point. Worth making it right. Definitely arguments in URL should _always_ be in sync with the UI (your uploads checkbox is broken in this respect) and should contain _all_ active filters/searches (unlike some other tools :wink:).

The question of should the search box be the same i.e. contain all active searches/filters is separate. It can become quite crowded (to counter that maybe everything but the "proper" query could be dimmed and at the end of the search box?). But maybe stuffing all that text into a search box isn't such a good idea (more so, because it's there in the URL to play with). Maybe there could be something suggesting that there are filters active  on narrower BPs? Workflow-style bar at the left of "search filters" "button" or half-Workflow-style at the bottom, not on the left?

Just thinking out loud... This is looking great! Let @paperboyo play on TEST...

kahuna/public/stylesheets/main.css
kenoir: @akash1810 will take a look at the crop page.

> Should we use mixpanel stats to find the most common size? 

Probably :)

alastairjardine:
Cracking effort @kenoir – thank you so much for filling in the blanks. I'd like to get this out, even with the inconsistent behaviour on the archive button.

alastairjardine:
> I think the search box should probably represent the current filter state

I'm going to propose we have a Workflow-esque filter panel rather than search. Advanced search is great for advanced users, but many will not have the skills to pick up the syntax in occasional/few times a day usage. Start the debate…here!

kenoir:
@alastairjardine going to put this on TEST for folks to have a look at, and it probably needs another code review as i've added some more media query rules.

kenoir:
@alastairjardine I think if the complexity of the search parameters regular users will be using increases it makes sense to break them out into a panel like in Workflow. I think the discussion probably requires a new issue.

kenoir:
@paperboyo @alastairjardine this branch is now on `TEST`. Please have a look and see what you think.

akash1810:
meep

dance

kenoir:
@akash1810 totally a feature ...

jamesgorrie:
Bugs:
- [ ] Upload button bounces when clicked
- [ ] Calendar is a bit weird
- [ ] Search on uploads needs ico
- [ ] Archive button isn't completely clickable

jamesgorrie:
Could we get rid of recrop while we're here?

akash1810:
Could we also set the breakpoint for hiding text on buttons to be the same for all buttons - it looks strange when some buttons have icon + text but others have icon only.

akash1810:
other than the above, looks great!!!

kenoir:
@jamesgorrie 
> Archive button isn't completely clickable

Going to leave this for later.

kenoir:
Should have addressed all the issues now. 

Can I get some thumbs?

kahuna/public/js/search/view.html
```diff
@@ -5,7 +5,9 @@
     </gr-top-bar-nav>
 
     <gr-top-bar-actions>
-        <a class="top-bar-item" ui:sref="upload">your recent uploads</a>
+        <a class="top-bar-item" ui:sref="upload" title="your recent uploads">
+            <i class="material-icons">photo_library</i>
+            <span class="top-bar-item--label-text"> your recent uploads</span></a>
```
akash1810: use padding rather than a leading space

kahuna/public/js/components/gr-top-bar/gr-top-bar.css
```diff
@@ -13,15 +13,19 @@ gr-top-bar-nav {
...
+    min-width: 100px;
+    flex-direction: row;
+    justify-content: flex-end;
 }
 
 
 .gr-top-bar-inner {
     background-color: #333;
-    padding: 10px;
+    padding: 0px 10px 0px 0px;
```
jamesgorrie: ?

kahuna/public/js/crop/view.html
```diff
@@ -1,28 +1,29 @@
...
+        <label class="top-bar-item" title="landscape">
+            <input type="radio" ng:model="ctrl.aspect" value="{{ctrl.landscapeRatio}}"/>
+            <span class="top-bar-item--label-text">landscape </span><i class="material-icons">crop_landscape</i> ({{::ctrl.getRatioString(ctrl.landscapeRatio)}})
         </label>
 
-        <label class="top-bar-item">
+        <label class="top-bar-item" title="portrait">
             <input type="radio" ng:model="ctrl.aspect" value="{{ctrl.portraitRatio}}" />
-            portrait ({{::ctrl.getRatioString(ctrl.portraitRatio)}})
+            <span class="top-bar-item--label-text">portrait </span><i class="material-icons">crop_portrait</i> ({{::ctrl.getRatioString(ctrl.portraitRatio)}})
```
jamesgorrie: If we're using BEM - `top-bar-item__label` or `top-bar-label` would make more sense.

kahuna/public/js/edits/archiver.html
```diff
@@ -6,11 +6,15 @@
     <!-- TODO: Start looking at a new loader to replace our "ing…" loader -->
     <span ng:show="archiver.isArchived" title="unarchive">
         <i class="material-icons">star</i>
-        <span ng:show="archiver.withText">unarchiv<span ng:show="!archiver.archiving">e</span><span ng:show="archiver.archiving">ing…</span></span>
+        <span class="archiver--label-text">
```
jamesgorrie: As with above, perhaps `archiver__label` makes more sense?

kahuna/public/js/image/view.html
```diff
@@ -1,25 +1,31 @@
 <gr-top-bar>
     <gr-top-bar-nav>
         <a class="top-bar-item"
-           ui:sref="search">← search</a>
+           ui:sref="search"><i class="material-icons">search</i>
+           <span class="top-bar-item--label-text">search</span></a>
```
jamesgorrie: [^](https://github.com/guardian/grid/pull/932/files#r34697945)

kahuna/public/stylesheets/main.css
```diff
@@ -418,6 +458,61 @@ textarea.ng-invalid {
     display: inline-block;
 }
 
+.search__filters li {
```
akash1810: @jamesgorrie and @theefer don't like element targeting... i forget why...

kahuna/public/js/search/view.html
jamesgorrie: [^](https://github.com/guardian/grid/pull/932/files#r34697945)

kahuna/public/stylesheets/main.css
```diff
@@ -306,32 +306,70 @@ textarea.ng-invalid {
 /* ==========================================================================
    Top bar
    ========================================================================== */
+@media screen and (max-width: 700px) {
+    .top-bar-item.top-bar-item--surplus {
```
jamesgorrie: The point of BEM is to avoid specificity and the modifiers should work using the cascade of the stylesheet.
i.e. `.top-bar-item--surplus {`.

kahuna/public/stylesheets/main.css
```diff
@@ -306,32 +306,70 @@ textarea.ng-invalid {
...
+        display: none;
+    }
+}
+   
+@media screen and (max-width: 1100px) {
+    .top-bar-item--label-text {
+        display: none;
+    }
+    .top-bar-item .archiver--label-text {
```
jamesgorrie: Do we need two selectors?

kahuna/public/stylesheets/main.css
```diff
@@ -306,32 +306,70 @@ textarea.ng-invalid {
...
 
 .home-link {
     text-indent: -9999px;
-    width: 24px;
-    height: 24px;
+    width: 50px;
+    height: 50px;
     background: url(/assets/images/grid-logo-32.png) no-repeat center;
-    margin-right: 10px;
+    margin-right: 0px;
```
jamesgorrie: Convention is 0, but I won't say anything else till I get a linter in.

kahuna/public/stylesheets/main.css
```diff
@@ -306,32 +306,70 @@ textarea.ng-invalid {
...
+    }
+}
+   
+@media screen and (max-width: 1100px) {
+    .top-bar-item--label-text {
+        display: none;
+    }
+    .top-bar-item .archiver--label-text {
+        display: none;
+    }
```
akash1810: should we combine this rule with the above one? Is `archiver--label-text` used? I can't seem to find it in any of the templates.

kahuna/public/stylesheets/main.css
jamesgorrie: Could we name these instead of using the element? e.g. `.filter-item`?

kahuna/public/stylesheets/main.css
```diff
@@ -418,6 +458,61 @@ textarea.ng-invalid {
...
+        position: absolute;
+        top: 25px;
+        left: 0px;
+        width: 150px;
+        background-color: #333;
+        position: absolute;
+        padding: 10px;
+        box-shadow: 0 1px 5px rgba(0, 0, 0, 0.5);
+    }
+    .search__filters.search__filters-hide {
```
jamesgorrie: This a weird set of classes? Do we mean `.search__filters--hide`?

kahuna/public/stylesheets/main.css
```diff
@@ -519,10 +623,11 @@ input.search-query__input {
 }
 
 .file-uploader__select-files {
+    margin-top: 12px;
     padding: 5px 10px;
 }
 
-@media screen and (max-width: 1200px) {
+@media screen and (max-width: 1100px) {
```
jamesgorrie: Is this based on any of the data from mixpanel?

kahuna/public/js/crop/view.html
```diff
@@ -30,20 +31,25 @@
...
             <div ng:switch-when="conditional"
                  class="cost cost--conditional"
-                 title="{{ctrl.image.data.usageRights.restrictions}}">
+                 title="restricted use: {{ctrl.image.data.usageRights.restrictions}}">
                 <!-- As `conditional` can only be set with usageRights, let's
                      just assume it's here. We might need to revisit this. -->
-                restricted use
-                <span class="ico ico--circled ico--help">?</span>
+                     <i class="material-icons">warning</i><span class="top-bar-item--label-text"> restricted use
+                         <span class="ico ico--circled ico--help">?</span></span>
```
akash1810: Not part of this feature, per se - can we replace this with the [icon](https://www.google.com/design/icons/#ic_help_outline)? We'd be able to remove some css for `ico` then too.

paperboyo:
Hi, sorry don’t have time to properly dig in now, but the calendar(s) will need some love in the future to be properly responsive (placed vertically on narrower BPs, at least). They would gain, maybe, from loosing one of two Clear buttons? There is also some issue plankton in https://github.com/guardian/grid/issues/672.

Also, search explainer can be opened on wide BP, but can’t be closed on narrow BP (not that anybody would do that other than me).

kahuna/public/js/search/query.html
```diff
@@ -24,38 +24,56 @@
...
-                   ng:false-value="true"
-                   ng:checked="{{!searchQuery.filter.nonFree}}" />
-            free to use only
-        </label>
+    <div class="search__filters-container">
+    <button class="search__filters-toggle"
+            type="button"
+            ng:click="filtersShown = !filtersShown ">
+        <span ng:hide="filtersShown">▸</span>
+        <span ng:show="filtersShown">▾</span>
```
akash1810: minor:

```html
<span ng:show="!filtersShown">▸</span>
<span ng:show="filtersShown">▾</span>
```

Is a bit simpler imo, but not fussed.

akash1810:
>  It can become quite crowded (to counter that maybe everything but the "proper" query could be dimmed and at the end of the search box?).

We've been thinking about possibly using [chips](https://material.angularjs.org/latest/#/demo/material.components.chips) as an answer to this (though it hasn't gotten past the thinking stage).

paperboyo:
Earth-shattering contribution: a whooping 20x size reduction!
271fee40-2a33-11e5-9ee9-4a7109f36bd8_opt

And invisible, too!

kahuna/public/js/components/gr-top-bar/gr-top-bar.css
kenoir: Ah, I was using the Chrome console as an editor, so I guess this is an artefact of that.

kahuna/public/stylesheets/main.css
kenoir: No, i'm just picking values at which the UI overflows to break on.

kenoir:
Beautified the CSS and markup as per comments. 

Merg? :ship: :pray:

kenoir:
@paperboyo yup, the calendar needs looking at a bit in another PR I think.

jamesgorrie:
:+1:

paperboyo:
Super-minor: the browser's status bar reads only base URL part when hovering over search-back button while in viewer, like so:

search-back

If possible, it should show full URL, ``https://media.[...]/search?query=Anita%20Bhagwandas`` in the example above.

jamesgorrie:
@paperboyo this is because we'reusing `ui-router`'s sticky states, I can explain it in finer detail if you'd like, but unless we change the lib we're using for this, this will remain.

paperboyo:
Thanks! Definitely not worth changing anything just because of that!

PR #934, opened 2015-07-15
Client image service

jamesgorrie:
After a chat with @akash1810 on Friday we realised it might be useful to have a service that does a few of the logical bits as a service then passes the new data / object to the editor.

This makes the editors a lot more simple and easy to understand.

After talking to @kenoir this might also be useful to make the editors deal with batch editing.

Hard to diff until #919 goes out.

### DIFF
--- a/kahuna/public/js/edits/image-editor.js
+++ b/kahuna/public/js/edits/image-editor.js
@@ -2,7 +2,7 @@ import angular from 'angular';
 import template from './image-editor.html!text';
 
 import './service';
-import './usage-rights-editor';
+import '../usage-rights/usage-rights-editor';
 
 
 export var imageEditor = angular.module('kahuna.edits.imageEditor', [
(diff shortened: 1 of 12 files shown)


kenoir:
:+1:

kahuna/public/js/edits/image-editor.js
```diff
@@ -71,6 +73,8 @@ imageEditor.controller('ImageEditorCtrl',
         ctrl.error = true;
     }
 
+
+    ctrl.usageRights = imageService(ctrl.image).usageRights;
```
akash1810: nice

kahuna/public/js/usage-rights/usage-rights-editor.html
```diff
@@ -0,0 +1,82 @@
...
+                    ng:switch-when="text"
+                    ng:required="property.required"></textarea>
+
+            </div>
+        </label>
+    </div>
+
+    <div class="ure__submit">
+        <button class="button" type="submit"
+            ng:class="{ 'button--green': ctrl.saved }"
```
akash1810: the pedant in me doesn't like this class name - `button--saved` would be better.

PR #936, opened 2015-07-15
Display upload time as relative to now, up to 7 days ago.

akash1810:
screen shot 2015-07-15 at 14 49 43

### DIFF
--- a/kahuna/public/js/preview/image.html
+++ b/kahuna/public/js/preview/image.html
@@ -47,8 +47,15 @@
 
     <div class="preview__bottom-bar bottom-bar">
         <div class="bottom-bar__meta">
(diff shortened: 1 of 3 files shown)


alastairjardine:
Quite a contested space – would it be worth trying to fix the major structural problems instead? Slide from the UI review.

media service ui review 001

If this change is going to stick, could we fix the gestalt issues between the date and the time for the >7 case please?

akash1810:
@alastairjardine made them the same weight:

screen shot 2015-07-15 at 15 23 19

alastairjardine:
:+1: 

Might be worth adding a bit more padding between the crop logo and the time, but otherwise c'est bon.

Issue #787, opened 2015-06-09
Add all categories to `usageRights`


jamesgorrie:
- [x] Subscription agency
- [x] Staff photographer
- [x] Contracted photographer
- [x] Commissioned photographer
- [x] Social media news image
- [x] Guardian Witness
- [x] Handout
- [x] Screengrab
- [x] Supplied for obituary
- [x] Pool / Rota
- [x] PR Image

akash1810:
Looks complete.

PR #938, opened 2015-07-15
Usage rights all categories

jamesgorrie:
[All categories are now there](https://github.com/guardian/grid/issues/787).
This will allow the component to work completely on the panel too.

## Next up:
* Add no-rights
* UI


common-lib/src/main/scala/com/gu/mediaservice/lib/config/UsageRightsConfig.scala
```diff
@@ -0,0 +1,113 @@
...
+    "PA",
+    "Reuters",
+    "Rex Features"
+  )
+
+  val suppliersCollectionExcl = Map(
+    "Getty Images" -> payGettySourceList
+  )
+
+}
```
jamesgorrie: This is now common as we use it in the API and Edits.

jamesgorrie:
:camel: As no longer WIP

common-lib/src/main/scala/com/gu/mediaservice/lib/config/UsageRightsConfig.scala
```diff
@@ -0,0 +1,113 @@
+package com.gu.mediaservice.lib.config
+
+object UsageRightsConfig {
+
+  // Note: we filter exclusively on matching source, rather than combining credit=Getty and source=X
+  // this is assumed to be good enough as it's unlikely other provider will use the same source.
+  val payGettySourceList = List(
+    "ASAblanca",
```
akash1810: Should we put this list in an external file?

jamesgorrie: This has been in Config since the start.
We could do - but what would probably be best is to have it in a store, as, eventually, this will become editable.

akash1810:
:+1:

PR #940, opened 2015-07-16
Add retained to json response.

akash1810:
Add `retained` to signify if:

- The image is from picdar (by presence of picdarUrn identifier)
- The image has at least one crop
- The image has been manually archived by user.

TODO:
- [ ] Update UI to signify the value of `retained`.
- [ ] Expose in search end-point and simplify reaper.

### DIFF
--- a/media-api/app/lib/ImageResponse.scala
+++ b/media-api/app/lib/ImageResponse.scala
@@ -48,7 +48,7 @@ object ImageResponse {
 
     val valid = ImageExtras.isValid(source \ "metadata")
 
-    val isRetained = image.identifiers.contains(Config.persistenceIdentifier) ||
+    val isPersisted = image.identifiers.contains(Config.persistenceIdentifier) ||
       image.exports.length > 0 ||
       image.userMetadata.exists(_.archived)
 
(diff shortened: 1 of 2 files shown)


media-api/app/lib/ImageResponse.scala
```diff
@@ -48,22 +48,29 @@ object ImageResponse {
 
     val valid = ImageExtras.isValid(source \ "metadata")
 
+    val isAutoRetained = image.identifiers.contains(Config.persistenceIdentifier) || image.exports.length > 0
```
jamesgorrie: Why not just keep this in one val? It's a little confusing.
Auto is an odd name as cropping isn't really auto - it's similar to archive?

Or potentially use `isArchived` but I would argue `image.userMetadata.exists(_.archived)` is clear enough?

media-api/app/lib/ImageResponse.scala
```diff
@@ -48,22 +48,29 @@ object ImageResponse {
...
-  def imageLinks(id: String, secureUrl: String, withWritePermission: Boolean, valid: Boolean) = {
+  def imageLinks(id: String, secureUrl: String, withWritePermission: Boolean, valid: Boolean, isAutoRetained: Boolean) = {
+    val metadataUri = s"${Config.metadataUri}/metadata/$id"
+
     val cropLink = Link("crops", s"${Config.cropperUri}/crops/$id")
-    val editLink = Link("edits", s"${Config.metadataUri}/metadata/$id")
+    val editLink = Link("edits", metadataUri)
     val optimisedLink = Link("optimised", makeImgopsUri(new URI(secureUrl)))
     val imageLink = Link("ui:image",  s"${Config.kahunaUri}/images/$id")
+    val archiveLink = Link("archived", s"$metadataUri/archived")
```
jamesgorrie: What is this for?

jamesgorrie:
:+1:

theefer:
:bikeshed: we talk about persistence elsewhere: retained -> persisted?

theefer:
:+1: either way

akash1810:
Changed to `persisted` for consistency.

PR #941, opened 2015-07-16
`gr-icon` directive.

akash1810:
Typing `<i class="material-icons">icon</i>` is long...

TODO
- [ ] icons everywhere - looking at you ▸ and ▾

### DIFF
--- a/kahuna/public/js/components/gr-icon/gr-icon.css
+++ b/kahuna/public/js/components/gr-icon/gr-icon.css
@@ -1,7 +1,6 @@
-.gr-icon {
+gr-icon {
     font-family: 'Material Icons';
     font-weight: normal;
-    font-style: normal;
     font-size: 16px;  /* Preferred icon size */
     display: inline-block;
     width: 1em;
@@ -22,6 +21,10 @@
     vertical-align: middle;
 }
 
-.gr-icon--small {
+gr-icon .gr-icon { /* avoid targeting i element */
+    font-style: normal;
+}
+
+gr-icon .gr-icon--small {
     font-size: 1.2rem;
 }
--- a/kahuna/public/js/components/gr-icon/gr-icon.css
+++ b/kahuna/public/js/components/gr-icon/gr-icon.css
@@ -1,0 +1,27 @@
+.gr-icon {
+    font-family: 'Material Icons';
+    font-weight: normal;
+    font-style: normal;
+    font-size: 16px;  /* Preferred icon size */
+    display: inline-block;
+    width: 1em;
+    height: 1em;
+    line-height: 1;
+    text-transform: none;
+    letter-spacing: normal;
+    word-wrap: normal;
+
+    /* Support for all WebKit browsers. */
+    -webkit-font-smoothing: antialiased;
+    /* Support for Safari and Chrome. */
+    text-rendering: optimizeLegibility;
+
+    /* Support for Firefox. */
+    -moz-osx-font-smoothing: grayscale;
+
+    vertical-align: middle;
+}
+
+.gr-icon--small {
+    font-size: 1.2rem;
+}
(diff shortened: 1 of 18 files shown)


kahuna/public/js/preview/image.html
```diff
@@ -61,7 +61,7 @@
                   title="this image has crops"
                   ng:if="image | hasExportsOfType:'crop'">
 
-                <i class="material-icons material-icons--small">crop</i>
+                <gr-icon small="true">crop</gr-icon>
```
akash1810: couldn't work out a way to use html5 attributes, i.e. `<gr-icon small>crop</gr-icon>`

kahuna/public/js/components/gu-date-range/gu-date-range.html
```diff
@@ -1,7 +1,7 @@
 <div class="gu-date-range" ng:init="ctrl.showOverlay = false">
     <div class="gu-date-range__display" ng:click="ctrl.showOverlay = !ctrl.showOverlay">
         <div class="gu-date-range__display__icon gu-date-range__display__icon__calendar">
-            &#128197;
+            <gr-icon>event</gr-icon>
```
akash1810: bye bye emoji date icon!

kahuna/public/js/components/gu-date-range/gu-date-range.html
```diff
@@ -1,9 +1,6 @@
 <div class="gu-date-range" ng:init="ctrl.showOverlay = false">
     <div class="gu-date-range__display" ng:click="ctrl.showOverlay = !ctrl.showOverlay">
-        <div class="gu-date-range__display__icon gu-date-range__display__icon__calendar">
-            &#128197;
-        </div>
-
+        <gr-icon>event</gr-icon>
```
akash1810: bye bye calendar emoji icon

kahuna/public/js/components/gr-icon/gr-icon.css
```diff
@@ -0,0 +1,27 @@
+.gr-icon {
```
jamesgorrie: Could we not just style the `gr-icon` element.... @theefer was using that as a pattern elsewhere and it's quite nice.

jamesgorrie:
Have you included this in all the component where it's used?

akash1810:
Yep - cheated and [put it in the root](https://github.com/guardian/grid/pull/941/files#diff-a727a260b31fe8cd84ba0780a0ee0370R62) so its only required once.

jamesgorrie:
Yea - I was wondering if that's the right thing to do?
That way if a component is removed from the root scope, it'll break - do you think it adds to much gumpf to add to each directive?

kahuna/public/js/components/gr-icon/gr-icon.css
akash1810: Could do, would mean we'd also have to have

```css
gr-icon i {
    font-style: normal;
}
```

To override the default browser styling of `<i>`.

kahuna/public/js/preview/image.html
jamesgorrie: You might want to call the attr gr:small too?

jamesgorrie: You could remove the scope and have the link:
```JavaScript
{
  link: function(scope, elem, attrs) { scope.small = angular.exists(attrs.grSmall) }
}
```

akash1810:
> do you think it adds to much gumpf to add to each directive?

Yeah. You'd end up doing `import 'components/gr-icon/gr-icon';` and `angular.module('myModule', ['grIcon']);` everywhere you want to icon, which is more cumbersome than just `<i class="material-icons">icon</i>` in a template.

kahuna/public/js/components/gr-icon/gr-icon.css
akash1810: Done. Targeting `gr-icon .gr-icon` to avoid `gr-icon i`.

jamesgorrie:
:+1: it'd be interesting to chat about BEM and the structure of our CSS, as in a componentalised world I'm not sure it makes the most sense.

kenoir:
:+1: nice

akash1810:
> it'd be interesting to chat about BEM and the structure of our CSS, as in a componentalised world I'm not sure it makes the most sense.

Yeah. I guess it depends on the size of the components; though you could argue that large components should be split up into smaller ones...

PR #942, opened 2015-07-16
Leg work for no rights

jamesgorrie:
I was hoping to have the no-rights added for this.
Unfortunately not as to add the filter to Elastic is too hard / annoying (you can't search for empty maps).

The next steps, is to write the parsers and reindex, that will kill this pain.

This does fix a deletion bug that comes from Theseus that I will fix there when I have more time.

Below shows:
* Image uploaded with Corbis image, no usage rights overrides
* Override with Handout and restriction
* Delete override, image is reverted to original usage rights

delete-usage-rights


kahuna/public/js/edits/service.js
```diff
@@ -51,10 +57,14 @@ service.factory('editsService',
      */
     function isEmpty(edit, image) {
         // find that matching resource
-        return findMatchingEditInImage(edit, image).then(matchingEdit =>
-            angular.equals(matchingEdit.data, {}) || angular.equals(matchingEdit.data, []) ?
+        return findMatchingEditInImage(edit, image).then(matchingEdit => {
+            console.log(matchingEdit.data);
```
akash1810: meep

jamesgorrie: Sorry - just ran the lint and removed a few.

kahuna/public/js/usage-rights/usage-rights-editor.js
```diff
@@ -88,6 +89,7 @@ usageRightsEditor.controller('UsageRightsEditorCtrl',
     }
 
     function save(data) {
+        ctrl.error = null;
```
akash1810: What's this for?

theefer:
> you can't search for empty maps

I'm lacking the context, but can't you search for missing category prop?

jamesgorrie:
@theefer You can't distinguish between `{ "prop": null }`, `{ "prop": [null] }`, `{ "prop": [] }` and `{ "prop": {} }`.

This is a very useful feature - we could add a [`null_value`](https://www.elastic.co/guide/en/elasticsearch/reference/1.6/query-dsl-exists-filter.html#_literal_null_value_literal_mapping) to the mapping, but I feel that would be bastardising the mappings to do a hack, whereas it is probably more useful to, later down the line, search for `missing`: `usageRights`, which the current mapping allows.

kahuna/public/js/usage-rights/usage-rights-editor.js
jamesgorrie: We receive errors from the server, which are rendered [here](https://github.com/guardian/grid/blob/master/kahuna/public/js/usage-rights/usage-rights-editor.html#L80).

jamesgorrie:
@theefer there's a good article on it [here](https://www.elastic.co/guide/en/elasticsearch/guide/current/_dealing_with_null_values.html). I've chatted to Phil too and he can't think of a non-polluting way around it.

kahuna/public/js/image/service.js
```diff
@@ -19,10 +19,13 @@ imageService.factory('imageService', ['editsService', function(editsService) {
...
         const resource = image.data.userMetadata.data.usageRights;
 
-        const save = newData => editsService.update(resource, newData, image);
-        const remove = () => editsService.remove(resource, image);
+        const save = newData =>
+            editsService.update(resource, newData, image).then(resource => resource.data);
 
-        return { data, resource, image, save, remove };
+        const remove = () =>
+            editsService.remove(resource, image).then((/*resoure*/) => image.data.usageRights);
```
akash1810: why `/*resource*/`?

kahuna/public/js/usage-rights/usage-rights-editor.html
```diff
@@ -77,6 +76,10 @@
                 <span ng:if="ctrl.saving">Saving</span>
                 <span ng:if="ctrl.saved">Saved</span>
         </button>
+        <button class="ico-button ure__remove" type="button" ng:click="ctrl.remove()"
+            title="remove usage rights overrides">
+            <i class="material-icons">delete</i>
```
akash1810: before merging to master, this'll need to become `<gr-icon>delete</gr-icon>` - `.material-icons` is dead now.

kahuna/public/js/image/service.js
jamesgorrie: I've put it there so you know that it's there. I really don't like the fact that jslint make you remove unused parameters, as they are a true representation of what's there (slightly closer to å  typed system).
Does it feel noisy?

akash1810: > Does it feel noisy?

A little, yeah.

jamesgorrie:
@akash1810 - new confirm delete:
saveur

akash1810:
I'd maybe have delete on the LHS.

:+1:

jamesgorrie:
> I'd maybe have delete on the LHS.

I did that, and it sort of looked like a main action when floating on it's own.

PR #943, opened 2015-07-17
Change label text hide breakpoint

kenoir:
Changes the label text breakpoint to a smaller width.

This changes accommodates the embedded version of the grid in Composer:

crop


akash1810:
:+1:

PR #944, opened 2015-07-17
Usage rights: Order select boxes (e.g. photographer name)

kenoir:
It should be ordered rather than random

screen shot 2015-07-17 at 10 09 57

### DIFF
--- a/kahuna/public/js/usage-rights/usage-rights-editor.html
+++ b/kahuna/public/js/usage-rights/usage-rights-editor.html
@@ -42,7 +42,7 @@
                         ng:if="property.options"
                         ng:model="ctrl.model[property.name]"
                         ng:required="property.required"
-                        ng:options="o for o in property.options">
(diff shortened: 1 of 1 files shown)


kenoir:
crop

akash1810:
Should we order by surname?

kenoir:
@akash1810 Might be impossible given the underlying implementation.

paperboyo:
> Should we order by surname?

There aren't (will never be) that many... although "everybody is a photographer now" :wink: Also, new Editorial Users server sorts __all__ users by first name, haha.

akash1810:
> Might be impossible really annoying to implement given the underlying implementation.

Could do some funky split on ' ' then move the last index to the first and sort... but horrible.

:+1:

jamesgorrie:
:-1: Bad implementation - we're looking at shifting all logic to the server.

kenoir:
Meh :crying_cat_face:

jamesgorrie:
not fussed :+1:

PR #951, opened 2015-07-17
order uploads by filename

jamesgorrie:
Makes sense as I have seen more than just @paperboyo use filenames for order.


akash1810:
:+1:

paperboyo:
Thx!

PR #952, opened 2015-07-17
icons on upload page thumbnails.

akash1810:
Missed a couple in https://github.com/guardian/grid/pull/947

screen shot 2015-07-17 at 16 09 54

Will componentise when I have more time.

PR #953, opened 2015-07-20
show image actions on upload

jamesgorrie:
We merged a bug in #952.
The CSS is starting to get very couply and unruly.
I'd like to take a look at it before it becomes more so.

Perhaps @theefer's suggestion of GSS?

### DIFF
--- a/kahuna/public/js/edits/image-editor.html
+++ b/kahuna/public/js/edits/image-editor.html
@@ -22,7 +22,8 @@
                  gr:image-fade-on-load
              />
 
-            <ul class="image-actions">
+            <ul class="image-actions"
+                ng:if="ctrl.image.data.valid">
                 <li>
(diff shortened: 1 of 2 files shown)


akash1810:
:+1:

theefer:
re GSS, I still like the idea of it for layout and it'd be interesting to prototype, although the project didn't seem to get the traction (and regular updates) I would have hoped, so might be a bit of a dead-end.

However GSS wouldn't necessarily solve the problems we have with CSS around modularity and composability. For these we may want to look at Sass (I've started working on a [Sass plugin for JSPM](https://github.com/theefer/plugin-sass), which works but is slightly blocked by JSPM/System bugs) or [Interoperable CSS](https://github.com/css-modules/icss) (see [intro blog post](http://glenmaddern.com/articles/interoperable-css)).

PR #954, opened 2015-07-20
Update main.css

alastairjardine:
Removing dotted line to reduce visual interference
**Before**
screen shot 2015-07-20 at 11 36 12

**After**
screen shot 2015-07-20 at 11 35 55

### DIFF
--- a/kahuna/public/stylesheets/main.css
+++ b/kahuna/public/stylesheets/main.css
@@ -905,7 +905,7 @@ FIXME: what to do with touch devices
     width: 24px;
     color: #999;
     display: block;
-    border-top: 1px dashed rgba(255, 255, 255, .5);
+    border-top: 1px solid rgba(255, 255, 255, .3);
 }
 .image-action--first {
     border-top: 0;


theefer:
Seems innocuous enough but where is this (screenshot?)?

alastairjardine:
@theefer Updated main comment above.

theefer:
:+1: 

(or they could even be transparent (margin instead of border)? either way removing the dashed line makes sense imo too)

alastairjardine:
@theefer Yeah, it's just a bit cleaner. Unsure what you mean about border/margin?

theefer:
I meant: should the separator be an actual coloured line, or should it just be a transparent gap? Just a thought, I don't particularly mind.

alastairjardine:
Happy with just transparency for the sake of speed. Perhaps one for a refactor in the future (as @akash1810 & I have spoken about)

akash1810:
If we're striving for consistency, should we also change the dashed separator on the uploads page? If so, transparent wouldn't work.

alastairjardine:
Yep, but I thought that was pretty high up the agenda for rejiggery anyway?

akash1810:
True, true.

theefer:
Actually while the gap works on lighter background:

screenshot-2015-07-20t14 35 40

It's invisible on darker background:

screenshot-2015-07-20t14 35 32

So let's just ship @alastairjardine's version with the border for now, works more generally.

PR #955, opened 2015-07-20
show no usage rights only to images that initially don't have any

jamesgorrie:
We only show no rights on images that don't have them to start with to stop the confusion of the interface defaulting to say PR Image.

This will disappear when we have reindexed and allow the override of no-rights.


theefer:
:+1:

PR #956, opened 2015-07-20
set the category in the right place

jamesgorrie:
Annoyingly a commit was missed in #955.


jamesgorrie:
To explain - we set the no rights at the beginning, not when the value has been changed.

theefer:
:+1:

PR #959, opened 2015-07-20
Using icons for pay and restricted images in results.

akash1810:
- Pay is a £ character (material icons doesn't have one)
- Restricted is a flag

Also removing `.ico` classes in favour of material icon `help`.

Why? Provides more space for `persisted` icon to go.

screen shot 2015-07-20 at 12 15 32

### DIFF
--- a/kahuna/public/js/crop/view.html
+++ b/kahuna/public/js/crop/view.html
@@ -34,8 +34,9 @@
                  title="restricted use: {{ctrl.image.data.usageRights.restrictions}}">
                 <!-- As `conditional` can only be set with usageRights, let's
                      just assume it's here. We might need to revisit this. -->
-                     <gr-icon>warning</gr-icon><span class="top-bar-item__label"> restricted use
-                         <span class="ico ico--circled ico--help">?</span></span>
+                <gr-icon>flag</gr-icon>
+                <span class="top-bar-item__label">restricted use</span>
+                <gr-icon>help</gr-icon>
             </div>
         </div>
 
--- a/kahuna/public/js/preview/image.html
+++ b/kahuna/public/js/preview/image.html
@@ -72,7 +72,8 @@
             <div class="preview__cost bottom-bar__action" ng:switch="ctrl.states.cost">
                 <div ng:switch-when="pay"
                      class="cost cost--pay">
-                    <gr-icon>attach_money</gr-icon>
+                    <!-- material icons doesn't have a £ icon -->
+                    <span title="pay to use">£</span>
                 </div>
 
                 <div ng:switch-when="conditional"
--- a/kahuna/public/js/preview/image.html
+++ b/kahuna/public/js/preview/image.html
@@ -69,7 +69,9 @@
                 image="ctrl.image"
                 disabled="ctrl.selectionMode"></ui-archiver>
 
-            <div class="preview__cost bottom-bar__action" ng:switch="ctrl.states.cost">
+            <div class="preview__cost bottom-bar__action"
+                 ng:if="ctrl.states.cost !== 'free'"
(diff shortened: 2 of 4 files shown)


kahuna/public/js/crop/view.html
```diff
@@ -34,8 +34,9 @@
                  title="restricted use: {{ctrl.image.data.usageRights.restrictions}}">
                 <!-- As `conditional` can only be set with usageRights, let's
                      just assume it's here. We might need to revisit this. -->
-                     <gr-icon>warning</gr-icon><span class="top-bar-item__label"> restricted use
-                         <span class="ico ico--circled ico--help">?</span></span>
+                <gr-icon>flag</gr-icon>
+                <span class="top-bar-item__label">restricted use</span>
+                <gr-icon>help</gr-icon>
```
akash1810: screen shot 2015-07-20 at 11 53 11

jamesgorrie:
`$` is american no?

akash1810:
yeah, there isn't a pound [icon](https://www.google.com/design/icons/) (could just use the `£` text I suppose...)

theefer:
Do we think the icon will be obvious enough compared to existing text? I'd hope so but people are often amazing at not guessing right.

cc @alastairjardine 

I'd go with a pound sign character (we could detect and make it dollar if in the US, but let's not).

kahuna/public/js/preview/image.html
```diff
@@ -72,7 +72,7 @@
             <div class="preview__cost bottom-bar__action" ng:switch="image.data.cost">
                 <div ng:switch-when="pay"
                      class="cost cost--pay">
-                    pay
+                    <gr-icon>attach_money</gr-icon>
```
theefer: We'd want a `title` attribute here.

alastairjardine:
It's a good start. I feel that the dollar sign is a bit whispy and hard to make out. It needs to be bolder.

Care with the shopping cart. If the model is indeed such that you have to pay as soon as you crop, rather than if it's used on the site/in the paper, then the shopping cart might not send the right messages. People might ascribe payment to the wrong bit of the process.

alastairjardine:
PS like the flag – models well with flag and correct.

theefer:
Shopping cart is really not obvious compared to some currency symbol imho. Plus what @alastairjardine said.

akash1810:
updated screenshot with `£`

kahuna/public/js/preview/image.html
```diff
@@ -72,7 +72,8 @@
             <div class="preview__cost bottom-bar__action" ng:switch="image.data.cost">
                 <div ng:switch-when="pay"
                      class="cost cost--pay">
-                    pay
+                    <!-- material icons doesn't have a £ icon -->
+                    <span title="paid to use">£</span>
```
theefer: Should it read "pay to use" instead?

theefer:
Could be worth running past Jo quickly in case she has anything to say about this.

alastairjardine:
It would be easy to test this by the way –

alastairjardine:
slippy fingers – clicked the wrong button. Closing is too easy. bad UI.

theefer:
It's easy to recover from the mistake though. :-)

kahuna/public/js/preview/image.html
akash1810: yep!

akash1810:
spoke to Jo and she didn't have any issue with it - she'd like to see it before it goes live though, so will release to TEST.

akash1810:
demoed to Jo on TEST and she was happy with the changes :smile:

theefer:
:+1:

PR #960, opened 2015-07-20
Fix image fade-in

theefer:
The fade-in directive wasn't actually working. This fixes it.

It also removes the existing initial fading in of search results, to avoid having two layers of fading in.

Note: images still only fade in if they are not preloaded faster enough (50ms, typically from browser cache).

### DIFF
--- a/kahuna/public/js/directives/gr-image-fade-on-load.js
+++ b/kahuna/public/js/directives/gr-image-fade-on-load.js
@@ -16,7 +16,7 @@ imageFade.directive('grImageFadeOnLoad',
             // If not loaded after animationThreshold, hide and wait
             // until loaded to fade in
             $timeout(() => {
-                if (! isLoaded) {
+                if (! isLoaded()) {
                     hide();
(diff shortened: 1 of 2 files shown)


kahuna/public/js/directives/gr-image-fade-on-load.js
```diff
@@ -50,13 +50,15 @@ imageFade.directive('grImageFadeOnLoad',
...
-                    transition: `opacity ${animationDuration}ms ease-out`
+                    opacity: 0
                 });
             }
 
             function reveal() {
-                element.css({opacity: 1});
+                element.css({
+                    opacity: 1,
+                    transition: `opacity ${animationDuration}ms ease-out`
```
jamesgorrie: Do you know if there is any perf problems with applying these to every  element?
The only literature I could find on it [was here](http://www.stubbornella.org/content/2009/03/27/reflows-repaints-css-performance-making-your-javascript-slow/#inline) - which makes sense.

jamesgorrie:
:+1:

PR #961, opened 2015-07-20
Show placeholders while items are lazy loading

theefer:
Note: the merge target for this PR is the base infinite-scroll branch.

Placeholders are positioned with the same formula as cells, and we show [as many extra rows of placeholders](https://github.com/guardian/grid/compare/sc-infinite-scroll...sc-infinite-scroll+placeholders-bis?expand=1#diff-e67d7a61b905b41b4f3d88c228e1ef95R98) as we preload beyond the viewport. For example, the preloaded-rows attribute is currently 4, so we load 4 rows of results before and after the current viewport; that means we should 8 rows of placeholders before and after the current viewport. This is to make it harder to scroll where there are no placeholders.

Jumping further down can still briefly reveal an area with no placeholder, but that's just a performance trade-off and it's acceptable imo.

### DIFF
--- a/kahuna/public/js/components/gu-lazy-table/gu-lazy-table.js
+++ b/kahuna/public/js/components/gu-lazy-table/gu-lazy-table.js
@@ -97,7 +97,8 @@ lazyTable.controller('GuLazyTableCtrl', [function() {
         // Placeholders
         const placeholderExtraCount$ = mult$(columns$, preloadedRows$);
         const placeholderRangeStart$ = max$(sub$(loadedRangeStart$, placeholderExtraCount$), 0);
-        const placeholderRangeEnd$ = min$(add$(loadedRangeEnd$, placeholderExtraCount$), itemsCount$);
+        const placeholderRangeEnd$ = min$(add$(loadedRangeEnd$, placeholderExtraCount$),
+                                          itemsCount$);
 
         const placeholderIndexes$ = combine$(
             items$, placeholderRangeStart$, placeholderRangeEnd$,
--- a/kahuna/public/js/components/gu-lazy-table/gu-lazy-table.js
+++ b/kahuna/public/js/components/gu-lazy-table/gu-lazy-table.js
@@ -10,8 +10,8 @@ import './gu-lazy-table-cell';
 
 import {
     combine$,
-    add$, sub$, mult$, div$,
-    floor$, ceil$, max$
+    add$, sub$, mult$, div$, mod$,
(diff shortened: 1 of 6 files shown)


kahuna/public/js/components/gu-lazy-table/gu-lazy-table.js
```diff
@@ -92,24 +94,59 @@ lazyTable.controller('GuLazyTableCtrl', [function() {
...
+        const placeholderExtraCount$ = mult$(columns$, preloadedRows$);
+        const placeholderRangeStart$ = max$(sub$(loadedRangeStart$, placeholderExtraCount$), 0);
+        const placeholderRangeEnd$ = min$(add$(loadedRangeEnd$, placeholderExtraCount$),
+                                          itemsCount$);
+
+        const placeholderIndexes$ = combine$(
+            items$, placeholderRangeStart$, placeholderRangeEnd$,
+            (items, placeholderRangeStart, placeholderRangeEnd) => {
+                let indexes = [];
+                for (let i = placeholderRangeStart; i < placeholderRangeEnd; i++) {
```
jamesgorrie: No `range`? Damn ES6.

theefer: I think @akash1810 may have written a range elsewhere using generators actually (?).

Happy to replace if so and no performance hit.

jamesgorrie:
Whilst testing I still have to wait for the API to return to get the placeholders - this isn't mean to be the case is it?

kahuna/public/js/components/gu-lazy-table/gu-lazy-table.js
akash1810: > I think @akash1810 may have written a range elsewhere using generators actually (?).

[Yep](https://github.com/guardian/grid/blob/master/kahuna/public/js/search/results.js#L169).

jamesgorrie:
^ To clarify the above it when you revisit a scroll location you've been to prior.

kahuna/public/js/components/gu-lazy-table/gu-lazy-table.js
theefer: Now using range helper.

theefer:
Can't seem to reproduce issue with scrolling up.

jamesgorrie:
Could just be my refresh being super slow. :love_hotel: OSX.

jamesgorrie:
:+1: Looks good.

PR #962, opened 2015-07-20
update PR image to be optionally restricted

jamesgorrie:
Seems I was wrong - Jo B, Jonny W etc have all said this is Free and _could_ be restricted.


theefer:
:+1:

PR #963, opened 2015-07-20
Moving `hasExportsOfType` to image service.

akash1810:
sorry for the big diff...

### DIFF
--- a/kahuna/public/js/image/service.js
+++ b/kahuna/public/js/image/service.js
@@ -7,7 +7,8 @@ export const imageService = angular.module('gr.image.service', ['kahuna.edits.se
 imageService.factory('imageService', ['editsService', function(editsService) {
     function forImage(image) {
         return {
-            usageRights: usageRights(image)
+            usageRights: usageRights(image),
+            states: states(image)
         };
     }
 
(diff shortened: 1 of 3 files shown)


kahuna/public/js/image/service.js
```diff
@@ -28,6 +29,18 @@ imageService.factory('imageService', ['editsService', function(editsService) {
         return { data, save, remove };
     }
 
+    function hasExportsOfType(image, type) {
+        return image.data.exports &&
+                image.data.exports.some(ex => ex.type === type);
+    }
+
+    function states(image) {
```
theefer: Could we stick to making functions verbs describing their action?

kahuna/public/js/preview/image.html
```diff
@@ -66,8 +66,8 @@
 
             <ui-archiver
                 class="bottom-bar__action"
-                image="image"
-                disabled="selectionMode"></ui-archiver>
+                image="ctrl.image"
+                disabled="ctrl.selectionMode"></ui-archiver>
 
             <div class="preview__cost bottom-bar__action" ng:switch="image.data.cost">
```
theefer: I think you missed one `image` here and below.

akash1810: oops, fixed now.

kahuna/public/js/image/service.js
```diff
@@ -7,7 +7,8 @@ export const imageService = angular.module('gr.image.service', ['kahuna.edits.se
 imageService.factory('imageService', ['editsService', function(editsService) {
     function forImage(image) {
         return {
-            usageRights: usageRights(image)
+            usageRights: usageRights(image),
+            states: getStates(image)
```
jamesgorrie: Not sure states sounds right here i.e. It's in the hasCrops state?
I can't think of anything better though :bike:...

jamesgorrie:
:+1:

PR #967, opened 2015-07-20
consistant casing.

akash1810:


kahuna/public/js/search/query.js
```diff
@@ -55,7 +55,7 @@ query.controller('SearchQueryCtrl',
 
             // don't track changes to `query` as it would trigger on every keypress
             if (key !== 'query') {
-                track.success('Query Change', { field: key, value: newVal });
+                track.success('Query change', { field: key, value: newVal });
```
jamesgorrie: Ditto with these:
```{ 'Field': key, 'Value': newVal }```

PR #968, opened 2015-07-20
[WIP] Batch usage rights

jamesgorrie:
Thought this might make sense, let's see.

@kenoir - perhaps a simpler way of looking at batch usage rights, deferring the logic that works with multiple usage-rights higher up. An alternative universe to #922 :dizzy:.

Seems like we might need less refactoring and have more control for different situations i.e. multiple select vs uploads page.

Maybe not, let's chat tomorrow.


kenoir:
Closing in favour of https://github.com/guardian/grid/pull/922

PR #969, opened 2015-07-20
fix route.

akash1810:
[oops](https://github.com/guardian/grid/pull/963/files#diff-b0adc273d4eb8af691690a0f30abcd26R29)


jamesgorrie:
Ha, that's almost the opposite of what we've done in the past.
:+1:

PR #970, opened 2015-07-21
Remove filter on deprecated cost property

theefer:
It was removed from the data some time ago so not useful in the query anymore.


theefer:
cc @jamesgorrie

jamesgorrie:
:+1:

PR #972, opened 2015-07-21
Usage rights on image page

kenoir:
Adds usage rights to image page.

crop

Depends: https://github.com/guardian/grid/pull/922

See changes: https://github.com/guardian/grid/compare/usage-right-on-search-page...usage-rights-on-image-page

**TODO:**

- [ ] Fix error with handing around non-promise "resource.getUri is not a function"


kenoir:
Much Copy Pasting.

kahuna/public/js/image/view.html
```diff
@@ -49,6 +49,35 @@
...
 
         <div class="image-info__group">
+            <div class="image-info__wrap metadata-line drop-menu image-info__usage-rights"
+                 ng:init="showUsageRights = false">
+                usage rights category
+                <div class="image-info__usage-rights drop-menu__items drop-menu__items--right"
+                     ng:show="showUsageRights">
+
+                    <button type="button" class="drop-menu__close" ng:click="showUsageRights = false">
+                        <i class="material-icons">close</i>
```
akash1810: `.material-icons` has been removed now - `<gr-icon>close</gr-icon>`

theefer:
The "Close" text (is it supposed to be an icon, as per @akash1810's comment above?) for the edit modal box is only on the image page, not on the rights from the search side-panel or from the uploads page. Given that there is a cancel/save at the bottom, is it needed?

image

kahuna/public/js/image/view.html
jamesgorrie: Removed as per #972

jamesgorrie:
@theefer removed as per #972

kenoir:
Weird error in this branch where resources are sometimes not promises so the editService balks. Maybe @jamesgorrie can take a look with me?

kenoir:
Merging after addressing problem identified by @akash1810 .

akash1810:
:+1: for fun

PR #973, opened 2015-07-22
using correct property for image validity.

akash1810:
Another oops caused by changes in https://github.com/guardian/grid/pull/963.

We render the crop symbol on the search page based on this property.


theefer:
:+1:

Issue #931, opened 2015-07-13
Batch remove all labels doesn't work on upload screen


akash1810:
When you upload more than one image, the upload page is in `batch` state, meaning the double headed arrows are shown, allowing you to apply the metadata from one image to all the images in the current upload job.

When there are no labels, clicking the double headed arrow results in a 500 error as the POST payload is `{data: []}`, whereas it should be a DELETE call that is made to remove the labels.

Is this the right workflow for this use case? Should there be a `delete all labels` button rather than inferring what the arrow will do when there are no labels?

Relates to #902.

paperboyo:
Hi,
Just my 2gr. Since Grid doesn’t export labels as metadata yet (https://github.com/guardian/grid/issues/823), there is no way that the current batch of uploads would contain __any__ labels at the beginning of upload procedure. Applying a no label state is thus an entirely uncommon scenario (one would have to create label(s) first and then decide to actually remove them). An extra button is complicating things too much, then.

That said, ideally, the behaviour/UX/UI of all batch operations should be consistent. Currently, it differs between Upload screen and Mutliple Selection interface (even the Apply to all icon is different!). Maybe Upload screen should move closer to Multiple Selection UX (ideally, disappearing entirely :wink:).

Regards
Mateusz

theefer:
@akash1810 Is it worth delaying fixing this use case until we unify the batch labelling with that on the search page? As long as we prevent any user action from resulting in noisy 500s on the servers.

PR #974, opened 2015-07-22
Perform DELETE request on uploads page when applying no labels.

akash1810:
Fixes https://github.com/guardian/grid/issues/931.

delete-all-labels

### DIFF
--- a/kahuna/public/js/edits/labeller.html
+++ b/kahuna/public/js/edits/labeller.html
@@ -22,10 +22,19 @@
         Add<span ng:show="ctrl.adding">ing…</span>
     </button>
 
-    <button
-        class="labeller__apply-all"
-        title="Apply these labels to all"
-        ng:if="ctrl.withBatch"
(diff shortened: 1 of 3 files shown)


jamesgorrie:
Are we sure we want to do this? At least without confirmation as you could make a big :hankey: here?

akash1810:
> Are we sure we want to do this?
I thought we decided it was a good feature to have here - https://github.com/guardian/grid/pull/925

jamesgorrie:
Indeed - wandering if we should have a confirmation on this?

jamesgorrie:
I suppose the question is are you happy to have this as a feature without a confirmation until someone messes up?

Has someone asked for this?

akash1810:
> I suppose the question is are you happy to have this as a feature without a confirmation until someone messes up?

No. WIPing whilst I add that.

jamesgorrie:
Besides the new colour - :

jamesgorrie:
:+1:

PR #975, opened 2015-07-23
[WIP] add new archiver using stream of images

jamesgorrie:
After trying to do #922 with @kenoir it seemed there was a little bit of :spaghetti:.

Using [`Rx.BehavoiurSubjects`](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/subjects/behaviorsubject.md) we're trying to remove the need for the `image-updated` magic event and move on to having components that take collections / streams of resources, [similar to what we've done](https://github.com/guardian/grid/pull/922/files#diff-595baa1028c67c62a14e9f2a63a98e95R112) with [usage rights](https://github.com/guardian/grid/pull/922/files#diff-533b4c2674310e7ac52f494543d3e8a0R36).

## TODO
- [x] add label editor
- [x] add metadata editor
- [x] add archive editor
- [x] add usage rights editor
- [ ] deprecate old label editor
- [ ] deprecate old add metadata editor
- [ ] deprecate old add archive editor
- [ ] deprecate old usage rights editor
- [ ] clean up the array of service
- [ ] fix structure of app
- [ ] ♥ lint

## Next PR
* Move over every editor to use these batch editors.


kahuna/public/js/components/gr-panel/gr-panel.js
```diff
@@ -43,6 +229,17 @@ grPanel.controller('GrPanel', [
...
 
+
+        ctrl.archivedService = archivedService(selection.images$);
+        ctrl.archivedService.onUpdate(updates => selection.updateImages(updates));
+
+        ctrl.metadataService = metadataService(selection.images$);
+        ctrl.metadataService.onUpdate(updates => selection.updateImages(updates));
+
+        ctrl.labelsService = labelsService(selection.images$);
+        ctrl.labelsService.onUpdate(updates => selection.updateImages(updates));
```
jamesgorrie: Not really sure that this makes the most sense - the idea @theefer had of merging an edits stream with this one would be good, but it least it's moved away from controlling the stream out of context.

jamesgorrie: Just so I remember for Monday:
```JavaScript
// service.js
const imageUpdates$ = new Rx.Subject();
//...
return { add, remove, stuff, imageUpdates };
//...
$q.all(edits).then(updates => imageUpdates$.onNext(updates));

//... gr-panel.js
ctrl.archivedService = archivedService(selection.images$);
selection.imageUpdates(ctrl.archivedService.imageUpdates$);

//... preview-selection.js
funcrion imageUpdate(imageUpdates$) {
  imageUpdate.subscribe(updates => { /*zip images*/ });
}
```

jamesgorrie: [Done](https://github.com/guardian/grid/commit/52a3d4e70e5e3d08c97f029a52e05de5244f47c1)

jamesgorrie:
Thinking about it yesterday - this work is going to make it a lot nicer to add a new / better uploader by having stream of images that we can tap into.

jamesgorrie:
@alastairjardine - would you want to take a look at one of the panel components that we could get right on the UX side to try and get this in?

Basically it's tidying up the selection / editing, but we could use it as a case to improve the panel from a usability perspective.

kahuna/public/js/image/service.js
```diff
@@ -44,3 +45,274 @@ imageService.factory('imageService', ['editsService', function(editsService) {
...
+            }
+        }
+    }
+
+    return labelsService;
+
+}]);
+
+
+imageService.factory('metadataService',
```
akash1810: It feels a bit strange that everything is now on the `imageService` rather than being standalone services.

jamesgorrie: Bare with me, I am doing a quick cleanup...

PR #977, opened 2015-07-23
ES query optimisations

theefer:
Manual tests showed huge improvements in query latency.

Will try on TEST to verify correctness of results.

Tips borrowed from: https://www.found.no/foundation/optimizing-elasticsearch-searches/


jamesgorrie:
:+1:

PR #978, opened 2015-07-23
Removing the archive filter from the UI.

akash1810:
From user feedback and [a couple of days of stats](https://goo.gl/4dn4ss) the "archived" checkbox filter isn't used.

User's feedback has been that its important to be able to archive images, but searching for just those images isn't useful - scanning down the whole grid and seeing visual indicators is the default use case.

We *could* add it to the advanced query syntax, however given the usage of this filter, I don't think its necessary right now.

As a result, we can break at a smaller screen size!

Haven't removed this query string from the media api, as its used by reaper atm.

### DIFF
--- a/kahuna/public/js/search/index.js
+++ b/kahuna/public/js/search/index.js
@@ -36,7 +36,7 @@ search.config(['$stateProvider',
     });
 
     $stateProvider.state('search.results', {
-        url: 'search?query&ids&since&nonFree&archived&uploadedBy&until',
+        url: 'search?query&ids&since&nonFree&uploadedBy&until',
         data: {
             title: function(params) {
                 return params.query ? params.query : 'search';
--- a/kahuna/public/js/search/query.html
+++ b/kahuna/public/js/search/query.html
@@ -56,16 +56,6 @@
             </li>
 
             <li class="search__filter-item">
-                <label class="search__filter">
-                    <input type="checkbox"
-                           ng:true-value="'true'"
-                           ng:false-value="undefined"
-                           ng:model="searchQuery.filter.archived" />
-                    archived
-                </label>
-            </li>
-
-            <li class="search__filter-item">
                 <gu-date-range class="search__filter search__date"
                                gu:start-date="searchQuery.filter.since"
                                gu:end-date="searchQuery.filter.until"
--- a/kahuna/public/stylesheets/main.css
+++ b/kahuna/public/stylesheets/main.css
@@ -439,7 +439,7 @@ textarea.ng-invalid {
     }
 }
 
-@media screen and (max-width: 1100px) {
+@media screen and (max-width: 1000px) {
 
     .search__filters-toggle {
         display: inherit;


theefer:
:+1: :tada:

paperboyo:
> We could add it to the advanced query syntax

Would be cool. Used it once or twice, mostly to make sure that I archived everything, so that it won't get flushed (agree that it shouldn't be in the UI). BTW, I still fail to get all that "archive" business... _Most_ of it seems entirely redundant legacy psychological bull. :cow2: One functionality+many functions=mess. Maybe only me, though.

Keep all the thingz, flush only the expired ones (+display expiry date)? If not that, then at least do not flush manually uploaded/touched (subject to expiry from rights management). Disconnect the need for "picks" from all that (and make "picks" more granular). Or not. :beer:

akash1810:
> I still fail to get all that "archive" business

We're doing work now to simplify it :smile: 

> +display expiry date

This would be interesting to explore - I've a feeling it may not be too useful until we can choose the display order, otherwise you'll have to scroll to the bottom (or use a specific date filter) to see them.

> then at least do not flush manually uploaded/touched

Are you referring the the current use of the `Modified` label? i.e. don't delete images that have this label?

> Disconnect the need for "picks" from all that

Talking with others, we've come to the same conclusion :fireworks: :tada: .

paperboyo:
> Are you referring the the current use of the Modified label?

Yeah, that is redundant that I still click the star, even though I add label (if adding labels triggers archiving, which is not clear and I do not know, would have to ask you and waste your time). But the whole idea that something I've uploaded even **can** disappear just seems **crazy** to me. Always has. Eg. when someone edits the metadata, but do not add labels, crops - will the image disappear? Why? I only understand disappearance of the stuff that not only we cannot buy after the free expires, but that we absolutely **cannot** use, like nevah-evah, even if we pay gazillions. Or if it contains naked Snowden, but he uploaded it by mistake and asked nicely.

The idea that stuff should disappear because it's not relevant and pollutes me-nice-view-that-I-like-to-scroll-around seems to me very 19th century. Just get better search/filtering/sorting/weighting.

Ok, maybe I would flush sport :stuck_out_tongue_winking_eye: (better categories, better categories, better categories!). 

> Talking with others, we've come to the same conclusion

:muscle:

akash1810:
> I still click the star, even though I add label

Its been suggested to auto archive specific labels - only suggested, not fully thought through the mechanics (e.g. what happens if the label is there, but I un-star it too? Should the star even be shown? Would it be confusing if it wasn't?)

> But the whole idea that something I've uploaded even can disappear just seems crazy to me.

Keeps CDN costs down.

paperboyo:
> I've a feeling it may not be too useful until we can choose the display order

True that. And even then not massively...

> Its been suggested to auto archive specific labels

Why only specific? Why not to automatically archive anything that was manually "touched" in any way (even simple metadata edit)? (again, subject to rights expiry)

And, if it's possible/realistic to keep everything apart from the stuff that we've lost rights too, then the whole archive functionality is redundant. Users wouldn't need to ever concern themselves with an archive action, just delete action? And then

> what happens if the label is there, but I un-star it too? Should the star even be shown? Would it be confusing if it wasn't?

would be a non-issue.

Stars/ratings/colour labels/counted votes/flags/whatevah would be used solely to indicate importance/value. These could be made into more collaborative tools with weighted vote counting  and combined with crops usage stats to work out eg. a very good photo depicting a certain subject that picdesk high echelon has in high regard and that readers are sharing on Pinterest like crazy but which was nevertheless overused on a prticular front in the last 3 weeks.

"Keep everything", apart from obvious storage costs (about which I would have no idea and I exclude that from these thoughts on purpose), may have just one downside: "why do we even keep all that ephemeral crap that is never going to be used and makes finding the important stuff hard". To that I say: better search/filters/sorting/weighting...

For now, and to come back to Earth, if the star is to still indicate both archive status AND suggest a pick (a conflation that I think is wrong and the longer it persists the worse - harder to then untangle when making right), then maybe stuff that is archived implicitly by means of cropping/labelling/whatever should have a greyed-out star that's fixed? But then it's not yellow and it's harder to see and I may miss this important photo and I want to give it a star, cause it's brilliant and all that which only shows the downside of the entanglement.

Gee, sorry for the length!

PR #979, opened 2015-07-23
[wip] Dulling down the archive status.

akash1810:
Following https://github.com/guardian/grid/pull/978, the presence of the icon is a strong enough indicator.

archive-colour

The hover state is also the same as the other items in the top bar too:

archive-colour-ii


theefer:
:+1:

PR #980, opened 2015-07-24
A few styling tweaks for consistency.

akash1810:
Panel's left border matches that of the top bar. The width of the batch panel also aligns to the recent uploads button (in "full" mode).

Before
screen shot 2015-07-24 at 10 30 18

After
screen shot 2015-07-24 at 10 28 44

Preview panel also has the same background colour as the batch panel.

### DIFF
--- a/kahuna/public/js/components/gr-panel/gr-panel.css
+++ b/kahuna/public/js/components/gr-panel/gr-panel.css
@@ -4,7 +4,7 @@
 }
 
 .panel--side {
-    width: 280px;
+    width: 284px;
     height: calc(100vh - 48px);
 }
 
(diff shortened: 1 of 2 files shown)


kenoir:
Pinging @alastairjardine so he can see.

kenoir:
Also :+1:

kahuna/public/js/components/gr-panel/gr-panel.css
```diff
@@ -4,7 +4,7 @@
 }
 
 .panel--side {
-    width: 280px;
+    width: 284px;
```
theefer: Can we document (as a comment) these magic values to people who come across them later know what to watch out for? It also alleviates fear of change if they no longer apply because the reason has gone away.

akash1810: Yep - https://github.com/guardian/grid/commit/08df850c3cb631b7c5701871407ef4f07e4db197.

PR #982, opened 2015-07-24
Adding dimensions to image page.

akash1810:
Adding image dimensions to the metadata panel of the image page.

screen shot 2015-07-24 at 14 46 16

Also using more MD icons!
screen shot 2015-07-24 at 14 46 31

### DIFF
--- a/kahuna/public/js/image/controller.js
+++ b/kahuna/public/js/image/controller.js
@@ -1,9 +1,14 @@
 import angular from 'angular';
 
+import './service';
 import '../edits/service';
 import '../analytics/track';
 
-var image = angular.module('kahuna.image.controller', ['kahuna.edits.service', 'analytics.track']);
+var image = angular.module('kahuna.image.controller', [
+    'gr.image.service',
+    'kahuna.edits.service',
+    'analytics.track'
+]);
 
 image.controller('ImageCtrl', [
     '$rootScope',
@@ -14,6 +19,7 @@ image.controller('ImageCtrl', [
     'optimisedImageUri',
     'cropKey',
     'mediaCropper',
+    'imageService',
     'editsService',
     'track',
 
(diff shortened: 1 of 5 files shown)


kahuna/public/js/image/controller.js
```diff
@@ -41,6 +48,9 @@ image.controller('ImageCtrl', [
         };
 
         ctrl.image = image;
+
+        ctrl.imageDimensions = imageService(ctrl.image).dimensions;
```
theefer: Not too important here, but I may be more keen on having dedicated accessor helpers as we have in Composer (e.g. `imageAccessor.getDimensions(image)`) rather than one function like here. The function here does a lot of work computing lots of information that isn't required, making this slower than necessary, and also harder to track what code uses what computed information.

theefer:
:+1:

kahuna/public/js/image/controller.js
akash1810: Essentially a model in mvc terms, right? Or have I misunderstood?

akash1810:
Reverting as crops are displaying the dimensions of the original image and not of the crop :-1:

kahuna/public/js/image/controller.js
theefer: In Composer we've stayed away from "models" (rich objects that wrap data) and preferred a more functional approach using functions that transform plain data. I think it worked quite well, but happy to discuss.

paperboyo:
Our fonts contain proper multiplication sign × glyph even in limited Latin1 version (we should load the widest one, though, for other reasons and the italic!) at  U+00D7 (HTML ``&#215;`` or ``&times;``). It looks better, IMHO, and no need for icon trickery :smiley:

capture

akash1810:
@paperboyo I think I've gone icon crazy as this didn't even occur to me! Not only does it look better, its a better solution! (Try copy pasting the version with the icon and you get `3000close2000` :cry:). Will change, thanks!!

paperboyo:
:laughing: Not at all, Sir! Crazy I went once optimising icon to be 184 B smaller :laughing: https://github.com/guardian/frontend/pull/6670#issuecomment-60177389

PR #984, opened 2015-07-24
Display dimensions.

akash1810:
Like https://github.com/guardian/grid/pull/982 but working for crops too.

dimensions

### DIFF
--- a/kahuna/public/js/crop/controller.js
+++ b/kahuna/public/js/crop/controller.js
@@ -50,10 +50,9 @@ crop.controller('ImageCropCtrl',
         }
     });
 
-    var cropWidth = () => Math.round(ctrl.coords.x2 - ctrl.coords.x1);
-    var cropHeight = () => Math.round(ctrl.coords.y2 - ctrl.coords.y1);
-    ctrl.cropSize = () => cropWidth() + ' x ' + cropHeight();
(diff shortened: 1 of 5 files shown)


theefer:
Feels a bit weird to be updating this information for each crop. The information in the right panel tends to pertain to the original picture, not each individual crop you may select (e.g. uploaded etc all relate to the original image).

Could possibly work around by displaying the original dimensions and labelling them "original dimensions"?

This may have a more natural location once we rework the way we show original and crops, as discussed.

Not strongly against releasing this, but slightly reluctant to introduce a precedent of crop-related info here.

akash1810:
> slightly reluctant to introduce a precedent of crop-related info here

The dimensions is the only thing that differs between the original and the crop. It felt a bit strange to not update the dimensions as there was a direct disconnect between the information shown in the panel and what the user has an inclination on - though maybe I chose a particularly small crop...

theefer:
Will leave it up to you and @alastairjardine to decide. It'll all get tidied up soon with the crops rework anyway.

paperboyo:
Oops, I should have commented here, sorry: https://github.com/guardian/grid/pull/982#issuecomment-124648977

akash1810:
now displaying original dimensions when viewing a crop (see updated gif).

theefer:
:+1: for expliciteness

PR #986, opened 2015-07-24
Resume vertical position when navigating back to search results

theefer:
As long requested and wished for.

Bit fiddly due to browser scrolling being slightly odd, but works well enough.

Demo: https://i.imgur.com/sn4rH7F.gif

## Known issue

Scroll down the default search, click on an image to preview it, then click on the Home/Grid button. One would expect to get back at the top of the results, but the view is actually scrolled down to the previous position.

Note that the Home/Grid button correctly shows the top of the results for any other search than the default.

### DIFF
--- a/kahuna/public/js/components/gu-lazy-table/gu-lazy-table.js
+++ b/kahuna/public/js/components/gu-lazy-table/gu-lazy-table.js
@@ -297,6 +297,7 @@ lazyTable.directive('guLazyTable', ['$window', 'observe$',
                 // with other cell rendering updates
                 scope.$applyAsync(() => {
                     element.css('height', viewHeight + 'px');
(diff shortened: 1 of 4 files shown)


kenoir:
Nice.

theefer:
Demo'd IRL to @alastairjardine.

theefer:
Noticed an issue on PROD whereby the search results aren't "locked" to what they were when you navigated away from then. In other words, if the search is matching a set of results that changes quickly (e.g. no filter at all), when navigating back new images might have arrived and although you're at the same vertical position, it's not the same set of results and things will have moved, which is confusing.

I don't think it's worth rolling back given how broken the previous behaviour was in the first place (everyone is currently working around it), however I'm going to work on a fix before advertising the feature.

PR #987, opened 2015-07-26
Expose delete argo action on images that can be deleted

theefer:
Expose a hypermedia affordance to declaratively describe a delete action on images, using an experimental extension to the argo media-type (`"actions"` property on resource entities).

By default, a user can only delete their own uploaded images if they haven't been persisted. A new `deleteImage` permission list has been added to list users who can delete any non-persisted image.

This will allow the UI(s) to know when to expose a 'Delete' button on an image.

Example (used both in search results and individual image):

``` json
{  
   "data":{  
      "id":"1ef04f13576e87a053ddd3decbef2cf4e2f21f02",
      "uploadTime":"2015-07-09T15:04:11Z",
     [...]
      "valid":true,
      "cost":"free",
      "persisted":false
   },
   "links":[  
      {  
         "rel":"crops",
         "href":"https://cropper.grid.com/crops/1ef04f13576e87a053ddd3decbef2cf4e2f21f02"
      },
      [...]
   ],
   "actions":[  
      {  
         "name":"delete",
         "href":"https://api.media.grid.com/images/1ef04f13576e87a053ddd3decbef2cf4e2f21f02",
         "method":"DELETE"
      }
   ]
}
```

Note that there is no support for this in theseus yet, but preliminary tests seem to indicate that the extra data doesn't break existing clients.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/argo/model/Action.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/argo/model/Action.scala
@@ -1,0 +1,20 @@
+package com.gu.mediaservice.lib.argo.model
+
+import java.net.URI
+
+import play.api.libs.json._
+import play.api.libs.functional.syntax._
+
+
+// TODO: add specification of parameters and body structure, mimeType
+case class Action(name: String, href: URI, method: String)
+
+object Action {
+
+  implicit val actionWrites: Writes[Action] = (
+    (__ \ "name").write[String] ~
+      (__ \ "href").write[String].contramap((_: URI).toString) ~
+      (__ \ "method").write[String]
+    )(unlift(Action.unapply))
+
+}
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/argo/model/EmbeddedEntity.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/argo/model/EmbeddedEntity.scala
@@ -11,7 +11,8 @@ import com.gu.mediaservice.lib.argo.WriteHelpers
 case class EmbeddedEntity[T](
   uri: URI,
   data: Option[T],
-  links: List[Link] = Nil
+  links: List[Link] = Nil,
+  actions: List[Action] = Nil
 )
 
 object EmbeddedEntity extends WriteHelpers {
@@ -19,7 +20,8 @@ object EmbeddedEntity extends WriteHelpers {
   implicit def embeddedEntityWrites[T: Writes]: Writes[EmbeddedEntity[T]] = (
     (__ \ "uri").write[String].contramap((_: URI).toString) ~
       (__ \ "data").writeNullable[T] ~
-      (__ \ "links").writeNullable[List[Link]].contramap(someListOrNone[Link])
+      (__ \ "links").writeNullable[List[Link]].contramap(someListOrNone[Link]) ~
+      (__ \ "actions").writeNullable[List[Action]].contramap(someListOrNone[Action])
     )(unlift(EmbeddedEntity.unapply[T]))
 
 }
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Store.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Store.scala
@@ -66,6 +66,7 @@ class KeyStore(bucket: String, credentials: AWSCredentials) extends BaseStore[St
 object PermissionType extends Enumeration {
   type PermissionType = Value
   val EditMetadata = Value("editMetadata")
+  val DeleteImage  = Value("deleteImage")
 }
 
 class PermissionStore(bucket: String, credentials: AWSCredentials) extends BaseStore[PermissionType, List[String]](bucket, credentials) {
--- a/media-api/app/controllers/MediaApi.scala
+++ b/media-api/app/controllers/MediaApi.scala
@@ -107,12 +107,12 @@ object MediaApi extends Controller with ArgoHelpers {
     ElasticSearch.getImageById(id) flatMap {
       case Some(source) => {
         val withWritePermission = canUserWriteMetadata(request, source)
+        val withDeletePermission = canUserDeleteImage(request, source)
 
-        withWritePermission.map {
-          permission => {
-            val (imageData, imageLinks) = ImageResponse.create(id, source, permission, include)
-            respond(imageData, imageLinks)
-          }
+        Future.sequence(List(withWritePermission, withDeletePermission)).map {
+          case List(writePermission, deletePermission) =>
+            val (imageData, imageLinks, imageActions) = ImageResponse.create(id, source, writePermission, deletePermission, include)
+            respond(imageData, imageLinks, imageActions)
         }
       }
       case None => Future.successful(ImageNotFound)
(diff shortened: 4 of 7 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/argo/model/Action.scala
```diff
@@ -0,0 +1,20 @@
+package com.gu.mediaservice.lib.argo.model
+
+import java.net.URI
+
+import play.api.libs.json._
+import play.api.libs.functional.syntax._
+
+
+// TODO: add specification of parameters and body structure, mimeType
```
theefer: I started with the simplistic DELETE action to avoid having to solve this.

media-api/app/controllers/MediaApi.scala
```diff
@@ -120,10 +131,32 @@ object MediaApi extends Controller with ArgoHelpers {
...
+              // TODO: use respond
+              Accepted.as(ArgoMediaType)
+            } else {
+              ImageDeleteForbidden
+            }
+          }
+        }
+
+      case None => Future.successful(ImageNotFound)
+    }
```
theefer: This logic is quite meh, but haven't found an easier way to express it as it mixes Future, Options and Booleans...

common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Store.scala
```diff
@@ -66,6 +66,7 @@ class KeyStore(bucket: String, credentials: AWSCredentials) extends BaseStore[St
 object PermissionType extends Enumeration {
   type PermissionType = Value
   val EditMetadata = Value("editMetadata")
+  val DeleteImage  = Value("deleteImage")
```
theefer: Currently nobody has this permission, which is OK. Need to check if the list would initially be the same as `editMetadata`.

common-lib/src/main/scala/com/gu/mediaservice/lib/argo/model/EmbeddedEntity.scala
```diff
@@ -19,7 +20,8 @@ object EmbeddedEntity extends WriteHelpers {
   implicit def embeddedEntityWrites[T: Writes]: Writes[EmbeddedEntity[T]] = (
     (__ \ "uri").write[String].contramap((_: URI).toString) ~
       (__ \ "data").writeNullable[T] ~
-      (__ \ "links").writeNullable[List[Link]].contramap(someListOrNone[Link])
+      (__ \ "links").writeNullable[List[Link]].contramap(someListOrNone[Link]) ~
+      (__ \ "actions").writeNullable[List[Action]].contramap(someListOrNone[Action])
```
jamesgorrie: A concern about this being none - with links - we know we are going to have a specific set of links.
With actions they seem more conditional, i.e. you can't delete, no delete action.

This leaves the API with two shapes:
```JSON
{
  "data": {},
  "actions": [ {"detete": {} }]
}
```
& 
```JSON
{
  "data": {}
}
```

On the client this would mean we would have have annoying client code of:
```JavaScript
if (actions ** getAction('delete')) {/*magical sparkles*/}
```

Although thinking about it - I am assuming this will be done in theseus - so it's a non-problem?

jamesgorrie: Thinking about this further - I'm assuming it'll be promised based?
e.g.
```JavaScript
resource.getAction('delete').then(action => { /* show action content */ })
```
Either way this is probably a discussion for the theseus client.

I suppose the question here is whether to have an empty array to say there are no actions, or no array.

theefer: Yes the idea is to abstract this in theseus, something like:

``` js
resource.perform('action').
  then(resp => /* result of performing the request */).
  catch(err => {
    // no such action error, or request failed error, or callback failed error
  })
```

The same way as we do:

``` js
resource.follow('link').
  then(res => /* resource after following link */`).
  catch(err => {
    // no such rel error, or request to `resource` error, or callback failed error
  })
```

Links are omitted if none and I think it makes sense to do the same for actions. It makes the response lighter and also more backward compatible.

media-api/app/lib/ImageResponse.scala
```diff
@@ -24,7 +24,16 @@ object ImageResponse {
 
   def fileMetaDataUri(id: String) = URI.create(s"${Config.rootUri}/images/$id/fileMetadata")
 
-  def create(id: String, esSource: JsValue, withWritePermission: Boolean, included: List[String] = List()): (JsValue, List[Link]) = {
+  // TODO: move this as a method of Image (fiddly due to dep on Config)
+  def imageIsPersisted(image: Image) = {
```
jamesgorrie: Strange place to live as we use it in other places bar the response.

theefer: Can you think of any better place? I'd be keen to move this to the model if possible, but it's a bit of a yakshave. :-(

media-api/app/lib/ImageResponse.scala
```diff
@@ -24,7 +24,16 @@ object ImageResponse {
...
+    image.identifiers.contains(Config.persistenceIdentifier) ||
+      image.exports.length > 0 ||
```
jamesgorrie: `image.exports.nonEmpty`

media-api/app/controllers/MediaApi.scala
```diff
@@ -120,10 +131,32 @@ object MediaApi extends Controller with ArgoHelpers {
     }
   }
 
-  def deleteImage(id: String) = Authenticated {
-    Notifications.publish(Json.obj("id" -> id), "delete-image")
-    // TODO: use respond
-    Accepted.as(ArgoMediaType)
+
+  val ImageCannotBeDeleted = respondError(MethodNotAllowed, "cannot-delete", "Cannot delete persisted images")
```
jamesgorrie: As these aren't types why are they capitalised?

theefer: Sticking to the convention in this file. I think it originated in these being objects, similarly to `MethodNotAllowed` etc.

I don't feel particularly strong about it, just seems to make sense to me given how they're used. I don't mind changing but could do outside this PR.

jamesgorrie: Not fussed, just wondering.

media-api/app/lib/ImageResponse.scala
jamesgorrie: I like the idea of it being part of the companion object for the model, but if it's a pain, we'll address it when it's more of a pain not to have.

common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Store.scala
```diff
@@ -66,6 +66,7 @@ class KeyStore(bucket: String, credentials: AWSCredentials) extends BaseStore[St
 object PermissionType extends Enumeration {
```
jamesgorrie: icon_offtopic
What do we think of enumerations? That's what I was trying to avoid by using case objects and traits in the usage rights, which seems a [common](http://stackoverflow.com/questions/1898932/case-objects-vs-enumerations-in-scala) pattern (or even better things like [enumeratum](https://github.com/lloydmeta/enumeratum)). The basic task to try and get matching working.

theefer: I also would tend to prefer case objects and traits, we could change this in a future PR if nobody has a particular preference for `Enumeration`.

jamesgorrie:
:+1: Definitely useful (delete and actions).

media-api/app/lib/ImageResponse.scala
theefer: Done

theefer:
Seems to break upload in TEST, reverting.

PR #988, opened 2015-07-27
Archiving images where the byline is a staff photographer.

akash1810:


akash1810:
Closing as we are writing to ES rather than dynamo in this PR (which is wrong!).

Discussed with @theefer about having a stream of incoming images for the services to listen/respond to which will be a better (and more extendable) solution for this.

The stream could be SNS/SQS or Kinesis - research required around what'll work best.

Issue #946, opened 2015-07-17
Show more date information.


akash1810:
Display date taken and ingest date on the thumbnails. NB: We don't always have date taken information.

This will likely be tied to the ability to date search on date taken too (atm, date search is only on ingest date).

Questions:
- [ ] should the implementation be click to toggle between or should both be displayed at all times?
- [ ] should they be relative time stamps (5 minutes ago) or full timestamps (17/07/2015 11:33)?

paperboyo:
Now there is an ability to search by `dateTaken` (although, the last set of values does not look like it makes a difference - miliseconds?).

>We don't always have date taken information.

We may be reading it wrong, see: https://github.com/guardian/grid/issues/1925

> should the implementation be click to toggle between or should both be displayed at all times?

Definitely not all at all times (not before we will get user preferences to show/hide all the badges). But, I would show the ones that are relevant for the current search (and then click to toggle temporarily, why not):

- `dateTaken` for default searches (that are ordered - newest/oldest - by `dateTaken`, I suppose?)
- `uploadTime` for searches using Upload time in date picker
- `lastModified` for searches using Last modified in date picker (would that be oroginal image date modified or last in-Grid modifications as well?

akash1810:
Closing as no recent activity and the use case isn't clear.

paperboyo:
> the use case isn't clear

Will be quite clear when https://github.com/guardian/grid/issues/1925 is implemented. If.

PR #989, opened 2015-07-27
OrderBy filter in the UI.

akash1810:
It already exists in the API, so lets use it!

screen shot 2015-07-27 at 17 54 26

### DIFF
--- a/kahuna/public/js/search/query.html
+++ b/kahuna/public/js/search/query.html
@@ -62,7 +62,7 @@
                                gu:preset-dates="searchQuery.sinceOptions"
                                gu:first-day="1">
                 </gu-date-range>
-            <li>
+            </li>
         </ul>
     </div>
 </form>
--- a/kahuna/public/js/search/results.html
+++ b/kahuna/public/js/search/results.html
@@ -13,14 +13,39 @@
 
     <div ng:if="ctrl.totalResults > 0">
 
-        <div class="image-results-count">
-            {{ctrl.totalResults | toLocaleString}} matches
-            <span ng:if="ctrl.inSelectionMode()">({{ctrl.selectedImages.size}} selected)</span>
-            <button ng:show="ctrl.newImagesCount > 0"
-                    ng:click="ctrl.revealNewImages()"
-                    class="image-results-count__new">
-                {{ctrl.newImagesCount | toLocaleString}} new
-            </button>
+        <div class="results-toolbar">
+            <div class="image-results-count">
+                {{ctrl.totalResults | toLocaleString}} matches
+                <span ng:if="ctrl.inSelectionMode()">({{ctrl.selectedImages.size}} selected)</span>
+                <button ng:show="ctrl.newImagesCount > 0"
+                        ng:click="ctrl.revealNewImages()"
+                        class="image-results-count__new">
+                    {{ctrl.newImagesCount | toLocaleString}} new
+                </button>
+            </div>
+            <div class="sort-direction">
+                <div class="radio-list">
+                    <div class="radio-list__item">
+                        <input type="radio"
+                               id="sort-direction__desc"
+                               class="radio-list__circle"
+                               name="sort-direction">
+                        <label for="sort-direction__desc" class="radio-list__label">
+                            Newest First
+                        </label>
+                    </div>
+                    
+                    <div class="radio-list__item">
+                        <input type="radio"
+                               id="sort-direction__asc"
+                               class="radio-list__circle"
+                               name="sort-direction">
+                        <label for="sort-direction__asc" class="radio-list__label">
+                            Oldest First
+                        </label>
+                    </div>
+                </div>
+            </div>
         </div>
 
         <ul class="results"
--- a/kahuna/public/js/search/results.html
+++ b/kahuna/public/js/search/results.html
@@ -30,8 +30,8 @@
                                id="sort-direction__desc"
                                class="radio-list__circle"
                                name="sort-direction"
-                               value="DESC"
-                               ng:model="ctrl.filter.sortDirection">
+                               ng:value="undefined"
+                               ng:model="ctrl.filter.orderBy">
                         <label for="sort-direction__desc" class="radio-list__label">
                             Newest first
                         </label>
@@ -42,8 +42,8 @@
                                id="sort-direction__asc"
                                class="radio-list__circle"
                                name="sort-direction"
-                               value="ASC"
-                               ng:model="ctrl.filter.sortDirection">
+                               ng:value="'oldest'"
+                               ng:model="ctrl.filter.orderBy">
                         <label for="sort-direction__asc" class="radio-list__label">
                             Oldest first
                         </label>
--- a/kahuna/public/js/search/results.html
+++ b/kahuna/public/js/search/results.html
@@ -29,17 +29,21 @@
                         <input type="radio"
                                id="sort-direction__desc"
                                class="radio-list__circle"
-                               name="sort-direction">
+                               name="sort-direction"
+                               value="DESC"
+                               ng:model="ctrl.filter.sortDirection">
                         <label for="sort-direction__desc" class="radio-list__label">
-                            Newest First
+                            Most Recent First
                         </label>
                     </div>
-                    
+
                     <div class="radio-list__item">
                         <input type="radio"
                                id="sort-direction__asc"
                                class="radio-list__circle"
-                               name="sort-direction">
+                               name="sort-direction"
+                               value="ASC"
+                               ng:model="ctrl.filter.sortDirection">
                         <label for="sort-direction__asc" class="radio-list__label">
                             Oldest First
                         </label>
--- a/kahuna/public/js/search/results.html
+++ b/kahuna/public/js/search/results.html
@@ -33,7 +33,7 @@
                                value="DESC"
                                ng:model="ctrl.filter.sortDirection">
                         <label for="sort-direction__desc" class="radio-list__label">
-                            Most Recent First
+                            Newest first
                         </label>
                     </div>
 
@@ -45,7 +45,7 @@
                                value="ASC"
                                ng:model="ctrl.filter.sortDirection">
                         <label for="sort-direction__asc" class="radio-list__label">
-                            Oldest First
+                            Oldest first
                         </label>
                     </div>
                 </div>
--- a/kahuna/public/js/search/results.html
+++ b/kahuna/public/js/search/results.html
@@ -32,8 +32,10 @@
                                name="sort-direction"
                                ng:value="undefined"
                                ng:model="ctrl.filter.orderBy">
-                        <label for="sort-direction__desc" class="radio-list__label">
-                            Newest first
+                        <label for="sort-direction__desc" class="radio-list__label"
+                               ng:class="{'radio-list--selected': ctrl.filter.orderBy === undefined}">
+                            <div class="radio-list__selection-state"></div>
+                            <div class="radio-list__label-value">Newest first</div>
                         </label>
                     </div>
 
@@ -44,8 +46,10 @@
                                name="sort-direction"
                                ng:value="'oldest'"
                                ng:model="ctrl.filter.orderBy">
-                        <label for="sort-direction__asc" class="radio-list__label">
-                            Oldest first
+                        <label for="sort-direction__asc" class="radio-list__label"
+                               ng:class="{'radio-list--selected': ctrl.filter.orderBy === 'oldest'}">
+                            <div class="radio-list__selection-state"></div>
+                            <div class="radio-list__label-value">Oldest first</div>
                         </label>
                     </div>
                 </div>
--- a/kahuna/public/js/search/results.js
+++ b/kahuna/public/js/search/results.js
@@ -242,7 +242,8 @@ results.controller('SearchResultsCtrl', [
                 until:      until,
                 since:      since,
                 offset:     offset,
-                length:     length
+                length:     length,
+                orderBy:    $stateParams.orderBy
             }));
         }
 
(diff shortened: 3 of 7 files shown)


kahuna/public/js/search/query.js
```diff
@@ -44,6 +44,13 @@ query.controller('SearchQueryCtrl',
         {label: 'Past week',     value: pastWeek}
     ];
 
+    ctrl.orderByOptions = [
+        {label: 'Upload Time (desc)'},
```
akash1810: This is the default; don't set a value to keep the URL clean.

kahuna/public/js/search/query.js
```diff
@@ -44,6 +44,13 @@ query.controller('SearchQueryCtrl',
...
+        {label: 'Upload Time (asc)', value: 'uploadTime'},
```
theefer: Is there any desire for ASC ordering from users?

kahuna/public/js/search/query.html
```diff
@@ -62,7 +62,14 @@
...
                 </gu-date-range>
-            <li>
+            </li>
+
+            <li class="search__filter-item">
+                <select class="search__filter"
+                        ng:model="searchQuery.filter.orderBy"
+                        ng:options="sortOrder.value as sortOrder.label for sortOrder in searchQuery.orderByOptions">
+                </select>
+            </li>
```
theefer: I wonder if having it in the search filters bar and right next to the date range picker could be misinterpreted by users in thinking that:

1. It's a search filter
2. It applies to the date range

It's also worth noting there isn't any hint that this is about ordering.

Would it help to add a label and perhaps move it away from the filters and the date range picker to avoid confusion? As a "post filter" and more advanced option, could it live closer to the results, by the result count?

cc @alastairjardine

kahuna/public/js/search/query.js
```diff
@@ -44,6 +44,13 @@ query.controller('SearchQueryCtrl',
...
+        {label: 'Date Taken (desc)', value: '-dateTaken'},
+        {label: 'Date Taken (asc)',  value: 'dateTaken'}
```
jamesgorrie: Something very useful that has come up is ordering by dateModified - would this be easy to wedge in?

kahuna/public/js/search/query.html
jamesgorrie: In terms of the label - you could try adding "order by ____" in the options?

kahuna/public/js/search/results.js
```diff
@@ -196,7 +196,8 @@ results.controller('SearchResultsCtrl', [
                 until:      until || $stateParams.until,
                 since:      since || $stateParams.since,
                 offset:     offset,
-                length:     length
+                length:     length,
+                orderBy:    $stateParams.orderBy
```
jamesgorrie: Assuming this will be undefined at this point and defaulted on the server?
Do we want to be explicit for the app? Or do we want the default set by the server?

kahuna/public/js/search/query.html
akash1810: > It applies to the date range

It sort of does. Maybe this has been implemented the wrong way - there should be an option in the date picker to specify the field to use, then another option to specify the ordering. That way you can search for all images taken today, for example. Rather than all images uploaded today, but then order by date taken (which could be before the upload time).

I think https://github.com/guardian/grid/issues/946 is also required to make the dates on the thumbnails relevant for the order/filter you're using.

> could it live closer to the results, by the result count

Arguably, the result count would have to be sticky for this to happen.

> In terms of the label - you could try adding "order by ____" in the options to save space?

Don't like...

screen shot 2015-07-28 at 10 34 39

kahuna/public/js/search/query.js
akash1810: Currently its not supported by the API, but should be simple enough.

kahuna/public/js/search/query.js
jamesgorrie: Joanna Ruck asked for it for pics of the day as they search, go to the bottom, and then go up.

alastairjardine:
Some UI changes before rollout, please :)

I'm not sure that a dropdown is the best UI control here. It would be much better to pull these out into separate buttons. If the current UI doesn't support that, we need to sort the UI so it can.

jamesgorrie:
> pull these out into separate buttons

Could you elaborate / show an example of? Not sure I understand.

alastairjardine:
Radios better for limited numbers of mutually exclusive options. The list isn't large enough to warrant a dropdown.

kahuna/public/js/search/query.html
akash1810: Re saving space, Drive's solution:

screen shot 2015-07-28 at 11 21 42


Not recommending, but its interesting. I like that its size is small (just an icon) but I don't like that its not immediately obvious what you've got selected - you have to open the dropdown and see what's ticked.

akash1810:
Radio buttons in the date filter drop-down. Feels a bit busy to me:

screen shot 2015-07-28 at 11 59 04

jamesgorrie:
I like the bluntness of "What" -> But "Order by" is probably right.
If you select one of the "Whats" will Dec be selected automatically.

theefer:
I find this rather confusing to put in the date range picker. Wouldn't mind chatting about it but I'm in a meeting until 1pm

alastairjardine:
Yeah, let's all chat about this. There may be a nice way to achieve this some of the work @kenoir is looking to pick up soon re: toolbars etc. Iconography would work well here, amongst other things to improve the visual appeal.

paperboyo:
> The list isn't large enough to warrant a dropdown.

But isn't it very likely to grow (dateModified, resolution, fileType, prominenceOnFronts etc.)?

Sorting indeed feels to me a thing detached from filters (a "postfilter", @theefer well said). That's why I can't see it residing within a date picker (which, itself, will likely to be used in many more places to set date ranges for different kind of temporal metadata). Most, if not all, filters should be available as a sorting option.

>> could it live closer to the results, by the result count

> Arguably, the result count would have to be sticky for this to happen.

Ah yes. The Status bar :-) (Lightroom, Finder, Windows Explorer, everything else). While filters/search are not visible/collapsed this, when visible (not hidden), always provides a hint on "what am I seeing now" + "what do I have selected/actionable" (can also provide a shortcut to specific filters/views when a breadcrumb)

[x new, y new matching current search/filters; All>modified:DD-MM-DD-MM>country:z; keyword:a>b available (c selected)]

alastairjardine:
>But isn't it very likely to grow (dateModified, resolution, fileType, prominenceOnFronts etc.)?
Possibly, but they're different things, which I'd like to tackle in a different manner. 

It shouldn't live in the date picker, please.

akash1810:
lets talk about this today

akash1810:
Moved it to the same level as the total image count and only ordering by the upload field in the first instance:

screen shot 2015-07-31 at 13 11 09

/cc @alastairjardine

kahuna/public/stylesheets/main.css
```diff
@@ -1751,6 +1764,7 @@ ui-archiver .archiver:hover {
    background-color: #444;
    text-align: center;
    border: 1px solid #555;
+   cursor:pointer;
```
theefer: Missing space?

kahuna/public/js/search/results.js
```diff
@@ -283,6 +286,20 @@ results.controller('SearchResultsCtrl', [
             }
         };
 
+        ctrl.filter = {
+            sortDirection: $stateParams.orderBy === 'uploadTime' ? 'ASC' : 'DESC'
+        };
+
+        $scope.$watch(() => ctrl.filter.sortDirection, onValChange(newVal => {
+            var filter = {orderBy: undefined};
```
theefer: Just to check, is that different from `{}`?

kahuna/public/js/search/query.html
```diff
@@ -62,7 +62,7 @@
                                gu:preset-dates="searchQuery.sinceOptions"
                                gu:first-day="1">
                 </gu-date-range>
-            <li>
+            </li>
```
theefer: Oops, good catch!

kahuna/public/js/search/results.html
```diff
@@ -13,14 +13,43 @@
...
+                <div class="radio-list">
+                    <div class="radio-list__item">
+                        <input type="radio"
+                               id="sort-direction__desc"
+                               class="radio-list__circle"
+                               name="sort-direction"
+                               value="DESC"
+                               ng:model="ctrl.filter.sortDirection">
+                        <label for="sort-direction__desc" class="radio-list__label">
+                            Most Recent First
```
theefer: I suggested "Newest first" to @akash1810 instead.

cc @alastairjardine

alastairjardine: "Most recent first" (sentence case)

theefer:
:+1:

kahuna/public/js/search/results.html
```diff
@@ -13,14 +13,43 @@
...
+
+                    <div class="radio-list__item">
+                        <input type="radio"
+                               id="sort-direction__asc"
+                               class="radio-list__circle"
+                               name="sort-direction"
+                               value="ASC"
+                               ng:model="ctrl.filter.sortDirection">
+                        <label for="sort-direction__asc" class="radio-list__label">
+                            Oldest First
```
alastairjardine: "Oldest first" (sentence case)

kahuna/public/js/search/results.html
alastairjardine: @theefer Yep, that also works (and is shorter!). Sentence case whatever it happens to be!

alastairjardine:
@akash1810 Did you try with an icon? Would be interested to see how it looks. Other than the minor sentence case stuff :+1: Nice work

kahuna/public/stylesheets/main.css
jamesgorrie: Missing lint?

kahuna/public/stylesheets/main.css
```diff
@@ -749,6 +749,20 @@ input.search-query__input {
...
+    float: left;
+}
+
+.image-results-count,
+.sort-direction {
+    display: inline-block;
+}
+
+.sort-direction {
+    float: right;
```
jamesgorrie: Not liking floats, flex has nearly always been better...
Up to you though.

kahuna/public/js/search/results.html
```diff
@@ -13,14 +13,43 @@
...
+                </button>
+            </div>
+            <div class="sort-direction">
+                <div class="radio-list">
+                    <div class="radio-list__item">
+                        <input type="radio"
+                               id="sort-direction__desc"
+                               class="radio-list__circle"
+                               name="sort-direction"
+                               ng:value="undefined"
```
akash1810: keep the url clean! :smile:

kahuna/public/js/search/results.html
```diff
@@ -13,14 +13,43 @@
...
+                            Newest first
+                        </label>
+                    </div>
+
+                    <div class="radio-list__item">
+                        <input type="radio"
+                               id="sort-direction__asc"
+                               class="radio-list__circle"
+                               name="sort-direction"
+                               ng:value="'oldest'"
```
akash1810: Query param becomes `orderBy=oldest`, with the hope that we'll soon support something like:

`orderOn=<uploadTime|lastModified|dateTaken>`

kahuna/public/js/search/results.js
```diff
@@ -135,14 +139,6 @@ results.controller('SearchResultsCtrl', [
...
         };
 
-        // Safer than clearing the timeout in case of race conditions
-        // FIXME: nicer (reactive?) way to do this?
-        var scopeGone = false;
-        $scope.$on('$destroy', () => {
-            scopeGone = true;
-        });
-
-
```
akash1810: moved to EOF

kahuna/public/js/search/results.js
```diff
@@ -157,11 +153,6 @@ results.controller('SearchResultsCtrl', [
             then(() => scrollPosition.resume($stateParams)).
             then(scrollPosition.clear);
 
-        $scope.$on('$destroy', () => {
-            scrollPosition.save($stateParams);
-        });
-
-
```
akash1810: moved to EOF

akash1810: moved to EOF

akash1810:
Everything but polling when sorted oldest first works now :-/

alastairjardine:
Got a couple of relatively minor stylin' tweaks that I'd like applied before release to make it more obvious which is the selected state.

akash1810:
the resulting css is horrendous, but something like this @alastairjardine?
order

alastairjardine:
Boom. Yep, that's nearly there @akash1810. Just needs a nadge more left & right padding to allow the text to breath. Top stuff (no pun intended).

theefer:
Looks nice

akash1810:
Side padding at 10px

screen shot 2015-08-04 at 17 22 18

I'll make the radio list a component soon; probably not part of this PR.

kahuna/public/js/search/results.js
```diff
@@ -86,8 +90,23 @@ results.controller('SearchResultsCtrl', [
...
+            /*
+            Maintain `lastSearchFirstResultTime` regardless of sorting order.
+
+            If we're sorting in ascending order, we need to get the upload time of the
+            last image, so we make a request for 1 image, then make a further request
+            where the offset is the total images - 1 from the initial request.
+             */
+            return angular.isUndefined(ctrl.filter.orderBy) ?
+                ctrl.searched = search({length: 1}) :
+                ctrl.searched = search({length: 1}).then((images) => {
```
theefer: As mentioned orally, small risk of race condition here if the new results have come in before the second search, although we don't think it's break the UI.

theefer:
:+1:

PR #990, opened 2015-07-27
actually upgrade to 1.4

jamesgorrie:
For some reason, jspm didn't update my config?
I will try to find out why, but a reset worked.

### DIFF
--- a/kahuna/public/config.js
+++ b/kahuna/public/config.js
@@ -11,25 +11,25 @@ System.config({
 
 System.config({
   "map": {
-    "angular": "github:angular/bower-angular@1.3.15",
-    "angular-animate": "github:angular/bower-angular-animate@1.3.15",
-    "angular-bootstrap": "github:angular-ui/bootstrap-bower@0.13.0",
-    "angular-elastic": "github:monospaced/angular-elastic@2.4.2",
-    "angular-messages": "github:angular/bower-angular-messages@1.3.15",
+    "angular": "github:angular/bower-angular@1.4.3",
+    "angular-animate": "github:angular/bower-angular-animate@1.4.3",
+    "angular-bootstrap": "github:angular-ui/bootstrap-bower@0.13.1",
+    "angular-elastic": "github:monospaced/angular-elastic@2.5.0",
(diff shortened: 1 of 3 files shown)


kahuna/public/js/components/gu-date-range/gu-date-range.js
```diff
@@ -90,7 +90,7 @@ guDateRange.directive('guDateRange', [function () {
             });
 
             $scope.$watch('pikaStartValue', function (pikaStartValue) {
-                var date = pikaStartValue && new Date(pikaStartValue);
+                var date = (pikaStartValue && new Date(pikaStartValue)) || new Date();
```
jamesgorrie: This is currently passing an empty string about, which is dealt with silently until now.

theefer: Is this causing an error? Have you tested that the behaviour is still as expected?

cc @akash1810 who worked on that bit.

theefer:
:+1:

kahuna/public/js/components/gu-date-range/gu-date-range.js
jamesgorrie: Yip - everything is also working as expected.
Passing an empty string to `Date` should error irrespective, so this code is more typesafe anyway.

kahuna/public/config.js
```diff
@@ -11,25 +11,25 @@ System.config({
...
     "jcrop": "github:tapmodo/Jcrop@0.9.12",
     "jquery": "github:components/jquery@2.1.1",
-    "mixpanel-js": "github:mixpanel/mixpanel-js@2.4.2",
-    "moment": "github:moment/moment@2.9.0",
+    "mixpanel-js": "github:mixpanel/mixpanel-js@2.5.2",
+    "moment": "github:moment/moment@2.10.5",
     "pandular": "npm:pandular@0.1.5",
-    "pikaday": "github:dbushell/Pikaday@1.3.2",
-    "raven-js": "github:getsentry/raven-js@1.1.18",
+    "pikaday": "github:dbushell/Pikaday@1.3.3",
```
jamesgorrie: This was updates which doesn't accept bad dates any moer.

PR #996, opened 2015-07-28
Add `msd-elastic` attribute to special instructions in `gr-panel`.

akash1810:
[For reference](https://github.com/monospaced/angular-elastic)

We use it on the description field already.


theefer:
:+1:

PR #999, opened 2015-07-28
Preserve the same set of results when navigating back to a previous search

theefer:
This prevents new matches to shift your results when you click away and come back to the same vertical position of your search, by constraining the search when you come back to match the same set of results.

This proved quite fiddly and I tried a various range of solutions using the uiRouter, but I had real trouble persisting the state as part of the `$stateParams` without having it in the URI (which I don't think would have been appropriate). Pushing it into the `$stateParams` didn't seem to play well with the deep-state redirect, i.e. when navigating back to a previous search e.g. from a preview page (using DSR), the non-URL state wasn't recovered.

I finally resolved to using a global state in the controller (could be its own angular module, but seemed slightly overkill with no obvious benefit). I also had to expose whether the search controller is loaded as part of a new search or a redirect (recover an old search), to know whether to refresh the state or not.


theefer:
Nota bene: I hate this kind of stateful logic but can't find a way to escape it in this case :cry:

kahuna/public/js/search/results.js
```diff
@@ -19,6 +19,16 @@ function compact(array) {
...
 
+
+// Global session-level state to remember the uploadTime of the first
+// result in the last search.  This allows to always paginate the same
+// set of results, as well as recovering the same set of results if
+// navigating back to the same search.
+// Note: I tried to do this using non-URL $stateParams and it was a
+// rabbit-hole that doesn't seem to have any end. Hence this slightly
+// horrid global state.
+let lastSearchFirstResultTime;
```
jamesgorrie: We could store it in a factory perhaps that contains the state?

kahuna/public/js/search/results.js
```diff
@@ -64,18 +76,38 @@ results.controller('SearchResultsCtrl', [
...
 
+        // If not reloading a previous search, discard any previous
+        // state related to the last search
+        if (! isReloadingPreviousSearch) {
+            lastSearchFirstResultTime = undefined;
+        }
+
         // TODO: avoid this initial search (two API calls to init!)
-        ctrl.searched = search({length: 0}).then(function(images) {
+        ctrl.searched = search({length: 1}).then(function(images) {
```
jamesgorrie: Why the `1`?

theefer: So we can [pull out the uploadTime of the first result](https://github.com/guardian/grid/pull/999/files#diff-b8c6d4770b5b04294cc8c88d27ae65d3R107) in this very first request.

jamesgorrie: Ah yes.

kahuna/public/js/search/results.js
theefer: Yes I thought so, but I couldn't really find any reason to justify the accompagnying boilerplate. Can you think of any?

One would be to help mock and write tests but we don't do that at the moment...

jamesgorrie: Besides taking up abou as much space as the comment, not really.

jamesgorrie:
:+1:

kahuna/public/js/search/results.js
theefer: The comment still needs to be there to explain the rationale either way...

PR #1000, opened 2015-07-28
Expose delete argo action, take 2

theefer:
Same as #987 but with a tiny theseus change (https://github.com/argo-rest/theseus/pull/17) to avoid breaking the client.


jamesgorrie:
:+1:

PR #1001, opened 2015-07-29
use a form that is ownedby us

jamesgorrie:
You can't transfer between domains and it was unfortunately / accidentally created on my personal account.


theefer:
Presumably it also notifies the list of submissions? :+1:

jamesgorrie:
Yip.

PR #1002, opened 2015-07-30
Make select elements full width in usage rights editor.

akash1810:
Because its a bit jumpy.

Before:
full-width

After:
screen shot 2015-07-30 at 14 17 41

### DIFF
--- a/kahuna/public/stylesheets/main.css
+++ b/kahuna/public/stylesheets/main.css
@@ -1752,10 +1752,14 @@ ui-archiver .archiver:hover {
    text-align: center;
    border: 1px solid #555;
 }
-.radio-list__circle:checked + .radio-list__label { 
+.radio-list__circle:checked + .radio-list__label {
    background-color:#666;
    border: 1px solid #777;
 }
 .radio-list__circle {
    display: none;
+}
+
+.full-width {
+    width: 100%;
 }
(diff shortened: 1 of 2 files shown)


kahuna/public/js/usage-rights/usage-rights-editor.css
```diff
@@ -35,3 +35,7 @@
 .ure__multiple-rights-warning {
     color: orange;
 }
+
+.ure--full-width-field {
```
theefer: `--` is a modifier state. Not sure you even need to namespace with `ure` here, but if you want to, you could use either `ure-full-width-field` or `ure__full-width-field`.

kahuna/public/stylesheets/main.css
```diff
@@ -1752,10 +1752,14 @@ ui-archiver .archiver:hover {
    text-align: center;
    border: 1px solid #555;
 }
-.radio-list__circle:checked + .radio-list__label { 
+.radio-list__circle:checked + .radio-list__label {
```
akash1810: thanks intellij...

kahuna/public/js/usage-rights/usage-rights-editor.css
akash1810: > Not sure you even need to namespace with ure here

Agreed - renamed to `full-width` and put in `main.css`.

theefer:
:+1:

PR #1005, opened 2015-07-30
Updates readme

tsop14:
removes bower and adds imagemagick from  installation requirements in readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,10 +32,9 @@ You will need to install:
 * sbt
 * JDK 8
 * Nginx
-* [bower](http://bower.io/)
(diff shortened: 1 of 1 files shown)


theefer:
:+1: 

cc @kenoir what was the conclusion on using GraphicsMagick for DEV on Macs in the end?

kenoir:
I would prefer to use GraphicsMagick as there are inconsistencies in how it behaves compared to ImageMagick (when used through Im4Java at least). Making changes to anything that modifies images can be a pain otherwise, as you might not get what you expect in production if you use ImageMagick.

PR #1006, opened 2015-07-30
Avoid superfluous vertical scrollbar

theefer:
# Before

image

# After

image

It's almost like we could do with CSS variables :sweat:


akash1810:
:+1:

PR #1007, opened 2015-07-31
Appalling hack to avoid jcrop.gif 404s

theefer:
It really pains me to write this but not quite as much as getting erroneous alarms on kahuna. Hopefully can get this fixed in the CSS plugin soon.


akash1810:
:+1:

PR #1010, opened 2015-07-31
Allow scrolling through 20k images if no query

theefer:
Cap to 20,000 images when browsing results with no search query (i.e. everything). This allows people who want to see everything (typically a handful of picture editors) to browse about one day's worth of images.

For everyone else, we keep the 5,000 image limit whenever there are search terms so as to keep the scrollbar less sensitive.

This seems like the simplest solution to accommodate both use cases.


jamesgorrie:
:+1:

PR #1011, opened 2015-07-31
Upgrade panda library to 0.2.7

theefer:
This new version drops the new cookie for the upcoming asymmetric auth. The old cookie is still issued and used for verification.


theefer:
cc @lindseydew @adamnfish

jamesgorrie:
:+1:

PR #1012, opened 2015-07-31
Fix leaking and errors on image update

theefer:
1. We should always free global listeners in controllers.
2. imagesAll can contain holes (`undefined`) so we need to protect against it (else we're seeing `Cannot read property 'data' of undefined`)


theefer:
Oddly the `image-updated` event seems to be triggered twice, not sure why.

theefer:
Also added another fix for an error that occurs every time the gr-panel is closed.

jamesgorrie:
:+1: Nice catches.

jamesgorrie:
Standing by my :+1:

PR #1013, opened 2015-07-31
New Cost Model: Deal with publication usage rights on old images.

jamesgorrie:
Making sure we set images to staff, contract or commissioned photographers.

I am waiting on a list that is comprehensive from Robert H with both staff and contracted, so I can search for those, and fallback to commissioned if they are in neither. This can be done as a second PR, we must just make sure we wait for that before reindexing.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
@@ -49,14 +49,6 @@ object PhotographerParser extends ImageProcessor {
           metadata    = image.metadata.copy(credit = Some(publication))
         )
 
-        // commissioned photographer
-        case (byline, Some("the guardian"), None, None) => image.copy(
-          usageRights = CommissionedPhotographer(byline, "The Guardian")
-        )
-        case (byline, Some("the observer"), None, None) => image.copy(
-          usageRights = CommissionedPhotographer(byline, "The Observer")
-        )
-
         case _ => image
       }
     }
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
@@ -21,51 +21,48 @@ object SupplierProcessors {
     PaParser,
     ReutersParser,
     RexParser,
-    StaffPhotographerParser
+    PhotographerParser
   )
 
   def process(image: Image): Image =
     all.foldLeft(image) { case (im, processor) => processor(im) }
 }
 
-object StaffPhotographerParser extends ImageProcessor {
-  def apply(image: Image): Image = {
-    image.metadata.byline.map(p => (p, PhotographersList.getPublication(staffPhotographers, p))) match {
-      case Some((photographer, Some(publication))) => image.copy(
-        usageRights = StaffPhotographer(photographer, publication),
-        metadata    = image.metadata.copy(credit = Some(publication))
-      )
-      case _ => image
-    }
-  }
-}
-
-object ContractedPhotographerParser extends ImageProcessor {
-  def apply(image: Image): Image = {
-    image.metadata.byline.map(p => (p, PhotographersList.getPublication(contractedPhotographers, p))) match {
-      case Some((photographer, Some(publication))) => image.copy(
-        usageRights = ContractPhotographer(photographer, publication),
-        metadata    = image.metadata.copy(credit = Some(publication))
-      )
-      case _ => image
-    }
-  }
-}
-
-object CommissionedPhotographerParser extends ImageProcessor {
+object PhotographerParser extends ImageProcessor {
   def apply(image: Image): Image = {
     image.metadata.byline.map { byline =>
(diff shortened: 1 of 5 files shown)


jamesgorrie:
:dromedary_camel:

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -20,25 +21,49 @@ object SupplierProcessors {
...
+        case (byline, Some("the observer"), None, None) => image.copy(
+          usageRights = CommissionedPhotographer(byline, "The Observer")
+        )
+
+        case _ => image
+      }
     }
-  }
+  }.getOrElse(image)
 }
```
jamesgorrie: [I had these as separate parsers](https://github.com/guardian/grid/commit/8f6649aeca6174e6e280a707d6571a0ba4af11ab#diff-4b854d8f2e0bc0725c5b4804936b7875R43), but they are actually intrinsically the same, this is more readable, and contains repeated logic.

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -20,25 +21,49 @@ object SupplierProcessors {
...
+          metadata    = image.metadata.copy(credit = Some(publication))
+        )
+
+        // commissioned photographer
+        case (byline, Some("the guardian"), None, None) => image.copy(
+          usageRights = CommissionedPhotographer(byline, "The Guardian")
+        )
+        case (byline, Some("the observer"), None, None) => image.copy(
+          usageRights = CommissionedPhotographer(byline, "The Observer")
+        )
```
theefer: Can you explain the rationale behind these two?

theefer: IOW unless the other lists are exhaustive, are we not going to incorrectly mark staff/contract photographers as commissioned?

common-lib/src/test/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessorsTest.scala
```diff
@@ -20,13 +20,27 @@ class SupplierProcessorsTest extends FunSpec with Matchers with MetadataHelper {
...
     }
+
+    it("should match ContractPhotographer byline") {
+      val image = createImageFromMetadata("byline" -> "Linda Nylind")
+      val processedImage = applyProcessors(image)
+      processedImage.usageRights should be(ContractPhotographer("Linda Nylind", "The Guardian"))
+      processedImage.metadata.credit should be(Some("The Guardian"))
+    }
+
+    it("should match CommissionedPhotographer byline") {
```
theefer: It's not so much the byline as the credit?

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
jamesgorrie: Some images that come through from the past are not on our staff / contract lists as we just don't
have all the data from RCS.

This will then account for those not on those lists, but we have explicitly set the credit to be ourselves.

theefer: Would be interested to see the data to alleviate my irrational fear that this would aggressively attribute pictures to commissioned photographers when they should be contract/staff.

I've just had a related conversation with Jo about import picdar library copyright groups, esp for commissioned photographers, and using that to feed the new model. We concluded it'd be good to have someone work with her on any extra data that can't be automated.

jamesgorrie: > can't be automated
Do you mean not from the current metadata, but could from the copyright groups? It'd also be good to see how many we would catch with that compared to anywhere else as that would be once off work that would be thrown out after reingest.

jamesgorrie: I'll also wait for the list to do a query on the API to get numbers

theefer: I think it'd be easier to have a chat about this.

Commissioned cannot be automated from a list of bylines because commissioned people typically also have sent us pictures that were *not* part of a commission, and they shouldn't be marked as commissioned.

jamesgorrie: Yip - which is why we're not making the assumption through the byline, but when the credit says itwas for us.

common-lib/src/test/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessorsTest.scala
jamesgorrie: Agreed - updated

jamesgorrie:
@theefer removed commissioned photographer logic.

theefer:
:+1:

PR #1016, opened 2015-08-03
Show delete button on image page if allowed

theefer:
This allows a handful of allowed people to delete images without the clunky "Delete from Grid" bookmarklet.

image

This feature makes use of the new `actions` in the argo responses :fireworks:

### DIFF
--- a/kahuna/public/js/image/controller.js
+++ b/kahuna/public/js/image/controller.js
@@ -16,6 +16,7 @@ var image = angular.module(
 image.controller('ImageCtrl', [
     '$rootScope',
     '$scope',
+    '$state',
     'onValChange',
     'image',
     'mediaApi',
@@ -29,6 +30,7 @@ image.controller('ImageCtrl', [
 
     function ($rootScope,
               $scope,
+              $state,
               onValChange,
               image,
               mediaApi,
@@ -130,6 +132,10 @@ image.controller('ImageCtrl', [
         }
 
         function updateAbilities(image) {
+            mediaApi.canDelete(image).then(deletable => {
+                ctrl.canBeDeleted = deletable;
+            });
+
             mediaCropper.canBeCropped(image).then(croppable => {
                 ctrl.canBeCropped = croppable;
             });
@@ -168,6 +174,17 @@ image.controller('ImageCtrl', [
                      */
                     return 'failed to save (press esc to cancel)';
                 });
+        };
+
+        ctrl.delete = function() {
+            const confirmed = window.confirm('Are you sure you want to delete this image from the Grid?');
+            if (confirmed) {
+                mediaApi.delete(image).then(() => {
+                    window.alert('The image will be deleted shortly!');
+                    // Can't stay on the page of a deleted image
+                    $state.go('search');
+                });
+            }
         };
 
         $scope.$on('$destroy', function() {
--- a/kahuna/public/js/image/controller.js
+++ b/kahuna/public/js/image/controller.js
@@ -177,6 +177,7 @@ image.controller('ImageCtrl', [
         };
 
         ctrl.delete = function() {
+            // TODO: use inline confirmation as per other tools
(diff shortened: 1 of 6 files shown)


kahuna/public/js/image/controller.js
```diff
@@ -170,6 +176,17 @@ image.controller('ImageCtrl', [
                 });
         };
 
+        ctrl.delete = function() {
+            const confirmed = window.confirm('Are you sure you want to delete this image from the Grid?');
```
jamesgorrie: We have confirm delete pattern [here](https://github.com/guardian/grid/blob/master/kahuna/public/js/components/gr-confirm-delete/gr-confirm-delete.js) that matches the rest of the tools.

kahuna/public/js/image/controller.js
```diff
@@ -170,6 +176,17 @@ image.controller('ImageCtrl', [
...
+            if (confirmed) {
+                mediaApi.delete(image).then(() => {
+                    window.alert('The image will be deleted shortly!');
```
jamesgorrie: You could probably extend [the above](https://github.com/guardian/grid/pull/1016/files#r36070673) to have a confirmation message.

media-api/app/controllers/MediaApi.scala
```diff
@@ -148,8 +148,7 @@ object MediaApi extends Controller with ArgoHelpers {
           canUserDeleteImage(request, source) map { canDelete =>
             if (canDelete) {
               Notifications.publish(Json.obj("id" -> id), "delete-image")
-              // TODO: use respond
-              Accepted.as(ArgoMediaType)
+              Accepted
```
jamesgorrie: Could we return the "The image will be deleted shortly!" to the client as a `confirmation-message` or siilar?

theefer: Any rationale for returning the message? Isn't that what `Accepted` means?

jamesgorrie: I was thinking that the fact that the image won't be completely deleted until the message is consumed would be the reason, but realise that that's actually coupling the systems... never mind me.

alastairjardine:
Good, but we should add a confirmation as this action is relatively destructive and hard to back out of. Would prefer our usual two-phase delete, but at a pinch we could compromise on an alert box with some appropriate wordage.

jamesgorrie:
@alastairjardine [the confirmation is already there](https://github.com/guardian/grid/pull/1016#discussion-diff-36070726R180), I've built a confirm delete from last time and suggested it to @theefer, it might be easy to wedge that in.

theefer:
There is a confirmation using a browser confirm window. @jamesgorrie pointed at a reusable component for that so I will try and use that instead if possible.

alastairjardine:
Ahhh, thanks @theefer @jamesgorrie.

kahuna/public/js/image/controller.js
```diff
@@ -170,6 +176,18 @@ image.controller('ImageCtrl', [
                 });
         };
 
+        ctrl.delete = function() {
+            // TODO: use inline confirmation as per other tools
+            const confirmed = window.confirm('Are you sure you want to delete this image from the Grid?');
+            if (confirmed) {
+                mediaApi.delete(image).then(() => {
+                    window.alert('The image will be deleted shortly!');
```
alastairjardine: Drop the '!'.

theefer:
I spent some time trying to get the existing (though unused) directive to do the right thing, and when that failed, to write another one to do the right thing. It's fiddly for a variety of reasons and I'd rather leave that inline confirmation for later rather than doing a botched job of it and pretend it's solved.

kahuna/public/js/image/controller.js
theefer: Oops didn't see that before I merged, I did that on master directly.

jamesgorrie:
:+1: then - I'll have a bash at it.

Issue #1015, opened 2015-07-31
Editing a selected image leaves selection in buggy state


theefer:
Steps to reproduce:
1. Open search results
2. Select an image
3. Edit the description
4. Save

Expected:
The description is altered and the image is still selected (blue border) and the panel open.

Actual:
The description is altered, the panel remains open but the image isn't visually selected (although the panel still says "1 image selected").

This is quite confusing so would be good to fix ASAP.

PR #1019, opened 2015-08-03
remove extra event emission.

akash1810:
Fixes https://github.com/guardian/grid/issues/1015.

As of https://github.com/guardian/grid/commit/4423078cec5109430ebeb0ac676407203c6e0642#diff-17227494f078d3840f7fa4a485a88e2fR112 we're emitting the `image-updated` event in the `update` function.

Firing it again in `updateMetadataField` causes [this](https://github.com/guardian/grid/issues/1015) behaviour.

### DIFF
--- a/kahuna/public/js/edits/service.js
+++ b/kahuna/public/js/edits/service.js
@@ -254,7 +254,6 @@ service.factory('editsService',
         return update(image.data.userMetadata.data.metadata, changed, image)
             .then(() => {
                 return image.get().then(updatedImage => {
-                    $rootScope.$emit('image-updated', updatedImage, image);
                     return updatedImage;
                 });
             });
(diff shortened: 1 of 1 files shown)


kahuna/public/js/edits/service.js
```diff
@@ -254,7 +254,6 @@ service.factory('editsService',
         return update(image.data.userMetadata.data.metadata, changed, image)
             .then(() => {
                 return image.get().then(updatedImage => {
-                    $rootScope.$emit('image-updated', updatedImage, image);
                     return updatedImage;
                 });
```
theefer: The `then` is no longer necessary?

theefer:
Good catch, minor comment toward further simplication :+1:

kahuna/public/js/edits/service.js
akash1810: this is true!

PR #1021, opened 2015-08-03
remove extra event emission again.

akash1810:
See https://github.com/guardian/grid/pull/1019.

We use the image resource that's returned in `then` [here](https://github.com/guardian/grid/blob/master/kahuna/public/js/image/controller.js#L156), so https://github.com/guardian/grid/pull/1019 wasn't entirely correct!

We should probably unify the way we use the `update` function - either listen to an event, or use the promise (the batch editing uses the promise).


theefer:
:+1: :smile:

PR #1022, opened 2015-08-03
Archive batch editor

jamesgorrie:
A smaller version of #975.
Using a new stream of images to be able to do things.

After a chat with @kenoir we thought this was better than having the strange event based version we have now.

__Next:__
- [x] Deprecate old archive service
- [ ] apply to other editors


kahuna/public/js/image/streamService.js
```diff
@@ -0,0 +1,48 @@
+import Rx from 'rx';
+
+function imageStream(images_) {
+
+    const images = new Set();
```
kenoir: Why convert to a set here when we just convert back to an array below?

kenoir:
Much nicer that what we currently have.

kahuna/public/js/edits/archiver.js
```diff
@@ -1,41 +1,54 @@
...
-            archiveService.unarchive(ctrl.image) :
-            archiveService.archive(ctrl.image);
+    function error() {
+        $window.alert('Failed to save the changes, please try again.')
+    }
 
-        promise
-            .catch(()  => $window.alert('Failed to save the changes, please try again.'))
-            .finally(() => ctrl.archiving = false);
+    function saved() {
```
akash1810: does this need to be a function, why not set `ctrl.saving` in the `finally` of the promises above?

akash1810:
As were dealing with streams, can we deprecate the use of `batch-x` classes?

kahuna/public/js/image/streamService.js
jamesgorrie: It's easier to add  / delete form a set than from an array... this is more of an internal representation, hiding it from clients of this as they shouldn't have to worry how things are added / removed.

We then return an array as that where you get to do the nice functional stuff to it.

jamesgorrie:
@akash1810 I've left the styling etc as this stuff is moving soon and opens this up to be a buggy PR.

kahuna/public/js/edits/archiver.js
jamesgorrie: It seemed a repeated bit of logic (albeit a small bit of logic) and seemed neater - happy to know it back out.

jamesgorrie:
Thumbs anyone?

theefer:
Damn forgot about this PR, will have a read today.

kahuna/public/js/archiver/archiver.js
```diff
@@ -0,0 +1,42 @@
...
+    ctrl.count = 0;
+    ctrl.notArchivedCount = 0;
+    ctrl.archivedCount = 0;
+    ctrl.hasMixed = false;
+    ctrl.archiving = false;
+
+    ctrl.service.count$.subscribe(i => ctrl.count = i);
+    ctrl.service.archivedCount$.subscribe(i => ctrl.archivedCount = i);
+    ctrl.service.notArchivedCount$.subscribe(i => ctrl.notArchivedCount = i);
+    ctrl.service.hasMixed$.subscribe(hasMixed => ctrl.hasMixed = hasMixed);
```
theefer: doesn't subscribe fire with the initial value, so you don't have to initialise above?

theefer: also subscribe will NOT be freed when this controller is destroyed. have a look at my rx helpers (in the lazy table), there is a helpful subscribe helper that auto-frees itself.

theefer: Given this pattern of subscribe to inject as a value on the scope is growing, we could even have another helper layer on top like:

`inject$(scope, 'count', ctrl.service.count$)`

kahuna/public/js/image/stream.js
```diff
@@ -0,0 +1,48 @@
+import Rx from 'rx';
+
+function imageStream(images_) {
+
+    const images = new Set();
+    const images$ = new Rx.BehaviorSubject([]);
+
+    if (images_) {
+        images_.forEach(image => images.add(image));
```
theefer: Can this be passed to the `Set` constructor instead?

kahuna/public/js/archiver/archiver.js
jamesgorrie: Nice, thanks.

kahuna/public/js/image/stream.js
jamesgorrie: It could but I didn't want to have a
```JavaScript
if (images_) {
   const images = new Set(images_)
} else {
  const images = new Set();
}
```

Seems to defeat the point of `const`.

theefer: Could `images_` just be defaulted to `[]` in the function parameter list?

jamesgorrie: I forget JS is :bomb: now.

kahuna/public/js/image/stream.js
```diff
@@ -0,0 +1,47 @@
+import Rx from 'rx';
+
+function imageStream(images_ = []) {
+
+    const images = new Set(images_);
```
theefer: It's tempting to use an immutable set (from immutable-js?) here in the future, would make updates a lot more sexily functional.

kahuna/public/js/edits/archiver.js
```diff
@@ -1,41 +1,53 @@
...
 
-    $scope.$watch(() => ctrl.image.data.userMetadata.data.archived.data, onValChange(newState => {
-        ctrl.isArchived = newState;
-    }));
+    ctrl.archive = () => {
+        ctrl.saving = true;
+        service.
+            archive().
+            catch(error).
+            finally(saved);
```
theefer: Detail, but I find it easier to read if the same locality of code is responsible for setting/clearing the saving flag. Either by mutating it here both times, or having two helper functions defined near each other that set and reset it.

theefer: Alternatively can be a helper on the promise: `injectSaving(promise)`

PR #1023, opened 2015-08-03
add filename to image case class

### LINKED ISSUE #1004: Index and expose original filename
We resisted it for a while but talking to users it's actually necessary to retain the original filename of the uploaded file in the image metadata. The reason is that often people receive images, and then later on more instructions or descriptions for these images, and the only way to relate the two is the filename. Currently the filename isn't kept or shown anywhere.

**Question: what to do if an image is re-uploaded?**
I suggest we override with the new filename if provided, and keep the previous one if not.

jamesgorrie:
Fixes #1004.
Saving the filename if it's there.

Should we be doing:

```JavaScript
{
  "uploadInfo": {
    "filename": "thingamajig.jpg"
  }
}
```

### DIFF
--- a/image-loader/app/controllers/Application.scala
+++ b/image-loader/app/controllers/Application.scala
@@ -114,7 +114,7 @@ class ImageLoader extends Controller with ArgoHelpers {
 
     Logger.info(s"Received ${uploadRequestDescription(uploadRequest)}")
 
-    val supportedMimeType = Config.supportedMimeTypes.contains(mimeType_)
+    val supportedMimeType = Config.supportedMimeTypes.contains(Some(mimeType_))
 
     if (supportedMimeType) storeFile(uploadRequest) else unsupportedTypeError(uploadRequest)
   }
--- a/image-loader/app/controllers/Application.scala
+++ b/image-loader/app/controllers/Application.scala
@@ -45,8 +45,8 @@ class ImageLoader extends Controller with ArgoHelpers {
   val indexResponse = {
     val indexData = Map("description" -> "This is the Loader Service")
     val indexLinks = List(
-      Link("load",   s"$rootUri/images{?uploadedBy,identifiers,uploadTime}"),
-      Link("import", s"$rootUri/imports{?uri,uploadedBy,identifiers,uploadTime}")
+      Link("load",   s"$rootUri/images{?uploadedBy,identifiers,uploadTime,filename}"),
+      Link("import", s"$rootUri/imports{?uri,uploadedBy,identifiers,uploadTime,filename}")
     )
     respond(indexData, indexLinks)
   }
--- a/image-loader/app/controllers/Application.scala
+++ b/image-loader/app/controllers/Application.scala
@@ -55,16 +55,16 @@ class ImageLoader extends Controller with ArgoHelpers {
   def createTempFile(prefix: String) = File.createTempFile(prefix, "", new File(Config.tempDir))
 
 
-  def loadImage(uploadedBy: Option[String], identifiers: Option[String], uploadTime: Option[String]) =
-    AuthenticatedUpload.async(DigestBodyParser.create(createTempFile("requestBody")))(loadFile(uploadedBy, identifiers, uploadTime))
+  def loadImage(uploadedBy: Option[String], identifiers: Option[String], uploadTime: Option[String], filename: Option[String]) =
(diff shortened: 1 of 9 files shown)


theefer:
Do the Mappings needs updating?

jamesgorrie:
I was going to see what people thought about this API shape?
It feels like it should probably be on the root?

jamesgorrie:
Do you think something like:
```JSON
{
  "id": "asmaklsjdalskjdlkasjdklasd",
  "fileData": { "filename": "happyclap.jpg" }
}
```

might make more sense?

theefer:
Any other information we'd be expecting you can think of?

Also while we're :bike::house:ing, would something like uploadFilenamed/uploadedFilename/originalFilename be more descriptive?

jamesgorrie:
Thinking about it - what about sticking it in the identifiers block as `uploadedId`?

theefer:
Not sure what I think of non-unique identifiers...

jamesgorrie:
Good point.

jamesgorrie:
I have been searching and it does seem anyone else really keeps the filename - except Flickr in the title field.

What about having something like this that might be useful later?
```JavaScript
{
  "uploadInfo": {
    "originalTime": "2015-08-04T09:04:27:67Z",
    "latestTime": "2015-12-04T14:20:33:54Z",
    "method": "FTP | web | import", // this would allow for a "uploaded by users" filter
    "file": { "name": "thingamajig.jpg" }
  }
}
```

theefer:
> it doesn't seem anyone else really keeps the filename

Getty exposes it in their magical XMP as Original Filename.

I'm keen to also adopt a naming that makes it more of an anecdotal piece of gathered submission metadata rather than an attribute of the image object, which by definition doesn't have a file or a name.


In your sample JSON, what else would `file` include in the future? If we can't think of anything else, can we just store the filename as plain property?

jamesgorrie:
> filename as plain property

That makes more sense as it's something that actually travels with the upload request - not drawn from the file itself. Updated.

jamesgorrie:
Updated in description that is.

jamesgorrie:
> I'm keen to also adopt a naming that makes it more of an anecdotal piece of gathered submission metadata rather than an attribute of the image object, which by definition doesn't have a file or a name.

Yip - This is more like `lastUpdated`, which is more metadata to the record that to what the record is representing (i.e. an image) - does the shape above feel good to run with?

theefer:
Method seems a bit odd to me, but the rest is OK. How would you propose moving to that model from what we have now though?

theefer:
It's also missing the "by" bit btw. Call me lazy, but I can't help wonder if the shape change is really such an improvement that it's  worth the hassle just for adding filename...?

jamesgorrie:
Yea it is - the reason I thought for investigating this is because having filename on the root seems a bit :hankey:. It also looked like a good opportunity to look at adding the method by which something was uploaded.

We could do it in two different steps, but what about:
``` JavaScript
{
  "uploadInfo": {
    "method": "FTP | web | import", // this would allow for a "uploaded by users" filter
    "filename": "thingamajig.jpg"
  }
}
```
else the PR stands as is.

theefer:
Let's chat about this tomorrow, would be good to get the nice filename work out!

kenoir:
+1 to having an `uploadInfo` block. Seems to fit nicely with other things we might specifically want to store about the images' ingest.

jamesgorrie:
Updated with the new `uploadInfo`.

image-loader/app/controllers/Application.scala
```diff
@@ -106,20 +108,22 @@ class ImageLoader extends Controller with ArgoHelpers {
...
       uploadedBy = uploadedBy_,
-      identifiers = identifiers_
+      identifiers = identifiers_,
+      uploadInfo = uploadInfo_
     )
 
     Logger.info(s"Received ${uploadRequestDescription(uploadRequest)}")
 
-    val supportedMimeType = Config.supportedMimeTypes.exists(Some(_) == mimeType_)
+    val supportedMimeType = Config.supportedMimeTypes.contains(mimeType_)
```
theefer: Does that work? I thought `mimeType_` was an `Option[String]` and `Config.supportedMimeTypes` a `List[String]`.

theefer:
Small question about unrelated refactor above. Else :+1: nice one!

image-loader/app/controllers/Application.scala
jamesgorrie: Strangely the compiler lets you do this - repaired and tested

PR #1024, opened 2015-08-03
Retire "your past 50 uploads" from the upload screen

theefer:
This also changes the "your recent uploads" link to simply load up your uploads (no free filter) in the search results. Happy to debate whether we still want that link for a while to get people used to this new flow, or whether to retire it altogether.

recorded

This is just one piece of the puzzle, I still think we can work on the uploads screen to make it clearer that images are immediately uploaded and it's a transient panel.

/cc @alastairjardine

### DIFF
--- a/kahuna/public/js/search/query.js
+++ b/kahuna/public/js/search/query.js
@@ -24,7 +24,8 @@ query.controller('SearchQueryCtrl',
                  function($scope, $state, $stateParams, onValChange , mediaApi, track) {
 
     var ctrl = this;
-    ctrl.filter = {
+    ctrl.filter = {};
+    ctrl.pseudoFilter = {
         uploadedByMe: false
     };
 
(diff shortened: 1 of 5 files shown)


akash1810:
Is there a need for both the checkbox and the link now?

jamesgorrie:
Worth getting rid of the word "recent" uploads.

I personally feel the editing in the panel should be made a wee bit easier - but it's worth running through that with users.

Could we get some tracking into the panel?

kenoir:
> Worth getting rid of the word "recent" uploads.

+1

theefer:
> Worth getting rid of the word "recent" uploads.

I initially did that, but then remembered that we'd added it to make it less confusing to have two controls labels exactly the same ("your uploads") very near to each other.

Shall we answer @akash1810's question first:

> Is there a need for both the checkbox and the link now?


Re

> Could we get some tracking into the panel?

Sure, what do we want to track? (Doesn't seem related to this PR though.)

jamesgorrie:
The tracking would be to see who uses the link and who uses the filter.
If we knew that we would know which one to get rid of first if we're looking to ease people into this.

If not we can just chose which one is better, and do it. Potentially the link makes more sense as you very rarely would want to see only free an my uploads at risk of not seeing all of them (I have seen people be confused by this before) and stops people getting into the habit of using the free to use filter as it's there as an emergency button really.

theefer:
Ah, I misunderstood, OK to check for clicks on that link.

Note that we can't really get rid of the filter completely, because there needs to be some kind of visual manifestation of the state of only seeing one's own uploads.

theefer:
Note that the link automatically unselects the "free to use" filter.

jamesgorrie:
> some kind of visual manifestation

You could have an on state for the uploads link - potentially making it no-longer a link.

>  automatically unselects the "free to use" filter

Sounds like that\s the way it should be so it's better than the filter.

theefer:
Added tracking now.

jamesgorrie:
So we keep both for now with tracking then choose which to deprecate?

theefer:
I think so. We all agree this is just a transitional state.

kahuna/public/js/search/query.js
```diff
@@ -65,18 +66,28 @@ query.controller('SearchQueryCtrl',
...
-        ctrl.filter.uploadedByMe = ctrl.uploadedBy === ctrl.user.email;
+        ctrl.pseudoFilter.uploadedByMe = ctrl.filter.uploadedBy === ctrl.user.email;
     });
 
+    // Bi-directional sync uploadedByMe <=> uploadedBy(my email)
+    $scope.$watch(() => $stateParams.uploadedBy, onValChange(uploadedBy => {
+        ctrl.pseudoFilter.uploadedByMe = uploadedBy === ctrl.user.email;
+    }));
+    $scope.$watch(() => ctrl.pseudoFilter.uploadedByMe, onValChange(uploadedByMe => {
+        ctrl.filter.uploadedBy = uploadedByMe ? ctrl.user.email : undefined;
```
jamesgorrie: Does this change the URL?

jamesgorrie:
If the answer to [my question](https://github.com/guardian/grid/pull/1024#discussion_r36206585) is yes, brillo :+1:

kahuna/public/js/search/query.js
theefer: No, this changes filter, and then the block below sets that into the state, which is reflected in the url.

theefer:
Not sure about this anymore, will rework and investigate more.

PR #1025, opened 2015-08-04
Making the `/usage-rights` endpoint resolve in all cases.

akash1810:
Stop getting 500s when asking for usage rights when there isn't any. - e.g:

[PROD-metadata-api]/metadata/79472bf085aecb85a2bb50fba591e1ea1cfdf774/usage-rights

Instead, respond with:

```json
{
    "data": {}
}
```

### DIFF
--- a/metadata-editor/app/controllers/EditsController.scala
+++ b/metadata-editor/app/controllers/EditsController.scala
@@ -144,8 +144,10 @@ object EditsController extends Controller with ArgoHelpers {
 
   def getUsageRights(id: String) = Authenticated.async {
     dynamo.jsonGet(id, "usageRights").map { dynamoEntry =>
-      val usageRights = (dynamoEntry \ "usageRights").as[UsageRights]
-      respond(usageRights)
+      (dynamoEntry \ "usageRights").asOpt[UsageRights] match {
+        case Some(usageRights: UsageRights) => respond(usageRights)
+        case _ => respond(NoRights)
+      }
     }
   }
 


jamesgorrie:
@theefer what does argo-spec suggest.
At the moment, as an embedded entity with no data we have:
```JSON
{
  "usageRights": {
    "uri": "https://..."
  }
}
```

When we are a Response Entity - how would we reflect this as the spec says `data` is required?

theefer:
The [argo](https://github.com/argo-rest/spec) spec [requires](https://github.com/argo-rest/spec#data) there to be a `data` property on response entities, it's only optional on embedded entities. That is because if there is no data, why is it a 200? It should be a 404.

The question here is how do we represent a missing override. There are two possible representations:

``` json
{
  "data": {}
}
```

or a 404 (resource does not exist).

But actually the first representation (empty object override) is what we'd use to override an existing usageRights with nothing, i.e. blank out the default usageRights (e.g. because it was wrong). So it makes sense for the "missing override" to be represented by the second solution, which is a 404.

Contrarily to mathematical evidence, 404 > 500.

akash1810:
It feels a bit odd to include a uri that resolves to 404.

jamesgorrie:
It's an interesting one as we don't 404 when you hit the `/metadata/{id}` endpoint, we just return what we consider an empty object.

As a lot of images would have a missing entry in the DB, this seems troublesome.

Why do we require a `data`in the spec when no data can be reflective of a state of an entiry?

theefer:
Do you want to talk about this live?

akash1810:
closing as @kenoir's got another branch.

Issue #993, opened 2015-07-28
Simplify metadata panel


kenoir:
**Problem:**
Reading the items in the metadata panel can be awkward.

![mpanel_jpeg](https://cloud.githubusercontent.com/assets/953792/8930305/abb8a74a-3524-11e5-951d-4bf1b8d5fff5.jpg)

**Solution:**
Use a simpler definition list driven view.

![unnamed](https://cloud.githubusercontent.com/assets/953792/8930248/5907ab40-3524-11e5-9adb-415d536942de.png)

/cc @alastairjardine

alastairjardine:
One of the ideas behind this [non-final but designed as a conversation starter] 'sectioned' panel, to enable the sections to be reused and also grouping things more naturally. A future enhancemnet could be collapsable sections, as we can easily add UI to enable that.

theefer:
Makes sense to me.

alastairjardine:
Just spoken to @kenoir, and I've posited the idea that we could have a think about making this panel responsive to help work with the small screen challenge.

alastairjardine:
Few things we should think about and resolve
- [ ] Usage rights panel overlaying rather than pushing down
- [ ] Show full metadata – decide if we'll just inline with other metadata, or have a collapsable panel
  - [ ] If collapsable, we'll adopt the pattern Composer uses for panels
  - [ ] If inline, we'll put it to the bottom
- [ ] Resolve underline weighting
- [ ] move to a data to a dl list
- [ ] Resolve new pattern/design for pay to use and incomplete warnings
- [ ] update hover style so it's obvious what connects to what 
![Uploading Screen Shot 2015-07-31 at 14.45.52.png…]()
- [ ] maybe more…

akash1810:
Thought - looking at the Solution image above, and the placement of the `Add label` button, we could move away from the "hover to reveal edit pencil" behaviour and just have the edit link next to the field header?

<img width="300" alt="screen shot 2015-08-04 at 20 42 13" src="https://cloud.githubusercontent.com/assets/836140/9070783/7d74a878-3ae9-11e5-823f-0208b5d7908f.png">
 
(ignore styling)

kenoir:
Closing as done, (other udpates should have more specific issues).

PR #1026, opened 2015-08-04
Metadata panel styling

tsop14:
#993
screenshot-media local dev-gutools co uk 2015-08-04 11-24-57
screenshot-media local dev-gutools co uk 2015-08-04 11-25-17

### DIFF
--- a/kahuna/public/js/components/gr-panel/gr-panel.html
+++ b/kahuna/public/js/components/gr-panel/gr-panel.html
@@ -177,10 +177,10 @@
 
 
         <div class="image-info__group">
-            <table>
+            <table class="image-info__group--list">
                 <tbody>
                     <tr class="image-info__wrap metadata-line image-info__byline">
-                        <th class="metadata__key">by</th>
+                        <th class="metadata-line__key">by</th>
                         <td class="image-info__byline metadata-line__info">
                             <button class="image-info__edit"
                                     ng:if="ctrl.userCanEdit"
@@ -212,7 +212,7 @@
                         </td>
                     </tr>
                     <tr class="metadata-line image-info__wrap metadata-line image-info__credit">
-                        <th class="metadata__key">credit</th>
+                        <th class="metadata-line__key">credit</th>
                         <td "image-info__credit metadata-line__info">
                             <button class="image-info__edit"
                                 ng:if="ctrl.userCanEdit"
--- a/kahuna/public/js/components/gr-panel/gr-panel.html
+++ b/kahuna/public/js/components/gr-panel/gr-panel.html
@@ -131,8 +131,8 @@
             </div>
         </div>
 
-        <div class="image-info__group">
-            <div class="image-info__wrap" ng:if="ctrl.rawMetadata.specialInstructions">
+        <div class="image-info__group" ng:if="ctrl.rawMetadata.specialInstructions">
+            <div class="image-info__wrap">
                 <button class="image-info__edit"
                         ng:if="ctrl.userCanEdit"
                         ng:click="specialInstructionsEditForm.$show()"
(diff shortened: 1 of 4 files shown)


kahuna/public/js/image/view.html
```diff
@@ -120,104 +120,102 @@ <h2 class="image-info__title">
...
-                <span class="metadata-line__info">{{ctrl.dimensions.width}} &times; {{ctrl.dimensions.height}}</span>
-            </div>
-        </div>
-
-        <div class="image-info__group" ng:if="ctrl.crop">
-            <div class="metadata-line">
-                dimensions (crop)
-                <span class="metadata-line__info">{{ctrl.dimensions.width}} &times; {{ctrl.dimensions.height}}</span>
-            </div>
+            <table>
```
kenoir: Should these be a `<dl>`?

theefer: We talked about that and in brief, yes, but it's also less convenient than tables because table columns automatically adapt to their widest cell, which you can't do with `dl` (you have to hardcode the width). We suggested to go with the simpler table for now to fix the visual style and move to dl as a future step.

kahuna/public/js/components/gr-panel/gr-panel.html
```diff
@@ -31,7 +31,7 @@ <h1 class="panel__title">{{ctrl.selectedImages.size}} image<span ng:if="ctrl.sel
         </ul>
 
         <div class="image-info__group">
-            <div class="image-info__wrap metadata-line drop-menu image-info__usage-rights"
+            <div class="image-info__wrap metadata-line drop-menu image-info__usage-rights metadata__key"
```
theefer: `metadata__key` here is a bit odd because according to [BEM](https://css-tricks.com/bem-101/) it means "the `key` child element of the `metadata` block", but there is no parent `metadata` block here.

theefer:
Minor comment about `metadata__key` class, but otherwise looks great! :+1:

kahuna/public/js/image/view.html
jamesgorrie: Flex?

kenoir: Flex all the things.

kahuna/public/stylesheets/main.css
```diff
@@ -1046,21 +1045,31 @@ FIXME: what to do with touch devices
...
+.image-info__group {
+    padding: 10px;
+    border-bottom: 1px solid #565656;
+} 
+
+.image-info:last-child .image-info__group{
+    border-bottom: none;
+}   
+
+.image-info__group table {
```
jamesgorrie: Any chance we could give these tables classnames, so if we do move to `dl` or whatever we can just use those?

jamesgorrie:
Same as @theefer - minor comment - :+1:

kahuna/public/js/image/view.html
```diff
@@ -240,13 +238,15 @@ <h2 class="image-info__title">
...
-    <dl class="image-info">
-        <dt class="image-info__heading image-info__heading--first">Labels</dt>
-        <dd>
-            <ui-labeller image="ctrl.image">
-            </ui-labeller>
-        </dd>
-    </dl>
+    <div class="image-info">
+        <dl class="image-info__group">
+            <dt class="image-info__heading image-info__heading--first">Labels</dt>
```
alastairjardine: CSS transform to uppercase in another PR?

alastairjardine:
It would be nice to make a start on sorting out the casing. If not this PR, the next one.

alastairjardine:
Might be nice to tweak the colour of the underline – it's still really strong and visually discordant.

alastairjardine:
We'll do a wash up and collect minor tweaks. 

Good stuff :+1:

kahuna/public/stylesheets/main.css
theefer: I'd generally agree although in the case of `border-collapse`, it only makes sense on a table. The `width: 100%` would be correct regardless of the tag on the other hand.

tsop14:
Releasing this branch but comments regarding dl instead of a table are being addressed in another branch now in progress.

PR #1027, opened 2015-08-04
"auto" archive staff photographer's images.

akash1810:
"auto" because the workflow is:

- upload image
- image loader will post a message to a (new) SNS topic
- metadata api has a thrall like message consumer, which will detect if an image is staff, and write to dynamo if so then post a metadata-edit message to thrall's SNS topic
- thrall processes the queue as usual, writing to ES.

There is a delay between the last two steps where the UI may not have a filled in star to denote the image has been archived; possible solutions include:

- move `archived` to the root of the document
- server side notifications to broadcast that the SNS message has been processed post upload and so toggle the star state.


kenoir:
@akash1810 what happens if the message from the metadata-editor is faster to reach ES than the first message from a Thrall?

theefer:
Ah, I didn't realise the image-loader was sending the message. In my mind the thrall would let people know, using a new topic, that it has indexed an image (or done something else). That'd also solve the racing issue.

kenoir:
A sequence diagram would be really handy in understanding what's going on here.

Is this what's going on?

blank flowchart - new page

jamesgorrie:
>  thrall would let people know

This would then be adding logic to thrall i.e. if staff - archive.
It feels as though thrall should remain dumb?

akash1810:
@kenoir re diagram, yes

kenoir:
@jamesgorrie yeah, but also adding the logic the metadata-api seems incongruent too (as it gives it another slightly tangential purpose). Another service perhaps?

akash1810:
> This would then be adding logic to thrall i.e. if staff - archive.

Not necessarily - atm, the Upload Queue processor has this logic. Thrall would just send a `new image` message.

jamesgorrie:
@akash - I was talking about if we let thrall let people know as @theefer thought it worked

theefer:
> This would then be adding logic to thrall i.e. if staff - archive.

No, thrall should just say "hey, here is a new image I just ingested". Other services would still apply the logic.

kenoir:
Maybe like this:

blank flowchart - new page 2

theefer:
What @kenoir said.

akash1810:
Updated to work per @kenoir's diagram.

akash1810:
@theefer @jamesgorrie @kenoir @tsop14 thumbs?

metadata-editor/app/lib/DynamoEdits.scala
```diff
@@ -0,0 +1,27 @@
+package lib
+
+import com.gu.mediaservice.lib.aws.DynamoDB
+import com.gu.mediaservice.model.Edits
+import play.api.libs.json.{Json, JsObject}
+import play.api.libs.concurrent.Execution.Implicits._
+
+object DynamoEdits {
```
theefer: Can we have a comment about what this is and why it's here (vs the controller), as it looks transitional (for a reason)?

metadata-editor/app/lib/DynamoEdits.scala
```diff
@@ -0,0 +1,27 @@
...
+      .map(publish(id))
+  }
+
+  def publish(id: String)(metadata: JsObject): Edits = {
+    val message = Json.obj(
+      "id" -> id,
+      "data" -> metadata
+    )
+
+    Notifications.publish(message, "update-image-user-metadata")
```
theefer: Didn't we just change this to be type-safer, or has this not been merged? cc @jamesgorrie

metadata-editor/app/lib/MetadataMessageConsumer.scala
```diff
@@ -0,0 +1,37 @@
...
+  Config.queueUrl, Config.awsEndpoint, Config.awsCredentials, MetadataEditorMetrics.processingLatency) {
+
+  override def chooseProcessor(subject: String): Option[JsValue => Future[Any]] =
+    PartialFunction.condOpt(subject) {
+      case "image-indexed" => processIndexedImage
+    }
+
+  def isStaffPhotographer(image: Image) = {
+    image.metadata.byline match {
+      case Some(byline: String) => StaffPhotographers.store.contains(byline)
```
theefer: Do we only archive staff photographer's images (not contributed/commissioned)?

metadata-editor/app/lib/DynamoEdits.scala
jamesgorrie: Just merged that change as there was some things to think about...

theefer:
Nice!

A few comments/questions. Would be good to trial on TEST first?

Also note that the branch needs master merged in.

metadata-editor/app/lib/MetadataMessageConsumer.scala
jamesgorrie: We should do all of them.

jamesgorrie:
Something to think about - how do we do this retrospectively.

theefer:
> how do we do this retrospectively

The reindex message should be sent if we send a media-api reindex (@akash1810 is it?). So we could just query by name or category, trigger a reindex and that should reset the overrides?

Still not 100% this is the right solution, even though I suggested it.

@kenoir @jamesgorrie @akash1810 thoughts welcome.

akash1810:
The message is send when thrall processes an `image` message iirc that message is only sent on new upload.

> how do we do this retrospectively.

I'm not sure this is a concern, as this PR is automating a process that users have been spending time doing manually already.

theefer:
We should probably also send the message on [`update-image`](https://github.com/guardian/grid/blob/master/media-api/app/controllers/MediaApi.scala#L183)? Thoughts?

akash1810:
> We should probably also send the message on update-image? Thoughts?

end of week tiredness kicking in....Thrall processes `image` and `update-image` in the [same way](https://github.com/guardian/grid/blob/master/thrall/app/lib/MessageConsumer.scala#L76)... so we're already sending the message on `update-image` :smile:

metadata-editor/app/lib/MetadataMessageConsumer.scala
akash1810: any hints as to where these are listed?

metadata-editor/app/lib/MetadataMessageConsumer.scala
```diff
@@ -0,0 +1,34 @@
...
+object MetadataMessageConsumer extends MessageConsumer(
+  Config.queueUrl, Config.awsEndpoint, Config.awsCredentials, MetadataEditorMetrics.processingLatency) {
+
+  override def chooseProcessor(subject: String): Option[JsValue => Future[Any]] =
+    PartialFunction.condOpt(subject) {
+      case "image-indexed" => processIndexedImage
+    }
+
+  def isArchivingCategory(category: String) = {
+    val archiveCategory = List("staff-photographer", "contract-photographer", "commissioned-photographer")
```
akash1810: magic strings are magic...

jamesgorrie: You could match on the types.
```Scala
isArchivingUsageRights[T <: UsageRights](usageRights: T) = usageRights match {
  case _:StaffPhotographer | _:ContractPhotographer | _:CommissionedPhotographer => true
  case _ => false
}
```

(there might be a better syntax, but it's nicer than using the strings).

akash1810: Much nicer!

The `case _:T1 | _:T2 => foo()` syntax looks to be pretty common and readable too.

akash1810: Much nicer!

The `case _:T1 | _:T2 => foo()` syntax looks to be pretty common and readable too.

akash1810:
One thing to note: manual updates to the byline will not trigger an auto archive atm.

jamesgorrie:
@akash - after talking to @theefer I'm not sure that we should do things like that as it's a little hard for a user to reason with. Similar to the usage rights - we might prompt a user, after setting a staff-photographer, would they like to copy the data to the usage rights from the client, but not automatically.

Similar to this, we could suggest archiving it.

akash1810:
> we might prompt a user, after setting a staff-photographer, would they like to copy the data to the usage rights from the client, but not automatically.

+1

theefer:
Sounds like it might be worth having another catch up about this to work
out the different scenarios?

Note that a lot of this work is likely to be useful regardless, for
instance to clear up entries in dynamo for images after they are deleted.

On Tue, 11 Aug 2015 10:19 James Gorrie <notifications@github.com> wrote:

> @akash <https://github.com/akash> - after talking to @theefer
> <https://github.com/theefer> I'm not sure that we should do things like
> that as it's a little hard for a user to reason with. Similar to the usage
> rights - we might prompt a user, after setting a staff-photographer, would
> they like to copy the data to the usage rights from the client, but not
> automatically.
>
> Similar to this, we could suggest archiving it.
>
> —
> <https://github.com/guardian/grid/pull/1027#issuecomment-129786273>.
>

-- 

in England Number 908396

akash1810:
(this comment is more of a reminder for me for tomorrow...)

Where do we stand on this?

theefer:
Let's catch up tomorrow

jamesgorrie:
@theefer is it tomorrow yet?

theefer:
Let's try and find some time later today.

akash1810:
Caught up w/master now.

Note: GH seems to show a `git mv` MessageConsumer.scala -> ThrallMessageConsumer.scala as a delete and new file :cry:

akash1810:
This work is superseded by https://github.com/guardian/grid/pull/1202.

PR #1029, opened 2015-08-05
use new photgrapher list

jamesgorrie:
Left out a reference - need to check TC to make sure it failed and was just my human error.


akash1810:
:+1:

PR #1033, opened 2015-08-05
Image metadata dl

tsop14:
#993

Changes the `<table>` values to `<dl>` for the metadata on selection view and image view, and the drop down full-metadata list on image view.

### DIFF
--- a/kahuna/public/js/components/gr-panel/gr-panel.html
+++ b/kahuna/public/js/components/gr-panel/gr-panel.html
@@ -175,78 +175,72 @@
             </div>
         </div>
 
-
         <div class="image-info__group">
-            <table class="image-info__group--list">
-                <tbody>
-                    <tr class="image-info__wrap metadata-line image-info__byline">
-                        <th class="metadata-line__key">by</th>
-                        <td class="image-info__byline metadata-line__info">
-                            <button class="image-info__edit"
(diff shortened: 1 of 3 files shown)


kahuna/public/stylesheets/main.css
```diff
@@ -1099,6 +1099,57 @@ FIXME: what to do with touch devices
     position: relative;
 }
 
+.image-info__group--dl {
+  width: 100%;
+  display: -webkit-box;
+  display: -webkit-flex;
```
theefer: Is that really necessary for the browsers we support (IIRC Chrome 39+, FF 33+)?

kahuna/public/js/components/gr-panel/gr-panel.html
```diff
@@ -175,78 +175,72 @@ <h2 class="image-info__title" ng:if="!ctrl.hasMultipleValues(ctrl.rawMetadata.ti
             </div>
         </div>
```
theefer: Do we also want to use DL for title/description/special instructions?

theefer:
:+1:

jamesgorrie:
Side note - [some nice gitness](https://github.com/blog/1506-closing-issues-via-pull-requests)

PR #1035, opened 2015-08-05
Show 'None' as category if none

theefer:
Previously was just missing any value.

image

Note: my editor also removed trailing whitespace.


akash1810:
:+1:

PR #1036, opened 2015-08-05
Match up es.cluster name between services and ES config itself

theefer:
My hope is for this to reduce noise in logs like:

```
2015-08-05 15:26:07,283 - [WARN] - from org.elasticsearch.client.transport in elasticsearch[Justine Hammer][generic][T#1] 
[Justine Hammer] node null not part of the cluster Cluster [media-api], ignoring...
```

Will want to triple check effect on TEST first.

### DIFF
--- a/media-api/conf/application.conf
+++ b/media-api/conf/application.conf
@@ -3,7 +3,7 @@ application.langs="en"
 session.httpOnly=false
 session.secure=true
 
-es.cluster = "media-api"
+es.cluster = "media-service"
 es.host = "localhost"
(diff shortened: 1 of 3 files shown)


akash1810:
:+1: 

[Related](https://github.com/guardian/grid/pull/899) - I'll try and find the doc page where I read about the ES REST API not using the cluster name.

theefer:
Going to trial on thrall on TEST now.

theefer:
Looks sane on test (works and no errors)

PR #1037, opened 2015-08-05
Makes sure publish message is valid on edits

jamesgorrie:
A simple version to make sure we aren't sending gumpth through to Elastic.
Simple version of #1038


theefer:
:+1: please test that it works. you could even test with that dangerous image we found :)

jamesgorrie:
@theefer Do you think this is better that #1038?

theefer:
It's Friday,

jamesgorrie:
Last thing I do - I swear.

PR #1038, opened 2015-08-05
Use types to publish to SNS

jamesgorrie:
Went down a rabbit whole - not sure if this makes sense, but seems nice to publish types, not `JsValue`s.

Problem is the error handling as, chances are, you've saved something, now the message isn't valid, so you're in a half way house of saved and not saved.

More complex version of #1037.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/SNS.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/SNS.scala
@@ -4,7 +4,8 @@ import com.amazonaws.auth.AWSCredentials
 import com.amazonaws.services.sns.{AmazonSNSClient, AmazonSNS}
 import com.amazonaws.services.sns.model.PublishRequest
 import play.api.Logger
-import play.api.libs.json.{Json, JsValue}
+import play.api.libs.json.{Writes, Json, JsValue}
+import scala.util.Try
 import scalaz.syntax.id._
 
 
@@ -19,5 +20,16 @@ class SNS(credentials: AWSCredentials, topicArn: String) {
     val result = client.publish(new PublishRequest(topicArn, Json.stringify(message), subject))
     Logger.info(s"Published message: $result")
   }
+
+  def publish[T](id: String, obj: T, subject: String)(implicit tjs: Writes[T]): Option[T] = {
+    val jsonOpt = Try(Json.toJson(obj)).toOption
+
+    jsonOpt.map(json => {
+      publish(message(id, json), subject)
+      obj
+    })
+  }
+
+  def message(id: String, data: JsValue) = Json.obj("id" -> id, "data" -> data)
 
 }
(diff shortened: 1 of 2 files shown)


kenoir:
Nice. :+1: 

Raises the question, should we save changes if we can't write to ES? What would we do if `metadata-editor` stopped publishing notifications for a while and ES got out of sync?

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/SNS.scala
```diff
@@ -20,4 +21,15 @@ class SNS(credentials: AWSCredentials, topicArn: String) {
     Logger.info(s"Published message: $result")
   }
 
+  def publish[T](id: String, obj: T, subject: String)(implicit tjs: Writes[T]): Option[T] = {
+    val jsonOpt = Try(Json.toJson(obj)).toOption
+
+    jsonOpt.map(json => {
+      publish(message(id, json), subject)
+      obj
+    })
```
theefer: Can/should we log if the `Option` is `None`? It should never be, right?

metadata-editor/app/controllers/EditsController.scala
```diff
@@ -48,6 +48,7 @@ object EditsController extends Controller with ArgoHelpers {
 
   val Authenticated = EditsApi.Authenticated
   val dynamo = new DynamoDB(Config.awsCredentials, Config.dynamoRegion, Config.editsTable)
+  def couldNotPublishMessage(id: String) = respondError(InternalServerError, "message-not-published", s"Data saved, but indexing could not take place")
```
theefer: For precision, "notification/message could not be published/sent"? (this component just sends a message, what is done with it is not its concern)

theefer:
Cool :+1: 

> Raises the question, should we save changes if we can't write to ES? What would we do if metadata-editor stopped publishing notifications for a while and ES got out of sync?

Last time this happened (SNS/SQS outage), we talked about two solutions, ideally both to be implemented:

1. An always-active backup SNS/SQS setup in another region (e.g. Frankfurt) that would automatically get used if the main channel isn't accessible. Publishers would write to it if the first write to eu-west-1 failed, and consumers (thralls) would always read from both.
2. A way to reindex all data from a given datetime range for each service, so missed indexes can be re-applied.

The second point is possibly quite expensive with cropper given that I don't think there'd be a way to find all the crops produced within a period without listing the date creation times of *all* objects in the S3 bucket...

Metadata editor could be easier if we can do a query on a date field in DynamoDB.

jamesgorrie:
I have a better idea what about:

user -> server -> validate json
* invalid -> badrequest
* valid -> keep going
-> save to dynamo -> notify -> respond

Basically suggesting that we take the validation out of the save / notify steps, but if something does go wrong, throw a 500 error as that's what's happened. We can catch that respond accordingly. Logging hre is also then free.

theefer:
Aren't we doing that with the forms in the controller to some extent? Or is that not as reliable as reading the JSON into the relevant case class? I know we meant to change the former to the latter in the API controller anyway, is that related to your suggestion?

jamesgorrie:
There feels there is not enough thought abou here - we are type safe from the simple version (#1037) for now.

PR #1039, opened 2015-08-05
Add tracking on recent uploads button clicks

theefer:
I'm still feeling like being overly cautious about releasing #1024 (retire 50 recent uploads), so this adds tracking on the recent uploads button so we can see how much it's actually being used.


jamesgorrie:
:+1: 
I have been asked twice so far if we could add that button to the image detail page - which is interesting.

PR #1040, opened 2015-08-06
New getty exclusions

jamesgorrie:
New list from Jo B.

## Free count:
`before: 1645942`
`after: 1651972`


common-lib/src/main/scala/com/gu/mediaservice/lib/config/UsageRightsConfig.scala
```diff
@@ -5,90 +5,35 @@ object UsageRightsConfig {
...
-    "Ulrich Baumgarten",
-    "Universal Images",
-    "World Kabbadi League",
-    "ullstein bild"
+    "Sports Illustrated Classic",
+    "Terry O'Neill",
+    "The Asahi Shimbun Premium Archive",
+    "The LIFE images collection",
+    "The LIFE picture collection Editorial",
+    "The LIFE premium collection"
```
theefer: capitalisation has changed. I think capitalisation may matter -- is it worth checking for matches?

theefer:
Comment about capitalisation, else :+1:

jamesgorrie:
> capitalisation

You were right, nice catch, updated and checked the rest from the list.

common-lib/src/main/scala/com/gu/mediaservice/lib/config/UsageRightsConfig.scala
```diff
@@ -5,90 +5,35 @@ object UsageRightsConfig {
...
-    "Her Og Nu",
-    "Hero Images",
-    "Hulton Royals Collection",
-    "IAAF World Athletics",
-    "ICC",
-    "Imperial War Museums",
-    "Interact Images",
+    "Fashion Window",
+    "Gamma-Legend",
+    "Gamma-Rapho",
```
theefer: This has also had a hyphen materialise. Can we check if one is the canonical one and which one?

common-lib/src/main/scala/com/gu/mediaservice/lib/config/UsageRightsConfig.scala
```diff
@@ -5,90 +5,35 @@ object UsageRightsConfig {
...
-    "Caiaimage",
-    "Carnegie Museum Art",
+    // TODO: Remove once we are on the new UsageRights model
+    // This is not excluded as we have a deal with Barcroft,
+    // and there are only a tiny amount from Getty
+    // "Barcroft Media",
     "Catwalking",
-    "Champions Hockey League",
-    "City-Press",
+    "Contour by Getty Images",
```
theefer: Was "Contour" before.

common-lib/src/main/scala/com/gu/mediaservice/lib/config/UsageRightsConfig.scala
```diff
@@ -5,90 +5,35 @@ object UsageRightsConfig {
...
+    "Fashion Window",
+    "Gamma-Legend",
+    "Gamma-Rapho",
     "International Center of Photography",
-    "Kommersant",
     "Lichfield Archive",
-    "LightRocket",
-    "MLB Platinum",
-    "Mark Leech Sports Photography",
+    "MLB Major League Baseball Platinum",
```
theefer: Was "MLP Platinum" before.

common-lib/src/main/scala/com/gu/mediaservice/lib/config/UsageRightsConfig.scala
```diff
@@ -5,90 +5,35 @@ object UsageRightsConfig {
...
-    "Toronto Star",
-    "UEFA.com",
-    "UK Press",
-    "Ulrich Baumgarten",
-    "Universal Images",
-    "World Kabbadi League",
-    "ullstein bild"
+    "Sports Illustrated Classic",
+    "Terry O'Neill",
+    "The Asahi Shimbun Premium Archive",
```
theefer: These two also had other syntax before, can we check?

theefer:
Sorry but there are more discrepancies in the previous vs new collection names that I think we should check.

common-lib/src/main/scala/com/gu/mediaservice/lib/config/UsageRightsConfig.scala
jamesgorrie: `"Contour by Getty Images"` is correct.

common-lib/src/main/scala/com/gu/mediaservice/lib/config/UsageRightsConfig.scala
jamesgorrie: The hyphen is correct for `"Gamma-Rapho"` - Legend doesn't have any, but assuming it's the same.

common-lib/src/main/scala/com/gu/mediaservice/lib/config/UsageRightsConfig.scala
jamesgorrie: Neither match - but assuming as the new ones seem to be correct, this must be correct.

common-lib/src/main/scala/com/gu/mediaservice/lib/config/UsageRightsConfig.scala
jamesgorrie: Which two?

theefer:
Great thanks for checking. If we really wanted to check we could probably download images from the Getty archive in the respective collections, but given there are no images now hopefully they're not collections under heavy use.

common-lib/src/main/scala/com/gu/mediaservice/lib/config/UsageRightsConfig.scala
theefer: Terry O'Neill (curly apostrophe vs straight single quote), and "The Asahi Shimbun Premium [Archive]".

jamesgorrie: The single quote must have been a copy paste balls up as it's the apostrophe.
Checking with Jo on the [Archive].

jamesgorrie: Only ones in the archive are excluded.

theefer: OK great! The master version has the straight single quote for Terry O'Neill, are you saying it should be the curly one?

jamesgorrie: It's meant to be the straight on (non match the curly, so we must have been getting it wrong).
It is now correct.

theefer: OK :sunny:

PR #1045, opened 2015-08-06
correcting `panel__content` height.

akash1810:
`panel-bottom` housed the cancel button, which has been promoted to the toolbar, can get rid of that class now.

(big gifs because recorded on retina screen :cry:)

before
before

after
after


kenoir:
@akash1810 again thanks :+1:

alastairjardine:
Great work all. Loving the way this is coming together. Nice UI to build on in the future.

theefer:
Neat :+1:

PR #1047, opened 2015-08-06
Locking metadata panel doesn't in older versions of Chrome

kenoir:
Older versions of chrome don't like the %- inside the css calc


kenoir:
Passes inspection by @jamesgorrie

theefer:
Funny that

PR #1051, opened 2015-08-07
delete image component.

akash1810:
Reusing `gr-confirm-delete` to show a confirm message before performing a delete on an image.

In first instance, adding this to the preview page:

delete-image

cc @alastairjardine

### DIFF
--- a/kahuna/public/js/components/gr-confirm-delete/gr-confirm-delete.js
+++ b/kahuna/public/js/components/gr-confirm-delete/gr-confirm-delete.js
@@ -1,4 +1,5 @@
 import angular from 'angular';
+import './gr-confirm-delete.css!';
 
 export const confirmDelete = angular.module('gr.confirmDelete', []);
 
@@ -6,10 +7,14 @@ confirmDelete.directive('grConfirmDelete', ['$timeout', function($timeout) {
 
     return {
         restrict: 'E',
+        transclude: true,
         template: `
-            <button class="button-ico" type="button" ng:click="showConfirm = true">
-                <span class="confirm" ng:if="showConfirm">Sure?</span>
-                <gr-icon>undo</gr-icon>
+            <button class="gr-confirm-delete" type="button"
+                ng:click="showConfirm = true"
+                ng:class="{'gr-confirm-delete--confirm': showConfirm}">
+                <gr-icon>delete</gr-icon>
+                <ng-transclude ng:if="!showConfirm"></ng-transclude>
+                <span class="gr-confirm-delete__label" ng:if="showConfirm">Confirm delete</span>
             </button>`,
 
         link: function(scope, element, attrs) {
--- a/kahuna/public/js/components/gr-delete-image/gr-delete-image.js
+++ b/kahuna/public/js/components/gr-delete-image/gr-delete-image.js
@@ -1,0 +1,41 @@
+import angular from 'angular';
+import '../gr-confirm-delete/gr-confirm-delete';
+
+export const deleteImage = angular.module('gr.deleteImage', ['gr.confirmDelete']);
+
+deleteImage.controller('grDeleteImageCtrl', ['$rootScope', '$scope', 'mediaApi',
+    function ($rootScope, $scope, mediaApi) {
+        var ctrl = this;
+
+        ctrl.delete = function () {
+            mediaApi.delete(ctrl.image)
+                .then((resp) => {
+                    if (angular.isDefined(ctrl.onSuccess)) {
+                        ctrl.onSuccess(resp);
+                    }
+                })
+                .catch((err) => {
+                    if (angular.isDefined(ctrl.onError)) {
+                        ctrl.onError(err);
+                    }
+                });
+        };
+}]);
+
+deleteImage.directive('grDeleteImage', [function () {
+    return {
+        restrict: 'E',
+        template: `
+            <gr-confirm-delete class="gr-delete-image" gr-on-confirm="ctrl.delete()">
+                <span class="gr-delete-image__label">Delete</span>
+            </gr-confirm-delete>`,
+        controller: 'grDeleteImageCtrl',
+        controllerAs: 'ctrl',
+        bindToController: true,
+        scope: {
+            image: '=',
+            onSuccess: '=?',
+            onError: '=?'
+        }
+    };
+}]);
--- a/kahuna/public/js/components/gr-delete-image/gr-delete-image.js
+++ b/kahuna/public/js/components/gr-delete-image/gr-delete-image.js
@@ -3,23 +3,22 @@ import '../gr-confirm-delete/gr-confirm-delete';
 
 export const deleteImage = angular.module('gr.deleteImage', ['gr.confirmDelete']);
 
-deleteImage.controller('grDeleteImageCtrl', ['$rootScope', '$scope', 'mediaApi',
-    function ($rootScope, $scope, mediaApi) {
-        var ctrl = this;
+deleteImage.controller('grDeleteImageCtrl', ['mediaApi', function (mediaApi) {
+    var ctrl = this;
 
-        ctrl.delete = function () {
-            mediaApi.delete(ctrl.image)
-                .then((resp) => {
-                    if (angular.isDefined(ctrl.onSuccess)) {
-                        ctrl.onSuccess(resp);
-                    }
-                })
-                .catch((err) => {
-                    if (angular.isDefined(ctrl.onError)) {
-                        ctrl.onError(err);
-                    }
-                });
-        };
+    ctrl.delete = function () {
+        mediaApi.delete(ctrl.image)
+            .then((resp) => {
+                if (angular.isDefined(ctrl.onSuccess)) {
+                    ctrl.onSuccess(resp);
+                }
+            })
+            .catch((err) => {
+                if (angular.isDefined(ctrl.onError)) {
+                    ctrl.onError(err);
+                }
+            });
+    };
 }]);
 
 deleteImage.directive('grDeleteImage', [function () {
--- a/kahuna/public/js/image/controller.js
+++ b/kahuna/public/js/image/controller.js
@@ -3,13 +3,15 @@ import angular from 'angular';
 import '../image/service';
 import '../edits/service';
 import '../analytics/track';
+import '../components/gr-delete-image/gr-delete-image';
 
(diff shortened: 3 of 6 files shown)


kahuna/public/js/edits/image-editor.js
```diff
@@ -73,6 +75,18 @@ imageEditor.controller('ImageEditorCtrl',
         ctrl.error = true;
     }
 
+    mediaApi.canDelete(ctrl.image).then(deletable => {
+        ctrl.canBeDeleted = deletable;
+    });
+
+
+    ctrl.onDeleteSuccess = () => {
+        $state.go('upload', {}, {reload: true});
```
theefer: Does this re-load the whole page?

kahuna/public/js/components/gr-delete-image/gr-delete-image.js
```diff
@@ -0,0 +1,33 @@
...
+
+export const imageDelete = angular.module('gr.imageDelete', ['gr.confirmDelete']);
+
+imageDelete.controller('grImageDeleteCtrl', ['$scope', 'mediaApi', function ($scope, mediaApi) {
+    var ctrl = this;
+
+    ctrl.delete = function () {
+        mediaApi.delete(ctrl.image)
+            .then((resp) => ctrl.onSuccess(resp))
+            .catch((err) => ctrl.onError(err));
```
theefer: Note that if either of these callbacks are not provided this will error out. If they are required, could check for their presence when initialising the directive (fail early)?

kahuna/public/js/edits/image-editor.js
akash1810: Yeah :cry:

Ideally just that `li` would be removed from the page. Will take a look.

alastairjardine:
Cool – could we update the copy so its:

Delete —> Confirm delete.

Thanks

kahuna/public/js/components/gr-confirm-delete/gr-confirm-delete.js
```diff
@@ -6,10 +7,14 @@ confirmDelete.directive('grConfirmDelete', ['$timeout', function($timeout) {
...
         template: `
-            <button class="button-ico" type="button" ng:click="showConfirm = true">
-                <span class="confirm" ng:if="showConfirm">Sure?</span>
-                <gr-icon>undo</gr-icon>
+            <button class="gr-confirm-delete button-ico" type="button"
+                ng:click="showConfirm = true"
+                ng:class="{'gr-confirm-delete--confirm': showConfirm}">
+                <gr-icon>delete</gr-icon>
+                <ng-transclude ng:if="!showConfirm"></ng-transclude>
+                <span class="confirm" ng:if="showConfirm">are you sure?</span>
```
alastairjardine: "Confirm delete"

jamesgorrie: > "Confirm delete"

:computer::lips:

kahuna/public/js/components/gr-delete-image/gr-delete-image.js
jamesgorrie: Would this not error from instantiating the directive without them as you've required them in the [scope](https://github.com/guardian/grid/pull/1051/files#diff-98f40c89c6f7ce3607c23deb9b0b1856R29)?

jamesgorrie: You would use '=?' if you wanted them as optional.

kahuna/public/js/image/controller.js
```diff
@@ -176,16 +178,15 @@ image.controller('ImageCtrl', [
...
-                });
+        ctrl.onDeleteSuccess = function () {
+            $state.go('search');
+        };
+
+        ctrl.onDeleteError = function (err) {
+            if (err.body.errorKey === 'image-not-found') {
+                $state.go('search');
+            } else {
+                window.alert(err.body.errorMessage);
```
jamesgorrie: `$window`?

kahuna/public/js/edits/image-editor.js
```diff
@@ -73,6 +75,18 @@ imageEditor.controller('ImageEditorCtrl',
...
+    };
+
+    ctrl.onDeleteError = (err) => {
+        window.alert(err.body.errorMessage);
```
jamesgorrie: `$window`?

jamesgorrie:
There is the question on the uploads page as to whether this acts more like a "cancel" than a delete, not sure we should just bodge it in there?

kahuna/public/js/components/gr-delete-image/gr-delete-image.js
theefer: Oh really? I thought they'd end up undefined. If that errors automatically then great.

theefer:
"cancel" would perpetuate the false myth that images aren't uploaded yet. They are. I think we should make it clearer.

jamesgorrie:
> They are. I think we should make it clearer.

:+1: +

kahuna/public/js/components/gr-delete-image/gr-delete-image.js
```diff
@@ -0,0 +1,40 @@
...
+            })
+            .catch((err) => {
+                if (angular.isDefined(ctrl.onError)) {
+                    ctrl.onError(err);
+                }
+            });
+    };
+}]);
+
+imageDelete.directive('grImageDelete', [function () {
```
jamesgorrie: Is there a reason why the directive / ctrl and file name are different - cold be confusing when searching.

kahuna/public/js/edits/image-editor.html
```diff
@@ -75,6 +75,14 @@
             ></ui-archiver>
         </div>
 
+        <div class="result-editor__delete-image">
+            <gr-image-delete image="ctrl.image"
```
jamesgorrie: We should conventionalise on naming i.e. `-` vs `:`

jamesgorrie: ^ Again - making it hard to search the codebase

theefer: I think we always use `-` in element names for browser compat/styling/etc reasons?

kahuna/public/js/components/gr-delete-image/gr-delete-image.js
akash1810: I didn't even notice that :fearful:

kahuna/public/js/edits/image-editor.html
jamesgorrie: ope, not what I've been doing, I'll go clean it up.

kahuna/public/js/components/gr-confirm-delete/gr-confirm-delete.js
akash1810: I might as well update all the items in the toolbar in this PR too - it looks weird otherwise.

jamesgorrie: clean up as we go along is a good way to do it.

theefer: Just a note: let's keep separate functional changes in separate PRs, that makes it easier to review, revert, avoids blocking individual functionalities when discussing unrelated changes, etc. Thanks.

kahuna/public/js/edits/image-editor.js
akash1810: Yep.

Will make it talk to the error service in a later PR too.

kahuna/public/js/edits/image-editor.js
theefer: Suggested emitting a notification, similar to image-updated, which the image upload service could listen to to update its state (remove deleted images).

jamesgorrie: After @kenoir  and I worked on the image result list with this - it was confusing.
Hence the push to move to the [stream thing](https://github.com/guardian/grid/pull/1022).

As we can explicitly tie the list and edits, I think we should do it that way (have the image list listen to edits somehow, not on a global handler).

akash1810: In this instance, I don't think a global event emission is required. Instead, the `gr-delete-image` shouldn't actually belong in `ui-image-editor` instead, it should be [here](https://github.com/guardian/grid/blob/master/kahuna/public/js/upload/view.html#L28). That way, the upload controller can remove the image from `this.myUploads` in the `onSuccess` argument of `gr-delete-image`.

jamesgorrie: @akash1810 :+1:

kenoir:
@akash1810 do you reckon we should keep the button a consistent size so when the text changes the width remains the same?

akash1810:
@kenoir definitely!

jamesgorrie:
> a consistent size

Perhaps setting a `min-width`, just incase it outgrows itself.

akash1810:
Updated this PR to only change the Preview page - will add to Uploads page in a subsequent PR.

kahuna/public/js/components/gr-delete-image/gr-delete-image.js
```diff
@@ -0,0 +1,41 @@
+import angular from 'angular';
+import '../gr-confirm-delete/gr-confirm-delete';
+
+export const deleteImage = angular.module('gr.deleteImage', ['gr.confirmDelete']);
+
+deleteImage.controller('grDeleteImageCtrl', ['$rootScope', '$scope', 'mediaApi',
+    function ($rootScope, $scope, mediaApi) {
```
jamesgorrie: `$rootScope` <= redundant?

jamesgorrie: `$scope` too?

kahuna/public/js/edits/image-editor.js
```diff
@@ -9,12 +9,12 @@ import '../usage-rights/usage-rights-editor';
 export var imageEditor = angular.module('kahuna.edits.imageEditor', [
     'kahuna.edits.service',
     'gr.image.service',
-    'kahuna.edits.usageRightsEditor'
+    'kahuna.edits.usageRightsEditor',
```
jamesgorrie: Trailing commas at end of array can cause problems...

jamesgorrie:
:+1:

PR #1052, opened 2015-08-07
Label clean up

tsop14:
Displays labels in full besides single image. 
Maintains single line view on the multi-view but removes scroll bar so users can view and click on labels. 

NOW 
screenshot-media local dev-gutools co uk 2015-08-07 10-40-02

BEFORE
screenshot-media local dev-gutools co uk 2015-08-04 11-38-57

NOW
screenshot-media local dev-gutools co uk 2015-08-07 11-04-23

BEFORE
screenshot-media local dev-gutools co uk 2015-08-07 11-04-07


akash1810:
Does this affect the Upload page too?

tsop14:
@akash1810 - no changes do not affect any other page. 

At the moment lots of labels will push the upload page to an ever increasing width: 
screenshot-media test dev-gutools co uk 2015-08-07 12-19-42

Should a scroll bar be added? Or, row wrap with increased height as happens with long descriptions.

jamesgorrie:
I wouldn't worry about the upload page in edge cases.

jamesgorrie:
How do you see the rest of the labels on a result?
Do we think that clicking on it to show the metadata panel is good enough?

If so, we could potentially give some visual cue that there is more images there incase people just assume a label hasn't been added etc... like a fade covering the end of the last label.

akash1810:
On the results page, the add label button is currently at the end of the list, so with a long list of labels, this button will be lost off screen (you have to scroll to get to it). Should we move it to the start instead, to prevent this?

tsop14:
@jamesgorrie you see the rest of the labels by scrolling (same functionality as before but the scrollbar is now hidden).

Will look at creating a gradient to create a clue that there are more labels not displayed...

jamesgorrie:
Ah, brilliant - hard to tell from the screen grab as not animated, nice work.

kahuna/public/stylesheets/main.css
```diff
@@ -1430,12 +1431,22 @@ FIXME: what to do with touch devices
     overflow: hidden;
 }
 
+.image-info__group--fixed-panel .labeller{
```
jamesgorrie: Could we not just add this as a state to the labeller e.g. `.labeller--single-row`?

jamesgorrie: Or are we avoiding that as we wold either have to create another view or send over a var as a state variable to the directive?

jamesgorrie:
:+1:

PR #1053, opened 2015-08-07
Remove duplicates from search results

theefer:
Keep a map of image id -> position, and when we detect an image is being loaded a second time at a different position, delete its former position to only keep the latest one.

This may leave holes or miss images under certain conditions, but at least it should fix the blank results bug so many people have experienced.

Tested locally be removing an image in the result set and scrolling back up from the end. Duplicates correctly detected and removed. I haven't found an easy way to inject out-of-order new images but the logic should hold for that case too.

### DIFF
--- a/kahuna/public/js/search/results.js
+++ b/kahuna/public/js/search/results.js
@@ -151,7 +151,7 @@ results.controller('SearchResultsCtrl', [
             ctrl.imagesAll = [];
             ctrl.imagesAll.length = Math.min(images.total, ctrl.maxResults);
 
-            imagesPositions = new Map;
+            imagesPositions = new Map();
 
             checkForNewImages();
 
--- a/kahuna/public/js/search/results.js
+++ b/kahuna/public/js/search/results.js
@@ -37,6 +37,7 @@ results.controller('SearchResultsCtrl', [
     '$stateParams',
     '$window',
     '$timeout',
+    '$log',
     'delay',
     'onNextEvent',
     'scrollPosition',
@@ -52,6 +53,7 @@ results.controller('SearchResultsCtrl', [
              $stateParams,
              $window,
              $timeout,
+             $log,
              delay,
              onNextEvent,
              scrollPosition,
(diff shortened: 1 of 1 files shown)


kahuna/public/js/search/results.js
```diff
@@ -146,6 +151,8 @@ results.controller('SearchResultsCtrl', [
             ctrl.imagesAll = [];
             ctrl.imagesAll.length = Math.min(images.total, ctrl.maxResults);
 
+            imagesPositions = new Map;
```
jamesgorrie: You don't need to `()`?

kahuna/public/js/search/results.js
```diff
@@ -170,12 +177,24 @@ results.controller('SearchResultsCtrl', [
...
+
+                    // If image already present in results at a
+                    // different position (result set shifted due to
+                    // items being spliced in or deleted?), get rid of
+                    // item at its previous position to avoid
+                    // duplicates
+                    const existingPosition = imagesPositions.get(imageId);
+                    if (angular.isDefined(existingPosition) &&
+                        existingPosition !== position) {
+                        $log.info(`Detected duplicate image ${imageId}, ` +
```
jamesgorrie: Why `$log`? Could be good when we debug on users desktop - just wondering.

jamesgorrie:
:+1:

kahuna/public/js/search/results.js
theefer: Yes, this problem is such a pain I thought it's good to give a bit of info in case we need to debug in the wild.

kahuna/public/js/search/results.js
theefer: No. If there are no params you don't need `()` to `new` functions.

theefer: Meh, we enforce parens. Vaguely convinced by [the explanation](https://jslinterrors.com/missing-invoking-a-constructor), so will add them here.

theefer:
Going to ship this, we can rollback if it causes unforeseen issues.

PR #1054, opened 2015-08-07
screengrabs should be free

jamesgorrie:
Agreed with downstairs that screengrabs can be free, so should be optionally restricted.


theefer:
:+1:

PR #1055, opened 2015-08-07
Add Your recent uploads in "more" menu

theefer:
Also replaces vertical dots with icon.

# Before

screenshot-2015-08-07t16 35 53

# After

screenshot-2015-08-07t16 37 36

### DIFF
--- a/kahuna/public/stylesheets/main.css
+++ b/kahuna/public/stylesheets/main.css
@@ -1203,7 +1203,7 @@ FIXME: what to do with touch devices
 }
 
 /*flex-basis values chosen to accommodate longest key*/
-.image-info__group--dl__key { 
+.image-info__group--dl__key {
     flex-basis: 40%;
 }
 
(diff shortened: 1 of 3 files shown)


kahuna/public/stylesheets/main.css
```diff
@@ -1225,7 +1225,7 @@ FIXME: what to do with touch devices
     flex-basis: 45%;
 }
 
-.image-info__group--dl__value--full-metadata {    
+.image-info__group--dl__value--full-metadata {
```
jamesgorrie: Always found it funny to call a class after an element.

jamesgorrie:
:+1:

alastairjardine:
I know it's merged, but could we make it a fully clickable button so that the target area is larger? IIRC the target area is quite a bit smaller than the other buttons? Create an issue? It'll be used a lot more now it's got some "real" functionality

theefer:
Will take a look.

PR #1056, opened 2015-08-07
only show restrictions when they're required

jamesgorrie:
Inspiration:
* Talking to Jonny
* Talking to Joanna
* Looking through the data and finding lots of "No restrictions"

This only shows the restrictions box if they are required or you explicitly say that you want to add some.

restrictions-ui

### DIFF
--- a/kahuna/public/js/usage-rights/usage-rights-editor.html
+++ b/kahuna/public/js/usage-rights/usage-rights-editor.html
@@ -12,7 +12,6 @@
         <!-- We reset the model here else properties remain attached to the\
         model, even though they don't exist in the form -->
         <select
-            class="full-width"
             ng:model="ctrl.category"
             ng:disabled="ctrl.saving"
             ng:options="category as category.name for category in ctrl.categories track by category.value"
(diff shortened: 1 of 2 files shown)


jamesgorrie:
As @theefer noticed, it shows and requires the restrictions on restricted categories.

kahuna/public/js/usage-rights/usage-rights-editor.html
```diff
@@ -39,94 +38,113 @@
...
                         name="{{ property.name }}"
-                        ng:if="property.optionsMap"
                         ng:model="ctrl.model[property.name]"
-                        ng:required="property.required"
-                        ng:options="o for o in ctrl.getOptionsFor(property)">
-                        <option value="">None</option>
-                    </select>
-
+                        ng:switch-when="text"
+                        ng:required="property.required">
```
theefer: Stupid question: having this whitespace in the textarea doesn't affect its initial value, or does it?

theefer:
:+1: 

Bonus: would be nice to focus the restrictions field when checking the "restricted" checkbox, but these things are often fiddly in Angular so not worth spending too much time on.

jamesgorrie:
I want to see how useful this is first and then I'll work on the focus.

PR #1058, opened 2015-08-07
Actually send Angular errors to Sentry

theefer:
Was having a quick look and it turns out because Angular errors don't bubble to the window, they were never sent to Sentry. Doh.

This code snippet is borrowed from Composer, and tested locally. In theory we could try to use [angular-raven](https://github.com/gdi2290/angular-raven), but in practice it doesn't give us much more than the 7 lines of code below (and more configuration hassle).

### DIFF
--- a/kahuna/public/js/sentry/sentry.js
+++ b/kahuna/public/js/sentry/sentry.js
@@ -3,6 +3,12 @@ import raven from 'raven-js';
 
 export var sentry = angular.module('sentry', []);
 
+const httpErrorProps = ['config', 'data', 'headers', 'status', 'statusText'];
+function isHttpError(obj) {
+    const objKeys = Object.keys(obj);
+    return httpErrorProps.every(key => objKeys.indexOf(key) !== -1);
+}
+
 sentry.factory('sentryEnabled', ['sentryDsn', function(sentryDsn) {
     return angular.isString(sentryDsn);
 }]);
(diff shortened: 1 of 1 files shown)


kahuna/public/js/sentry/sentry.js
```diff
@@ -3,20 +3,18 @@ import raven from 'raven-js';
...
-sentry.factory('sentry', [function() {
-
-    function trigger(errorMessage, extra) {
-        raven.captureException(new Error(errorMessage), { extra });
-    }
-
-    return { trigger };
+// TODO: Alternatively could investigate angular-raven
+sentry.config(['$provide', function ($provide) {
+    $provide.decorator('$exceptionHandler', ['$delegate', function ($delegate) {
```
jamesgorrie: Does this capture http errors too? If so, it probably shouldn't.

jamesgorrie:
> we could try to use angular-raven

I haven't really found any that give that much value either.
Nice find :+1:.

kahuna/public/js/sentry/sentry.js
theefer: Good catch, it is.

I was hoping to be able to do something clever where only uncaught HTTP error would bubble, unfortunately **all** throw exceptions (as opposed to returning `$q.reject(...)`) bubble to the global error handler no matter whether you catch them or not, and that includes HTTP errors.

I've added a fix in another commit.

theefer:
I seem to be getting a fair number of 429 when testing locally, not entirely sure why.

jamesgorrie:
Ah - to sentry? Is that a pay for thing?

jamesgorrie:
:+1: to the new commit

theefer:
We pay for Sentry.

PR #1061, opened 2015-08-10
Turn robots away

theefer:
Our service is pretty much private, so let's tell robots not to index anything (even though there is little access without auth).


jamesgorrie:
:+1:

PR #1062, opened 2015-08-10
Load non moving results

jamesgorrie:
Makes sure we're not loading in a mutating set of results.

~~e.g. When we search for today we're search for `2015-08-10T23:00:00` which will be having images added to it.~~

# before
__Search for a data rage__ =>
search between `2015-08-10T23:00:00.000Z` - `2015-08-11T22:59:59.999Z`

__scroll__
search between `2015-08-10T23:00:00.000Z` - `2015-08-11T22:59:59.999Z`

# after
__Search for a data rage__ =>
search between `2015-08-10T23:00:00.000Z` - `2015-08-11T22:59:59.999Z`

__scroll__
search between `2015-08-10T23:00:00.000Z` - `lastSearchFirstResultTime`

### DIFF
--- a/kahuna/public/js/search/results.js
+++ b/kahuna/public/js/search/results.js
@@ -277,11 +277,15 @@ results.controller('SearchResultsCtrl', [
         function search({until, since, offset, length} = {}) {
             // FIXME: Think of a way to not have to add a param in a million places to add it
 
+            // If we have a lastSearch time we should always respect it.
+            // The reason for this is that we should never be looking into the future, of our
+            // initial search so that we are returning an immutable / identical set, just in
+            // at different offsets. An example would be search for "today", we would first search
+            // for 2015-08-10T23:00:00, get the latest result and then search for that time.
             // Default explicit until/since to $stateParams
             if (angular.isUndefined(until)) {
-                // if we have a `lastSearchTime`, use that instead of `$stateParams` so that if the
-                // stateParams.until was in the future, we're not loading a mutating set of results.
                 until = lastSearchFirstResultTime || $stateParams.until;
+                console.log('until', until)
             }
             if (angular.isUndefined(since)) {
                 since = $stateParams.since;
--- a/kahuna/public/js/search/results.js
+++ b/kahuna/public/js/search/results.js
@@ -222,7 +222,7 @@ results.controller('SearchResultsCtrl', [
         function checkForNewImages() {
             $timeout(() => {
                 const latestTime = lastSearchFirstResultTime;
-                // Blank any 'until' parameter to look for new images
+                // Use explicit `until`, or blank it to find new images
                 search({since: latestTime, length: 0, until: $stateParams.until || null}).then(resp => {
                     // FIXME: minor assumption that only the latest
                     // displayed image is matching the uploadTime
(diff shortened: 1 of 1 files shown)


theefer:
I don't really understand this, could you elaborate or give more examples, esp when the various parameters are or aren't set?

jamesgorrie:
To be honest - I'm not sure the date selecting code is massively clear - I am having a run through it at the moment...

theefer:
I think I may just have understood what you're trying to do. Because there are three possible inputs for `until`, there may be a clearer & safer way to write this.

jamesgorrie:
Just updated it to respect `lastSearchFirstResultTime` if it's set - otherwise it'll use the `$stateParams.until` for the first time you do a dated search.

Seems simpler / easier to understand now?

kahuna/public/js/search/results.js
```diff
@@ -223,9 +223,7 @@ results.controller('SearchResultsCtrl', [
             $timeout(() => {
                 const latestTime = lastSearchFirstResultTime;
                 // Blank any 'until' parameter to look for new images
-                // TODO: if a manual until was set (e.g. using date
-                // picker), don't check for new images until now
-                search({since: latestTime, length: 0, until: null}).then(resp => {
+                search({since: latestTime, length: 0, until: $stateParams.until || null}).then(resp => {
```
theefer: Could update the comment above to make this clearer (e.g. "Use explicit 'until' parameter, or else blank 'until'")

kahuna/public/js/search/results.js
```diff
@@ -281,7 +279,9 @@ results.controller('SearchResultsCtrl', [
 
             // Default explicit until/since to $stateParams
             if (angular.isUndefined(until)) {
-                until = $stateParams.until || lastSearchFirstResultTime;
+                // if we have a `lastSearchTime`, use that instead of `$stateParams` so that if the
+                // stateParams.until was in the future, we're not loading a mutating set of results.
+                until = lastSearchFirstResultTime || $stateParams.until;
```
theefer: So this is based on the assumption that `lastSearchFirstResultTime` is always earlier than `$stateParams.until`, because `$stateParams.until` would have been used the first time to fetch the results which resulted in `lastSearchFirstResultTime` being set?

Makes sense, but given how fiddly that rationale is, and brittle in the face of change, might be worth expliciting it here in the comments so the next person (who may well be us :-) ) doesn't have to think for too long.

theefer:
This change seems to only fix an issue when `until` is set in `$stateParams` (i.e. in the URL), but that didn't seem to be the case for the user who had the issue this morning? Selecting Today doesn't set `until` AFAICS.

jamesgorrie:
This gets set whenever until is not explicitly sent over.
The only time we ever do this is when we are searching for new images.

theefer:
:+1: although I'm still sceptical whether that's meant to solve the issue you reported, as I don't think it could?

jamesgorrie:
I've just realised my description is false to what I am trying to solve.
I've updated it hopefully not describing something different.

theefer:
# Data Rage!!

data rage

theefer:
:+1: 

Let's keep an ear out in case anyone else reports the same problems with super flaky results?

PR #1063, opened 2015-08-10
Fix command to configure github registry on CI

theefer:
Hoping this will resolve the current failures to build kahuna on TC.


jamesgorrie:
So this has never worked?

:+1:

theefer:
No I think it got renamed with a JSPM upgrade and somehow things still worked in the meantime, maybe because someone else (frontend?) was helping set this global auth on the TC box.

theefer:
Ah JSPM 0.16 beta just got a new version out, likely related.

theefer:
Not sure this branch has fixed the issue actually.

PR #1065, opened 2015-08-10
Show filename (if any) on image preview page

theefer:
Note:
* Also split out uploader onto its own row
* Filename gets cropped with an ellipsis if too long. The full name is revealed as a hover tooltip.

image


jamesgorrie:
:+1:

PR #1066, opened 2015-08-10
Make ftp-watcher send filenames to image-loader

theefer:
Had to fiddle with the encoding to get proper support for accents etc.


jamesgorrie:
:+1:

PR #1067, opened 2015-08-11
better logic behind showing restrictions box

jamesgorrie:
The box isn't remembering, when you return to the page, that the image has restrictions.
Now it does.


jamesgorrie:
Explainers:

# :boom: 
usage-rights-boken

# :hammer: 
usage-rights-broken-not

theefer:
:+1:

PR #1071, opened 2015-08-11
Rename JSPM env var to avoid conflict

theefer:
`JSPM_GITHUB_AUTH_PASSTOKEN` is used in latest jspm to include the hashed user+token, rather than just the token. This avoid this conflicts by renaming the env var so we can use either version.


theefer:
I've added the new var in TC so let's see if it builds.

theefer:
It built fine.

PR #1073, opened 2015-08-12
Centre thumbnails on upload page.

akash1810:
Most thumbnails are centred, already, this is for the odd vertical image:

Before:
screen shot 2015-08-12 at 09 45 40

After:
screen shot 2015-08-12 at 09 45 25


jamesgorrie:
:+1:

PR #1074, opened 2015-08-12
add 'crown copyright' usage rights category

jamesgorrie:
There are 750+ images from picdar with this set from the copyright groups.
There are still more being added (there's a few from 2015).

This means it's needed for legacy and present-day.


common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
```diff
@@ -302,3 +304,18 @@ object Pool {
  implicit val jsonReads: Reads[Pool] = Json.reads[Pool]
  implicit val jsonWrites: Writes[Pool] = UsageRights.defaultWrites
 }
+
+
+case class CrownCopyright(restrictions: Option[String] = None)
```
jamesgorrie: I thought we could have something like:
```
Copyright(copyright: String) {
  def cost = copyright match {
    case "crown-copyright" => Free
    case _ => Pay
  }
}
```

But it starts to feel hierarchical i.e. `UsageRights :> Copyright :> CopyrightType`

theefer:
:+1:

PR #1075, opened 2015-08-12
Adding 16:9 cropping ratio for video.

akash1810:
screen shot 2015-08-12 at 12 43 40

cc @blishen @na7hangood

### DIFF
--- a/kahuna/public/js/crop/controller.js
+++ b/kahuna/public/js/crop/controller.js
@@ -21,6 +21,7 @@ crop.controller('ImageCropCtrl',
     // Standard ratios
     ctrl.landscapeRatio = 5 / 3;
     ctrl.portraitRatio = 2 / 3;
+    ctrl.videoRatio = 16 / 9;
     ctrl.freeRatio = null;
 
     const originalDimensions = image.data.source.dimensions;
@@ -59,6 +60,8 @@ crop.controller('ImageCropCtrl',
             return '5:3';
         } else if (Number(aspect) === ctrl.portraitRatio) {
             return '2:3';
+        } else if (Number(aspect) === ctrl.videoRatio) {
+            return '16:9';
         }
         // else undefined is fine
     };
--- a/kahuna/public/js/crop/view.html
+++ b/kahuna/public/js/crop/view.html
@@ -18,6 +18,11 @@
             <gr-icon>crop_portrait</gr-icon><span class="top-bar-item__label">portrait</span> ({{::ctrl.getRatioString(ctrl.portraitRatio)}})
         </label>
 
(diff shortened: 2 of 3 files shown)


na7hangood:
Proved - this does fix the blurry image problem on video embeds.

paperboyo:
15:9. 16:9. :sob: Maybe one day we should think about safe-area guides and some autocropping? Guidess would help with "bisons" as well.

akash1810:
What do you mean by bisons? (a google search was, obviously, not immediately helpful...)

theefer:
:+1:

alastairjardine:
Just a thought – we should update the button styling here to match the styling in the The Guardian | The Observer toggle in the rights picker. Radios are so last month.

akash1810:
@alastairjardine I've already made a note of that on my pad :smile: :smile:

paperboyo:
Showcased main media on feature tone articles cropping automatically from 5:3 to 5:2 on widest breakpoint, like [INTERNAL RESOURCE, look for "bison"] https://sites.google.com/a/guardian.co.uk/next-gen-fronts-design. Users are never sure what exactly gonna get cropped (harder to tell, the more initial main media strays from 5:3).

theefer:
When we want a specific ratio, we should be able to crop that ratio imo, rather than wild guessing and trying to hint. If several ratios should be allowed for the same image slot, then we should support that too.

Until we support multiple ratios, if we can do a hack to preview the 5:2 out of the 5:3, maybe that's worth it. However I don't think that applies to the 16:9, which is a single known other ratio. Or do you disagree? It's true that it's close to 5:3...

alastairjardine:
@akash1810 
moar-cat

paperboyo:
Just the closeness of it, "wasting" precious toolbar space and possible user errors... Idea was to use guides to indicate what will get changed/cropped WOULD that 15:9 become 16:9 for video.

"Bison" helpers are separate issue, obviously, something for the future (the whole logic is less and less obvious for main medias that are NOT 5:3 to begin with).

Above are just thoughts. For the future. If that fixes something, go for it, do not listen to my rumblings! Sorry, have to run...

theefer:
To avoid user errors, we should apply clear restrictions in Composer where we know there are some. E.g. can't pick a non-5:3 as trail, can't pick a non-16:9 as video poster image.

paperboyo:
> we should apply clear restrictions in Composer

Yes. Maybe, also, when Grid is called for modally for certain asset types, certain crop options should not be available at all?

theefer:
Yes agreed, that's part of what I had in mind.

paperboyo:
Thanks! :smiley:

PR #1078, opened 2015-08-12
boolFilter for archived field

akash1810:
Using `boolFilter` on the `archived` field to determine if an image can be deleted.


jamesgorrie:
Does missing then get considered as `false`?

theefer:
As long as what @jamesgorrie is true :+1:

akash1810:
> Does missing then get considered as false?

Yep.

PR #1079, opened 2015-08-12
Adding delete option within an upload job.

akash1810:
Similar to the uploads page, just within an upload job.

### DIFF
--- a/kahuna/public/js/upload/controller.js
+++ b/kahuna/public/js/upload/controller.js
@@ -50,7 +50,7 @@ upload.controller('UploadCtrl', [
 
         ctrl.onDeleteError = function (err) {
             if (err.body.errorKey === 'image-not-found') {
-                $state.go('search');
+                $state.go('upload', {}, {reload: true});
             } else {
                 $window.alert(err.body.errorMessage);
             }
--- a/kahuna/public/js/upload/jobs/upload-jobs.html
+++ b/kahuna/public/js/upload/jobs/upload-jobs.html
@@ -28,7 +28,14 @@
 
         <ui-image-editor ng:if="job.image"
                          image="job.image"
-                         with-batch="ctrl.jobs.length > 1"
+                         with-batch="ctrl.jobs.length > 1">
         </ui-image-editor>
+
+        <gr-delete-image class="right"
+                         ng:if="job.canBeDeleted"
+                         image="job.image"
+                         on-success="ctrl.onDeleteSuccess"
+                         on-error="ctrl.onDeleteError">
+        </gr-delete-image>
     </li>
 </ul>
--- a/kahuna/public/js/upload/jobs/upload-jobs.js
+++ b/kahuna/public/js/upload/jobs/upload-jobs.js
@@ -8,8 +8,8 @@ export var jobs = angular.module('kahuna.upload.jobs', ['kahuna.preview.image', 
 
 
 jobs.controller('UploadJobsCtrl', [
-    '$scope', '$state', '$window', 'apiPoll', 'track', 'mediaApi',
-    function($scope, $state, $window, apiPoll, track, mediaApi) {
+    '$window', 'apiPoll', 'track', 'mediaApi',
+    function($window, apiPoll, track, mediaApi) {
 
     var ctrl = this;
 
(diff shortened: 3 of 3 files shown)


kahuna/public/js/upload/controller.js
```diff
@@ -46,7 +46,7 @@ upload.controller('UploadCtrl', [
 
         ctrl.onDeleteError = function (err) {
             if (err.body.errorKey === 'image-not-found') {
-                $state.go('search');
+                $state.go('upload', {}, {reload: true});
```
theefer: Is that going to cause going from Your current uploads to Last 50 uploads?

kahuna/public/js/upload/jobs/upload-jobs.js
```diff
@@ -60,6 +66,21 @@ jobs.controller('UploadJobsCtrl',
...
+        var index = ctrl.jobs.findIndex(i => i.image.data.id === image.data.id);
+
+        if (index > -1) {
+            ctrl.jobs.splice(index, 1);
+        }
+    };
+
+    ctrl.onDeleteError = function (err) {
+        if (err.body.errorKey === 'image-not-found') {
+            $state.go('upload', {}, {reload: true});
```
akash1810: not sure if this is desirable or not as it will mean the batch operations disappear.

kahuna/public/js/upload/controller.js
akash1810: better behaviour imo.

kahuna/public/js/upload/jobs/upload-jobs.js
```diff
@@ -2,12 +2,14 @@ import angular from 'angular';
...
+import '../../components/gr-delete-image/gr-delete-image';
 
 export var jobs = angular.module('kahuna.upload.jobs', ['kahuna.preview.image', 'analytics.track']);
 
 
-jobs.controller('UploadJobsCtrl',
-                ['$scope', 'apiPoll', 'track', function($scope, apiPoll, track) {
+jobs.controller('UploadJobsCtrl', [
+    '$scope', '$state', '$window', 'apiPoll', 'track', 'mediaApi',
+    function($scope, $state, $window, apiPoll, track, mediaApi) {
```
theefer: `$scope` not needed?

kahuna/public/js/upload/jobs/upload-jobs.js
theefer: Yeah, that sounds annoying? Is this necessary?

kahuna/public/js/upload/controller.js
akash1810: > Is that going to cause going from Your current uploads to Last 50 uploads?

This is within the `Last 50 Uploads`... if a delete fails, reload the page rather than go to search page, which is better behaviour imo.

kahuna/public/js/upload/jobs/upload-jobs.js
akash1810: Might be better to window.error it in this instance.

kahuna/public/js/upload/controller.js
theefer: Ah yes :+1:

kahuna/public/js/upload/jobs/upload-jobs.js
theefer: I'd have thought all we want to do is remove the image from the list?

akash1810: But the deletion would have failed, so when the page is refreshed/you view your uploads, the image you thought you deleted somehow reappears. Might be confusing?

kahuna/public/js/upload/jobs/upload-jobs.html
```diff
@@ -28,7 +28,14 @@
 
         <ui-image-editor ng:if="job.image"
                          image="job.image"
-                         with-batch="ctrl.jobs.length > 1"
+                         with-batch="ctrl.jobs.length > 1">
```
theefer: Browsers are really too kind.

kahuna/public/js/upload/jobs/upload-jobs.js
```diff
@@ -32,6 +34,10 @@ jobs.controller('UploadJobsCtrl',
                 jobItem.image = image;
                 jobItem.thumbnail = image.data.thumbnail;
 
+                mediaApi.canDelete(image).then(deletable => {
```
theefer: I think this needs updating with latest master.

theefer:
:+1:

Issue #830, opened 2015-06-16
Add Batch download


paperboyo:
Hello,

Tell me if feature requests don't belong here, I try to put here only the obvious ones, but hey - you know me :)

Downloading more than one image is currently a pain. Even more so in Firefox which does not support our downloading mechanism properly (opens an image in a new tab from where one has to Save the image as...).

An ability to download multiple selected photos would be cool. Most probably they would have to be zipped (stored), would be even better if they wouldn't (user would just have to specify a destination folder).

Regards
Mateusz

alastairjardine:
UI perspective: selection would enable this button on the new fixed toolbar.

paperboyo:
Done in https://github.com/guardian/grid/pull/1080. Thanks. Some very minor things to fix (or not) in https://github.com/guardian/grid/issues/1123#issuecomment-133298077.

Issue #1041, opened 2015-08-06
Add delete button to toolbar on picture grid page


akash1810:
_From @alastairjardine on August 6, 2015 9:57_

We have a toolbar, we should use it!

_Copied from original issue: guardian/grid-infra#107_

akash1810:
Any ideas on the UX on this when multiple images are selected, as currently:
- anyone can delete their own uploads
- the ability to delete any image is permission based

What'll happen when your selection includes an image that you're not allowed to delete for whatever reason?

akash1810:
I was thinking of having adding the delete action to the hover items. Good/bad?

akash1810:
_From @alastairjardine on August 6, 2015 10:6_

Perhaps enable only for the people who have the permission in the first instance.

akash1810:
_From @theefer on August 6, 2015 10:7_

> I was thinking of having adding the delete action to the hover items. Good/bad?

It's such a rare action, seems unwarranted noise?

akash1810:
btw, wrong repo?

akash1810:
_From @alastairjardine on August 6, 2015 10:15_

@akash1810 Probably :(

akash1810:
_From @alastairjardine on August 6, 2015 10:16_

@akash1810 I would try to keep the image functions to the most frequently used.

akash1810:
> It's such a rare action, seems unwarranted noise?

and

>  I would try to keep the image functions to the most frequently used.

Slightly related - presumably this also goes for a download button? i.e. the download button goes to the toolbar.

theefer:
I would have said so.

alastairjardine:
> Slightly related - presumably this also goes for a download button? i.e. the download button goes to the toolbar. 

Yep, _especially_ as you might want to download a multiple image selection

kenoir:
> Any ideas on the UX on this when multiple images are selected

Actions that cannot be performed on the full batch result in a disabled button.

If you're looking for actions to go in the toolbar, i'd try the archiver as it's probably pretty straightforward to implement and a reasonably common action.

akash1810:
Any pointers on the behaviour of the search results? My current thinking is the thumbnail should disappear but the placeholder square should remain (until reload).

alastairjardine:
Sounds good. Perhaps ghost in a trashed symbol to make it super clear
that's what happened?

On Friday, 14 August 2015, Akash A <notifications@github.com> wrote:

[...]

Issue #823, opened 2015-06-12
Allow downloading original images with metadata overrides


paperboyo:
Hello,
Currently, when I upload an image to Grid, fill in metadata like e.g. Description and then download it from Grid, the downloaded image does not contain that Description.

Currently, when I download an original agency image from Grid, work on it and upload that modified variant image, there is no way for Grid to be sure that it’s a derivative and of which image exactly.

While not a priority, I think it would be beneficial for the Grid to add it’s own metadata to “original” images.

Then it would be possible to create a tool that, for example, opens several Grid “originals” as layers in Photoshop and then Grid upon upload would correctly identify sources and mixed usage rights of that composite image.

Regards
Mateusz

paperboyo:
Some useful resources:

https://wwwimages2.adobe.com/content/dam/acom/en/devnet/xmp/pdfs/XMP%20SDK%20Release%20cc-2016-08/XMPSpecificationPart1.pdf

https://wwwimages2.adobe.com/content/dam/acom/en/devnet/xmp/pdfs/XMP%20SDK%20Release%20cc-2016-08/XMPSpecificationPart3.pdf#page=52&zoom=auto,-163,613

http://www.metadataworkinggroup.com/pdf/mwg_guidance.pdf#page=37&zoom=140,-169,464 (and further).

PR #1080, opened 2015-08-12
Client download

jamesgorrie:
Downloading the originals in one tidy Zip.
Files will have either their filename or hash.

The filename for the zip is generated by JSZip - I want to change this, but will do it in another PR which will do the same to the single download.

downloader3


cloud-formation/dev-template.json
```diff
@@ -1,6 +1,12 @@
 {
     "AWSTemplateFormatVersion": "2010-09-09",
     "Description": "Media Service Dev",
+    "Parameters": {
```
jamesgorrie: Not sure if this is still needed, I will try without it.

jamesgorrie: Actually, using xhr, this will be required.

kenoir:
Nice. :+1:

media-api/conf/application.conf
```diff
@@ -4,7 +4,7 @@ session.httpOnly=false
 session.secure=true
 
 es.cluster = "media-service"
-es.host = "localhost"
+es.host = "ec2-52-16-168-116.eu-west-1.compute.amazonaws.com"
```
kenoir: Should this be here?

jamesgorrie: No, I am removing it from history.

kahuna/public/js/components/gr-panel/gr-panel.html
```diff
@@ -35,13 +35,15 @@
...
-                    <gr:usage-rights-editor
+                    <gr-usage-rights-editor
                         gr:usage-rights="ctrl.usageRights"
                         gr:on-save="showUsageRights = false"
                         gr:on-cancel="showUsageRights = false">
-                    </gr:usage-rights-editor>
+                    </gr-usage-rights-editor>
                 </div>
 
+                <gr-downloader gr:images="ctrl.images"></gr-downloader>
```
theefer: Should this be at the bottom of the panel instead (by the archiver), while we wait for a bar to put all those actions on?

kahuna/public/js/components/gr-panel/gr-panel.js
```diff
@@ -37,6 +40,7 @@ grPanel.controller('GrPanel', [
         $scope,
         $window,
         $q,
+        $http,
```
theefer: Not used?

kahuna/public/js/components/gr-panel/gr-panel.html
jamesgorrie: Yip - still working out if `xhr` is the way - which it is I think.

jamesgorrie:
Ready!

cloud-formation/dev-template.json
```diff
@@ -58,7 +64,25 @@
...
+                            "AllowedMethods": [
+                                "GET"
+                            ],
+                            "AllowedOrigins": [
+                                "*"
+                            ],
+                            "ExposedHeaders": [
+                                "Date"
+                            ],
+                            "MaxAge": "3600"
```
theefer: Can we check if this affects the headers served through nginx imgops? Why 1h actually?

kahuna/public/stylesheets/main.css
```diff
@@ -1192,7 +1192,12 @@ FIXME: what to do with touch devices
     border-bottom: 1px solid #565656;
 }
 
-.image-info:last-child .image-info__group{
+.image-info__group--last {
+    border-bottom: 0;
+    clear: both;
+}
+
+.image-info:last-child .image-info__group {
```
theefer: How is that different from the above?

theefer:
Could we customize the ZIP filename, e.g. `grid-download-DATE.zip`?

alastairjardine:
Button for download needs to go into the toolbar. This is the place for actions like this now. Metadata panel needs to have one function, not many. Thanks!

kahuna/public/stylesheets/main.css
jamesgorrie: That must have been left there from somewhere, I've removed it.
One was targeting `.image-info:last-child` were as I wanted the last `.image-info__group`.

I suppose I could use `.image-info__group:last-child` - but I try to avoid it to be explicit.

cloud-formation/dev-template.json
jamesgorrie: It doesn't - but this was a copy paste muck up.
If we leave them out it respects the normal settings.

jamesgorrie:
@alastairjardine We have a few problems with this technically i.e. panel is the only thing that is aware of the selection.

alastairjardine:
@jamesgorrie Oh no! In what timeframe can we get this fixed? Uncomfortable about letting things go out that don't make sense in the new structure. 

If this does go out, can we chase it quickly with a PR to make the toolbar selection aware?

akash1810:
@alastairjardine this also applies to the [Delete button](https://github.com/guardian/grid/issues/1041) I'm working on atm :pensive: 

@kenoir started work on the refactoring to allow us to easily add buttons to the toolbar, but it turned into a rabbit hole - I'll pick it up w/@theefer as discussed on Slack this morning.

alastairjardine:
@akash1810 It sounds like there are enough cases where it would be useful. Can I possibly ask y'all to push it a bit up your priority list?

theefer:
@jamesgorrie :+1: (any thoughts on name of ZIP file?)

jamesgorrie:
I was going to just have `grid-downloads-{count}-{date}.zip`?

jamesgorrie:
@alastairjardine Hopefully we can. We won't publicise it until it's in the rights place.

theefer:
> I was going to just have grid-downloads-{count}-{date}.zip?

+1

theefer:
Actually you might want to put the date before any count for saner ordering purposes if ordered by filename (i.e. gets ordered by date, rather by the number of images you downloaded).

jamesgorrie:
I thought grid-down... so it doesn't get mixed with other numbers / dates.

jamesgorrie:
Wait, got you.

paperboyo:
This is super-useful, fixes https://github.com/guardian/grid/issues/830, I'll be using it a lot, thanks!

> Files will have either their filename or hash.

At the danger of sounding like a user who requires his phone to make him coffee and have a shower, could I suggest for the files to contain both the original filename (before) **and** the hash (after) in the filename.

Reason (subject to above reservation) being that sometimes original filenames are useful for sorting in Finder/Explorer, while - lacking https://github.com/guardian/grid/issues/823 - the hash can be used manually to locate the original. And also can be used programatically as a poor man's replacement when not having https://github.com/guardian/grid/issues/823.

> Actually you might want to put the date before any count

+1

jamesgorrie:
@paperboyo I will work on the naming in another PR to keep this one (slightly) cleaner and we can discuss it there.

jamesgorrie:
@theefer @alastairjardine added a little something here - shall we mix these together or potentially do this-then-that?

https://github.com/guardian/grid/pull/1087

alastairjardine:
Nice. Happy to get it out with no publicity then chase quickly with a bump into the toolbar

theefer:
:+1:

PR #1086, opened 2015-08-13
refactoring

akash1810:
Move `canDelete` to image service.

### DIFF
--- a/kahuna/public/js/image/service.js
+++ b/kahuna/public/js/image/service.js
@@ -38,7 +38,8 @@ imageService.factory('imageService', ['editsService', function(editsService) {
         return {
             cost: image.data.cost,
             hasCrops: hasExportsOfType(image, 'crop'),
-            isValid: image.data.valid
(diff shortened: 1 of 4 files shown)


jamesgorrie:
Descriptive PR titles are nice for searching, notifications etc :smile_cat:
:+1:

kahuna/public/js/image/service.js
```diff
@@ -38,7 +38,8 @@ imageService.factory('imageService', ['editsService', function(editsService) {
         return {
             cost: image.data.cost,
             hasCrops: hasExportsOfType(image, 'crop'),
-            isValid: image.data.valid
+            isValid: image.data.valid,
+            canDelete: image.getAction('delete').then(action => !! action)
         };
```
theefer: I think we should rework this service soon, it doesn't make sense to do all the work it does if you're only interested in a single one of them.

theefer:
:+1:

PR #1087, opened 2015-08-13
Client download + action bar

jamesgorrie:
Added it to the action bar instead.
I haven't moved it out as a component, this is more proof of concept (and potentially first-pass).

downloader-action-panel

### DIFF
--- a/kahuna/public/js/components/gr-panel/gr-panel.html
+++ b/kahuna/public/js/components/gr-panel/gr-panel.html
@@ -267,7 +267,7 @@
             </ul>
         </div>
 
-        <div class="image-info__group batch-archive" ng:switch="ctrl.archivedState"
+        <div class="image-info__group image-info__group--last batch-archive" ng:switch="ctrl.archivedState"
             ng:class="{'batch-archive--archiving': ctrl.archiving}">
             <div ng:switch-when="mixed" class="batch-archive--mixed">
                 <div class="metadata-line">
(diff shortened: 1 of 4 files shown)


alastairjardine:
Nice. I would range it left rather than right. Actually, we should probably have 

[ Ordering | Count | Clear selection | Download <space space space> Meta control (pinned right) ]

jamesgorrie:
I've done it in that order - I personally quite like to Google way of adding the seletion count and close together as they explain each other...

__as above__
action-bar

__similar to Photo's below__
action-bar-2

__Photos__
close-selection

theefer:
I also find clearer to have the selection count (not necessarily the total count, which is unrelated) and selection clear together, as well as selection actions. I know @kenoir had tried something around that but he wasn't convinced by the end result.

Probably not worth bike shedding too deep here as we're talking about more changes to the action bar on selection.

jamesgorrie:
> we're talking about more changes to the action bar on selection

Is this a decent enough test to put infront of people?
A big question people have asked, nearly every-time I've shown multi-select, is, "how do I get out of here". Either way - I thought this a good place to start a conversation.

theefer:
Not sure moving the Clear in the middle of an indistinguishable mix of labels and buttons makes it particularly stand out..? :-)

alastairjardine:
I suspect it might be more clear than on the RHS as an explicity labelled thing, especially if the positioning is relatively fixed and consistent. I'd love to link them together a la google, but I think we'll have more going on longer term.

jamesgorrie:
We could have the actions on the images not be separated?

action-bar-downloads

PR #1089, opened 2015-08-13
Support API params for filtering by date modified/taken since/until

theefer:
Also retired the fromDate/toDate which was redundant with since/until.

### DIFF
--- a/media-api/README.md
+++ b/media-api/README.md
@@ -12,15 +12,15 @@
 | offset     | Start offset for results                  | 0
 | length     | Maximum number of results                 | 10
 | orderBy    | Field for ordering (prepend '-' for DESC) | -uploadTime
-| fromDate   | Search images uploaded since this time    | [no lower bound]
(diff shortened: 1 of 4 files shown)


jamesgorrie:
Good to watch performance on these as the ordering was a little slower (I'm just wondering how ES deals with dates).

:+1:

theefer:
> as the ordering was a little slower

Where did you see that? Just curious as I don't think the current ordering field (`uploadTime`) is indexed specially?

jamesgorrie:
I remember @akash1810 saying it and also in the demo this morning, although that could have been the wifi.

akash1810:
The inclusion of the `offset` param makes the performance a little worse.

`https://api.[domian]/images?q=&offset=1649680&length=1&orderBy=uploadTime&free=true`

vs 

`https://api.[domain]/images?q=length=1&orderBy=uploadTime&free=true`

theefer:
I don't think this has anything to do with the existing performance issue, so merging.

PR #1091, opened 2015-08-13
[wip] update labels on thumbnail when batch edited.

akash1810:
WIP because I need to work out if we can get rid of the [root event](https://github.com/guardian/grid/blob/master/kahuna/public/js/services/label.js#L15).

before (the blank frame is the page being manually refreshed)
before

after
after


akash1810:
changes have been made in https://github.com/guardian/grid/pull/1095

PR #1095, opened 2015-08-14
`gr-add-label` component.

akash1810:
Standalone button to add labels to an image.

I had this in a branch for a while, and after this was reported:

picture 496

I thought I might as well finish up the work as it fixes the issue by moving the Add button next to the field label (see https://github.com/guardian/grid/issues/993 for an artist's impression).

Here's how it looks now:

# Preview
add-label-preview

# Uploads
add-label-current-uploads

# Image Page
add-label-image

### DIFF
--- a/kahuna/public/js/components/gr-add-label/gr-add-label.css
+++ b/kahuna/public/js/components/gr-add-label/gr-add-label.css
@@ -1,0 +1,8 @@
+gr-add-label .button-shy {
+    box-shadow: none;
+}
+
+gr-add-label .small {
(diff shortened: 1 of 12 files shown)


akash1810:
On TEST now.

theefer:
Good stuff :+1:

paperboyo:
Minor UI point: the plus button does not have a help tooltip consistently with other action buttons on search page (instead an image tooltip is shown). It also does not have a tooltip on your recent uploads page.

PR #1096, opened 2015-08-16
Picdar rights override

jamesgorrie:
Not completely done.

The point here is to run through images that have had their rights set from Picdar, and munge them with rights already set.

The plan is:
* Read images with rights set from Picdar
* Read the API
* Look if we should override
* Mark if overridden or not

The process is explained [here](https://github.com/guardian/grid/pull/1096/files#diff-2e5785c62cc88d070fab1ce07fe913feR54).


picdar-export/src/main/scala/com.gu.mediaservice.picdarexport/lib/cleanup/UsageRightsOverride.scala
```diff
@@ -151,4 +151,8 @@ object UsageRightsOverride {
 
   def getUsageRights(copyrightGroup: String, metadata: ImageMetadata) =
     copyrightGroupToUsageRightsMap.get(copyrightGroup).flatMap(func => func(metadata))
+
+  def getOverrides(currentRights: UsageRights, picdarRights: UsageRights) = {
+
```
jamesgorrie: Still need to figure what the logic is here.
Might be in order of importance:
* Media API overrides
* Picdar rights
* Media API auto rights

We will need to make sure we send over the overiddes from the Media API then.

theefer:
Filled in the missing bits and now proven to work against TEST.

Still need to further cleanup of agency names from what we find in Picdar before we can run this in PROD, to be confirmed with Jo.

theefer:
Merging as now working and generally inert.

Next step will be to only record `agency` images if the supplier is one in our allowed known list. If not, we should error out so that we can review whether the agency needs to be added or supported somehow. Also likely will need to map variants of supplier names to their canonical name, e.g. `PA WIRE` => `PA`.

PR #1097, opened 2015-08-17
Simpler, faster initial query to determine result set size and latest results

theefer:
When ordering by Oldest first currently, we do an extra request to determine the uploadTime of the latest result. That request is very slow and expensive, likely because we request the very last result of the matches. This PR changes the logic to simply always query the result set size and latest result by using a "newest first" ordering.


akash1810:
nice :+1:

PR #1098, opened 2015-08-17
Show errors above top-bar

theefer:
# Before

image

# After

image

### DIFF
--- a/kahuna/public/stylesheets/main.css
+++ b/kahuna/public/stylesheets/main.css
@@ -406,7 +406,7 @@ textarea.ng-invalid {
     left: 0;
     right: 0;
     top: 5px;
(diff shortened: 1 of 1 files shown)


kahuna/public/stylesheets/main.css
```diff
@@ -406,7 +406,7 @@ textarea.ng-invalid {
     left: 0;
     right: 0;
     top: 5px;
-    z-index: 2;
+    z-index: 10;
```
akash1810: is it worth using a similar strategy/style-guide as frontend regarding z-index values? (unfortunately, can't find the link).

theefer: Yes.

PR #1099, opened 2015-08-17
Fix bug with publication selection in usage-rights-editor

theefer:
This was preventing choosing the publication for a photographer on the uploads page, because all fake radios had the same ID. This uses the unique scope `$id` to make the ids unique.

Also fixes the other use of dynamic id to produce [valid IDs](http://www.w3.org/TR/REC-html40/types.html#type-name) (i.e. start with letters, not numbers).


akash1810:
:+1:

PR #1100, opened 2015-08-17
Disable the label hyperlink when in selection mode.

akash1810:
Currently, labels remain clickable when in selection mode - clicking a label when you're in selection mode performs a search and thus cancels your selected images. This stops that - clicking on a label will select/deselect the image.


theefer:
:+1:

PR #1101, opened 2015-08-17
Batch delete images.

akash1810:
Haven't spent too long on styling as this'll move to the toolbar soon (hopefully!).

screen shot 2015-08-17 at 17 41 51

### DIFF
--- a/kahuna/public/js/components/gr-delete-image/gr-delete-image.js
+++ b/kahuna/public/js/components/gr-delete-image/gr-delete-image.js
@@ -3,35 +3,39 @@ import '../gr-confirm-delete/gr-confirm-delete';
 
 export const deleteImage = angular.module('gr.deleteImage', ['gr.confirmDelete']);
 
-deleteImage.controller('grDeleteImageCtrl', ['mediaApi', function (mediaApi) {
+deleteImage.controller('grDeleteImageCtrl', ['$q', 'mediaApi', function ($q, mediaApi) {
     var ctrl = this;
 
-    ctrl.delete = function () {
-        mediaApi.delete(ctrl.image)
+    ctrl.delete = function (image) {
+        return mediaApi.delete(image)
             .then((resp) => {
                 if (angular.isDefined(ctrl.onSuccess)) {
-                    ctrl.onSuccess(resp, ctrl.image);
+                    ctrl.onSuccess(resp, image);
                 }
             })
             .catch((err) => {
                 if (angular.isDefined(ctrl.onError)) {
-                    ctrl.onError(err, ctrl.image);
+                    ctrl.onError(err, image);
                 }
             });
     };
+
+    ctrl.batchDelete = function () {
+        return $q.all(ctrl.images.map(image => ctrl.delete(image)));
+    }
 }]);
 
 deleteImage.directive('grDeleteImage', [function () {
     return {
         restrict: 'E',
         template: `
-            <gr-confirm-delete class="gr-delete-image" gr-on-confirm="ctrl.delete()">
+            <gr-confirm-delete class="gr-delete-image" gr-on-confirm="ctrl.batchDelete()">
             </gr-confirm-delete>`,
         controller: 'grDeleteImageCtrl',
         controllerAs: 'ctrl',
         bindToController: true,
         scope: {
-            image: '=',
+            images: '=',
             onSuccess: '=?',
             onError: '=?'
         }
--- a/kahuna/public/js/components/gr-delete-image/gr-delete-image.js
+++ b/kahuna/public/js/components/gr-delete-image/gr-delete-image.js
@@ -18,7 +18,7 @@ deleteImage.controller('grDeleteImageCtrl', [
         ctrl.delete = function () {
             // HACK to wait for thrall to process the message so that when we
             // poll the api, it will be up to date.
-            return $q.all(ctrl.images.map(image => ctrl.deleteImage(image)))
+            return $q.all(Array.from(ctrl.images.values()).map(image => ctrl.deleteImage(image)))
                 .then(() => {
                     $timeout(() => {
                         $rootScope.$emit('images-deleted', ctrl.images);
--- a/kahuna/public/js/components/gr-delete-image/gr-delete-image.js
+++ b/kahuna/public/js/components/gr-delete-image/gr-delete-image.js
@@ -1,15 +1,29 @@
 import angular from 'angular';
 import '../gr-confirm-delete/gr-confirm-delete';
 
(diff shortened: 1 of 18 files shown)


alastairjardine:
It does need to move soon…v v soon!

kahuna/public/js/components/gr-panel/gr-panel.css
```diff
@@ -61,26 +61,17 @@
...
 
 .batch-archive {
-    float: left;
     width: 100%;
-    box-sizing: border-box;
-    /* ^ because it has elements floating in it */
-}
-
-.batch-archive .metadata-line {
-    float: left;
```
akash1810: :heart: [flexbox](https://github.com/guardian/grid/pull/1101/files#diff-616e7ea78eaec4cf50c56788e7f9711dR272)

akash1810:
I've released this to TEST now.

kahuna/public/js/search/results.js
```diff
@@ -370,6 +370,23 @@ results.controller('SearchResultsCtrl', [
...
+
+            if (imageIndex > -1) {
+                ctrl.images.splice(imageIndex, 1);
+            }
+
+            if (imageAllIndex > -1) {
+                ctrl.imagesAll.splice(imageIndex, 1);
+            }
+
+            ctrl.totalResults += -1;
```
theefer: did you mean `ctrl.totalResults -= 1` or even `ctrl.totalResults--`? :-)

kahuna/public/js/search/results.js
```diff
@@ -370,6 +370,23 @@ results.controller('SearchResultsCtrl', [
             }
         });
 
+        const freeImageDeleteListener = $rootScope.$on('image-deleted', (e, image) => {
```
theefer: Bit crappy to have to do this kind of stuff here, but agree it's necessary until we tidy up all this stateful business.

theefer:
Looks good, but I just tried it in TEST and it looks a bit buggy when deleting two consecutive images. The second doesn't disappear.

I've got a suspicion that images don't like their index changing in the lazy table (an assumption was made that the index never changes iirc). One way to avoid this might be not to remove them from the `images` and `imagesAll` arrays, but instead replace them by `undefined`. Although actually now I think about it it would mean they are not loaded and would be reloaded, causing other duplicates, etc.

Let's catch up tomorrow morning, it's probably a reasonably easy fix.

akash1810:
this is on TEST again, it looks ok, if a little slow because of the setTimeout.

kahuna/public/js/components/gr-delete-image/gr-delete-image.js
```diff
@@ -3,23 +3,30 @@ import '../gr-confirm-delete/gr-confirm-delete';
...
+        };
+
+        ctrl.delete = function () {
+            // HACK to wait for thrall to process the message so that when we
+            // poll the api, it will be up to date.
+            return $q.all(ctrl.images.map(image => ctrl.deleteImage(image)))
+                .then(() => {
+                    $timeout(() => {
+                        $rootScope.$emit('images-deleted', ctrl.images);
+                    }, 1000);
```
akash1810: I wonder if we should make the timeout relative to the number of images? `Math.min(ctrl.images.length * 500, 2000)`?

theefer: Why? They should be processed within milliseconds if the queue is consumed normally.

akash1810: There was an image that took a bit longer to delete on TEST - resulting in the crazy disappearing images. A page refresh sorted, but obv not ideal.

akash1810:
Caught up with master now! This means the delete button is in the toolbar, rather than the panel!

alastairjardine:
Sweet

On Friday, 21 August 2015, Akash A <notifications@github.com> wrote:

> Caught up with master now! This means the delete button is in the toolbar,
> rather than the panel!
>
> —
> <https://github.com/guardian/grid/pull/1101#issuecomment-133504804>.
>


-- 
--
Alastair Jardine, Senior UX Architect

-- 

in England Number 908396

kahuna/public/js/components/gr-delete-image/gr-delete-image.js
```diff
@@ -3,23 +3,30 @@ import '../gr-confirm-delete/gr-confirm-delete';
...
+            return mediaApi.delete(image)
+                .catch((err) => {
+                    $rootScope.$emit('image-delete-failure', err, image);
+                });
+        };
+
+        ctrl.delete = function () {
+            // HACK to wait for thrall to process the message so that when we
+            // poll the api, it will be up to date.
+            return $q.all(Array.from(ctrl.images.values()).map(image => ctrl.deleteImage(image)))
```
jamesgorrie: Why not use the polling mechanic we have else where for checking when thralls done?
Or look into something like [`SSE`s](http://www.html5rocks.com/en/tutorials/eventsource/basics/)?

theefer: I don't like all this polling we're doing, but given how much pain we're having working around issues with this, that might be the best/simplest/mose reliable idea.. @akash1810 what do you think?

akash1810: I'm tempted to try SSEs though am conscious this PR has been open (and wip-ed) for too long, so will time box it.

theefer: Can we try and deliver this feature first? I think @jamesgorrie's suggestion could help fix the current issue blocking release and seems sensible to me.

Not sure how we'd implement SSE in this case without architectural changes, but could be interesting to chat about it when @jamesgorrie is back in so we have an idea if and when it fits on the roadmap.

akash1810: updated w/polling

theefer: Nice one!

theefer:
:+1: great

kahuna/public/js/components/gr-delete-image/gr-delete-image.js
jamesgorrie: Yea, also didn't want to splinter off onto another way of doing things, as this is starting to complicate the code (having multiple ways of doing things). If we'd like to we could abstract it to a checkActionComplete service or something and then muck around with the complexities and different implementations in there to keep it in one place... Cheers @akash1810.

jamesgorrie:
bonza :+1:

PR #1102, opened 2015-08-17
Decouple auth from initial asset load (trigger login via JS, not kahuna page load)

theefer:
As a step toward a simpler setup where the Angular app (incl. `index.html` page) is served statically rather than behind a Play controller that enforces auth. The need for auth'ing is instead detected by a probe API call that, if failed with a 401, triggers a redirect to the login link provided in the response (with the current page URL set as `redirectUri`).

This will lead to a brief flash of the page loading before the user is sent for auth the first time (when they don't have a valid panda cookie), but that's actually quite rare. Most of the time the auth just expires and is refreshed in the background.

If this code works, I'm keen to move it to our libraries so it can be reused.

**Notes**
* This would make a few of our `meta` configuration public, including Mixpanel token, Sentry DSN and wat page URL. The main one I'd suggest protecting is the Sentry DSN; we could expose it in the API instead. Not sure about Mixpanel, as it's nice to have it in the page as early as possible to report errors even if other things are broken. The wat page I'm not too fussed about, it's nice to be able to statically extract it from scripts imho.
* This leaves us more vulnerable to faulty browser settings (the infamous Third-party cookies setting preventing transparent re-auth in the background). Refreshing the page would no longer fix the issue automatically, although that may not be such a bad thing to identify misconfigured browsers as constantly having to reload the page isn't really a great UX. Something to keep an eye on.

### DIFF
--- a/kahuna/app/controllers/Application.scala
+++ b/kahuna/app/controllers/Application.scala
@@ -1,16 +1,11 @@
 package controllers
 
-import play.api.mvc.Controller
+import play.api.mvc.{Action, Controller}
 import lib.Config
-import com.gu.mediaservice.lib.auth.PanDomainAuthActions
 
-// TODO: retire Panda entirely from kahuna, let the JS app and the APIs manage auth
+object Application extends Controller {
 
-object Application extends Controller with PanDomainAuthActions {
-
-  override lazy val authCallbackBaseUri = Config.rootUri
-
-  def index(ignored: String) = AuthAction { req =>
+  def index(ignored: String) = Action { req =>
     Ok(views.html.main(
       Config.mediaApiUri,
       Config.watUri,
--- a/kahuna/app/controllers/Panda.scala
+++ b/kahuna/app/controllers/Panda.scala
@@ -39,11 +39,11 @@ object Panda extends Controller
     }
   }
 
-  // Trigger the auth cycle and return a dummy page
-  // Typically used for automatically re-auth'ing in the background
-  def doLogin = AuthAction { req =>
-    // Note: returning NoContent as iframe content seems to make browsers unhappy
-    Ok("logged in")
+  // Trigger the auth cycle
+  // If a redirectUri is provided, redirect the browser there once auth'd,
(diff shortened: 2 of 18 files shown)


kahuna/app/controllers/Application.scala
```diff
@@ -1,16 +1,11 @@
...
 
-// TODO: retire Panda entirely from kahuna, let the JS app and the APIs manage auth
+object Application extends Controller {
 
-object Application extends Controller with PanDomainAuthActions {
-
-  override lazy val authCallbackBaseUri = Config.rootUri
-
-  def index(ignored: String) = AuthAction { req =>
+  def index(ignored: String) = Action { req =>
```
theefer: This is the main change: the index is no longer auth-protected.

kahuna/app/controllers/Panda.scala
```diff
@@ -12,11 +14,14 @@ import scala.util.{Success, Try, Failure}
...
   with PanDomainAuthActions
+  with ArgoHelpers
   with ArgoErrorResponses {
 
   override lazy val authCallbackBaseUri = Config.rootUri
-  def loginUri: String = Config.loginUri
+  override lazy val loginUriTemplate    = Config.loginUriTemplate
 
+
+  // FIXME: how to use usual Authenticated action helper for this? separate Controller?
```
theefer: I think it's doable but not directly related to this, so leaving for the next PR.

akash1810:
I had a look at this last night - it looks fine to me, though I didn't understand why. Can we talk irl tomorrow?

theefer:
Short answer is: to make kahuna (i.e. our angular client) completely static. It shouldn't need to be part of a deployed Play application. All the Play app does is enforce auth (this removes this, it's not necessary as we can do it in the API layer instead as we already do) and inject some configuration values (could be done at build time). So the goal is to create a standalone micro-service just for auth (essentially the `Panda` controller here) and statically inject the configuration at build time so kahuna is just a set of static files we can host in S3. Makes it simpler and faster to deploy.

PR #1105, opened 2015-08-19
Fixes 'cd is not a command' error

kenoir:
Passes command for imgops and elasticsearch to bash so "cd" is interpreted as a command.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-x-terminal-emulator -T elasticsearch -e "cd elasticsearch; ./dev-start.sh" &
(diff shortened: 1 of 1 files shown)


akash1810:
:+1:

theefer:
Odd (what shell are you using that doesn't have cd?), but  :+1: if it helps you!

kenoir:
I think those commands get passed to `x-terminal-emulator`, which in a default ubuntu setup gets passed to gnome-terminal that then executes the commands without proxying it through a shell. As I understand it `cd` is a builtin and only works because the shell implements the POSIX standard for such things. 

To see your terminal emulator on ubuntu do `ll /etc/alternatives/x-terminal-emulator`

theefer:
Interesting, thanks!

PR #1106, opened 2015-08-19
Using ini file for settings as its simpler.

akash1810:
.ini files are slightly more universal than .py files and running `./main.py` is a lot easier - see the [readme](https://github.com/guardian/grid/blob/aa-dot-props/scripts/dot-properties/README.md)


kenoir:
:+1:

kenoir:
How does this fit with keeping some of this config in S3?

akash1810:
> How does this fit with keeping some of this config in S3?

Similar to before - copy down one single file, add your stack name, run script.

Similar as its now one file, vs two - however ini files are friendlier imo.

theefer:
:+1:

PR #1107, opened 2015-08-19
icon label directive to reduce css repetition

akash1810:
We've got a few `media screen and (max-width: 800px)` - this creates a `gr-icon-label` component that should keep it DRY.

I've not updated everywhere...yet! But that'll be another PR as the remaining (the selection based ones) have the icon on the right.


kenoir:
:+1:

theefer:
Nice :+1:

PR #1109, opened 2015-08-19
Moves simple ordering to search bar (use checkbox for binary state)

kenoir:
gifrecord_2015-08-19_155119

### DIFF
--- a/kahuna/public/js/search/query.html
+++ b/kahuna/public/js/search/query.html
@@ -25,6 +25,7 @@
     </button>
 
     <div class="search__filters-container">
+
     <button class="search__filters-toggle"
             type="button"
             ng:click="filtersShown = !filtersShown ">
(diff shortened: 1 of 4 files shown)


kenoir:
@alastairjardine preparation for moving more options to subnav.

alastairjardine:
All good – I like the visual nature of the previous style, but happy for it to go out like this for now!

theefer:
:-1: from me, I find it really confusing to be a checkbox and to be placed in the middle of filtering options. It's also not very future proof as we have discussed allowing to order by last modified too.

Happy to chat and/or be proven wrong.

alastairjardine:
Could we, for kicks, try the previous style buttons up there instead?

kenoir:
@theefer @alastairjardine continuing this work in https://github.com/guardian/grid/pull/1110.

PR #1110, opened 2015-08-19
Better action bar

kenoir:
Continues @jamesgorrie's work in https://github.com/guardian/grid/pull/1087.

gifrecord_2015-08-20_115851

### DIFF
--- a/kahuna/public/js/downloader/downloader.js
+++ b/kahuna/public/js/downloader/downloader.js
@@ -12,7 +12,7 @@ downloader.controller('DownloaderCtrl',
     ctrl.download = () => {
         const zip = new JSZip();
         const imageHttp = url => $http.get(url, { responseType:'arraybuffer' });
-        const imagesAddedToZip = ctrl.images.map(image =>
+        const imagesAddedToZip = Array.from(ctrl.images.values()).map(image =>
             imageHttp(image.data.source.secureUrl)
                 .then(resp => zip.file(imageName(image), resp.data))
         );
@@ -44,7 +44,7 @@ downloader.directive('grDownloader', function() {
         template: `
             <button type="button" ng:click="ctrl.download()">
                 <gr-icon>file_download</gr-icon>
-                Download {{ctrl.images.length}} images
+                Download {{ctrl.images.size}} images
             </button>`
     };
 });
--- a/kahuna/public/js/downloader/downloader.js
+++ b/kahuna/public/js/downloader/downloader.js
@@ -42,9 +42,9 @@ downloader.directive('grDownloader', function() {
             images: '=grImages' // crappy two way binding
         },
         template: `
-            <button type="button" ng:click="ctrl.download()">
+            <button type="button" title="Download images" ng:click="ctrl.download()">
                 <gr-icon>file_download</gr-icon>
-                Download {{ctrl.images.size}} images
+                <span class="icon-label">Download</span>
             </button>`
     };
 });
--- a/kahuna/public/js/search/query.html
+++ b/kahuna/public/js/search/query.html
@@ -24,57 +24,87 @@
         Advanced
     </button>
 
-    <div class="search__filters-container">
+    <div class="search__modifier-container">
+        <button class="search__modifier-toggle"
+                type="button"
+                ng:click="filtersShown = !filtersShown ">
+            <gr-icon class="search__modifier-toggle__icon">filter_list</gr-icon>
+            <span class="search__modifier-toggle__text">Search filters</span>
+        </button>
 
-    <button class="search__filters-toggle"
-            type="button"
-            ng:click="filtersShown = !filtersShown ">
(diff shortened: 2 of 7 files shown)


kenoir:
Pinging @jamesgorrie @alastairjardine.

kenoir:
Work to add archiving is another PR (possibly dependent on https://github.com/guardian/grid/pull/1022).

akash1810:
What're the thoughts around https://github.com/guardian/grid/pull/1087#issuecomment-130708683, that is, the Google Photos method of one toolbar, that changes once in selection mode.

kenoir:
@akash1810 I like it, but I think these changes are a step in that direction. Once we have a toolbar specifically for selection mode we can think about hiding the search bar when in that mode. 

What do you think @alastairjardine?

alastairjardine:
Yep, let's build on this. I don't mind exploring the more google like approach the button, but I worry it's not actually that good :) Something about the clear on the left concerns me. Happy to give it a go though to see how it feels on the grid!

kahuna/public/js/search/query.js
```diff
@@ -24,6 +24,15 @@ query.controller('SearchQueryCtrl',
                  function($scope, $state, $stateParams, onValChange , mediaApi, track) {
 
     var ctrl = this;
+
+    ctrl.ordering = {
+	orderBy: $stateParams.orderBy
```
theefer: TABs? :disappointed:

kahuna/public/js/downloader/downloader.js
```diff
@@ -12,7 +12,7 @@ downloader.controller('DownloaderCtrl',
     ctrl.download = () => {
         const zip = new JSZip();
         const imageHttp = url => $http.get(url, { responseType:'arraybuffer' });
-        const imagesAddedToZip = ctrl.images.map(image =>
+        const imagesAddedToZip = Array.from(ctrl.images.values()).map(image =>
```
theefer: Why this change?

kahuna/public/js/search/query.html
```diff
@@ -55,6 +56,17 @@
             </li>
 
             <li class="search__filter-item">
+                <label class="search__filter">
+                    <input type="checkbox"
+                           ng:model="searchQuery.ordering.orderBy"
+                           ng:false-value="undefined"
+                           ng:true-value="'oldest'"
+                           ng:checked="searchQuery.ordering.orderBy === 'oldest'" />
+		    Oldest first
```
theefer: I find it quite confusing to have this alongside the filters :crying_cat_face:

kahuna/public/js/search/query.js
kenoir: Need to setup my editorconfig clearly.

theefer:
I like the idea, but:

1. Why move the order widget to the query filters? Makes it more confusing imho.
2. Should it be flushed to the right, to live with the meta panel tab which is also another selection-specific control?

kahuna/public/js/downloader/downloader.js
kenoir: Not sure, this change is from the branch @jamesgorrie started.

kahuna/public/js/search/query.html
kenoir: I think it should be in the search bar, but perhaps it should be obviously distinct. Perhaps some kind of separator between this and the filter options?

theefer: In that case it could be to the right of the date filter, which is also a filter?

I wonder if a dropdown would be clearer than a checkbox, if we want to avoid the space-intensive styled radio we had before?

kahuna/public/js/downloader/downloader.js
theefer: OK. Looks unrelated, but probably worth shipping regardless of this PR, so :+1:

kenoir:
> Why move the order widget to the query filters? Makes it more confusing imho.

Having buttons on the sub nav that "reset" the current search feels wrong and this is "a thing which you do to the search query" rather than a "thing you do to the selection". 

I don't really like the two buttons we had before in this context as they take up a lot of space and have an weird visual style for the top & sub nav.

I agree ordering should be distinct from filters.

> Should it be flushed to the right, to live with the meta panel tab which is also another selection-specific control?

Yeah maybe, I think when I tried it with that it looks strange because of all the separators. We should experiment a bit.

paperboyo:
> I don't really like the two buttons we had before in this context (...)

+1 (or is that -1?) Me neither. If only because there will be more values to sort by (related https://github.com/guardian/grid/pull/989#issuecomment-126271799).

kenoir:
@alastairjardine @theefer using the old buttons (slightly tweaked), now behave responsively like the filters. 

Moved order options to the right (and only show them when in selection mode).

kenoir:
Useful for:
- https://github.com/guardian/grid/pull/1101
- https://github.com/guardian/grid/pull/1022

theefer:
:+1: minus the tab

PR #1112, opened 2015-08-19
[WIP] Initial draft work on service worker support for faster load performance

theefer:
**This is an experiment to showcase what's possible and how easy it is**

The idea will be to serve more assets from the cache, possibly even serving stale while revalidating newer versions in the background.

I'd like to improve our baseline caching story first though, bundling assets together and hashing them and serving them with a far expiry time. That will be a better foundation for further optims using SW.

Also adds "offline" support, although quite useless for this app really. API calls etc are not very gracefully failing (no error message, poor recovery when back online).

### DIFF
--- a/kahuna/app/Global.scala
+++ b/kahuna/app/Global.scala
@@ -5,12 +5,13 @@ import play.api.{Logger, Application, GlobalSettings}
 import play.api.mvc.WithFilters
 import play.filters.gzip.GzipFilter
 
-import lib.{Config, ForceHTTPSFilter}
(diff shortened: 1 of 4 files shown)


jamesgorrie:
>  bundling assets together 
Definitely seems like a more sensible solution as this might be quite hard to maintain.

Not for this PR but could be worth thinking of how we might do that once we move away from having kahuna as a play app, but hosted in S3 (if that's the plan).

theefer:
Needs more work, so closing, but will leave the branch around in case someone finds it useful to implement this in the future.

PR #1114, opened 2015-08-20
ensure errorMessage exists.

akash1810:
from sentry alarm.


theefer:
:+1: 

Uh we have more Sentry errors now, we should take a look at some point soon.

PR #1118, opened 2015-08-20
Allow editing copyright field

theefer:
We should allow people to edit this field in case it's wrong. (We should allow editing more/all fields but that's just one more easy step.)

I decided not to show the input field on the upload page if there is no initial copyright metadata (i.e. you can edit existing but not add missing). This is because we don't make use of copyright anywhere so not worth confusing people at upload time by prompting for unnecessary information. I also noticed even agencies rarely fill in the copyright field, making it more of a nice-to-have field.

Copyright field can be added from the image view or search meta panel.

image

screenshot-2015-08-20t14 55 16

### DIFF
--- a/kahuna/public/js/components/gr-panel/gr-panel.html
+++ b/kahuna/public/js/components/gr-panel/gr-panel.html
@@ -238,6 +238,37 @@
                         </span>
                     </span>
                 </dd>
+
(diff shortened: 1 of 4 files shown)


akash1810:
:+1: 

On the Upload page, should the field be required if it pre-exists i.e. you shouldn't be able to delete it?

theefer:
Why would we make it required?

akash1810:
> Why would we make it required?

Because its been provided previously...?

theefer:
What if it was wrong and the user wants to remove it? Seems weird to force them to invent spurious information to replace it. The byline is the not mandatory and can be removed for instance.

PR #1119, opened 2015-08-20
Fix close button position, use clear icon

theefer:
# Before

image

# After

screenshot-2015-08-20t16 33 48


akash1810:
:+1:

alastairjardine:
Great. Just checking a usability detail: Is the square-area clickable or only the ✖ detail?

theefer:
It's larger than the X, thought not quite the whole area.

Actually here was easy change so I just fixed it to be the full area, see dark outline:

image

alastairjardine:
Perfect @theefer

PR #1120, opened 2015-08-20
inline ure form on upload page.

akash1810:
inline-ure

### DIFF
--- a/kahuna/public/js/edits/image-editor.js
+++ b/kahuna/public/js/edits/image-editor.js
@@ -13,8 +13,8 @@ export var imageEditor = angular.module('kahuna.edits.imageEditor', [
 ]);
 
 imageEditor.controller('ImageEditorCtrl',
-                       ['$scope', '$timeout', 'editsService', 'imageService',
-                        function($scope, $timeout, editsService, imageService) {
+                       ['$scope', '$timeout', 'editsService', 'editsApi', 'imageService',
+                        function($scope, $timeout, editsService, editsApi, imageService) {
 
     var ctrl = this;
 
@@ -56,6 +56,8 @@ imageEditor.controller('ImageEditorCtrl',
     function onSave() {
         return ctrl.image.get().then(newImage => {
             ctrl.image = newImage;
+            ctrl.usageRights = imageService(ctrl.image).usageRights;
+            updateUsageRightsCategory();
             ctrl.status = ctrl.image.data.valid ? 'ready' : 'invalid';
             ctrl.saving = false;
 
@@ -75,6 +77,16 @@ imageEditor.controller('ImageEditorCtrl',
 
 
     ctrl.usageRights = imageService(ctrl.image).usageRights;
+
+    function updateUsageRightsCategory() {
+        let category = ctrl.categories.find(cat => cat.value === ctrl.usageRights.data.category);
+        ctrl.usageRightsCategory = category && category.name;
+        ctrl.showUsageRights = ctrl.usageRightsCategory === undefined;
+    }
+
+    editsApi.getUsageRightsCategories()
+        .then(cats => ctrl.categories = cats)
+        .finally(() => updateUsageRightsCategory());
 }]);
 
 
--- a/kahuna/public/stylesheets/main.css
+++ b/kahuna/public/stylesheets/main.css
@@ -2006,11 +2006,12 @@ ui-archiver .archiver:hover {
 }
 
 .radio-list__label {
-   flex-grow: 1;
-   color: #CCC;
-   background-color: #444;
-   text-align: center;
-   cursor: pointer;
+    flex-grow: 1;
+    color: #CCC;
+    background-color: #444;
+    text-align: center;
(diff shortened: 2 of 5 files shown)


kahuna/public/stylesheets/main.css
```diff
@@ -1181,31 +1181,29 @@ FIXME: what to do with touch devices
...
-.result-editor__labeller-container gr-add-label {
+.result-editor__field-container gr-add-label {
     font-size: 1.3rem;
     padding-right: 10px;
 }
 
-.result-editor__label {
+.result-editor__field-label {
     width: 100px;
+    flex-shrink: 0;
```
akash1810: this took far too long to work out! :crying_cat_face:

akash1810:
not complete - css woes...

screen shot 2015-08-20 at 17 43 18

theefer:
Nice start! It feels like it would be nice to show the current value even when the panel is collapsed, otherwise you don't know whether you need to expand it or not without expanding it (we also talked of having it expanded by default if no usage rights present -- that'd be complementary).

Should we try to reuse the layout (rights above metadata) and editing activation (the pen) as on other metadata panels?

image

akash1810:
> It feels like it would be nice to show the current value even when the panel is collapsed

and 

> having it expanded by default if no usage rights present

Agreed! Feels like a separate PR to me.

Would have done the latter as part of this PR, but it took stupidly long to work out the why the css wasn't playing nice :neutral_face: 

> Should we try to reuse the layout (rights above metadata)

I did that, but it looked strange... maybe because I'm used to the other fields being fist on the Uploads.

> and editing activation (the pen) as on other metadata panels

:+1:

theefer:
:+1: Let's chat tomorrow about the next steps ^

akash1810:
Updated now to:

- Auto expand the form when there are no usage rights on an image
- Follow the same design pattern of hover to reveal pencil -> click pencil to enter edit mode

inline-ure

kahuna/public/stylesheets/main.css
```diff
@@ -282,7 +282,7 @@ a:focus {
     width: 100%;
 }
 
-.right {
+.flex-right {
```
akash1810: meep, this shouldn't really be part of this PR. `flex-right` felt like a more appropriate name for this class.

theefer:
Looks like "Usage rights category" doesn't fit. Should we just use Usage rights here (and maybe elsewhere)?

theefer:
Sorry to nitpick but the pen at the far right is very far from any content, could easily be overlooked. What do you think?

akash1810:
> Should we just use Usage rights here

:+1: looks nicer (will fix alignment issues)

screen shot 2015-08-21 at 12 06 15

akash1810:
> could easily be overlooked

By design. My thinking was, if there aren't any usage rights, the form will already be expanded.

If there are initial usage rights, then its what the system thinks they should be - having the edit far away should mean less accidental changes (unsure if this happens IRL). There might also be an argument in that if we keep manually overriding the system usage rights, the system's formula might actually be wrong.

akash1810:
> but the pen at the far right is very far from any content

@alastairjardine had suggested a nice change to the edit pattern on the metadata panels - on hover, the entire row will highlight to better distinguish what field the pencil belongs to. Might see how that looks.

theefer:
Not convinced about making it artificially confusing to edit. If the user wants to take the action, it should be easy to figure out how?

Is the usage rights category text smaller on purpose? Can it be normal text style?

akash1810:
Can't quite get row highlight to work properly across the site - additional PR.

Changes @theefer suggested:

screen shot 2015-08-21 at 13 15 15

The above is the hover state.

kahuna/public/js/edits/image-editor.js
```diff
@@ -75,6 +77,16 @@ imageEditor.controller('ImageEditorCtrl',
 
 
     ctrl.usageRights = imageService(ctrl.image).usageRights;
+
+    function updateUsageRightsCategory() {
+        let category = ctrl.categories.find(cat => cat.value === ctrl.usageRights.data.category);
+        ctrl.usageRightsCategory = category && category.name;
+        ctrl.showUsageRights = ctrl.usageRightsCategory === undefined;
```
theefer: Is that going to expand it every time the user saves `None` as a usage right?

theefer:
:+1: Looks like another step in the right direction. Can you ping us when it's on TEST for one last look?

kahuna/public/js/edits/image-editor.js
akash1810: In theory, yes. However, `None` isn't an option...

screen shot 2015-08-21 at 14 22 39

akash1810:
on TEST now.

theefer:
So once you've set a category you can't unset it?

Seems like a usability issue to me. Can we fix in another branch, or at least discuss (Slack?)?

theefer:
Tried this on TEST, looks great!

PR #1121, opened 2015-08-20
hover state and title on clear query button.

akash1810:
hover-clear-query

### DIFF
--- a/kahuna/public/js/search/query.html
+++ b/kahuna/public/js/search/query.html
@@ -14,7 +14,7 @@
                 title="Clear query"
                 ng:show="searchQuery.filter.query"
                 ng:click="searchQuery.resetQueryAndFocus()">
-            <gr-icon>clear</gr-icon>
+            <gr-icon>cancel</gr-icon>
         </button>
     </span>
 
--- a/kahuna/public/js/search/query.html
+++ b/kahuna/public/js/search/query.html
@@ -11,6 +11,7 @@
         />
         <button class="search-query__icon search-query__clear clear-button"
                 type="button"
+                title="Clear query"
                 ng:show="searchQuery.filter.query"
                 ng:click="searchQuery.resetQueryAndFocus()">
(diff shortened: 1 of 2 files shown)


theefer:
Is the hover state really necessary? The cursor already changes and it looks a bit weird. Also it overlays over the border.

akash1810:
> Is the hover state really necessary? 

Thought it makes it uniform with the rest of the buttons across the site.

theefer:
But it's not really the same button style as elsewhere; the others have a border all around, which makes the background colour less weird than if there is no border. Not a biggie but seems inelegant and superfluous to me in this instance.

akash1810:
> the others have a border all around, which makes the background colour less weird than if there is no border

Agreed. What about treating it similar to batch label buttons (apply/delete from all images):

hover-clear-query

alastairjardine:
The other option, and the one I actually prefer, is to give the control a bit more weight. What about this?
https://www.google.com/design/icons/#ic_highlight_off

Full area still clickable, of course

alastairjardine:
Or https://www.google.com/design/icons/#ic_cancel @akash1810 

The added weightiness would be good here, especially as it's on quite a dark background.

theefer:
As long as we don't introduce inconsistency (currently same clear for clearing search and selection and cancelling an edit etc).

alastairjardine:
The bolder control works well in this context – I don't view it as an inconsistency in this case. Context matters :)

akash1810:
to confirm - no hover state, but different icon?

alastairjardine:
hover state on different icon – make it blue?

akash1810:
updated (obv. you can't screen shot the mouse cursor...):

screen shot 2015-08-21 at 15 15 46

akash1810:
Changed to white as its our main hover colour (see toolbar hover states).

Related https://github.com/guardian/grid/pull/1135.

jamesgorrie:
:+1:

PR #1122, opened 2015-08-20
Fix bug where the delete button disappears after editing usage rights on uploads page.

akash1810:
Steps to reproduce bug:

- Visit `/upload`
- Find an image which you can delete
- Edit the usage rights of that image
- On saving the usage rights, observe the delete button disappear


theefer:
As per discussion on Slack, I was curious what the root cause was that this hack (?) fixes.

theefer:
:+1:

Issue #819, opened 2015-06-12
Investigate Word Stemming of indexed content


kenoir:
Currently `title` and `description` are stemmed using the Snowball stemming algorithm.

This can lead to some misleading search results, e.g.

- Snoop (dogg) / Snooping
- Park / Parking

jamesgorrie:
@mchv How do you guys deal with fields that you need two different analysers on?

jamesgorrie:
Having looked at this a while ago mulit-fields seemed good?
https://www.elastic.co/guide/en/elasticsearch/reference/current/_multi_fields.html

mchv:
@jamesgorrie we currently don't. I think that is really difficult to get it right, and especially having good tuning that works for all your use case with multiple analyser. I think a better way, but may be more iterative, is to use exclusion when you find a case that does not work, see [configuring language   analyzers](https://www.elastic.co/guide/en/elasticsearch/guide/current/configuring-language-analyzers.html)

kenoir:
@mchv that looks like a much simpler solution.

kenoir:
@theefer it'll be where there's a semantic collision between the word senses of stemmed and unstemmed words. The classic snooping --> snoop (searches for "snoop dogg"). I'm not really sure how we'd work around this in the general case.

kenoir:
I propose we use the "S-stemmer" described here: http://www.researchgate.net/publication/220433848_How_effective_is_suffixing

```
IF a word ends in “ies,” but not “eies” or “aies”
THEN “ies” - “v”
IF a word ends in “es,” but not “aes,” “ees,” or “oes”
THEN “es” - “e”
IF a word ends in 3,” but not “us” or ‘3s”
THEN 3” - NULL 
```

Stemming algorithm can be chosen as detailed here: https://www.elastic.co/guide/en/elasticsearch/guide/current/choosing-a-stemmer.html

kenoir:
Currently I think we use the "Porter" stemming algorithm which seems more far reaching in what it chooses to stem.

kenoir:
From http://www.researchgate.net/publication/220433848_How_effective_is_suffixing:

```
Porter (1980) developed a stemming algorithm that
looks for about 60 suffixes, producing word variant conflation
intermediate between a simple singular-plural technique
and Lovins algorithm. The multistep process uses a
successive removal of short suffixes, rather than a single
removal of the longest possible suffix.
```

theefer:
Looks good

[...]

PR #1126, opened 2015-08-21
Adds custom analyzer to index using s_stemmer

kenoir:
See https://github.com/guardian/grid/issues/819

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Settings.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Settings.scala
@@ -6,34 +6,38 @@ import play.api.libs.json.Json
 object IndexSettings {
 
   val englishSStemmer = Json.obj(
+    "type" -> "custom",
     "tokenizer" -> "standard",
     "filter" -> Json.arr(
+      "gu_stopwords",
       "lowercase",
       "s_stemmer"
     )
   )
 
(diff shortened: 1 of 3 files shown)


theefer:
:+1: LGTM, scary part will be the reindex

jamesgorrie:
RE: reindexing - We should potentially (if it doesn't add overhead) be looking at our [own tool](https://github.com/guardian/elasticsearch-remap-tool) for that.

kenoir:
Going to merge this for now as it has no impact apart from when creating a new index.

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Settings.scala
```diff
@@ -0,0 +1,44 @@
...
+  )
+
+  val filter = Json.obj(
+    "s_stemmer" -> Json.obj(
+      "type" -> "stemmer",
+      "language" -> "minimal_english"
+    ),
+    "gu_stopwords" -> Json.obj(
+      "type" -> "stop",
+      "stopwords" -> "_english_"
```
akash1810: Might be missing something, but is this the standard list of stopwords?

kenoir: It is, you might be able to get away with not defining tit.

PR #1128, opened 2015-08-21
Avoid exposing mixpanel token in auth-free index page, return in media-api config instead

theefer:
Had to change some of the client code to handle the fact that the token now comes asynchronously.

See related CF change https://github.com/guardian/grid-infra/pull/116

### DIFF
--- a/kahuna/public/js/mixpanel/mixpanel.js
+++ b/kahuna/public/js/mixpanel/mixpanel.js
@@ -6,17 +6,14 @@ import UAParser from 'ua-parser-js';
 
 export var mp = angular.module('mixpanel', []);
 
-mp.constant('mixpanelEnabled', ['mixpanelToken', function(mixpanelToken) {
-    return angular.isString(mixpanelToken);
-}]);
-
 /**
  * This module is to allow the rest of the app to include mixpanel as a dependency
  * and not have to deal with the global `var`.
  */
-mp.factory('mixpanel', ['$window', 'mixpanelEnabled', function($window, mixpanelEnabled) {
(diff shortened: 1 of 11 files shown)


kahuna/public/js/mixpanel/mixpanel.js
```diff
@@ -47,6 +46,6 @@ mp.factory('mixpanel', ['$window', 'mixpanelEnabled', function($window, mixpanel
     return {
         init: init,
         track: track,
-        isEnabled: () => mixpanelEnabled
+        isEnabled: () => initialised
```
akash1810: we don't seem to need this any more

theefer: It is still used [in one place](https://github.com/guardian/grid/blob/sc-mixpanel-token-in-api/kahuna/public/js/analytics/track.js#L87).

akash1810: ah!

Feels better to have that check in [here](https://github.com/guardian/grid/blob/sc-mixpanel-token-in-api/kahuna/public/js/mixpanel/mixpanel.js#L38) so that the state is contained. Though that's a separate issue.

akash1810:
:+1:

kahuna/public/js/mixpanel/mixpanel.js
theefer: Fair point. After some experimenting I figured it's actually better to keep that state in the analytics service. Have a look at the latest commits and let me know if you think that's better!

akash1810: nice. :+1:

PR #1129, opened 2015-08-21
Use PandaAuth helper in Panda controller

theefer:
Minor tidy up to avoid duplicating error mapping by using existing auth helper.


akash1810:
:+1:

PR #1130, opened 2015-08-21
Validate agency supplier name, other cleanups

theefer:
Don't allow creating agency rights for unknown suppliers.

Other tidy ups


theefer:
Inert so merging.

PR #1131, opened 2015-08-21
[WIP] Break panda out of kahuna and into a new auth service

theefer:
This change basically moves the Panda controller out of kahuna and into a new service called "auth" (`media-auth.$domain`).

This is a further step toward making kahuna a dump HTTP server of static assets, and auth split out into its own micro-service.

Pending:

- [x] Don't [hardcode](https://github.com/guardian/grid/blob/master/kahuna/public/js/main.js#L38) static reauth URI in kahuna, ~~read it from the 419 error response instead (requires work on panda-session and/or pandular)~~ read it from the HTML page where kahuna injects it instead (not as flexible but a lot simpler).
- [x] Corresponding [grid-infra change](https://github.com/guardian/grid-infra/pull/118).
- [ ] Create resources in the TEST and PROD stack and get the new ELBs DNS'd up.


jamesgorrie:
I might be missing a conversation - but should this be tying in with the other tools login?

theefer:
Not sure what you mean, but the swimlaning idea is that each major project manages its own auth (which is shared using the panda cookie as it is now). Additionally there is a new standalone login service to be used by minor projects that don't warrant their own auth service. Does that answer this?

kenoir:
Merged master as some of the port mapping here conflict with usage.

theefer:
Thanks. Need to pick this up again, but it's quite big and not quite as high priority atm..

theefer:
I've brought this up to date with master and fixed the main issues. Next step is to split this and the grid-infra PR into smaller pieces that can be safely released incrementally. Will look into this the week after next.

theefer:
Superseded by https://github.com/guardian/grid/pull/1729 and more PRs

PR #1133, opened 2015-08-24
raise exception if no stack is defined in settings.ini.

akash1810:
the properties generator script appears to select an arbitrary stack otherwise - better to fail early.


theefer:
:+1:

PR #1134, opened 2015-08-24
Add logging to ELK for all services except ES and ftp-watcher

kenoir:
Adds logging config to log to the shared ELK stack for most apps.

We can't do ES right now because it's not a Play app and it also perhaps doesn't make sense to push it's logs off to another ES instance. ftp-watcher (in our implementation) is not an AWS app so needs further work to have the right permissions to put to a kinesis stream.

- Depends: https://github.com/guardian/grid-infra/pull/119

### DIFF
--- a/cropper/app/Global.scala
+++ b/cropper/app/Global.scala
@@ -5,10 +5,16 @@ import play.filters.gzip.GzipFilter
 
 import controllers.{Application => App}
 
+import lib.{LogConfig, Config}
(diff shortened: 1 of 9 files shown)


theefer:
:+1: w00t, amazing work!

ES is more persistent anyway so at least we don't loose the logs all the time.

ftp-watcher -- let's kill it instead :-)

PR #1135, opened 2015-08-24
add label hover is white as its an action.

akash1810:
looks nicer...

add-label-hover

### DIFF
--- a/kahuna/public/js/components/gr-add-label/gr-add-label.html
+++ b/kahuna/public/js/components/gr-add-label/gr-add-label.html
@@ -1,5 +1,4 @@
-<button class="button-shy"
-        ng:class="{'small': ctrl.grSmall}"
+<button ng:class="{'small': ctrl.grSmall}"
         ng:click="ctrl.addLabel()"
         ng:disabled="ctrl.adding"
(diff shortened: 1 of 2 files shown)


kahuna/public/js/components/gr-add-label/gr-add-label.html
```diff
@@ -1,5 +1,4 @@
-<button class="button-shy"
-        ng:class="{'small': ctrl.grSmall}"
+<button ng:class="{'small': ctrl.grSmall}"
```
jamesgorrie: Could we remove this as we don't have a `.small` class.

jamesgorrie:
:+1:

kahuna/public/js/components/gr-add-label/gr-add-label.html
akash1810: This [component](https://github.com/guardian/grid/blob/master/kahuna/public/js/components/gr-add-label/gr-add-label.js#L47) does. Its possible misleading as small in this case means no button text, just the icon...

akash1810:
merging - revising nomenclature for another PR.

PR #1136, opened 2015-08-24
Update photographers

jamesgorrie:
New list from Jo B / Robert H.


common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
```diff
@@ -22,41 +22,85 @@ object PhotographersList {
...
+    "Graham Turner"  -> "The Guardian",
+    "Sean Smith"     -> "The Guardian",
+    "Jill Mead"      -> "The Guardian",
+    "Roger Tooth"    -> "The Guardian",
+    "James Mann"     -> "The Guardian",
+    "Rachel Vere"    -> "The Guardian",
+    "Alicia Canter"  -> "The Guardian",
+    "Bill Code"      -> "The Guardian",
+
+    // Past
```
jamesgorrie: Jo B has assured me that these people shouldn't come in through the FTP stuff, so we should be safe.

The alternative would be to move this list to the Picdar import, but then when things are uploaded, we might not get the benefit of adding this automagically.

:roller_coaster:

jamesgorrie: As chatted with @theefer :kiss: might be the solution here.

common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
```diff
@@ -22,25 +22,72 @@ object PhotographersList {
...
+
+  val contractedPhotographers: Map[String, String] = Map(
+    "Antonio Zazueta"     -> "The Guardian",
+    "Christian Sinibaldi" -> "The Guardian",
     "Christopher Thomond" -> "The Guardian",
-    "Sean Smith"          -> "The Guardian",
     "David Levene"        -> "The Guardian",
-    "David Sillitoe"      -> "The Guardian",
-    "Eamonn Mccabe"       -> "The Guardian",
-    "Felicity Cloake"     -> "The Guardian",
```
theefer: She's gone altogether?

theefer: Note that we have thoughts of her pics

theefer: Just checked with Robert and she's a freelance writer for us (food). He's going to check the contracts.

Either way she doesn't seem to put herself as byline so we couldn't catch it with an automated filter. Instead, we may want to just do a batch override of her rights (where she's the byline through the picdar metadata overrides) and mark her as a contributor photographer. Someone would then have to do the same going forward when they import or use her pics.

theefer:
Merging as Felicity doesn't belong in either staff or contractor list anyway.

theefer:
Going to need to batch `/reindex` images matching these bylines.

common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
jamesgorrie: We definitely have her in the [byline quite a lot](https://api.media.gutools.co.uk/images?q=byline%3A%22Felicity%20Cloake%22).

She is under the "contract photographers" In the Picdar copyright groups, so she will be picked up there.

There is only one that has been uploaded by us, and I suppose she would be flagged up there as a commissioned photographer (pushing the need for a "copy to metadata from usage rights" button)?

PR #1138, opened 2015-08-25
New cost model option

jamesgorrie:
An `newCostModel` flag so we can test the numbers, specifically with the rights-inspector.

### DIFF
--- a/media-api/app/controllers/MediaApi.scala
+++ b/media-api/app/controllers/MediaApi.scala
@@ -313,7 +313,7 @@ case class SearchParams(
   uploadedBy: Option[String],
   labels: List[String],
   hasMetadata: List[String],
-  newCostModel: Boolean
+  costModelDiff: Boolean
 )
 
 object SearchParams {
@@ -350,7 +350,7 @@ object SearchParams {
       request.getQueryString("uploadedBy"),
       commaSep("labels"),
       commaSep("hasMetadata"),
-      request.getQueryString("newCostModel").nonEmpty
+      request.getQueryString("costModelDiff").nonEmpty
     )
   }
 
@@ -375,7 +375,8 @@ object SearchParams {
       "free"              -> searchParams.free.map(_.toString),
       "uploadedBy"        -> searchParams.uploadedBy,
       "labels"            -> listToCommas(searchParams.labels),
-      "hasMetadata"       -> listToCommas(searchParams.hasMetadata)
+      "hasMetadata"       -> listToCommas(searchParams.hasMetadata),
+      "costModelDiff"     -> Some(searchParams.costModelDiff.toString())
     ).foldLeft(Map[String, String]()) {
       case (acc, (key, Some(value))) => acc + (key -> value)
       case (acc, (_,   None))        => acc
--- a/media-api/app/controllers/MediaApi.scala
+++ b/media-api/app/controllers/MediaApi.scala
@@ -313,7 +313,7 @@ case class SearchParams(
   uploadedBy: Option[String],
   labels: List[String],
   hasMetadata: List[String],
-  newCostModel: Option[Boolean]
+  newCostModel: Boolean
 )
 
 object SearchParams {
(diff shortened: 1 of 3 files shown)


media-api/app/lib/elasticsearch/SearchFilters.scala
```diff
@@ -90,4 +79,9 @@ trait SearchFilters extends ImageFields {
     "pool"
   )
 
+  def filterOrFilter(filter: Option[FilterBuilder], orFilter: Option[FilterBuilder]): Option[FilterBuilder] = (filter, orFilter) match {
```
theefer: :+1: I swear there has to be some Scalaz incantation to do this in a clever, succinct, type-agnostic way.

@bmjames would know...

media-api/app/controllers/MediaApi.scala
```diff
@@ -348,7 +349,8 @@ object SearchParams {
       request.getQueryString("free").map(_.toBoolean),
       request.getQueryString("uploadedBy"),
       commaSep("labels"),
-      commaSep("hasMetadata")
+      commaSep("hasMetadata"),
+      request.getQueryString("newCostModel").nonEmpty
```
theefer: Should we map this to `_.toBoolean` (default to `false`)?

jamesgorrie: I'm using it as a flag, as it's off by default (i.e. doesn't have a `false` option). So the QS would be `/search?newCostModel`. Don't really care as it's temporary, but felt like a flag, not an option.

media-api/app/lib/elasticsearch/SearchFilters.scala
jamesgorrie: You could have some implicits potentially:
``` Scala
implicit class FilterBuilderHelpers(filter: Option[FilterBuilder]) {
    def ||(orFilter: Option[FilterBuilder]): Option[FilterBuilder] = (filter, orFilter) match {
      case (Some(someFilter), Some(orSomeFilter)) => Some(filters.or(someFilter, orSomeFilter))
      case (filterOpt,    orFilterOpt)    => filterOpt orElse orFilterOpt
    }
  }
```

and then the implementation looks like:
```freeMetadataFilter || freeUsageRightsFilter```

But that's still got the weird pattern matching code.

I need to look more into [elastic4s](https://github.com/sksamuel/elastic4s)...

**Comment by jamesgorrie** [comment]
I've added it that it now does a diff, I didn't see the need of adding both parameters as that adds complexity if they're both set.

It's hijacking the query slightly, but it's for dev purposes only.

**Comment by jamesgorrie** [comment]
A few points on testing it with the live ES data:

* total => 390221
* credit:"The Guardian" => 205107
* credit:"Guardian" => 2455

From a rudimentary check - these credits seem to be set as overrides.

The solution might be worth that when we reindex, we set the usage rights after applying the overrides (I think we do).

It also shows we should probably put the usage rights at the top of the uploads editor and get that "copy to metadata" button.

**Comment by jamesgorrie** [comment]
> we set the usage rights after applying the overrides

This isn't actually happening with a reindex, I will investigate.

**Comment by jamesgorrie** [comment]
__re:__ "The Guardian" credits - we could just use @theefer's new action tool to potentially query them and `POST` the usage rights on these images - although it would be hard to tell what they're meant to be.

Some examples:
```JSON
[{
"description": "Breaking News Graphic",
"credit": "Guardian",
"byline": "Guardian Graphics",
"keywords": []
},
{
"description": "dfc350cb70b7074daf4c4239a6c85d1c0b7e83a7 modified",
"credit": "The Guardian",
"byline": "Samuel Gibbs",
"keywords": []
}]
```

theefer:
:+1:

media-api/app/lib/elasticsearch/SearchFilters.scala
bmjames: @theefer I can only think of:
```scala
import scalaz.syntax.apply._, scalaz.std.option._

^(filter, orFilter)(filters.or) orElse filter orElse orFilter
```

You might one day want to generalise it to any number of filters:

```scala
List(filter, orFilter, anotherFilter).flatten.reduceLeft(filters.or)
```

PR #1139, opened 2015-08-25
crop image link component.

akash1810:
componentise all the things!

This is the first step towards putting the toolbar of image actions on the uploads page.

Spoiler alert `gr-image-action-bar` component will follow soon...


theefer:
:+1:

PR #1142, opened 2015-08-25
Add logger config to normalise log location and level

kenoir:
This should resolve the problem of play logs not appearing in logstash.


theefer:
Enterprise Java Magic

magic

theefer:
:+1:

PR #1144, opened 2015-08-25
Creates a crop side panel

tsop14:
Moves 'crop' button from top-bar to the top of the crop side panel. 
Moves 'dimensions' from its own div to below image thumbnail.
To show which image is currently selected, a blue border-bottom appears and dimensions text is whiter. 
Removes 'original' button from top-bar as user can click on the thumbnail to return to original image.
Aligns images to the left (rather center) to avoid it looking as if it displays the placement of the crop inside the original image. 

Now 
screenshot-media local dev-gutools co uk 2015-08-25 14-31-16

Before
screenshot-media test dev-gutools co uk 2015-08-25 10-52-22


jamesgorrie:
Out of interest - why'd we choose the right hand side?

kahuna/public/stylesheets/main.css
```diff
@@ -1265,16 +1265,39 @@ FIXME: what to do with touch devices
     border-left: 1px solid #565656;
     background-color: #444;
 }
+
+.image-details--crop {
+    float: left;
```
jamesgorrie: It'd be good to try and avoid floats as they always seem to create layout issues...

jamesgorrie:
Could we run this on TEST in front of people as the button position could be a little obscure at first.

paperboyo:
Wow left panel! Great! :smiley_cat:

Just thinking out loud...

> looking as if it displays the placement of the crop inside the original image

Why the lighter background then where the crop does not match 5:3? This gives exactly that impression. Maybe it would be even clearer that it's not the case with that lighter background being the same colour as the one under dimensions etc. (darker)?

As a plus, it would be clearer then what to drag as well and what's an entity (since we've lost a full blue frame to underline for tools UI consistency, I presume, which I'm not sure about - kingdom for showing me another draggable underlined button). Active one disappears even despite text styling tweaks... 

Are crops wider than 5:3 aligned to top of that uniformly-sized container?

tsop14:
@paperboyo 

> Are crops wider than 5:3 aligned to top of that uniformly-sized container?

Not currently, to take up as minimum vertical space and avoid a scroll bar for as long as possible. Wider than 5:3 crop shown in screen shot below. 

> Maybe it would be even clearer that it's not the case with that lighter background being the same colour as the one under dimensions etc. (darker)?

Done.

> since we've lost a full blue frame to underline for tools UI consistency, I presume, which I'm not sure about - kingdom for showing me another draggable underlined button

Underline picked as UI similar to selected filter, and I thought it looked less noisy. I didn't realise this wasn't used for draggable elements. Reverted back to full border. 

What do you think?

screenshot-media local dev-gutools co uk 2015-08-25 16-15-23

paperboyo:
> Not currently, to take up as minimum vertical space and avoid a scroll bar for as long as possible. Wider than 5:3 crop shown in screen shot below.

Makes sense!

Thanks for other changes. Normally I'm all for consistency between tools and strict conventions between UI elements, but I think full border makes it much clearer what's selected. Other may disagree.

Not sure about the lighter line still separating image and dimensions - it brakes the entities apart and may suggest there is a lighter strip (e.g. wrongly cropped off agency badge) at the bottom of the image like in the above active example.

But! You must know **all** my contributions here **must** be treated with the grain of salt and better be discussed. Acted upon only if you agree with them! Me an impostor here :stuck_out_tongue_winking_eye:

tsop14:
@jamesgorrie 

> Out of interest - why'd we choose the right left hand side?

I think it flows well to read across: all the image options --> selected image --> its details. 
If the crop goes to the right of the metadata etc. panel there is a scroll bar separating them. If as shown in the image, it separates the image from its details. 

screenshot-media local dev-gutools co uk 2015-08-25 16-58-02

jamesgorrie:
@tsop14 Thanks for the explainer - :+1: to the code.

It would be nice to see if people spot the new location of the button just incase we frustrate a whole bunch of people, especially as we have more users now.

We _could_, as an introduction, make the button stand out a little more and then bring it back inline with what you've done once people know that the button has moved.

akash1810:
> Could we run this on TEST in front of people as the button position could be a little obscure at first.

:+1: 

I'd be in favour of keeping the button in the top bar, with an additional `Add` button next to the `Crops` text (like we do with `Labels`). Though that might prove a bit bloaty...

What's the behaviour when there aren't any crops yet?

kahuna/public/js/components/gr-crop-image/gr-crop-image.html
```diff
@@ -1,7 +1,9 @@
-<a ng:if="ctrl.canBeCropped" ui:sref="crop({imageId: ctrl.image.data.id})">
-    <gr-icon-label gr-icon="crop">Crop</gr-icon-label>
+<a class="image-info__crop-item image-info__crop-item__button"
```
akash1810: I'd be tempted to avoid putting page specific classes in the base component as it makes it a bit harder to reuse elsewhere as the styling carries through, instead I'd put it [here](https://github.com/guardian/grid/pull/1144/files#diff-845918b04f61694e3ed10f91930a51a3R28).

kahuna/public/stylesheets/main.css
```diff
@@ -917,7 +917,7 @@ input.search-query__input {
 
 .results-toolbar-item__metadata-panel-controls {
     line-height: 35px;
-    width: 130px;
+    width: 129.5px;
```
akash1810: :smiley_cat:

tsop14: Haha, to make its border line up with the button on top's border :see_no_evil:

theefer: Are sub-pixel dimensions a thing? Is that only on retina?

Seems a little odd to me too, but maybe easiest is to show us!

tsop14:
@akash1810 

There is a thumbnail of the original image, and its dimensions. 

screenshot-media local dev-gutools co uk 2015-08-27 11-52-10

kahuna/public/js/components/gr-crop-image/gr-crop-image.html
theefer: +1

tsop14:
screenshot-media local dev-gutools co uk 2015-08-27 17-29-20

theefer:
Cute! :+1:

paperboyo:
O! Weight? Why, if I may ask.

paperboyo:
Weight can be useful on Original to indicate compression level/crappiness when taken together with pixel dimensions. I would see it near dimensions in metadata panel, though. Not many users would be familiar with how to interpret it. I may be wrong, though, most probably I haven't though of something.

Also, how the readout behaves when that JPG is 70MB. Or 1,5GB, for that matter. Or wait, Grid dies then, checked that ;-)

theefer:
Karen on Culture (and Ranj) said that it was a useful piece of information for them to decide on pictures to use in the paper. Not entirely sure why but they seemed confident they used it.

We'd also like to be able to show DPI (and maybe cm dimensions...?), but that's in the fileMetadata so needs some work to extract out.

kahuna/public/stylesheets/main.css
```diff
@@ -1477,15 +1526,16 @@ FIXME: what to do with touch devices
...
+    min-width: 100%;
     box-sizing: border-box;
+    background-color: #333;
 }
+
 .image-crop--selected,
 .result--selected,
 .result--selected:hover {
-    border-color: #00adee;
+    border: solid 2px #00adee;
```
akash1810: Are we aware on the effect this `2px` border has on `result--selected`?

border

Pre selection, `result` has a [transparent 5px border](https://github.com/guardian/grid/blob/master/kahuna/public/stylesheets/main.css#L718), decreasing it to 2px makes it look a little jumpy.

paperboyo:
> to decide on pictures to use in the paper

Ah! Clear. No need to look into the metadata. DPI recorded there in huge majority of our JPGs does not matter one iota. Pixel dimensions is enough. 20x20px is 1in wide at 20DPI, but 2in wide at 10DPI. It can also be 0,17cm wide at 300DPI.

I think all our print products are now printed at 300DPI (I may double-check tomorrow). From quality POV, the picture would better be ~300DPI for magazines and ~240DPI for newspaper products (it’s all relative: Imaging sometimes processes images at, let’s say 80DPI, just because there is no bigger original). But these are safe values.

Assuming 300 DPI (PPI, really, Pixels Per Inch) and working out from that an (ideal) print size in centimetres, or whatever crazy royal thing is a thing, having given pixel dimensions, will give users an idea of how big a picture can go, much better than the file size which is worthless given a variance in JPG compression. Indeed, file size is still used by some to assess print size (wrongly), but they are the age of my grandparents where it was all losslessly-compressed TIFFs, so there was no variance. And my grandparents are, sadly, dead now...

It’s true that an image (even a JPG) may hold genuine DPI information (where it was sized specifically), but it not the case with agency-supplied or camera-sourced imagery. And even then, file size would be worthless to asses a good quality print size (given the same JPG compression, an image of 20x20px with an info that it’s 20DPI is exactly the same file size as 20x20px image with an info that it’s 300DPI).

As to where that information would be most useful to display... Hmm. The earlier, the better, I would say, so the metadata panel near the pixel dimensions (which, TBH, should be enough) would be a right place - if I’m correct to assume that the above file size near the Original crop will show up only **after** you’ve done the first crop.

Sorry for the logorrhea, I’ve been slightly gorried... Happy to try to explain in person.

paperboyo:
34a681b6-4ce1-11e5-8d63-3f5546468194

The above Original is, actually, a perfect example of size not being a good indicator of print size. Nobody in their right mind would judge a 547KB picture to be suitable for a print bigger than stamp-sized. But that 3500x2377 px picture can be printed ~A4 at 300DPI and will probably look good enough thanks to well-compressable empty background and (most probably, too high) compression in general. That said, I haven't actually seen that Original :laughing:

theefer:
Awesome, thanks Mat for the detailed write-up, very useful and makes
complete sense!

It seems file size shouldn't be needed, and I'm generally reluctant to add
things unless we can prove they're useful.

We could alternatively hide it away and tell people where it is and check
if they genuinely find it so useful that they go and expand some extra menu
for example.

In this case, I'm inclined to play the political card and add the thing
they want (possibly not, as Alastair would say, the thing they need), to
show good will to a desk we're working hard to migrate. Favours can go a
long way, and trying to do the Right Thing (retrain people to use better
heuristics than the one they're used to to assess print quality) is likely
not going to be simple or perceived as cooperation from our part.

Happy to discuss but let's keep in mind it's one line of code and it
can/will be changed in the future :-)

On Fri, 28 Aug 2015 09:28 Mateusz <notifications@github.com> wrote:

> [image: 34a681b6-4ce1-11e5-8d63-3f5546468194]
> <https://cloud.githubusercontent.com/assets/6032869/9541493/75be5974-4d60-11e5-9653-919e1259baec.png>
>
> The above Original is, actually, a perfect example of size not being a
> good indicator of print size. Nobody in their right mind would judge a
> 547KB picture to be suitable for a print bigger than stamp-sized. But that
> 3500x2377 px picture can be printed ~A4 at 300DPI and will probably look
> good enough thanks to well-compressable empty background and (most
> probably, too high) compression in general. That said, I haven't actually
> seen that Original [image: :laughing:]
>
> —
> <https://github.com/guardian/grid/pull/1144#issuecomment-135674068>.
>

-- 

in England Number 908396

paperboyo:
> show good will

Sure thing!

> retrain people to use better heuristics (...) is likely not going to be simple

True that!

And - it's not that this info is completely useless. Or in totally wrong place. I would definitely repeat it in the metadata panel at some point, though.

kahuna/public/stylesheets/main.css
jamesgorrie: We should probably hve something like `preview--selected` and `crop--selected` to stop side effects.

PR #1148, opened 2015-08-26
Ignore invalid arguments to avoid 500 in API

theefer:
Next step will be to 400 with an error when they are invalid.

### DIFF
--- a/media-api/app/controllers/MediaApi.scala
+++ b/media-api/app/controllers/MediaApi.scala
@@ -389,7 +389,7 @@ object SearchParams {
       "uploadedBy"        -> searchParams.uploadedBy,
       "labels"            -> listToCommas(searchParams.labels),
       "hasMetadata"       -> listToCommas(searchParams.hasMetadata),
-      "costModelDiff"     -> Some(searchParams.costModelDiff.toString())
+      "costModelDiff"     -> Some(searchParams.costModelDiff.toString)
     ).foldLeft(Map[String, String]()) {
       case (acc, (key, Some(value))) => acc + (key -> value)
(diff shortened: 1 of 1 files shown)


kenoir:
:+1:

media-api/app/controllers/MediaApi.scala
```diff
@@ -341,25 +345,25 @@ object SearchParams {
...
       request.getQueryString("missingIdentifier"),
-      request.getQueryString("valid").map(_.toBoolean),
-      request.getQueryString("free").map(_.toBoolean),
+      request.getQueryString("valid") flatMap parseBooleanFromQuery,
+      request.getQueryString("free") flatMap parseBooleanFromQuery,
       request.getQueryString("uploadedBy"),
       commaSep("labels"),
       commaSep("hasMetadata"),
-      request.getQueryString("costModelDiff").nonEmpty
+      request.getQueryString("costModelDiff") flatMap (s => Try(s.toBoolean).toOption) getOrElse false
```
jamesgorrie: Assuming this is now a `Boolean` in the QS e.g. `?costModelDiff=true`?

jamesgorrie: If so - why don't we use the `parseBooleanFromQuery`?

theefer: Thanks, oversight, will add!

jamesgorrie: So it is now a bool val with default as false?

theefer: Yes. That way our query params are uniform

PR #1151, opened 2015-08-26
inlining the usage rights form on the image page.

akash1810:
ure-image-page

Related https://github.com/guardian/grid/pull/1120.


theefer:
Nice! :+1: 

Does the gr-panel need the same treatment? (presumably this doesn't break it?)

akash1810:
> Does the gr-panel need the same treatment?

Yep. Separate PR.

> presumably this doesn't break it?

Nope.

PR #1154, opened 2015-08-26
Delete crops

jamesgorrie:
We can now delete crops.

__Next__:
* Add action to media-api
* Expose in UI

TODO:
- [x] add permission
- [x] erroring
- [x] ~~throw s3 errors~~ TODO in another PR as it affect the whole `S3StorageFile`

### DIFF
--- a/cropper/app/controllers/Application.scala
+++ b/cropper/app/controllers/Application.scala
@@ -1,5 +1,7 @@
 package controllers
 
+import com.gu.mediaservice.lib.auth.PermissionType.PermissionType
+
 import scala.concurrent.Future
 
 import _root_.play.api.data._, Forms._
@@ -11,7 +13,7 @@ import _root_.play.api.Logger
 import _root_.play.api.Play.current
 
 import com.gu.mediaservice.lib.auth
-import com.gu.mediaservice.lib.auth.{PermissionStore, AuthenticatedService, PandaUser, KeyStore}
+import com.gu.mediaservice.lib.auth._
 import com.gu.mediaservice.lib.argo.ArgoHelpers
 import com.gu.mediaservice.lib.argo.model.Link
 import com.gu.mediaservice.model.{Crop, SourceImage, CropSource, Bounds}
@@ -21,6 +23,9 @@ import org.joda.time.DateTime
 import lib.imaging.ExportResult
 
 import lib._
+
+import scalaz.{Validation, ValidationNel}
+import scalaz.syntax.validation._
 
 
 object Application extends Controller with ArgoHelpers {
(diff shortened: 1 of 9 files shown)


jamesgorrie:
Out of progress!

cropper/app/controllers/Application.scala
```diff
@@ -107,6 +110,19 @@ object Application extends Controller with ArgoHelpers {
...
+    validateUserWithPermissions(httpRequest.user, PermissionType.DeleteCrops) map { user =>
+      Crops.deleteCrops(id).map { _ =>
+        Notifications.publish(Json.obj("id" -> id), "delete-image-exports")
+        Accepted
+      } recover {
+        case _ => respondError(BadRequest, "deletion-error", "Could not delete crops")
+      }
+    } recover {
+      case PermissionDeniedError => Future.successful(respondError(Unauthorized, "permission-denied", "You cannot delete crops"))
+    } flatMap (a => a) // <- flattening the futures for the response
```
theefer: Weird, we should not have to do that. Can we not use `flatMap` above?

cropper/app/lib/Permissions.scala
```diff
@@ -0,0 +1,30 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+import com.gu.mediaservice.lib.auth._
+
+
+object PermissionDeniedError extends Throwable("Permission denied")
+
+object Permissions {
+  val permissionStore = new PermissionStore(Config.configBucket, Config.awsCredentials)
+
+  def validateUserWithPermissions(user: Principal, permission: PermissionType.PermissionType)(implicit ec: ExecutionContext): Future[Principal] = {
```
theefer: Wrap long line?

cropper/app/lib/Permissions.scala
```diff
@@ -0,0 +1,30 @@
...
+    val failFuture = Future.failed(PermissionDeniedError)
```
theefer: Could be hoisted outside this function to avoid creating it each time (even if successful)?

theefer: Only works if the object is not stateful though.. I *think* that's safe here..?

cropper/app/lib/Permissions.scala
```diff
@@ -0,0 +1,30 @@
+package lib
+
+import scala.concurrent.{ExecutionContext, Future}
+import com.gu.mediaservice.lib.auth._
+
+
+object PermissionDeniedError extends Throwable("Permission denied")
+
+object Permissions {
```
theefer: Nice to have a helper for that!

theefer:
A few comments but looks good!

cropper/app/controllers/Application.scala
jamesgorrie: Yea, I was battling with the syntax a little. Updated.

cropper/app/lib/Permissions.scala
jamesgorrie: :+1:

cropper/app/lib/Permissions.scala
jamesgorrie: Safe.

cropper/app/lib/Permissions.scala
jamesgorrie: We can hopefully use this to do things like multiple permission etc later... probably move it over to common when / if we need to.

theefer:
Nice :+1:

PR #1156, opened 2015-08-26
Use stream to keep up with changes to a cell's `item`.

akash1810:
The emission of the `image-updated` event means the image has changed and so the cell's `item` is stale.

Converting to stream means we keep up to date with changes.

### DIFF
--- a/kahuna/public/js/components/gu-lazy-table/gu-lazy-table-cell.js
+++ b/kahuna/public/js/components/gu-lazy-table/gu-lazy-table-cell.js
@@ -7,8 +7,8 @@ export var lazyTableCell = angular.module('gu.lazyTableCell', [
 ]);
 
 lazyTableCell.directive('guLazyTableCell',
-                        ['subscribe$',
-                         function(subscribe$) {
+                        ['subscribe$', 'observe$',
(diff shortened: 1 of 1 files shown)


kahuna/public/js/components/gu-lazy-table/gu-lazy-table-cell.js
```diff
@@ -17,8 +17,8 @@ lazyTableCell.directive('guLazyTableCell',
         transclude: true,
         template: '<ng-transclude ng:if="guLazyTableCellVisible"></ng-transclude>',
         link: function(scope, element, attrs, ctrl) {
-            const item = scope.$eval(attrs.guLazyTableCell);
-            const position$ = ctrl.getItemPosition$(item);
+            const item$ = observe$(scope, attrs.guLazyTableCell);
+            const position$ = item$.flatMapLatest((item) => ctrl.getItemPosition$(item));
```
theefer: Note I think this could be simplified to `item$.flatMapLatest(ctrl.getItemPosition$)`

theefer:
Cosmetic suggestion, else :+1:

PR #1158, opened 2015-08-26
remove ability to run scripts from thall

jamesgorrie:
Seems like a very dangerous thing to have lying around.
:boom:


theefer:
I wanted to do that the other day and forgot, nice one :+1:

theefer:
maybe don't release thrall now though

theefer:
let's do that tomorrow morning rather than so late in the day

jamesgorrie:
Thrall looks to have caught up with everything and is healthy - :ship:

PR #1160, opened 2015-08-26
add config bucket to cropper

jamesgorrie:
As we're going to want to have this for deleting images.

### DIFF
--- a/cropper/app/controllers/Application.scala
+++ b/cropper/app/controllers/Application.scala
@@ -1,11 +1,9 @@
 package controllers
 
-import java.net.{URI, URL}
-
 import scala.concurrent.Future
 
 import _root_.play.api.data._, Forms._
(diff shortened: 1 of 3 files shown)


akash1810:
:+1:

cropper/app/controllers/Application.scala
```diff
@@ -30,6 +28,7 @@ object Application extends Controller with ArgoHelpers {
   import Config.{rootUri, loginUriTemplate, kahunaUri}
 
   val keyStore = new KeyStore(Config.keyStoreBucket, Config.awsCredentials)
+  val permissionStore = new PermissionStore(Config.configBucket, Config.awsCredentials)
```
theefer: Does it need to be started?

jamesgorrie: Nope - removed.

PR #1163, opened 2015-08-27
Delete crops action

jamesgorrie:
Reliant on guardian/grid/pull/1154.
Adding the action to delete crops if you can.

I've added this to the media-api - as this is where it seems like it would be more useful / semantic after trying to add it to the UI.

I'll catch the old permissions stuff up once this is out so I can test this works independently.


jamesgorrie:
Wondering if this makes more sense on the media-api image response - thoughts?

common-lib/src/main/scala/com/gu/mediaservice/lib/auth/PermissionsHandler.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  def validateUserWithPermissions(user: Principal, permission: PermissionType.PermissionType)
+                                 (implicit ec: ExecutionContext): Future[Principal] =
+    getPermissionValForUser(permission, user) flatMap {
+      case true  => Future.successful(user)
+      case false => failFuture
+    }
+
+  def getPermissionValForUser(permission: PermissionType.PermissionType, user: Principal): Future[Boolean] = {
+    println(user)
```
theefer: Hello World

common-lib/src/main/scala/com/gu/mediaservice/lib/auth/PermissionsHandler.scala
```diff
@@ -0,0 +1,30 @@
...
+
+
+object PermissionDeniedError extends Throwable("Permission denied")
+
+trait PermissionsHandler {
+  val failFuture = Future.failed(PermissionDeniedError)
+  val permissionStore: PermissionStore
+
+  def validateUserWithPermissions(user: Principal, permission: PermissionType.PermissionType)
+                                 (implicit ec: ExecutionContext): Future[Principal] =
```
theefer: Why return the user?

jamesgorrie: Going on the discussion yesterday with API design - should this return an `Option` of something and `None` if it doesn't match instead of failing the future?

I prefer the the failing future.

jamesgorrie:
Just realised I am going to have to move this back to the cropper so we don't have weird issues of the API not having caught up and thus not showing the action as we hide on export length === 0.

jamesgorrie:
Updated to have the action in cropper.

common-lib/src/main/scala/com/gu/mediaservice/lib/auth/PermissionsHandler.scala
theefer: Yeah I'm not sure it necessarily makes sense to return a failed future. Why not just a `Future[Boolean]` though?

jamesgorrie: There's two different methods here which are trying to express solutions different situations:

__[`validateUserWithPermissions`](https://github.com/guardian/grid/pull/1163/files#diff-3047d82dcca27850c68ce76fae73e54cR12)__
This is when you want to do something once we know the user is validated to do it, e.g. delete crops. This is similarly to our Authenticated thing. [I was playing with the idea of using Scalaz`s Validations](https://github.com/guardian/grid/commit/4df32cd6fbc46d295383c04efd9145623ee3c8ef#diff-f1869cbcbc4bbc5ee4ddb9c12741f6e9R133) -> but felt it overkill where promises can express the same thing.

__[getPermissionValForUser](https://github.com/guardian/grid/pull/1163/files#diff-3047d82dcca27850c68ce76fae73e54cR19)__
When you want the value of the permission.

> Future[Boolean]
Not sure it's massively useful, but then again, neither is the user.

theefer:
Makes sense to me to have the action in the cropper btw.

Issue #573, opened 2015-03-23
Files with leading space aren't uploaded or rejected by ftp-watcher


theefer:
We had a minor incident where files whose filename started with a space couldn't be uploaded (the ftp-watcher would upload an empty string as file content) and were not successfully moved to the failed/ folder either. It is likely a bug in the ftp commands.

paperboyo:
Safe to close now?

kenoir:
Yup.

PR #1164, opened 2015-08-27
S3 Watcher

kenoir:
Intended to replace/supplement ftp-watcher by allowing uploads to s3 bucket.

Depends: https://github.com/guardian/grid-infra/pull/103/files


deploy_s3watcher.sh
```diff
@@ -0,0 +1,24 @@
+#/bin/bash
```
theefer: Not urgent, but would be lovely to have a standardised way to push a build to S3. Would be useful to deploy kahuna as static files, or other things like the wat pages and grid-inspector. I think there may be an sbt plugin that does this but obviously a lot of these things are *not* sbt projects, so would have to be more generic.

Just a thought.

s3watcher/package.json
```diff
@@ -0,0 +1,17 @@
...
+    "async": "~1.4.2",
+    "fs": "0.0.2",
+    "request": "~2.61.0"
+  },
+  "devDependencies": {},
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "BSD-2-Clause"
```
theefer: Our default is typically Apache-2

theefer:
Exciting :-)

deploy_s3watcher.sh
kenoir: I wonder if @sihil has any thoughts on deploying lambda functions & deploying static sites via riffraff.

s3watcher/package.json
kenoir: This is the default generated by npm. Will fix.

s3watcher/S3Watcher.js
```diff
@@ -0,0 +1,118 @@
...
+
+exports.handler = function(event, context) {
+    var srcBucket = event.Records[0].s3.bucket.name;
+    var srcKey = decodeURIComponent(event.Records[0].s3.object.key.replace(/\+/g, " "));
+    var splitSrcKey = srcKey.split("/");
+    var simpleKey = splitSrcKey[(splitSrcKey.length - 1)];
+    var filename = '/tmp/' + simpleKey;
+
+    // TODO: Constants we want set from elsewhere!
+    var baseUrl = "nope";
```
kenoir: @gidsg do we have an established pattern for injecting secrets into Lambda functions? I was thinking perhaps pull and parse a properties file from S3.

s3watcher/S3Watcher.js
```diff
@@ -0,0 +1,31 @@
+var Rx = require('rx');
+
+var Lambda   = require('./lib/Lambda');
+var Transfer = require('./lib/Transfer');
+
+
+process.env['NODE_TLS_REJECT_UNAUTHORIZED'] = '0';
```
theefer: Does this disable TLS verification? Do we want to have that in PROD?

kenoir: Yes & no.

s3watcher/lib/Upload.js
```diff
@@ -0,0 +1,38 @@
...
+            body: data,
+            forever: true
+        };
+
+        const uploadRequest = request.post(options);
+
+        return Rx.Observable.create(function(observer){
+           uploadRequest.on("response", function(res){
+               observer.onNext(res);
+           }.bind(observer));
```
theefer: Why is the extra `function` required here?

s3watcher/lib/Upload.js
```diff
@@ -0,0 +1,38 @@
...
+    buildUpload: function(config, s3Event) {
+         const headers = {
+            'Content-Length': s3Event.size,
+            'Content-Type': 'application/octet-stream',
+            'X-Gu-Media-Key': config.apiKey
+        };
+
+        return  {
+            key: config.apiKey,
+            url: config.baseUrl + "/images?filename=" + s3Event.filename,
```
theefer: Could the filename be passed along as a `params` to the `request` call, to ensure proper encoding of it?

We will need to send the `uploadedBy` information too (folder?).

s3watcher/lib/Upload.js
```diff
@@ -0,0 +1,38 @@
...
+            size: s3Event.size,
+            headers: headers
+        };
+    },
+    postData: function(upload, data) {
+        const options = {
+            url: upload.url,
+            headers: upload.headers,
+            body: data,
+            forever: true
```
theefer: Worth documenting why this is needed?

s3watcher/S3Watcher.js
```diff
@@ -0,0 +1,31 @@
...
+
+exports.handler = function(event, context) {
+
+    const lambda   = Lambda.init(event, context);
+    const transfer = Transfer.init(lambda.s3Event);
+
+    const operation = lambda.config.flatMap(transfer.operation).share();
+
+    operation.subscribeOnError(lambda.fail);
```
theefer: Is this necessary? Should be caught by line 23 and 29?

s3watcher/lib/Transfer.js
```diff
@@ -0,0 +1,58 @@
...
+            console.log("Deleting from ingest bucket.");
+            return S3Helper.deleteS3Object({
+                Bucket: s3Event.bucket,
+                Key: s3Event.key
+            });
+        };
+
+        const fail = function(result) {
+            console.log("Upload failed with status code: " + result.response.statusCode);
+            console.log("Copying to fail bucket.");
```
theefer: Note that in the FTP watcher we tried three times.

s3watcher/lib/Lambda.js
```diff
@@ -0,0 +1,48 @@
...
+            const srcSize = e.object.size;
+            const srcFilename = filenameFromKey(srcKey);
+
+            return {
+                bucket: srcBucket,
+                key: srcKey,
+                filename: srcFilename,
+                size: srcSize
+            };
+        })();
```
theefer: Bit odd to have an IIFE, could be split into an actual function?

s3watcher/lib/Transfer.js
```diff
@@ -0,0 +1,58 @@
+var S3Helper     = require('./S3Helper');
+var Upload       = require('./Upload');
+
+
+module.exports = {
+    init: function(s3Event){
+        const operation = function(config) {
+            console.log("Downloading from ingest bucket.");
+            const upload = Upload.buildUpload(config, s3Event);
```
theefer: Redundant with line 16.

s3watcher/lib/Lambda.js
kenoir: Do you mean it's a bit IIFEy? *rimshot*

s3watcher/S3Watcher.js
```diff
@@ -0,0 +1,31 @@
+var Rx = require('rx');
```
jamesgorrie: These could be `const`s.

akash1810:
Not part of this PR, but would we able to add a filter to the objects we get from s3 to fix https://github.com/guardian/grid/issues/573?

theefer:
I don't think #573 will be an issue when we use this, as it's specifically a bug in the FTPWatcher.

s3watcher/lib/Transfer.js
```diff
@@ -0,0 +1,51 @@
...
+            console.log("Downloading from ingest bucket.");
+
+            const upload = Upload.buildUpload(config, s3Event);
+
+            return S3Helper.getS3Object({
+                Bucket: s3Event.bucket,
+                Key: s3Event.key
+            }).flatMap(function(data){
+                console.log("Uploading to image-loader.");
+
```
theefer: Could put in a TODO hint to try and use the stream API for better performance (with the caveat that it's a big PITA)?

theefer:
Great stuff! :+1: 

Let's prove in TEST and talk to Adrian about getting this properly setup!

PR #1167, opened 2015-08-28
Performance improvement to recover previous optim from .first()

theefer:
We've identified that the performance issue seems to be related to removing the `.first()` optimisation.

This change seems to recover much of the lost performance by avoiding firing when the index doesn't actually change, which is 99% of the time.

Tried locally vs the PROD API and can definitely see the improvement in terms of memory use.


jamesgorrie:
Nice :+1:

akash1810:
:+1:

PR #1168, opened 2015-08-28
Delete crops ui

jamesgorrie:
Waiting on #1163 

Allow users to delete crops.
If they're on the special list - so far, I am.

There is a gotcha / race condition that already exists, if the state is refreshed before the API has heard from thrall, the delete button will not be there yet. This doesn't seem to be the general case on PROD.

delete-cropss


kahuna/public/js/components/gr-delete-crops/gr-delete-crops.js
```diff
@@ -0,0 +1,59 @@
...
+
+        ctrl.active = false;
+
+        mediaCropper.canDeleteCrops(ctrl.image).then(deleteCrops => {
+            const deleteConfirmText = 'DELETE';
+            if (deleteCrops) {
+                activate();
+                ctrl.delete = () => {
+                    const superSure = $window.prompt(
+                        'If ANY of these crops are used on the site, or on any other pubic ' +
```
theefer: :neckbeard:

theefer:
:+1:

PR #1174, opened 2015-09-01
Tooltip on description only to avoid obstructing view

theefer:
Previously the tooltip would often obstruct the Add label button, etc. Now more targeted.


jamesgorrie:
I thought this made it quite hard to target when people were going through loads of them?

theefer:
I'm not entirely sure how much people use this tooltip in the first place. OTOH there have been multiple complaints about it. The better solution would be to have a custom tooltip of our own that stays out of the way, includes more data, etc.

This seems like an easy fix, and we can monitor for feedback if this is preventing people from accessing the tooltip.

jamesgorrie:
:+1: Agreed - I was just making sure we weren't going back on complaints we had had in the past...

PR #1175, opened 2015-09-01
Start using new cost model with Guardian credits

jamesgorrie:
We need to start moving to the new rights model.

Part of the problem while I am going through is that we are promoting people to use the credit/source model to set the cost of the image.

If we don;t switch over we are going to be digging a hole deeper for our selves.  

This PR switches over so that we use the `usageRights` info and the fact that we've marked an image with a Guardian creditas free.

--- 

2015: 535474 of 535812 (99.8%)
2014: 135586  of 136023 (99.6%)
2013: 102453 of 103185 (99.2%)
2012: 96689 of 100713 (96%)


theefer:
Note that presumably a lot should be matched once we ingested the Picdar rights. I'm keen to do this ASAP but waiting for the reindex to finish. Hopefully can start tomorrow.

jamesgorrie:
Ready to go :boom::boom::boom:

jamesgorrie:
Could we have a decision on this code wise so when we get through the philosophy - we can go?

media-api/app/lib/elasticsearch/SearchFilters.scala
```diff
@@ -16,9 +16,40 @@ trait SearchFilters extends ImageFields {
...
+  // We're showing `Conditional` here too as we're considering them potentially
+  // free. We could look into sending over the search query as a cost filter
+  // that could take a comma separated list e.g. `cost=free,conditional`.
+  val freeUsageRightsFilter = freeToUseCategories.toNel.map(filters.terms(usageRightsField("category"), _))
+
+  val freeFilter = filterOrFilter(freeSupplierFilter, freeUsageRightsFilter)
+  val nonFreeFilter = freeFilter.map(filters.not)
+
+  val guardianCreditFilter = guardianCredits.toNel.map(cs => filters.terms(metadataField("credit"), cs))
+  val freeFilterOrGuardianCredits = filterOrFilter(freeFilter, guardianCreditFilter)
```
theefer: Can we make it clear this is a temporary hack?

common-lib/src/main/scala/com/gu/mediaservice/lib/config/UsageRightsConfig.scala
```diff
@@ -57,4 +57,19 @@ object UsageRightsConfig {
     "Getty Images" -> payGettySourceList
   )
 
+  // FIXME: There must be a better way (._.).
+  // Potentially making cost a lookup again?
+  val freeToUseCategories: List[String] = List(
```
theefer: I swear we already had this list somewhere?

theefer:
:+1:

common-lib/src/main/scala/com/gu/mediaservice/lib/config/UsageRightsConfig.scala
jamesgorrie: It was moved here a while ago.

media-api/app/lib/elasticsearch/SearchFilters.scala
jamesgorrie: Yip - done.

theefer:
:+1:

theefer:
:tada:

PR #1176, opened 2015-09-01
Getty as agency when credit is there

jamesgorrie:
Based on the finding of #1175 we would make quite a big saving when matching on these credits.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
@@ -141,22 +141,12 @@ object GettyXmpParser extends ImageProcessor {
 }
 
 object GettyCreditParser extends ImageProcessor {
-  import UsageRightsConfig.payGettySourceList
-
-  def apply(image: Image): Image = {
-    val isPayForGettySource = image.metadata.source.exists(payGettySourceList.contains)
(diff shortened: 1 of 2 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -139,6 +140,25 @@ object GettyParser extends ImageProcessor {
   }
 }
 
+object GettyCreditParser extends ImageProcessor {
+  import UsageRightsConfig.payGettySourceList
+
+  def apply(image: Image): Image = {
+    val isPayForGettySource = image.metadata.source.exists(payGettySourceList.contains)
+    val gettyCredits = List("getty images", "afp/getty images")
+    val isGettyCredit = image.metadata.credit.map(_.toLowerCase).exists(gettyCredits.contains)
```
theefer: Why are we filtering by source here rather than storing the `suppliersCollection` and letting the normal algorithm determine cost?

jamesgorrie: This is for images without the XMP data, so we wouldn't have supplier or the collection at this point.

jamesgorrie: Sorry, misunderstood, it now matches irrespective of source.

theefer:
:+1:

PR #1178, opened 2015-09-01
remove 'Guardian' from free credits

jamesgorrie:
After using batcher there are no more `credit:Guardian`s.
Removing it now will stop it being used in the future.


akash1810:
:+1: 

Will we need to run the batcher again before merging?

jamesgorrie:
There's a few I will do manually

PR #1179, opened 2015-09-02
Cursor to indicate description title hover

jamesgorrie:
[See here for what it looks like](https://css-tricks.com/almanac/properties/c/cursor/).
Not exactly a custom tooltip, but I think that will introduce more trouble that it's worth.


alastairjardine:
Where is this popping up? On the image cell description?

jamesgorrie:
Closing as it's actually stupid...

jamesgorrie:
@alastairjardine Yip - the title is on the description text.

PR #1180, opened 2015-09-03
add missing property to thrall.properties.template

akash1810:


jamesgorrie:
Brilliant :+1:

akash1810:
@kenoir any reason why we're using the properties file here, and in [other places](https://github.com/guardian/grid/commit/1430d4d0834a196d3774cb28843d1414b5af3e5a) we're hardcoded?

kenoir:
@akash181 a reindexing thrall uses a different value.

PR #1183, opened 2015-09-03
Ensure all log messages appear in Kinesis Stream

kenoir:
Trying to resolve the issue where not all logs are recorded to centralised logging service.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/LogConfig.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/LogConfig.scala
@@ -13,8 +13,8 @@ import org.slf4j.{Logger => SLFLogger, LoggerFactory}
 import com.gu.mediaservice.lib.config.CommonPlayAppConfig
 import com.gu.logback.appender.kinesis.KinesisAppender
 
-import play.api.Logger
-import play.api.{Logger => PlayLogger, LoggerLike}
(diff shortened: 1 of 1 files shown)


kenoir:
Tested in CODE. All log messages now appearing!

kenoir:
Ping @theefer.

theefer:
w00t! What was the issue?

Also do all the logs show up in the logfiles on the apps? I noticed there is no more application.log, only stdout.log now?

kenoir:
There are a couple of loggers being used by play (framework & app level). The previous code only directed app logs to kinesis, this captures the root logger and sends all logs.

We do not log to a file appender only to console (however our upstart script directs stdout to `stdout.log` :astonished:).

theefer:
Ah OK. Would it be possible (or sensible) to write to disk as well, do you think?

kenoir:
Possibly. I think log rotation is already taken of for media-service apps by logrotate.d config we previously set up. 

I'd lean towards doing that in another PR as it is another modification to existing behavior.

theefer:
Yep that fine. :+1: 

I'm just thinking of us looking rather stupid if we have issues and Logstash is down and we have 0 logs.

kenoir:
Turns out as logback is configure to push all logs (at debug level to STDOUT) and our upstart script pipes that to stdout.log (and logrotate.d already takes care of these logs) we should not need to make further changes.

PR #1184, opened 2015-09-03
reset ure form to original values when cancelled

akash1810:
GIFs may not illustrate the bug. Steps to reproduce:
- Open the usage rights form for any image
- Make a change but don't save, instead press cancel to close the form
- Open the usage rights form again, and the value of the form is that of the change from above rather than the original.

# Before
ure-reset-on-cancel

# After
ure-reset-on-cancel-after

I've also refactored to remove a few unused functions and move instantiation of controller properties higher.

### DIFF
--- a/kahuna/public/js/usage-rights/usage-rights-editor.js
+++ b/kahuna/public/js/usage-rights/usage-rights-editor.js
@@ -18,6 +18,12 @@ usageRightsEditor.controller(
 
     var ctrl = this;
 
+    ctrl.saving = false;
+    ctrl.saved = false;
+    ctrl.categories = [];
+    ctrl.originalCats = [];
+    ctrl.model = angular.extend({}, ctrl.usageRights.data);
+
     const multiRights = { name: 'Multiple categories', value: '' };
     const noRights = { name: 'None', value: '' };
     const catsWithNoRights = () => [noRights].concat(ctrl.originalCats);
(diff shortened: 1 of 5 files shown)


kahuna/public/js/usage-rights/usage-rights-editor.js
```diff
@@ -187,6 +183,8 @@ usageRightsEditor.controller(
     function uiError(error) {
         ctrl.error = error.body.errorMessage;
     }
+
+    init();
```
jamesgorrie: This means we will be doing a call to the API twice which isn't needed.

We could either memoize the categories in the service, or have two methods here, `fetchCategories` & `setCategory`.

theefer:
Suggested to @akash1810 to make the ure conditional behind an `ng:if` instead of `ng:show`. That should have the benefit of killing the DOM and controller when the dialog is closed, and re-initialised it fully every time it gets opened. That way the dialog doesn't have to be aware of some kind of "reset" when hidden and shown. This is slightly more expensive from a JS/DOM POV, but makes the logic much simpler and less stateful.

jamesgorrie:
@theefer :+1::fireworks:

akash1810:
Updated. `ng:if` is so much nicer!

kahuna/public/js/usage-rights/usage-rights-editor.js
akash1810: > We could either memoize the categories in the service

The service already does this - https://github.com/guardian/grid/blob/master/kahuna/public/js/services/api/edits-api.js#L13

jamesgorrie:
:+1:

PR #1186, opened 2015-09-03
add AAP credit check with usage rights

jamesgorrie:
Noticed we don't match on this while going through the list.


jamesgorrie:
Will reindex post merge.

theefer:
:+1: 

Interesting I hadn't seen AAP as credit before, so either new or I'd missed it!

jamesgorrie:
I think it's relatively new... they've been coming uploaded by `aapimages` since about 05/01/15.

PR #1187, opened 2015-09-03
allow empty publication in usage when contract or commissioned

akash1810:
Allow omission of publication in PUT body for usage rights when contract or commissioned photographer.

Allows the body of the PUT request to be:

```json
{"data":{"category":"contract-photographer","photographer":"foo"}}
```

Meaning we won't be storing empty string for the `None` publication.


theefer:
:+1:

jamesgorrie:
Out of interest - besides the bug, was there any request from users for this?

akash1810:
@jamesgorrie no request; some had noticed the bug though.

theefer:
I'd also argue we should always make reverting actions possible in the UI, as a rule, unless there is a good reason not to.

PR #1188, opened 2015-09-03
File name on upload page

tsop14:
File name appears below the thumbnail. Ellipsis text overflow if long file name. 

screenshot-media local dev-gutools co uk 2015-09-03 17-14-45

screenshot-media local dev-gutools co uk 2015-09-03 17-14-30

### DIFF
--- a/kahuna/public/js/edits/image-editor.html
+++ b/kahuna/public/js/edits/image-editor.html
@@ -73,10 +73,6 @@
                 class="result-editor__archiver"
                 image="ctrl.image"
             ></ui-archiver>
-
-            <div class="result-editor__meta-data">
-                <span>{{ctrl.image.data.uploadInfo.filename}}</span>
-            </div>
         </div>
         <span class="result-editor__save-status-container">
             <span class="result-editor__save-status"
@@ -134,6 +130,10 @@
                     gr:on-cancel="ctrl.showUsageRights = false">
                 </gr-usage-rights-editor>
             </div>
+        </div>
+        <div class="result-editor__field-container" ng:if="ctrl.image.data.uploadInfo.filename">
+            <div class="result-editor__field-label">File name</div>
+            <div>{{ctrl.image.data.uploadInfo.filename}}</div>
         </div>
 
     </div>
--- a/kahuna/public/js/edits/image-editor.html
+++ b/kahuna/public/js/edits/image-editor.html
@@ -87,28 +87,6 @@
     </div>
 
     <div class="result-editor__editor">
-        <ui-required-metadata-editor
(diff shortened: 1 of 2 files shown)


akash1810:
Any reason why you chose to put it under the thumbnail vs. a field on the right? As the filename in the first image is a random string, it looks a bit strange/random at first...

theefer:
I don't have strong opinions on where it goes, the only thing I'll point out is that unlike the rest of the forms on the right, filename is not editable.

kahuna/public/js/edits/image-editor.html
```diff
@@ -73,8 +73,11 @@
                 class="result-editor__archiver"
                 image="ctrl.image"
             ></ui-archiver>
-        </div>
 
+            <div class="result-editor__meta-data">
+                <span>{{ctrl.image.data.uploadInfo.filename}}</span>
```
akash1810: can we add a `title` attribute for the tooltip?

alastairjardine:
Pure opinion, but under the thumbnail's a bit odd. Would be better inlined on the right as a read only field. Consider it a title, I guess.

paperboyo:
So does it mean that filename doesn’t go into a Desc field (when that’s empty) any more?

+1 for tooltip (in general tooltips should display the full thing whenever the thing is truncated for all thingz)



> the only thing I'll point out is that unlike the rest of the forms on the right, filename is not editable.

+1 Purer opinion than the rest: under thumbnail :stuck_out_tongue_winking_eye:

theefer:
I think it should still also go in description if description is empty as apparently sometimes that's a useful bootstrap for the description.

paperboyo:
It currently does exactly that. I was just asking if it’s gonna change seeing the first screenshot (I don’t think it should, as it’s useful proxy).

tsop14:
screenshot-media local dev-gutools co uk 2015-09-03 18-02-58

tsop14:
screenshot-media local dev-gutools co uk 2015-09-03 18-42-00

theefer:
We may be able to find a visual treatment to differentiate it more from other writeable fields (esp Usage Rights, which looks identical but *is* editable).

As a next step/PR I suggest moving Usage rights to the top of the fields, as per @tsop14's second screenshot. Might mitigate the confusion.

theefer:
Ah sorry missed out that Usage rights is already moved. Fine by me!

:+1:

jamesgorrie:
:+1:

akash1810:
> esp Usage Rights, which looks identical but is editable

That didn't even occur to me :flushed:.

:+1:

PR #1189, opened 2015-09-04
Clean null unicode

jamesgorrie:
Realised while going through some data Getty, lo and behold, adds some pretty strange info to the byline and source.

e.g.
* `Ian Waldie\u0000\u0000`
* `Getty Images\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0003\u0000\u0000\u0000\u0005\u0000\u0000\u0000\nï¿½ï¿½ï¿½ï¿½`

Superb, I'll try and find a way of searching for this and reindex on change.


theefer:
0 is the end of string char. We should review why they get included...

jamesgorrie:
Nt sure I know what you mean - will wait till you're in.

jamesgorrie:
Closing due to more investigation needed.

PR #1193, opened 2015-09-04
use bloomberg via getty as indication of coming from getty

jamesgorrie:
There seems to be quite a few uploaded by users - I think once they're downloaded from Getty or Bloomberg which don't contain Getty XMP metadata.


theefer:
:+1:

PR #1194, opened 2015-09-04
Fix bug where ure wouldn't close.

akash1810:
ng:if creates a new scope when the element is restored, use controller property instead.

related - https://github.com/guardian/grid/pull/1184


theefer:
:+1:

PR #1195, opened 2015-09-04
log on receipt of heartbeat.

akash1810:


thrall/app/lib/MessageConsumer.scala
```diff
@@ -1,28 +1,27 @@
...
-import scala.concurrent.{Future, ExecutionContext}
 import scala.concurrent.duration._
-
-import com.amazonaws.services.sqs.AmazonSQSClient
-import com.amazonaws.services.sqs.model.{Message => SQSMessage, DeleteMessageRequest, ReceiveMessageRequest}
-import org.elasticsearch.action.update.UpdateResponse
-import _root_.play.api.libs.json._
-import _root_.play.api.libs.functional.syntax._
-import akka.actor.ActorSystem
+import scala.concurrent.{ExecutionContext, Future}
```
akash1810: *sigh* intellij!

thrall/app/lib/MessageConsumer.scala
```diff
@@ -81,7 +80,7 @@ object MessageConsumer {
     }
 
   def heartbeat(msg: JsValue) = Future {
-    None
+    Logger.info("received heartbeat")
```
akash1810: this is the only change

akash1810:
I'm being stupid, this is not actually needed.

In debugging thrall on TEST I can't see any heartbeat messages being sent - that's the actual problem, rather than receipt of heartbeats.

PR #1196, opened 2015-09-04
Add packer provisioning script for imgops AMI

theefer:
Previously was piggybacking on obsolete frontend AMI definition. This now keeps the service in our own hands.

I've created an AMI with this and tested in TEST. Will upgrade the CF ref in grid-infra next.


jamesgorrie:
:+1:

PR #1198, opened 2015-09-07
Update metadata from usage rights

jamesgorrie:
Adding the endpoint that will allow for an endpoint that converts usage rights into metadata.

The problem is editors aren't updating the usage rights, as they then have to update the user metadata afterwards. This makes UR a first class citizen.

This is trying to make sure that people start focusing more on usage rights than metadata for the cost of an image, but also making it as easy as possible.

For the UI - I suggest we have a "apply to metadata" button at first and show editors that, and if it is used most of the time, we can make it automatic (potentially with a new queue, or even in the same update).


metadata-editor/conf/routes
```diff
@@ -19,6 +19,8 @@ GET     /metadata/:id/usage-rights        controllers.EditsController.getUsageRi
 PUT     /metadata/:id/usage-rights        controllers.EditsController.setUsageRights(id: String)
 DELETE  /metadata/:id/usage-rights        controllers.EditsController.deleteUsageRights(id: String)
 
+POST    /metadata/:id/ur-to-md            controllers.EditsController.applyMetadataFromUsageRights(id: String)
```
theefer: Better name please?

jamesgorrie: `usage-rights-to-metadata`?

metadata-editor/conf/routes
```diff
@@ -1,30 +1,32 @@
...
+PUT     /metadata/:id/metadata                      controllers.EditsController.setMetadata(id: String)
 
-GET     /metadata/:id/usage-rights        controllers.EditsController.getUsageRights(id: String)
-PUT     /metadata/:id/usage-rights        controllers.EditsController.setUsageRights(id: String)
-DELETE  /metadata/:id/usage-rights        controllers.EditsController.deleteUsageRights(id: String)
+GET     /metadata/:id/usage-rights                  controllers.EditsController.getUsageRights(id: String)
+PUT     /metadata/:id/usage-rights                  controllers.EditsController.setUsageRights(id: String)
+DELETE  /metadata/:id/usage-rights                  controllers.EditsController.deleteUsageRights(id: String)
+
+POST    /metadata/:id/usage-rights-to-metadata      controllers.EditsController.applyMetadataFromUsageRights(id: String)
```
theefer: Can we have a verb in the URI to reinforce how Naughty:tm: the endpoint is? `set-metadata-from-usage-rights` or whatnot

jamesgorrie: Updated - don't you think the `POST` is suggestive enough?
What about:
`POST /metadata/:id/metadata/from-usage-rights`?

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
```diff
@@ -128,6 +128,14 @@ class DynamoDB(credentials: AWSCredentials, region: Region, tableName: String) {
         new ValueMap().withMap(":value", valueMapWithNullForEmptyString(value))
     )
 
+  def jsonPatch(id: String, key: String, value: Map[String, String])
```
jamesgorrie: @theefer added this so that we can patch the JSON map.

This will come in handy as we could potentially split the metadata form into single inputs not doing all the merging / validation on the client.

metadata-editor/conf/routes
jamesgorrie: Updated to the above, as it seems to make more sense.

jamesgorrie:
[The no-whatspace diff makes it nicer on viewing the routes](https://github.com/guardian/grid/pull/1198/files?w=1)

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
```diff
@@ -186,4 +195,19 @@ class DynamoDB(credentials: AWSCredentials, region: Region, tableName: String) {
...
+  // As AttributeExpressions are deprecated, Amazon recommends using `UpdateExpression`s
+  // This is essentially a script builder.
+  // http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html#DDB-UpdateItem-request-UpdateExpression
+  def partialMapUpdate(parentKey: String, value: Map[String, String]) = {
+    val valueMap = new ValueMap()
+    // add the value to the map and then return the update expression
+    val expression  = "SET " + value
+      .map { case (k, v) => {
+        valueMap.withString(s":$k", v)
+        s"""$parentKey.$k=:$k""".stripMargin
```
theefer: What is the `stripMargin` for? `"""` needed?

jamesgorrie: Because IntelliJ says so... removed.

metadata-editor/app/controllers/EditsController.scala
```diff
@@ -193,6 +201,16 @@ object EditsController extends Controller with ArgoHelpers {
     edits
   }
 
+  def metadataFromUsageRights(usageRights: UsageRights): Option[ImageMetadata] = {
+    val toImageMetadata: PartialFunction[UsageRights, ImageMetadata] = (ur: UsageRights) => ur match {
+      case u: StaffPhotographer        => ImageMetadata(byline = Some(u.photographer), credit = Some(u.publication))
+      case u: ContractPhotographer     => ImageMetadata(byline = Some(u.photographer), credit = u.publication)
+      case u: CommissionedPhotographer => ImageMetadata(byline = Some(u.photographer), credit = u.publication)
```
theefer: Can we double check (again..) with Jo that we always want the publication as credit (i.e. `... for the ...`) even for commissioned photographers? I know they don't for illustrators (for which we still need to add a category), which they credit as `Illustration: Joe Doh`, but maybe they do for photographers..?

jamesgorrie: Illustrations fall in to the weird world of not being _Photgraphers_. Agreed - we're going to have to create a new category for those.

Photographers: 
Contract & Commissioned = "______ for the Guardian"
Staff: "The Guardian"

So I think the above stands, we might just have to make Composer aware of staff photographers.

jamesgorrie:
Bump on thoughts?

jamesgorrie:
To clarify my thinking on this:
As asked by Jonny
> my thought is might it make more sense to do it the other way around, given that the byline and credit fields are usually provided already

At the moment we infer the usage rights from the original metadata with our [processors](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala).

When a user is updating the image data (data referring to usage rights, archived etc) though, we want to create the mental shift from `credit = Guardian ∴ cost = free` to "it's a staff photographer, so we can use it".

We can then offer the user a way of saying we can infer the metadata from the usage rights if they'd like.

The other option is to infer the usage rights from the metadata (reapplying the [processors](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala) would work some of the time) - but then it we are just doing what we do now, which is `special-metadata-combo = free`.

Given that this could shift the perception of what is free and what isn't and is more related to the Contracts that we have, we should stick to the former.

theefer:
Agreed

jamesgorrie:
Merge conflict fixed.

theefer:
:+1:

PR #1199, opened 2015-09-07
surface 500 and 503 errors to ng:messages.

akash1810:
Further to this, we should be using `ng:messages` more, rather than [`$window.alert`](https://github.com/guardian/grid/blob/master/kahuna/public/js/components/gr-add-label/gr-add-label.js#L15) so that it doesn't block the UI. Though we'd most likely need to give the `ng:message` context so that the message is more meaningful/helpful.

### DIFF
--- a/kahuna/public/js/errors/global.html
+++ b/kahuna/public/js/errors/global.html
@@ -22,4 +22,14 @@
             to reauthenticate.
         </div>
     </div>
+
+    <div class="global-error warning" ng:message="unknown">
+        An unknown error has occurred.
+        If the problem persists please <a href="mailto:thegrid.dev@theguardian.com" class="coloured-link">email the Grid team</a>.
+    </div>
+
+    <div class="global-error error" ng:message="server">
+        A server error has occurred; the Grid may not be fully functional at this time.
+        If the problem persists please <a href="mailto:thegrid.dev@theguardian.com" class="coloured-link">email the Grid team</a>.
+    </div>
 </ng:messages>
--- a/kahuna/public/js/main.js
+++ b/kahuna/public/js/main.js
@@ -151,16 +151,34 @@ kahuna.run(['$rootScope', 'mediaApi',
 
 // Intercept 401s and emit an event
 kahuna.config(['$httpProvider', function($httpProvider) {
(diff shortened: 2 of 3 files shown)


kahuna/public/js/errors/global.html
```diff
@@ -22,4 +22,14 @@
             to reauthenticate.
         </div>
     </div>
+
+    <div class="global-error warning" ng:message="unknown">
```
akash1810: `warning` class as the error code/severity isn't known.

kahuna/public/js/main.js
```diff
@@ -151,16 +151,29 @@ kahuna.run(['$rootScope', 'mediaApi',
...
-kahuna.factory('httpUnauthorisedInterceptor',
+kahuna.factory('httpErrorInterceptor',
                ['$q', '$rootScope', 'httpErrors',
                 function($q, $rootScope, httpErrors) {
     return {
         responseError: function(response) {
-            if (response.status === httpErrors.unauthorised.errorCode) {
-                $rootScope.$emit('events:error:unauthorised');
+            switch (response.status) {
+                case 0:
```
theefer: When does that happen?

akash1810: I found it happened on POST failure - I changed the [shape of the payload](https://github.com/guardian/grid/blob/master/kahuna/public/js/services/label.js#L25) in the add label message in DEV to test this (`.post({data_: labels})` vs. `.post({data: labels})`). This caused a server error which returned 500 in the network tab, but angular always showed it as 0.

theefer: That's slightly worrying, could we investigate some more? Were there any security errors in the console?

akash1810: qq - would it be better to use `$provide.decorator` like [pandular does](https://github.com/guardian/pandular/blob/master/src/heal.js)?

theefer: Not necessarily, the decorator is useful so we can remap the promise (until after we have tried to reauth) but we don't want to do this here, just spy on errors.

theefer:
:+1:

PR #1200, opened 2015-09-07
Refactor selection to use reactive state

theefer:
A new take on rethinking how we manage mutable state in the application.

The main idea is to compose simpler but more powerful primitives: **reactive streams** (each new state being emitted as a new value on the stream) and **immutable data structures** (more efficient and convenient to transform, more performant to compare *if they only contain immutable values*).

Currently we hold a lot of copies of mutable state throughout the app, e.g. image resources in search results and selection where they need to be kept in sync if modified. Mutable state also means that if you have derived state (e.g. the list of labels of all selected images), you have to keep track of when any of the state you depend on changed. This gives rise to patterns like [global notifications/observers](https://github.com/guardian/grid/blob/0390c4279656478de66d7f71f6f05cf4f35e8685/kahuna/public/js/search/results.js#L361-L377) and [fiddly cached state](https://github.com/guardian/grid/blob/0390c4279656478de66d7f71f6f05cf4f35e8685/kahuna/public/js/services/preview-selection.js#L33-L61).

Reactive streams allow to depend on state declaratively, by expressing derived state as a function or (or transformation/mapping of) the state you depend on. Even better, streams support composition (using various operators such as `zip` or `combineLatest`)

The result also ideally follows the principle of [unidirectional flow](http://staltz.com/unidirectional-user-interface-architectures.html) (see also: Flux, CycleJS, Reactive Loop, etc): Controllers should expose actions as a stream, which is applied to the state. State is injected into the view via the controller scope.

The state is now hoisted into the router for each given application state (e.g. search contains results and selection). We should be able to move all the currently [complicated logic](https://github.com/guardian/grid/blob/0390c4279656478de66d7f71f6f05cf4f35e8685/kahuna/public/js/search/results.js#L132-L191) that lives in the controller into pure services and helpers. That opens the door to easier testing etc.

There are a lot more opportunities to transition to a reactive model throughout, in particular around the search results. This is just an initial step.

If you have any questions please come over and ask me!

## TODO

- [x] Remove all mutations of image objects, which cause mutations that can't be tracked sanely in immutable collections. This is the case when editing labels and archived state. Revert to polling instead (which eventually yields a new Image resource) until we can find a simpler way.
- [x] Extract transformations (e.g. given a set of images return the common labels, etc) into helper functions.
- [x] Fix usageRights batch editing


theefer:
Now fully usable and ready for a review :-)

kahuna/.jscsrc
```diff
@@ -2,7 +2,16 @@
...
-    "requireCurlyBraces": true,
+    "requireCurlyBraces": [
+        "if",
+        "else",
+        "for",
+        "while",
+        "do",
+        "try",
+        "catch"
+        // Not needed for case and default
```
akash1810: > // Not needed for case and default

Yay! :+1: 

Need to add `finally` to that list?

kahuna/public/js/components/gr-panel/gr-panel.js
```diff
@@ -60,67 +75,127 @@ grPanel.controller('GrPanel', [
...
+        inject$($scope, selectedUsageCategory$,    ctrl, 'usageCategory');
+
+
+        // FIXME: distinct?
+        const selectedMetadata$ = selectedImagesList$.
+              map(imageList.getMetadata).
+              map(imageList.getSetOfProperties);
+        const rawMetadata$ = selectedMetadata$.map(selectedMetadata => {
+            return selectedMetadata.map((values) => {
+                switch (values.size) {
```
akash1810: Why `switch` and not `if`?

kahuna/.jscsrc
theefer: Thanks, good `catch`! (:boom::tada:)

kahuna/public/js/components/gr-panel/gr-panel.js
theefer: I think I carried that over from the previous version of this code, and I find it reads quite well. Except I noticed a bug in my copy, so thanks :-)

kahuna/public/js/search/results.js
```diff
@@ -136,9 +145,15 @@ results.controller('SearchResultsCtrl', [
...
 
             // imagesAll will be a sparse array of all the results
+            const totalLength = Math.min(images.total, ctrl.maxResults);
             ctrl.imagesAll = [];
             ctrl.imagesAll.length = Math.min(images.total, ctrl.maxResults);
 
+            // TODO: ultimately we want to manage the state in the
+            // results stream exclusively
+            results.clear();
+            results.resize(totalLength);
```
akash1810: why do we resize?

theefer: It's the equivalent of [this line](https://github.com/guardian/grid/pull/1200/files#diff-b8c6d4770b5b04294cc8c88d27ae65d3R150). We make it a sparse array with the total length.

kahuna/public/js/search/index.js
```diff
@@ -68,6 +74,33 @@ search.config(['$stateProvider',
...
+                return listFactory();
+            }],
+            // equivalent of `ctrl.images`
+            compactResults$: ['results', function(results) {
+                return results.items$.
+                    map(items => items.filter(angular.identity)).
+                    shareReplay(1);
+            }],
+            // set of selected images resources
+            selectedImages$: ['selection', 'compactResults$', function(selection, compactResults$) {
```
akash1810: do the selected images live here now then, as opposed to the selection service?

theefer: Yes, the preview-selection service is now [delete](https://github.com/guardian/grid/pull/1200/files#diff-533b4c2674310e7ac52f494543d3e8a0L1). One of the differences is that previously the service was a singleton that lives across the whole lifecycle of the app, and therefore needs to be manually [purged](https://github.com/guardian/grid/pull/1200/files#diff-b8c6d4770b5b04294cc8c88d27ae65d3L422) when leaving the search view that uses it. Now the selection is instantiated as part of the search state. Less global state.

kahuna/public/js/services/image-accessor.js
```diff
@@ -0,0 +1,50 @@
...
+imageAccessor.factory('imageAccessor', function() {
+
+    /* == Extractors ==  (return embedded Resources) */
+
+    function extractUserMetadata(image) {
+        return image.data.userMetadata;
+    }
+
+
+    /* == Readers ==  (return data) */
```
akash1810: minor: why readers over getters?

kahuna/public/js/services/label.js
```diff
@@ -2,18 +2,43 @@ import angular from 'angular';
...
         var labelIndex = existingLabels.findIndex(lbl => lbl.data === label);
-
         if (labelIndex !== -1) {
             return existingLabels[labelIndex]
                 .delete()
-                .then(newLabels => {
-                    image.data.userMetadata.data.labels = newLabels;
-                    $rootScope.$emit('image-updated', image, image);
-                    return image;
+                .then(newLabels => apiPoll(() => untilLabelsEqual(image, newLabels.data)))
```
akash1810: Why are we polling now? The response of POSTing and DELETEing to the labels endpoint returns the new labels already.

theefer: Yes, but the problem is that we want a new instance of the whole image resource to replace it throughout. Otherwise we're just monkey patching the resource object and that means immutable collections holding image resources can't detect the resource has changed (without doing a deep comparison which would be very inefficient).

In the future I'd like to be able to reflect the change on the client-side only, as we used to do, but it should still replace the whole image resource so the change can be detected. Introducing polling is the simplest workaround to do this for now but I agree it's lame.

Does that make sense?

kahuna/public/js/services/image-accessor.js
theefer: That's how we named then in Composer, possibly to make it clearer it's a function and not a method on an object. I don't have a strong preference so happy to rename if others prefer getters?

kahuna/public/js/services/label.js
akash1810: > without doing a deep comparison which would be very inefficient

ATM, each component is watching their part (?) of the image resource only, rather than the whole image e.g. https://github.com/guardian/grid/blob/master/kahuna/public/js/edits/labeller.js#L19.

> I'd like to be able to reflect the change on the client-side only

Don't quite understand what you mean...

jamesgorrie:
Quite a big PR (I don't suppose it could be any other way).
It feels like a lot of the [new](https://github.com/guardian/grid/pull/1200/files#diff-a6009712c9e46fa06c9bf9b4b9821a09R103) [subscriptions](https://github.com/guardian/grid/pull/1200/files#diff-a6009712c9e46fa06c9bf9b4b9821a09R154) (old mutations) could be shifted into their own services instead of living in the panel, making the panel more of a view that understands those services - but that would just make the PR larger (thinking out loud)...

Nice move into finding a single, neat way of dealing with this... :+1:

kahuna/public/js/services/label.js
theefer: We need to avoid mutating image resources in place so the object identity can be used to compare two versions of an image resource (e.g. [image objects in the results collection](https://github.com/guardian/grid/pull/1200/files#diff-04b33f0ec2971bf016d5d80534267132R102)). Basically mixing an immutable model with mutable objects defeats all optimisations and capacity to reason about changes. We should move toward more immutability.

That said we should be able to do this without having to rely on the server to hand us fresh image resource objects. If modifying an image (or sub resources of an image), we should be able to update our internal state ("the state of that sub resource and its parent image resource") without polling the API. The server said the update was successful so we can reflect it onto our in-memory model without doing a read-after-write, which is an anti-pattern.

Happy to talk more about ideas around this.

akash1810:
:+1:

theefer:
> It feels like a lot of the new subscriptions (old mutations) could be shifted into their own services instead of living in the panel

I started on that but it seemed a bit artificial because it's not really reused anywhere else. It could also be moved to the state `resolve` block, as they are just declarative mappings of other streams. I suggest rethinking where such code is best to live as we use this pattern more.

jamesgorrie:
>  I suggest rethinking
:+1::thought_balloon:

PR #1201, opened 2015-09-07
Fix meta-panel button styling.

akash1810:
[Let the meta-panel button grow](http://i.giphy.com/85dCPlhUW5bYk.gif) by itself.

Fixes issue where the "Unlock meta panel" text is positioned incorrectly. Also fixes the responsive view.

Before
screen shot 2015-09-07 at 15 23 59

After
screen shot 2015-09-07 at 15 24 54

Before
screen shot 2015-09-07 at 15 24 15

After
screen shot 2015-09-07 at 15 24 42


theefer:
:+1:

PR #1202, opened 2015-09-07
update persisted formula.

akash1810:
include photographer usage rights categories.

Related to https://github.com/guardian/grid/pull/1027. https://github.com/guardian/grid/pull/1027 won't need to send messages to a new queue now...

TODO:
- [x] add a `persisted` query to the media-api to simplify reaper.

### DIFF
--- a/media-api/app/lib/ImageResponse.scala
+++ b/media-api/app/lib/ImageResponse.scala
@@ -26,7 +26,7 @@ object ImageResponse {
 
   def isPhotographerCategory[T <: UsageRights](usageRights: T) =
     usageRights match {
-      case _:StaffPhotographer | _:ContractPhotographer | _:CommissionedPhotographer => true
+      case _:Photographer => true
       case _ => false
     }
 
--- a/media-api/app/lib/ImageResponse.scala
+++ b/media-api/app/lib/ImageResponse.scala
@@ -24,11 +24,18 @@ object ImageResponse {
 
   def fileMetaDataUri(id: String) = URI.create(s"${Config.rootUri}/images/$id/fileMetadata")
 
+  def isPhotographerCategory[T <: UsageRights](usageRights: T) =
+    usageRights match {
+      case _:StaffPhotographer | _:ContractPhotographer | _:CommissionedPhotographer => true
+      case _ => false
+    }
+
   // TODO: move this as a method of Image (fiddly due to dep on Config)
   def imageIsPersisted(image: Image) = {
     image.identifiers.contains(Config.persistenceIdentifier) ||
       image.exports.nonEmpty ||
-      image.userMetadata.exists(_.archived)
+      image.userMetadata.exists(_.archived) ||
+      isPhotographerCategory(image.usageRights)
   }
 
   def create(id: String, esSource: JsValue, withWritePermission: Boolean,
--- a/media-api/app/lib/elasticsearch/ElasticSearch.scala
+++ b/media-api/app/lib/elasticsearch/ElasticSearch.scala
@@ -2,7 +2,7 @@ package lib.elasticsearch
 
 import java.util.regex.Pattern
 
-import org.elasticsearch.index.query.FilteredQueryBuilder
+import org.elasticsearch.index.query.{FilterBuilder, FilteredQueryBuilder}
 import org.elasticsearch.search.aggregations.bucket.terms.StringTerms
 
 import scala.concurrent.{ExecutionContext, Future}
@@ -91,14 +91,39 @@ object ElasticSearch extends ElasticSearchClient with SearchFilters with ImageFi
     val missingIdentifier= params.missingIdentifier.map(idName => filters.missing(NonEmptyList(identifierField(idName))))
     val uploadedByFilter = params.uploadedBy.map(uploadedBy => filters.terms("uploadedBy", NonEmptyList(uploadedBy)))
 
-    val validityFilter   = params.valid.flatMap(valid => if(valid) validFilter else invalidFilter)
+    val validityFilter: Option[FilterBuilder] = params.valid.flatMap(valid => if(valid) validFilter else invalidFilter)
 
     val costFilter       =
       if (params.costModelDiff) params.free.flatMap(free => if (free) freeDiffFilter else nonFreeDiffFilter)
       else                      params.free.flatMap(free => if (free) freeFilter else nonFreeFilter)
 
+
+    val persistedFilter = params.persisted map {
+      case true => {
+        filters or (
+          filters.bool.must(filters.existsOrMissing("exports", true)),
+          filters.exists(NonEmptyList(identifierField(Config.persistenceIdentifier))),
+          filters.bool.must(filters.term(editsField("archived"), true)),
+          filters.bool.must(filters.term(usageRightsField("category"), "staff-photographer")),
+          filters.bool.must(filters.term(usageRightsField("category"), "contract-photographer")),
+          filters.bool.must(filters.term(usageRightsField("category"), "commissioned-photographer"))
+        )
+      }
+      case false => {
+        filters and (
+            filters.bool.must(filters.existsOrMissing("exports", false)),
+            filters.missing(NonEmptyList(identifierField(Config.persistenceIdentifier))),
+            filters.existsOrMissing(editsField("archived"), false),
+            filters.bool.mustNot(filters.term(usageRightsField("category"), "staff-photographer")),
+            filters.bool.mustNot(filters.term(usageRightsField("category"), "contract-photographer")),
+            filters.bool.mustNot(filters.term(usageRightsField("category"), "commissioned-photographer"))
+        )
+      }
+    }
+
+
     val filterOpt = (
-      metadataFilter.toList ++ labelFilter ++ archivedFilter ++
+      metadataFilter.toList ++ persistedFilter ++ labelFilter ++ archivedFilter ++
       uploadedByFilter ++ idsFilter ++ validityFilter ++ costFilter ++
       hasExports ++ hasIdentifier ++ missingIdentifier ++ dateFilter
     ).toNel.map(filter => filter.list.reduceLeft(filters.and(_, _)))
--- a/media-api/app/lib/elasticsearch/ElasticSearch.scala
+++ b/media-api/app/lib/elasticsearch/ElasticSearch.scala
@@ -20,7 +20,7 @@ import scalaz.NonEmptyList
 
 
 import com.gu.mediaservice.syntax._
-import com.gu.mediaservice.lib.elasticsearch.ElasticSearchClient
+import com.gu.mediaservice.lib.elasticsearch.{ImageFields, ElasticSearchClient}
 import controllers.{AggregateSearchParams, SearchParams}
 import lib.{MediaApiMetrics, Config}
 
--- a/media-api/app/lib/elasticsearch/ElasticSearch.scala
+++ b/media-api/app/lib/elasticsearch/ElasticSearch.scala
@@ -98,32 +98,14 @@ object ElasticSearch extends ElasticSearchClient with SearchFilters with ImageFi
       else                      params.free.flatMap(free => if (free) freeFilter else nonFreeFilter)
 
 
-    val persistedFilter = params.persisted map {
-      case true => {
-        filters or (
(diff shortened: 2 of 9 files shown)


theefer:
Cool! I think we also need to update the ES query that is used to match archived and also the one in thrall to check whether an image can be deleted?

thrall/app/lib/ElasticSearch.scala
```diff
@@ -74,7 +74,10 @@ object ElasticSearch extends ElasticSearchClient {
           missingOrEmptyFilter(s"identifiers.$persistenceIdentifier"),
           boolFilter.should(
             missingOrEmptyFilter("userMetadata.archived"),
-            boolFilter.must(termFilter("userMetadata.archived", false))
+            boolFilter.must(termFilter("userMetadata.archived", false)),
+            boolFilter.mustNot(termFilter("usageRights.category", "staff-photographer")),
+            boolFilter.mustNot(termFilter("usageRights.category", "contract-photographer")),
+            boolFilter.mustNot(termFilter("usageRights.category", "commissioned-photographer"))
```
akash1810: This query now becomes:

```sh
curl -XGET 'http://localhost:9200/_search' -d '
{
    "filter": {
      "and": {
        "filters": [
          {
            "missing": {
              "field": "exports",
              "null_value": true,
              "existence": true
            }
          },
          {
            "missing": {
              "field": "identifiers.picdarUrn",
              "null_value": true,
              "existence": true
            }
          },
          {
            "bool": {
                "should": [
                    { 
                        "missing": {
                            "field": "userMetadata.archived",
                            "null_value": true,
                            "existence": true
                        } 
                    },
                    {
                        "bool": {
                            "must": [
                                { "term": { "userMetadata.archived": false } }
                            ]
                        }
                    }
                ],
                "must_not": {
                  "terms": {
                      "usageRights.category": ["staff-photographer"]
                  }
                },
                "must_not": {
                  "terms": {
                      "usageRights.category": ["contract-photographer"]
                  }
                },
                "must_not": {
                  "terms": {
                      "usageRights.category": ["commissioned-photographer"]
                  }
                }
            }
          }
        ]
      }
    }
}'
```

theefer:
Great! Was there another instance of the formula in the media-api for the archived API param?

akash1810:
> Was there another instance of the formula in the media-api for the archived API param?

Don't think its necessary as we're changing the definition for `persisted` rather than `archived`.

akash1810:
Should probably update the UI to surface persisted though - next PR.

thrall/app/lib/ElasticSearch.scala
```diff
@@ -74,7 +74,10 @@ object ElasticSearch extends ElasticSearchClient {
           missingOrEmptyFilter(s"identifiers.$persistenceIdentifier"),
           boolFilter.should(
             missingOrEmptyFilter("userMetadata.archived"),
-            boolFilter.must(termFilter("userMetadata.archived", false))
+            boolFilter.must(termFilter("userMetadata.archived", false)),
+            boolFilter.mustNot(termFilter("usageRights.category", "staff-photographer")),
```
jamesgorrie: We have [these helpers](https://github.com/guardian/grid/blob/master/media-api/app/lib/elasticsearch/ImageFields.scala) to try mitigate string usage.

jamesgorrie: We could also use the `StaffPhotographer.category` here.

media-api/app/lib/ImageResponse.scala
```diff
@@ -24,11 +24,18 @@ object ImageResponse {
 
   def fileMetaDataUri(id: String) = URI.create(s"${Config.rootUri}/images/$id/fileMetadata")
 
+  def isPhotographerCategory[T <: UsageRights](usageRights: T) =
```
jamesgorrie: [Hopefully going to get a new trait in which would help with this](https://github.com/guardian/grid/pull/1205/files#diff-b5fd3ed3a620cfe97ed5b6a40a586bd3R80).

akash1810: oooh shiny!!!!

jamesgorrie: :+1: it and it's all yours.

thrall/app/lib/ElasticSearch.scala
```diff
@@ -1,29 +1,27 @@
...
 import org.joda.time.DateTime
-import groovy.json.JsonSlurper
-import _root_.play.api.libs.json._
-
-import com.gu.mediaservice.lib.elasticsearch.ElasticSearchClient
-import com.gu.mediaservice.syntax._
 
-import ThrallMetrics._
+import scala.collection.convert.decorateAll._
+import scala.concurrent.{ExecutionContext, Future}
```
akash1810: IDE import optimisation... all I did was import `ImageFields`

media-api/app/lib/ImageResponse.scala
akash1810: updated.

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -91,14 +91,39 @@ object ElasticSearch extends ElasticSearchClient with SearchFilters with ImageFi
...
 
-    val validityFilter   = params.valid.flatMap(valid => if(valid) validFilter else invalidFilter)
+    val validityFilter: Option[FilterBuilder] = params.valid.flatMap(valid => if(valid) validFilter else invalidFilter)
 
     val costFilter       =
       if (params.costModelDiff) params.free.flatMap(free => if (free) freeDiffFilter else nonFreeDiffFilter)
       else                      params.free.flatMap(free => if (free) freeFilter else nonFreeFilter)
 
+
+    val persistedFilter = params.persisted map {
```
jamesgorrie: Could we shift these into the [`SearchFilters`](https://github.com/guardian/grid/blob/master/media-api/app/lib/elasticsearch/SearchFilters.scala)?

media-api/app/lib/elasticsearch/filters.scala
```diff
@@ -36,6 +36,9 @@ object filters {
   def term(field: String, term: String): FilterBuilder =
     termFilter(field, term)
 
+  def term(field: String, value: Boolean): FilterBuilder =
```
jamesgorrie: worth calling it `bool`?

jamesgorrie: Although nonplussed as it's conflicting wither way.

akash1810: > worth calling it bool?

The function of the parameter name?

jamesgorrie: the method...

media-api/app/lib/elasticsearch/ElasticSearch.scala
akash1810: Sure - any particular reason?

jamesgorrie: Just to keep this file more readable / expressive of what we're doing, leaving the actual implementation in the filters file. e.g.
```
case true => true persistedFilter
case false => notPersistedFilter
```
Makes easily readable sense.

media-api/app/lib/elasticsearch/filters.scala
akash1810: Might be better to change the signature of the `term` above to `field: String, term: Object` as the [es library supports it](https://github.com/elastic/elasticsearch/blob/v1.3.4/src/main/java/org/elasticsearch/index/query/FilterBuilders.java#L127).

jamesgorrie: That looks like nasty Java laziness.
Perhaps just stick to what we know the results of (If we do pass a case class - what happens?)

media-api/app/lib/elasticsearch/ElasticSearch.scala
akash1810: > Makes easily readable sense.

It really does!!

jamesgorrie: > Makes easily readable sense.

^ Is not a easily readable, sensible sentence.

akash1810:
can I get a show of thumbs? :smile:

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -91,14 +91,21 @@ object ElasticSearch extends ElasticSearchClient with SearchFilters with ImageFi
...
 
     val costFilter       =
       if (params.costModelDiff) params.free.flatMap(free => if (free) freeDiffFilter else nonFreeDiffFilter)
       else                      params.free.flatMap(free => if (free) freeFilter else nonFreeFilter)
 
+
+    val persistFilter = params.persisted map {
+      case true   => persistedFilter
+      case false  => nonPersistedFilter
+    }
```
theefer: Could reuse the `if`/`else` syntax from above and the rest of the file, but I don't mind particularly.

media-api/app/lib/elasticsearch/SearchFilters.scala
```diff
@@ -82,6 +84,27 @@ trait SearchFilters extends ImageFields {
...
+  val persistedFilter = filters or (
+    filters.bool.must(filters.existsOrMissing("exports", true)),
+    filters.exists(NonEmptyList(identifierField(Config.persistenceIdentifier))),
+    filters.bool.must(filters.boolTerm(editsField("archived"), true)),
+    filters.bool.must(filters.term(usageRightsField("category"), "staff-photographer")),
+    filters.bool.must(filters.term(usageRightsField("category"), "contract-photographer")),
+    filters.bool.must(filters.term(usageRightsField("category"), "commissioned-photographer"))
+  )
+
+  val nonPersistedFilter = filters and (
```
theefer: Can this not just be the negation of the filter above?

theefer:
Minor comments, looks good generally!

media-api/app/lib/elasticsearch/SearchFilters.scala
akash1810: Partially - a negation of the `archived` filter isn't strictly correct though as it doesn't cater for the missing case.

media-api/app/lib/elasticsearch/SearchFilters.scala
```diff
@@ -82,6 +84,17 @@ trait SearchFilters extends ImageFields {
...
+  val persistedFilter = filters.or(
+    filters.bool.must(filters.existsOrMissing("exports", true)),
+    filters.exists(NonEmptyList(identifierField(Config.persistenceIdentifier))),
+    filters.bool.must(filters.boolTerm(editsField("archived"), true)),
+    filters.bool.must(filters.term(usageRightsField("category"), "staff-photographer")),
+    filters.bool.must(filters.term(usageRightsField("category"), "contract-photographer")),
+    filters.bool.must(filters.term(usageRightsField("category"), "commissioned-photographer"))
+  )
+
+  val nonPersistedFilter = filters.not(persistedFilter)
```
akash1810: This does:

```
curl -XGET 'http://localhost:9200/_search' -d '
{
  "query": {
      "bool": {
        "must": {
          "match_all": {
            
          }
        }
      }
    },
    "filter": {
      "bool": {
        "must": [
          {
            "not": {        <= the important bit!
              "filter": {
                "bool": {
                  "should": [
                    {
                      "bool": {
                        "must": {
                          "exists": {
                            "field": "exports"
                          }
                        }
                      }
                    },
                    {
                      "exists": {
                        "field": "identifiers.picdarUrn"
                      }
                    },
                    {
                      "bool": {
                        "must": {
                          "term": {
                            "userMetadata.archived": true
                          }
                        }
                      }
                    },
                    {
                      "bool": {
                        "must": {
                          "term": {
                            "usageRights.category": "staff-photographer"
                          }
                        }
                      }
                    },
                    {
                      "bool": {
                        "must": {
                          "term": {
                            "usageRights.category": "contract-photographer"
                          }
                        }
                      }
                    },
                    {
                      "bool": {
                        "must": {
                          "term": {
                            "usageRights.category": "commissioned-photographer"
                          }
                        }
                      }
                    }
                  ]
                }
              }
            }
          }
        ]
      }
    }
}'
```

akash1810:
will merge tomorrow if no objections

PR #1204, opened 2015-09-08
Manage concurrent update requests to the same Resource using a pool, last one wins

theefer:
This should fix the "saving failed" errors a lot of people are seeing when saving metadata. These errors are caused by multiple saving requests (+ background polling to detect when the saved changes have been written) being issued in close succession. Intermediate requests end up failing because subsequent requests have been written back to media-api already (eventual consistency of multiple PUTs).

Bit fiddlier than I'd hoped, hopefully we can simplify further in the future.

### DIFF
--- a/kahuna/public/js/edits/service.js
+++ b/kahuna/public/js/edits/service.js
@@ -1,4 +1,5 @@
 import angular from 'angular';
+import Rx from 'rx';
 
 import '../services/api/edits-api';
 import '../services/api/media-api';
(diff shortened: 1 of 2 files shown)


jamesgorrie:
A little fiddly, but neat :+1:

theefer:
Also [removed the alert](https://github.com/guardian/grid/commit/1a07996b8b77aebd928b60c80a46d05865e344d2) as we're already showing the save error in the page and it was redundant and spammy.

image

PR #1206, opened 2015-09-08
update byline with correct case of photographer's name

jamesgorrie:
Spotted we weren't correcting the casing on TEST.
Now we do.


akash1810:
:+1:

PR #1207, opened 2015-09-08
afp as getty

jamesgorrie:
Images from AFP don't have Getty XMP data.

Quote from Roger H:
> They are acquired as part of our Getty deal...


theefer:
> Images from AFP don't have Getty XMP data.

Is that always true? I'm pretty sure I saw some with XMP.

theefer:
but :+1:

jamesgorrie:
I was only looking at the diff ones.
I can dig more, but there are definitely some which don't.

theefer:
I assume they are just the older images from the archive that didn't have the XMP then.

PR #1208, opened 2015-09-08
Update dev-template.json

akash1810:
remove unused param - we accept [* as the origin](https://github.com/guardian/grid/blob/master/cloud-formation/dev-template.json#L186) for downloading images in DEV.


theefer:
:+

PR #1209, opened 2015-09-08
use more explicit field paths

jamesgorrie:
We're matching on ALL fields that are called what we call them, causing negation issues and will catch us later.

__TODO:__
- [x] update tests


theefer:
If it works and solves the issue :+1:

jamesgorrie:
Tested on test (and tests updated).

PR #1212, opened 2015-09-09
use autocomplete to update staff photographers

jamesgorrie:
We've been told that people who are staff, but are not official photographers still submit photography, especially editors in the field.

As far as I can tell and have been told this photography belongs to use and should be flagged as such.

To this end, we can make the photographer field in the usage rights box an autocomplete input.

__TODO__
- [ ] Swap data list dependant on publication
- [ ] Stop enter submitting the form

### DIFF
--- a/kahuna/public/js/forms/datalist.js
+++ b/kahuna/public/js/forms/datalist.js
@@ -6,12 +6,13 @@ import '../util/eq';
 export var datalist = angular.module('kahuna.forms.datalist', ['util.eq']);
 
 
-datalist.directive('grDatalist', [function() {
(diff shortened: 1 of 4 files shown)


theefer:
Do we prefer this to a select + "other" option? Intuitively that feels like it'd be faster for known photographers, and less error prone, but slower for people who aren't in the list. Possibly fiddlier to implement too.

I don't particularly want to argue for that other option, just raise it as an alternative.

jamesgorrie:
I was thinking that - thought I would put this out there as an option, although after using it, I suppose the list + other would be more useful in this case.

PR #1213, opened 2015-09-09
Other in staff photographer

jamesgorrie:
We've been told that people who are staff, but are not official photographers still submit photography, especially editors in the field.

From Robert H
> These should be treated as staff photography

other-photographer

### DIFF
--- a/kahuna/public/js/usage-rights/usage-rights-editor.html
+++ b/kahuna/public/js/usage-rights/usage-rights-editor.html
@@ -114,7 +114,7 @@
                              ng:if="property.optionsMap"
                              ng:init="otherValue = false">
                             <select
-                                class="full-width"
+                                class="full-width input-"
                                 name="{{ property.name }}"
                                 ng:model="ctrl.model[property.name]"
                                 ng:required="property.required"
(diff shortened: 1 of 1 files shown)


kahuna/public/js/usage-rights/usage-rights-editor.html
```diff
@@ -110,15 +110,29 @@
...
-                            ng:model="ctrl.model[property.name]"
-                            ng:required="property.required"
-                            ng:options="o for o in ctrl.getOptionsFor(property)">
-                            <option value="">None</option>
-                        </select>
+                        <div class="flex-container"
+                             ng:if="property.optionsMap"
+                             ng:init="otherValue = false">
+                            <select
+                                class="full-width input-"
```
akash1810: Extra `input-` class?

theefer:
Simple enough :+1:

akash1810:
:+1:

theefer:
Random though (please ignore as far as this PR is concerned): later we could add a button to "It was me actually!!1" which would use the logged in user's name from their session.

PR #1214, opened 2015-09-09
Allstar agency

jamesgorrie:
There aren't many, but from the powers that be:
> Allstar are a picture library we have a contract with http://www.allstarpl.com


jamesgorrie:
Tests passing.

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -75,6 +76,16 @@ object AlamyParser extends ImageProcessor {
   }
 }
 
+object AllStarParser extends ImageProcessor {
+  def apply(image: Image): Image = image.metadata.credit match {
+    case Some("Allstar Picture Library") => image.copy(
+      usageRights = Agency("Allstar Picture Library"),
+      metadata = image.metadata.copy(credit = Some("Allstar Picture Library"))
```
theefer: This seems redundant as the `case` was checking that in the first place?

theefer:
Minor comment :+1:

PR #1215, opened 2015-09-09
Upgrade theseus to fix bug, remove occurrence of the bug in the first place

theefer:
New version of Theseus should fix the issue with the `data` property of empty embedded resources.

This PR actually retires the only (?) known place where this was an issue.


jamesgorrie:
:+1:

PR #1216, opened 2015-09-09
Fix method name when deleting selected images

theefer:
Minor typo.

Note that when deleting one selected image, it disappears and then some weird logic kicks in and reveals the image again. That's a bug, but it seems to be already happening on PROD...


theefer:
Actually that bug was only happening in DEV because I didn't have thrall running, I think.

PR #1217, opened 2015-09-09
Quick fix to render changed labels and archived state

theefer:
Nasty, it will get better when we use more reactive streams throughout to model state.


jamesgorrie:
:+1:

PR #1218, opened 2015-09-09
explicitly respond with 400 on bad label POST

akash1810:
if you, intentionally, POST to /labels with a bad payload (e.g. `{"foo":["bar"]}` as opposed to the expected `{"data":["yo"]}`), a dynamo exception is raised on the server and play returns it as a 500 - a 400 is more accurate (and more explicit).

discovered in https://github.com/guardian/grid/pull/1199.

(apologies for the import madness!)

### DIFF
--- a/metadata-editor/app/controllers/EditsController.scala
+++ b/metadata-editor/app/controllers/EditsController.scala
@@ -5,21 +5,17 @@ import java.net.URI
 import java.net.URLDecoder.decode
 
 import com.gu.mediaservice.lib.argo.ArgoHelpers
+import com.gu.mediaservice.lib.argo.model._
+import com.gu.mediaservice.lib.aws.{DynamoDB, NoItemFound}
 import com.gu.mediaservice.model._
+import lib._
+import play.api.data.Forms._
+import play.api.data._
+import play.api.libs.concurrent.Execution.Implicits._
+import play.api.libs.json._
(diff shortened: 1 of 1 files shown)


metadata-editor/app/controllers/EditsController.scala
```diff
@@ -113,7 +109,9 @@ object EditsController extends Controller with ArgoHelpers {
           .setAdd(id, "labels", labels)
           .map(publish(id))
           .map(edits => labelsCollection(id, edits.labels.toSet))
-          .map {case (uri, labels) => respondCollection(labels)}
+          .map {case (uri, labels) => respondCollection(labels)} recover {
+          case _ => BadRequest
```
theefer: Can we be more specific on what we want to recover from? Is there an Amazon exception class we can catch?

Being too broad means we may mistakenly catch unrelated errors and mask them from our logs and attention (e.g. network down, DNS gone, out of memory, disk full, etc).

theefer: (Arguably we should validate the payload before we talk to Dynamo, but OK for a quick fix for now if it's a hassle to do the validation)

akash1810: > Is there an Amazon exception class we can catch?

Yep. Done.

theefer:
:+1:

PR #1219, opened 2015-09-09
Compare label list without order

theefer:
Final fix for the broken label on upload page


theefer:
@jamesgorrie sitting next to me says :+1:

PR #1220, opened 2015-09-09
Correctly map agencies when importing from Picdar

theefer:
Also whitelist "commissioned-agency" and Ronald Grant Archive now.


jamesgorrie:
:+1:

PR #1221, opened 2015-09-09
show other photographer if it is set

jamesgorrie:
We do not set `isOther` on load so if it was set, we just show and empty dropdown.
This makes sure we do.

other-photo

### DIFF
--- a/kahuna/public/js/usage-rights/usage-rights-editor.js
+++ b/kahuna/public/js/usage-rights/usage-rights-editor.js
@@ -134,6 +134,14 @@ usageRightsEditor.controller(
         return property.optionsMap[val];
     };
 
+    ctrl.isOther = property => {
+        const missingVal =
+            !ctrl.getOptionsFor(property)
+                .find(option => option === ctrl.model[property.name]);
+
+        return missingVal;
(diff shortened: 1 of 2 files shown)


kahuna/public/js/usage-rights/usage-rights-editor.js
```diff
@@ -134,6 +134,14 @@ usageRightsEditor.controller(
         return property.optionsMap[val];
     };
 
+    ctrl.isOther = property => {
```
akash1810: minor: would `isOtherOption` be more descriptive?

akash1810:
:+1:

kahuna/public/js/usage-rights/usage-rights-editor.js
jamesgorrie: I've gone with `isOtherValue`?

theefer:
:+1:

kahuna/public/js/usage-rights/usage-rights-editor.js
akash1810: :+1:

PR #1224, opened 2015-09-09
Disable mangling of var names when minifying for better debug

theefer:
Bit of a cop out, but I have spent a fair amount of time digging into why `names` mappings are not correct in our source maps and I could not find a solution. I was able to reproduce with a very simple test case, and it is the minification that gets in the way. I tried both Traceur and Babel, and a variety of code changes, to no avail.

See https://github.com/jspm/jspm-cli/issues/1104 for details.

This change disables the mangling of variable names in the minification stage (to make them shorter and reduce payload size). While it sounds a little dramatic, the bundle file size only goes up from 307KB to 380KB gzipped (1.1MB to 1.5MB not gzipped), which isn't that bad considering it should hopefully get Sentry working again and let us use the debugger tools in TEST/PROD with proper variable names.

### DIFF
--- a/kahuna/package.json
+++ b/kahuna/package.json
@@ -59,7 +59,7 @@
   "scripts": {
     "test": "npm run lint",
     "lint": "jscs public/js/ && jshint public/js --exclude public/js/dist/build.js",
-    "dist": "jspm bundle js/main public/js/dist/build.js -m --inject",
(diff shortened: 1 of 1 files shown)


jamesgorrie:
Worth it :+1:

theefer:
Not entirely sure if it will totally appease Sentry but worth a try.

PR #1225, opened 2015-09-10
validate empty labells

jamesgorrie:
Picked up from a Sentry error.
* Validate the label input.
* I remove the cancel on blur as it cancels when you leave the input, including when you try to click the buttons
* changed naming / structure to match other component

cc: @tsop14 

label-validation

### DIFF
--- a/kahuna/public/js/components/gr-add-label/gr-add-label.html
+++ b/kahuna/public/js/components/gr-add-label/gr-add-label.html
@@ -5,17 +5,21 @@
         title="Add label">
 
     <gr-icon ng:class="{'spin': ctrl.adding}">add_box</gr-icon>
-        <span ng:hide="ctrl.grSmall">
-            Add<span ng:show="ctrl.adding">ing...</span>
-        </span>
+    <span ng:hide="ctrl.grSmall">
+        Add<span ng:show="ctrl.adding">ing...</span>
+    </span>
 </button>
-<form class="gr-add-label__form" ng:if="ctrl.active" ng:submit="ctrl.addLabel()" >
+<form
+    class="gr-add-label__form"
+    ng:if="ctrl.active"
+    ng:submit="ctrl.save()">
     <input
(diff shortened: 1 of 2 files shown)


kahuna/public/js/components/gr-add-label/gr-add-label.html
```diff
@@ -3,19 +3,36 @@
...
-            Add<span ng:show="ctrl.adding">ing...</span>
-        </span>
+    <span ng:hide="ctrl.grSmall">
+        Add<span ng:show="ctrl.adding">ing...</span>
+    </span>
 </button>
-<form class="gr-add-label__form" ng:if="ctrl.active" ng:submit="ctrl.addLabel()" >
-    <input class="gr-add-label__form__input" type='text' ng:model="ctrl.newLabel" gr:auto-focus placeholder="label1, label2, label3…" ng:blur="ctrl.cancel()"></input>
+<form
+    class="gr-add-label__form"
```
theefer: Is this a new syntax for aligning attributes we want to introduce?

theefer:
:+1:

kahuna/public/js/components/gr-add-label/gr-add-label.html
jamesgorrie: I've been doing it like for ages... just realised there is the other way too - the other way doesn't play nice with IntelliJ. How's it with ~Vim~ emacs?

tsop14:
Noted. Thanks. :smile:

kahuna/public/js/components/gr-add-label/gr-add-label.html
theefer: GNU/Emacs:tm: handles both fairly well. I tend to prefer the way we've done it so far for readability reasons, but if it's a pain for all but me to edit those, and the format above works better for all, I don't mind switching to that.

PR #1228, opened 2015-09-10
Upload metadata bug

tsop14:
Allows users to delete required metadata on uploads page (which they might do in the process of editing data). 
Stops input-blur kicking in halfway through editing. 
Ensures copyright is visible and therefore editable even if it has been deleted by a user.

### DIFF
--- a/kahuna/public/js/upload/jobs/required-metadata-editor.js
+++ b/kahuna/public/js/upload/jobs/required-metadata-editor.js
@@ -30,7 +30,7 @@ jobs.controller('RequiredMetadataEditorCtrl',
         var cleanMetadata = {};
         Object.keys(ctrl.metadata).forEach(key => {
             if (ctrl.metadata[key] !== ctrl.saveWhenChangedFrom[key]) {
-                cleanMetadata[key] = ctrl.metadata[key];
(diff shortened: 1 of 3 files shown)


kahuna/public/js/upload/jobs/required-metadata-editor.js
```diff
@@ -30,7 +31,7 @@ jobs.controller('RequiredMetadataEditorCtrl',
         var cleanMetadata = {};
         Object.keys(ctrl.metadata).forEach(key => {
             if (ctrl.metadata[key] !== ctrl.saveWhenChangedFrom[key]) {
-                cleanMetadata[key] = ctrl.metadata[key];
+                cleanMetadata[key] = ctrl.metadata[key] || '';
```
jamesgorrie: I don't understand why this is here? If it's been edited, angular sets it to `''` no?

theefer: Not if the `input` is `required`. If `required`, Angular uses `undefined` as the empty value. I swear we'd fixed this before but evidently not.

theefer:
:+1:

PR #1222, opened 2015-09-09
Illustration

jamesgorrie:
Adding Illustrator as a usage rights category.
Once we have the list we can also add parsers.

- [x] Put illustrator list in once we have it

PR #1229, opened 2015-09-10
add illustrators

jamesgorrie:
Missed this in #1222 for some peculiar reason.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
@@ -114,7 +114,19 @@ object MetadataConfig {
     "Suki Dhanda"         -> "The Observer"
   )
 
-  val contractIllustrators = List("Winshluss", "Robert Crumb", "Steve Bell")
+  val contractIllustrators = List(
+    "Robert G Fresson",
+    "Bill Bragg",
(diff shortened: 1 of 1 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
```diff
@@ -114,7 +114,19 @@ object MetadataConfig {
...
+    "Bill Bragg",
+    "Sebastien Thibault",
+    "Eva Bee",
+    "Nate Kitch",
+    "Joe Magee",
+    "Jasper Rietman",
+    "Matt Kenyon",
+    "Andreij Krauze",
+    "Ellie Foreman-Peck",
+    "Ben Jennings"
```
theefer: Alphabetical order please for general sanity.

theefer:
:+1:

common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
paperboyo: Gee... It's Andrzej Krauze, not Andreij! Of e.g. this less famous one http://grafik.rp.pl/grafika2/817291,9.jpg. I haven't checked spelling on the others, though. What about Observer's, like e.g. Chris Riddell?

PR #1231, opened 2015-09-11
Try and fix Sentry errors

theefer:
Sourcemaps are better now though not 100% reliable still possibly.

### DIFF
--- a/kahuna/public/js/image/controller.js
+++ b/kahuna/public/js/image/controller.js
@@ -194,10 +194,12 @@ image.controller('ImageCtrl', [
         });
 
         const freeImageDeleteFailListener = $rootScope.$on('image-delete-failure', (err, image) => {
-            if (err.body && err.body.errorMessage) {
+            if (err && err.body && err.body.errorMessage) {
                 $window.alert(err.body.errorMessage);
             } else {
-                $window.alert(`Failed to delete image ${image.data.id}`);
+                // Possibly not receiving a proper image object sometimes?
+                const imageId = image && image.data && image.data.id || 'Unknown ID';
+                $window.alert(`Failed to delete image ${imageId}`);
             }
         });
 
--- a/kahuna/public/js/main.js
+++ b/kahuna/public/js/main.js
@@ -125,7 +125,7 @@ kahuna.run(['$log', '$rootScope', 'mediaApi', function($log, $rootScope, mediaAp
                     authAndRedirect(loginUriTemplate);
                 } else {
                     // Couldn't extract a login URI, die noisily
-                    throw error;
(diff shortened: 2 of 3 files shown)


kahuna/public/js/image/controller.js
```diff
@@ -194,10 +194,12 @@ image.controller('ImageCtrl', [
...
 
         const freeImageDeleteFailListener = $rootScope.$on('image-delete-failure', (err, image) => {
-            if (err.body && err.body.errorMessage) {
+            if (err && err.body && err.body.errorMessage) {
                 $window.alert(err.body.errorMessage);
             } else {
-                $window.alert(`Failed to delete image ${image.data.id}`);
+                // Possibly not receiving a proper image object sometimes?
+                const imageId = image && image.data && image.data.id || 'Unknown ID';
+                $window.alert(`Failed to delete image ${imageId}`);
```
theefer: Possibly https://beta.getsentry.com/the-guardian/the-grid/group/81091607/events/3770690434/

kahuna/public/js/usage-rights/usage-rights-editor.js
```diff
@@ -200,7 +200,9 @@ usageRightsEditor.controller(
     }
 
     function uiError(error) {
-        ctrl.error = error.body.errorMessage;
+        // ♫ Very superstitious ♫
+        ctrl.error = error && error.body && error.body.errorMessage ||
+            'Unexpected error';
```
theefer: https://beta.getsentry.com/the-guardian/the-grid/group/81090041/

kahuna/public/js/main.js
```diff
@@ -125,7 +125,7 @@ kahuna.run(['$log', '$rootScope', 'mediaApi', function($log, $rootScope, mediaAp
                     authAndRedirect(loginUriTemplate);
                 } else {
                     // Couldn't extract a login URI, die noisily
-                    throw error;
+                    throw new Error('Failed to redirect to auth, no login URI found');
```
theefer: Unlikely but possibly https://beta.getsentry.com/the-guardian/the-grid/group/80626909/ (?)

akash1810: worth doing `$log.info(error)` too?

theefer: Errors thrown in a Promise chain should be surfaced on the window by Angular anyway, so I think that would be superfluous.

theefer: Note that we normally don't throw errors in Promise chains, we `$q.reject` instead. We should only throw if we want it to be surfaced noisily in our errors.

akash1810:
:+1:

PR #1232, opened 2015-09-11
More aggressively attribute images to Getty based on credit

theefer:
Basically any image that includes "Getty Images" somewhere in the credit line, or that contains "... via Getty" or ".../Getty", will be attribute to Getty.

This is to work around the poor standardisation of metadata in Getty Images, particularly from the archive pre-2014. Particularly helpful to attribute images downloaded from the Getty archive to Getty when manually uploaded.


akash1810:
:+1:

PR #1235, opened 2015-09-11
don't encode slash in query

jamesgorrie:
Boring.
Fixes #1059 

[Related PR](https://github.com/angular-ui/ui-router/issues/1759)

PR #1236, opened 2015-09-11
fix the problem where entering / in the query gets encoded in the url and the input field.

akash1810:


akash1810:
alternative to https://github.com/guardian/grid/pull/1235 @jamesgorrie just beat me to it :smile:

PR #1238, opened 2015-09-11
Replaces panel add label button with gr-add-label

tsop14:
Makes label input in metadata-panel consistent with everywhere else. Hurray. 

labels4

### DIFF
--- a/kahuna/public/js/components/gr-add-label/gr-add-label.js
+++ b/kahuna/public/js/components/gr-add-label/gr-add-label.js
@@ -13,12 +13,8 @@ export var addLabel = angular.module('gr.addLabel', [
 ]);
 
 addLabel.controller('GrAddLabelCtrl', [
-    '$scope',
-    '$window',
-    'labelService',
-    function ($scope,
-              $window,
-              labelService) {
+    '$scope', '$window', 'labelService',
+    function ($scope, $window, labelService) {
 
         let ctrl = this;
 
@@ -38,12 +34,13 @@ addLabel.controller('GrAddLabelCtrl', [
         function save(label, imageArray) {
             ctrl.adding = true;
 
-            labelService.batchAdd(imageArray, label).then(image => {
-                        ctrl.image = image;
-                        reset();
-                    })
-                    .catch(saveFailed)
-                    .finally(() => ctrl.adding = false);
+            labelService.batchAdd(imageArray, label)
+                .then(image => {
+                    ctrl.image = image;
+                    reset();
+                })
+                .catch(saveFailed)
+                .finally(() => ctrl.adding = false);
 
         }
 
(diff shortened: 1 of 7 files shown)


kahuna/public/js/components/gr-add-label/gr-add-label.js
```diff
@@ -12,7 +12,8 @@ export var addLabel = angular.module('gr.addLabel', [
     'gr.autoFocus'
 ]);
 
-addLabel.controller('GrAddLabelCtrl', ['$window', 'labelService',
+addLabel.controller('GrAddLabelCtrl', [
+    '$window', 'labelService',
```
jamesgorrie: `$window` isn't used?

jamesgorrie:
It'd be nice to see if this could work with @theefer new reactive services, but it's so simple, let's get it in :+1:

kahuna/public/js/components/gr-add-label/gr-add-label.js
```diff
@@ -58,7 +61,8 @@ addLabel.directive('grAddLabel', [function () {
         scope: {
             image: '=',
             grSmall: '=?',
-            active: '='
+            active: '=',
+            selected: '='
```
theefer: Could we rename `selected` to `images`? Otherwise the concept of selection is leaking into this generic component, which is unnecessary and confusing.

(We could even reduce the API surface and just have `images`, and when needed with a single image, it could just be passed `images="[ctrl.theImage]"`, but I don't mind too much.)

theefer:
Some minor comments, else :+1:

kahuna/public/js/components/gr-add-label/gr-add-label.js
tsop14: function saveFailed() {
            $window.alert('Something went wrong when saving, please try again!');
        }

Still used here.

jamesgorrie: Sorry, my fault for not reading coccrectly.

kahuna/public/js/components/gr-panel/gr-panel.html
```diff
@@ -269,8 +269,15 @@
             </dl>
         </div>
 
-        <div class="image-info__group batch-label">
-            <div class="metadata-line">Labels</div>
+        <div class="image-info__group">
+            <div class="image-info__heading">
+                <span>Labels</span>
+                <gr-add-label ng:blur="ctrl.cancel"
+                              image="ctrl.image"
```
theefer: Is this still needed?

tsop14: No, removing.

kahuna/public/js/components/gr-add-label/gr-add-label.js
```diff
@@ -21,24 +22,26 @@ addLabel.controller('GrAddLabelCtrl', ['$window', 'labelService',
...
 
-        function save(labels) {
+        function save(label, imageArray) {
             ctrl.adding = true;
 
-            labelService.add(ctrl.image, labels)
+            labelService.batchAdd(imageArray, label)
                 .then(image => {
-                    ctrl.image = image;
+                    ctrl.images = image;
```
theefer: Just to check that `image` here is indeed the list of images?

theefer:
:+1:

PR #1233, opened 2015-09-11
style uploading and recent uploads the same.

akash1810:
Also styling uploading images the same as recently uploaded images for consistency.

style

PR #1239, opened 2015-09-11
recent-uploads component

akash1810:
Decoupling for easier refactoring of upload jobs.

### DIFF
--- a/kahuna/public/js/upload/view.html
+++ b/kahuna/public/js/upload/view.html
@@ -29,20 +29,7 @@
             <a ng:controller="SessionCtrl" class="coloured-link" ui:sref="search.results({uploadedBy: user.email, nonFree: true})">View all your uploads</a>
         </div>
 
-        <p ng:if="! ctrl.myUploads">
-            Loading…
-        </p>
-        <p ng:if="ctrl.myUploads.data.length == 0">
-            You haven’t uploaded anything yet.
-        </p>
-        <ul ng:if="ctrl.myUploads.data.length > 0">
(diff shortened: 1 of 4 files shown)


akash1810:
bah! stupidly depends on https://github.com/guardian/grid/pull/1233 because i didn't branch off master...oops

akash1810:
wip-ing as it breaks uploading

kahuna/public/js/upload/view.html
```diff
@@ -22,27 +22,14 @@
 
     <file-prompt class="section" ng:if="!ctrl.latestJob"></file-prompt>
 
-    <section class="section">
+    <section class="section" ng:if="!ctrl.latestJob">
```
theefer: Is this new?

theefer:
1 comment, else :+1:

kahuna/public/js/upload/view.html
akash1810: Yeah. Its so that the recent uploads isn't displayed when we're currently uploading. It makes the UI a bit simpler/cleaner imo.

theefer: How do you change "mode" of the upload page to reveal the last 50 uploads, having just uploaded images?

akash1810: navigation/page refresh...

theefer: I do wonder whether that kind of modality really makes the UI simpler...?

akash1810: removed this `ng:if` now.

PR #1242, opened 2015-09-14
fix failed upload navigation

akash1810:
To reproduce bug:
- Upload a bad image (e.g. a png)
- Navigate to search
- Navigate back to uploads
- Witness:
  - the bad upload is still there
  - the upload prompt isn't shown
  - drag and drop uploading doesn't work

Now the flow is:
- Upload a bad image
- Navigate to search
- Navigate back to uploads
- Witness:
  - You're able to upload in the usual ways.

You can also delete a failed upload from the upload job.

Before
current

After
after

### DIFF
--- a/kahuna/public/js/upload/manager.js
+++ b/kahuna/public/js/upload/manager.js
@@ -30,7 +30,7 @@ upload.factory('uploadManager',
 
         // once all `jobItems` in a job are complete, remove it
         // TODO: potentially move these to a `completeJobs` `Set`
-        $q.all(promises).then(() => jobs.delete(job));
+        $q.all(promises).finally(() => jobs.delete(job));
     }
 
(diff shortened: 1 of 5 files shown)


kahuna/public/js/upload/manager.js
```diff
@@ -30,7 +30,7 @@ upload.factory('uploadManager',
 
         // once all `jobItems` in a job are complete, remove it
         // TODO: potentially move these to a `completeJobs` `Set`
-        $q.all(promises).then(() => jobs.delete(job));
+        $q.all(promises).finally(() => jobs.delete(job));
```
akash1810: This is the fix for uploading a bad image, navigating around then navigate back to /upload.

theefer:
:+1:

PR #1243, opened 2015-09-14
add guardian witness to the categories list

jamesgorrie:
Adding the GuardianWitness category to the usage rights.

We don't allow you to edit the restrictions if there are defaults. This might not be completely desired, but is the simplest for now in terms of what we surface to the user.

guardian-witness-category

### DIFF
--- a/kahuna/public/js/usage-rights/usage-rights-editor.html
+++ b/kahuna/public/js/usage-rights/usage-rights-editor.html
@@ -49,13 +49,24 @@
                            ng:disabled="property.required" />
                     Restricted
                 </label>
-                <div ng:if="ctrl.isRestricted(property)">
+
+                <div ng:if="ctrl.isRestricted(property)"
+                     ng:switch="ctrl.category.defaultRestrictions === undefined">
+                    <!-- We don't allow you to set the restrictions if there are defaults.
+                    This might not be the behaviour we want in the future, but works for now. -->
                     <textarea
                         class="form-input-text"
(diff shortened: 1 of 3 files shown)


kahuna/public/js/usage-rights/usage-rights-editor.html
```diff
@@ -49,13 +49,24 @@
...
                         name="{{ property.name }}"
                         placeholder="What restrictions apply to this image? e.g. 'Use in relation to the Windsor Triathlon only'"
+                        ng:switch-when="true"
                         ng:model="ctrl.model[property.name]"
                         ng:required="ctrl.isRestricted(property)"></textarea>
+
+                    <textarea
+                        class="form-input-text"
+                        disabled="disabled"
+                        ng:switch-default>{{ctrl.category.defaultRestrictions}}</textarea>
```
theefer: Would a text be less weird/frustrating than a disabled textarea? I don't mind too much, just a thought.

theefer:
:+1:

kahuna/public/js/usage-rights/usage-rights-editor.html
jamesgorrie: Not sure this is better - the `textarea` says, "we've set then, and that's that". I think we can revisit this when we decide exactly what we need to allow for things with default restrictions.

gw-retrcitions

theefer: :+1:

PR #1245, opened 2015-09-14
use proper JSON serialiser and update mapping for illustrator

jamesgorrie:
Complete lack of testing on my behalf - I'll get this out and then think about what to do with it.
:cry:


jamesgorrie:
__TODO__
- [ ] update mappings

akash1810:
:+1:

PR #1203, opened 2015-09-08
Labels input

tsop14:
Changes labels prompt to an input box on:
*  Image page
* Preview page
* Upload page
 
Allows multiple label entry using a comma between values.

screenshot-media local dev-gutools co uk 2015-09-07 18-02-50
labels

PR #1247, opened 2015-09-15
Auto-complete labels

tsop14:
When user starts typing, label suggestions appear as dropdown list
When label list is visible, enter key replaces input with the selected label (user will need to press escape to enter a label which is a substring of an existing label)
When no suggestion list is visible, enter key saves the input value as a label
When user clicks away from input box and then clicks back to input box, label suggestions appear

labels5


theefer:
> When label list is visible, enter key replaces input with the selected label (user will need to press escape to enter a label which is a substring of an existing label)

Is it worth testing with some users to check that this isn't more annoying than useful?

theefer:
Code looks good! :+1: 

Just the question about usability, worth asking @alastairjardine and possibly testing with some people.

akash1810:
Nice! :+1:

akash1810:
What does this mean for inputting multiple labels with a comma? Do we need to worry about the UX for that? i.e. should entering a comma initiate a new suggestion? (will that even work?)

tsop14:
As soon as you input a comma the suggestions go away :cry:  It is trying to match the string including the comma so no results are returned.

> i.e. should entering a comma initiate a new suggestion?

I think it should if people are using the comma to add multiple entries... which they might not be. Need to find out.

akash1810:
>  It is trying to match the string including the comma so no results are returned.

Ah! That makes sense.

:ship:

tsop14:
labels6

paperboyo:
That’s super-useful, if only as a discovery mechanism in absence of labels panel :) Very thorough as well!

>> When label list is visible, enter key replaces input with the selected label (user will need to press escape to enter a label which is a substring of an existing label)

> Is it worth testing with some users to check that this isn't more annoying than useful?

Yeah, not sure... Hmm... When checking how suggestions work currently in the Credit field on Upload page, found a bug - ESC won’t cancel the overwriting of a Credit :-( In general, something is fishy with that suggester there, for a long time (misspelled The Guardian would provide a way to see how much).

> I think it should if people are using the comma to add multiple entries... which they might not be. Need to find out.

:wave: I am, I am! Very useful and only just introduced so not many people may know...

> should entering a comma initiate a new suggestion? (will that even work?)

That would be ideal, obviously (if it would work).

tsop14:
labels8

tsop14:
Label suggestions for each label in input. If suggestions are visible, when user presses enter button the highlighted label (from the list )is added to the input box. If no suggestions are visible, when user presses enter button all label in the input box are saved. 

labels8

alastairjardine:
:top: stuff

theefer:
Brilliant :+1:

paperboyo:
:clap: :fireworks:

alastairjardine:
It's getting there ;)

A few IxD changes, but spoken directly to @tsop14 about them.

* command &/ control enter to save to ensure people don't accidentally add tags
* full width box
* if you're at the end of the box, selecting a term from the dropdown doesn't shift the text to the end, thus making it look like it's not done anything
* ESC to cancel when not in the dropdown state

paperboyo:
> ESC to cancel when not in the dropdown state

Yes! ESC does not escape editing mode, just hides the suggestions, second hit on ESC shows suggestions again :-(

> command &/ control enter to save to ensure people don't accidentally add tags

Hmm. Not going to be easy to teach that. It's easy to remove the label after committing (on all but main page that is). It's much faster to just hit Return/Enter. For all these reasons, I'm against. But, obviously, that's just me. Me, the people ;-)

> full width box

It may be useful to see labels that were already there before entering editing mode (available in all but main page). Currently, they disappear, so the full width box makes sense. But if they would not disappear, I would prefer to see (just part, true) of them. The other solution would be for them to be all listed in the editing box, but then we run into https://github.com/guardian/grid/pull/1203#issuecomment-138566588...

In general, I'm not sure I like how our suggester works. Currently, if you've put `pa` and suggester highlights existing `part`, an immediate commit _from the keyboard_ results in `part`. But! Immediate click on OK button results in just `pa`. That's inconsistent. Both ways have theirs pros and cons, so I'm not sure. I've committed zylions of wrong Credit fields (even more so, because suggester used to highlight a wrong Credit as the first one, I think it may be fixed now). But having to press down arrow to get to the first suggestion constitutes a major pain in the butt as well... Well... Maybe apart from first ESC hiding a suggester, also Backspace should hide it? Last point: should second ESC escape editing mode loosing all labels from editing box, or should it just loose the last, edited, one? Hard to write about it, easier just to swear while clicking :-)

:beer:

paperboyo:
Hmm. Can't reproduce if I go to the Upload page after the fact. But on upload, committing labels with Return does not work - it just hides and shows suggester repeatedly. I had that also on other pages, but whenever I want to find a reproducible procedure, I fail...

PR #1248, opened 2015-09-15
[BLOCKED] Use standard pipeline to produce thumbnail and correct colour profile

theefer:
Port the cropper colour profile work to the thumbnail production, so that thumbnails are faithful to the output we'd produce.

One marginally sad side-effect is that common-lib now depends on imaging libraries. There is a risk to pile more and more things into common and blow it up in size. Let's worry about it later though.

## TODO

- [ ] Ingest a batch in TEST to check for sanity and performance
- [ ] Check behaviour on images with flaky colour profiles which currently trip up the cropper (do we have any in stock we know about?)


theefer:
This is broken by images with an invalid profile, so needs more work before can be released.

PR #1250, opened 2015-09-15
update photographer list

akash1810:
per Jo's request.


jamesgorrie:
I want to run this past RCS first as that's where the list came from.

jamesgorrie:
As mentioned in the email - let's just move the people over to the Observer

akash1810:
> let's just move the people over to the Observer

Done.

jamesgorrie:
:+1:

PR #1251, opened 2015-09-15
Auto update md from ur

jamesgorrie:
When a user sets the usage rights on an image, they will need to re-set the metadata with information that can be inferred from the usage rights i.e. Staff photographer => credit & byline.

This adds a call to the server to set that automatically where we can after setting the usage rights.

__TODO__
- [x] Write PR description
- [x] Comment all the hacks
- [x] What to do if no metadata can be found

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/api/Transformers.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/api/Transformers.scala
@@ -1,8 +1,10 @@
 package com.gu.mediaservice.api
 
-import java.net.URLEncoder
+import java.net.{URI, URLEncoder}
 
+import com.gu.mediaservice.lib.argo.model.{EmbeddedEntity, Action}
 import com.gu.mediaservice.lib.config.Services
+import com.gu.mediaservice.model.{Edits, ImageMetadata}
 import play.api.libs.json._
 
 class Transformers(services: Services) {
@@ -46,6 +48,18 @@ class Transformers(services: Services) {
       )
     }
 
+  def entityUri(id: String, endpoint: String = ""): URI =
+    URI.create(s"$metadataBaseUri/metadata/$id$endpoint")
+
+  type MetadataEntity = EmbeddedEntity[ImageMetadata]
+
+  def metadataEntity(id: String, m: ImageMetadata): Option[MetadataEntity] =
+    Edits.noneIfEmptyMetadata(m).map(i =>
+      EmbeddedEntity(entityUri(id, "/metadata"), Some(i), actions = List(
+        Action("set-metadata-from-usage-rights", entityUri(id, "/metadata/set-from-usage-rights"), "POST")
+      ))
+    )
+
   def wrapMetadata(id: String): Reads[JsObject] =
     __.read[JsObject].map { metadata =>
       Json.obj(
@@ -55,7 +69,9 @@ class Transformers(services: Services) {
         // always returns Array(). This is just making sure that bug is replicated
         // so we can do equalities to see if the services are synced. This will
         // be rectified when we use Argo here.
-        "data" -> (metadata ++ Json.obj("keywords" -> Json.arr()))
+        "data" -> (metadata ++ Json.obj("keywords" -> Json.arr())),
+        "actions" -> List(Action("set-from-usage-rights",
+          URI.create(s"$metadataBaseUri/metadata/$id/metadata/set-from-usage-rights"), "POST"))
       )
     }
 
--- a/kahuna/public/js/edits/service.js
+++ b/kahuna/public/js/edits/service.js
@@ -149,8 +149,10 @@ service.factory('editsService',
         return existingRequestPool.promise;
     }
 
-    function perform(resource, action, originalImage) {
-        const newRequest = resource.perform(action).
+    // HACK: This is a very specific action that we use the `updateRequestPool` ast this action
+    // actually updates the metadata as a sideeffect.
+    function updateMetadataFromUsageRights(resource, originalImage) {
+        const newRequest = resource.perform('set-from-usage-rights').
               then(edit => getSynced(originalImage, newImage => matches(edit, newImage)));
 
         const existingRequestPool = updateRequestPools.get(resource) ||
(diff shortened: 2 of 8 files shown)


common-lib/src/main/scala/com/gu/mediaservice/api/Transformers.scala
```diff
@@ -46,6 +48,18 @@ class Transformers(services: Services) {
       )
     }
 
+  def entityUri(id: String, endpoint: String = ""): URI =
+    URI.create(s"$metadataBaseUri/metadata/$id$endpoint")
+
+  type MetadataEntity = EmbeddedEntity[ImageMetadata]
```
jamesgorrie: I moved these to a common response thing - but realised that's for another PR.

kahuna/public/js/edits/service.js
```diff
@@ -149,6 +149,17 @@ service.factory('editsService',
         return existingRequestPool.promise;
     }
 
+    function perform(resource, action, originalImage) {
+        const newRequest = resource.perform(action).
+              then(action => resource.get().then(edit => getSynced(originalImage, newImage => matches(edit, newImage))));
```
jamesgorrie: We perform the action and then refetch it and compare it.

jamesgorrie: @theefer - should an action return itself in the response, or the response it's modified. At the moment we're doing a mix of returning the resource with the `uri` of the action.

kahuna/public/js/upload/jobs/required-metadata-editor.js
```diff
@@ -20,9 +20,12 @@ jobs.controller('RequiredMetadataEditorCtrl',
 
     ctrl.saving = false;
     ctrl.disabled = () => Boolean(ctrl.saving || ctrl.externallyDisabled);
-    ctrl.metadata = metadataFromOriginal(ctrl.originalMetadata);
     ctrl.saveOnTime = 750; // ms
-    ctrl.copyrightWasInitiallyThere = !!ctrl.metadata.copyright;
+    ctrl.copyrightWasInitiallyThere = !!ctrl.originalMetadata.copyright;
+
+    $scope.$watch(() => ctrl.originalMetadata, newMetadata => {
```
theefer: Let's comment this.

kahuna/public/js/usage-rights/usage-rights-editor.js
```diff
@@ -182,11 +182,17 @@ usageRightsEditor.controller(
             const image = usageRights.image;
             const resource = image.data.userMetadata.data.usageRights;
             return editsService.update(resource, data, image).
-                then(resource => resource.data);
+                then(resource => resource.data).
+                then(() => updateMetadataFromUsageRights(image));
```
theefer: Let's comment this.

metadata-editor/app/controllers/EditsController.scala
```diff
@@ -158,7 +158,7 @@ object EditsController extends Controller with ArgoHelpers with DynamoEdits {
       .getOrElse(Future.failed(EditsValidationError("no-matching-metadata-found", "Couldn't find any matching metadata")))
     } recover {
       case NoItemFound => respondError(NotFound, "item-not-found", "Could not find image")
-      case e: EditsValidationError => respondError(BadRequest, e.key, e.message)
+      case e: EditsValidationError => respondError(NotFound, e.key, e.message)
```
theefer: Why?

kahuna/public/js/edits/service.js
```diff
@@ -149,6 +149,17 @@ service.factory('editsService',
...
+
+        const existingRequestPool = updateRequestPools.get(resource) ||
+            registerUpdateRequest(resource, originalImage);
```
theefer: I don't think it will give the correct behaviour to reuse the update pool for other actions..?

kahuna/public/js/edits/service.js
theefer: What is your POST returning currently?

jamesgorrie: This:
```JSON
{
    "data": {
        "credit": "The Guardian",
        "byline": "James Gorrie",
        "copyright": "Copyright",
        "keywords": []
    }
}
```

We could add the URI to the metadata endpoint as that is what it's modified, does that feel correct or like we're I'm retrofitting a reason to a solution?

kahuna/public/js/upload/jobs/required-metadata-editor.js
jamesgorrie: > Comment all the hacks

metadata-editor/app/controllers/EditsController.scala
jamesgorrie: User: "Set metadata from usage rights?"
Computer: "No metadata for those usage rights found."

kahuna/public/js/edits/service.js
jamesgorrie: I think they're pretty much one and the same thing - would it make more sense to just have a `requestPool`? What's the value in being able to differentiate between `update`s and `perform`s?

kahuna/public/js/edits/service.js
theefer: Yes I think that would make sense: return the created/altered entity, with the URI to identify it.

kahuna/public/js/edits/service.js
theefer: Let's have a chat after lunch.

metadata-editor/app/controllers/EditsController.scala
theefer: :+1:

kahuna/public/js/edits/service.js
jamesgorrie: We now return:

```Json
{
    "uri": "https://media-metadata.local.dev-gutools.co.uk/metadata/ebeda54078668254bb13e109619cd4787662683b/metadata",
    "data": {
        "description": "Cleaning up your mess",
        "credit": "The Guardian",
        "byline": "Alicia Canter",
        "keywords": []
    }
}
```

kahuna/public/js/edits/service.js
jamesgorrie: [Updated to the way more explicit method name](https://github.com/guardian/grid/commit/50bdf59eb1e4e591bc9f6175357f8e6ab6b185fa) to make sure we don't confuse it with anything else later.

theefer:
:+1:

PR #1252, opened 2015-09-16
split photographers between internal and external

jamesgorrie:
Further discussions with Jo B raises that these are people who have worked the picture desk and taken photos for us, but aren't photographers.

Understandably this clutters the list, but we need to make sure that we ingest these people correctly from Picdar or reindex them.

I've removed them from the UI, whilst still ensuring we categorise them correctly.

cc: @akash1810 with #1250

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
@@ -78,6 +78,7 @@ object MetadataConfig {
   // them correctly.
   // TODO: Think about removin these once Picdar is dead.
   val internalStaffPhotographers = List(
+    "E Hamilton West"       -> "The Guardian",
     "Emma Baddeley"         -> "The Guardian",
     "Harriet St Johnston"   -> "The Guardian",
     "James Michelson"       -> "The Guardian",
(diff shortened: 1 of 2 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
```diff
@@ -83,6 +73,24 @@ object MetadataConfig {
...
+  val internalStaffPhotographers = List(
+    "Emma Baddeley"         -> "The Guardian",
+    "Harriet St Johnston"   -> "The Guardian",
+    "James Michelson"       -> "The Guardian",
+    "Lorna Roach"           -> "The Guardian",
+    "Marcus Mays"           -> "The Guardian",
+    "Millie Burton"         -> "The Guardian",
+    "Rachel Vere"           -> "The Guardian",
+    "Richard Blake"         -> "The Guardian",
+    "Ted West"              -> "The Guardian"
```
akash1810: isn't `Ted West` aka `E Hamilton West`?

jamesgorrie: E Ha... Much better name.

akash1810:
:+1:

theefer:
Nice.

PR #1253, opened 2015-09-16
Use CROSS_FIELDS type in multi_match query.

akash1810:
The [default](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-match-query.html#multi-match-types) is [`BEST_FIELDS`](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-match-query.html#type-best-fields) which is most useful when searching for multiple words best found in the same field.

[`CROSS_FIELDS`](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-match-query.html#type-cross-fields) treats fields with the same analyzer as though they were one big field. Looks for each word in any field.

If we have the image:

```json
{
  "title": "foo",
  "description": "bar"
}
```

and we search for `foo bar`, `BEST_FIELDS` does not find it because its expecting the whole search term to exist in a single field.

Using `CROSS_FIELDS` means all the fields are (essentially) concatenated together and the search is run on that, and so a match is found.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Settings.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Settings.scala
@@ -3,6 +3,8 @@ package com.gu.mediaservice.lib.elasticsearch
 import play.api.libs.json.Json
 
 object IndexSettings {
+  // TODO rename `english_s_stemmer` as its an analyzer not a stemmer - would require a reindex.
+  val guAnalyzer = "english_s_stemmer"
 
   val englishSStemmer = Json.obj(
     "type" -> "custom",
@@ -32,7 +34,7 @@ object IndexSettings {
   )
 
   val analyzer = Json.obj(
-    "english_s_stemmer" -> englishSStemmer
+    guAnalyzer -> englishSStemmer
   )
 
   val analysis = Json.obj(
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
@@ -10,7 +10,7 @@ object Mappings {
   val nonAnalyzedString = Json.obj("type" -> "string", "index" -> "not_analyzed")
   val nonIndexedString  = Json.obj("type" -> "string", "index" -> "no")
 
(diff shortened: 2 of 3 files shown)


jamesgorrie:
Do we know if there's a performance hit on this?

theefer:
Worth checking that it works across the fields that have different analysers, e.g. description and byline.

akash1810:
> Do we know if there's a performance hit on this?

Docs don't say. Will have a search online and run a few queries on PROD to check.

akash1810:
> Worth checking that it works across the fields that have different analysers

From the [docs](https://www.elastic.co/blog/multi-field-search-just-got-better):

> All fields used in a cross_fields query should use the same analyzer so that they all produce the same list of query terms. If fields with different analyzers are queried, then they will be grouped together by analyzer. Each group will be queried with the cross_fields approach, then the scores from all groups will be combined with a bool query.

> Alternatively, you can force the same analyzer across all fields by specifying an analyzer in the query:
```json
{
    "multi_match": {
        "query":    "Quick brown fox",
        "fields": [ "title", "body" ],
        "type":     "cross_fields",
        "analyzer": "standard"
    }
}
```

Not sure we want to force the same analyzer.

akash1810:
This is on TEST now and working pretty well - performance doesn't seem to be impacted, though its hard to tell because elastic is quick anyway!

Search for `aardvark`, `oliphaunt` and `aardvark oliphaunt` for a demonstration.

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Settings.scala
```diff
@@ -3,6 +3,8 @@ package com.gu.mediaservice.lib.elasticsearch
 import play.api.libs.json.Json
 
 object IndexSettings {
+  // TODO rename `english_s_stemmer` as its an anayzer not a stemmer - would require a reindex.
```
jamesgorrie: Typo

jamesgorrie:
It might be good to surface somewhere: "For exact searches use `""`" or something - but only if it is a problem. :+1:

paperboyo:
> good to surface somewhere

There is quite a lot of syntax not documented in Advanced drop down. Maybe it won't all fit there? A link from there to a doc with _all_ syntax would be cool or from three-dot-menu... How was it, again, `@yesterday`?

PR #1255, opened 2015-09-16
Pass uri as part of image object for Composer to pick up

theefer:
I noticed `mediaApiUri` was missing in Composer when drag and dropping an image from the Grid into Composer. This seems to be caused by the `uri` missing on the image object in the drag transfer data. It may have been missing for many months, as I can't see where it'd have changed recently.

I'm curious if this going to reveal a higher usage of Grid images than we thought.

cc @cb372 who inadvertently helped reveal this issue.


jamesgorrie:
:+1:

PR #1256, opened 2015-09-16
Trim filename argument on upload

theefer:
When we finally support uploading files with leading/trailing whitespace, we'll want to trim the filename.


image-loader/app/controllers/Application.scala
```diff
@@ -166,4 +166,10 @@ class ImageLoader extends Controller with ArgoHelpers {
       }
     }
   }
+
+
+  // Find this a better home if used more widely
+  implicit class NonEmpty(s: String) {
```
jamesgorrie: :heartbeat: `implicit`s except when you can't find them.

jamesgorrie:
:+1:

image-loader/app/controllers/Application.scala
theefer: I think IDEs can help sometimes.

PR #1257, opened 2015-09-16
use same font size in all fields on uploads page.

akash1810:
Pointed out by @paperboyo 

# Before
screen shot 2015-09-16 at 13 35 20

# After
screen shot 2015-09-16 at 13 34 44

Hint: look at `Usage rights` and `File name`.


theefer:
:+1:

PR #1258, opened 2015-09-16
missed e hammilton west

jamesgorrie:
Doh :eyes:


akash1810:
oh, oops!

:+1:

paperboyo:
@jamesgorrie That list still contains `Andreij Krauze`. Correct name is `Andrzej Krauze`.

PR #1261, opened 2015-09-17
Make S3 Watcher logging more verbose

kenoir:
Adds current state to logs to make debugging simpler.

**Note:** This will log the API key when making an upload request.


theefer:
:+

PR #1262, opened 2015-09-17
update ordering to match supplied order

jamesgorrie:
Something we can test and improve on - but this was the suggestion from Jo B.


theefer:
:+1:

PR #1263, opened 2015-09-17
Creative commons

jamesgorrie:
Adding creative commons with a caution.
I've used the word caution as I think we will need to allow users to add warnings later, and didn't want to use `defaultWarning` as that implies it can be overrides as is the case with restrictions.

creative-commons

Release strategy - Kahuna then Metadata-editor - we can stagger this if we think it is completely nesseary. It will not break anything, but only show the caution when you have the new version of the js app.


theefer:
:+1: though please check wording etc with Robert?

PR #1264, opened 2015-09-17
Revert "Auto update md from ur"

jamesgorrie:
Reverts guardian/grid#1251

Found a bug that the jsonPatch screws up when you're inserting a new record...


jamesgorrie:
https://forums.aws.amazon.com/thread.jspa?threadID=162907

Issue #598, opened 2015-03-31
Surface "persisted" content in the UI


kenoir:
Persisted content should be shown in the UI with the reason for persistence:

- Cropped
- Staff photographer?
- User requested (by who)

Provide method for user to request persistence (perhaps reuse "archived/starred"

PR #1266, opened 2015-09-17
surface persisted state and reasons to ui

akash1810:
# Uploads page

screen shot 2015-09-17 at 16 00 05

# Image page

screen shot 2015-09-17 at 15 59 44

# Search
persisted

fixes https://github.com/guardian/grid/issues/598

### DIFF
--- a/kahuna/public/js/components/gr-image-persist-status/gr-image-persist-status.css
+++ b/kahuna/public/js/components/gr-image-persist-status/gr-image-persist-status.css
@@ -1,3 +1,9 @@
 gr-image-persist-status {
     display: block;
 }
+
+gr-image-persist-status ui-archiver,
+gr-image-persist-status .archiver {
+    display: inline-block;
+    height: inherit;
+}
--- a/kahuna/public/js/components/gr-image-persist-status/gr-image-persist-status.css
+++ b/kahuna/public/js/components/gr-image-persist-status/gr-image-persist-status.css
@@ -1,0 +1,3 @@
+gr-image-persist-status {
+    display: block;
+}
--- a/kahuna/public/js/components/gr-image-persist-status/gr-image-persist-status.html
+++ b/kahuna/public/js/components/gr-image-persist-status/gr-image-persist-status.html
@@ -1,5 +1,8 @@
-<gr-icon ng:if="! ctrl.states.canArchive">account_balance</gr-icon>
+<gr-icon-label ng:if="! ctrl.states.canArchive" gr-icon="account_balance">
+    <span ng:show="ctrl.withText">Persisted</span>
+</gr-icon-label>
 
 <ui-archiver ng:if="ctrl.states.canArchive"
              image="ctrl.image"
-             disabled="ctrl.disabled"></ui-archiver>
+             disabled="ctrl.disabled"
+             with-text="ctrl.withText"></ui-archiver>
--- a/kahuna/public/js/components/gr-image-persist-status/gr-image-persist-status.html
+++ b/kahuna/public/js/components/gr-image-persist-status/gr-image-persist-status.html
@@ -1,5 +1,5 @@
 <gr-icon-label ng:if="! ctrl.states.canArchive" gr-icon="lock"
-    title="Kept in library because: {{ctrl.states.persistedReasons}}">
+    title="Kept in library because image is: {{ctrl.states.persistedReasons}}.">
     <span ng:show="ctrl.withText">Kept in library</span>
 </gr-icon-label>
 
--- a/kahuna/public/js/components/gr-image-persist-status/gr-image-persist-status.html
+++ b/kahuna/public/js/components/gr-image-persist-status/gr-image-persist-status.html
@@ -1,6 +1,6 @@
-<gr-icon-label ng:if="! ctrl.states.canArchive" gr-icon="account_balance"
-    title="Persisted because: {{ctrl.states.persistedReasons}}">
-    <span ng:show="ctrl.withText">Persisted</span>
+<gr-icon-label ng:if="! ctrl.states.canArchive" gr-icon="lock"
+    title="Kept in archive because: {{ctrl.states.persistedReasons}}">
+    <span ng:show="ctrl.withText">Kept in archive</span>
 </gr-icon-label>
 
 <ui-archiver ng:if="ctrl.states.canArchive"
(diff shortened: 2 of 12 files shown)


jamesgorrie:
https://www.google.com/design/icons/#ic_lock_outline ?

akash1810:
@jamesgorrie that icon is already in use on the metadata panel.

akash1810:
for the record, I'm not convinced `account_balance` is the correct icon.

kahuna/public/js/components/gr-image-persist-status/gr-image-persist-status.css
```diff
@@ -0,0 +1,3 @@
+gr-image-persist-status {
```
jamesgorrie: Ido wish we had a `gr-* { display: block; }`

akash1810: that's cheating...

kahuna/public/js/edits/image-editor.html
```diff
@@ -69,10 +69,8 @@
                 </span>
             </div>
 
-            <ui-archiver
-                class="result-editor__archiver"
-                image="ctrl.image"
-            ></ui-archiver>
+            <gr-image-persist-status class="result-editor__archiver"
+                                     image="ctrl.image"></gr-image-persist-status>
```
jamesgorrie: We should set the convention of `image` vs `gr:image`

akash1810: This applies to all attributes too, right?

jamesgorrie: Any custom attributes that we stick on `scope`.

jamesgorrie: It's a little annoying as you land up with:
```JS
{
  image: '=grImage'
}
// or
delete(ctrl.grImage);
```

akash1810: > `{ image: '=grImage' }`

:+1: 

(will change any other existing custom attributes to this standard as I see them).

kahuna/public/js/components/gr-image-persist-status/gr-image-persist-status.js
```diff
@@ -0,0 +1,51 @@
...
+        ctrl.states = imageService(ctrl.image).states;
+
+        const freeUpdateListener = $rootScope.$on('image-updated', (e, updatedImage) => {
+            if (ctrl.image.data.id === updatedImage.data.id) {
+                ctrl.states = imageService(updatedImage).states;
+                ctrl.image = updatedImage;
+            }
+        });
+
+        $scope.$on('$destroy', () => freeUpdateListener());
```
jamesgorrie: `$scope.$on('$destroy', freeUpdateListener);`?

kahuna/public/js/image/service.js
```diff
@@ -25,11 +25,16 @@ imageService.factory('imageService', [function() {
...
     function getStates(image) {
+        const persistReasons = image.data.persisted.reasons;
+
         return {
             cost: image.data.cost,
             hasCrops: hasExportsOfType(image, 'crop'),
             isValid: image.data.valid,
-            canDelete: image.getAction('delete').then(action => !! action)
+            canDelete: image.getAction('delete').then(action => !! action),
+            canArchive: image.data.persisted.value === false ||
```
jamesgorrie: So are we saying you can't archive this if it's already persisted?
If so - why not? That's not true on the API level is it?

akash1810: > So are we saying you can't archive this if it's already persisted?

Yep. Archiving an image that already has a crop has no real effect, other than possibly going the user reassurance that it won't be deleted. If we remove/lock the option of archiving in this scenario, then it means less clicks for the user.

> That's not true on the API level is it?

Correct. The API allows you to POST to archived regardless of the value of persisted.

paperboyo:
>>So are we saying you can't archive this if it's already persisted?

>Yep. Archiving an image that already has a crop has no real effect

By "archive" you mean "star"? I thought the whole decoupling meant that the star can at last be used to mean "good pic", instead of meaning all these different things... The pic can be good whether or not it has crops, comes from PicDarLib, has labels etc. The above suggests I won't be able to say that it's good in any of these scenarios. If that's true and intentional, maybe we should get rid of a star and replace it with that LibraryofCongress icon (at least it's not a diskette :)?

theefer:
Nice!

A few comments, all fairly minor:

It's not obvious that the "[open lock] Kept in library" label on the image page is a button (especially as, when a [closed lock], it's not). Should it perhaps change label on hover to "Remove from library"?

image

The visual difference between the open and closed lock is really minor. Hard to see, especially at small size (e.g. on upload page). Not sure what to do with it, maybe let's just monitor feedback on that...?

"Kept in library because: exports" may not be very obvious to people? I think it makes sense to send that key semantically from the API, but maybe we could remap it to "cropped" or something?

Same with "Kept in library because: persistence-identifier". Would have to allude to the Picdar Library to make it clearer though, which makes it less generic :cry:

kahuna/public/js/components/gr-image-persist-status/gr-image-persist-status.html
```diff
@@ -0,0 +1,9 @@
+<gr-icon-label ng:if="! ctrl.states.canArchive" gr-icon="lock"
+    title="Kept in library because: {{ctrl.states.persistedReasons}}">
```
jamesgorrie: Could we build up nicer sentences here.
"Kept in library because of: exports" => "Kept in library because image has exports".

Also not sure how people feel about having the [help cursor](http://www.echoecho.com/csscursors.htm) on tootips as the title takes a smidgen of a second to appear.

akash1810:
> Should it perhaps change label on hover to "Remove from library"?

Done.

> "Kept in library because: exports" may not be very obvious to people? I think it makes sense to send that key semantically from the API, but maybe we could remap it to "cropped" or something?

> Same with "Kept in library because: persistence-identifier". Would have to allude to the Picdar Library to make it clearer though, which makes it less generic :cry:

Done.

I've released this branch to TEST too.

kahuna/public/js/components/gr-image-persist-status/gr-image-persist-status.html
akash1810: > Could we build up nicer sentences here.

Done.

theefer:
:+1: great stuff!

PR #1267, opened 2015-09-17
add short descriptions

jamesgorrie:
Concerns:
* References to GNM (removed when originally done) make it very GNMy... not necessary really.
* ~~All caps to make a point - perhaps this should be a caution (see: Social Media)?~~ Made it a caution.

Otherwise they're almost what we had.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
@@ -235,8 +235,10 @@ case class SocialMedia(restrictions: Option[String] = None)
     val name = "Social Media"
     val description =
       "Images grabbed from social media to support breaking news where no other image is available " +
-      "from usual sources. APPROVAL NEEDED FROM SENIOR EDITOR IF PERMISSION FROM OWNER CANNOT BE " +
-      "ACQUIRED."
+      "from usual sources."
+
+   override val caution =
+    Some("Approval needed from senior editor if permission from ownercannot be aqcuired")
   }
 object SocialMedia {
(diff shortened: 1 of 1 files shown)


common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
```diff
@@ -233,11 +234,11 @@ case class SocialMedia(restrictions: Option[String] = None)
...
-      "Taken from public websites and social media to support " +
-      "breaking news where no other image is available from usual sources. " +
-      "Permission should be sought from the copyright holder, but in " +
-      "extreme circumstances an image may be used with the approval of " +
-      "a senior editor."
+      "Images grabbed from social media to support breaking news where no other image is available " +
+      "from usual sources."
+
+   override val caution =
+    Some("Approval needed from senior editor if permission from ownercannot be aqcuired")
```
theefer: Missing space between words.

theefer: typo in "aqcuired"

common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
```diff
@@ -214,8 +216,7 @@ case class GuardianWitness(restrictions: Option[String] = None)
     val defaultCost = Some(Conditional)
     val name = "Guardian Witness"
     val description =
-      "Provided by readers in response to callouts and assignments on " +
-      "GuardianWitness."
+      "Images provided by readers in response to callouts and assignments on Guardian Witness."
```
theefer: There is no space between `Guardian` and `Witness`.

theefer:
A few typos, else :+1:

PR #1273, opened 2015-09-18
deprecate old cost model from cost calculator

jamesgorrie:
:disappointed:  Either way - this will make sure that the cost is reflected in the UI. :disappointed:


theefer:
:+1:

Issue #1172, opened 2015-09-01
Top bar baselines out of whack


alastairjardine:
![my eyes](https://cloud.githubusercontent.com/assets/4292469/9601462/9bef6d4a-5099-11e5-9c4c-dad60419a623.png)

Baselines are all over the place – would be lovely to tidy them up.
![css](https://cloud.githubusercontent.com/assets/4292469/9601478/b97b84fc-5099-11e5-8a83-4f79b0fc0fc5.gif)

PR #1276, opened 2015-09-18
allow uploads after uploading

akash1810:
Display the file upload button once all uploads have finished. The behaviour is:
1. Upload image
2. Fill in metadata
3. Press Upload button
4. Images from 1. are moved to recent uploads

Note, this is the current behaviour if you replace 3. with drag and drop more images.

upload-after-upload

I view this as a temporary solution until we have "add to upload queue" abilities.

### DIFF
--- a/kahuna/public/js/upload/prompt/prompt.css
+++ b/kahuna/public/js/upload/prompt/prompt.css
@@ -5,8 +5,12 @@ file-prompt {
 }
 
 file-prompt file-uploader {
-    display: flex;
-    justify-content: center;
+    display: inline-block;
+    padding: 0 5px;
+}
+
+file-prompt file-uploader .file-uploader__select-files {
+    margin-top: 0;
 }
 
 file-prompt .message {
--- a/kahuna/public/js/upload/prompt/prompt.html
+++ b/kahuna/public/js/upload/prompt/prompt.html
@@ -1,9 +1,3 @@
-<file-uploader></file-uploader>
-
-<p class="message">
-    Select files to upload.
-</p>
-
-<p class="message">
-    Either drag 'n drop images onto this screen or click the upload button to get your images on the Grid.
-</p>
+<div class="message">
+    Either drag 'n drop images onto this screen or click <file-uploader></file-uploader> to get your images on the Grid.
+</div>
--- a/kahuna/public/js/upload/prompt/prompt.js
+++ b/kahuna/public/js/upload/prompt/prompt.js
@@ -11,6 +11,7 @@ prompt.directive('filePrompt', [function () {
     return {
         restrict: 'E',
         transclude: 'replace',
+        scope: {}, // ensure isolated scope
         template: template
     };
 }]);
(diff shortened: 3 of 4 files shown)


theefer:
It seems odd to add it as part of the Your current uploads header. Seems to imply that it may add to the current uploads (a requested feature), but I don't think it does? Equally risky is to re-introduce the confusion of some users as so whether dropped/selected files are uploaded or not once shown in Your current uploads, and lead them to think that to finish the upload you have to click on the button?

Would it be simpler/clearer to just leave the upload area all the time? Not sure it would, just wondering..

akash1810:
> Seems to imply that it may add to the current uploads (a requested feature), but I don't think it does?

No it doesn't. I view this as an quick win until we add the "add to upload job" feature.

> Equally risky is to re-introduce the confusion of some users as so whether dropped/selected files are uploaded or not once shown in Your current uploads, and lead them to think that to finish the upload you have to click on the button?

Fair point.

> Would it be simpler/clearer to just leave the upload area all the time? Not sure it would, just wondering..

Will try it out.

akash1810:
using full prompt, rather than upload button.

upload

theefer:
Is there a particular reason for hiding the prompt temporarily during the upload? It makes the whole page jump down when the uploads finish.

akash1810:
> Is there a particular reason for hiding the prompt temporarily during the upload?

Can't get it to work properly otherwise :unamused:. Most likely to do with isolating scope, will look at again.

akash1810:
upload-upload-upload

jamesgorrie:
Jonny sent through this a while ago - which could make that bar at the top a lot smaller (and also makes more sense I think). The reason I mention it is I can see it getting annoying to have to scroll, even if all you've done is upload 1-2 images on a small screen.

instructions2

akash1810:
@jamesgorrie ooh, I like! Updated.

kahuna/public/js/upload/prompt/prompt.css
```diff
@@ -5,8 +5,12 @@ file-prompt {
...
 
 file-prompt file-uploader {
-    display: flex;
-    justify-content: center;
+    display: inline-block;
+    padding: 0 5px;
+}
+
+file-prompt file-uploader .file-uploader__select-files {
+    margin-top: 0;
```
akash1810: Bit annoying that we have to do this. It looks like `margin-top` is baked into the component for positioning in `gr-top-bar` of the search page, which isn't ideal as its coupling page specific styles into components. I tried (quickly) decoupling, but it got messy and made https://github.com/guardian/grid/issues/1172 worse.

jamesgorrie:
The only thing I can think of to check with users is whether they expect their images to be added to their current uploads queue (to apply batch edits) or if jumping to complete is expected.

Either way this is current behaviour and can tested / iterated on later :+1:

jamesgorrie:
Assuming we want to remove the need for this banner at some stage too?

PR #1277, opened 2015-09-18
No usage rights

jamesgorrie:
When embarking on doing this, I was left trying to figure out what changed what state, which is a little frustrating.

This hopefully makes things a lot more understandable and leaves us able to enhance this in the future.

Weirdly I noticed that I am doing things similar to what's being done in [`gr-panel`](https://github.com/guardian/grid/blob/master/kahuna/public/js/components/gr-panel/gr-panel.js#L108-L135). Hopefully these will bring us closer to one abstraction to manage this (and coming to the same solution twice is mostly a good sign).

Another addition is we can now have a more correct model i.e. If two images are of different categories, but have they same restrictions - this will be reflected.

__TODO__
- [x] Make sure we have feature parity
- [x] remove legacy code

### DIFF
--- a/kahuna/public/js/usage-rights/usage-rights-editor.html
+++ b/kahuna/public/js/usage-rights/usage-rights-editor.html
@@ -11,7 +11,7 @@
             ng:model="ctrl.category"
             ng:disabled="ctrl.saving"
             ng:options="category as category.name for category in ctrl.categories track by category.value"
-            ng:change="ctrl.reset()">
+            ng:change="ctrl.changeCategory()">
         </select>
     </label>
 
--- a/kahuna/public/js/usage-rights/usage-rights-editor.html
+++ b/kahuna/public/js/usage-rights/usage-rights-editor.html
@@ -7,7 +7,7 @@
         ng:if="ctrl.multipleImages()">
         multiple usage rights
     </div>
-    
+
     <label class="ure__category">
         <!-- We reset the model here else properties remain attached to the\
         model, even though they don't exist in the form -->
@@ -92,10 +92,11 @@
                             ng:required="property.required" />
 
                         <div ng:if="property.options"
-                             ng:switch="property.options.length > 3">
+                             ng:switch="property.options.length > 3"
+                             ng:init="options = ctrl.getOptsFor(property)">
                                 <div class="radio-list" ng:switch-when="false">
                                     <div class="radio-list__item"
-                                        ng:repeat="o in property.options">
+                                         ng:repeat="o in options">
 
                                         <input type="radio"
                                             ng:value="o.value"
@@ -121,7 +122,7 @@
                                     ng:switch-when="true"
                                     ng:model="ctrl.model[property.name]"
                                     ng:required="property.required"
-                                    ng:options="o.value as o.key for o in property.options">
+                                    ng:options="o.value as o.key for o in options">
                                 </select>
                         </div>
 
--- a/kahuna/public/js/usage-rights/usage-rights-editor.html
+++ b/kahuna/public/js/usage-rights/usage-rights-editor.html
@@ -11,7 +11,7 @@
             ng:model="ctrl.category"
             ng:disabled="ctrl.saving"
             ng:options="category as category.name for category in ctrl.categories track by category.value"
-            ng:change="ctrl.changeCategory()">
+            ng:change="ctrl.reset()">
         </select>
     </label>
 
--- a/kahuna/public/js/usage-rights/usage-rights-editor.html
+++ b/kahuna/public/js/usage-rights/usage-rights-editor.html
@@ -7,7 +7,7 @@
         ng:if="ctrl.multipleImages()">
         multiple usage rights
     </div>
-
+    
     <label class="ure__category">
         <!-- We reset the model here else properties remain attached to the\
         model, even though they don't exist in the form -->
--- a/kahuna/public/js/usage-rights/usage-rights-editor.html
+++ b/kahuna/public/js/usage-rights/usage-rights-editor.html
@@ -93,7 +93,7 @@
 
                         <div ng:if="property.options"
                              ng:switch="property.options.length > 3"
-                             ng:init="options = ctrl.getOptsFor(property)">
+                             ng:init="options = ctrl.getOptionsFor(property)">
                                 <div class="radio-list" ng:switch-when="false">
                                     <div class="radio-list__item"
                                          ng:repeat="o in options">
@@ -135,7 +135,7 @@
                                 name="{{ property.name }}"
                                 ng:model="ctrl.model[property.name]"
                                 ng:required="property.required"
-                                ng:options="o for o in ctrl.getOptionsFor(property)"
+                                ng:options="o for o in ctrl.getOptionsMapFor(property)"
                                 ng:if="!otherValue">
                                 <option value="">None</option>
                             </select>
--- a/kahuna/public/js/usage-rights/usage-rights-editor.html
+++ b/kahuna/public/js/usage-rights/usage-rights-editor.html
@@ -3,11 +3,6 @@
       name="usageRights"
       ng:submit="usageRights.$valid && ctrl.save()">
 
-    <div class="ure__multi"
-        ng:if="ctrl.multipleImages()">
-        multiple usage rights
-    </div>
-
     <label class="ure__category">
         <!-- We reset the model here else properties remain attached to the\
         model, even though they don't exist in the form -->
@@ -30,7 +25,7 @@
 
     <div class="ure__properties" ng:if="ctrl.category">
         <div class="ure__multiple-rights-warning"
-            ng:if="ctrl.multipleUsageRights()">
+            ng:if="ctrl.usageRights.length > 1">
             Multiple usage rights
             <gr-icon
                 title="Some specific information about the usage rights of these images might be missing below and will be overridden.">
--- a/kahuna/public/js/usage-rights/usage-rights-editor.html
+++ b/kahuna/public/js/usage-rights/usage-rights-editor.html
@@ -174,7 +174,7 @@
         <button class="ure__action button-ico button-save"
             type="submit"
             title="save usage rights overrides"
-            ng:disabled="ctrl.savingDisabled() || !usageRights.$valid">
+            ng:disabled="ctrl.savingDisabled || !usageRights.$valid">
             <gr-icon-label ng:if="!ctrl.saving" gr-icon="check">Save</gr-icon-label>
             <gr-icon ng:if="ctrl.saving">timelapse</gr-icon>
         </button>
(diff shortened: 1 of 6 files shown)


jamesgorrie:
norights

kahuna/public/js/usage-rights/usage-rights-editor.js
```diff
@@ -6,125 +6,115 @@ import './usage-rights-editor.css!';
...
-    };
-
-
-    const getGroupCategory =
-        (usageRights, cats) => cats.find(cat => cat.value === usageRights.reduce(
-            (m, o) => (m == o.data.category) ? o.data.category : {},
-            usageRights[0].data.category
-        ));
+    const usageRights$ = new Rx.BehaviorSubject([]);
+    $scope.$watch(() => ctrl.usageRights, usageRightsList => {
```
theefer: Would `observe$` not work for this?

kahuna/public/js/usage-rights/usage-rights-editor.js
```diff
@@ -134,70 +124,60 @@ usageRightsEditor.controller(
...
 
-    function updateSuccess(data) {
-        ctrl.model = data;
-        ctrl.onSave();
-        setCategory(data.category);
-        uiSaved();
-        ctrl.saving = false;
+    function unique(arr) {
+        return arr.reduce((prev, curr) =>
+            prev.indexOf(curr) !== -1 ? prev : prev.concat([curr]), []);
```
theefer: Note that you can pass items to `concat`, e.g. `prev.concat(curr)`.

kahuna/public/js/usage-rights/usage-rights-editor.js
jamesgorrie: Yip - I hadn't discovered that when I started - changed.

kahuna/public/js/usage-rights/usage-rights-editor.js
```diff
@@ -134,70 +124,60 @@ usageRightsEditor.controller(
...
     }
 
-    function uiSaved() {
-        ctrl.saved = true;
-        $timeout(() => ctrl.saved = false, 1500);
+    // takes an array of objects and turns it into an object with an array of unique values
+    // e.g. [{ a: 1, b: 2 }, { a: 2, b: 2, c: 3 }] => { a: [1,2], b: [2], c: [3] }
+    function reduceObjectsToArrays(objects) {
```
theefer: You may be able to use `getSetOfProperties` from `imageList` (though it uses immutable data structures rather than Object/Array)?

kahuna/public/js/usage-rights/usage-rights-editor.js
jamesgorrie: Didn't realise that - thanks.

kahuna/public/js/usage-rights/usage-rights-editor.js
jamesgorrie: What is `getSetOfProperties` expecting given that you can run `flatMap` on it?

theefer: Ah, it expects an immutable List as input.

jamesgorrie: Ah - is it a stream of a array?

theefer: No, it's not an observable, just an [immutable List](https://facebook.github.io/immutable-js/docs/#/List).

kahuna/public/js/usage-rights/usage-rights-editor.js
```diff
@@ -6,125 +6,112 @@ import './usage-rights-editor.css!';
...
-        ctrl.categories = catsWithNoRights();
-        ctrl.category = noRights;
-    };
-    const setMultiRightsCats = () => {
-        ctrl.categories = catsWithMultiRights();
-        ctrl.category = multiRights;
-    };
-
+    // @return Stream.<Array.<UsageRights>>
+    const usageRights$ = observe$($scope, () => ctrl.usageRights).startWith([]);
```
theefer: Would `observe$($scope, 'ctrl.usageRights')` work too? Not sure it's genuinely better, or more performant, but I noticed you use the construct a lot in this file.

kahuna/public/js/usage-rights/usage-rights-editor.js
```diff
@@ -6,125 +6,112 @@ import './usage-rights-editor.css!';
...
+            }
+            return model;
+        }, {});
+    });
 
-                    cat.properties[i].options = propertyOptions;
-                }
-            });
-        });
+    const savingDisabled$ = category$.combineLatest(category$, cat => cat === multiCat);
```
theefer: `combineLatest` with itself? Is that a `map`?

kahuna/public/js/usage-rights/usage-rights-editor.js
```diff
@@ -6,125 +6,112 @@ import './usage-rights-editor.css!';
...
-    // TODO: How do we make this more synchronous? You can only resolve on the
-    // routeProvider, which is actually bound to the UploadCtrl in this instance
-    // SEE: https://github.com/angular/angular.js/issues/2095
-    ctrl.save = () => {
-        // FIXME [1]: We do this as images that didn't have usagerights in the first place will have
-        // the no rights option, this is to make sure we don't override but rather delete.
-        const data = modelToData(ctrl.model);
+    const showRestrictionsChange$ = observe$($scope, () => ctrl.showRestrictions);
+    const showRestrictions$ = showRestrictionsChange$.combineLatest(
+        forceRestrictions$, model$, (forceRestrictions, showRestrictions, model) => {
```
theefer: I'm a little confused by this self-referential stream.

Also, are the arguments in the `resultSelector` function in the correct order?

jamesgorrie: >  are the arguments in the resultSelector function in the correct order

No they were't.

I've renamed the streams - it's an complex situation.
* If there are no restrictions and they are not required => false
* If there are restrictions but they are not required => false - but you can turn them off
* If they are required => require them

Does that make sense - probably better :panda_face: 2 :dragon_face: if not.

kahuna/public/js/usage-rights/usage-rights-editor.js
jamesgorrie: Happy to chat through this - but I'd rather :bike::house: it as this current method provides the data structure the template needs.

kahuna/public/js/usage-rights/usage-rights-editor.js
```diff
@@ -134,70 +118,61 @@ usageRightsEditor.controller(
...
+        }));
+    }
+
+    function getUniqueCats(usageRights) {
+        return unique(usageRights.map(ur => ur.data.category));
     }
 
-    function updateSuccess(data) {
-        ctrl.model = data;
-        ctrl.onSave();
```
theefer: I noticed `onSave` is no longer called from anywhere?

jamesgorrie: Added it as before...

kahuna/public/js/usage-rights/usage-rights-editor.js
jamesgorrie: It does work - I've always seemed to have settled for the function method as it fells less magic.

We seem to have most commnly gone with the function method on `$watch`es, but assuming this is because a `$watch` can't take a string.

I can't find any documentation on performance, so happy to go with either.

theefer: `$watch` can take a string. The string has to be an expression that evaluates on the scope though, so not some custom function, and presumably because we sometimes needed the freedom of custom functions, we also used functions even when a string would have worked.

I don't mind too much either way, just pointing it out.

theefer:
Can we catch up on this tomorrow? Looks nicer generally, I just have some concerns about the circularity of some of the state.

jamesgorrie:
Yip - I thought I removed most circularity - I will take another run through and let's catch up.

kahuna/public/js/usage-rights/usage-rights-editor.js
jamesgorrie: It is and has been changed.

kahuna/public/js/usage-rights/usage-rights-editor.js
jamesgorrie: Just updated it as it's much nicer - we might want to move that function to a util function - another PR though as too many moving parts.

jamesgorrie:
@theefer - removed cycle around `showRestrictions` by... removing it.
Removed the category cycle by injecting category, but using the `userChange` in the streams that need it  i.e. `savingDisabled`.

I think I made the user changing things integral - where it's actually not for most things.

kahuna/public/js/usage-rights/usage-rights-editor.html
```diff
@@ -16,7 +11,7 @@
             ng:model="ctrl.category"
             ng:disabled="ctrl.saving"
             ng:options="category as category.name for category in ctrl.categories track by category.value"
-            ng:change="ctrl.reset()">
+            ng:change="ctrl.changeCategory()">
```
theefer: I can't find where this is declared?

kahuna/public/js/usage-rights/usage-rights-editor.js
```diff
@@ -8,123 +15,108 @@ import '../components/gr-confirm-delete/gr-confirm-delete.js';
...
+            const uniqeCat = uniqueCats[0] || '';
+            return cats.find(cat => cat.value === uniqeCat);
         } else {
-            setStandardCats();
+            return multiCat;
         }
+    });
 
+    // @return Stream.<Category>
+    const categoryFromUserChange$ = observe$($scope, () => ctrl.category).filter(cat => !!cat);
```
theefer: Can we just document the `filter` hack?

kahuna/public/js/usage-rights/usage-rights-editor.html
jamesgorrie: That was debugging sorry.

kahuna/public/js/usage-rights/usage-rights-editor.js
jamesgorrie: Yea, dug a little deeper with no luck...

jamesgorrie:
[So we're a little out of whack where the category is no longer the canonical category when the user changes it](https://github.com/guardian/grid/commit/82bf15b6b76b23467549bf8278c1e5c6bcb01c88#diff-81f57b8057236574f9266cb5d4cb54e3R48). It'd be good to explore how we deal with user input / streams - but this works for now.

theefer:
Nifty :+1:

jamesgorrie:
@theefer - thanks for the `Rx`y help on this.

PR #1278, opened 2015-09-21
safer typecasting to fix endpoint

akash1810:
When visiting `https://api.media....co.uk/images/edits/label{?q}` the following is thrown:

```
ClassCastException: org.elasticsearch.search.aggregations.bucket.terms.UnmappedTerms cannot be cast to org.elasticsearch.search.aggregations.bucket.terms.StringTerms
```

Cast as the more abstract `InternalTerms` which both `UnmappedTerms` and `StringTerms` extend.

### DIFF
--- a/media-api/app/lib/elasticsearch/ElasticSearch.scala
+++ b/media-api/app/lib/elasticsearch/ElasticSearch.scala
@@ -3,7 +3,7 @@ package lib.elasticsearch
 import java.util.regex.Pattern
 
 import org.elasticsearch.index.query.{FilterBuilder, FilteredQueryBuilder}
-import org.elasticsearch.search.aggregations.bucket.terms.StringTerms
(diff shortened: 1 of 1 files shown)


jamesgorrie:
:+1:

theefer:
:+1: though ideally we should not use casting, which is unsafe.

jamesgorrie:
@theefer we should hopefully be deprecating this for the suggesters, so didn't think it mattered.
I remember @philwills mentioning he was doing (but potentially didn't complete) some work around getting results in a better way - as this is rubbish.

philwills:
Unfortunately you're right @jamesgorrie. I got a PR to ES to make this more possible, then realised I'd changed the access to the wrong property and couldn't face going back to say that I'd screwed up.

PR #1279, opened 2015-09-21
Restrict image with the Guardian credit

jamesgorrie:
I've added this into the hack method to keep it all self contained although there might have been a way to change the ordering of the whole thing to look at restrictions only once - but that felt like pollution.


theefer:
:+1: but we should work to retire the Guardian hack soon.

jamesgorrie:
> retire... soon
:+1:

PR #1280, opened 2015-09-21
extract labeller css into its own file.

akash1810:
This is a non-functional change, it just makes the css easier to manage.

I've noticed the styling of labels on PROD is a bit off when in selection mode (both in the search results and the gr-panel). Will fix in a follow-up PR.


jamesgorrie:
:+1:

kahuna/public/js/edits/labeller.css
```diff
@@ -0,0 +1,75 @@
+/* ==========================================================================
+   Labeller
+   ========================================================================== */
+.labeller {
+    font-size: 1.3rem;
+    display: flex;
+}
+
+.label {
```
theefer: This class seems to appear in other places outside of the labeller directive (e.g. in the gr-panel component). Let's try and be mindful of isolating classes that have side-effects elsewhere.

Issue #765, opened 2015-05-27
Index Colorspace


kenoir:
In order to produce "correct" crops for images with no color profile we ned to be able to infer the color model for images with no ICC profile.

We previously tried to do this here: https://github.com/guardian/media-service/pull/748 but ran into problems.

PR #1281, opened 2015-09-21
Correct colour space before cropping

theefer:
Currently, the cropper fails in at least two cases:
* If cropping an image where the colour space in the ICC profile doesn't match the actual colour model of the image data. e.g. an image with RGB image data with a CMYK profile (note that both such an image is technically incorrect). This is because when trying to convert from CMYK to sRGB profile, GraphicsMagick detects that the image data isn't actually CMYK and it fails with an error about mismatching colour models.
* If cropping an image where the image data is not in RGB (e.g. CMYK or grayscale) and there is no embedded ICC profile. In that case, adding the profile doesn't trigger any conversion; the sRGB profile is simply embedded, but it doesn't match the image data, resulting in an image that doesn't look correct in browsers.

The solution is to always read out both the embedded ICC profile colour space and the image data colour model and compare them before any conversion.

If they match, great, we can just apply our normal conversion to sRGB profile (from whatever profile is correctly embedded).

If they don't match, either because there is no embedded ICC profile, or because the embedded ICC profile is not matching the actual colour model, then we strip any embedded profile and replace it with a generic ICC profile that matches the actual colour model of the image data. Note that there is no guarantee that the profile will exactly match the image data, but it's a best effort estimate with standard ICC profiles. We can then resume the normal flow of converting the image to the desired output profile (sRGB), etc.

Initial tests with various images with mismatching or missing ICC profiles gave correct results.

This will also be used to correct the colour space of thumbnails we generate, in a future PR.

Thanks to @paperboyo for his big help with this!

## TODO

- [x] More manual testing
- [x] Think of whether we can write automated tests for this somehow


common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/im4jwrapper/ImageMagick.scala
```diff
@@ -18,9 +21,20 @@ object ImageMagick {
...
   def addDestImage(op: IMOperation)(dest: File) = op <| (_.addImage(dest.getAbsolutePath))
   def crop(op: IMOperation)(b: Bounds): IMOperation = op <| (_.crop(b.width, b.height, b.x, b.y))
   def profile(op: IMOperation)(profileFileLocation: String): IMOperation = op <| (_.profile(profileFileLocation))
+  def thumbnail(op: IMOperation)(width: Int): IMOperation = op <| (_.thumbnail(width))
   def scale(op: IMOperation)(dimensions: Dimensions): IMOperation = op <| (_.scale(dimensions.width, dimensions.height))
+  def format(op: IMOperation)(definition: String): IMOperation = op <| (_.format(definition))
+
   def runConvertCmd(op: IMOperation): Future[Unit] = Future((new ConvertCmd).run(op))
+  def runIdentifyCmd(op: IMOperation): Future[List[String]] = Future {
+    val cmd = new IdentifyCmd()
```
kenoir: Are you presuming that the identify command infers the colorspace from (a) some heuristic on the image data itself, or (b) uses image metadata?

When looking at this previously I was defeated by the ImageMagick source code being impenetrable to me. I _think_ testing indicated that it is in fact (a). If this is the case then images with incorrect or missing metadata will still fail this test.

To confirm which situation is occurring you could intentionally strip all metadata from an image and then run the identify command to see whether it can correctly infer the colorspace.

paperboyo:
:gem:! Thanks, @theefer, great explanation!

Just some thoughts as I can’t read code:

`Note: ICC identified as "icm" here`

Are we stripping both `icm` AND `icc` (the latter being much more common)?

I’m worried about robustness of our colour model inference via `jpeg-colorspace-name`. This is not available in IM whose recent, better colour management changes may have not been ported to GM yet. It also reads `jpeg` ;-)

In general with that, I would like to see if among all our imagery there is even one image that is not correctly identified by the above but opens in Photoshop. If yes, I would like to find more robust way of identification. Only completely broken images have no colour model...

The above detection could inform our db (it may be useful to list colour model in Show all metadata and be able to search for it): https://github.com/guardian/grid/issues/765

common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/im4jwrapper/ImageMagick.scala
paperboyo: > To confirm which situation is occurring you could intentionally strip all metadata from an image and then run the identify command to see whether it can correctly infer the colorspace.

Yes, it (GM) can. After running `-strip`, the `jpeg-colorspace-name` is still present.

kenoir:
> jpeg-colorspace-name

This. I think I gave up last time I tried this because it's not possible to work out the colorspace without the metadata, (which kind of defeats the point). I seem to remember there are Java libraries that do run heuristics on the actual image data (possibly ImageJ) or indeed as @paperboyo says it's quite possibly in the most recent versions of ImageMagick.

common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/im4jwrapper/ImageMagick.scala
kenoir: @theefer i'd try this with the same setup as in production (can't actually remember if we are using IM or GM, and the version may have an impact).

theefer:
As far as we could see, `jpeg-colorspace-name` is using the image data, because using it fixes the known bugs we have, and it likely wouldn't be the case if it was just reading metadata.

@paperboyo what are your worries specifically? Could we run some tests to verify them?

> Are we stripping both icm AND icc?

No, I ran into an issue when stripping just ICC and I thought ICM was the canonical name, but if they're genuinely different we could try and strip both?

paperboyo:
@kenoir `jpeg-colorspace-name` is not present in most recent IM. It is present in most recent GM, though. In most recent IM this property _may_ have been extracted earlier in the output as `Colorspace:` (although, watch out - it reads `sRGB` which just means `RGB`, really).

In both GM and IM, these properties are still present after `-strip`, though. I wonder what would make them disappear, but still make image readable...

common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/im4jwrapper/ImageMagick.scala
theefer: Seems to work fine on the cropper instance from some quick tests.

kenoir:
If this PR fixes the problem in TEST then it's all fine but it's worth trying. I seem to remember being frustrated with the difference between my sandbox and a "real" instance.

paperboyo:
This is a comparison of latest IM & GM properties we are interested in, that __survive__ `-strip` commands, run, in order, on an RGB, a CMYK and a grayscale JPG file:

| IM  | GM |
| ------------- | ------------- |
| (very beggining) `Colorspace:` `sRGB`, `CMYK`, `Gray`  | `JPEG-Colorspace-Name:` `RGB`, `CMYK`, `GRAYSCALE` |
| (under `properties`) `jpeg:colospace:` `2`, `4`, `1`  | `JPEG-Colorspace:` `2`, `4`, `1`  |
| (very beggining) `Type:` `TrueColor`, `ColorSeparation`, `Palette` | (very beggining) `Type:` `true color`, `color separated`, `grayscale` |

paperboyo:
> No, I ran into an issue when stripping just ICC and I thought ICM was the canonical name, but if they're genuinely different we could try and strip both?

As fare as I know, when they differ, it pertains only to filename extensions. Part of the image spec is ICC, so I’m not sure why IM or GM even lets you differentiate between them (running any should strip ALL colour profile data).

Okay, I’just checked: added an ICM profile to one image and ICC profile to another (the same profile, really, just different filenames :). Then run `+profile "icc,icm"` on both of these in _both_ IM and GM. Both apps correctly removed profiles from both images. This command seems safe to use.

Annoyingly, if you do just `+profile "ICC"` but run it on an image where you’ve added a profile with an extension ICM, then the profile does __not__ get removed. I blame GM & IM, they make no sense here.

That said, on an unrelated note, I can’t make latest GM to accept `+profile "!ICC,*" command to strip everything __but__ the colour profile. It works correctly in IM. Most probably, just a bug in GM. We should be careful, would we like to use it.

jamesgorrie:
>  ... GM to accept +profile "!ICC,*"

I have played around quite a lot with that command and it differs quite a lot dependant on OS too.

paperboyo:
> it differs quite a lot dependant on OS too

Oh, no! That means that ICC profiles are treated by both apps as _files_ instead of image characteristics. It’s convenient when adding and extracting, but definitely not in this instance... At least IM is listing `icc:` properties (like copyright, description, copyright, model - very useful to _identify_ specific profiles) irrespective of if the image had ICC or ICM profile added. GM lists only `Profile-color: [size in bytes]`, though :cry: So GM is useless if you would like to know _what_ actual profile an image has...

@theefer re:identifying specific profiles

theefer:
Now stripping both icm and icc cc @paperboyo 

Added some tests for the colour model detection using a few sample images with and without embedded profiles.

We should consider adding more tests around the `cropImage` function in the future, to check output profile and colour model.

common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  // Optionally apply transforms to the base operation if the colour space
+  // in the ICC profile doesn't match the colour model of the image data
+  private def correctColour(base: IMOperation)(iccColourSpace: Option[String], colourModel: Option[String]) = {
+    (iccColourSpace, colourModel) match {
+      // If matching, all is well, just pass through
+      case (icc, model) if icc == model => base
+      // If no colour model detected, we can't do anything anyway so just hope all is well
+      case (_,   None) => base
```
akash1810: is it worth logging out these cases?

akash1810:
:+1:

common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
paperboyo: > is it worth logging out these cases?

I do think so. It's generally interesting to me what is the composition of the whole db in terms of colour models, colour profiles. Not that (all) the metadata isn't interesting ;)

PR #1282, opened 2015-09-21
fix a couple of label css

akash1810:
# search
## before
label-before

## after
label

# gr-panel
## before
screen shot 2015-09-21 at 14 32 00

## after
screen shot 2015-09-21 at 14 30 18

upload remains as it is.

### DIFF
--- a/kahuna/public/js/edits/labeller.css
+++ b/kahuna/public/js/edits/labeller.css
@@ -19,6 +19,10 @@ ui-labeller-compact {
     background-color: #99ddff;
 }
 
+/*
+label--partial is part of gr-panel.
+TODO refactor gr-panel to use common labeller component
+*/
 .label--partial .label__value,
 .label--partial .label__add,
 .label--partial .label__remove {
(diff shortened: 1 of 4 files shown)


kahuna/public/js/edits/labeller.css
```diff
@@ -17,13 +19,18 @@
     background-color: #99ddff;
 }
 
-.label--partial,
-.label--partial gr-icon {
+/*
+label--partial is part of gr-panel.
+TODO refactor gr-panel to use common labeller component
+*/
+.label--partial .label__value,
```
jamesgorrie: Could we not just apply this to `label--partial`? Why have every building block have it?

akash1810: To win on specificity against https://github.com/guardian/grid/pull/1282/files#diff-493a18df09d616f3d63820529e14407cR33 ... I agree its not very nice.

jamesgorrie: Ah, crapola - this works for now though - let's tidy it up once this is out.

jamesgorrie:
:+1:

PR #1285, opened 2015-09-22
Hides suggestion list when an option has been picked with enter key

tsop14:
Makes pressing enter key consistent with clicking on a result using the mouse. 

This was a problem when the input generated more than one suggestion result e.g. various case combinations appear for the same input ('the guardian' and 'The Guardian') or the input was a substring ('sun' and 'sunday').

Now
labels15

Before
labels13


theefer:
:+1:

paperboyo:
:sparkles:

PR #1286, opened 2015-09-22
ensures selected index is within results list

tsop14:
Now
labels16

Before
labels17


jamesgorrie:
Looks like this might need some `Rx`ing.
:+1:

theefer:
Shouldn't we just reset to 0 when the input changes? (Intuitively it seems that'd make the code simpler too?)

I've had a look on Facebook, Google and Twitter and they all reset any selection index to the default (either no list item selected, or the first one) when the input change.

jamesgorrie:
Nice and neat - :+1: ?

PR #1288, opened 2015-09-22
update gitignore to ignore .json files created in the re-index process.

akash1810:
Ignore `add_alias.json` and `swap_alias.json`.


theefer:
:+1:

PR #1290, opened 2015-09-23
better no rights description

jamesgorrie:
Pointed out that this is more clear as nearly all images have rights, we just don't have a contract that corresponds.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
@@ -99,7 +99,7 @@ case object NoRights
     val restrictions = None
     val name = "No Rights"
     val description =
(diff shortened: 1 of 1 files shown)


akash1810:
Do you thing we need to say they're pay to use? Should they be in the Grid otherwise?

jamesgorrie:
Nope, I wouldn't say so, hopefully this is clear enough to insinuate those.
They should definitely be in the Grid - we don't only have free images?

jamesgorrie:
> I wouldn't say so
This definition came from the picture editors - so they at least understand it.

theefer:
:+1: but let's make sure to keep the Holy Spreadsheet in sync to avoid confusion and salty tears.

akash1810:
:+1:

jamesgorrie:
@seb this one is not holy i.e. it's not in the spreadsheet as this is more of a "remove".

PR #1291, opened 2015-09-23
Add a few category mappings of picdar rights

theefer:
As provided by Jo.

There are a few others (Keystone, ANSA) but it's less clearcut and low numbers so going to leave them out.


jamesgorrie:
:+1:

PR #1292, opened 2015-09-23
extending creative commons model

jamesgorrie:
The current model of creative commons is not extensive enough - this adds:
* `licence`
* `source`
* `creator`
* `contentLink`

And Maps to:
* `credit => licence / source`
* `creator => byline`
* `contentLink => creditLink`

- [ ] Get actual list from RH for licenses (I've included the current commercial licences)
- [ ] Do mappings to metadata

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
@@ -383,7 +383,8 @@ object CommissionedIllustrator {
  )(i => (i.category, i.creator, i.restrictions))
 }
 
-case class CreativeCommons(licence: String, source: String, creator: String, restrictions: Option[String] = None)
+case class CreativeCommons(licence: String, source: String, creator: String, contentLink: String,
+                           restrictions: Option[String] = None)
   extends UsageRights {
     val category = "creative-commons"
     val defaultCost = Some(Free)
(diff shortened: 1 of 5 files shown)


common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
```diff
@@ -396,5 +397,12 @@ case class CreativeCommons(restrictions: Option[String] = None)
   }
 object CreativeCommons {
  implicit val jsonReads: Reads[CreativeCommons] = Json.reads[CreativeCommons]
- implicit val jsonWrites: Writes[CreativeCommons] = UsageRights.defaultWrites
+ implicit val jsonWrites: Writes[CreativeCommons] = (
```
theefer: Why does it need to be explicit now?

jamesgorrie: As the `UsageRights.defaultRights` can only deals with category and restrictions.
I have a branch looking at composing said function (or using partials) which cleans up a lot of this manual labour - I just haven't had time to completely finish / figure it out.

jamesgorrie: We can't use Json.writes[CreativeCommons] either as that then leaves out category et al as it only takes in the parameters of the case class.

theefer:
:+1:

PR #1293, opened 2015-09-23
Log on esresponse

jamesgorrie:
Not sure what this is for - but we can use it for this.


theefer:
I'd rather not have side-effects in case class constructors :-/

Could we log explicitly instead?

PR #1294, opened 2015-09-24
sleep for a second before scrolling.

akash1810:
should hopefully reduce the impact a reindex has on overall performance.


theefer:
:+1:

PR #1297, opened 2015-09-24
script to display percentage completion.

akash1810:
NB: env vars are from https://github.com/guardian/grid-infra/blob/master/platform/elasticsearch/reindex.md

### DIFF
--- a/elasticsearch/reindex/progress.sh
+++ b/elasticsearch/reindex/progress.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+# Usage: [watch -t] ./progress.sh <current index doc count>
+
+DOC_COUNT=$1
+NEW_DOC_COUNT=$(curl --silent $ES_URL:9200/$NEW_INDEX/_count | jq ".count")
+
+echo "($NEW_DOC_COUNT/$DOC_COUNT)*100" | bc -l


elasticsearch/reindex/progress.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+# Usage: [watch -t] ./progress.sh <current index doc count>
+
+DOC_COUNT=$1
```
jamesgorrie: Could we not do:
`OLD_DOC_COUNT=$(curl --silent $ES_URL:9200/$OLD_INDEX/_count | jq ".count")`

and just pass the two index names in?

theefer: Technically the old count is the count at the time you started the ingestion, right?

akash1810: > Technically the old count is the count at the time you started the ingestion, right?

Yes.

jamesgorrie: But it's the old index that's being updated so on each request you will get the new count, which would make this more, or less, accurate dependant on how you want to look at it.

i.e.
If you have 100 images, start reindexing, we add 20 images and we get to 30/120 - 25%. But we are actually 30% on this step. But we are 25% complete of the whole process.

akash1810: > we add 20 images and we get to 30/120

It'll still be 30/100. During re-index, we clone the index from a point in time. The changes are replayed during the syncing stage.

jamesgorrie: Seems the % completion of this step is more useful so automatically setting the `old_index_count` would give you incorrect results.

PR #1298, opened 2015-09-24
adding logging to images that aren't deleted

jamesgorrie:
After mucking around way to much - here's the solution.
There could be a bit of housekeeping around here to express exactly what it is that we're doing, but that's for a rainy day.


theefer:
:+1:

PR #1299, opened 2015-09-24
add more expressive deletion logic to thrall

jamesgorrie:
Just testing some logic out as the old logic threw me (even though I wrote it) and didn't make immediate sense to @lindseydew either. 

@lindseydew - does this make more sense to you? It's a little clogged up due to our complex logic around deletion.

### DIFF
--- a/thrall/app/lib/ElasticSearch.scala
+++ b/thrall/app/lib/ElasticSearch.scala
@@ -86,7 +86,7 @@ object ElasticSearch extends ElasticSearchClient with ImageFields {
 
   def deleteImage(id: String)
                        (implicit ex: ExecutionContext):
-                        Future[Either[ImageDeletedResponse, ImageNotDeletableResponse]] = {
+                        Future[Either[ImageNotDeletableResponse, ImageDeletedResponse]] = {
 
     // We search for the image first as the `prepareDelete` doesn't return whether
     // it has delete anything or not. We need this info further up the chain to know
@@ -106,11 +106,11 @@ object ElasticSearch extends ElasticSearchClient with ImageFields {
           .incrementOnFailure(failedDeletedImages) { case _:ImageNotDeletableResponse => true }
 
         deleteFuture map { deleteResponse =>
-          Left(ImageDeletedResponse(id))
+          Right(ImageDeletedResponse(id))
         } recoverWith {
           case r:ImageNotDeletableResponse => {
             Logger.info(s"Could not delete image $id")
-            Future.successful(Right(r))
+            Future.successful(Left(r))
           }
         }
       }
--- a/thrall/app/lib/ElasticSearch.scala
+++ b/thrall/app/lib/ElasticSearch.scala
@@ -1,11 +1,11 @@
 package lib
 
-import _root_.play.api.libs.json._
+import play.api.libs.json._
+import play.api.Logger
 import com.gu.mediaservice.lib.elasticsearch.{ElasticSearchClient, ImageFields}
 import com.gu.mediaservice.syntax._
 import groovy.json.JsonSlurper
 import lib.ThrallMetrics._
-import org.elasticsearch.action.deletebyquery.DeleteByQueryResponse
 import org.elasticsearch.action.update.{UpdateRequestBuilder, UpdateResponse}
 import org.elasticsearch.action.updatebyquery.UpdateByQueryResponse
 import org.elasticsearch.client.UpdateByQueryClientWrapper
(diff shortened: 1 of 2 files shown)


thrall/app/lib/ThrallMessageConsumer.scala
```diff
@@ -39,21 +39,15 @@ object ThrallMessageConsumer extends MessageConsumer(
   def updateImageUserMetadata(metadata: JsValue): Future[UpdateResponse] =
     withImageId(metadata)(id => ElasticSearch.applyImageMetadataOverride(id, metadata \ "data"))
 
-  def deleteImage(image: JsValue): Future[EsResponse] =
+  // The Unit, Unit is due to the two sied effects
```
theefer: What Unit?

thrall/app/lib/ElasticSearch.scala
```diff
@@ -62,40 +63,56 @@ object ElasticSearch extends ElasticSearchClient with ImageFields {
...
+      missingOrEmptyFilter(identifierField(Config.persistenceIdentifier)),
+      boolFilter.mustNot(termFilter(editsField("archived"), true)),
+      boolFilter.mustNot(termFilter(usageRightsField("category"), "staff-photographer")),
+      boolFilter.mustNot(termFilter(usageRightsField("category"), "contract-photographer")),
+      boolFilter.mustNot(termFilter(usageRightsField("category"), "commissioned-photographer"))
     )
+  )
 
-    val deleteQuery = client
+  private def deleteImageQuery(id: String) =
```
theefer: It's a bit confusing that similarly named `deleteFilterQuery` and `deleteImageQuery` are two completely different things: one returns an ES query object, the other runs an ES  query command.

akash1810: any particular reason for `private`?

thrall/app/lib/ElasticSearch.scala
```diff
@@ -62,40 +63,56 @@ object ElasticSearch extends ElasticSearchClient with ImageFields {
...
+          case 1 => deleteImageQuery(id)
+          case _ => Future.failed(ImageNotDeletableResponse(id))
         }
         deleteFuture
           .incrementOnSuccess(deletedImages)
-          .incrementOnFailure(failedDeletedImages) { case ImageNotDeletable => true }
+          .incrementOnFailure(failedDeletedImages) { case _:ImageNotDeletableResponse => true }
+
+        deleteFuture map { deleteResponse =>
+          Left(ImageDeletedResponse(id))
```
theefer: Normally, `Left` is used for failures and `Right` for success. Is the inversion here intentional?

thrall/app/lib/ThrallMessageConsumer.scala
```diff
@@ -39,21 +39,15 @@ object ThrallMessageConsumer extends MessageConsumer(
...
+  def deleteImage(image: JsValue): Future[Either[ImageDeletedResponse, ImageNotDeletableResponse]] =
     withImageId(image) { id =>
-      // if we cannot delete the image as it's "protected", succeed and delete
-      // the message anyway.
-      ElasticSearch.deleteImage(id).map {
-        case r: DeleteByQueryResponse => {
+      ElasticSearch.deleteImage(id).map { deleteResponse =>
+        deleteResponse.left.map { imageDeletedResponse =>
```
theefer: Not entirely sure what the point of using an `Either` is if we only ever care about the success case?

theefer:
Not convinced it's clearer this way, but it might be just me :-)

thrall/app/lib/ElasticSearch.scala
jamesgorrie: Nope, just being goofy.

thrall/app/lib/ElasticSearch.scala
jamesgorrie: I would only want people to access `deleteImage` as a client.

thrall/app/lib/ThrallMessageConsumer.scala
jamesgorrie: Old comment - removed

thrall/app/lib/ThrallMessageConsumer.scala
jamesgorrie: The reason being that we definitely care that the image was allowed to be deleted i.e. an image with that `id` and `deletableFilter` exists and was deleted.

We were using a future before, but that could fail as the request was unsuccessful i.e. Elastic broke, or because we decided it would fail because it wasn't deltable.

We also use to care if the image was not deleted so we could log it - but I moved the logging into eh ES client as that's where the rest of the logic if. I suppose this could be a `Future[Option[ImageDeletedResponse]]` as we no longer care about the failure of an image to be deleted.

jamesgorrie:
@theefer -> Moved it to an `Option`, if this still isn't clear, happy to close as it's not important.

PR #1300, opened 2015-09-24
make reindex log more verbose by displaying percentage completion.

akash1810:
Alternate (but less precise) version of https://github.com/guardian/grid/pull/1297. However, this PR means one less script to run when running a re-index.


jamesgorrie:
:+1:

PR #1301, opened 2015-09-24
Usage App

kenoir:
Polls CAPI for changes -> Extracts Image elements from content -> Store usages in DynamoDB


usage/app/lib/Config.scala
```diff
@@ -19,4 +19,7 @@ object Config extends CommonPlayAppProperties with CommonPlayAppConfig {
   val rootUri = services.metadataBaseUri
   val kahunaUri = services.kahunaBaseUri
   val loginUriTemplate = services.loginUriTemplate
+
+  val pollerTable = "media-service-DEV-ContentPoller"
```
theefer: Make it a property?

theefer:
Nice! :+1:

kenoir:
Maybe this should not refer to image, but media for future extensibility.

jamesgorrie:
@kenoir - That's what I was doing here:
https://github.com/guardian/grid/pull/1322/files#diff-254da095d43c90244fef48d415c92d0eR3

usage/app/lib/UsageStream.scala
```diff
@@ -0,0 +1,37 @@
...
+      createUsage(createUsageId(imageElement.id, index), content.id, imageElement)
+    }})
+
+  def createStatus(container: ContentContainer) = container match {
+    case PreviewContentItem(_,_) => PendingUsageStatus(container.lastModified)
+    case LiveContentItem(_,_) => PubishedUsageStatus(container.lastModified)
+  }
+
+  def createUsage(id: String, group: String, imageElement: Element) =
+    MediaUsage(id, group, imageElement)
```
kenoir: image -> media

usage/Global.scala
```diff
@@ -12,11 +12,20 @@ import com.gu.mediaservice.lib.play.RequestLoggingFilter
 
 object Global extends WithFilters(RequestLoggingFilter, new GzipFilter) with GlobalSettings {
 
+  // TODO: We shouldn't need to lazy val this ideally
+  lazy val usageRecorder = UsageRecorder.observable.subscribe(println(_))
```
akash1810: stray `println`?

usage/app/model/UsageStatus.scala
```diff
@@ -0,0 +1,14 @@
...
+
+trait UsageStatus {
+  override def toString = this match {
+    case _:PendingUsageStatus => "pending"
+    case _:PubishedUsageStatus => "published"
+  }
+}
+
+case class PendingUsageStatus() extends UsageStatus
+case class PubishedUsageStatus() extends UsageStatus
```
theefer: I think you typo'd `PubicUsageStatus`.

usage/app/lib/UsageRecorder.scala
```diff
@@ -0,0 +1,31 @@
...
+  val usageStream = UsageStream.observable
+
+  def recordUpdates(usageGroup: UsageGroup) = {
+    UsageRecordTable.matchUsageGroup(usageGroup).map(dbUsageGroup => {
+
+      val deletes = (dbUsageGroup.usages -- usageGroup.usages).map(mediaUsage => {
+        UsageRecordTable.delete(mediaUsage.grouping, mediaUsage.usageId)
+      })
+
+      val updates = usageGroup.usages.map(UsageRecordTable.update(_))
```
theefer: So these are upserts (new and updates)?

usage/app/model/MediaUsage.scala
```diff
@@ -0,0 +1,59 @@
...
+      createUsageId(media.id, status, index),
+      grouping,
+      media.id,
+      "web",
+      "image",
+      status,
+      lastModified
+    )
+
+  // The purpose of this hash is to obfuscate the usage_id
```
theefer: What's the reason for obfuscating? Better hash distribution?

usage/Global.scala
```diff
@@ -12,11 +12,20 @@ import com.gu.mediaservice.lib.play.RequestLoggingFilter
...
+
   override def beforeStart(app: Application): Unit = {
     LogConfig.init(Config)
   }
 
   override def onStart(app: Application) {
     UsageApi.keyStore.scheduleUpdates(Akka.system(app).scheduler)
+
+    usageRecorder
```
theefer: Can we have a comment here explaining how this hack links with the lazy val?

usage/app/lib/UsageRecorder.scala
kenoir: Yes. At the moment there's not a need to distinguish.

common-lib/src/main/scala/com/gu/mediaservice/lib/argo/ArgoHelpers.scala
```diff
@@ -24,6 +24,27 @@ trait ArgoHelpers extends Results {
     serializeAndWrap(response, Ok)
   }
 
+  def respondCollections[T](
+    uri: Option[URI] = None,
+    data: Map[String, Seq[T]],
+    links: List[Link] = Nil
+  ) (implicit writes: Writes[T]): Result = {
+    val collections = data.map({ case (key, collection) => {
+      key -> CollectionReponse(
```
theefer: Embedded entities [must](https://github.com/argo-rest/spec#uri) have a `uri` parameter. Otherwise they are not embedded resources, if they don't have a `uri`. What's the main reason for wrapping them with data/length in this case?

kahuna/public/js/components/gr-image-usage/gr-image-usage.html
```diff
@@ -0,0 +1,35 @@
+<div ng:show="ctrl.usage.pending || ctrl.usage.published">
```
theefer: In future I suggest keeping UI changes separate from back-end work (e.g. on a separate branch based on this one), in case we want to merge/release/revert them independently. Unless there was a reason to do this here?

usage/conf/routes
```diff
@@ -1,5 +1,7 @@
 GET     /                                               controllers.UsageApi.index
 
+GET     /usage/media/:mediaId                           controllers.UsageApi.forMedia(mediaId: String)
```
theefer: `usages`? (usually plural nouns in the rest of the APIs)

usage/app/controllers/UsageApi.scala
```diff
@@ -22,10 +22,17 @@ object UsageApi extends Controller with ArgoHelpers {
...
     respond(indexData, indexLinks)
   }
-
   def index = Authenticated { indexResponse }
+
+  def forMedia(mediaId: String) = Authenticated.async {
+    val usagesFuture = UsageTable.queryByImageId(mediaId)
+
+    usagesFuture.map[play.api.mvc.Result](UsageResponse.buildCollection).recover {
+      case NoItemFound => respond(false)
```
theefer: What does that translate as in terms of HTTP response?

usage/app/lib/UsageMetrics.scala
```diff
@@ -0,0 +1,9 @@
+package lib
+
+import com.gu.mediaservice.lib.metrics.CloudWatchMetrics
+import Config.{awsCredentials, stage}
+
+object UsageMetrics extends CloudWatchMetrics(s"$stage/Thrall", awsCredentials) {
```
theefer: Do you mean `$stage/Usage`?

common-lib/src/main/scala/com/gu/mediaservice/lib/argo/ArgoHelpers.scala
kenoir: Convenience for the client. They will get a `uri` if there's a reason to have separate endpoints for pending and published sets.

kahuna/public/js/components/gr-image-usage/gr-image-usage.html
kenoir: Just fiddling in this branch. Agreed UI work should be in another branch here, but the project split makes them effectively releasable separately. I think @akash1810 is working on UI stuff at the moment, maybe we could cherry pick these commits into another branch and remove them here.

akash1810: Yep to both working on the UI and moving it off this branch.

usage/conf/routes
kenoir: Guess I wasn't thinking about presenting a list of all usages on `/usage` so don't care about pluralising in the REST style. But come to think of it we might actually might do that so it's probably worth it.

common-lib/src/main/scala/com/gu/mediaservice/lib/argo/ArgoHelpers.scala
theefer: OK, just note that theseus will *not* understand these as embedded resources (or parse them as `Resource` objects) because they're not valid resources according to the spec. I guess it's just data that happens to look similar to a subresource and possibly helps future proofing (although could be slightly confusing).

usage/app/controllers/UsageApi.scala
kenoir: 200 OK

```json
{"data":{}}
```

usage/conf/usage.conf
```diff
@@ -0,0 +1,28 @@
+env USER=media-service
+env USER_HOME=/home/media-service
+env JAR=/home/media-service/app.jar
+
+# Workaround for https://github.com/aws/aws-sdk-java/issues/123
+env JAVA_OPTS="-XX:MaxRAMFraction=2 -XX:InitialRAMFraction=2 -XX:+PrintGCDetails -XX:+PrintGCDateStamps -Xloggc:gc.log"
```
theefer: Just curious how and why this differs from the other "conf" run scripts? e.g. no `-Dfile.encoding=UTF-8` etc?

kenoir: Borrowed this from one of the Ophan services. It sets the RAM usage for the JVM as a fraction rather than a hard coded value.

I'm not sure of the consequence of removing the default file encoding here, so I suspect we _should_ be setting it.

theefer: I'd probably supersticiously set it, yeah. re memory, whichever is best I don't mind, but might be good to have consistency across the different run scripts, or else document why they differ?

kenoir: I think they should all use the ratio method here TBH.

On Thu, 22 Oct 2015 at 13:55 Sébastien Cevey <notifications@github.com>
wrote:

> In usage/conf/usage.conf
> <https://github.com/guardian/grid/pull/1301#discussion_r42743845>:
>
> > @@ -0,0 +1,28 @@
> > +env USER=media-service
> > +env USER_HOME=/home/media-service
> > +env JAR=/home/media-service/app.jar
> > +
> > +# Workaround for https://github.com/aws/aws-sdk-java/issues/123
> > +env JAVA_OPTS="-XX:MaxRAMFraction=2 -XX:InitialRAMFraction=2 -XX:+PrintGCDetails -XX:+PrintGCDateStamps -Xloggc:gc.log"
>
> I'd probably supersticiously set it, yeah. re memory, whichever is best I
> don't mind, but might be good to have consistency across the different run
> scripts, or else document why they differ?
>
> —
> <https://github.com/guardian/grid/pull/1301/files#r42743845>.
>

-- 

in England Number 908396

kenoir:
Merging as this app is disjoint from the others.

PR #1302, opened 2015-09-24
Caret bug fix

tsop14:
Caret no longer jumping to the start/end of the input text when scrolling through suggestion list using nav keys. 

Default behaviour: down-arrow keydown moves the caret to the end of input text; up-arrow keydown moves the caret to the start of input text. This default behaviour is prevented when suggestion list is active.   

Now
labels21

Before
labels20


theefer:
:+1:

PR #1303, opened 2015-09-24
No original link on image crop page

theefer:
# Before
screenshot-2015-09-24t16 00 38

# After
screenshot-2015-09-24t16 01 00


akash1810:
:+1:

PR #1305, opened 2015-09-25
order by uploadTime rather than uploadedBy.

akash1810:
This is a change to the query used during a reindex - ordering by `uploadTime` means we process images chronologically.


tsop14:
'uploadTime' could also mean, the time it took to complete the upload process. Would 'timeUploaded' or 'timeOfUpload' be clearer?

theefer:
:+1:

akash1810:
> 'uploadTime' could also mean, the time it took to complete the upload process. Would 'timeUploaded' or 'timeOfUpload' be clearer?

`timeOfUpload` would be a little clearer, however `uploadTime` its part of the [model](https://github.com/guardian/grid/blob/master/media-api/app/lib/ImageResponse.scala#L186) and would be painful to change.

PR #1306, opened 2015-09-25
Correctly convert thumbnail colour space to sRGB

theefer:
This applies the same conversions as we do with crops to thumbnails.

863e0b9694bc31edbb4baf16db730dbe66b07da1 is a good example of where it can go horribly wrong currently.

# TODO

- [x] Review desired quality and image refinement flags with @paperboyo 
- [x] Review performance on TEST vs current


paperboyo:
> Record image colourModel in fileMetadata

Cool! What do we use to read that info? GM? Then, it does not give us an option to also read colour profile name (and switching from GM>IM, which does, is probably to much). But if not GM, maybe we could read and store colour profile name as well? Also, if we ever surface that info in the UI, it should say that it pertains to the original image (cause the crops will be sRGB/whatever RGB).

If we just apply current pipeline, the thumbs will grow by 2,6KB*no. of thumbs. We can alleviate that by converting them to e.g. AdobeRGB or TinysRGB (~600b).

> desired quality and image refinement flags
It's always gonna ba a compromise between quality and filesize, but my (limited) tests show that we may as well aim for, maybe, halving the filesize.

theefer:
> What do we use to read that info? GM?

Yes, it's that `identify -format '%[JPEG-Colorspace-Name]'` command.

> Then, it does not give us an option to also read colour profile name (and switching from GM>IM, which does, is probably to much).

Not sure what you mean by that. Do you mean the ICC `Color space` or `Profile Description`?

> If we just apply current pipeline, the thumbs will grow by 2,6KB*no. of thumbs.

I'm not sure that's true, currently I don't think we touch the profile so the thumbnail contains however much ICC profile information there is in the image (potentially 400KB), AFAIK.

theefer:
But yes we could optimise the embedded profile, as long as it remains visually correct.

paperboyo:
GM, then. As per https://github.com/guardian/grid/pull/1281#issuecomment-142001285 it doesn't give us what IM would give us. A name of the profile (taken from one of the fields IM describes the profile with).

> I'm not sure that's true

Yes, you're right. That's how will grow only the thumbs of images without embedded profiles.

> as long as it remains visually correct

Always!

paperboyo:
> GM, then. As per #1281 (comment) it doesn't give us what IM would give us. A name of the profile (taken from one of the fields IM describes the profile with).

Sorry, I’m an idiot. Ignore that and all above blabbing about storing colour profile identity. I’ve learned that we already do it. Doh!

jamesgorrie:
:+1: to code :+1: to testing perf.

theefer:
I want to merge #1621 into this and then this can go out, likely on Thursday. Latency looks similar to the current one, but with correct colour space conversion and reduced thumbnail file size (often half the size).

PR #1307, opened 2015-09-25
scripts to semi-automate starting all services in dev.

akash1810:
Set of scripts to launch each service with debug ports attached for interactive debugging in IntelliJ.

I've been using these scripts locally and they work fairly well. After sharing them with @tsop14 I thought I'd check them in.

I hadn't been using iTerm profiles before (I was running them one by one), but I've just set them up (plus window arrangements) and starting a dev environment is a lot simpler now - launch iTerm, `CMD + Shift + R` and away you go. :smile: 

Could definitely do the same in tmux, but I'm still a novice when it comes to that...

### DIFF
--- a/dev/00-run-elastic-search.sh
+++ b/dev/00-run-elastic-search.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+cd ../elasticsearch
+source dev-start.sh
--- a/dev/01-run-media-api.sh
+++ b/dev/01-run-media-api.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+cd ../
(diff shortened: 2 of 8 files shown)


jamesgorrie:
Not sure if this is much easier than just running the sbt commands, although the interactive debugging is great. It's also a good start that we can evolve :+1:

akash1810:
The more I think about this PR, the more I want to close it without merging as its prescribing the use of iTerm and IntelliJ for not much benefit (read automation).

I'll close until in favour of getting [run.sh](https://github.com/guardian/grid/blob/master/run.sh) working cross platform.

jamesgorrie:
@akash1810 :+1:

PR #1309, opened 2015-09-28
use elastic's count search type

jamesgorrie:
[As suggested by the ES docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations.html#_returning_only_aggregation_results).

I'd like to monitor performance on this, if it's bad, it's an easy rollback.

### DIFF
--- a/media-api/app/lib/elasticsearch/ElasticSearch.scala
+++ b/media-api/app/lib/elasticsearch/ElasticSearch.scala
@@ -150,7 +150,6 @@ object ElasticSearch extends ElasticSearchClient with SearchFilters with ImageFi
       .map{ response =>
         val buckets = response.getAggregations.getAsMap.get(name).asInstanceOf[InternalTerms].getBuckets
         val results = buckets.toList map (s => BucketResult(s.getKey, s.getDocCount))
-        println(response)
 
         AggregateSearchResults(results, buckets.size)
       }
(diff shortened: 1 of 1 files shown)


media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -144,13 +144,13 @@ object ElasticSearch extends ElasticSearchClient with SearchFilters with ImageFi
...
     search
-      .setFrom(0)
-      .setSize(0)
+      .setSearchType(SearchType.COUNT)
       .executeAndLog("metadata aggregate search")
       .toMetric(searchQueries, List(searchTypeDimension("aggregate")))(_.getTookInMillis)
       .map{ response =>
         val buckets = response.getAggregations.getAsMap.get(name).asInstanceOf[InternalTerms].getBuckets
         val results = buckets.toList map (s => BucketResult(s.getKey, s.getDocCount))
+        println(response)
```
akash1810: stray println?

akash1810:
:+1:

media-api/app/lib/elasticsearch/ElasticSearch.scala
jamesgorrie: Thanks :eyes:

PR #1310, opened 2015-09-28
Label collections

jamesgorrie:
First pass attempt at adding a search for sibling labels.
e.g.
```
image[0].labels = ["hat", "shoes", "wig"]
image[1].labels = ["hat", "shoes", "sandals"]

GET /suggest/labels/hat
{
  "count": 3
  "data": [
        {
            "key": "shoes",
            "count": 2
        },
        {
            "key": "wig",
            "count": 1
        },
        {
            "key": "sandals",
            "count": 1
        }
    ]
}
```

__TODO__
- [x] sort by recently modified
- [x] tidy code
- [ ] test performance (needs to go live)


media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -128,6 +129,43 @@ object ElasticSearch extends ElasticSearchClient with SearchFilters with ImageFi
       }
   }
 
+  def siblingLabelsSearch(siblingLabel: String)(implicit ex: ExecutionContext): Future[AggregateSearchResults] = {
+    val name = "siblingLabels"
+    val lastModifiedField = "lastModified"
+    val labelsField = editsField("labels")
+
+    val sortByDateAggr =
```
jamesgorrie: As much as ordering by `lastModified` puts something at the top of the list when you add a label, another record would usurp it if you edit, say, the metadata. We might want to add the idea of lastModified to each `Resource`? Perhaps something for the argo-rest/spec?

jamesgorrie:
Happy for review!

akash1810:
Looks good :+1: .

Would we need the URI of the label in the response too?

jamesgorrie:
> Would we need the URI of the label in the response too?

Are you suggesting having an array of label resources? Why would that be useful?

akash1810:
> Are you suggesting having an array of label resources?

Yep.

> Why would that be useful?

Not entirely sure - was more thinking out loud.

jamesgorrie:
Labels don't actually have their own resource, they are attached to an image. i.e.
[`/metadata/:id/labels/*label`](https://github.com/guardian/grid/blob/master/metadata-editor/conf/routes#L13).

what would we have the ID set as - and if you did a DELETE against, say, `/metadata/labels/toupee`, would we delete all instances of toupee? I think the problem here is we have a list of suggestions, not resources as we can't interact with them.

akash1810:
Yep, you're right - all valid points!! :smile: 

:+1:

PR #1311, opened 2015-09-29
reaper as a python script.

akash1810:
Currently reaper is hidden away in the cloud formation script, meaning we can't run it locally. This is a simple script that allows reaper to be run locally.

I'd like to use this to further debug the failed deleted images graph spikes we see in PROD when performing a reindex - though the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-scroll.html) suggest the main cause of this problem is related to scrolling.

> The results that are returned from a scroll request reflect the state of the index at the time that the initial search request was made, like a snapshot in time. Subsequent changes to documents (index, update or delete) will only affect later search requests.

(Our reaper instances currently do not have java installed, hence not scala.)


jamesgorrie:
So this isn't being used by CF / our stack yet? Do you think it will, or shall we keep it like the imgopts, a local version of the same thing?

akash1810:
> Do you think it will

I think it should be re-written as a scala app if it will be used. That way it can be built and deployed in the same way as the other apps to reduce complexity.

> a local version of the same thing?

I view it as that - the last commit enforces that too (we're ignoring SSL warnings, which is ok in DEV).

akash1810:
> we're ignoring SSL warnings

Curiously, I didn't have to do this on my personal mac...

jamesgorrie:
> same way as the other apps to reduce complexity

I would argue the small cron job etc is a lot less complex than a play app doing something similar.

jamesgorrie:
Are we logging on failed `DELETE`s?

akash1810:
> Are we logging on failed DELETEs?

Nope. Do we need to, as the media-api does that already?

jamesgorrie:
Good point :+1:

kenoir:
> I would argue the small cron job etc is a lot less complex than a play app doing something similar.

I agree with @jamesgorrie here.

We could perhaps replace the reaper instances with a Lambda function triggered by something like: https://alestic.com/2015/05/aws-lambda-recurring-schedule/

akash1810:
Using aws lambda's would be interesting. Currently, we delete 100 images every minute - as the UTC notifies every 15 minutes, we'd have to increase that number accordingly. That's a different problem though.

theefer:
Nice one!

It'd be nice to move from the script in CF to this one. Should not be too hard to bundle this into an artifact (maybe taking inspiration of the annoying way we bundle ICC profile files) and release via RiffRaff.

(We could have used something like the [grid-batcher](https://github.com/guardian/grid-batcher), which would make it easy to use the hypermedia structure to perform the actions, but might be more tedious to bundle a Node app)

PR #1312, opened 2015-09-29
Unify edits response

jamesgorrie:
I've had to add an action / link / field to two places too many times now, so have unified the response in the API as well as the edits service. This work was need to add more links for the label siblings UI.


akash1810:
Has the shape of the api changed now? From the diff, Its difficult to tell what's new - I can see `Transformers` has been removed, but not much more than that... :cry:

akash1810:
:+1:

PR #1313, opened 2015-09-29
centre crop thumbnail.

akash1810:
Also removing a `ul` with a single `li` to keep the DOM (slightly) lighter.

before
screen shot 2015-09-29 at 16 54 24

after
screen shot 2015-09-29 at 16 54 03


jamesgorrie:
That's been annoying me for ages - :boom: :+1:

PR #1314, opened 2015-09-29
Automatically add labels to uploads

tsop14:
Allows users to create a list of labels that will be automatically added to every upload. 

These labels are stored in local storage. 

labels22

TODO

- [x] Allow users to add more than one label at a time to their list of preset labels using a comma (to make consistent with original add label)
- [x] Move presetLabeller files out of prompt folder


jamesgorrie:
Not sure about the default labels thing - could itbe something 100% explicit e.g. "Apply `LABEL INPUT` to label all uploads". It's subjective, but we can run it past the Observer bunch tomorrow too.

tsop14:
Yes, should be made clearer. I think it is good to have the labels in the middle but it also needs to look sensible when there are no labels. 

:smile:  screenshot-media local dev-gutools co uk 2015-09-29 18-01-03 


:disappointed:  screenshot-media local dev-gutools co uk 2015-09-29 18-07-45 

@alastairjardine

jamesgorrie:
@tsop14 you could not hide the input box, showing the `e.g. culture` if there aren't any labels?

kahuna/public/js/components/gr-add-preset-label/gr-add-preset-label.css
```diff
@@ -0,0 +1,31 @@
...
+
+.gr-add-preset-label__form {
+    display: flex;
+}
+
+.gr-add-preset-label__form__input {
+    width: 150px;
+}
+
+.gr-add-preset-label__form__buttons {
```
jamesgorrie: BEM is only really effective with one layer deep, otherwise it almost destroys it's own usefulness.

You could have something like `gr-add-preset-label__form-input` or to be more descriptive `gr-add-preset-label__label-input`.

jamesgorrie: The reason being, it's relationship to the `form` is not important.

kahuna/public/js/components/gr-add-preset-label/gr-add-preset-label.css
```diff
@@ -0,0 +1,31 @@
...
+    display: flex;
+    height: 25px;
+    padding: 5px;
+}
+
+gr-add-preset-label:hover {
+    color: white;
+}
+
+gr-add-preset-label .small {
```
jamesgorrie: Is the class `.small` used anywhere else? If not could we make it a `--small` modifier.

kahuna/public/js/components/gr-add-preset-label/gr-add-preset-label.css
```diff
@@ -0,0 +1,31 @@
...
+    border: none;
+    padding: 0;
+}
+
+.gr-add-preset-label__form {
```
jamesgorrie: We could just use `flex-container`?

kahuna/public/js/components/gr-add-preset-label/gr-add-preset-label.css
```diff
@@ -0,0 +1,31 @@
...
+    display: flex;
+}
+
+.gr-add-preset-label__form__input {
+    width: 150px;
+}
+
+.gr-add-preset-label__form__buttons {
+    display: flex;
+    font-size: 13px;
```
jamesgorrie: We might want to make a `small-text` class?

kahuna/public/js/components/gr-add-preset-label/gr-add-preset-label.html
```diff
@@ -0,0 +1,38 @@
+<button ng:class="{'small': ctrl.grSmall}"
```
jamesgorrie: As we use this in only one place - we don't really use the small do we?

kahuna/public/js/upload/prompt/preset-labeller.js
```diff
@@ -0,0 +1,43 @@
...
+    return {
+        restrict: 'E',
+        controller: 'PresetLabellerCtrl',
+        controllerAs: 'ctrl',
+        bindToController: true,
+        template: template
+    };
+}]);
+
+
```
jamesgorrie: Is there really a need for a gr-add-preset-label and this? We haven't used it anywhere else, and don't plan on it do we?

tsop14:
labels23

kahuna/public/js/components/gr-preset-labels/gr-preset-labels.js
```diff
@@ -0,0 +1,75 @@
...
+
+            if (newPresetLabelList) {
+                save(newPresetLabelList);
+            }
+        };
+
+        ctrl.cancel = reset;
+
+        ctrl.removePresetLabel = labelToRemove => {
+            let updatedPresetList = ctrl.presetLabels.filter( label => label !== labelToRemove);
```
jamesgorrie: We can use `const` for a lot of these as they aren't mutated.

kahuna/public/js/services/preset-label.js
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    return {
+        getLabels,
+        setLabels
+    };
+
+}]);
+
+export default presetLabelService;
```
jamesgorrie: If we used [`Set`s](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Set) here, adding and removing would be a little easier (we could do `addLabel`, `removeLabel` as opposed to doing the whole set).

kahuna/public/js/upload/jobs/upload-jobs.js
```diff
@@ -4,19 +4,39 @@ import '../../preview/image';
...
+            $scope,
+            $window,
+            apiPoll,
+            track,
+            imageService,
+            labelService,
+            presetLabelService) {
 
     var ctrl = this;
+    let presetLabels = presetLabelService.getLabels();
```
jamesgorrie: `const`

jamesgorrie:
A small comment around how we add / remove labels vs setting them - but we can :bike::house: that for now... :+1:

kahuna/public/js/components/gr-preset-labels/gr-preset-labels.js
```diff
@@ -0,0 +1,67 @@
...
+        ctrl.cancel = reset;
+
+        ctrl.removePresetLabel = labelToRemove => {
+            presetLabelService.removeLabel(labelToRemove);
+            ctrl.presetLabels = presetLabelService.getLabels();
+        };
+
+        function save(labels) {
+            ctrl.adding = true;
+            ctrl.active = false;
```
jamesgorrie: `reset` already does this?

kahuna/public/js/components/gr-preset-labels/gr-preset-labels.html
```diff
@@ -0,0 +1,56 @@
...
+
+<div class="add-preset-label">
+    <button ng:click="ctrl.active = true"
+            ng:disabled="ctrl.adding"
+            ng:if="!ctrl.active"
+            title="Add a new label">
+
+        <gr-icon ng:class="{'spin': ctrl.adding}">add_box</gr-icon>
+        <span>
+           <span ng:show="ctrl.adding">Adding...</span>
```
akash1810: why nested `span`?

kahuna/public/js/services/preset-label.js
```diff
@@ -0,0 +1,39 @@
...
+
+    const presetLabelsKey = 'preset labels';
+
+    function getLabels() {
+        return JSON.parse($window.localStorage.getItem(presetLabelsKey));
+    }
+
+    function addLabels(newLabels) {
+        let labels = new Set(getLabels());
+        newLabels.map( newLabel => labels.add(newLabel) );
```
jamesgorrie: It's generally best practice to use `forEach` when we aren't expecting the result of a `map`

kahuna/public/js/services/preset-label.js
```diff
@@ -0,0 +1,39 @@
...
+                            ['$window', function ($window) {
+
+    const presetLabelsKey = 'preset labels';
+
+    function getLabels() {
+        return JSON.parse($window.localStorage.getItem(presetLabelsKey));
+    }
+
+    function addLabels(newLabels) {
+        let labels = new Set(getLabels());
```
jamesgorrie: This can be `const`. When you use the `add` / `remove` it's not actually mutating the object (which is what makes these quite nice).

jamesgorrie:
A few small points of info :+1:

kahuna/public/js/services/preset-label.js
```diff
@@ -0,0 +1,39 @@
+import angular from 'angular';
+
+var presetLabelService = angular.module('kahuna.services.presetLabel', []);
+
+presetLabelService.factory('presetLabelService',
+                            ['$window', function ($window) {
+
+    const presetLabelsKey = 'preset labels';
```
akash1810: Keys with a space scare me :fearful:...

akash1810:
:+1:

kahuna/public/js/services/preset-label.js
jamesgorrie: It's a good point, `presetLabels` would more be our convention

theefer:
Nice! Do we know if it's being used?

PR #1316, opened 2015-09-30
always have metadata in edits response

jamesgorrie:
We should always have metadata in the response as it's a resource, so we won't be able to `PUT` is it's just not there.


kenoir:
:+1:

akash1810:
:+1:

PR #1318, opened 2015-10-02
catch no-link error for related labels

jamesgorrie:
Is the link is missing - we don't care.
With help from @kenoir and his `Rx`yness.


akash1810:
:+1:

PR #1319, opened 2015-10-02
a few bits of styling on the results tolbar

jamesgorrie:
* Got rid of some superfluous css
* used the generic CSS
* Made related labels not break

[Easy to read PR](https://github.com/guardian/grid/pull/1319/files?w=1)

# All bust up
results-toolbar-boken

# Wow!
results-toolbar-fix

### DIFF
--- a/kahuna/public/stylesheets/main.css
+++ b/kahuna/public/stylesheets/main.css
@@ -904,10 +904,6 @@ input.search-query__input {
     padding: 0 10px;
 }
 
-.results-toolbar-item--split {
-    margin-right: auto;
-}
-
 .results-toolbar-item--left {
     border-left: 0;
 }
--- a/kahuna/public/stylesheets/main.css
+++ b/kahuna/public/stylesheets/main.css
@@ -879,13 +879,20 @@ input.search-query__input {
     position: fixed;
     z-index: 2;
 }
+
+.results-toolbar__right {
+    margin-left: auto;
+    line-height: 35px;
+}
+
 .results-toolbar-item {
     vertical-align: middle;
-    border-width: 0px 1px;
+    border-width: 0 1px;
     border-color: #565656;
     border-style: solid;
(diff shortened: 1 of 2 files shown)


akash1810:
Horizontal scrolling in a responsive design :fearful: Could we make suggested labels a drop menu?

jamesgorrie:
> Horizontal scrolling in a responsive design
Why not? Swipe works fine with it? We have tested it on the live site with the nav and it's quite intrinsic.

jamesgorrie:
The reason for horizontal scrolling is that we don't know how long the list is.
If we are going to have three - we don't need to hide them until the screen is really small - but you would need to if you had 15.

The ultimate solution is to have the excess labels go into a dropdown, but I can't find a lib that does that and it would take a while to write, essentially this is what horizontal scroll is.

jamesgorrie:
This is what I mean by dynamically hiding menu-items / labels behind a dropdown:
dissappearing-menu-items

kahuna/public/stylesheets/main.css
```diff
@@ -942,15 +944,19 @@ input.search-query__input {
     height: 36px;
 }
 
-/* TODO: remove this once eveything is in the action-bar */
-.results-count {
-    float: left;
-    padding: 0 10px;
+.related-labels {
+    line-height: 35px;
```
akash1810: Bit annoying to repeat `line-height` in all the toolbar items.

Can we set `line-height: 35px` on `results-toolbar` then

```css
.related-labels {
  line-height: inherit;
}
```

akash1810:
:+1: 

less css the better!

kahuna/public/stylesheets/main.css
jamesgorrie: Annoyingly not, line-height with px / united values is poorly supported or adhered to.

PR #1320, opened 2015-10-02
DRY-ing out.

akash1810:
Looks scary as its kind of self referential... it compiles though :dancers:

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
@@ -13,11 +13,12 @@ sealed trait UsageRights {
   val defaultRestrictions: Option[String] = None
   val caution: Option[String] = None
 }
+
 object UsageRights {
   val defaultWrites: Writes[UsageRights] = (
     (__ \ "category").write[String] ~
-    (__ \ "restrictions").writeNullable[String]
-  )(u => (u.category, u.restrictions.orElse(u.defaultRestrictions)))
+      (__ \ "restrictions").writeNullable[String]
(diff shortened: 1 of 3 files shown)


akash1810:
Future TODO:
- [ ] do the same for [all strings](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala#L33-L50).

common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
```diff
@@ -60,24 +60,24 @@ object UsageRights {
...
-        case _                           => None
+        case "chargeable"                       => json.asOpt[Chargeable]
+        case "agency"                           => json.asOpt[Agency]
+        case "commissioned-agency"              => json.asOpt[CommissionedAgency]
+        case "PR Image"                         => json.asOpt[PrImage]
+        case "handout"                          => json.asOpt[Handout]
+        case "screengrab"                       => json.asOpt[Screengrab]
+        case "guardian-witness"                 => json.asOpt[GuardianWitness]
+        case "social-media"                     => json.asOpt[SocialMedia]
+        case "obituary"                         => json.asOpt[Obituary]
```
jamesgorrie: Could we not do all of them while we're here?

akash1810: Yep...though now theres loads of whitespace changes in the diff :crying_cat_face:

jamesgorrie:
Fabulous :+1:

theefer:
Nice, makes uses a lot more discoverable at compilation-time/type-level.

PR #1322, opened 2015-10-02
basic routes for usage

jamesgorrie:
__WIP__

Adding the basic usage routes so we know what the shape of the API should be.

### DIFF
--- a/usage/conf/routes
+++ b/usage/conf/routes
@@ -1,5 +1,8 @@
 GET     /                                               controllers.UsageApi.index
 
+GET     /usage/media/:mediaId                           controllers.UsageApi.forMedia(mediaId: String)
+GET     /usage/content/:contentId                       controllers.UsageApi.forContent(contentId: String)
+
 # Management
 GET     /management/healthcheck                         com.gu.mediaservice.lib.management.Management.healthCheck
 GET     /management/manifest                            com.gu.mediaservice.lib.management.Management.manifest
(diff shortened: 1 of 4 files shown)


usage/conf/routes
```diff
@@ -1,5 +1,9 @@
 GET     /                                               controllers.UsageApi.index
 
+GET     /usage/image/:mediaId                           controllers.UsageApi.forMedia(mediaId: String)
```
jamesgorrie: For the Grid

usage/conf/routes
```diff
@@ -1,5 +1,9 @@
...
+GET     /usage/article/:articleId                       controllers.UsageApi.forArticle(articleId: String)
+GET     /usage/date/:date                               controllers.UsageApi.forDate(date: String)
```
jamesgorrie: For RCS

jamesgorrie: THis could probably be an aggregation of CAPI vs this API unless we want print in it?

kenoir:
Good way to make sure we meet in the middle is trying to put usage records into the DB via the API.

jamesgorrie:
^ Explain more to me on Monday

usage/conf/routes
```diff
@@ -1,5 +1,8 @@
 GET     /                                               controllers.UsageApi.index
 
+GET     /usage/media/:mediaId                           controllers.UsageApi.forMedia(mediaId: String)
+GET     /usage/content/:contentId                       controllers.UsageApi.forContent(contentId: String)
```
akash1810: What's this route for?

jamesgorrie: TO query via a content ID or a media ID. We need to gather more requirements for this.

kenoir:
Merged into capi-poller branch.

PR #1324, opened 2015-10-05
Allow deleting of images without an export.

akash1810:
These changes will allow any image that doesn't have an export to be manually deleted via a DELETE request to the media-api.

This change adds the DELETE action to an image response in the media-api iff there are no exports.

Reaper is not affected as `persisted=false` will still [omit `archived`, `staff-photographer` and `picdarUrn` images](https://github.com/guardian/grid/blob/aa-api-allow-delete-from-photographer/media-api/app/lib/elasticsearch/SearchFilters.scala#L100-L107).

### DIFF
--- a/media-api/app/lib/ImageResponse.scala
+++ b/media-api/app/lib/ImageResponse.scala
@@ -96,9 +96,7 @@ object ImageResponse extends EditsResponse {
 
     val links = imageLinks(id, secureUrl, withWritePermission, valid)
 
-    val noExports = persistenceReasons.isEmpty || !persistenceReasons.contains("exports")
-
-    val isDeletable = noExports && withDeletePermission
+    val isDeletable = !persistenceReasons.contains("exports") && withDeletePermission
 
     val actions = imageActions(id, isDeletable, withWritePermission)
 
--- a/media-api/app/lib/ImageResponse.scala
+++ b/media-api/app/lib/ImageResponse.scala
@@ -96,7 +96,12 @@ object ImageResponse extends EditsResponse {
 
     val links = imageLinks(id, secureUrl, withWritePermission, valid)
 
-    val actions = imageActions(id, isPersisted, withWritePermission, withDeletePermission)
+    val isDeletableOnPersistence = persistenceReasons.isEmpty ||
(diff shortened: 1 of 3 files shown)


jamesgorrie:
But this will keep it with the state of `retained`?

jamesgorrie:
i.e. we have two states, `retained` and `deletable`?

thrall/app/lib/ElasticSearch.scala
```diff
@@ -70,10 +69,7 @@ object ElasticSearch extends ElasticSearchClient with ImageFields {
       andFilter(
         missingOrEmptyFilter("exports"),
         missingOrEmptyFilter(identifierField(Config.persistenceIdentifier)),
-        boolFilter.mustNot(termFilter(editsField("archived"), true)),
-        boolFilter.mustNot(termFilter(usageRightsField("category"), StaffPhotographer.category)),
-        boolFilter.mustNot(termFilter(usageRightsField("category"), ContractPhotographer.category)),
-        boolFilter.mustNot(termFilter(usageRightsField("category"), CommissionedPhotographer.category))
+        boolFilter.mustNot(termFilter(editsField("archived"), true))
```
theefer: I was wondering if we should also allow deleting archived images, and perhaps even images with a picdarUrn?

akash1810: Updated now.

media-api/app/lib/ImageResponse.scala
```diff
@@ -96,7 +96,11 @@ object ImageResponse extends EditsResponse {
 
     val links = imageLinks(id, secureUrl, withWritePermission, valid)
 
-    val actions = imageActions(id, isPersisted, withWritePermission, withDeletePermission)
+    val noExports = persistenceReasons.isEmpty || !persistenceReasons.contains("exports")
```
jamesgorrie: Why does `noExports` = all the persistence reasons?

akash1810: Because I like to make things complicated... Updated with more clarity.

theefer:
:+1: but in the absence of tests please quintuple check that it does what you think and doesn't allow deleting images with exports.

PR #1326, opened 2015-10-05
Match truncated for the Gu/Obs bylines

### LINKED ISSUE #1325: Match photographers even if byline is truncated
e.g. 38e4b6756d0c13014a35f35f760a42042627658a has "Christopher Thomond for The Gua" due to the limit in length of 32 chars for the byline field. Other of his pictures have an untruncated longer by-line, so the truncation is probably due to an app rewriting the metadata before we get the file.

1. Can we get the full untruncated data from somewhere, as hinted by @paperboyo elsewhere? A quick exiftool dump of the metadata doesn't show the full version in any other field unfortunately.
2. Can we detect truncated for the GU bylines optimistically?

theefer:
Fixes #1325.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/ExtractGuardianCreditFromByline.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/ExtractGuardianCreditFromByline.scala
@@ -5,12 +5,29 @@ import com.gu.mediaservice.model.ImageMetadata
 object ExtractGuardianCreditFromByline extends MetadataCleaner {
 
   val BylineForTheGuardian = """(?i)(.+) for the (Guardian|Observer)[.]?""".r
+  val BylineForTheTrunc = """(?i)(.+) for the (.+)[.]?""".r
+
+  val GuardianPrefix = Prefix("Guardian")
+  val ObserverPrefix = Prefix("Observer")
 
   override def clean(metadata: ImageMetadata): ImageMetadata = metadata.byline match {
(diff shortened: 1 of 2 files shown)


akash1810:
:+1:

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/ExtractGuardianCreditFromByline.scala
```diff
@@ -5,12 +5,29 @@ import com.gu.mediaservice.model.ImageMetadata
...
+
+  val GuardianPrefix = Prefix("Guardian")
+  val ObserverPrefix = Prefix("Observer")
 
   override def clean(metadata: ImageMetadata): ImageMetadata = metadata.byline match {
     case Some(BylineForTheGuardian(byline, org)) => {
       val orgName = org.toLowerCase.capitalize
       metadata.copy(byline = Some(byline), credit = Some(s"The $orgName"))
     }
+    // Catch truncated bylines (EXIF allows 32 chars only)
```
paperboyo: Technically, that's IPTC's legacy `By-line` field. In EXIF it's called `Artist` and does not have said restrictions. Minor point, obv.

theefer: Oops you're right, my typo.

theefer: Patched on master, thanks.

PR #1330, opened 2015-10-06
Composite

jamesgorrie:
We need something for composites so we can retire the old cost model.
This will hopefully leave us in a situation where we can just split the string later. It would be great to have references to the images used too, but that's going to need quite a lot of off platform integration, perhaps with the mighty skills of @blishen.

* Added composite
* Added examples to placeholder
* Small code tidy

- [x] Get approval & wording from Robert H and Jo B

composite

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
@@ -80,7 +80,8 @@ object Mappings {
     "creator" -> nonAnalyzedString,
     "licence" -> nonAnalyzedString,
     "source" -> nonAnalyzedString,
-    "contentLink" -> nonAnalyzedString
+    "contentLink" -> nonAnalyzedString,
+    "suppliers" -> standardAnalysedString
   )
 
   val exportsMapping =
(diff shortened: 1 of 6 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -80,7 +80,8 @@ object Mappings {
     "creator" -> nonAnalyzedString,
     "licence" -> nonAnalyzedString,
     "source" -> nonAnalyzedString,
-    "contentLink" -> nonAnalyzedString
+    "contentLink" -> nonAnalyzedString,
+    "suppliers" -> standardAnalysedString
```
akash1810: does this mean a re-index is needed for rollout?

metadata-editor/app/model/UsageRightsProperty.scala
```diff
@@ -31,69 +34,76 @@ object UsageRightsProperty {
...
       photographerField(externalPhotographersMap, "publication")
     )
-
-    case _:CommissionedPhotographer => List(
+    case _:ContractPhotographer =>
+    List(
       publicationField(false),
-      photographerField
+      photographerField(
+        Random.shuffle(contractedPhotographers)
```
akash1810: nice

akash1810:
:+1:

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
jamesgorrie: Nope, just adding it to the mapping

PR #1335, opened 2015-10-08
Suggest parent label

jamesgorrie:
As a way to discover that we have related labels.
Off the back of conversations with G2 & Observer (people who use folders etc).

related-label-select 2

### DIFF
--- a/kahuna/public/js/search/results.html
+++ b/kahuna/public/js/search/results.html
@@ -35,7 +35,7 @@
                             <gr-datalist class="related-labels__datalist" gr:search="ctrl.labelSearch(q)">
                                 <!-- TODO: autogrow this element -->
                                 <input class="related-labels__parent" type="text"
-                                       placeholder="e.g. culture"
+                                       placeholder="e.g. observer"
                                        ng:model="ctrl.parentLabel"
                                        ng:model-options="{ updateOn: 'gr:datalist:update' }"
                                        ng:change="ctrl.setParentLabel()"
--- a/kahuna/public/js/search/results.html
+++ b/kahuna/public/js/search/results.html
@@ -26,13 +26,24 @@
             </div>
 
             <div class="results-toolbar-item results-toolbar-item--static results-toolbar-item--left
-                        results__related-labels related-labels flex-container text-small">
+                        results__related-labels related-labels flex-container text-small"
+                        ng:class="{'related-labels--with-labels': ctrl.relatedLabels.length !== 0}">
                 <div class="related-labels__title">
-                    Related to
-                    <form ng:submit="ctrl.setParentLabel()" style="display: inline-block">
-                        <!-- TODO: autogrow this element -->
-                        <input class="related-labels__parent" type="text" ng:model="ctrl.parentLabel" />
-                    </form>:
+                    Related to:
+                    <span ng:switch="ctrl.relatedLabels.length === 0">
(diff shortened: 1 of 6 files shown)


media-api/app/controllers/MediaApi.scala
```diff
@@ -306,12 +307,29 @@ object MediaApi extends Controller with ArgoHelpers {
...
+  private val getSuggestedLabelsLink =
+    Link("suggested-labels", s"$rootUri/suggest/edits/labels")
+
   def suggestMetadataCredit(q: Option[String], size: Option[Int]) = Authenticated.async { request =>
     ElasticSearch
       .completionSuggestion("suggestMetadataCredit", q.getOrElse(""), size.getOrElse(10))
       .map(c => respondCollection(c.results))
   }
 
+  def suggestLabels() = Authenticated {
```
jamesgorrie: This is where I am thinking it should be a subset of super labels. We can then autocomplete the [preset labels](https://github.com/guardian/grid/blob/master/kahuna/public/js/services/preset-label.js) too

media-api/app/controllers/MediaApi.scala
```diff
@@ -306,12 +307,29 @@ object MediaApi extends Controller with ArgoHelpers {
...
+    val pseudoFamousLabels = List(
```
theefer: Sad to hardcode in the codebase :crying_cat_face: 

Could it be an application config thing?

I'd start with it hardcoded to prove the point though, so :+1:

kahuna/public/js/search/results.html
```diff
@@ -27,8 +27,24 @@
...
-                <div class="related-labels__title">Related to {{ctrl.parentLabel}}:</div>
+                        ng:class="{'related-labels--with-labels': ctrl.relatedLabels.length !== 0}">
+                <div class="related-labels__title">
+                    Related to:
+                    <span ng:switch="ctrl.relatedLabels.length === 0">
+                        <form class="inline-block" ng:switch-when="true" ng:submit="ctrl.setParentLabel()">
+                            <gr-datalist class="related-labels__datalist" gr:search="ctrl.labelSearch(q)">
+                                <!-- TODO: autogrow this element -->
+                                <input class="related-labels__parent" type="text"
+                                       placeholder="e.g. culture"
```
tsop14: On uploads I said 'e.g. Observer'. I think we should be suggesting consistent label styles as they are case sensitive? so 'observer' and 'culture'... ?

theefer:
The "Related to" prompt and user flow seems weird to me? Related to what? And why does selecting something in there make it search?

theefer:
Actually looking at the screenshot the second part of the flow kind of makes sense. Might just be the "Related to __" that's odd to me. Maybe I'm wrong and @alastairjardine can pitch in.

kahuna/public/js/search/results.html
jamesgorrie: I think case insensitive as that's what others have been doing so far.
I'll change this to the "observer", but we could populate the example from the list later on.

media-api/app/controllers/MediaApi.scala
jamesgorrie: Yea, I think the point would be to have "leading" which are guardian specific, then, as they become the most popular labels, use the standard labels suggester.

theefer: OK

theefer:
:+1:

jamesgorrie:
@theefer it is now.

PR #1338, opened 2015-10-08
More matchers for Allstar agency pictures

theefer:
Should cover around 35,000 pictures that really belong to Allstar.


jamesgorrie:
Nice :+1:

PR #1339, opened 2015-10-08
typo in feature

jamesgorrie:
Oops

### DIFF
--- a/media-api/app/controllers/MediaApi.scala
+++ b/media-api/app/controllers/MediaApi.scala
@@ -321,7 +321,7 @@ object MediaApi extends Controller with ArgoHelpers {
       "obsbizcash", "obscomment", "obsfestivals", "obsfocus", "obsforeignnews", "obsgalleries",
       "obshomenews", "obsmastheads", "obssports", "obssuppliments",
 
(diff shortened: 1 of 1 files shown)


paperboyo:
This feature is a bug!

media-api/app/controllers/MediaApi.scala
```diff
@@ -321,7 +321,7 @@ object MediaApi extends Controller with ArgoHelpers {
       "obsbizcash", "obscomment", "obsfestivals", "obsfocus", "obsforeignnews", "obsgalleries",
       "obshomenews", "obsmastheads", "obssports", "obssuppliments",
```
theefer: Is it really "obssuppliments"?

jamesgorrie:
@paperboyo Why?
@theefer it's taken directly from Picdar.

paperboyo:
> it's taken directly from Picdar

But it's not really connected to Picdar any more, right? I, for one, like when people fix my grammar...

Unfortunately, they cannot fix my sense of humour, which is the answer to another of @jamesgorrie's questions.

jamesgorrie:
Ah wait, @theefer - that was [spotted and fixed](https://github.com/guardian/grid/blob/master/media-api/app/controllers/MediaApi.scala#L322). My bad.

theefer:
Cool

PR #1340, opened 2015-10-08
Import image by URI

tsop14:
Any supported file type can be imported to the grid from the browser by dragging and dropping. 
Must drag the image - cannot drag the uri text.

importuri

### DIFF
--- a/kahuna/public/js/upload/dnd-uploader.js
+++ b/kahuna/public/js/upload/dnd-uploader.js
@@ -24,6 +24,7 @@ dndUploader.controller('DndUploaderCtrl',
     ctrl.uploadFiles = uploadFiles;
     ctrl.importWitnessImage = importWitnessImage;
     ctrl.isWitnessUri = witnessApi.isWitnessUri;
+    ctrl.loadAndUpdateUriImage = loadAndUpdateUriImage;
 
     function uploadFiles(files) {
         // Queue up files for upload and go to the upload state to
@@ -53,6 +54,14 @@ dndUploader.controller('DndUploaderCtrl',
             return Promise.all([metadataUpdate, rightsUpdate]).
                 then(() => fullImage.data.id);
         });
+    }
+
+   function loadAndUpdateUriImage(fileUri) {
+        return loaderApi.import(fileUri).then(mediaResp =>
+            // Wait until image indexed
+            apiPoll(() => mediaResp.get()).
+                then(() => $state.go('upload', {}, { reload: true }))
+        );
     }
 
     function importWitnessImage(uri) {
@@ -170,9 +179,22 @@ dndUploader.directive('dndUploader', ['$window', 'delay', 'safeApply', 'track',
                         ctrl.importing = false;
                     });
                     track.action(trackEvent, dropAction('Witness'));
-                } else {
+                } else if (uri) {
+                    ctrl.importing = true;
+                    ctrl.loadAndUpdateUriImage(uri)
+                        .catch(error => {
+                            if (error.body && error.body.errorMessage) {
+                                $window.alert(error.body.errorMessage);
+                            } else {
+                                $window.alert('An error occurred while importing the ' +
+                                'image, please try again');
+                            }
+                        })
+                        .finally(() => ctrl.importing = false);
+                }
+                else {
                     $window.alert('You must drop valid files or ' +
-                                  'GuardianWitness URLs to upload them');
+                        'URLs to upload them');
 
                     track.action(trackEvent, dropAction('Invalid'));
                 }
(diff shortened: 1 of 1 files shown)


kahuna/public/js/upload/dnd-uploader.js
```diff
@@ -170,9 +179,22 @@ dndUploader.directive('dndUploader', ['$window', 'delay', 'safeApply', 'track',
                         ctrl.importing = false;
                     });
                     track.action(trackEvent, dropAction('Witness'));
-                } else {
+                } else if (uri) {
+                    ctrl.importing = true;
+                    ctrl.loadUriImage(uri)
+                        .catch(error => {
+                            if (error.body && error.body.errorMessage) {
+                                $window.alert(error.body.errorMessage);
```
theefer: What is that in the case of dragging either something that is not an image, or an image that isn't a JPG? Let's just make sure the message is understandable to human beings (aka non programmers), like @akash1810 did for the upload error.

jamesgorrie:
As mentioned earlier - I was wondering about whether this might promote people dragging images from, say, Getty without downloading the high-res version, just using the one from their site. Do we think this is a concern?

theefer:
I thought about that but -- they will learn?

Also people who access agency websites tend to be picture editors, so might be slightly more used to that?

paperboyo:
I have no idea, but I don't suppose such a d'n'd operation could carry a source URL so that we could automagically store it?

theefer:
It does, that's what we download the image from. Why store it?

paperboyo:
Do not understand a question :laughing: Let me come up with an example...

Let's say...
- I search for all images that were uploaded this way to see how many of them are of terribly poor quality
- I do a primary sort by resolution and secondary sort by usages; find the smallest one with most usage
- instead of using Google reverse Image search, I check the stored URL and can see that this is a low res image from flickr
- I click the stored URL and replace the image with the highest resolution original downloadable from flickr
- all the crops are recreated and they update in all the places that the image was used

That's before there is an automated highest-res-flick-URL-resolver added...

paperboyo:
Ah, so there would be two URLs here. One for where the image was downloaded from. And the second one - URL of a page it was dreagged from. Hmm. Second more useful, I think.

theefer:
I don't think we'd get the uri of the page where the image came from, only the image uri itself.

paperboyo:
> I don't think we'd get the uri of the page where the image came from

:sadpanda: So I will be adding that manually, as I try to do for Screengrabs...

P.s. Obviously, I imaging wrongly above: primary sort should be by the usages and secondary - by resolution

jamesgorrie:
@paperboyo I imagine that where we will have images of poor quality where there are great quality images available are mostly going to be "huge sites with external asset servers". e.g. Getty, Rex, AAP etc. If we find we have a lot of those going in, we could build specific importers, similar to GW. Let's see if that's what happens.

@tsop14 - Perhaps we could store something in `uploadInfo`? Not sure what, as the URI / filename isn't going to be of much use.

paperboyo:
> I imagine that where we will have images of poor quality where there are great quality images available are mostly going to be "huge sites with external asset servers"

That may be.

> the URI / filename isn't going to be of much use

Until it will be. Sometimes it's the only place to contain some info helping you to locate/identify (like e.g. subject matter, source etc.). Not everybody uses hashes for filenames, luckily :stuck_out_tongue_winking_eye:

tsop14:
> What is that in the case of dragging either something that is not an image, 

The provided 'uri' could not be downloaded

> or an image that isn't a JPG?

Unsupported mime-type: image/png. Supported: image/jpeg

jamesgorrie:
@paperboyo filenames are useful thus far because people have been looking for images that they have named. It's a personal reference, hence it was added for that benefit.

When you drag an image from a website, the filename was created by someone else, probably a machine, where it is going to much less useful. It'd be good to find the need to store the name before we do it.

paperboyo:
For some big sites, it's hashes indeed. Which to me looks like they are just forcing me to look at their ugly internals, while I would much prefer them not messing with what was *always* useful data in the first place (even if only hinting at the camera manufacturer and the fact that it's a digicam file...). But some are more thoughtful and do not impose their db internals, nobody cares about, onto the world. 
She looks pretty (and I knew that **before** I clicked!): https://upload.wikimedia.org/wikipedia/commons/f/f5/James_McNeill_Whistler_-_La_Princesse_du_pays_de_la_porcelaine_-_brighter.jpg.

theefer:
Feels like it could be useful to store the sourceUri, but wouldn't say it's a blocker.

kahuna/public/js/upload/dnd-uploader.js
jamesgorrie: @tsop14 [for reference](https://github.com/guardian/grid/blob/2ef8bf576dd4aad8849d8d3e8cdbf5477664a960/kahuna/public/js/upload/jobs/upload-jobs.js#L98-L100). We could probably have these in a value store to ensure consistency.

theefer:
:+1: btw, nice one!

PR #1345, opened 2015-10-12
`q` in related labels

jamesgorrie:
* retain previous search when adding a mainLabel
* take `q` into consideration when searching for related labels (previously we were returning labels that wouldn't have any results as we weren't including q)

related-labels-filter

### DIFF
--- a/kahuna/public/js/search-query/service.js
+++ b/kahuna/public/js/search-query/service.js
@@ -18,15 +18,11 @@ searchQueryService.factory('searchQueryService', ['observe$', function(observe$)
     }
 
     function addLabel(label) {
-        query$.onNext(state => {
-            return `${state.trim()} #${label}`;
-        });
+        query$.onNext(q => hasLabel(q, label) ? q : `${q} #${label}`);
     }
 
     function removeLabel(label) {
-        query$.onNext(state => {
-            return state.replace(`#${label}`, '').trim();
-        });
+        query$.onNext(q => q.replace(`#${label}`, ''));
     }
 
     function set(q) {
--- a/kahuna/public/js/search-query/service.js
+++ b/kahuna/public/js/search-query/service.js
@@ -6,8 +6,8 @@ export const searchQueryService = angular.module('gr.searchQuery.service', ['uti
 
 searchQueryService.factory('searchQueryService', [function() {
     // TODO: When we update rx, we need to swap the seed / acc around
-    const q$ = (query$, startWith) => query$.
-        scan(startWith || '', (state, fn) => fn(state).trim()).
+    const q$ = (query$, startWith = '') => query$.
+        scan(startWith, (state, fn) => fn(state).trim()).
         distinctUntilChanged().
         shareReplay(1);
 
--- a/kahuna/public/js/search-query/service.js
+++ b/kahuna/public/js/search-query/service.js
@@ -1,0 +1,37 @@
+import angular from 'angular';
+import Rx from 'rx';
+import '../util/rx';
+
+export const searchQueryService = angular.module('gr.searchQuery.service', ['util.rx']);
+
+searchQueryService.factory('searchQueryService', ['observe$', function(observe$) {
+    const query$ = new Rx.Subject();
+
+    // TODO: When we update rx, we need to swap the seed / acc around
+
+    const q$ = query$.scan('', (state, fn) => fn(state).trim());
+
+    function hasLabel(q, label) {
+        // TODO: It'd be nice to build this up from an API
+        const r = new RegExp(`[#|label\:]${label}\\b`);
+        return r.test(q);
+    }
+
+    function addLabel(label) {
+        query$.onNext(state => {
+            return `${state.trim()} #${label}`;
+        });
+    }
+
+    function removeLabel(label) {
+        query$.onNext(state => {
+            return state.replace(`#${label}`, '').trim();
+        });
+    }
+
+    function set(q) {
+        query$.onNext(() => q || '');
+    }
+
+    return { addLabel, removeLabel, set, q$ };
+}]);
--- a/kahuna/public/js/search-query/service.js
+++ b/kahuna/public/js/search-query/service.js
@@ -5,11 +5,9 @@ import '../util/rx';
 export const searchQueryService = angular.module('gr.searchQuery.service', ['util.rx']);
 
 searchQueryService.factory('searchQueryService', [function() {
-    const query$ = new Rx.Subject();
-
     // TODO: When we update rx, we need to swap the seed / acc around
-    const q$ = query$.
-        scan('', (state, fn) => fn(state).trim()).
+    const q$ = (query$, startWith) => query$.
+        scan(startWith || '', (state, fn) => fn(state).trim()).
         distinctUntilChanged().
         shareReplay(1);
 
@@ -22,17 +20,28 @@ searchQueryService.factory('searchQueryService', [function() {
         return labelMatch(label).test(q);
     }
 
-    function addLabel(label) {
-        query$.onNext(q => hasLabel(q, label) ? q : `${q} ${createLabel(label)}`);
+    function addLabel(query$) {
+        return label =>
+            query$.onNext(q => hasLabel(q, label) ? q : `${q} ${createLabel(label)}`);
     }
 
-    function removeLabel(label) {
-        query$.onNext(q => q.replace(labelMatch(label), ''));
+    function removeLabel(query$) {
+        return label => query$.onNext(q => q.replace(labelMatch(label), ''));
     }
 
-    function set(q) {
-        query$.onNext(() => q || '');
+    function setQuery(query$) {
+        return q => query$.onNext(() => q || '');
     }
 
-    return { q$, addLabel, removeLabel, 'set': set };
+    return q => {
+        const query$ = new Rx.Subject();
+        const service = {
+            q$:          q$(query$, q),
+            addLabel:    addLabel(query$),
+            removeLabel: removeLabel(query$),
+            setQuery:    setQuery(query$)
+        };
+
+        return service;
+    };
 }]);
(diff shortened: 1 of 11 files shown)


media-api/conf/routes
```diff
@@ -19,7 +19,7 @@ GET     /images                                       controllers.MediaApi.image
 
 # completion
 GET     /suggest/metadata/credit                      controllers.MediaApi.suggestMetadataCredit(q: Option[String], size: Option[Int])
-GET     /suggest/edits/labels/:label/sibling-labels   controllers.MediaApi.suggestLabelSiblings(label: String, selectedLabels: Option[String])
```
theefer: What will happen to old clients talking to the updated version of the API?

kahuna/public/js/search-query/service.js
```diff
@@ -0,0 +1,32 @@
...
+
+searchQueryService.factory('searchQueryService', [function() {
+    const query$ = new Rx.Subject();
+
+    // TODO: When we update rx, we need to swap the seed / acc around
+    const q$ = query$.scan('', (state, fn) => fn(state).trim());
+
+    function hasLabel(q, label) {
+        // TODO: It'd be nice to build this up from an API
+        const r = new RegExp(`[#|label\:]${label}\\b`);
```
theefer: I don't think `[#|label\:]` does what you think here.

media-api/conf/routes
jamesgorrie: Changed it back as we're not 100% sure what these are just yet, so no point in naming it.

kahuna/public/js/search/query.js
```diff
@@ -62,6 +69,16 @@ query.controller('SearchQueryCtrl',
...
           .forEach(setAndWatchParam);
 
+    ctrl.changeQuery = () =>
+        searchQueryService.set(ctrl.filter.query);
+
+    // FIXME: This is here to stop circular injection of the model.
+    // Avoiding: queryChange => set() => emit() => queryChange()
+    searchQueryService.q$.filter(q =>
+        q !== ctrl.filter.query
+    ).subscribe(q => ctrl.filter.query = q);
```
theefer: Is this ever unsubscribed?

kahuna/public/js/search-query/service.js
```diff
@@ -0,0 +1,32 @@
...
+        return r.test(q);
+    }
+
+    function addLabel(label) {
+        query$.onNext(q => hasLabel(q, label) ? q : `${q} #${label}`);
```
theefer: @tsop14 was working on a fix for labels that contain spaces. I can't remember if it was released or not?

kahuna/public/js/search-query/service.js
```diff
@@ -0,0 +1,32 @@
...
+import Rx from 'rx';
+import '../util/rx';
+
+export const searchQueryService = angular.module('gr.searchQuery.service', ['util.rx']);
+
+searchQueryService.factory('searchQueryService', [function() {
+    const query$ = new Rx.Subject();
+
+    // TODO: When we update rx, we need to swap the seed / acc around
+    const q$ = query$.scan('', (state, fn) => fn(state).trim());
```
theefer: Just to check: does `trim` not get in the way of typing several words (i.e. does it trim the trailing space while typing)?

theefer: Also: check the [other reactive state helpers](https://github.com/guardian/grid/blob/master/kahuna/public/js/lib/data-structure/list-factory.js#L17-L19), I suspect you may want to add `distinctUntilChanged` (to skip identical values) and `shareReplay` (to share subscription and replay the last value to new subscribers).

kahuna/public/js/search/query.js
```diff
@@ -62,6 +69,16 @@ query.controller('SearchQueryCtrl',
     Object.keys($stateParams)
           .forEach(setAndWatchParam);
 
+    ctrl.changeQuery = () =>
+        searchQueryService.set(ctrl.filter.query);
+
+    // FIXME: This is here to stop circular injection of the model.
+    // Avoiding: queryChange => set() => emit() => queryChange()
+    searchQueryService.q$.filter(q =>
+        q !== ctrl.filter.query
```
theefer: Is this necessary? I would have thought that if `q === ctrl.filter.query`, then `ctrl.filter.query = q` is essentially a no-op?

kahuna/public/js/search-query/service.js
jamesgorrie: No, you're right it doesn't.
[Updated](https://github.com/guardian/grid/commit/d84af857546d103811715310a0abf993407e1cb3)

kahuna/public/js/search/query.js
jamesgorrie: * User types `"culture"`
* `q$.emit("culture")`
* `filter.query = q`

We just set q twice unessecarily, which could also cause problems. This is the same [challenge we had in the usage rights editor](https://github.com/guardian/grid/blob/master/kahuna/public/js/usage-rights/usage-rights-editor.js#L47-L58)

kahuna/public/js/search/query.js
jamesgorrie: [I'm using the inject now](https://github.com/guardian/grid/commit/dd56ac3da86f44699e549aeb97be6bda3b397792#diff-32fea343d36b21c11521e08ea0059cafR78), so it will be.

kahuna/public/js/search-query/service.js
```diff
@@ -0,0 +1,38 @@
+import angular from 'angular';
+import Rx from 'rx';
+import '../util/rx';
+
+export const searchQueryService = angular.module('gr.searchQuery.service', ['util.rx']);
+
+searchQueryService.factory('searchQueryService', [function() {
+    const query$ = new Rx.Subject();
+
+    // TODO: When we update rx, we need to swap the seed / acc around
```
theefer: We should do that soon, and yes we'll need to be careful about that :+1:

kahuna/public/js/search/index.js
```diff
@@ -47,6 +49,11 @@ search.config(['$stateProvider',
                 });
                 return {state: $dsr$.redirect.state, params};
             }]
+        },
+        resolve: {
+            searchQueryService: ['$stateParams', 'searchQueryService', function($stateParams, searchQueryService) {
+                return searchQueryService($stateParams.query);
```
jamesgorrie: @theefer the problem I have here is that `$stateParms.query` is formed when we're in the `search.results` view, but we need it in both. I was wondering if the view hierarchy that we have at the moment is necessarily correct.

jamesgorrie:
@theefer I've now updates this so it works using the router as the place for the abstraction of the `searchQueryService`.

[There is a bit of what looks like a hack](https://github.com/guardian/grid/commit/b6ee6fc5fcf1072dc9f9c52cabcf41e32ddc21f7#diff-04b33f0ec2971bf016d5d80534267132R42), although it is actually what we are trying to do, but I think it might be the current view hierarchy that makes it a bit strange.

kahuna/public/js/search/results.html
```diff
@@ -44,7 +44,7 @@
                                        gr:datalist-input />
                             </gr-datalist>
                         </form>
-                        <span ng:switch-default>{{ctrl.parentLabel}}:</span>
+                        <span ng:switch-default>{{ctrl.parentLabel}}</span>:
```
theefer: Does that mean it will always be there (even when showing the input dropdown)?

kahuna/public/js/search/index.js
```diff
@@ -36,7 +39,9 @@ search.config(['$stateProvider',
     $stateProvider.state('search', {
         // FIXME [1]: This state should be abstract, but then we can't navigate to
         // it, which we need to do to access it's deeper / remembered chile state
-        url: '/',
+        // FIXME: We now share the `query` param between the `query` and results view.
+        // I'm not sure if this is exactly what we want.
+        url: '/?query',
```
theefer: Does it still get copied into `ctrl.filter.query`? Is that even an issue?

kahuna/public/js/search/index.js
```diff
@@ -47,11 +52,16 @@ search.config(['$stateProvider',
                 });
                 return {state: $dsr$.redirect.state, params};
             }]
+        },
+        resolve: {
+            searchQueryService: ['$stateParams', 'searchQueryService', function($stateParams, searchQueryService) {
```
theefer: Is *Service* appropriate here? Should it just be `searchQuery`?

jamesgorrie: Yea, I was trying not to break things while refactoring, I'll get on the naming...

kahuna/public/js/search/results.html
jamesgorrie: Yip, which is what I would expect, but happy for it not to be that way.

jamesgorrie:
Noticed a bug with the new routing that we lose focus on the input while typing, so looking into that.

kahuna/public/js/search/results.html
theefer: I personally wouldn't expect a `:` after the input field followed by nothing, but YMMV. Up to you.

kahuna/public/js/search-query/service.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    function setQuery(query$) {
+        return q => query$.onNext(() => q || '');
+    }
+
+    return q => {
+        const query$ = new Rx.Subject();
+        const service = {
+            q$:          q$(query$, q),
```
jamesgorrie: What do we think about building `classes` this way?
Ot is perhaps using classes better?

theefer: I find it slightly confusing, but perhaps it's just me or just naming issues (`q$` is a function that returns a new version of `query$` which we then call `q$`, for example), and having to pass in the internal state to each "method" explicitly.

Contrast with [previous data structures](https://github.com/guardian/grid/blob/master/kahuna/public/js/lib/data-structure/ordered-set-factory.js) for comparison. Not saying those are perfect either, but it seems to have slightly fewer closures.

theefer: PS: One of my issues with classes is that you can't sanely make anything private (e.g. `query$` here).

jamesgorrie:
Closing in favour of the direction of #1399

PR #1346, opened 2015-10-12
Misc cropper cleanups

theefer:
A set of code cleanups prior to extending the cropper to support full exports.

This should not introduce any functional change.

See individual commits for a list of changes.


common-lib/src/main/scala/com/gu/mediaservice/model/Export.scala
```diff
@@ -12,8 +12,7 @@ case class Export(
   date: Option[DateTime],
   specification: CropSource,
   master: Option[Asset],
-  assets: List[Asset],
-  exportType: String
```
theefer: This was not used by anything AFAIK.

cropper/app/lib/Crops.scala
```diff
@@ -60,24 +61,25 @@ object Crops {
...
-      val cropTooBig = bounds.width > dimensions.width || bounds.height > dimensions.height
+    asset.dimensions.exists { dimensions =>
+      val positiveCoords       = List(bounds.x,     bounds.y     ).forall(_ >= 0)
+      val strictlyPositiveSize = List(bounds.width, bounds.height).forall(_  > 0)
+      val withinBounds = (bounds.x + bounds.width  <= dimensions.width ) &&
+                         (bounds.y + bounds.height <= dimensions.height)
 
-      outOfBounds || cropTooSmall || cropTooBig
-    } getOrElse true
+      positiveCoords && strictlyPositiveSize && withinBounds
```
theefer: Inverted the logic and made the check more exhaustive (previously, the crop bounds could overflow the actual image dimensions and still be considered valid).

kenoir:
:+1:

common-lib/src/main/scala/com/gu/mediaservice/model/Export.scala
jamesgorrie: We do have it in the [Mappings](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala#L90) - although I can see it's been moved in your [other PR](https://github.com/guardian/grid/pull/1349/files#diff-fc7552667a451d2b7835c23fa4739ac8R70).

Looking at the ES record though, we are recording it (truncated):
```JSON
{
    "exports": [
        {
            "specification": {},
            "id": "0_161_2447_1467",
            "author": "them.tham@there.co.uk",
            "assets": [
                {},
                {}
            ],
            "date": "2015-06-23T18: 12: 30Z",
            "type": "crop",
            "master": {
                "dimensions": {
                    "height": 1467,
                    "width": 2447
                },
                "secureUrl": ".../0_161_2447_1467/master/2447.jpg",
                "file": .../0_161_2447_1467/master/2447.jpg",
                "mimeType": "image/jpeg",
                "size": 2994287
            }
        }
    ]
}
```
There is also a hard reference to let it be stored in ES [here](https://github.com/guardian/grid/blob/master/cropper/app/controllers/Application.scala#L81).

I don't think it's a problem, just FYI.

jamesgorrie: Although I see you've remove the `"type" -> "crop"` in the other PR too.

PR #1347, opened 2015-10-12
Label styling and recognising labels-with-spaces

tsop14:
**Labels-with-spaces**
Filters labels-with-spaces (e.g. #NEW nauru pics from BD) so they become enclosed in quote marks and therefore searchable
Adds case to match phrases in imageSearch so that labels-with-spaces have .selected assigned to them
- label turns blue when selected
- label cannot be added twice

NOW:
labels

BEFORE:
labels2

**Styling**
Removes scroll bar under related labels
Moves semicolon to after the parent label

NOW:
screenshot-media local dev-gutools co uk 2015-10-12 16-24-58

BEFORE: 
screenshot-media local dev-gutools co uk 2015-10-12 16-17-18

### DIFF
--- a/kahuna/public/js/search/results.js
+++ b/kahuna/public/js/search/results.js
@@ -55,6 +55,7 @@ results.controller('SearchResultsCtrl', [
     'panelService',
     'range',
     'isReloadingPreviousSearch',
+    'queryLabelFilterFilter',
     function($rootScope,
              $scope,
              $state,
@@ -73,7 +74,8 @@ results.controller('SearchResultsCtrl', [
              results,
              panelService,
              range,
-             isReloadingPreviousSearch) {
+             isReloadingPreviousSearch,
+             queryLabelFilterFilter) {
 
         const ctrl = this;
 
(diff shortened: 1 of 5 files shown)


kahuna/public/js/search/results.js
```diff
@@ -207,7 +210,8 @@ results.controller('SearchResultsCtrl', [
 
         ctrl.setParentLabel = () => {
             if (ctrl.parentLabel) {
-                $state.transitionTo($state.current, { query: `#${ctrl.parentLabel}` }, {
+                const parentLabel = queryLabelFilterFilter(ctrl.parentLabel);
+                $state.transitionTo($state.current, { query: `${parentLabel}` }, {
```
theefer: Note that ``` `${x}` ``` is a bit of a tautology for just `x`.

theefer:
:+1: but this will conflict with @jamesgorrie's PR

kahuna/public/stylesheets/main.css
```diff
@@ -974,7 +974,8 @@ input.related-labels__parent {
 }
 
 .related-labels__labels {
-    overflow: auto;
+    overflow-x: auto;
```
jamesgorrie: This won't work due to the restrictions behind `overflow`. TL;DR; Anything with `overflow: auto` will have the other axis set to auto too.

kahuna/public/stylesheets/main.css
```diff
@@ -957,7 +957,7 @@ input.search-query__input {
 .related-labels--with-labels {
     /* This is to allow scrolling of labels, but show the
     autocomplete dropdown */
-    overflow: auto;
+    overflow-x: auto;
```
jamesgorrie: This will set both axis to auto AFAIK?

kahuna/public/stylesheets/main.css
theefer: Not sure what you mean, we tried with @tsop14 and it solved the problem?

jamesgorrie: I'm getting it mixed up - it's when you mix it with `visible`, this doesn't wokr.
:+1:

jamesgorrie:
> conflict

[I've added support in my branch](https://github.com/guardian/grid/commit/da5b57510e790128edd36b7cb6947a3c5a523ad9)

kahuna/public/stylesheets/main.css
jamesgorrie: Ignore - I'm being silly.

PR #1348, opened 2015-10-12
Make client send export type (currently unused)

theefer:
This will come into play with the upcoming PR that reads the `type` argument.


jamesgorrie:
:+1:

PR #1349, opened 2015-10-12
Add support for "full" export in cropper

theefer:
i.e. exporting a "crop" that contains the entire image.

(In the future, we may be able to make that smarter to automatically exclude Getty banners and the like. Currently the full export would include the banner.)

To create a full export, all you have to do is `POST` to `cropper.gu.com/crops` the following JSON:

```
{
  "type": "full",
  "source": "uri of the image resource"
}
```

## TODO

- [ ] Add a test to show API error if no "type" sent.
- [x] Surface error if Media API is not available.


cropper/app/lib/Crops.scala
```diff
@@ -21,8 +21,8 @@ object Crops {
   import scala.concurrent.ExecutionContext.Implicits.global
   import Files._
 
-  def outputFilename(source: SourceImage, bounds: Bounds, outputWidth: Int, isMaster: Boolean = false): String = {
-    s"${source.id}/${Crop.getCropId(bounds)}/${if(isMaster) "master/" else ""}$outputWidth.jpg"
+  def outputFilename(source: SourceImage, bounds: Bounds, outputWidth: Int, exportType: String, isMaster: Boolean = false): String = {
+    s"${source.id}/${Crop.getCropId(bounds)}/$exportType/${if(isMaster) "master/" else ""}$outputWidth.jpg"
```
theefer: Note the change in output paths. This means paths won't be backwards compatible. Not entirely sure if it's worth distinguishing images in their path, rather than having them flat (there is still a `type` metadata property recorded on each S3 object), so open for discussion.

cropper/app/controllers/Application.scala
```diff
@@ -34,6 +34,9 @@ object Application extends Controller with ArgoHelpers {
 
   val mediaApiKey = keyStore.findKey("cropper").getOrElse(throw new Error("Missing cropper API key in key bucket"))
 
+  val parseExportRequest = parse.json(ExportRequest.readExportRequest)
```
theefer: @jamesgorrie This seems like a much nicer way to parse input than the former forms imho.

jamesgorrie: I really don't like those forms: We have [something similar here](https://github.com/guardian/grid/blob/master/metadata-editor/app/controllers/EditsController.scala#L174-L179).

I would like to explore typed requests e.g.
```
Authenticated.async(parse.json)[Export] { export =>
  //...
}
```

cropper/app/controllers/Application.scala
```diff
@@ -44,52 +47,24 @@ object Application extends Controller with ArgoHelpers {
...
+    val user = httpRequest.user
 
-  val cropSourceForm: Form[CropSource] = Form(
-    tuple("source" -> nonEmptyText, "x" -> number, "y" -> number, "width" -> number, "height" -> number, "aspectRatio" -> optional(nonEmptyText))
-      .transform[CropSource]({ case (source, x, y, w, h, r) => CropSource(source, Bounds(x, y, w, h), r) },
-                       { case CropSource(source, Bounds(x, y, w, h), r) => (source, x, y, w, h, r) })
-  )
+    executeRequest(exportRequest, user).map { case (imageId, export) =>
+      val cropJson = Json.toJson(export).as[JsObject]
+      val exports  = Json.obj(
```
jamesgorrie: Could we case class these?

cropper/app/controllers/Application.scala
theefer: This is kind of what I do here? Although the syntax you wrote seems a little cleaner still.

cropper/app/model/ExportRequest.scala
```diff
@@ -0,0 +1,45 @@
...
+  private val readCropRequest: Reads[CropRequest] = (
+    (__ \ "source").read[String] ~
+    __.read[Bounds] ~
+    (__ \ "aspectRatio").readNullable[String](pattern(aspectRatioLike))
+  )(CropRequest.apply _)
+
+  private val readFullExportRequest: Reads[FullExportRequest] =
+    (__ \ "source").read[String].map(FullExportRequest.apply)
+
+  val readExportRequest: Reads[ExportRequest] = Reads[ExportRequest]( jsvalue => (jsvalue \ "type") match {
```
jamesgorrie: if `(jsvalue \ "type")` isn't there, what does it return? Or does it error if we haven't told it to return an `Option`?

cropper/app/controllers/Application.scala
theefer: Bit overkill as it will really only be used here...?

cropper/app/model/ExportRequest.scala
theefer: It should then fall into the error case on line 35.

theefer: I might try and write a test for this actually :-)

cropper/app/controllers/Application.scala
```diff
@@ -44,52 +47,24 @@ object Application extends Controller with ArgoHelpers {
...
-
-    val author: Option[String] = httpRequest.user match {
-      case user: AuthenticatedService => Some(user.name)
-      case user: PandaUser => Some(user.email)
+      Notifications.publish(exports, "update-image-exports")
+      Ok(cropJson).as(ArgoMediaType)
+    } recover {
+      case InvalidImage => respondError(BadRequest, "invalid-image", InvalidImage.getMessage)
+      case MissingSecureSourceUrl => respondError(BadRequest, "no-source-image", MissingSecureSourceUrl.getMessage)
+      case InvalidCropRequest => respondError(BadRequest, "invalid-crop", InvalidCropRequest.getMessage)
```
jamesgorrie: Are we going to have a case of the media API being down as that is a dependency?

theefer: Good point, I'll see how this can be surfaced (will have to be a 5xx)

cropper/app/controllers/Application.scala
jamesgorrie: Exactly, I am agreeing that I think this is a much nicer way to go :dancers:

cropper/app/controllers/Application.scala
jamesgorrie: :+1:

cropper/app/controllers/Application.scala
theefer: [Nicer with your suggestion](https://github.com/guardian/grid/commit/f7cd23b9f052a047770020159858022766917686), thanks!

jamesgorrie: Bonza :boom:

theefer:
I've now [added better handling](https://github.com/guardian/grid/commit/fc61da4be2fe2af5d4bd5fd06e4a31891c36b22b) of a variety of errors that could occur. I did it by writing a test suite for the controller, but unfortunately it's still a little too hacky to commit (wouldn't pass on CI) so more work is needed to make the controller testable in isolation.

cropper/app/controllers/Application.scala
```diff
@@ -44,52 +50,30 @@ object Application extends Controller with ArgoHelpers {
 
   def index = Authenticated { indexResponse }
 
+  def export = Authenticated.async(parse.json) { httpRequest =>
+    httpRequest.body.validate[ExportRequest] map { exportRequest =>
```
jamesgorrie: Where does this Validate come from?

theefer: `body` is a JValue because of the `parse.json` BodyParser on the line above. Does that answer your question?

theefer: I believe `validate` is like `asOpt` but returning a more detailed error if failed.

cropper/app/controllers/Application.scala
```diff
@@ -44,52 +50,30 @@ object Application extends Controller with ArgoHelpers {
...
+        case InvalidSource => respondError(BadRequest, "invalid-source", InvalidSource.getMessage)
+        case ImageNotFound => respondError(BadRequest, "image-not-found", ImageNotFound.getMessage)
+        case InvalidImage => respondError(BadRequest, "invalid-image", InvalidImage.getMessage)
+        case MissingSecureSourceUrl => respondError(BadRequest, "no-source-image", MissingSecureSourceUrl.getMessage)
+        case InvalidCropRequest => respondError(BadRequest, "invalid-crop", InvalidCropRequest.getMessage)
+        case ApiRequestFailed => respondError(BadGateway, "api-failed", ApiRequestFailed.getMessage)
       }
-    )
+    } recoverTotal {
+      case e => Future.successful(respondError(BadRequest, "bad-request", e.errors.head._2.head.message))
```
theefer: I forgot this, which is quite hacky and side-effecty. Will improve.

cropper/app/lib/Crops.scala
jamesgorrie: As spoken about over IM - I can't see the harm in this, but  I can't see why it would be beneficial to have this?

Potentially if we wanted to differentiate between a full-size, freeform crop and a full export, this might be useful, but I would argue those are the same thing.

cropper/app/controllers/Application.scala
jamesgorrie: Nice

theefer: Note that I moved it from `Authenticated.async(parse.json[ExportRequest])` because that causes a nasty HTML page to be returned if the parsing failed, which is silly, we want a nice JSON error.

cropper/app/lib/Crops.scala
theefer: Reverted it to *not* include the export type.

jamesgorrie:
:+1:

PR #1350, opened 2015-10-12
use updated image when sending over metadata edit

jamesgorrie:
If we keep using the non-updated one, the metadata diff doesn't work correctly.
To reproduce: 
* get image with no description / credit
* update description
* update credit

Then the credit disappears as the image is the image we loaded the page with, which wouldn't have a description.

Surprised this hasn't come up earlier.


theefer:
:+1:

PR #1352, opened 2015-10-13
First pass at DRYing the logger xml.

akash1810:
Using an `include` directive DRY out logger configuration commonality.

Not all apps are created equal, however, so `include` isn't used everywhere. For example, `image-loader` has a root level of `ERROR` whereas other apps are `DEBUG`.

This also only logs AWS ERRORs:

```xml
<logger name="com.amazonaws">
  <level value="ERROR" />
</logger>
```

### DIFF
--- a/image-loader/conf/logger.xml
+++ b/image-loader/conf/logger.xml
@@ -15,21 +15,6 @@
         </encoder>
     </appender>
 
-    <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
-        <encoder>
-            <pattern>%date [%thread] %-5level GUUID=%X{GUUID} %logger{36}:%L - %msg%n%xException{15}</pattern>
-        </encoder>
-    </appender>
-
-    <logger name="play" level="INFO" />
-    <logger name="application" level="DEBUG" />
-    <logger name="com.amazonaws">
-        <level value="ERROR" />
-    </logger>
-
-    <root level="ERROR">
-        <appender-ref ref="CONSOLE"/>
-        <appender-ref ref="LOGFILE"/>
-    </root>
+    <include resource="logger-include.xml" />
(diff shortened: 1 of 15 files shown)


image-loader/conf/logger.xml
```diff
@@ -23,6 +23,9 @@
 
     <logger name="play" level="INFO" />
     <logger name="application" level="DEBUG" />
+    <logger name="com.amazonaws">
+        <level value="ERROR" />
+    </logger>
```
theefer: Why is image-loader different from the others?

metadata-editor/conf/logger.xml
```diff
@@ -2,17 +2,6 @@
...
-        </encoder>
-    </appender>
-
-    <logger name="play" level="INFO" />
-    <logger name="application" level="DEBUG" />
-
-    <root level="DEBUG">
-        <appender-ref ref="CONSOLE"/>
-    </root>
+    <include resource="logger-include.xml" />
```
theefer: Missing the linking file for metadata-editor?

akash1810: Oops! Added.

image-loader/conf/logger.xml
akash1810: Not sure. Have switched the root level to debug in DEV and the logs are not any louder. Will change.

kenoir:
Nice. :+1:

akash1810:
Reverting because TC doesn't support symlinks on server side checkouts because they use JGit on the server - https://youtrack.jetbrains.com/issue/TW-12648.

It looks like [JGit has fixed this](https://bugs.eclipse.org/bugs/show_bug.cgi?id=354367), so maybe a later version of TC will fix this.

PR #1353, opened 2015-10-13
Isolate scope in gr-preset-labels.

akash1810:
Fixes bug where the upload button does not work on the uploads page.

### DIFF
--- a/kahuna/public/js/components/gr-preset-labels/gr-preset-labels.js
+++ b/kahuna/public/js/components/gr-preset-labels/gr-preset-labels.js
@@ -64,6 +64,7 @@ presetLabels.directive('grPresetLabels', [function () {
         controller: 'GrPresetLabelsCtrl',
         controllerAs: 'ctrl',
         bindToController: true,
-        template: template
+        template: template,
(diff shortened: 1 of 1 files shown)


jamesgorrie:
:+1: for the fix.

What's actually happening though?

akash1810:
> What's actually happening though?

Steps to reproduce problem:
- Visit `/upload`.
- Press Upload button.
- Choose file.
- Witness nothing happen (no file uploaded, not console logs).

jamesgorrie:
Yip, get the interface problem, but why did this help?

akash1810:
Ah, sorry! Without an isolated scope, [`ctrl.latestJob`](https://github.com/guardian/grid/blob/master/kahuna/public/js/upload/controller.js#L14) (also [here](https://github.com/guardian/grid/blob/master/kahuna/public/js/upload/view.html#L21)) gets removed. Similar solution was added to [`prompt`](https://github.com/guardian/grid/blob/master/kahuna/public/js/upload/prompt/prompt.js#L15) too.

PR #1354, opened 2015-10-13
componentise invalid metadata message.

akash1810:
Part of the work around componentising the metadata panel on the image page to make adding usage simpler.


jamesgorrie:
:+1:

PR #1355, opened 2015-10-13
search with q in related labels

jamesgorrie:
Decoupling from #1345 
We now consider the query in the search for related labels.
We also don't clear the search when adding the main label.

### DIFF
--- a/kahuna/public/js/search/results.js
+++ b/kahuna/public/js/search/results.js
@@ -211,7 +211,8 @@ results.controller('SearchResultsCtrl', [
         ctrl.setParentLabel = () => {
             if (ctrl.parentLabel) {
                 const parentLabel = queryLabelFilterFilter(ctrl.parentLabel);
-                $state.transitionTo($state.current, { query: `${parentLabel}` }, {
+                const newQ = `${$stateParams.query.trim()} ${parentLabel}`;
+                $state.transitionTo($state.current, { query: newQ }, {
                     reload: true, inherit: false, notify: true
                 });
             }
--- a/media-api/app/controllers/MediaApi.scala
+++ b/media-api/app/controllers/MediaApi.scala
@@ -234,27 +234,12 @@ object MediaApi extends Controller with ArgoHelpers {
 
     val searchParams = SearchParams(request)
 
-    // We search if we have a label in the search, take the first one and then look up it's
-    // siblings s that we can return them as "related labels"
-    val labels = searchParams.structuredQuery.flatMap {
-      // TODO: Use ImageFields for guard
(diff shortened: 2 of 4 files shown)


kahuna/public/js/search/results.js
```diff
@@ -211,7 +211,8 @@ results.controller('SearchResultsCtrl', [
         ctrl.setParentLabel = () => {
             if (ctrl.parentLabel) {
                 const parentLabel = queryLabelFilterFilter(ctrl.parentLabel);
-                $state.transitionTo($state.current, { query: `${parentLabel}` }, {
+                const newQ = `${$stateParams.query.trim()} ${parentLabel}`;
```
jamesgorrie: This will hopefully be more intelligent when using a query service.

theefer:
:+1:

media-api/app/controllers/MediaApi.scala
```diff
@@ -299,16 +284,32 @@ object MediaApi extends Controller with ArgoHelpers {
...
-    )
-    val paramVars = paramMap.map { case (key, value) => key := value }.toSeq
-    val uri = uriTemplate expand (paramVars: _*)
+  private def getRelatedLabelsLink(searchParams: SearchParams) = {
+    // We search if we have a label in the search, take the first one and then look up it's
+    // siblings so that we can return them as "related labels"
+    val labels = searchParams.structuredQuery.flatMap {
+      // TODO: Use ImageFields for guard
+      case Match(field: SingleField, value:Words) if field.name == "userMetadata.labels" => Some(value.string)
+      case Match(field: SingleField, value:Phrase) if field.name == "userMetadata.labels" => Some(value.string)
```
akash1810: why pattern match on `Words` and `Phrase` separately?

jamesgorrie: How would you do it in one?

akash1810: `case Match(field: SingleField, value: Value) if field.name == "userMetadata.labels" => Some(value.toString)`

Though you'd have to create another trait to for Words and Phrase (as DateRange also extends Value)... which is not worth it.

jamesgorrie: DateRange is in there too.
And we might not want to include other, new values, I prefer to be explicit while matching, almost feels like the point of it.

jamesgorrie: > create another trait to for Words and Phrase

I'm starting to regret my use of `trait`s else where, so trying to get away from it. It always seems to add unnecessary complexity.

PR #1357, opened 2015-10-13
Don't trim undefined

jamesgorrie:
TypeScript as a solution?


theefer:
:+1: Only if we replace null/undefined by some kind of option..

jamesgorrie:
[?](http://www.scala-js.org/)

PR #1359, opened 2015-10-13
return more related labels

jamesgorrie:
For travel particularly.

### DIFF
--- a/media-api/app/lib/elasticsearch/ElasticSearch.scala
+++ b/media-api/app/lib/elasticsearch/ElasticSearch.scala
@@ -125,7 +125,8 @@ object ElasticSearch extends ElasticSearchClient with SearchFilters with ImageFi
       }
   }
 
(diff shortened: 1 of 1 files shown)


theefer:
:+1:

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -125,7 +125,8 @@ object ElasticSearch extends ElasticSearchClient with SearchFilters with ImageFi
       }
   }
 
-  def labelSiblingsSearch(structuredQuery: List[Condition], excludeLabels: List[String] = Nil)(implicit ex: ExecutionContext): Future[AggregateSearchResults] = {
+  def labelSiblingsSearch(structuredQuery: List[Condition], excludeLabels: List[String] = Nil, size: Int = 20)
```
jamesgorrie: I've added it as a param here, but don't want to open up the API to have a infinite number (like the rest of our API :scream:)

theefer: Elsewhere we should enforce a maximum.

jamesgorrie: Indeed.

PR #1360, opened 2015-10-13
encode label in link

jamesgorrie:
Not the greatest solution, but stops the search for `"` blowing up our `URITemplate` object.


akash1810:
:+1:

PR #1361, opened 2015-10-14
switch between suggested labels

jamesgorrie:
As per feedback and also not having completed the `searchQueryService`(#1345), the suggested labels now switch as opposed to toggle.

I image the toggle would be more useful in a slightly less hierarchical world, but for now we have two levels and this is how users are using it.

switch-related-tags


theefer:
:+1:

PR #1363, opened 2015-10-14
fixes issue where cropped symbol wasn't showing.

akash1810:
We no longer return `type` in the api response.


jamesgorrie:
Damn overengineering... :+1:

theefer:
:+1:

PR #1364, opened 2015-10-14
persist illustrators.

akash1810:


jamesgorrie:
Did we need to add this into ES anywhere?
I can't remember if we query on `isPersisted`?

akash1810:
good catch! 1 sec...

akash1810:
> Did we need to add this into ES anywhere?

Done now.

(Probably time to write tests now...)

jamesgorrie:
:+1:

theefer:
:+1:

PR #1365, opened 2015-10-14
Revert "DRYing the logger xml."

akash1810:
Reverts guardian/grid#1352


akash1810:
Reverting because TC doesn't support symlinks on server side checkouts because they use JGit on the server - https://youtrack.jetbrains.com/issue/TW-12648.

It looks like [JGit](https://bugs.eclipse.org/bugs/show_bug.cgi?id=354367) has fixed this, so maybe a later version of TC will fix this.

kenoir:
:crying_cat_face:

PR #1368, opened 2015-10-14
Image metadata as a component.

akash1810:
Part of the work around componentising the metadata panel on the image page to make adding usage simpler.

Related - https://github.com/guardian/grid/pull/1354.

Ideally, `gr-panel` would be used instead, however, the amount of data displayed differs between the two.

No functional change, just a refactoring.


kahuna/public/js/image/view.html
```diff
@@ -90,238 +90,9 @@
...
-                            {{keyword}}
-                        </a>
-                    </li>
-                </ul>
-            </dd>
-        </dl>
-    </div>
+    <gr-metadata-validity gr-image="ctrl.image"></gr-metadata-validity>
+    <gr-image-cost-message gr-image="ctrl.image"></gr-image-cost-message>
+    <gr-image-metadata gr-image="ctrl.image"></gr-image-metadata>
```
akash1810: The main reason for the change is here.

We'd be able to simply say:

```html
<gr-image-metadata ng:if="showMetadata" gr-image="ctrl.image"></gr-image-metadata>
<gr-image-usage ng:if="! showMetadata" gr-image="ctrl.image"></gr-image-usage>
```

jamesgorrie:
> the amount of data displayed differs between the two

Is this just a repercussion around the fact that they're not the same thing?

akash1810:
I'd argue they are the same thing - the difference being one is for a single image and the other for one or more. Probably want to work out what people want to see in these two views to be certain.

jamesgorrie:
Good point, it allows us that.

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
```diff
@@ -0,0 +1,145 @@
...
+
+    function ($rootScope,
+              $scope,
+              imageService,
+              editsService,
+              mediaApi,
+              editsApi,
+              track) {
+
+        let ctrl = this;
```
jamesgorrie: `const`?

jamesgorrie: Wait, being stupid.

kahuna/public/js/image/controller.js
```diff
@@ -1,13 +1,12 @@
...
 
 import '../image/service';
-import '../edits/service';
-import '../analytics/track';
 import '../components/gr-delete-image/gr-delete-image';
 import '../components/gr-add-label/gr-add-label';
 import '../downloader/downloader';
 import '../components/gr-crop-image/gr-crop-image';
 import '../components/gr-delete-crops/gr-delete-crops';
+import '../components/gr-image-metadata/gr-image-metadata';
```
jamesgorrie: What are you thoughts around having these `components/componentname/componentname`?
Why not just `components/componentname`?

jamesgorrie:
Looks sensible - but it is starting to raise questions around our app structure.

General best practice is to form components, but we're kind of breaking things into views and then services that live else where... Perhaps not yet, but I think it would be good to have a chat.

:+1:

kahuna/public/js/image/controller.js
akash1810: `components/componentname/componentname` allows for the template and module to exist in the same directory (`components/componentname`). Though I agree its a bit of a hassle.

akash1810:
> it would be good to have a chat

:+1: Starting to think maybe the `components` directory isn't needed anymore and the components in that dir can be moved up a level.

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
theefer: No it could be a `const` if we wanted to.

theefer:
> Starting to think maybe the components directory isn't needed anymore and the components in that dir can be moved up a level.

I find there's already quite a confusing mix in the root directory of "pages" and "components" and "utilities" etc. Would be nice to use the file tree to clearly delineate the different types of "things".

PR #1369, opened 2015-10-15
add extra rex parser

jamesgorrie:
We seem to own all of them.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
@@ -245,11 +245,14 @@ object ReutersParser extends ImageProcessor {
 }
 
 object RexParser extends ImageProcessor {
-  def apply(image: Image): Image = image.metadata.source match {
+  val rexAgency = Agency("Rex Features")
+  val SlashRex = ".+/ Rex Features".r
+
+
+  def apply(image: Image): Image = (image.metadata.source, image.metadata.credit) match {
     // TODO: cleanup byline/credit
-    case Some("Rex Features") => image.copy(
-      usageRights = Agency("Rex Features")
-    )
+    case (Some("Rex Features"), _) => image.copy(usageRights = rexAgency)
+    case (_, Some(SlashRex()))     => image.copy(usageRights = rexAgency)
     case _ => image
   }
 }
--- a/common-lib/src/test/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessorsTest.scala
+++ b/common-lib/src/test/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessorsTest.scala
@@ -385,6 +385,13 @@ class SupplierProcessorsTest extends FunSpec with Matchers with MetadataHelper {
       processedImage.metadata.credit should be(Some("Tim Ireland/REX Shutterstock"))
       processedImage.metadata.source should be(Some("Rex Features"))
     }
+
+    it("should match '*/ Rex Features' credit") {
+      val image = createImageFromMetadata("credit" -> "Bleddyn Butcher / Rex Features")
+      val processedImage = applyProcessors(image)
+
+      processedImage.usageRights should be (Agency("Rex Features"))
+    }
   }
 
 


common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -245,11 +245,14 @@ object ReutersParser extends ImageProcessor {
...
+  val SlashRex = ".+/ Rex Features".r
+
+
+  def apply(image: Image): Image = (image.metadata.source, image.metadata.credit) match {
     // TODO: cleanup byline/credit
-    case Some("Rex Features") => image.copy(
-      usageRights = Agency("Rex Features")
-    )
+    case (Some("Rex Features"), _) => image.copy(usageRights = rexAgency)
+    case (_, Some(SlashRex()))     => image.copy(usageRights = rexAgency)
```
theefer: Note you can use `case foo | bar =>` if you want.

common-lib/src/test/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessorsTest.scala
```diff
@@ -385,6 +385,13 @@ class SupplierProcessorsTest extends FunSpec with Matchers with MetadataHelper {
       processedImage.metadata.credit should be(Some("Tim Ireland/REX Shutterstock"))
       processedImage.metadata.source should be(Some("Rex Features"))
     }
+
+    it("should match '*/ Rex Features' credit") {
+      val image = createImageFromMetadata("credit" -> "Bleddyn Butcher / Rex Features")
+      val processedImage = applyProcessors(image)
+
+      processedImage.usageRights should be (Agency("Rex Features"))
```
theefer: Do we want to save the prefix as the collection, or is it other random things (byline etc)?

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -245,11 +245,14 @@ object ReutersParser extends ImageProcessor {
 }
 
 object RexParser extends ImageProcessor {
-  def apply(image: Image): Image = image.metadata.source match {
+  val rexAgency = Agency("Rex Features")
+  val SlashRex = ".+/ Rex Features".r
```
theefer: Is there always a space before `Rex` (e.g. in Getty there isn't, but Rex may be different)?

theefer:
:+1:

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
jamesgorrie: We're matching on source and credit here, hence the tuple. Unless I'm missing something?

common-lib/src/test/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessorsTest.scala
jamesgorrie: I know we're meant to keep the same credit.
The prefix can be a photographer, collection or agency, so I thought it best nt to assume we can figure that out (I remember trying that a while ago and almost going :rage4:).

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
jamesgorrie: Yip, again, trying not to be presumptive, from what I've seen and been shown, this is always the case.

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
theefer: I meant I think you can write a single `case` as `case (Some("Rex Features"), _) | (_, Some(SlashRex()))` if you prefer. I don't mind particularly.

common-lib/src/test/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessorsTest.scala
theefer: :+1:

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
jamesgorrie: Ah, never knew that, I'll give it a go.

PR #1371, opened 2015-10-15
Stop propagation to allow popout in selection mode

jamesgorrie:
Seemed to be a sensible suggestion by picture editors as they often go through selecting loads of image, but want a closer look at one mid-way.

popout-on-selection


theefer:
:+1:

PR #1375, opened 2015-10-15
more contrast between persisted and non-persisted.

akash1810:
From the upload page, it isn't obvious when a non-system persisted image has been persisted or not. Now the behaviour is:

- non-persisted is dark grey with hover white
- persisted is light grey with hover white (as before)
- system persisted is filled lock (as before)

lock

### DIFF
--- a/kahuna/public/js/edits/archiver.html
+++ b/kahuna/public/js/edits/archiver.html
@@ -18,7 +18,7 @@
             </span>
         </gr-icon-label>
     </span>
-    <span ng:show="!archiver.isArchived" title="Add to library">
+    <span ng:show="!archiver.isArchived" title="Add to library" class="archiver--unarchived">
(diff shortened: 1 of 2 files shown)


tsop14:
Should this be the same on the preview page?

akash1810:
> Should this be the same on the preview page?

I tried that, but there wasn't enough contrast and it looked a bit odd (though a different shade can be used).

preview-lock

On the preview page, the lock is hidden on a non-persisted image until you hover, which I thought was clear. What do you think?

theefer:
If we could make the shade (possibly a different one) work on the results as well, feels like that'd be nice?

:+1:

akash1810:
Tweaked.

Uploads:
uploads

Preview:
preview

theefer:
:+1:

PR #1376, opened 2015-10-16
blocks drag'n'drop uploading of grid thumbnails

tsop14:
Checks if URIs match pattern of Grid thumbnail images' URIs, only uploads images that do not match. 

NOW :v: 
uploads1

BEFORE 
uploads2

### DIFF
--- a/kahuna/public/js/upload/dnd-uploader.js
+++ b/kahuna/public/js/upload/dnd-uploader.js
@@ -21,6 +21,8 @@ dndUploader.controller('DndUploaderCtrl',
                             apiPoll, witnessApi) {
 
     var ctrl = this;
+    //hack to prevent grid thumbnails being re-added to the grid for now
+    //TODO make generic - have API tell kahuna S3 buckets' domain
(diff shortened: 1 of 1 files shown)


kahuna/public/js/upload/dnd-uploader.js
```diff
@@ -21,9 +21,12 @@ dndUploader.controller('DndUploaderCtrl',
                             apiPoll, witnessApi) {
 
     var ctrl = this;
+    const gridThumbnailPattern = /https:\/\/media-service([0-9-a-z]+)thumbbucket([0-9-a-z]+)/;
```
theefer: The only thing about this pattern is that it's coupled with CF configuration that lives outside this repo. To make this more generic the API would have to tell kahuna what the domain our S3 buckets live under. Sounds like quite a faff, so maybe this is OK, but could we have a TODO to highlight that this is a bit of a hack?

tsop14:
@theefer comment added.

theefer:
:+1:

jamesgorrie:
I wonder if we could pick up on the fact before showing the dnd window and just do nothing?

PR #1378, opened 2015-10-19
removes unnecessary obs labels from suggestion list

tsop14:
As per email from Observer:
We don't need

obsfestivals
obsgalleries

so delete away.

Thanks


theefer:
:+1:

PR #1379, opened 2015-10-19
Adding family label for the family section of the Guardian

jamesgorrie:
As requested by Family section who are moving over to using this.


theefer:
:+1:

Issue #1377, opened 2015-10-18
Byline field disappears if not matching in multiple-select


paperboyo:
Hello,

In the RHS meta panel, when there are multiple images selected all fields list data if it’s the same for all images or say “Multiple [titles, descriptions, credits etc.]. Byline field disappears completely though, when the data in the field is not matching between all selected images. It should behave like all the other fields and just say “Multiple bylines” for when the data is not matching within the selection.

Regards
Mateusz

paperboyo:
Turns out, it doesn’t work only for images that one has no rights to edit metadata of. Still, it should work (works for all other fields). Thanks, @tsop14!

PR #1380, opened 2015-10-19
stops byline and copyright disappearing in side panel when multiple images selected

tsop14:
ng-if for 'byline' and 'copyright' tags checks if they exist in ctrl.rawMetadata.x (which returns an array of the different values) rather than ctrl.metadata.x (which returns 'undefined' if there is more than one image selected and their values are different).  

FIXES: #1377


theefer:
:+1:

paperboyo:
Thanks!

PR #1382, opened 2015-10-19
Map "crop" that covers the whole image to a "full" export

theefer:
That way we avoid confusion between the two, as a crop that contains the whole image is always a "full" export, regardless of how the user generated it (e.g. freeform selection of the entire image, or "full export" shortcut).

Note that the day we make "full" exports smarter e.g. to exclude silly agency banners, this logic will also need to be adapted.


jamesgorrie:
Nice :+1:

PR #1383, opened 2015-10-19
Give more memory to cropper and image-loader

theefer:
Don't see why not, and we're seeing OutOfMemory errors on cropper (and rarely on loader too).

Could be a leak, but unlikely because it happened shortly after a release.

Bit of a finger in the air update to be honest, but there's 3GB of RAM on these boxes so let's try this first.

### DIFF
--- a/cropper/conf/cropper.conf
+++ b/cropper/conf/cropper.conf
@@ -2,7 +2,7 @@ env USER=media-service
 env USER_HOME=/home/media-service
 env JAR=/home/media-service/app.jar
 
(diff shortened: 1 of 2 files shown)


akash1810:
:+1: 

Would this mean we're restricted to a minimum instance type?

theefer:
Yes.

On Wed, 21 Oct 2015, 11:14 Akash A <notifications@github.com> wrote:

> [image: :+1:]
>
> Would this mean we're restricted to a minimum instance type?
>
> —
> <https://github.com/guardian/grid/pull/1383#issuecomment-149846357>.
>

-- 

in England Number 908396

PR #1385, opened 2015-10-20
Shift selecting in preview

tsop14:
Allows click on check box to go through to image and so select it. 
Stops selection process when you click on a checkbox to prevent text highlighting. 

selecting

### DIFF
--- a/kahuna/public/js/search/results.js
+++ b/kahuna/public/js/search/results.js
@@ -407,7 +407,7 @@ results.controller('SearchResultsCtrl', [
         ctrl.imageHasBeenSelected = (image) => ctrl.selectedItems.has(image.uri);
 
         const toggleSelection = (image) => selection.toggle(image.uri);
-        ctrl.select           = (image) => selection.add(image.uri);
+        ctrl.select           = (image) => { selection.add(image.uri), $window.getSelection().empty(); }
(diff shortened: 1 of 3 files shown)


kahuna/public/js/search/results.js
```diff
@@ -407,7 +407,7 @@ results.controller('SearchResultsCtrl', [
         ctrl.imageHasBeenSelected = (image) => ctrl.selectedItems.has(image.uri);
 
         const toggleSelection = (image) => selection.toggle(image.uri);
-        ctrl.select           = (image) => selection.add(image.uri);
+        ctrl.select           = (image) => { selection.add(image.uri), $window.getSelection().empty(); }
         ctrl.deselect         = (image) => selection.remove(image.uri);
```
theefer: Does it need to be done here too in `deselect`?

theefer:
Nice one! :+1:

kahuna/public/stylesheets/main.css
```diff
@@ -779,6 +779,10 @@ input.search-query__input {
     z-index: 1;
 }
 
+.result__select--allow-shift-click {
```
akash1810: Class name is a little confusing to me - it seems too strongly coupled to the controller logic? Could make it more generic so it can be used elsewhere? e.g. `no-pointer-events`.

PR #1386, opened 2015-10-20
allow an empty metapanel on results page

jamesgorrie:
[Files with no whitespace](https://github.com/guardian/grid/pull/1386/files?w=1)

After chatting to @tsop14 and seeing people battle with the "automagic" way the metadata panel works, I thought it might be worth looking into making it work off more explicit actions. On this note, I have seen the mouseover to show panel potentially frustrating as if you don't click it fas enough and then click it, you lock it. 

There is also an option here to potentially tab this panel (related labels?) similar to the way we've discussed with @akash1810 on the image view with usage rights.

## before
meta-panel--no-empty

## :sun_with_face:after:sun_with_face: 
meta-panel--empty 1


theefer:
re click to reveal vs lock, @kenoir and I both thought maybe the show/hide toggle should be separate from the locking option. In other words, the show control turning into a lock control is a little confusing, and it may be easier to have lock/unlock as an extra control separate from the show/hide control. Just a thought, which could be added to your change here for extra clarity?

kahuna/public/js/components/gr-panel/gr-panel.js
```diff
@@ -170,16 +173,6 @@ grPanel.controller('GrPanel', [
               map(allEditable => allEditable.every(v => v === true));
         inject$($scope, selectionIsEditable$, ctrl, 'userCanEdit');
 
-
-        subscribe$($scope, selection.isEmpty$, isEmpty => {
-            if (isEmpty) {
-                panelService.unavailable(panelName, false);
-            } else {
-                panelService.available(panelName, false);
-            }
```
theefer: Are these two methods still used after this?

theefer: Found one instance above, ignore me.

kahuna/public/js/search/results.html
```diff
@@ -69,22 +69,25 @@
                 </ul>
             </div>
 
-            <div class="results-toolbar__right flex-container" ng:if="ctrl.inSelectionMode">
-                <div class="results-toolbar-item results-toolbar-item--right results-toolbar-item--static">
+            <div class="results-toolbar__right flex-container">
+                <div class="results-toolbar-item results-toolbar-item--right results-toolbar-item--static"
+                     ng:if="ctrl.selectionCount > 0">
```
theefer: Any way to avoid repeating this conditional? Possibly not worth the bother?

jamesgorrie: I didn't think it was a massive problem as it keeps separated...  Happy to move it into a `selection-actions` component if you think that is more clear.

theefer: Don't mind either way.

theefer:
:+1: 

Maybe we could even have a little more help instructions in the empty panel on how to select images. But could be a separate PR.

paperboyo:
> re click to reveal vs lock

No sure if I said how much I think it makes no sense to have to click there **every time** to show the panel after making a selection (currently, one can only download without opening a panel, nothing else). How much I would love open/lock state to be remembered for me (ideally, between machines). If I haven't - I just did. If I have - I did it twice for emphasis :)

I think all the behaviours around panels are so fundamental, that they better not be messed around too often/without a proper plan ("could be added to your change here for extra clarity").

I, for what's it worth (not much!), do not mind having open and lock in one place. I quite like it.

paperboyo:
@jamesgorrie I can see a close button in the bottom screengrab?

- replicates open/close

- while showing over green UR banner, it suggest resetting UR, not closing the panel

How many different things will it suggest showing over all these different tabbed panels?

jamesgorrie:
@theefer I've added very explicit actions. I wonder if auto-hiding the shown-but-not-locked still makes sense?

> I think all the behaviours around panels are so fundamental, that they better not be messed around too often/without a proper plan ("could be added to your change here for extra clarity").

The reason this is being done is to try come up with that plan. The way it has been setup so far seems to have caught everyone who I see using it out.

> I can see a close button in the bottom screengrab?

It's gone now the buttons are more explicit.

Did anyone every ask for the slightly overlayed version or was that just our solution to not having someone lose there position when the images are reordered?

jamesgorrie:
explicit-panel

paperboyo:
> The way it has been setup so far seems to have caught everyone who I see using it out.

Yeah, I was afraid of that... :( Having two huge buttons there, though... (even less space in the toolbar for the future, ugly, redundant). My biggest pain with it was to have to open it every time. I think, the below would sort many of the problems (is the underlying problem?):

> lose there position when the images are reordered

That would be soo nice to have sorted. It can happen not only with panels pushing images, but ultimately with search/filters as well...

never loose position > can haz panel always open on selection/can haz it remember me preference

paperboyo:
Look, obviously, I'm skewed. I still think Lightroom's way is superior, if only because you do not have to travel to the top with your mouse (and a triangle takes less space). It can show/hide and lock. It cannot auto-open (I have nothing against that.)

Is it discoverable? No. Most probably, much less discoverable than Grid's implementation. But it only takes seeing it once. (+ keyboard shortcuts!). [Okay, the've added onboarding animations recently...).

I may be pushing for what I want, because I've seen it working. But forget about that (and me)! If anything, I'm saying: it's okay to struggle with some stuff. Even if every single person struggles. Even if millions struggle (let alone hundreds). If you are sure it's the right way - just make it so and wait. And for the love of God - do not measure!

That's that when it comes to me having a career in UX :D

alastairjardine:
I would be very happy to have a proper discussion about this – we've not thought about the usability of the tool enough. 

I'm not happy with the lock button, so please don't release.

alastairjardine:
Oh, and an empty state needs to be improved – please let's chat!

jamesgorrie:
Happy to revert out the changes then to the usability.

The actual PR request here was to have the panel not automatically close on the clearing of a selection as that was a usability for quite a lot of people.

Reverting to original PR.

jamesgorrie:
> Maybe we could even have a little more help instructions in the empty panel on how to select images. But could be a separate PR.

Nice idea

jamesgorrie:
> the empty state needs to be improved

> Maybe we could even have a little more help instructions in the empty panel on how to select images. But could be a separate PR.

?

paperboyo:
> Nice idea

Since I've already stamped all over everybody's work here, contradicted the Architect and insulted the whole profession, I would politely disagree here as well (if only for consistency :) ).

Instead of useful-just-once instructions/help, that IMHO belongs in three-dot-menu or once-shown-never-seen-again onboarding experience, I would imagine having here some useful statistics on the whole of (currently unselected) imagery (depending on current search).

What exactly would you propose then, boyo? No idea, TBH. Also, most probably, hard/impossible to compute all of that knowing how long it takes to compute it for just three fields on a selection of 50...

See?! Insult again!

Peace :metal:

jamesgorrie:
> I would imagine having here some useful statistics on the whole of (currently unselected) imagery

The PR description suggests that this is what the PR is moving towards (potentially not the exact content, but useful content non the less).

alastairjardine:
> Instead of useful-just-once instructions/help, that IMHO belongs in three-dot-menu or once-shown-never-seen-again on-boarding experience, I would imagine having here some useful statistics on the whole of (currently unselected) imagery (depending on current search).

Absolutely. the angle we should be going down is "If it's going to be open, it has to add value". Just showing instructional text is a cop out.

> The PR description suggests that this is what the PR is moving towards (potentially not the exact content, but useful content non the less).

This is why talking about larger features first and envisioning (NOTE: _not_ build) them fully is a good thing – half baked featuritis is a plague, and is to the detriment of overall software quality. Get to the end result, work out how to deliver them. Not take the first step and see where we go. It might be contradictory to how this team is set up, but I firmly believe getting to an end result conceptually is a stronger position to work from. There are way of validating without building to support that aims too.

alastairjardine:
> Is it discoverable? No. Most probably, much less discoverable than Grid's implementation. But it only takes seeing it once. (+ keyboard shortcuts!). [Okay, the've added onboarding animations recently...).

I do agree, @paperboyo. Lightroom is a pro tool, which many will use for 100s of hours, not just tens. The media service is in an interesting place. On the one hand, it needs to support complex workflows that picture editors need to manage a library. On the other hand, it needs to support people that fit in the  WJNAGDPFMA (Who Just Need A Goddamn Picture For My Article) category. They're two, _very_ different and complex needs wrapped up into a single solution, and I've made no bones about hiding those contradictory aims.

I would be fascinated to see how many people use Lightroom full screen vs not full screen, because the panel of which you speak is fucking fiddly to hit when not full screen, and is moderately annoying if it pops out accidentally by mousing through.

paperboyo:
> because the panel of which you speak is fucking fiddly to hit when not full screen

True! True also that Grid is much more likely to _not_ be used full screen (modal, windows side-by-side). That's why keyboard shortcuts would help. But yes - that is a legitimate argument _against_ full-length active edge _à la_ Lr. But, it is not one for two buttons, IMHO.

> On the one hand, it needs to support complex workflows

&

> On the other hand, it needs to support people

While it is a challenge (and fun!), I can't see a contradiction here. Nor do I see a reason to have two UI modes (never liked the idea). And I can testify, that I'm seeing both categories of users gaining greatly by the Grid. And both having some issues with it. Fun!

There is plenty of stuff that can be done (and is being done) to help both without hurting neither.

paperboyo:
> moderately annoying if it pops out accidentally by mousing through

Verging on off-topic (although I think - not), that helps:

lr

alastairjardine:
> True! True also that Grid is much more likely to not be used full screen (modal, windows side-by-side). That's why keyboard shortcuts would help. But yes - that is a legitimate argument against full-length active edge à la Lr. But, it is not one for two buttons, IMHO.

+++++++++ couldn't agree more.

jamesgorrie:
@alastairjardine & @paperboyo 
Moving the conversation on topic again (not negating, but bikeshedding the rest).

This PR just respects the lock. If you have locked the panel and clear your selection, the panel stays.

jamesgorrie:
I personally think the smaller change here is useful to users as I have seen it catch them a few times - but if @alastairjardine thinks it's too small a change i.e. not fully fledged, happy to close, or better, find what the more fledged solution is to add it here.

paperboyo:
@jamesgorrie Not that anybody should care, but I have/had nothing against the main aim of this here PR (only against sneaking-in the second - lock - button). Even more so, because https://github.com/guardian/grid/pull/1387 seems to depend on this, right? - so it wouldn't even be an "empty panel" any more...

jamesgorrie:
#1387 was put there as a POC. I will close it once it's been looked at.
It is still an empty panel.

paperboyo:
> It is still an empty panel

Maybe so but at least behaving as it should according to lock status instead of illogically, no?

jamesgorrie:
@paperboyo that's the point :point_up:

jamesgorrie:
Still uncertain as to whether this is blocked by the UX (basically the empty panel) or not?

theefer:
Is it worth rephrasing what this is solving?

How much of the confusion is due to the dual-function reveal/lock button, which isn't actually related to this change?

jamesgorrie:
> @alastairjardine & @paperboyo 
> Moving the conversation on topic again (not negating, but parking the rest).
> This PR just respects the lock. If you have locked the panel and clear your selection, the panel stays.

[said here](https://github.com/guardian/grid/pull/1386#issuecomment-149852597)

At the moment the description is correct - we just don't close the panel on selection clearing.

jamesgorrie:
One last bump until :closed_book: if people don't think it's worth it.

jamesgorrie:
Merging as it's a step forward.

PR #1388, opened 2015-10-22
use removeAllRanges for cross browser support

jamesgorrie:
To support Chrome and FF.


theefer:
:+1: if tested on both

tsop14:
Nice!

PR #1389, opened 2015-10-22
add obs and culture labels

jamesgorrie:
As much as we're probably going to be changing this - this will get more people using it, thus we can see how they use it.

@tsop14

### DIFF
--- a/media-api/app/controllers/MediaApi.scala
+++ b/media-api/app/controllers/MediaApi.scala
@@ -322,14 +322,20 @@ object MediaApi extends Controller with ArgoHelpers {
   }
 
   def suggestLabels(q: Option[String]) = Authenticated {
+    
     val pseudoFamousLabels = List(
-      "obsbizcash", "obscomment", "obsfocus", "obsforeignnews",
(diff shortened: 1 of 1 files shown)


theefer:
:+1: how big does it look now ?

jamesgorrie:
Big - but I think it's not really a problem worth solving just yet as we do have prefix search.
screen shot 2015-10-22 at 13 42 01

alastairjardine:
Okay, I'll bite: "thus we can see how they use it."

What will you be looking for? What does good look like? How will you filter the feedback? What are you hypotheses?

jamesgorrie:
> What will you be looking for? What does good look like? 

For now - people using it - this has unblocked some desks and people coming on board, and these were added for the rest. What would be better is for people to feedback what is wrong to add to the list @blishen has already collated.

> How will you filter the feedback?

The same way we have been for a while which is to go and ask for it, or, as people are now using it and have also been told to email us with feedback - from there.

> What are your hypotheses?

Most of what was discussed this morning. Clearer hierarchy will be needed, applying these to images is to tricky etc.

PR #1391, opened 2015-10-22
exclude newspix from getty

jamesgorrie:
We don't have a contract with them.

### DIFF
--- a/media-api/app/lib/elasticsearch/SearchFilters.scala
+++ b/media-api/app/lib/elasticsearch/SearchFilters.scala
@@ -18,7 +18,7 @@ trait SearchFilters extends ImageFields {
   val invalidFilter = Config.requiredMetadata.map(metadataField).toNel.map(filters.anyMissing)
 
   // New Cost Model
-  import UsageRightsConfig.{ suppliersCollectionExcl, freeSuppliers, payGettySourceList }
+  import UsageRightsConfig.{ suppliersCollectionExcl, freeSuppliers }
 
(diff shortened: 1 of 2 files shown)


media-api/app/lib/elasticsearch/SearchFilters.scala
```diff
@@ -18,7 +18,7 @@ trait SearchFilters extends ImageFields {
   val invalidFilter = Config.requiredMetadata.map(metadataField).toNel.map(filters.anyMissing)
 
   // New Cost Model
-  import UsageRightsConfig.{ suppliersCollectionExcl, freeSuppliers, payGettySourceList }
+  import UsageRightsConfig.{ suppliersCollectionExcl, freeSuppliers }
```
tsop14: Why did you remove 'payGettySourceList'?

jamesgorrie: It wasn't used.

theefer:
:+1: but please let's check if we're out of sync with the source list of excluded collections, or if the source list is wrong.

jamesgorrie:
I was wrong - it's newpix international.
They annoyingly* attached getty data then email pictures in.

* I say annoying, they actually attach composition data etc which is interesting, annoying that they do it using Getty's XMP

theefer:
What's the best solution in that case?  Should we still exclude them from the Getty parser for the time being?

jamesgorrie:
That's what I am looking at now - I wonder if other might use it too so when they upload to Getty (?) it get's categorised the way they want it.

jamesgorrie:
#1395

PR #1392, opened 2015-10-22
Better Lambda fail states and reporting

kenoir:
- Don't fail Lambda function when upload fails to avoid Lambda retries
- Report successful uploads & failures to CW

**Next steps:**
- [ ] Ask for all suppliers in Test.
- [ ] Move single supplier in Prod when we're satisfied with ^.

### DIFF
--- a/s3watcher/lib/Metrics.js
+++ b/s3watcher/lib/Metrics.js
@@ -8,17 +8,19 @@ module.exports = {
             "Value" : uploadResult.uploadedBy
         }]
 
+        const timestamp = new Date;
+
         return {
             MetricData: [{
                 MetricName: metricName,
                 Dimensions: dimensions,
-                Timestamp: new Date,
+                Timestamp: timestamp,
                 Unit: "Count",
                 Value: 1
             },
             {
                 MetricName: metricName,
-                Timestamp: new Date,
+                Timestamp: timestamp,
                 Unit: "Count",
                 Value: 1
             }],
--- a/s3watcher/lib/Metrics.js
+++ b/s3watcher/lib/Metrics.js
@@ -3,16 +3,21 @@ module.exports = {
         const metricName = uploadResult.succeeded ?
             "UploadedImages" : "FailedUploads";
 
-        const dimensions = uploadResult.succeeded ?
-            [{
-                "Name" : "UploadedBy",
-                "Value" : uploadResult.uploadedBy
-            }] : [];
+        const dimensions = [{
+            "Name" : "UploadedBy",
+            "Value" : uploadResult.uploadedBy
+        }]
 
         return {
             MetricData: [{
                 MetricName: metricName,
                 Dimensions: dimensions,
+                Timestamp: new Date,
+                Unit: "Count",
+                Value: 1
+            },
+            {
+                MetricName: metricName,
                 Timestamp: new Date,
                 Unit: "Count",
                 Value: 1
--- a/s3watcher/lib/Metrics.js
+++ b/s3watcher/lib/Metrics.js
@@ -1,0 +1,23 @@
+module.exports = {
+    create: function (uploadResult) {
+        const metricName = uploadResult.succeeded ?
+            "UploadedImages" : "FailedUploads";
+
+        const dimensions = uploadResult.succeeded ?
+            [{
+                "Name" : "UploadedBy",
+                "Value" : uploadResult.uploadedBy
+            }] : [];
+
+        return {
+            MetricData: [{
+                MetricName: metricName,
+                Dimensions: dimensions,
+                Timestamp: new Date,
+                Unit: "Count",
+                Value: 1
+            }],
+            Namespace: uploadResult.stage + "/S3watcher"
+        };
+    }
+}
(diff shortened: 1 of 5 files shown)


s3watcher/lib/Upload.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
          const headers = {
             'Content-Length': s3Event.size,
             'Content-Type': 'application/octet-stream',
-            'X-Gu-Media-Key': config.apiKey
+            'X-Gu-Media-Key': config.stage == "PROD" ? "xxx" : config.apiKey
```
jamesgorrie: `===`?

s3watcher/lib/Upload.js
```diff
@@ -44,16 +54,12 @@ module.exports = {
         const uploadRequest = request.post(options);
 
         return Rx.Observable.create(function(observer){
+            // check mime type, if not jpeg notify CW, stop!
```
jamesgorrie: Is this a TODO?

s3watcher/lib/Metrics.js
```diff
@@ -0,0 +1,30 @@
...
+            "UploadedImages" : "FailedUploads";
+
+        const dimensions = [{
+            "Name" : "UploadedBy",
+            "Value" : uploadResult.uploadedBy
+        }]
+
+        const timestamp = new Date;
+
+        return {
```
jamesgorrie: If I remember somewhere - CloudWatch can be throttled.

Would it be an idea to store a queue here and push out data points on a certain size or time the way we do in the other apps?

Or would that not work as we're lambdering?

theefer: Don't think you could share anything between lambda executions..?

s3watcher/lib/Upload.js
theefer: `"xxx"`? why not always the configured api key?

s3watcher/lib/Upload.js
```diff
@@ -44,16 +54,12 @@ module.exports = {
...
-                } else {
-                    const failMessage =
-                        "Failed upload with code: " + response.statusCode;
-                    observer.onError(failMessage);
-                }
-           });
-           uploadRequest.on("error", observer.onError.bind(observer));
+                observer.onNext(uploadResult(response));
+            });
+            uploadRequest.on("error", observer.onError.bind(observer));
```
theefer: Does the request error if a non 2xx response code (e.g. 415, 5xx)?

theefer:
Looks good generally, a few questions above. Could you outline how this affects the different kinds of errors (415, other) and if just an intermediate step, what the next steps are?

s3watcher/lib/Metrics.js
kenoir: Internal application state cannot be shared between executions.

Looks like Cloudwatch Logs have a maximum `POST` rate but there is none specified for `PutMetricData`. 

http://docs.aws.amazon.com/AmazonCloudWatch/latest/DeveloperGuide/cloudwatch_limits.html

s3watcher/lib/Upload.js
kenoir: Brain fart trying to hide the key in logs, this will do that but also not work at all so :crying_cat_face:

s3watcher/lib/Upload.js
kenoir: It was. But i'm not doing it, so i'll remove it.

s3watcher/lib/Upload.js
kenoir: Couldn't find the definitive answer to this in the docs. After some fiddling in the node console and figuring out how to simulate error responses with `netcat`:

```sh
while true; do { echo -e 'HTTP/1.1 500 NOPE\r\n'; } | nc -l 8080; done
```

It does not throw an error as long as the server returns a HTTP response (regardless of whether it's a 4xx or 5xx).

kenoir: The result of this is that while timeouts will cause a Lambda failure and consequently retries. Any response from the server that is not `202` will log an error to Cloudwatch but not cause a Lambda failure, so no retry.

kenoir:
@seb addressed what happens in error states above, (it's as we discussed I think). I decided not to fail fast on non jpeg mime types as it's not straightforward to add and can be addressed by modifying the Lambda trigger events if necessary (it's possible to only send events on *.jpg/*.jpeg files). Though our aspirations to handle pngs and the low frequency of "bad" files make me feel that it's unnecessary complexity.

theefer:
Do you think we would we want to retry (and not delete the file yet) if the upload fails with something else than 415 (e.g. 5xx)? Not sure myself.

Else looks good to me :+1:

kenoir:
@theefer think we should respect the API (i.e. a 4xx or 5xx will always be so), whereas a timeout is worth retrying.

theefer:
> i.e. a 4xx or 5xx will always be so

A 4xx will indeed be pointless to retry (client error, so always going to fail unless client changes), but in theory a 5xx is a temporary server error. We can deal with this later though.

PR #1394, opened 2015-10-23
add validation to search params

jamesgorrie:
In a move towards paginating labels - I didn't want to open up aggregate endpoints to infinite page sizes.

### DIFF
--- a/media-api/app/controllers/MediaApi.scala
+++ b/media-api/app/controllers/MediaApi.scala
@@ -252,7 +252,7 @@ object MediaApi extends Controller with ArgoHelpers {
 
     SearchParams.validate(searchParams).fold(
       // TODO: respondErrorCollection?
-      errors => Future.successful(respondError(UnprocessableEntity, InvalidUrlParams.errorKey,
+      errors => Future.successful(respondError(UnprocessableEntity, InvalidUriParams.errorKey,
         // Annoyingly `NonEmptyList` and `IList` don't have `mkString`
         errors.map(_.message).list.reduce(_+ ", " +_), List(searchLink))
       ),
@@ -428,9 +428,9 @@ case class SearchParams(
   persisted: Option[Boolean]
 )
 
-case class InvalidUrlParams(message: String) extends Throwable
-object InvalidUrlParams {
-  val errorKey = "invalid-url-parameters"
+case class InvalidUriParams(message: String) extends Throwable
+object InvalidUriParams {
+  val errorKey = "invalid-uri-parameters"
 }
 
 object SearchParams {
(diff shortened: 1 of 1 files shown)


media-api/app/controllers/MediaApi.scala
```diff
@@ -243,8 +247,21 @@ object MediaApi extends Controller with ArgoHelpers {
...
+      // TODO: respondErrorCollection?
+      errors => Future.successful(respondError(UnprocessableEntity, InvalidUrlParams.errorKey,
+        // Annoyingly `NonEmptyList` and `IList` don't have `mkString`
+        errors.map(_.message).list.reduce(_+ ", " +_), List(searchLink))
+      ),
+      params => respondSuccess(params)
+    )
   }
 
+  List("1", "2").mkString(",")
```
theefer: :-)

media-api/app/controllers/MediaApi.scala
```diff
@@ -243,8 +247,21 @@ object MediaApi extends Controller with ArgoHelpers {
...
       links = suggestedLabelsLink :: List(prevLink, nextLink, relatedLabelsLink).flatten
     } yield respondCollection(imageEntities, Some(searchParams.offset), Some(totalCount), links)
+
+    val searchParams = SearchParams(request)
+
+    SearchParams.validate(searchParams).fold(
+      // TODO: respondErrorCollection?
+      errors => Future.successful(respondError(UnprocessableEntity, InvalidUrlParams.errorKey,
+        // Annoyingly `NonEmptyList` and `IList` don't have `mkString`
+        errors.map(_.message).list.reduce(_+ ", " +_), List(searchLink))
```
theefer: I've been wondering if we should make it standard to allow returning multiple errors, instead of just one. OT for this though.

media-api/app/controllers/MediaApi.scala
```diff
@@ -411,6 +428,11 @@ case class SearchParams(
   persisted: Option[Boolean]
 )
 
+case class InvalidUrlParams(message: String) extends Throwable
+object InvalidUrlParams {
+  val errorKey = "invalid-url-parameters"
```
theefer: Can we make this `uri` for overall consistency?

media-api/app/controllers/MediaApi.scala
jamesgorrie: Yea, I was thinking of the client  - having to deal with multiple error responses would be annoying. I would argue multiple, as that can be the case.

media-api/app/controllers/MediaApi.scala
jamesgorrie: :-(

media-api/app/controllers/MediaApi.scala
jamesgorrie: Saying that - I was wondering if we could have multiple error keys? I wouldn't image so, as that starts edging towards multiple error Status'. 

It feels like we should return 1 error with multiple reasons.

media-api/app/controllers/MediaApi.scala
```diff
@@ -481,6 +503,23 @@ object SearchParams {
...
+    type SearchParamValidations = ValidationNel[InvalidUrlParams, SearchParams]
+
+    def validate(searchParams: SearchParams): SearchParamValidations = {
+      // we just need to return the first `searchParams` as we don't need to manipulate them
+      // TODO: try reduce these
+      (validateLength(searchParams).toValidationNel |@| validateOffset(searchParams).toValidationNel)((s1, s2) => s1)
+    }
+
+    def validateOffset(searchParams: SearchParams): SearchParamValidation = {
+      if (searchParams.offset < 0) InvalidUrlParams("offset cannot be less than 0").failure else searchParams.success
```
theefer: Parameterise into a variable?

theefer:
:+1:

PR #1395, opened 2015-10-23
exclude newpiz images even though they have getty data

jamesgorrie:
As per #1391 - it seems Newspix Int are not affiliated with Newspix Aus (who are affiliated with Getty), but they do attach Getty metadata. I can only theorise why, but not 100% sure.


theefer:
:+1: can we just reindex matching images?

jamesgorrie:
> reindex

Yip.

PR #1396, opened 2015-10-23
creates print layout for image page

tsop14:
[relevant main.css changes - without whitespace](https://github.com/guardian/grid/pull/1396/files?w=1)


theefer:
Cool! Bit confusing why GitHub is not showing the diff though?

akash1810:
Nice!

Can't see full diff and GH, strangely, doesn't allow commenting on the no-whitespace diff, so can't comment inline :crying_cat_face:.

Is the `@media screen` needed? It looks like the print style isn't restricted to the image page - how does the uploads page look? Do we need `!important`?

tsop14:
> Is the @media screen needed?

Without `@media screen` enclosing the pre-existing css rules, I break printing in Chrome (FF is fine). Do you have any ideas about this?

> It looks like the print style isn't restricted to the image page - how does the uploads page look?

Do you know how to apply the style sheet to just one page? The other pages don't look great - are they printed? 

 > Do we need !important?

:person_frowning: the component stylesheets load after main.css and overwrite the print styling.

jamesgorrie:
Nice.

> are they printed

I haven't had anyone ask about it.

Have you run these passed the people who use it?

tsop14:
No, who is best to email?

akash1810:
> Do you have any ideas about this?

Does the `media` attribute on the `link` element behave any better?

> Do you know how to apply the style sheet to just one page?

You could wrap [this page](
https://github.com/guardian/grid/blob/master/kahuna/public/js/image/view.html) in a div, apply a class and target it? Might not be that elegant though...

> The other pages don't look great - are they printed?

I've only heard one person asking to print a contact sheet from the results page.

theefer:
Roger T was interested in printing.

tsop14:
image the grid 💁image page

by- pacifi-si-rex shutterstock the grid 💁results page

Description div and images have a max-height in the hope most things fit on one page - couldn't find a way to make image height fill remaining space or use % for heights. 

Final layout varies between FF and Chrome. 

:deciduous_tree:

kahuna/public/stylesheets/main.css
```diff
@@ -2096,3 +2097,63 @@ FIXME: what to do with touch devices
 .radio-list__label-value {
     padding: 5px 10px;
 }
+
+
+
+@media print {
+
+    * {
+        color: black !important;
```
theefer: Can we leave a comment re why everything is important?

kahuna/public/stylesheets/main.css
```diff
@@ -2096,3 +2097,63 @@ FIXME: what to do with touch devices
 .radio-list__label-value {
     padding: 5px 10px;
 }
+
+
+
+@media print {
+
+    * {
```
theefer: This is a very expensive selector (selects every single element on the page). I don't know enough of how `@media` works to know when it is applied, but is there another way?

theefer:
Minor comments, but generally :+1:

PR #1397, opened 2015-10-26
Use the correct name for the Life picture collection

jamesgorrie:
Search for `source:"The LIFE Picture Collection Editorial"` and `source:"The LIFE Picture Collection"`. We have none in the former.

These are not included in out contract,


theefer:
:+1: Was it right/wrong in the spreadsheet we get from Getty? If they don't match, is it worth leaving a comment here for the person who'll sync up the lists next time?

jamesgorrie:
Good call - added

PR #1398, opened 2015-10-26
decode the URI params at the right place

jamesgorrie:
We aren't decoding the URL onload so when we looad a query with slashes in, it isn't decoded.
Done it universally as there might be other stuff from the query.

e.g. `/search?query=this%252Fthat`

### DIFF
--- a/kahuna/public/js/search/query.js
+++ b/kahuna/public/js/search/query.js
@@ -23,7 +23,7 @@ query.controller('SearchQueryCtrl',
                  ['$scope', '$state', '$stateParams', 'onValChange', 'mediaApi', 'track',
                  function($scope, $state, $stateParams, onValChange , mediaApi, track) {
 
-    var ctrl = this;
+    let ctrl = this;
 
     ctrl.ordering = {
         orderBy: $stateParams.orderBy
@@ -31,12 +31,6 @@ query.controller('SearchQueryCtrl',
 
     $scope.$watch(() => ctrl.ordering.orderBy, onValChange(newVal => {
         $state.go('search.results', {orderBy: newVal});
-    }));
-
-    // Boring - stops slashes being encoded in input - might be fixed with:
-    // https://github.com/angular-ui/ui-router/issues/1759
-    $scope.$watch(() => ctrl.filter.query, onValChange(newVal => {
-        ctrl.filter.query = angular.isDefined(newVal) ? decodeURIComponent(newVal) : '';
     }));
 
     ctrl.filter = {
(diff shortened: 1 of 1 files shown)


kahuna/public/js/search/query.js
```diff
@@ -23,7 +23,7 @@ query.controller('SearchQueryCtrl',
                  ['$scope', '$state', '$stateParams', 'onValChange', 'mediaApi', 'track',
                  function($scope, $state, $stateParams, onValChange , mediaApi, track) {
 
-    var ctrl = this;
```
theefer: Note that `const` would work too as it's mutated but not reassigned?

theefer:
:+1:

kahuna/public/js/search/query.js
jamesgorrie: I don't think it would as wee add properties to it. e.g. [`ordering`](https://github.com/guardian/grid/blob/decode-uri/kahuna/public/js/search/query.js#L28-L30)

theefer: `const` **doesn't** enforce immutability, only that the variable isn't reassigned. (It's the same as `val` in Scala, vs `var`, btw. Only that most types in Scala are immutable)

jamesgorrie: Ah, thanks.

jamesgorrie:
@theefer I've updated the PR to not decode `undefined` as `decodeURIComponent(undefined) === "undefined"`.

kahuna/public/js/search/query.js
```diff
@@ -62,11 +56,13 @@ query.controller('SearchQueryCtrl',
     Object.keys($stateParams)
           .forEach(setAndWatchParam);
 
-    // pass undefined to the state on empty to remove the QueryString
-    function valOrUndefined(str) { return str || undefined; }
+    // URL parameters are not decoded when taken out of the params.
+    // Might be fixed with: https://github.com/angular-ui/ui-router/issues/1759
+    // Pass undefined to the state on empty to remove the QueryString
+    function valOrUndefined(str) { return str ? decodeURIComponent(str) : undefined; }
```
theefer: Just to check that we do want to `decodeURIComponent` both when `ctrl.filter[key]` is initialised and updated below? (does seem like both could be unified, but this it OT here!)

jamesgorrie: `setAndWatchParam` implies that - we need to do this on load, then when it changes. Basically whenever we set it.
I would argue the `setAndWatchParam` paradigm could use a little work, but we do want it in both places here.

theefer: OK :+1:

theefer:
:+1: if tested

jamesgorrie:
tested

PR #1399, opened 2015-10-27
Add basic route as base route

jamesgorrie:
__WIP__

More exploratory - but also something that will win us a lot of users.

I think - the more I look at it that we might have one application state that is resolved through to a few other child states.

__URL structure:__
`/search</image/:imageId</crop>>{?q&uploadedBy...}`

__TODO__
- [x] Cropping
- [x] ~~Potentially stop call to image on image view activation as we already have all the information~~ should probably be some form of other caching
- [ ] What to do when we come from uploads
- [x] Stop scrolling on results page when you have the image overlay
- [ ] See if we can remove references to `isReloadingPreviousSearch`
- [ ] Pause new image lookup when we aren't in the direct `search.results` route
- [ ] Load single image without results
- [ ] ^ If you are from results show next / prev


jamesgorrie:
[Video](https://www.youtube.com/watch?v=LR4LfhqKkUI).
Note the URL. Seems to be quite a lot smoother than the current deep state redirect too.

jamesgorrie:
The route runs all the way through to crop now.

theefer:
Looks really promising. Would be interested to catch up on the routing implications and general approach.

jamesgorrie:
Also remembers selection:
remember-selection

jamesgorrie:
@theefer - definitely - let's catch up tomorrow.
The main structural thing I would like to clear up is if we define a `view` e.g. something that replies on these resolves to occur.

jamesgorrie:
@theefer It's starting to look like we might need to think about moving the search into a resolver, almost making it part of the application state. I've started making a move on it. After talking to @akash1810 we both think the results could do with a refactor - it might be good to chat about whether this is a valuable enough task to undertake.

kahuna/public/js/query/query.js
```diff
@@ -82,7 +91,7 @@ query.controller('SearchQueryCtrl',
 
     $scope.$watchCollection(() => ctrl.filter, onValChange(filter => {
         filter.uploadedBy = filter.uploadedByMe ? ctrl.user.email : undefined;
-        $state.go('search.results', filter);
+        //$state.go('search.results', filter);
```
jamesgorrie: This is quite a lot smoother as we aren't reloading the state on every search, we just update the searchParams stream, which is then mapped into the search results.

kahuna/public/js/results/results.js
```diff
@@ -221,58 +186,46 @@ results.controller('SearchResultsCtrl', [
...
-            then(() => delay(30)).
-            then(() => scrollPosition.resume($stateParams)).
-            then(scrollPosition.clear);
+        // TODO: reinstate this
+        //ctrl.suggestedLabelSearch = q =>
+        //    ctrl.searched.then(images =>
+        //        images.follow('suggested-labels').get({q}).then(labels => labels.data)
+        //    ).catch(() => []);
+
+        ctrl.loadRange = (start, end) => { Search.loadRange(start, end); };
```
jamesgorrie: TODO

theefer:
> we might need to think about moving the search into a resolver

OK, happy to have a chat, and if we do, would be good to break such work into separate incremental changes.

jamesgorrie:
Closing to break into smaller parts.

PR #1402, opened 2015-10-28
Also sync originalMetadata to store the cleaned version

### LINKED ISSUE #971: "Reuters" becomes "REUTERS" on any user interaction
Hello,

Ok, I don't know if "any", but:

1. Run https://media.gutools.co.uk/search?query=credit:Reuters
2. Add a label or archive image
3. Credit for that image becomes "REUTERS", so that it will disappear from the above search on refresh

Not a massive thing, just slightly annoying and triggers a lot of manual editing or wrong credit on live site. Also, baffling as to why it happens in the first place... :smile:

Regards
Mateusz

theefer:
This should fix #971.

Essentially any cleanup we do on the metadata when loading an image should be stored both on `originalMetadata` and `metadata`. A mismatch between the two (where only `metadata` has the cleaned version) caused any edit to recomputed `metadata` from `originalMetadata`, which then erased any cleaning done at load time.

This was an oversight as we [already do this](https://github.com/guardian/grid/blob/master/media-api/app/controllers/MediaApi.scala#L199-L202) when reindexing. Reindexing images fix the issue.

Note that the discrepancy will be widespread in all the images we have. It can be fixed by reindexing all images, now or later.

### DIFF
--- a/image-loader/app/model/ImageUpload.scala
+++ b/image-loader/app/model/ImageUpload.scala
@@ -52,8 +52,11 @@ case object ImageUpload {
         baseImage      = createImage(uploadRequest, sourceAsset, thumbAsset, fileMetadata, cleanMetadata)
         processedImage = SupplierProcessors.process(baseImage)
 
-        // FIXME: dirty hack to sync the originalUsageRights as well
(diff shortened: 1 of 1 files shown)


jamesgorrie:
:+1: Good catch.

PR #1403, opened 2015-10-28
Remove all remaining mentions of "archive" and replaced with Library

theefer:
In particular, the meta panel still read "archive". Updated:

image
image
image

The end goal would be to move them as actions in the action bar alongside Download etc, but requires more work due to the various states this can be in.

Also capitalised Library everywhere and removed `?` in `Remove from Library`.

### DIFF
--- a/kahuna/public/js/components/gr-image-persist-status/gr-image-persist-status.html
+++ b/kahuna/public/js/components/gr-image-persist-status/gr-image-persist-status.html
@@ -1,6 +1,6 @@
 <gr-icon-label ng:if="! ctrl.states.canArchive" gr-icon="lock"
-    title="Kept in library because image is: {{ctrl.states.persistedReasons}}.">
-    <span ng:show="ctrl.withText">Kept in library</span>
+    title="Kept in Library because image is: {{ctrl.states.persistedReasons}}.">
+    <span ng:show="ctrl.withText">Kept in Library</span>
 </gr-icon-label>
 
 <ui-archiver ng:if="ctrl.states.canArchive"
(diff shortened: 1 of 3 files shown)


akash1810:
Can we replace the stars with the lock too?

theefer:
Ah yes good point. How would we denote the "half star" though..?

akash1810:
Solid lock would be the most logical imo. Solid lock = (hard) persisted. Outline lock = (soft) persisted, but is too similar to unlocked.

tsop14:
Is it confusing that in other places the lock is something you click (image previews) or part of a button (top bar)?

akash1810:
> Is it confusing that in other places the lock is something you click (image previews) or part of a button (top bar)?

Potentially. I think this confusion will go away once these buttons are in the toolbar, however.

jamesgorrie:
@theefer we could just use the [half-selected thing](https://www.google.com/design/icons/#ic_indeterminate_check_box). 

I still think adding a "this image will be kept for another 5 days" would help people understand what this actually is, but I suppose they do because of Picdar.

alastairjardine:
> Potentially. I think this confusion will go away once these buttons are in the toolbar, however.

Any progress on this? It feels like a long time in coming...

jamesgorrie:
> these buttons are in the toolbar

We're going to have a double-lock situation there though...

alastairjardine:
This lock thing feels like it's getting contorted, and not ideal. Any chance you can have a chat with @akemitakagi who's pretty good at making icons... Material design's have reached a limitation here.

theefer:
Updated to use the lock icon as suggested:

image

Going to release this because it's an improvement regardless.

We can talk about moving to the action bar and improving the iconography separately.

alastairjardine:
Hmm, okay. Again, I'm not convinced it's as good as it can be. Standards!

theefer:
No, we all agree, it's definitely not as good as good as it can be.

I've just been to see Mee-Lai who complained about a number of things, including these exact two points (lack of clarity of the three archive states, and having to go in meta panel to batch archive). It remains on our list of priorities.

PR #1405, opened 2015-10-29
Crop shortcut

tsop14:
Creates a keyboard shortcut on the crop page - enter key crops the image.

### DIFF
--- a/kahuna/public/js/crop/controller.js
+++ b/kahuna/public/js/crop/controller.js
@@ -103,7 +103,7 @@ crop.controller('ImageCropCtrl',
         }).finally(() => {
             ctrl.cropping = false;
         });
-    }
+    };
 
      function cropReturnKeyShortcut(event) {
          // check if ENTER key
--- a/kahuna/public/js/crop/controller.js
+++ b/kahuna/public/js/crop/controller.js
@@ -4,13 +4,14 @@ var crop = angular.module('kahuna.crop.controller', []);
 
 crop.controller('ImageCropCtrl',
                 ['$scope', '$rootScope', '$stateParams', '$state',
-                 '$filter', 'mediaApi', 'mediaCropper',
+                 '$filter', '$document', 'mediaApi', 'mediaCropper',
                  'image', 'optimisedImageUri',
                  function($scope, $rootScope, $stateParams, $state,
-                          $filter, mediaApi, mediaCropper,
+                          $filter, $document, mediaApi, mediaCropper,
                           image, optimisedImageUri) {
 
     const ctrl = this;
+    const body = $document[0].body;
 
     var imageId = $stateParams.imageId;
     ctrl.image = image;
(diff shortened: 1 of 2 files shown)


akash1810:
nice :+1:

theefer:
Neat :+1:

tsop14:
A return keypress on the crop button bubbles up to the body, so the eventListener calls crop. The keypress then triggers ng:click, so crop is called again. To prevent this an if statement checks if the image is in the process of cropping and only performs the crop if it is not.

kahuna/public/js/crop/controller.js
```diff
@@ -66,7 +67,16 @@ crop.controller('ImageCropCtrl',
         // else undefined is fine
     };
 
-    ctrl.crop = function() {
+     ctrl.callCrop = function() {
+         //prevents return keypress on the crop button posting crop twice
+         if (ctrl.cropping === false) {
+             crop();
+         } else {
+             return;
```
theefer: Note that this `else` is superfluous as an empty return is the default anyway.

kahuna/public/js/crop/controller.js
```diff
@@ -66,7 +67,16 @@ crop.controller('ImageCropCtrl',
         // else undefined is fine
     };
 
-    ctrl.crop = function() {
+     ctrl.callCrop = function() {
+         //prevents return keypress on the crop button posting crop twice
+         if (ctrl.cropping === false) {
```
theefer: `if (! ctrl.cropping) {`?

kahuna/public/js/crop/controller.js
```diff
@@ -66,7 +67,16 @@ crop.controller('ImageCropCtrl',
...
+             crop();
+         } else {
+             return;
+         }
+     };
+
+    crop = () => {
```
theefer: You need to define it as a `function` or else put a `const` before it, otherwise it's on the global scope. (Does our linting not catch that?)

theefer:
A few minor comments, else looks :+1:

kahuna/public/js/crop/controller.js
tsop14: > Does our linting not catch that?

Not caught; the linter was happy.

tsop14: When I define it the linter complains: 'crop' is defined but never used.

kahuna/public/js/crop/controller.js
```diff
@@ -66,7 +67,16 @@ crop.controller('ImageCropCtrl',
...
+     ctrl.callCrop = function() {
+         //prevents return keypress on the crop button posting crop twice
+         if (!ctrl.cropping) {
+             crop();
+         } else {
+             return;
+         }
+     };
+
+    const crop = () => {
```
akash1810: why `const` over `function`?

jamesgorrie: `function` would make more sense, as if I call `callCrop` before this line, it'll fail as `crop` wouldn't have been defined yet.

kahuna/public/js/crop/controller.js
```diff
@@ -66,7 +67,16 @@ crop.controller('ImageCropCtrl',
         // else undefined is fine
     };
 
-    ctrl.crop = function() {
+     ctrl.callCrop = function() {
+         //prevents return keypress on the crop button posting crop twice
+         if (!ctrl.cropping) {
+             crop();
+         } else {
```
jamesgorrie: As @theefer mentioned, you don't need the `else` here.

PR #1406, opened 2015-10-29
Tooltip for new image button.

akash1810:
Example:

screen shot 2015-10-29 at 13 05 34

### DIFF
--- a/kahuna/public/js/search/results.html
+++ b/kahuna/public/js/search/results.html
@@ -20,7 +20,8 @@
                 {{ctrl.totalResults | toLocaleString}} matches
                 <button ng:show="ctrl.newImagesCount > 0"
                         ng:click="ctrl.revealNewImages()"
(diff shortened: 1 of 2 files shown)


kahuna/public/js/search/results.html
```diff
@@ -20,7 +20,8 @@
                 {{ctrl.totalResults | toLocaleString}} matches
                 <button ng:show="ctrl.newImagesCount > 0"
                         ng:click="ctrl.revealNewImages()"
-                        class="image-results-count__new">
+                        class="image-results-count__new"
+                        title="{{ctrl.newImagesCount}} new images since {{ctrl.lastestTimeMoment}}">
```
jamesgorrie: `as of` instead of `since`?
Since would be `since 12 Sep 2013`?

jamesgorrie:
:+1: after grammar.

PR #1407, opened 2015-10-29
Tidy up of MediaApiController into SuggestionController

jamesgorrie:
[Turns out you can't do pagination / scrolling on aggregates](http://stackoverflow.com/questions/27776582/aggregation-sorting-pagination-in-elastic-search), [it's very complicated](https://github.com/elastic/elasticsearch/issues/4915#issuecomment-33963055). I should have looked first.

The separation of controllers seemed to make sense though as it was becoming hard to plough through the `MediaApi` controller.

### DIFF
--- a/media-api/app/controllers/SuggestionController.scala
+++ b/media-api/app/controllers/SuggestionController.scala
@@ -5,10 +5,7 @@ import play.api.libs.json.{Reads, Json, Writes}
 import play.api.mvc.{AnyContent, Request, Controller}
 
 import com.gu.mediaservice.lib.argo.ArgoHelpers
-import com.gu.mediaservice.lib.auth
-import com.gu.mediaservice.lib.auth.{PermissionStore, KeyStore}
 
-import lib.Config
 import lib.elasticsearch.{AggregateSearchResults, ElasticSearch}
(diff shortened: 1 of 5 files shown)


media-api/app/controllers/SuggestionController.scala
```diff
@@ -0,0 +1,108 @@
...
+
+import lib.Config
+import lib.elasticsearch.{AggregateSearchResults, ElasticSearch}
+import lib.querysyntax.Parser
+
+import scala.util.Try
+
+object SuggestionController extends Controller with ArgoHelpers {
+
+  val keyStore = new KeyStore(Config.keyStoreBucket, Config.awsCredentials)
```
theefer: Seems odd to have a second keystore. Also it's never started in the Global file. Can we reuse the same keystore (and permission store) for both?

jamesgorrie: Ah yes, I forgot about that. [I've moved it over to it's own place](https://github.com/guardian/grid/commit/1884a5366ca494ac0d937f14f2cfa240f3672eef).

theefer:
:+1:

PR #1409, opened 2015-10-30
Usage UI

kenoir:
Provides a UI to the usage app.


alastairjardine:
Can I have a wee looksee?

kenoir:
@alastairjardine of course. We're struggling with just getting it all end to end right now, so best to just come poke us.

akash1810:
Example usage tab:

screen shot 2015-11-02 at 11 39 02

kenoir:
:+1:

kahuna/public/js/components/gr-image-usage/gr-image-usage-list.html
```diff
@@ -0,0 +1,22 @@
+<div class="image-info__group">
+    <span class="flex-container gr-image-usage__field_title">
+        {{ctrl.type}} ({{ctrl.usages.length}})
+    </span>
+
+    <ul>
+        <li class="gr-image-usage__record" ng-repeat="data in ctrl.usages.data">
+            {{data.title}}
+            <div class="more-info">
+                <span class="date-added">{{ctrl.formatTimestamp(data.dateAdded)}}</span>
```
theefer: Could be done as a filter, but equally :+1:

kahuna/public/js/components/gr-image-usage/gr-image-usage.css
```diff
@@ -0,0 +1,35 @@
+.gr-image-usage__field_title {
```
theefer: Does this belong here?

kahuna/public/js/components/gr-image-usage/gr-image-usage.js
```diff
@@ -0,0 +1,64 @@
...
+}]);
+
+module.directive('grImageUsage', [function() {
+    return {
+        restrict: 'E',
+        template: template,
+        controller: 'grImageUsageCtrl',
+        controllerAs: 'ctrl',
+        bindToController: true,
+        transclude: true,
```
theefer: Is this necessary?

kahuna/public/js/components/gr-image-usage/gr-image-usage.css
```diff
@@ -0,0 +1,35 @@
...
+    vertical-align: top;
+    text-align: left;
+}
+
+gr-image-usage-list {
+    display: block;
+}
+
+gr-image-usage-list .date-added {
+    font-size: 11px;
```
jamesgorrie: We could do something [similar to this](https://github.com/guardian/grid/blob/master/kahuna/public/stylesheets/main.css#L325) as we have it in a few places, perhaps `font-tiny`? Not a blocker, maybe a thought for later.

kahuna/public/js/components/gr-image-usage/gr-image-usage.js
```diff
@@ -0,0 +1,64 @@
...
+import moment from 'moment';
+
+import template from './gr-image-usage.html!text';
+import usageTemplate from './gr-image-usage-list.html!text';
+import './gr-image-usage.css!';
+
+export let module = angular.module('gr.imageUsage', []);
+
+module.controller('grImageUsageCtrl', ['mediaUsage', function (mediaUsage) {
+    let ctrl = this;
```
jamesgorrie: `const` works here too.

usage/app/model/ContentDetails.scala
```diff
@@ -20,7 +24,15 @@ object ContentDetails {
...
-      content.sectionId.getOrElse("none")
+      content.sectionId.getOrElse("none"),
+      composerUrl(content)
     )
   }
+
+  def composerUrl(content: Content): Option[URL] = content.fields
+    .flatMap(_.internalComposerCode)
+    .flatMap(composerId => {
+      Try(new URL(s"${Config.composerBaseUrl}/${composerId}")).toOption
```
theefer: In the future it could be nice to denormalise this into the usage metadata at collection time, rather than having to construct this here, which will imply knowledge of all the systems this integrates with.

kahuna/public/js/components/gr-image-usage/gr-image-usage.css
akash1810: Yes. It looks a bit out of place because it hasn't been restricted to `gr-image-usage-list` like rules below. Updating.

kahuna/public/js/components/gr-image-usage/gr-image-usage-list.html
jamesgorrie: I'm still always wary of filters when you see how often they actually run...

theefer: This function on the controller would run the same number of times though...?

In both cases, if the expression is immutable, you can/should make it a one-time binding by prefixing with `::`, e.g. `{{:: ctrl.formatTimestamp(data.dateAdded)}}`, which prevents the creation of a watcher for it.

alastairjardine:
@akash1810 Could you do a screenshot for the >1 usage line? Would like to have a quick look at it so that we can ensure the spacing is good :)

That 'g' icon looks quite small. Could we punch up the size a bit? Might be best doing it on screen – call me over if you have a second.

usage/app/model/ContentDetails.scala
kenoir: I'm not sure whether it makes sense to denormalise this in ES but _not_ in the metadata dynamo store. The composer base url is not necessarily constant but the ID is and can be used to extract further metadata from other systems (Workflow for example).

alastairjardine:
Nice one @akash1810

akash1810:
> Nice one @akash1810

and @kenoir !! :smile:

PR #1411, opened 2015-10-30
Stupid ui router

jamesgorrie:
Slashes keep breaking - now they look right, but are double encoded when read from the `stateParams`.
[Following the following advice](https://github.com/angular-ui/ui-router/issues/1119#issuecomment-134603427), this seems to work in all cases.

Sorry @akash1810 - I think this is what you did the first time.

### DIFF
--- a/kahuna/public/js/search/index.js
+++ b/kahuna/public/js/search/index.js
@@ -38,8 +38,8 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
     // see: https://github.com/angular-ui/ui-router/issues/1119#issuecomment-64696060
     // TODO: Fix this
     $urlMatcherFactoryProvider.type('Slashed', {
-        encode: val => val || '',
-        decode: val => val || '',
+        encode: val => angular.isDefined(val) ? val : undefined,
+        decode: val => angular.isDefined(val) ? val : undefined,
         // We want to always match this type. If we match on slash, it won't update
         // the `stateParams`.
         is: () => true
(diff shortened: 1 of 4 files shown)


kahuna/public/js/search/index.js
```diff
@@ -24,14 +24,26 @@ export var search = angular.module('kahuna.search', [
     'data-structure.list-factory',
     'data-structure.ordered-set-factory',
     'gr.topBar',
-    'grPanel'
+    'grPanel',
+    'ui.router'
```
jamesgorrie: This was added as it should be here.

kahuna/public/js/search/index.js
```diff
@@ -24,14 +24,26 @@ export var search = angular.module('kahuna.search', [
...
-               function($stateProvider) {
+search.config(['$stateProvider', '$urlMatcherFactoryProvider',
+               function($stateProvider, $urlMatcherFactoryProvider) {
+
+    // Query params slashes are encoded that mess up our seach query when we search for e.g. `24/7`.
+    // see: https://github.com/angular-ui/ui-router/issues/1119#issuecomment-64696060
+    // TODO: Fix this
+    $urlMatcherFactoryProvider.type('Slashed', {
+        encode: val => val || '',
+        decode: val => val || '',
```
akash1810: Will this decode `%2F` to `%2F` rather than `/`?

tsop14:
:+1: :tophat:

tsop14:
Happy to test this in TEST with the cases I saw.

kahuna/public/js/search/index.js
jamesgorrie: If you search for %2F, you will get %2F.
e.g.
* create label: %2F
* search for #%2F

input value: #%2F
url value: %23%252F

jamesgorrie: Saying that, I don't want to return a string for undefineds, so I have updated it.

PR #1413, opened 2015-10-30
Print styling enhancements

tsop14:
Keywords are sometimes really long and labels might be - delimits height of all metadata descriptions to try to keep things on the same page. 

Firefox had different display to Chrome - styling changes to make image appear in both.

### DIFF
--- a/kahuna/public/stylesheets/main.css
+++ b/kahuna/public/stylesheets/main.css
@@ -2131,6 +2131,7 @@ FIXME: what to do with touch devices
         max-width: 100%;
         max-height: 450px;
         transform: none;
+        position: static;
     }
 
     .image-canvas, .image-details {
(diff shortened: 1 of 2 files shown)


kahuna/public/stylesheets/main.css
```diff
@@ -2131,14 +2131,15 @@ FIXME: what to do with touch devices
...
     }
 
-    .image-details {
+    .image-canvas, .image-details {
         height: auto;
         width: auto;
     }
 
-    .image-info__description {
+    dd {
```
jamesgorrie: Do these `dd`s have classes?

jamesgorrie:
Could we just hide the keywords - or are they useful?
Otherwise :+1:

kahuna/public/stylesheets/main.css
tsop14: The dd used for keywords does not :disappointed:

jamesgorrie: Could it? We've avoided this just incase we accidentally target a stray `dd`...

tsop14:
> Could we just hide the keywords - or are they useful?

Would rather keep it, just in cases.

PR #1414, opened 2015-10-30
WIP UI create full crop

tsop14:
full crop


alastairjardine:
How about ghosting a full crop, rather than having a button? Happy to sketch out what I mean...

PR #1416, opened 2015-11-02
Also bundle image profiles with image-loader

theefer:
Small change toward #1306  that will allow us to release https://github.com/guardian/grid-infra/pull/141/files.


jamesgorrie:
:+1:

PR #1419, opened 2015-11-02
move location of settings.ini to outside of the repo to avoid accidental committing.

akash1810:
move location of settings.ini to outside of the repo to avoid accidental committing.

### DIFF
--- a/scripts/dot-properties/main.py
+++ b/scripts/dot-properties/main.py
@@ -3,7 +3,8 @@
 from ConfigParser import ConfigParser
 from generate import generate_files
 
-CONFIG_FILE = 'settings/settings.ini'
+CONFIG_FILE = '/etc/gu/grid-settings.ini'
+OUTPUT_DIR = '/etc/gu'
 
 def _default_option(config, section, option, default=None):
     return config.get(section, option) \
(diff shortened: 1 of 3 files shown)


theefer:
:+1:

jamesgorrie:
Eventually we should remove it from the `.gitignore` too.

scripts/dot-properties/main.py
```diff
@@ -21,11 +22,9 @@ def main():
     if not cf_stack:
         raise Exception('No stack found in {}'.format(CONFIG_FILE))
 
-    output_dir = _default_option(config, 'output', 'directory', 'output')
-
     properties = config._sections['properties']
 
-    generate_files(aws_profile_name, aws_region, cf_stack, output_dir, properties)
+    generate_files(aws_profile_name, aws_region, cf_stack, OUTPUT_DIR, properties)
```
tsop14: Is there a convention for when to use lower case and when to use upper case?

akash1810:
> Eventually we should remove it from the `.gitignore` too.

Yep - after everyone's deleted the file first!

scripts/dot-properties/main.py
akash1810: The python convention is upper-case for constants and lower for variables.

PR #1423, opened 2015-11-03
Style related labels scrollbar in Chrome to avoid overlap

theefer:
Should look something like:
image

Sadly it's seemingly impossible to style it in Firefox without re-implementing the scrollbar in JS, which is silly.


jamesgorrie:
:+1:

PR #1424, opened 2015-11-04
Use the guAnalyser correctly

jamesgorrie:
At the moment we index only `title` and `description` with the `english_s_stemmer`.
When we search on `Words` we apply the analyser to all fields. 

If a field hasn't been indexed with this analyser, it cannot be searched correctly with it. When searching for `byline`s that have an `s` in them e.g. "Charles Emerson" it then gets broken into "Charle Emerson" when using the `english_s_stemmer` as it hasn't been indexed to be searched like this.

This uses [this trick](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-match-query.html#CO114-1) to use the right search analyser dependant on how the field has been indexed.

So:

__Search:__ Jeremy Corbyn
__Matches:__
* description: Jeremy Corbyn's
* description: Jeremy Corbyn
* byline: Jeremy Corbyn

__Does not match__
* byline: Jeremy Corbyn's

### DIFF
--- a/media-api/app/lib/elasticsearch/QueryBuilder.scala
+++ b/media-api/app/lib/elasticsearch/QueryBuilder.scala
@@ -24,6 +24,7 @@ class QueryBuilder(matchFields: Seq[String]) {
     new MultiMatchQueryBuilder(value, fields: _*).`type`(MultiMatchQueryBuilder.Type.PHRASE)
 
   def makeMultiQuery(value: Value, fields: Seq[String]) = value match {
+    // We only want to search the fields that are indexed with the `guAnalyser` with itself.
     case Words(string) => boolQuery().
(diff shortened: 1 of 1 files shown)


theefer:
Does this still search across fields as expected?

e.g.

Search: cricket jenkins

should find an image with

* description: Another stupid match of cricket
* byline: Tom Jenkins
(only mention of jenkins is in the byline, not description)

jamesgorrie:
@theefer I'm not sure that that works today ("cricket jenkins" does as he puts his name in the caption). I'll have a closer look.

jamesgorrie:
It now works, but we are now searching with a `bool.should` match with both analysers. This doesn't feel massively correct, so I might have a go at a suggestion from @kenoir.

jamesgorrie:
Closing in favour of #1426

PR #1426, opened 2015-11-04
Use correct stemmer

jamesgorrie:
In favour of #1424.

`minimal_english` is described as follows (bold for emphasis):

> The EnglishMinimalStemmer in Lucene, which __removes plurals__

This means "charles" becomes "charle" because it's stupid.

We've actually gone with the analyser [here](https://www.elastic.co/guide/en/elasticsearch/guide/current/algorithmic-stemmers.html). It seems to do everything we want it to.

A few re-namings too to hopefully make it more clear.

__NOTE:__
Reindex necessary.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Settings.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Settings.scala
@@ -3,27 +3,26 @@ package com.gu.mediaservice.lib.elasticsearch
 import play.api.libs.json.Json
 
 object IndexSettings {
-  // TODO rename `english_s_stemmer` as its an analyzer not a stemmer - would require a reindex.
-  val guAnalyzer = "english_s_stemmer"
+  val englishAnalyzerName = "english_analyzer"
 
-  val englishSStemmer = Json.obj(
+  val englishAnalyzer = Json.obj(
     "type" -> "custom",
     "tokenizer" -> "standard",
     "filter" -> Json.arr(
-      "lowercase",
-      "asciifolding",
       "english_possessive_stemmer",
-      "gu_stopwords",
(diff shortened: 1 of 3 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Settings.scala
```diff
@@ -3,27 +3,26 @@ package com.gu.mediaservice.lib.elasticsearch
...
     "type" -> "custom",
     "tokenizer" -> "standard",
     "filter" -> Json.arr(
-      "lowercase",
-      "asciifolding",
       "english_possessive_stemmer",
-      "gu_stopwords",
-      "s_stemmer"
+      "lowercase",
+      "english_stop",
```
jamesgorrie: Having `lowercase` and `english_stop` in this order is important, as they are filtered and then sent to the stemmer.

theefer: Can you put this as a comment in the code?

akash1810:
> A few re-namings too to hopefully make it more clear.

Yay! As we only have a single analyser, could we also remove the `english` prefix from variable names? i.e. `englishAnalyzerName` => `analyzerName` etc.?

jamesgorrie:
> `englishAnalyzerName` => `analyzerName`

@akash1810 we might want to add some later, as this is searching the description / title - it seems to be a decent name for it.

theefer:
:+1:  but let's be thorough testing that it works as expected for all combinations of stemming (plurals, possessive apostrophes), stop word elimination (in, the, etc), and across multiple fields with different analysers.

jamesgorrie:
@theefer closing until I can find some actual working solution, which none of these are at the moment.

PR #1433, opened 2015-11-05
Allow single quotes through query filter

jamesgorrie:
This does raise the question of what we do with double quotes. e.g. `by:Tony "Towers" Tetnowski`. We could match to see if there are double quotes, then use single quotes, but then what if we have both?

This is is a simple solution - I'd like to see if we have a double quote problem before we try fix it.


theefer:
:+1:

PR #1434, opened 2015-11-05
increases suggested labels returned from 20 to 30

tsop14:
Short term fix as travel need to see more labels. 

> At the moment we can see 20 but are using more than that on a weekly basis.


akash1810:
How did we decide on 30?

theefer:
More than 20? :-)

:+1:

akash1810:
very good.

:+1:

jamesgorrie:
Interestingly enough the performance doesn't seem to change, but I suppose it's only 10.

PR #1418, opened 2015-11-02
Print usage recording

kenoir:

PR #1436, opened 2015-11-05
Usages - Fix for re-used images

kenoir:
Adds `dateRemoved` to `MediaUsage` equality forcing creates rather than updates.

**Note:**
As the Usage table DynamoDB hash (or range) key does not incorporate the date created / removed this will overwrite old records with a new created date. 

To create new usages for the same image in the same document after being removed and re-added we'll need to rejig things a bit.

Depends on https://github.com/guardian/grid/pull/1418


akash1810:
nice :+1:

PR #1439, opened 2015-11-06
removes seen from here feature.

akash1810:
feature isn't used and doesn't make sense when ordering by oldest first.

before
screen shot 2015-11-06 at 13 18 51

after
screen shot 2015-11-06 at 13 18 09

### DIFF
--- a/kahuna/public/js/search/results.html
+++ b/kahuna/public/js/search/results.html
@@ -121,8 +121,7 @@
                 gu:lazy-table-cell="image">
 
                 <div class="result"
-                     ng:class="{ 'result--seen': ctrl.imageHasBeenSeen(image),
-                                 'result--selected': ctrl.imageHasBeenSelected(image)}">
+                     ng:class="{'result--selected': ctrl.imageHasBeenSelected(image)}">
 
                     <!-- <HACK> -->
                     <!-- We don't want to add a property on `image` to determine the state of the checkbox, so render two individual inputs. -->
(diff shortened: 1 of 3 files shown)


kenoir:
:+1: This always confused me.

paperboyo:
Hurray! (Does it remove it completely or only when sorted by oldest first?)

akash1810:
> Does it remove it completely or only when sorted by oldest first?

Completely.

paperboyo:
:fireworks:

jamesgorrie:
This feature doesn't actually make sense in many instances - I might put tracking on it to see if we can scrap it.

jamesgorrie:
Oh wait - I see you've done that. Could we remove the JS code too then?

akash1810:
> Could we remove the JS code too then?

[This](https://github.com/guardian/grid/pull/1439/files#diff-b8c6d4770b5b04294cc8c88d27ae65d3L122) has been removed. Is there more (pointers please)?

jamesgorrie:
@akash1810 :eyeglasses: Sorry - I don't know how I missed that.

theefer:
Out of interest, I don't see any tracking code being removed, so how do we know it's not being used?

akash1810:
> I don't see any tracking code being removed, so how do we know it's not being used?

There wasn't any tracking. We know its not used by conversations in passing with others when talking about other features.

:fearful:

theefer:
OK. Let's see if anyone notices. IIRC it was mostly for people on news and print desk who want to scan a whole day, but maybe this has been improved since with lazy scrolling. :+1:

jamesgorrie:
> We know its not used by conversations in passing
We have quite a lot of users now, so I think future removals should probably have some tracking. This feature didn't work properly however, so it'll be interesting to see if we hear anything.

theefer:
> We have quite a lot of users now, so I think future removals should probably have some tracking

+1

akash1810:
> future removals should probably have some tracking

Agreed. Oversight on my part for this PR.

theefer:
In fact, could we make it a rule to add tracking wherever we add new features we expected being used. An obvious example would be the usage tab, as well as the "full frame" button (cc @tsop14).

PR #1440, opened 2015-11-06
updates print style sheets

tsop14:
If we add components they need to be hidden on the print stylesheet. :cry: 

This is because wanted to avoid using universal style rule (display:none) but might be better to do this?


akash1810:
:+1:

akash1810:
Might be worth creating a generic `no-print` class, that way the [list](https://github.com/guardian/grid/blob/em-ff-printing/kahuna/public/stylesheets/main.css#L2130-L2136) doesn't get to unwieldy?

PR #1441, opened 2015-11-09
S3Watcher - Fix for not moving to fail bucket

kenoir:
Noticed that failed images were not getting moved to the fail bucket (although they were being logged as failed). 

This PR provides a (tested) fix this problem.

### DIFF
--- a/s3watcher/lib/Transfer.js
+++ b/s3watcher/lib/Transfer.js
@@ -37,6 +37,23 @@ module.exports = {
             console.log(stateMessage, state);
         }
 
+       const success = function(result) {
+            log("delete", s3Object);
+
+            return S3Helper.deleteS3Object(s3Object);
+        };
+
+        const failGraceful = function(e) {
+            log("failed", e)
+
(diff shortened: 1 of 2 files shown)


s3watcher/lib/Transfer.js
```diff
@@ -37,6 +37,23 @@ module.exports = {
...
+            log("delete", s3Object);
+
+            return S3Helper.deleteS3Object(s3Object);
+        };
+
+        const failGraceful = function(e) {
+            log("failed", e)
+
+            log("copy", s3CopyObject);
+            return S3Helper.copyS3Object(s3CopyObject).flatMap(function(){
```
akash1810: Does the S3 api not support `mv`?

akash1810:
:+1:

s3watcher/lib/Transfer.js
kenoir: Had to double check, but it seems not: http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html

PR #1446, opened 2015-11-09
[WIP] Usage: Use usage URI in notification

kenoir:
I'd like to consider notifying by URI rather than including all the
content in the message. This means that Thrall makes an API call to
usage when it gets the update and pulls the state itself from the URI
included.

This lets us avoid having to worry about deserializing the content of
messages differently from API responses.

I think there may be other benefits too but haven't completely thought
through all the consequences of this approach.

Comments welcome.


kenoir:
Closing in favour of https://github.com/guardian/grid/pull/1447

PR #1447, opened 2015-11-10
Store and present media usage from Elastic

kenoir:
This PR allows the grid to store usages in ES and present them via the Media API. This in turn allows us to sensibly present (and search) usage data across large numbers of images.

**Image page:**
usage-image

**Search view:**
usage-search

[API Examples](https://github.com/guardian/grid/blob/usage-event-stream-lambda/usage/api-examples.md)

**Deployment strategy:**

Apps should be deployed in the following manner to avoid service disruption:

Preperation:
- Turn off Usage by scaling ASG to 0
- Drop & Recreate Usage Table

Deployment:
- ES Mapping Update
- Thrall
- Image Loader
- Media API
- Kahuna
- Usage CF Update: https://github.com/guardian/grid-infra/pull/181
- Usage


deploy_js_lambda.sh
```diff
@@ -0,0 +1,32 @@
+#/bin/bash
+
+STAGE=$1
+BUCKET=$2
```
akash1810: Do we need this parameter? The bucket for the lambda is currently [hard coded](https://github.com/guardian/grid-infra/blob/master/platform/cloud-formation/media-service.json#L3684) in the template.

kahuna/public/config.js
```diff
@@ -1,14 +1,16 @@
 System.config({
-  baseURL: "/assets",
-  defaultJSExtensions: true,
-  transpiler: "traceur",
-  paths: {
+  "baseURL": "/assets",
```
akash1810: Why are we enclosing the keys in inverted commas? Was it breaking before?

kenoir: Auto generated by whatever happens when i run `npm install`.

On Mon, 16 Nov 2015 at 11:26 Akash A <notifications@github.com> wrote:

> In kahuna/public/config.js
> <https://github.com/guardian/grid/pull/1447#discussion_r44912655>:
>
> > @@ -1,14 +1,16 @@
> >  System.config({
> > -  baseURL: "/assets",
> > -  defaultJSExtensions: true,
> > -  transpiler: "traceur",
> > -  paths: {
> > +  "baseURL": "/assets",
>
> Why are we enclosing the keys in inverted commas? Was it breaking before?
>
> —
> <https://github.com/guardian/grid/pull/1447/files#r44912655>.
>

-- 

in England Number 908396

theefer:
cc @alastairjardine

kahuna/public/js/components/gr-image-usage/gr-image-usage.js
```diff
@@ -5,25 +5,31 @@ import template from './gr-image-usage.html!text';
 import usageTemplate from './gr-image-usage-list.html!text';
 import './gr-image-usage.css!';
 
-export let module = angular.module('gr.imageUsage', []);
+import '../../image/service';
 
-module.controller('grImageUsageCtrl', ['mediaUsage', function (mediaUsage) {
-    const ctrl = this;
+export let module = angular.module('gr.imageUsage', [
```
theefer: We really should almost never have to use `let`. `const` means constant reference, which almost all our code (including this) does, *not* immutable value.

kahuna/public/js/image/service.js
```diff
@@ -8,6 +10,7 @@ imageService.factory('imageService', [function() {
     function forImage(image) {
```
theefer: I really don't like this helper that does loads of extra work for nothing. Even if not now, we should change it to individual helpers for each transformation we're interested in.

kahuna/public/js/image/service.js
```diff
@@ -19,6 +22,30 @@ imageService.factory('imageService', [function() {
         };
     }
 
+    function usages(image) {
+        function usageTitle(usage) {
+            const referenceType = usage.get("usageType") == "print" ? "indesign" : "frontend";
```
theefer: I thought we were jslinting to use single quotes everywhere, but didn't look like it failed here?

thrall/app/lib/ElasticSearch.scala
```diff
@@ -168,6 +182,10 @@ object ElasticSearch extends ElasticSearchClient with ImageFields {
     """.stripMargin
 
   // Create the exports key or add to it
+  private val addUsagesScript =
```
theefer: Nitpicking, but could be named `replace` instead of `add` to differentiate from the exports one, which *does* add (append).

usage-updater/index.js
```diff
@@ -0,0 +1,21 @@
+const Rx = require('rx');
+const Notifications = require('./lib/Notifications');
+const Lambda        = require('./lib/Lambda');
+const UsageRequest  = require('./lib/UsageRequest');
+
+
+exports.handler = function(event, context) {
```
theefer: Do we want any monitoring/CloudWatch for this? I'm not particularly convinced we do, but worth making it a conscious decision.

usage-updater/lib/Lambda.js
```diff
@@ -0,0 +1,38 @@
+const Rx = require('rx');
+const Config = require('./Config');
```
theefer: Where does `Config.js` come from?

theefer:
@kenoir I've got some notes to discuss when we're both back from lunch!

usage-updater/lib/Lambda.js
kenoir: `Config.js` is where the config is for the Lambda. I think I originally intended it to be generated (or pulled from S3) by the deploy script, will go back and update.

theefer:
Notes as discussed:

* In media-api responses, I suggest the `usages` are an embedded collection pointing to the original resources in the usage service. Additionally, both in the media-api and usage service, each usage object could be its own resource so it's uniquely identifiable. In media-api it'd look something like:

``` json
{
  "id": "foo",
  "...": "...",
  "usages": {
    "uri": "/usages/media/8322fc48ebdcb610757d7d8dde076bfed860dfb7",
    "data": [
      {
        "uri": "/usages/media/8322fc48ebdcb610757d7d8dde076bfed860dfb7/usageid",
        "data": { usage info... }
      }
    ]
}
```

* Could `referenceType` and `usageType` be just `type`?
* Why is the `usageType` capitalised (`"Image"`)?
* Is `web` the best `usageType`? Are we differentiating between content/fronts/membership/etc at that level, or is it a catch-all with a further clue as to the context of publication?
* How can we expose more useful structured information onto usages: date/page/section/etc for print, composer ID for content, fronts ID for fronts, etc? Could be useful in ES for querying too?
* Should the example include a Composer link and/or ID?
* The ES example is not up-to-date, still uses `source`, `usageType`, etc.
* Should we use ISO date times with milliseconds from the start (check with @jamesgorrie)?
* Return taken down content with a status of `takendown`?

Also to think about:
* How to release incompatible kahuna/usage changes with minimal impact?

jamesgorrie:
> Is `web` the best `usageType`...

Could we have `"platform": "capi || print"` (is `web` misconstruing as it's apps to?)
and `"location": { not sure, but page for print, section etc could be applicable }`.

> Should we use ISO date times with milliseconds from the start

I think so - I can start the PR for that now.

theefer:
> "platform": "capi || print"

That goes back to the question of whether "capi" (or "web") include fronts (fronts are stored in a different part of CAPI)?

> is `web` misconstruing as it's apps to?

That was part of my reason for questioning it, so apparently it's not just me ;-)

deploy_js_lambda.sh
kenoir: If we hard code the bucket name we expose it in the repo, i'm not comfortable making the name public (although the permissions are sufficient to prevent tampering). It also is another Guardian specific thing that consumers of the project would have to work around.

.gitignore
```diff
@@ -1,4 +1,8 @@
-s3watcher/node_modules
+S3Watcher/node_modules
```
akash1810: Any reason to have these ignore rules on the project root rather than, for example, in the S3Watcher directory?

deploy_js_lambda.sh
theefer: Could it be looked up by tags using the aws CLI? If not manual sounds like the best compromise.

.gitignore
theefer: I suggest just having `node_modules` which is ignored recursively unless prefixed with a `/`.

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -108,6 +108,24 @@ object Mappings {
...
+  val usageReference =
+    nonDynamicObj(
+      "referenceType" -> nonAnalyzedString,
+      "uri"  -> nonAnalyzedString,
+      "name" -> sStemmerAnalysedString
+    )
+
+  val usagesMapping =
+    nonDynamicObj(
+      "title"        -> sStemmerAnalysedString,
```
akash1810: Why analyse this field? I'd imagine we'd only ever search for exact article titles (phrase matching)?

jamesgorrie: I would imagine you'd want some analysis on it - not `non_analysed` so you could match on a partial title?

deploy_js_lambda.sh
akash1810: This works too - https://github.com/guardian/grid/blob/aa-dev-usage-lambda/cloud-formation/scripts/get-or-create-artifact-bucket.sh (part of https://github.com/guardian/grid/pull/1470).

usage-updater/index.js
kenoir: I think we safely use the "free" Lambda invocation/error statistics to monitor this service if necessary. Usage updates are already recorded via the `usage` app itself.

usage-updater/lib/Lambda.js
kenoir: Now done here: https://github.com/guardian/grid/commit/86f18f29adb953bebf44856b899142e8fc58ae1a

kahuna/public/js/image/service.js
kenoir: I've extracted image usages here: https://github.com/guardian/grid/commit/813123cc06aab8c18c2973d5e5bc8ea7c2c40d2a

usage-updater/deploy.sh
```diff
@@ -0,0 +1,22 @@
+#/bin/bash
```
akash1810: missing a `!`?

kenoir:
@theefer re-re-re-review?

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -123,6 +123,57 @@ object Mappings {
       "filename" -> nonAnalyzedString
     )
 
+  val usageReference =
+    nonDynamicObj(
+      "type" -> nonAnalyzedString,
+      "uri"  -> nonAnalyzedString,
+      "name" -> sStemmerAnalysedString
```
theefer: Is `name` still used? Couldn't see in the API examples.

kahuna/public/js/image/service.js
```diff
@@ -1,4 +1,5 @@
 import angular from 'angular';
+import Immutable from 'immutable';
```
theefer: Not used?

kahuna/public/js/services/image/usages.js
```diff
@@ -0,0 +1,60 @@
...
+                usage.get('platform') == 'print' ? 'indesign' : 'frontend';
+
+            const build = (usage, referenceType) => {
+                const reference = usage.get('references').find(u =>
+                    u.get('type') == referenceType);
+
+                return (reference && reference.get('name')) ? reference.get('name') : 'No title found.';
+            };
+
+            return build(usage, referenceType);
```
theefer: Is the extra `build` function needed/helping if you're calling it immediately anyway?

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
kenoir: `name` is used in `kahuna` it's a "friendly" way of describing a reference.

kahuna/public/js/services/image/usages.js
kenoir: I think I might have done this for clarity (I sometimes like to use functions in JS for readability rather than reuse) but i've tried it the other way and it's about the same.

kahuna/public/js/services/image/usages.js
```diff
@@ -0,0 +1,60 @@
...
+                    hasDigitalUsages: !filterByPlatform('digital').isEmpty()
+                };
+
+            });
+
+        return usages$
+
+    }
+
+    return image => forImage(image);
```
theefer: I'm personally not too keen on this pattern we've used elsewhere (the image service) of having a big helper that computes a set of miscellaneous useful values and returns them all. It's less granular which makes it harder to track which is used where (e.g. to know what to test or adapt when changing the service), and it's less performant (everything is computed every time even if only a subset if used).

The alternatives I can think of that we have used so far (around the list and ordered set collection/streams used for results etc) are to [return a set of streams](https://github.com/guardian/grid/blob/usage-event-stream-lambda/kahuna/public/js/lib/data-structure/ordered-set-factory.js#L46-L48), rather than the stream of sets as above (streams are lazy so only the ones used will be computed), or to define [helper functions](https://github.com/guardian/grid/blob/usage-event-stream-lambda/kahuna/public/js/services/image-list.js) that can be composed to transform the plain original stream. These two approaches can be combined too.

I also find it a little confusing that a "service" ends up being a function, rather than a module that holds more descriptive methods of what it does (`service(foo)` vs `service.getUsages(foo)`). Elsewhere we used "factory" instead to denote "a function that returns a new instance".

Now this is just my opinion so if you disagree or have better suggestions, please share!

And I don't think it's worth spending a lot of time refactoring code that works, so if we want to make some adjustments, it may be fair to do whatever is quickest/easiest so we can ship usage!

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
theefer: OK :+1:. Is it worth adding to the API examples for completeness?

Or it could be done in the API (between ES and returning the JSON) so all clients get a nice representation but we don't denormalise something which is derived and harder to change later? Up to you.

thrall/app/lib/ElasticSearch.scala
```diff
@@ -168,6 +181,10 @@ object ElasticSearch extends ElasticSearchClient with ImageFields {
     """.stripMargin
 
   // Create the exports key or add to it
+  private val replaceUsagesScript =
+    "ctx._source.usages = usages;"
```
theefer: Not an immediate concern but out of interest, do you have any thoughts on how/if we could use the most recent timestamp in the usages to make sure last-write-wins always even if SQS messages arrive out of order (i.e. discard older messages)? Feels like it would be doable by finding the latest lastModified of both the source and new usages and checking the new one is more recent?

Just thinking ahead, let's not get distracted with that right now.

usage/app/controllers/UsageApi.scala
```diff
@@ -1,53 +1,121 @@
...
     val usagesFuture = UsageTable.queryByImageId(mediaId)
 
-    usagesFuture.map[play.api.mvc.Result](UsageResponseCollection.build).recover { case error: Exception => {
+    usagesFuture.map[play.api.mvc.Result]((mediaUsages: Set[MediaUsage]) => {
+      val usages = mediaUsages.toList.map(UsageBuilder.build)
+
+      usages match {
+        case Nil => respondNotFound("No usages found.")
+        case usage :: _ => {
+          val uri = Try { new URI(s"${services.usageBaseUri}/usages/media/${mediaId}") }.toOption
```
theefer: Curious why the Try here?

usage/app/lib/Config.scala
```diff
@@ -27,6 +29,7 @@ object Config extends CommonPlayAppProperties with CommonPlayAppConfig {
   val capiPreviewUser = properties("capi.preview.user")
   val capiPreviewPassword = properties("capi.preview.password")
 
+  val topicArn = properties("sns.topic.arn")
```
theefer: Does it need adding to the `dot-properties` dev script too? Presumably already added to CF?

usage/app/model/UsageRecord.scala
```diff
@@ -17,9 +19,10 @@ case class UsageRecord(
   mediaType: Option[String] = None,
   lastModified: Option[DateTime] = None,
   usageStatus: Option[String] = None,
-  dataMap: Option[Map[String, String]] = None,
+  printUsageMetadata: Option[PrintUsageMetadata] = None,
+  digitalUsageMetadata: Option[DigitalUsageMetadata] = None,
   dateAdded: Option[DateTime] = None,
-  dateRemoved: Option[DateTime] = None
+  dateRemoved: Either[String, Option[DateTime]] = Right(None)
```
theefer: Is it worth adding a comment to explain the semantics of the `Either` here?

theefer:
Some minor comments but looks great overall! :+1:

kahuna/public/js/services/image/usages.js
kenoir: @theefer like this? https://gist.github.com/kenoir/63beadb9f00310f261a2

thrall/app/lib/ElasticSearch.scala
kenoir: Alternative is to use a non-disordering message queue e.g. `kinesis`.

usage/app/lib/Config.scala
kenoir: Not in the CF I think. Bit of an oversight, and needs adding to the script.

usage/app/controllers/UsageApi.scala
kenoir: `uri` is optional, and `java.net.URI` will throw if it can't parse the string to a URI. So although I don't expect this to throw, I prefer it to `Some(thing)`.

jamesgorrie: `URI.create` :dancer:

kahuna/public/js/services/image/usages.js
theefer: Yeah I personally think that looks great. What do you think?

thrall/app/lib/ElasticSearch.scala
theefer: Yes good point. Let's keep that in mind for the future.

kahuna/public/js/services/image/usages.js
jamesgorrie: :+1: from me.

kenoir: @theefer I concur.

usage/app/lib/Config.scala
kenoir: CF: https://github.com/guardian/grid-infra/pull/181

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
kenoir: Yeah, if we did it in the API it would also allow us to chuck a bit of wrangling we do in the client to fill in empty values.

kenoir: Back and forth over the code a few times and I don't really like any of the refactoring options very much so i'm going to leave it.

kenoir:
@theefer Addressed the problems raised. Are we good to go tomorrow (in between sessions at Scala Exchange)?

theefer:
Yes I don't see why not, as long as you're able to support any issues :-)

Great work! :+1:

s3watcher/deploy.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
akash1810: This file has been replaced with [`build.sh`](https://github.com/guardian/grid/blob/master/s3watcher/build.sh), might need to rebase master?

kenoir:
Recent changes reduce the number of messages sent:

screenshot from 2015-12-17 09 39 24

kenoir:
@theefer message frequency has been significantly curtailed by:
- `groupBy` mediaId
- `flatMap` resulting streams of `UsageNotice` using `distinctUntilChanged`
- Override `UsageNotice` `equals` and `hashCode` to ignore `lastModifed` in notice JSON
- The existing `ResetException` workaround prevents memory grouped streams consuming all memory by discarding streams after given period (now 30 seconds).

kenoir:
@theefer if you're happy with the current changes, i'll do some more testing tomorrow and aim for a release Monday.

PR #1449, opened 2015-11-11
adds tracking to create full frame button

tsop14:
> In fact, could we make it a rule to add tracking wherever we add new features we expected being used. An obvious example would be the usage tab, as well as the "full frame" button (cc @tsop14).

Done.


theefer:
:+1:

PR #1451, opened 2015-11-12
add filename to results side panel

jamesgorrie:
The filename is busy being added to the caption so they can do this at the moment (see the filename and results at the same time).

Seems to make sense to just put it there.

### DIFF
--- a/kahuna/public/js/services/image-accessor.js
+++ b/kahuna/public/js/services/image-accessor.js
@@ -29,6 +29,12 @@ imageAccessor.factory('imageAccessor', function() {
         return image.data.metadata;
     }
 
+    function readExtraInfo(image) {
+        return {
+            filename: image.data.uploadInfo && image.data.uploadInfo.filename
+        };
+    }
+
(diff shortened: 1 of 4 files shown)


kahuna/public/js/services/image-accessor.js
```diff
@@ -29,6 +29,12 @@ imageAccessor.factory('imageAccessor', function() {
         return image.data.metadata;
     }
 
+    function readExtraInfo(image) {
+        return {
+            filename: image.data.uploadInfo && image.data.uploadInfo.filename
```
jamesgorrie: I've gone with `extraInfo` as we might want something else in there at some stage.

theefer:
:+1:

tsop14:
Given how similar they are, would it make sense to have all the info on the image page in this search side panel?

theefer:
> Given how similar they are, would it make sense to have all the info on the image page in this search side panel?

+1

jamesgorrie:
@tsop14 :+1:

PR #1452, opened 2015-11-12
Bump panda to 0.2.9

theefer:
Version bump in preparation for rolling out fixes merged into panda's master. I wanted to decouple these changes from upgrading to the current latest version.

Note that this keeps our existing panda infrastructure but we should eventually migrate to use:
- [ ] STS AssumeRole instead of panda credentials stored on disk.
- [ ] All services apart from kahuna/auth to use just the verification library and not the auth issuing library.
- [ ] Separate out auth from kahuna, making kahuna just static files and auth a standalone service much like [login.gutools](https://github.com/guardian/login.gutools).


tsop14:
:panda_face: :+1:

PR #1453, opened 2015-11-13
Datastore bucket

jamesgorrie:
Adding datastore bucket - initially to store collection data. Thought it worth doing as we're adding Put permissions.

Waiting on corresponding PR.


media-api/app/lib/Config.scala
```diff
@@ -34,6 +34,7 @@ object Config extends CommonPlayAppConfig with CommonPlayAppProperties {
 
   val imageBucket: String = properties("s3.image.bucket")
   val thumbBucket: String = properties("s3.thumb.bucket")
+  val dataStoreBucket: String = properties("s3.datastore.bucket")
```
theefer: Do we really want the media-api to be responsible for this?

jamesgorrie: No, but I want to get something to test infront of users, while we're doing that I can start setting up another service.

theefer: OK. I think I'd have favoured co-opting a different less crucial service (maybe metadata-editor), in case it has any problematic side-effects.

jamesgorrie:
Renamed.

scripts/dot-properties/templates/media-api.properties.template
```diff
@@ -3,6 +3,7 @@ aws.id={{AwsId}}
 aws.secret={{AwsSecret}}
 s3.image.bucket={{ImageBucket}}
 s3.thumb.bucket={{ThumbBucket}}
+s3.datastore.bucket={{CollectionsBucket}}
```
theefer: This one needs renaming too to match up the value read in `Config`

theefer:
One missing rename, and let me know what you think of [abusing a different service](https://github.com/guardian/grid/pull/1453#discussion_r44784254) :-)

Else :+1:

scripts/dot-properties/templates/media-api.properties.template
jamesgorrie: Ah yes, missing in the find replace.
thanks.

media-api/app/lib/Config.scala
jamesgorrie: Good call.

PR #1454, opened 2015-11-13
Bump panda to 0.2.10

theefer:
This should hopefully silence our 500s in kahuna.

This doesn't intrinsically solve the "anti-forgery token did not match" error, but it exposes it as a 400 to the client, which should then show it as a global error to the user, rather than failing silently.


jamesgorrie:
Makes sense :+1:

PR #1455, opened 2015-11-13
add collections and actiondata models

jamesgorrie:
Been working with @tsop14 around creating a collection model.

From that we decided we'd need to track who created it (or added it to an image) - so there's a small attempt to try and standardise action data.

I have looked for a standard around this - [Schema.org has one](schema.org/Action) - but it's too verbose (it uses `startTime` and `endTime`).

This is in common lib as it'll be needed for `Image` and `Edits` soon enough.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/Collection.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/Collection.scala
@@ -1,0 +1,19 @@
+package com.gu.mediaservice.model
+
+import org.joda.time.DateTime
+import play.api.libs.json._
+
+case class Collection(path: List[String], actionData: ActionData)
+object Collection {
+  implicit def formats: Format[Collection] = Json.format[Collection]
+}
+
+// Following the crop structure
+// TODO: Use this in crop too
+case class ActionData(author: String, date: DateTime)
+object ActionData {
+  implicit def formats: Format[ActionData] = Json.format[ActionData]
+  // TODO: Use the generic formats for DateTime
+  implicit val dateWrites = Writes.jodaDateWrites("yyyy-MM-dd'T'HH:mm:ss.SSSZZ")
+  implicit val dateReads = Reads.jodaDateReads("yyyy-MM-dd'T'HH:mm:ss.SSSZZ")
+}


common-lib/src/main/scala/com/gu/mediaservice/model/Collection.scala
```diff
@@ -0,0 +1,18 @@
...
+  implicit def formats: Format[Collection] = Json.format[Collection]
+}
+
+// Following the crop structure
+// TODO: Use this in crop too
+case class actionData(author: String, date: DateTime)
+object actionData {
+  implicit def formats: Format[actionData] = Json.format[actionData]
+  implicit val dateWrites = Writes.jodaDateWrites("yyyy-MM-dd'T'HH:mm:ss.SSSZ")
+  implicit val dateReads = Reads.jodaDateReads("yyyy-MM-dd'T'HH:mm:ss.SSSZ")
```
theefer: I swear we have some shared helpers for that somewhere so you don't have to duplicate here.

theefer:
:+1: minor comment

common-lib/src/main/scala/com/gu/mediaservice/model/Collection.scala
jamesgorrie: Is this the date format we want to go for in the long run though?

jamesgorrie: [This is also the only reference I can find to any date stuff](https://github.com/guardian/grid/blob/collection-model/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala#L45)

theefer: See https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/formatting/package.scala

theefer: Although the difference is that currently we never store milliseconds, which is actually quite annoying. I'd argue maybe create a new standard (without the noMillis) in the common-lib file and start using that instead, and we could migrate the rest in the future.

common-lib/src/main/scala/com/gu/mediaservice/model/Collection.scala
```diff
@@ -0,0 +1,18 @@
...
+import play.api.libs.json._
+
+case class Collection(path: List[String], actionData: actionData)
+object Collection {
+  implicit def formats: Format[Collection] = Json.format[Collection]
+}
+
+// Following the crop structure
+// TODO: Use this in crop too
+case class actionData(author: String, date: DateTime)
```
theefer: Classes (including case ~) should start with a capital letter.

tsop14:
:+1:

common-lib/src/main/scala/com/gu/mediaservice/model/Collection.scala
jamesgorrie: I've updated this so that we always [write with millis](https://github.com/guardian/grid/commit/d230bb8e8d0a4be8a98ec1090bfa69d7f4c3b56a#diff-7e88df982621cdb0d9fcefd26e742452R10) even if we didn't have them.

When we read - we use [two parsers](https://github.com/guardian/grid/blob/d230bb8e8d0a4be8a98ec1090bfa69d7f4c3b56a/common-lib/src/main/scala/com/gu/mediaservice/lib/formatting/package.scala#L14-L15) so we should be good on both fronts.

I'd like to do a little more testing though to see if this screws up anything on the client (I can't image it would).

jamesgorrie: Saying that - I can (and will) write some automation around the parsing.

theefer: Cool, probably best as a separate PR?

Worth doing a full end-to-end as this data can show up in different places (ES index, exports S3 object metadata, etc).

jamesgorrie: Yea - I'll separate it out and then apply it here - I'm going to go with this for now though.

PR #1456, opened 2015-11-16
Adds some utils to run apps in panes with tmux, and to refresh with ping

kenoir:
- `run_tmux.sh` runs all your services in tmux panes
- `the_pingler.sh` pings all the services (to keep them fresh after code updates in dev)

### DIFF
--- a/the_pingler.sh
+++ b/the_pingler.sh
@@ -1,0 +1,33 @@
+#!/bin/zsh
+
+set -o shwordsplit
+
+API="http://localhost:9001/management/healthcheck"
+THRALL="http://localhost:9002/"
+IMAGE_LOADER="http://localhost:9003/management/healthcheck"
+FTP_WATCHER="http://localhost:9004/"
+KAHUNA="http://localhost:9005/management/healthcheck"
+CROPPER="http://localhost:9006/management/healthcheck"
+METADATA="http://localhost:9007/management/healthcheck"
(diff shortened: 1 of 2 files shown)


the_pingler.sh
```diff
@@ -0,0 +1,33 @@
...
+set -o shwordsplit
+
+API="http://localhost:9001/management/healthcheck"
+THRALL="http://localhost:9002/"
+IMAGE_LOADER="http://localhost:9003/management/healthcheck"
+FTP_WATCHER="http://localhost:9004/"
+KAHUNA="http://localhost:9005/management/healthcheck"
+CROPPER="http://localhost:9006/management/healthcheck"
+METADATA="http://localhost:9007/management/healthcheck"
+USAGE="https://localhost:9009/"
```
theefer: why no healthcheck for this one?

theefer:
:+1:

the_pingler.sh
kenoir: It's in a different place for `ftp-watcher`: `/health-check`. `thrall` reports failed health-check sometime because of the message frequency check.

theefer: I meant `usage`.

kenoir: `usage` health check is also a bit dodgy in dev (although it presents the *correct* management endpoints) as if the polling fails it can break the http interface.

PR #1457, opened 2015-11-16
add the id to the collection model

jamesgorrie:
Adding the ID so that we can have some sort of reference to the path for searching / getting them later.
As discussed with @tsop14 we're probably going to be storing the `path`, `pathId`, and potentially a hierarchically tokened version of the `pathId` so that we can experiment with what might be the most useful search to users.

`POST /collections { "data": ["g2", "art", "24/7"] }`
`GET /g2/art/24%2F7`

```JSON
{
  "uri": "media-metadata.com/collections/g2/art/24%2F7",
  "data" : {
    "pathId": "g2/art/24%2F7",
    "path": ["g2", "art", "24/7"],
    "actionData": {
      "author": "j@g.com",
      "date": "2015...",
    },
    "actions": [{
      "name": "delete",
      "method": "DELETE",
      "href": "media-metadata.com/collections/g2/art/24%2F7"
    }]
  }
}
```


theefer:
Not sure I understand what the `pathId` is used for that cannot be achieved with `path`?

jamesgorrie:
@theefer what would it look like as an identifier in the URL (take the `delete` action for example)?

I suppose we could do a `DELETE` to `/collections` with `{ "data": ["g2", "arts", "24/7"] }` - but that seems incredibly not `REST`y.

There is also the issue of not being able to match lists exactly in ES - the general advice is create another field that is hash or compilation of sorts of the list and store it in another field to exact match on.

theefer:
`GET /g2/art/24//7` (or probably better, `GET /g2/art/24%2F7`) and `DELETE` etc makes sense, but why can't we split the path in the server to match the `["g2", "arts", "24/7"]` path in ES? Does ES allow exact matching arrays (with order)?

jamesgorrie:
> Does ES allow exact matching arrays (with order)?

No, as far as @tsop14 and I can see - the recommended advice is to create a new searchable field. You can create an `AND` search but the order is not adhered to.

I think having the ability to exact search (an image will only exist in collections you put it in i.e. it won't inherently be in `g2` if it's in `g2/art` (if this was the case it's contrary to all other folder like structures)) - this would need us to have a way of representing an exact value in ES, URL encoding doesn't feel great though.

I think having this stored with the [opposite to this](https://www.elastic.co/guide/en/elasticsearch/reference/1.4/analysis-pathhierarchy-tokenizer.html) could be interesting to test with with very little complexity added.

With both of the above we can test as to what is best, and potentially use them in different circumstances (stats vs search for example).

jamesgorrie:
Saying that, we could look into writing a custom array analyser, although I'm unsure what the overhead would be.

jamesgorrie:
Or hash the array, making the fact that we have a constant hash the security of idempotentcy and the search for that, but I feel we lose out with the potential to test with the hierarchy then.

jamesgorrie:
@theefer I've added the ES mappings to give a more holistic view of what's going on.

So if we have:
``` JSON
{
    "image": {
        "collections": [
            {
                "path": "g2/film/art",
                "actionData": {
                    "date": "2015-11-19T00:00:00.000Z",
                    "author": "james.gorrie@guardian.co.uk"
                },
                "pathId": "g2/film/art",
                "pathList": [
                    "g2",
                    "film",
                    "art"
                ]
            }
        ]
    }
}
```

If we query with `g2/art` it will on collections.:
* path: true
* pathId: false
* pathList: false

If we query with `g2` it will on collections.:
* path: true
* pathId: false
* pathList: true

If we query with `g2/film/art` it will on collections.:
* path: true
* pathId: true
* pathList: false

I was going to add the `"index_name": "collection"` to the `path` but thought it might be best to actually just use the aliasing in parboiled so we can change it to the parameter we want to query making it more configurable.

theefer:
Looks good!

How are `/` escaped in paths if present in collection names?

Makes sense to translate in the query syntax parser imo.

jamesgorrie:
> How are / escaped in paths if present in collection names?

`List("news", "live", "24/7").pathId =  'news/live/24%2F7'.

> Makes sense to translate in the query syntax parser imo.

Are you talking about the search name i.e. collection?

theefer:
Cool, and yes :+1:

PR #1460, opened 2015-11-17
use correct list

akash1810:


theefer:
Meh :+1: 

Thanks for fixing!

theefer:
how long as it broken for? could we reindex some stuff (post hackday) to cleanup?

akash1810:
> how long as it broken for?

git history suggests August...!

PR #1461, opened 2015-11-18
[WIP] colour themes

tsop14:
Adds a new stylesheet which is disabled until a button is clicked by the user. 
Clicking the buttons enables/disables the stylesheet which swaps the colour theme. 

background

screen shot 2015-11-18 at 17 24 04

TODO
Store the result in local storage?

### DIFF
--- a/kahuna/app/views/main.scala.html
+++ b/kahuna/app/views/main.scala.html
@@ -19,6 +19,7 @@
     <link rel="sentry-dsn" href="@dsn" />
 }
     <link rel="stylesheet" href="@routes.Assets.at("stylesheets/main.css")" />
+    <link rel="alternate stylesheet" title="white" ng-disabled="true" href="@routes.Assets.at("stylesheets/white-background.css")" />
 
     <style>
         .ng-cloak { display: none }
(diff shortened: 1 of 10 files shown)


theefer:
Fun, but what's the rationale for that?

I don't mind user preferences, but I would be worried about the maintenance overhead vs actual benefit. Do you have any thoughts on how we could mitigate that?

paperboyo:
Just my nosy 2gr... I would rather see that switch in three-dot-menu not to crowd the toolbar. Maybe it would be better to have all the currently dark background to be white (or very light grey)?

I don't particularly need that, but user prefs in localstorage would work wonders to store panel lock status (it drives me crazy to have to click Show info panel every time I make a selection). That's the real reason for my comment, really :blush:

alastairjardine:
Looks a bit...broken! Without full changing all the UI chrome it's going to feel kinda odd. Not quite sure of the benefit?

All these types of prefs usually live in the user dropdown on the top right hand side, under the name.

tsop14:
> Fun, but what's the rationale for that?

Colours look different on black/white background and pictures are often placed on a white-ish background on the website. So, thought would be nice if they could swap background to have a looksee from within the Grid. 

> Looks a bit...broken! Without full changing all the UI chrome it's going to feel kinda odd. Not quite sure of the benefit?

:point_up_2: not sure this outweighs appearing 'broken'.

PR #1464, opened 2015-11-19
small script to download permissions file from s3

akash1810:
Helper script to download the `permissions.properties` file from any stack. Downloads to `/tmp` as its outside of the project directory so can't be checked in.

Usage: `./get-permissions.sh STACK_NAME`.

Not entirely convinced `/scripts` is the correct directory...

### DIFF
--- a/scripts/get-permissions.sh
+++ b/scripts/get-permissions.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+STACK_NAME=$1
+
+if [ -z ${STACK_NAME} ];
+then
+    echo "Please specify a stack name."
+    echo "Pro tip: you can get your stack name by running:"
+    echo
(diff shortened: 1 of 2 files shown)


tsop14:
:+1:

theefer:
Cool :+1: 

Note that the in an ideal world, we should use the permissions service so we can also grant others (e.g. Jo) to self-service permission management.

akash1810:
> Note that the in an ideal world, we should use the permissions service

Was thinking about this earlier. Given that this service isn't open (yet), what would be the best approach to integrate it whilst allowing the Grid to be used by the community?

theefer:
> Given that this service isn't open (yet), what would be the best approach to integrate it whilst allowing the Grid to be used by the community?

Excellent question. We should not tie the Grid to any closed project.

cc @lindseydew

PR #1466, opened 2015-11-19
The terminal is your friend for cloudformation updates

akash1810:
Why spend time logging into the aws web console all the time? I got tired doing that...

(`create-dev-stack.sh` added for community...)


theefer:
:+1:

PR #1474, opened 2015-11-23
update material icon font.

akash1810:
Additional icons are available - https://www.google.com/design/icons/.

Fonts taken from https://github.com/google/material-design-icons/tree/master/iconfont. Still waiting on https://github.com/google/material-design-icons/issues/133 to get the font resources via jspm.


jamesgorrie:
Are there any that we need - or are we keeping up to date?
:+1:

akash1810:
Just keeping up to date.

PR #1478, opened 2015-11-23
adds a progress bar for http requests.

akash1810:
Gives visibility on the progression of an image(s) download - there's a 2px progress bar at the top of the page.

I want to add toasts too, in a subsequent PR, using [`ng-messages`](https://docs.angularjs.org/api/ngMessages/directive/ngMessages). There'd need to be a bit of refactoring to the way we currently use `ng-messages` to display global errors however, so this felt like a very quick win.

There's an added bonus in that the progress bar is shown for any http request, e.g creating crops, metadata edits, search, upload (can be configured to ignore certain requests if needed).

progress-bar

### DIFF
--- a/kahuna/package.json
+++ b/kahuna/package.json
@@ -12,6 +12,7 @@
       "angular-ui-router": "github:angular-ui/ui-router@^0.2.15",
       "angular-ui-router-extras": "github:christopherthielen/ui-router-extras@^0.0.14",
       "angular-xeditable": "github:vitalets/angular-xeditable@^0.1.9",
+      "angular-loading-bar": "github:chieffancypants/angular-loading-bar@^0.8.0",
       "clean-css": "npm:clean-css@^3.3.5",
       "css": "github:systemjs/plugin-css@0.1.13",
       "github:tapmodo/Jcrop": "github:tapmodo/Jcrop@0.9.12",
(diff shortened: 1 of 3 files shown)


akash1810:
On TEST now.

cc @alastairjardine

jamesgorrie:
Besides the download - what other requests justify a loader that don't have a contextual one?

akash1810:
> Besides the download - what other requests justify a loader that don't have a contextual one?

Quite a few: removing label, image/crop deletion.

There are other places where the current progress indicator is adding "ing", for example "cropping". A progress bar looks a bit nicer in this case.

jamesgorrie:
> progress bar looks a bit nicer in this case

The only worry I have, similar to errors, is if they're not contextual, they're very often missed. Proven by the use of "warning" bars across the tools being almost completely ignored.

Would you expect this to usurp those loaders?

Because it doesn't :+1:

alastairjardine:
@jamesgorrie Fair thoughts :) Agree about them becoming missed – which is why the contextual ones are often better. 

There's also another dimension to consider as well (spoken with @akash1810 about it), which is time to complete actions. It feels fine to have the top progress bar for actions such as updating labels etc, which should complete in <1s. For actions that are generally ~1-10s blocking actions (i.e. user can't really do anything else), it makes more sense to have these as contextual and linked, or lightboxed and deliberately blocking. Finally, for longer, or indeterminate actions (downloading a lot of images, for example), it makes sense to have these occur in the background.

akash1810:
> Finally, for longer, or indeterminate actions (downloading a lot of images, for example), it makes sense to have these occur in the background.

Closing this PR in favour of implementing this behaviour.

PR #1479, opened 2015-11-23
Collection list editor

jamesgorrie:
Allows the ability to add / remove / read the lit of collections to show in the collection adventurer.
CC: @tsop14


common-lib/src/test/scala/com/gu/mediaservice/lib/collections/CollectionManagerTest.scala
```diff
@@ -0,0 +1,27 @@
+package com.gu.mediaservice.model
+
+import com.gu.mediaservice.lib.collections.CollectionsManager
+import org.scalatest.{FunSpec, Matchers}
+
+class CollectionManagerTest extends FunSpec with Matchers {
```
tsop14: :boar: should this be CollectionsManagerTest? There isn't a CollectionManager anymore?

jamesgorrie: yip - good :eyes:

common-lib/src/test/scala/com/gu/mediaservice/lib/collections/CollectionsManagerTest.scala
```diff
@@ -0,0 +1,27 @@
+package com.gu.mediaservice.model
+
+import com.gu.mediaservice.lib.collections.CollectionsManager
+import org.scalatest.{FunSpec, Matchers}
+
+class CollectionsManagerTest extends FunSpec with Matchers {
```
theefer: Nice to have some tests!

metadata-editor/app/controllers/CollectionsController.scala
```diff
@@ -0,0 +1,70 @@
...
+    } getOrElse Future.successful(invalidJson(req.body))
+  }
+
+  def addCollectionFromString(collection: String) = Authenticated.async { req =>
+    addCollection(CollectionsManager.stringToPath(collection), getUserFromReq(req))
+  }
+
+  private def addCollection(path: List[String], who: String) = {
+    val collection = Collection(path, ActionData(who, DateTime.now))
+    println(collection)
```
theefer: Hello world!

metadata-editor/app/controllers/CollectionsController.scala
```diff
@@ -0,0 +1,70 @@
...
+    CollectionsStore.add(collection).map { collection =>
+      respond(collection)
+    }
+  }
+
+  private def getUserFromReq(req: AuthenticatedRequest[_, Principal]) = req.user match {
+    case PandaUser(email, _, _, _) => email
+    case AuthenticatedService(name) => name
+    // We should never get here
+    case _ => "anonymous"
```
theefer: Can we log an error?

common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
```diff
@@ -0,0 +1,25 @@
...
+  val delimiter = "/"
+  val enc = "UTF-8"
+  def decodePathBit(s: String) = decode(s, enc)
+  def encodePathBit(s: String) = encode(s, enc)
+
+  def stringToPath(s: String) = s.split(delimiter).map(decodePathBit).toList
+  def pathToString(path: List[String]) = path.map(encodePathBit).mkString(delimiter)
+
+  def add(collection: Collection, collections: List[Collection]): List[Collection] =
+    collection :: collections.filter(col => col.path != collection.path)
```
theefer: Is the list sorted? Feels like it could benefit from being sorted by path?

metadata-editor/app/store/CollectionsStore.scala
```diff
@@ -0,0 +1,59 @@
...
+object CollectionsStore {
+  import Config.{awsCredentials, collectionsBucket}
+  val store = new JsonStore(collectionsBucket, awsCredentials, "collections")
+
+  def getAll: Future[List[Collection]] = store.getData flatMap { json =>
+    json.asOpt[List[Collection]].map(Future.successful) getOrElse Future.failed(InvalidCollectionJson(json))
+  }
+
+  def add(collection: Collection): Future[Collection] = {
+    store.getData flatMap { json =>
```
theefer: There's definitely a strong risk of race conditions here, especially as the store uses the previously cached data as based for the update (dirty read) and there are several servers with cached state.

I assume this is a placeholder for a store that will enforce strong concurrency (Dynamo?) ?

theefer: (The issue could be mitigated for now by forcing a store refresh before an update, although it still leaves the door open to concurrent writes, both on multiple and a single server)

theefer:
A few comments but looks nice generally! :+1:

metadata-editor/conf/routes
```diff
@@ -20,6 +20,12 @@ GET     /metadata/:id/usage-rights                      controllers.EditsControl
 PUT     /metadata/:id/usage-rights                      controllers.EditsController.setUsageRights(id: String)
 DELETE  /metadata/:id/usage-rights                      controllers.EditsController.deleteUsageRights(id: String)
 
+# Collections
+GET     /collections                                    controllers.CollectionsController.getCollections
+POST    /collections                                    controllers.CollectionsController.addCollectionFromJson
+PUT     /collections/*collection                        controllers.CollectionsController.addCollectionFromString(collection: String)
```
theefer: This shouldn't really be a PUT as it doesn't adhere to the HTTP spec of replacing the resource at the given URI with the provided entity.

common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
jamesgorrie: Can't think why - but happy to add it in.

metadata-editor/conf/routes
jamesgorrie: The `PUT` here is idempotent, except for updating the actionData due to [the way we add things](https://github.com/guardian/grid/pull/1479/files#diff-54243d7bda00027741bfd38375de1575R18). We could make a read endpoint to complete the puzzle, with little use though. 

I'm not sure I like it as it's nicer to get an actual list from the client that we then create the ID from, so I'm going to stick with the `POST` as we don't know what the ID is until we've made the collection.

common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
theefer: Easier to read manually (it's in order), and that should mean the server response is ordered as well, which makes it easier to browse visually, and potentially more efficient to aggregate on the client.

jamesgorrie: [Sorted](https://github.com/guardian/grid/commit/75dc7a9aade23ff1d3f5b49ff6b95e77c06f82b2)

metadata-editor/app/store/CollectionsStore.scala
jamesgorrie: :bike::house: until we need to, hopefully the abstraction of the CollectionStore will make it easy to swap out. this will work until we need to be a little more prod ready.

common-lib/src/test/scala/com/gu/mediaservice/lib/collections/CollectionsManagerTest.scala
jamesgorrie: [There's more](https://github.com/guardian/grid/commit/c7b9c6ff1ff65213e8deb38f83217db377b65148)

metadata-editor/conf/routes
theefer: You're right it's idempotent, but it's slightly odd to `PUT` with an empty body (technically the body could be anything as it's discarded by the code), and have that action result in some magical data materialising at that URI. The RFC says `The PUT method requests that the enclosed entity be stored under the supplied Request-URI.` -- technically the RFC doesn't specify precisely how that enclosed entity may vary from the entity returned by GETing that URI again, but the general assumption (which we have respected elsewhere in the Grid) is to return the exact same entity. The benefit of that is that you can respond with 204 No Content (cheaper) and still be confident of what data is at that resource (e.g. for caching or memoizing).

There's an interesting issue with server-generated information, such as the actionData, which generally led us to prefer POST in those cases.

Anyway this is all a little bit academic and you're removed the endpoint so I'll just shut up now :-)

Issue #1477, opened 2015-11-23
Getting error when trying to start any of the services: Cannot load plugin


ebrukson:
No source available, here is the exception stack trace:

An exception occurred during Plugin [play.api.GlobalPlugin] initialization

java.lang.ExceptionInInitializerError: 
     Global$.beforeStart(Global.scala:9)
     play.api.GlobalPlugin.<init>(GlobalSettings.scala:214)
     sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
     sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
     sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
     java.lang.reflect.Constructor.newInstance(Constructor.java:422)
     play.api.WithDefaultPlugins$$anonfun$plugins$1$$anonfun$apply$9.apply(Application.scala:132)
     play.api.WithDefaultPlugins$$anonfun$plugins$1$$anonfun$apply$9.apply(Application.scala:130)
     scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
     scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
     scala.collection.immutable.List.foreach(List.scala:381)
     scala.collection.TraversableLike$class.map(TraversableLike.scala:245)
     scala.collection.immutable.List.map(List.scala:285)
     play.api.WithDefaultPlugins$$anonfun$plugins$1.apply(Application.scala:130)
     play.api.WithDefaultPlugins$$anonfun$plugins$1.apply(Application.scala:166)
     play.utils.Threads$.withContextClassLoader(Threads.scala:21)
     play.api.WithDefaultPlugins$class.plugins(Application.scala:128)
     play.api.DefaultApplication.plugins$lzycompute(Application.scala:402)
     play.api.DefaultApplication.plugins(Application.scala:402)
     play.api.Play$$anonfun$start$1.apply$mcV$sp(Play.scala:91)
     play.api.Play$$anonfun$start$1.apply(Play.scala:91)
     play.api.Play$$anonfun$start$1.apply(Play.scala:91)
     play.utils.Threads$.withContextClassLoader(Threads.scala:21)
     play.api.Play$.start(Play.scala:90)
[...]

PR #1480, opened 2015-11-23
update to boto3.

akash1810:
From https://github.com/guardian/grid/issues/1477#issuecomment-159029232, I realised the properties generator was too configurable in comparison to other scripts in the project.

These changes mean:
- Only the profile can be configured via the settings file.
- The aws region is controlled by the value in the aws config file `~/.aws/config`, which is more conventional.
- The stack name is calculated in a similar way to [`stack-name.sh`](https://github.com/guardian/grid/blob/master/cloud-formation/scripts/stack-name.sh)

boto3 also logs more by default.

### DIFF
--- a/scripts/dot-properties/README.md
+++ b/scripts/dot-properties/README.md
@@ -28,6 +28,12 @@ NB: From the virtualenvwrapper docs:
 
 > You will want to add the command to `source /usr/local/bin/virtualenvwrapper.sh` to your shell startup file, changing the path to virtualenvwrapper.sh depending on where it was installed by pip.
 
+So:
+
+```sh
+echo "source /usr/local/bin/virtualenvwrapper.sh" >> ~/.profile
+```
+
 ## Usage
 
 ### Install requirements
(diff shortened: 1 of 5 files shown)


scripts/dot-properties/generate.py
```diff
@@ -9,27 +9,45 @@
...
 
         # Cropper service doesn't expect a protocol to be included,
         # remove the http:// string from the ImageOriginWebsite item.
-        outputs['ImageOriginWebsite'] = outputs['ImageOriginWebsite'].replace('http://', '')
+        if key == 'ImageOriginWebsite':
+            value = value.replace('http://', '')
 
-        return outputs
-    else:
-        LOGGER.warn('Stack not found: {stack}'.format(stack=stack_name_or_id))
```
akash1810: boto3 will raise a nicely formed exception if the stack doesn't exist, so we don't need to log anything.

scripts/dot-properties/main.py
```diff
@@ -15,16 +16,10 @@ def main():
     config = ConfigParser()
     config.read(CONFIG_FILE)
 
-    aws_profile_name = _default_option(config, 'aws', 'profile-name', 'media-service')
-    aws_region = _default_option(config, 'aws', 'region', 'eu-west-1')
```
akash1810: `aws_region` is controlled by `~/.aws/config`

scripts/dot-properties/main.py
```diff
@@ -15,16 +16,10 @@ def main():
...
-    cf_stack = _default_option(config, 'aws', 'stack-name', None)
```
akash1810: `cf_stack` is now calculated in a similar way to https://github.com/guardian/grid/blob/master/cloud-formation/scripts/stack-name.sh

scripts/dot-properties/main.py
```diff
@@ -15,16 +16,10 @@ def main():
...
-
-    if not cf_stack:
-        raise Exception('No stack found in {}'.format(CONFIG_FILE))
```
akash1810: `cf_stack` is now calculated in a similar way to https://github.com/guardian/grid/blob/master/cloud-formation/scripts/stack-name.sh

scripts/dot-properties/generate.py
```diff
@@ -9,27 +9,45 @@
...
         # remove the http:// string from the ImageOriginWebsite item.
-        outputs['ImageOriginWebsite'] = outputs['ImageOriginWebsite'].replace('http://', '')
+        if key == 'ImageOriginWebsite':
+            value = value.replace('http://', '')
+
+        LOGGER.info('{key} = {value}'.format(key=key, value=value))
 
-        return outputs
-    else:
-        LOGGER.warn('Stack not found: {stack}'.format(stack=stack_name_or_id))
```
akash1810: boto3 will raise a nicely formed exception if the stack doesn't exist, so we don't need to log anything.

scripts/dot-properties/README.md
```diff
@@ -28,6 +28,12 @@ NB: From the virtualenvwrapper docs:
 
 > You will want to add the command to `source /usr/local/bin/virtualenvwrapper.sh` to your shell startup file, changing the path to virtualenvwrapper.sh depending on where it was installed by pip.
 
+So:
+
+```sh
+echo "source /usr/local/bin/virtualenvwrapper.sh" >> ~/.profile
```
akash1810: convenience script for the line 29 above.

theefer:
:+1:

PR #1481, opened 2015-11-24
Add HTML linting to test process

theefer:
Should help catch malformed HTML, etc.

### DIFF
--- a/kahuna/package.json
+++ b/kahuna/package.json
@@ -52,6 +52,7 @@
     }
   },
   "devDependencies": {
+    "htmllint-cli": "0.0.4",
     "jscs": "^1.13.0",
     "jshint": "^2.5.11",
     "jspm": "~0.16.0"
--- a/kahuna/package.json
+++ b/kahuna/package.json
@@ -59,7 +59,7 @@
   },
   "scripts": {
     "test": "npm run lint",
-    "lint": "jscs public/js/ && jshint public/js --exclude public/js/dist/build.js",
+    "lint": "jscs public/js/ && jshint public/js --exclude public/js/dist/build.js && htmllint public/js/**/*.html",
     "dist": "jspm bundle js/main public/js/dist/build.js -m --inject --no-mangle",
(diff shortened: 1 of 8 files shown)


kahuna/public/js/usage-rights/usage-rights-editor.html
```diff
@@ -33,7 +33,7 @@
             </gr-icon>
         </div>
 
-        <label
+        <div
```
theefer: The label structure was not valid HTML, so I had to rejig things somewhat.

kahuna/package.json
```diff
@@ -52,13 +52,14 @@
...
   "devDependencies": {
+    "htmllint-cli": "0.0.4",
     "jscs": "^1.13.0",
     "jshint": "^2.5.11",
     "jspm": "~0.16.0"
   },
   "scripts": {
     "test": "npm run lint",
-    "lint": "jscs public/js/ && jshint public/js --exclude public/js/dist/build.js",
+    "lint": "jscs public/js/ && jshint public/js --exclude public/js/dist/build.js && htmllint public/js/**/*.html",
```
akash1810: presumably with this, we're encouraged to use standalone templates rather than embedding them in the directive?

kahuna/public/js/preview/image.html
```diff
@@ -30,6 +30,7 @@
        ui:drag-data="{{ctrl.image | asImageDragData}}">
        <div class="preview__fade"></div>
        <img class="preview__image"
+            alt=""
```
akash1810: `alt="{{::ctrl.image.data.metadata.description}}"` ?

akash1810:
:+1:

kahuna/package.json
theefer: Good point, it wouldn't check in the directives. We seem to be doing mostly separate templates but worth keeping in mind..

kahuna/public/js/preview/image.html
theefer: I wasn't sure about that but maybe that's better. I've updated.

PR #1482, opened 2015-11-24
Fix meta panel height and horizontal scrollbars

theefer:
Noticed that the labels dropdown was being cropped when talking to a user. Also noticed that long filenames caused horizontal scrolling in the gr-panel. This fixes both.

## Before
screenshot-2015-11-24t15 39 24
screenshot-2015-11-24t15 38 09

## After
screenshot-2015-11-24t15 38 32
screenshot-2015-11-24t15 37 57

### DIFF
--- a/kahuna/public/js/components/gr-panel/gr-panel.html
+++ b/kahuna/public/js/components/gr-panel/gr-panel.html
@@ -13,348 +13,348 @@
         </button>
     </div>
 
-    <div ng:if="ctrl.selectedImages.size > 0">
-        <div class="panel__top">
-            <ul class="costs">
-                <li ng:repeat="cost in ctrl.selectedCosts | orderBy:'data'">
-                    <div ng:switch="cost.data">
-                        <div ng:switch-when="free"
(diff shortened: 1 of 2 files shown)


kahuna/public/js/components/gr-panel/gr-panel.html
```diff
@@ -13,348 +13,348 @@
         </button>
     </div>
 
-    <div ng:if="ctrl.selectedImages.size > 0">
```
theefer: I decided to remove the extra `div` that caused issues, and move the `ng:if` onto both children. We could have kept the `div` but then would have to apply `height: 100%` onto it, which makes it an HTML+CSS fix that may be more likely to be broken again in the future (fix implies tight coupling of code far away). The HTML-only fixes seemed less brittle.

jamesgorrie: > fix implies tight coupling of code far away

:+1:

jamesgorrie:
:+1:

PR #1483, opened 2015-11-24
blocked cookies message

akash1810:
Adds a message for self diagnosis of blocked cookies.

screen shot 2015-11-24 at 15 45 57


akash1810:
the image is a little crude...

kahuna/public/js/errors/global.html
```diff
@@ -21,6 +26,11 @@
                target="_self" class="coloured-link">login again</a>
             to reauthenticate.
         </div>
+
+        <p>
+            If you can see this icon <img src="/assets/images/blocked-cookies.png" alt="blocked third party cookies" class="side-padded"/> in the address bar,
+            please follow <a href="https://docs.google.com/a/guardian.co.uk/document/d/1qlQhHM2NX3xWBIREB2O8D7S4ZtN0sYb7IDV2leE4Ubc/edit?usp=sharing" class="coloured-link" target="_blank">these instructions</a>.
```
theefer: Looks like you have to have a guardian.co.uk email to access this. Could you request it to be public for all instead?

theefer:
:+1:

theefer:
Nice!

kahuna/public/js/errors/global.html
akash1810: > Could you request it to be public for all instead?

@jamesgorrie I think you're the owner of the doc, can you change the permissions? I can't work out how to request access level changes from within Google Docs :cry:

jamesgorrie: Done.

paperboyo:
Hello,

Spaghetti Monster my witness that I’ve tried to create my first-ever PR, but git web interface does not allow uploading binary files, so maybe another time...

Here is an original Google’s blocked cookies icon, just optimised a bit better:

blocked_cookies_128colours

cookies icon before-after
[left: proposed (1,299 bytes); right: current (6,717 bytes)]

Regards
Mateusz

PR #1484, opened 2015-11-24
disable download button whilst downloading

akash1810:
first steps for feedback.

downloading

### DIFF
--- a/kahuna/public/js/downloader/downloader.html
+++ b/kahuna/public/js/downloader/downloader.html
@@ -1,0 +1,19 @@
+<button class="download"
+        type="button"
+        title="Download images"
+        ng:disabled="ctrl.downloading"
+        ng:if="ctrl.images.size > 1"
+        ng:click="ctrl.download()">
+    <gr-icon-label gr-icon="file_download">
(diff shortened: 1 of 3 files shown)


kahuna/public/js/downloader/downloader.html
```diff
@@ -0,0 +1,19 @@
...
+        ng:if="ctrl.images.size > 1"
+        ng:click="ctrl.download()">
+    <gr-icon-label gr-icon="file_download">
+        Download<span ng:show="ctrl.downloading">ing</span>
+    </gr-icon-label>
+</button>
+
+<a class="download"
+   ng:if="ctrl.images.size == 1"
+   ng:disabled="ctrl.downloading"
```
theefer: Do we ever get `downloading` when there's only one image?

theefer:
Nice and simple :+1:

kahuna/public/js/downloader/downloader.html
akash1810: Its a possibility, there are some pretty big images and some pretty slow connections...

PR #1486, opened 2015-11-25
tidy up the usage rights data model

jamesgorrie:
Having all those `?`s were really bothering me - so I thought I would try one last time at a tidy and actually got closer to getting to something that represents the model better.

An added benefit is the JSON parsing has been unified so that pattern now is: `case class data + category`.

This now reflects our data model more (`trait UsageRights`) and has separated out (almost) all of the business logic (`trait UsageRightsSpec`).


kenoir:
:+1:

PR #1487, opened 2015-11-25
Fix retry errors

kenoir:
Lambda functions being invoked multiple times clouds the lambda error logs, this PR causes such failure to **not** trigger a lambda error.

### DIFF
--- a/s3watcher/S3Watcher.js
+++ b/s3watcher/S3Watcher.js
@@ -14,7 +14,7 @@ exports.handler = function(event, context) {
 
     const duplicateTransfer = function(err) {
         const failState = err && (err.code === 'NotFound' || err.code === 'NoSuchKey');
-        const failComplete = Rx.Observable.from(["No source file."]);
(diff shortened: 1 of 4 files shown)


s3watcher/S3Watcher.js
```diff
@@ -12,9 +12,16 @@ exports.handler = function(event, context) {
         return Transfer.init(lambda.s3Event, config);
     });
 
+    const duplicateTransfer = function(err) {
+        const failState = err && (err.code === 'NotFound' || err.code === 'NoSuchKey');
+        const failComplete = Rx.Observable.from(["No source file."]);
```
theefer: Note that you can also use `Rx.Observable.return("No source file.")`.

theefer:
:+1:

s3watcher/S3Watcher.js
kenoir: Thanks, I was looking for that functionality.

Issue #1485, opened 2015-11-24
CloudFormation stack creation issue


ebrukson:
Getting this when attempting to create a stack

15:35:25 UTC-0500	CREATE_FAILED	AWS::DynamoDB::Table	LiveContentPollTable	Resource creation cancelled
Physical ID:media-service-DEV-ubuntu-LiveContentPollTable-1BTFSKA4UNA7Y
15:35:23 UTC-0500	CREATE_FAILED	AWS::Lambda::Function	UsageUpdaterFunction	Error occurred while getting the object from S3. S3 Error Code: NoSuchKey. S3 Error Message: The specified key does not exist.

kenoir:
You need to deploy the` usage-updater` artefact to S3 before running the CF uodate. 

Although you are probably better off omitting the CF for that service as the `usage` service is highly tied to Guardian architecture.

akash1810:
> Although you are probably better off omitting the CF for that service as the usage service is highly tied to Guardian architecture.

I'll take a look at putting these resources into a secondary, optional, stack as they're not core to running the Grid in general.

akash1810:
Hi @ebrukson, if you pull master you can use the [create-stack.sh](https://github.com/guardian/grid/blob/master/cloud-formation/scripts/create-dev-stack.sh) script.

ebrukson:
Thank you for looking into this.

I am getting an error now

ubuntu@GRID:/var/www/grid/public_html/cloud-formation/scripts$ ./create-dev-stack.sh
error: startswith is not defined
.Buckets[] | select(.Name | startswith("media-service-dist")) | .Name
                            ^^^^^^^^^^
1 compile error
[Errno 32] Broken pipe
Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='UTF-8'>
BrokenPipeError: [Errno 32] Broken pipe
Creating bucket media-service-dist-jahnishiengichiwieho
make_bucket: s3://media-service-dist-jahnishiengichiwieho/
usage: aws [options] <command> <subcommand> [parameters]
aws: error: argument subcommand: Invalid choice, valid choices are:

list
Uploading template to s3://media-service-dist-jahnishiengichiwieho/cloudformation/DEV/media-service-DEV-null.template
upload: ../dev-template.json to s3://media-service-dist-jahnishiengichiwieho/cloudformation/DEV/media-service-DEV-null.template
Creating stack media-service-DEV-null
[...]

PR #1488, opened 2015-11-25
split usage into separate stack.

akash1810:
As usage is strongly tied to CAPI, let it be its own stack.

Fixes https://github.com/guardian/grid/issues/1485.


kenoir:
:+1:

PR #1490, opened 2015-11-25
add note to create a permissions file

akash1810:
media-api will complain if this file doesn't exist.

`initialise-permissions.sh` is not run automatically because the stack creation needs to have finished (we could poll the cli, but that sounds a bit much...)

cc @nickpapacostas


theefer:
:+1:

PR #1492, opened 2015-11-26
persist commissioned agency images.

akash1810:


jamesgorrie:
:+1: Nice

jamesgorrie:
We should also send an email to our users explaining this at some stage.

PR #1494, opened 2015-11-26
remove collections mappings from mappings

jamesgorrie:
Removing for now as I would like to have more of a think as to where it should go.
Also we'd need a settings update, which is a little trickier than a remap.


kenoir:
:+1:

theefer:
Presumably that was never updated in TEST and PROD?

jamesgorrie:
@theefer apologies - the other PR #1494 fixes that.

PR #1495, opened 2015-11-26
update photographers list

jamesgorrie:
Checked against the master list and these are not meant to be there.
It's a repercussion of having been sent a few different list.

I think these are good now.


theefer:
Did you double check with Robert and Jo? I remember Jo talking about both these names, though can't recall the exact context.

jamesgorrie:
I've cross referenced these on previous discussions / emails / lists.
Happy to double check.

jamesgorrie:
Although Alicia definitely isn;t as she's a picture editor that sits with Joanna, so she would be under "other".

theefer:
Up to you, but at the very least I think we should email both of them to notify them of the change, so they can either confirm or push back, or just have a trail.

theefer:
Alicia definitely is a photographer though, search for `by:"Alicia Canter"`.

PR #1497, opened 2015-11-26
Remove Christian from photgraphers list

jamesgorrie:
Removing Christian.
Alicia stays as she is now staff.


theefer:
If checked :+1:

PR #1498, opened 2015-11-26
use cross region friendly s3 url

akash1810:
Buckets in the US don't follow the `https://s3-$REGION.amazonaws.com/$OBJECT_KEY` format for object urls. `https://$BUCKET.s3.amazonaws.com/$OBJECT_KEY` is more universal across regions.


jamesgorrie:
So this works in US and EU?
If so :+1:

akash1810:
> So this works in US and EU?

Yep. No idea why they'd use two different patterns though!

PR #1499, opened 2015-11-26
Collections app

jamesgorrie:
Basic app for collections.
This remove the `/collections` endpoint form the metadata app, but so what.


collections/app/lib/Config.scala
```diff
@@ -0,0 +1,30 @@
...
+object Config extends CommonPlayAppProperties with CommonPlayAppConfig {
+
+  val appName = "collections"
+
+  val properties = Properties.fromPath("/etc/gu/collections.properties")
+
+  val awsCredentials: AWSCredentials =
+    new BasicAWSCredentials(properties("aws.id"), properties("aws.secret"))
+
+  val dynamoRegion: Region = Region.getRegion(Regions.EU_WEST_1)
```
akash1810: Just realised we're hard coding the region. Presumably if I was to create a stack in another region nothing would work?

collections/app/lib/Config.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  val keyStoreBucket = properties("auth.keystore.bucket")
+
+  val collectionsBucket: String = properties("s3.collections.bucket")
```
akash1810: does this bucket need creating in cf?

collections/app/lib/Config.scala
theefer: I would not spend too much time worrying about that. That's an easy PR if someone is interested.

collections/conf/deploy.json
```diff
@@ -0,0 +1,24 @@
+{
+    "defaultStacks": ["media-service"],
+    "packages": {
+        "metadata-editor": {
```
theefer: That should be `collections` I think?

collections/app/lib/Config.scala
```diff
@@ -0,0 +1,30 @@
...
+  val imageCollectionsTable = properties("dynamo.table.imageCollections")
```
theefer: Will this need creating in CF?

project/Build.scala
```diff
@@ -57,6 +57,10 @@ object Build extends Build {
     .libraryDependencies(awsDeps ++ imagingDeps)
     .testDependencies(scalaTestDeps)
 
+  val collections = playProject("collections")
+    .libraryDependencies(awsDeps ++ playWsDeps)
```
theefer: Do you actually use `WS` (I didn't check).

theefer:
Nice one! :+1: 

Can I bother you with adding it to the `run*.sh` and `pingler` scripts in the root as well for completion?

collections/app/lib/Config.scala
jamesgorrie: It does [locally](https://github.com/guardian/grid/blob/master/cloud-formation/dev-template.json#L219-L240) and [publicly](https://github.com/guardian/grid-infra/blob/master/platform/cloud-formation/media-service.json#L609-L630).

project/Build.scala
jamesgorrie: Cheers, another copy paste

collections/conf/deploy.json
jamesgorrie: Copy paste :rage4:

collections/app/lib/Config.scala
jamesgorrie: [We do locally](https://github.com/guardian/grid/pull/1499/files#diff-a9f99695fda2ccf6f8fd009098777753R278) (which is what we need first) - I am adding the whole stack to the CF now.

jamesgorrie:
@theefer added - I still run crappy OSX so don't really use those so forgot.

collections/app/lib/Config.scala
akash1810: ah! I didn't check master :blush:

jamesgorrie:
@kenoir I am using 9010, which is greater than than 9009, I just typoed pingler.

PR #1502, opened 2015-11-27
warn of too big download.

akash1810:
Blobs have a limit of 500MB - warn of download failure.

Adds to https://github.com/guardian/grid/pull/1484 to give feedback on downloads.

### DIFF
--- a/kahuna/public/js/downloader/downloader.js
+++ b/kahuna/public/js/downloader/downloader.js
@@ -9,11 +9,12 @@ export const downloader = angular.module('gr.downloader', []);
 const maxBlobSize = 500 * 1024 * 1024;
 
 const bytesToSize = (bytes) => {
-    const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB'];
     if (bytes === 0) {
-        return 'n/a';
+        return '0 Bytes';
     }
-    const i = parseInt(Math.floor(Math.log(bytes) / Math.log(1024)));
+
+    const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB'];
+    const i = Math.floor(Math.log(bytes) / Math.log(1024));
 
     return i === 0 ?
         `${bytes} ${sizes[i]}` :
--- a/kahuna/public/js/downloader/downloader.js
+++ b/kahuna/public/js/downloader/downloader.js
@@ -6,7 +6,7 @@ import template from './downloader.html!text';
 export const downloader = angular.module('gr.downloader', []);
 
 // blob URLs have a max size of 500MB - https://github.com/eligrey/FileSaver.js/#supported-browsers
-const max_blob_size = 500 * 1024 * 1024;
+const maxBlobSize = 500 * 1024 * 1024;
 
 const bytesToSize = (bytes) => {
     const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB'];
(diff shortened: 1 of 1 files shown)


kahuna/public/js/downloader/downloader.js
```diff
@@ -24,9 +39,24 @@ downloader.controller('DownloaderCtrl',
...
+            if (blob.size <= max_blob_size) {
+                const url = $window.URL.createObjectURL(blob);
+                $window.location = url;
+            }
+            else {
+                const max_size = bytesToSize(max_blob_size);
+                const blob_size = bytesToSize(blob.size);
+
+                // the things we do for a happy linter...
+                const message = [
```
akash1810: I cried a little...

kahuna/public/js/downloader/downloader.js
```diff
@@ -5,6 +5,21 @@ import template from './downloader.html!text';
 
 export const downloader = angular.module('gr.downloader', []);
 
+// blob URLs have a max size of 500MB - https://github.com/eligrey/FileSaver.js/#supported-browsers
+const max_blob_size = 500 * 1024 * 1024;
```
theefer: We don't really use snake case in JS anywhere..?

kahuna/public/js/downloader/downloader.js
```diff
@@ -5,6 +5,21 @@ import template from './downloader.html!text';
...
+
+const bytesToSize = (bytes) => {
+    const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB'];
+    if (bytes === 0) {
+        return 'n/a';
+    }
+    const i = parseInt(Math.floor(Math.log(bytes) / Math.log(1024)));
```
theefer: What's the `parseInt` for?

kahuna/public/js/downloader/downloader.js
akash1810: oops :blush:

kahuna/public/js/downloader/downloader.js
```diff
@@ -24,9 +40,24 @@ downloader.controller('DownloaderCtrl',
...
+            else {
+                const maxSize = bytesToSize(maxBlobSize);
+                const blobSize = bytesToSize(blob.size);
+
+                // the things we do for a happy linter...
+                const message = [
+                    'Download is too big!',
+                    `Max file size: ${maxSize}. Your download: ${blobSize}.`,
+                    'Consider downloading smaller batches.'
+                ].join('\n');
```
jamesgorrie: Old school + new school.

jamesgorrie:
A little more fiddly - but could we split the downloads to make them fit?
:+1: as it's better than just failing.

kahuna/public/js/downloader/downloader.js
akash1810: Not needed, `Math.floor` already returns an int...

I've also refactored this function a little to make it simpler, returning fast when `bytes === 0`.

akash1810:
> could we split the downloads to make them fit?

Switching to server side downloading would be simpler I think. Might be worth time boxing in a separate PR to confirm?

theefer:
Let's talk to @blishen to review priorities re big downloads, as AFAIK it's only one user who seems to hit this problem (but yes it sucks for her).

paperboyo:
Before one hits 500MB download limit, one is likely to hit check-your-selection-for-holes-because-most-likely-stuff-in-the-middle-is-left-unselected selection limit. At least that's my experience. :-)

That's only a side note, warning on limit is a good idea!

jamesgorrie:
@paperboyo have you raised an issue for that?

paperboyo:
@jamesgorrie Ouch. You got me. :scream_cat: It's in my huge folder of "Stuff to report, that everybody knows of, just has no time, so no need to make an ass of myself reporting it".

Also: "No need to report that to him until we have something to report".

PR #1503, opened 2015-11-27
Foreman procfile

NickPapacostas:
allows you to run all the apps from one command


jamesgorrie:
Is this different to [this](https://github.com/guardian/grid/blob/master/run.sh
)

theefer:
run.sh only works on Linux. I don't mind the duplication as we explore better cross-platform ways to run the service. But yeah long term would be nice to have a single standard cross-platform way.

NickPapacostas:
one other advantage is you can run services individually a bit easier e.g. "foreman run media-api"

jamesgorrie:
:+1: bonza

PR #1504, opened 2015-11-27
adding profile to setup scripts and down casing username for s3 buckets

NickPapacostas:
made these changes while working through the setup process

### DIFF
--- a/cloud-formation/scripts/post-dev-stack-creation.sh
+++ b/cloud-formation/scripts/post-dev-stack-creation.sh
@@ -3,6 +3,8 @@
 source ./utils.sh
 source ./stack-name.sh
 
+export AWS_PROFILE=media-service
+
 CROPPER_KEY=cropper-`lower_case_random_string`
 CONFIG_BUCKET=`get_stack_resource_physical_id $STACK_NAME ConfigBucket`
 KEY_BUCKET=`get_stack_resource_physical_id $STACK_NAME KeyBucket`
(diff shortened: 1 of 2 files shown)


cloud-formation/scripts/post-dev-stack-creation.sh
```diff
@@ -3,6 +3,8 @@
 source ./utils.sh
 source ./stack-name.sh
 
+export AWS_PROFILE=media-service
```
jamesgorrie: Some people might have multiple profiles they work on - wouldn't this be inconvenient for them?

akash1810: would this need to go above `source ./stack-name.sh` so that its set before any call to the awscli is made?

theefer: Note that a shell script cannot set an environment variable outside of its execution. IOW `AWS_PROFILE` is only set within that script, unless you `source` the script instead of executing it.

theefer: @akash1810 We tried and no, since these scripts only define functions that call the awscli, they don't call the cli yet.

akash1810: Cool.

This has just made me realise `AWS_DEFAULT_PROFILE` is [legacy](https://github.com/aws/aws-cli/issues/1281#issuecomment-92509459) too! (We mention setting `AWS_DEFAULT_PROFILE` in the project root readme).

akash1810:
:+1:

theefer:
:+1:

Issue #1425, opened 2015-11-04
Keywords disappear when metadata edited


tsop14:
![disappearing keywords](https://cloud.githubusercontent.com/assets/8484757/10935946/32b28626-82e1-11e5-95b4-5eececf5b67d.gif)

theefer:
Looks like editing applies `keywords: []` (default) as an override on the `metadata`. It shouldn't.

PR #1506, opened 2015-11-27
Omit keywords if empty list

theefer:
Fixes https://github.com/guardian/grid/issues/1425, which is a bug with overrides caused by keywords defaulting to empty list.

In theory, we should allow both to omit `keywords` and have it set to `[]` in the overrides, as a way to override the keywords in the originalMetadata. In practice, we never allowed the keywords to be edited in the first place so that point is kind of moot.

I can't think of any reason why this change could cause any issues, particularly given how little we use keywords.

Every single image that has metadata edits (i.e. overrides) will currently have `keywords: []` in ES both for `userMetadata` and `metadata`. We'll need to clean that data once this is released with some kind of ES cleanup task.


tsop14:
:+1:

akash1810:
:+1: 

Do we have tracking on searches by keyword?

theefer:
Might be able to just search the access logs?

jamesgorrie:
Don't know why we didn't do this earlier... another PR would be to clean up a couple of hacks in the code to deal with this. :+1:

PR #1507, opened 2015-11-27
read aws region from property file

akash1810:
Allows the Grid to be run in any region.

We're currently hard-coding the dynamo region to eu-west-1. This will only work iff the stack is created in that region!


jamesgorrie:
Updated in CF for prod?

akash1810:
> Updated in CF for prod?

Have done now - https://github.com/guardian/grid-infra/pull/178.

jamesgorrie:
:+1:

PR #1510, opened 2015-11-30
[WIP] adds jonny weeks to staff photographers list

tsop14:


akash1810:
:+1:

theefer:
Can we respect the alphabetical order for ease of maintenance? Apart from that :+1:

jamesgorrie:
Still tentative from me - trying to figure out exactly what we're doing here - but if you think it solves an immediate problem that can be polished later :+1:

tsop14:
Not a good idea to add someone to a list they should not be on.

Issue #1459, opened 2015-11-16
Disappearing description.


akash1810:
Steps to reproduce:
- Upload an image with no description in its metadata (screenshots are a simplest).
- Witness the description field populated with the filename.
- Set the usage rights of the image.
- Witness the description field being cleared.

PR #937, opened 2015-07-15
use filename if description is missing

jamesgorrie:
A bit stateful, but this whole bit is really.

Waiting to hear back from Jo if certain characters often represent spaces, but we could add that later once she's used it as an iteration.

PR #1511, opened 2015-11-30
write filename to description if no description in image metadata

akash1810:
Fixes https://github.com/guardian/grid/issues/1459.


tsop14:
This was being done before #937 - was it removed or did it break?

akash1810:
> This was being done before #937

That PR was only setting the value on the client and wan't making a request to the metadata-editor to save it.

tsop14:
> Neither - it just didn't exist :smile:.

Until now... :+1:

theefer:
Cool. @NickPapacostas had already started working on this I think, my bad, so maybe we need to communicate more around who's working on what, either via GitHub issues or Slack or vocal-chord-chat.

akash1810:
ah. +1 for more comms!

PR #1513, opened 2015-12-01
read aws region from property file

akash1810:
correcting https://github.com/guardian/grid/pull/1507

`RegionUtils.getRegion` is what `Region.getRegion` does ([link](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/regions/Region.java#L59-L61))


akash1810:
merging as it fixes https://github.com/guardian/grid/pull/1512

PR #1514, opened 2015-12-01
deploy s3watcher w/teamcity + riffraff

akash1810:


s3watcher/build.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+get_function_name() {
+    echo $(aws cloudformation list-stack-resources --stack-name $1 \
```
theefer: Let's remember to always pass in the `--profile` explicitly with every `aws` command?

s3watcher/build.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
```
theefer: You may want a `set -e` here to abort if anything fails?

s3watcher/build.sh
theefer: Also is this meant to run on TC? How will it have the right credentials to do so?

theefer:
Nice, just a few questions above on the build script.

s3watcher/build.sh
akash1810: > How will it have the right credentials to do so?

Environment variables.

theefer: Can you have a chat with @adamnfish about whether this is OK security-wise?

akash1810:
merging so I can setup teamcity

PR #1517, opened 2015-12-01
Nodes as resources

jamesgorrie:
After much jiggery pokery with the JSON parsers again, there's a way to have lists of `EmbeddedEntity`s with the Subtype of the same class that they're a list of. i.e.
`case class A(a: String, b: List[A]) ` => `ResposeEntity(data: List[EmbeddedEntity[A]])`

This should hopefully allow us to simplify the client code. Not a refactor, probably just a few subtractions.

CC: @tsop14

### DIFF
--- a/collections/app/controllers/CollectionsController.scala
+++ b/collections/app/controllers/CollectionsController.scala
@@ -35,21 +35,20 @@ object CollectionsController extends Controller with ArgoHelpers {
   val Authenticated = ControllerHelper.Authenticated
 
   def uri(u: String) = URI.create(u)
+  val collectionUri = uri(s"$rootUri/collections")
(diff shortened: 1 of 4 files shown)


jamesgorrie:
[A verbose JSON representation](https://gist.github.com/jamesgorrie/c1be43224a393fb2fee1)

theefer:
Nice! :+1: 

Open question: do we still want the `content` wrapper? I don't mind too much, just curious why.

jamesgorrie:
Yea - it's starting to become a little useless, although it does contain the action data, which is useful. It's because I started working on a node creator, not a tree of collections. It might be useful if we ever want a tree of something else? :sunrise_over_mountains:

PR #1518, opened 2015-12-02
usage rights => rights & restrictions

jamesgorrie:
I quite likes Jonny's suggestion, so had a go and still like it.
rr-results
rr-uploads

### DIFF
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
@@ -1,7 +1,7 @@
 <div class="image-info">
     <div class="image-info__group">
         <dl class="image-info__wrap metadata-line image-info__usage-rights">
-            <dt class="metadata-line__key">Usage rights category</dt>
+            <dt class="metadata-line__key">Rights & restrictions</dt>
 
             <gr-usage-rights-editor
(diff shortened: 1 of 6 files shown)


kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
```diff
@@ -1,7 +1,7 @@
 <div class="image-info">
     <div class="image-info__group">
         <dl class="image-info__wrap metadata-line image-info__usage-rights">
-            <dt class="metadata-line__key">Usage rights category</dt>
+            <dt class="metadata-line__key">Rights & restrictions</dt>
```
akash1810: `&amp;` rather than `&`?

akash1810:
:+1:

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
jamesgorrie: I don't  think that's necessary nowadays?

theefer: Interestingly, I just tried in the W3C validator and it's not (anymore?). `a & b` is seemingly valid. However `a &x b` is not.

theefer:
Nice one :+1:

PR #1519, opened 2015-12-02
positive positive on drag logicc

jamesgorrie:
# Problem
At the moment we enable dnd when we aren't dragging something with drag data from the grid. This will create false positives on elements (not `img` or `a`) that have `draggable=true` (which we will be needing for collections).

# Solution
We now look to see is we are able to accept the draggable, by looking through the `dataTransfer.types`.

# TODO
~~I am looking into what we do about elements that have `uri-list`s from the grid i.e. they don't have `drag-data` but are images or links. This is a problem we have on the Grid already, so not in the PR.~~

### DIFF
--- a/kahuna/public/js/main.js
+++ b/kahuna/public/js/main.js
@@ -36,7 +36,14 @@ var config = {
     assetsRoot: assetsRootLink && assetsRootLink.getAttribute('href'),
 
     // Static config
-    'pandular.reAuthUri': '/login'
+    'pandular.reAuthUri': '/login',
+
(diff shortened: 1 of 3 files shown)


theefer:
:+1:

jamesgorrie:
@theefer I've added the mimeTypish data to the `dataTransfer` to stop images and links from the grid kicking off a upload dnd action. What do you think?

theefer:
Commented on the commit, besides that :+1:

jamesgorrie:
Commenting Done.

PR #1521, opened 2015-12-02
push s3-watcher logs to logstash

akash1810:
Push s3-watcher logs to logstash with a lambda subscribed to the s3-watcher cloud watch log group.

**TODO**
- [x] push logs to kinesis stream that's attached to logstash
- [x] work out deployment process - cloud formation?


cloudwatch-logs-to-logstash/lambda/index.js
```diff
@@ -0,0 +1,79 @@
...
+                const roleRequest = sts.assumeRole({
+                    RoleArn: config.roleArn,
+                    RoleSessionName: 'cloudwatch-logs-to-logstash'
+                });
+
+                roleRequest.send(function (err, data) {
+                    if (err) {
+                        context.fail(err);
+                    } else {
+                        const kinesis = new AWS.Kinesis({
```
kenoir: Can you extract Kinesis interaction into another lib?

akash1810: yep!

cloudwatch-logs-to-logstash/build.sh
```diff
@@ -0,0 +1,36 @@
...
+    echo 'PROD_FUNC_NAME is unset';
+    exit 1
+fi
+
+cd lambda
+npm install
+
+npm test
+
+# TODO get or create config.json
```
akash1810: I can't think of a decent way to do this; writing CI env vars to disk doesn't feel that nice.

I'm tempted to leave CI/CD for now...

kenoir: Difficult one. I wonder whether we really do need to worry about secrets in the CI environment?

akash1810: Opted for [manual deploy](https://github.com/guardian/grid/commit/ac5e89684d94a0e2aafb5d97a32d715b75fd0175) in this instance as this lambda won't update very often (if at all) as it just extracts json, then pushes to kinesis.

cloudwatch-logs-to-logstash/lambda/index.js
```diff
@@ -0,0 +1,41 @@
...
+
+            logs.forEach(function(log) {
+                const messageParts = log.message.split('\t');
+
+                if (! isNaN(Date.parse(messageParts[0]))) {
+                    const rawMessage = messageParts[messageParts.length - 1]
+                        .replace(/(')/gm, '"')
+                        .replace(/(\n)/gm, '');
+
+                    eval('const message = ' + rawMessage);
```
jamesgorrie: `eval` is never the answer, and can lead to security flaws...

jamesgorrie: Could we JSON parse it?

akash1810: > Could we JSON parse it?

Not easily. At this stage, `rawMessage` looks like:

```
'{ message: "Downloading from ingest bucket.",  state:    { Bucket: "media-service-test-s3watcheringestbucket-XXXXX",     Key: "YYYYY.jpg" } }'
```

Note the keys are not quote enclosed, so we'd need to do some funky regex magic first (I couldn't work out how). There are also (hidden) `\t` characters between key and value :cry:.

`eval` seemed like the nicest solution (albeit horrible).

Will add a note about this.

cloudwatch-logs-to-logstash/config.json
```diff
@@ -0,0 +1,4 @@
+{
```
jamesgorrie: Worth naming this file with `-template` or something, to make sure we don't commit accidentally?

cloudwatch-logs-to-logstash/lambda/.jscsrc
```diff
@@ -0,0 +1,43 @@
+{
+//    "disallowMultipleVarDecl": "exceptUndefined",
```
jamesgorrie: Are we going to go with one? Or just remove both?

cloudwatch-logs-to-logstash/lambda/index.js
```diff
@@ -0,0 +1,46 @@
...
+
+
+                    //`rawMessage` is a string that looks like
+                    //{message: "Downloading from ingest bucket", state: {Bucket "foo-bar"}}
+                    //As the keys in are not quoted, eval is the simplest way to convert to JSON.
+                    eval('const message = ' + rawMessage);
+
+                    console.log(message);
+
+                    records.push({
```
jamesgorrie: Why don't we just map to a new records var?

cloudwatch-logs-to-logstash/lambda/index.js
```diff
@@ -0,0 +1,46 @@
...
+        } else {
+            result = JSON.parse(result.toString('utf8'));
+
+            const logs = result.logEvents;
+            const records = [];
+
+            logs.forEach(function(log) {
+                const messageParts = log.message.split('\t');
+
+                if (! isNaN(Date.parse(messageParts[0]))) {
```
jamesgorrie: What's the case of not having the date?

cloudwatch-logs-to-logstash/lambda/index.js
```diff
@@ -0,0 +1,48 @@
...
+                    const rawMessage = messageParts[messageParts.length - 1]
+                        .replace(/(')/gm, '"')
+                        .replace(/(\n)/gm, '');
+
+
+                    // `rawMessage` is a string that looks like
+                    // {message: "Downloading from ingest bucket", state: {Bucket "foo-bar"}}
+                    // As the keys in are not quoted, eval is the simplest way to convert to JSON.
+                    eval('const message = ' + rawMessage);
```
theefer: Might want to extract this hack into its own function for clarity, and surround with a `try-catch` in case of errors?

cloudwatch-logs-to-logstash/lambda/index.js
akash1810: Filtering out log lines that were generated in the normal lifecycle of a lambda execution, e.g. duration, memory use. Log lines with timestamps were created explicitly from a lambda calling `console.log()`.

Added comment.

cloudwatch-logs-to-logstash/lambda/index.js
akash1810: Done.

cloudwatch-logs-to-logstash/lambda/index.js
akash1810: Done.

cloudwatch-logs-to-logstash/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Developing
+This lambda puts messages to a Kinesis stream. When run from within the AWS environment, the Lambda gets credentials from an execution role.
+In DEV, we can use our local credentials from `~/.aws/credentials` by setting the value `stage` to `DEV` in `lambda/config.json`;
+
+
+## Deploying
+This Lambda has not been added to CI/CD due to the slight complexity of getting secrets into the Lambda.
+
+To deploy, run `./deploy.sh` then update the Lambda function code.
```
theefer: As suggested, `deploy.sh` could be renamed to `build.sh` to better represent what it does.

theefer:
Besides rename, :+1: !

kenoir:
:+1: nice. Still be good to break this out into a separate repo, it definitely feels like something others might want to use by itself.

akash1810:
> Still be good to break this out into a separate repo, it definitely feels like something others might want to use by itself.

Yeah. I haven't thought of a nice way to do this yet as the cf template is horrible with parameters. Ideally the template for this would be nested in the parent service to avoid copy/paste... More thinking needed!

PR #1524, opened 2015-12-03
Option to download low-res image

theefer:
Uses the imgops service to generate an image at most 800x800 pixels, useful for syndication so they can send a low res preview of an image before sending the original once paid for.

One caveat is that the imgops service doesn't work well with non-sRGB images (i.e. they don't get converted to sRGB), but I think it's no worse than sending the original.

image

cc @paperboyo

### DIFF
--- a/kahuna/public/js/image/controller.js
+++ b/kahuna/public/js/image/controller.js
@@ -38,6 +38,7 @@ image.controller('ImageCtrl', [
     'image',
     'mediaApi',
     'optimisedImageUri',
+    'lowResImageUri',
     'cropKey',
     'mediaCropper',
     'imageService',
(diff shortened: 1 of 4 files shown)


jamesgorrie:
:+1:

paperboyo:
Just wanted to use it (who would have thought?) for emailing previews. But it didn't work for multiple selection, so I haven't. Also, in the far away future maybe user prefs could store the resolution (and quality) of these?

jamesgorrie:
>  Also, in the far away future maybe user prefs could store the resolution (and quality) of these?

What are the use cases for this? Just asking as it would add a bit of complexity?
Surely the settings now are okay for emailing?

> Multiselection
I think there was mention of this, we might try later.

paperboyo:
Okay they are. I would adjust them depending on may knowledge of the recipient's screen/email app, though. There may be other cases (like quick slideshow of the whole batch) that would go away after we get prev/next in image view :-) This is just thinking out loud!

> we might try later.

Again, I was just amazed that I wanted to use it in the first place and wanted to let you know. Not expecting to need that any time soon (just as I haven't expected to use it at all :).

PR #1527, opened 2015-12-04
More preset dates to filter.

akash1810:
Adding:
- Past 6 months
- Past year

screen shot 2015-12-04 at 11 26 18


theefer:
Was that following someone's request? :+1:

akash1810:
Yeah, sorry should have mentioned that. Requested from Mark on the Sports desk.

PR #1528, opened 2015-12-04
Image collections

jamesgorrie:
Adds collections to the Collections Dynamo DB.
The store is a `ImageId => List[Collection]` store.

I thought we could do a store of row = `ImageId => CollectionId => Collection`, but then when returning the list we would have to do a look up every time, so I thought keeping the list of collections together might be best.

The side-effect of this is that we can have multiple collections in a list (two people having added it), which might actually have a nice side-effect of being able to create a bit of an audit trail.

`POST collection.com/images/:imageId -d '{ "data":["g2", "film"] }'`
```JSON
{
	"data": [{
		"pathId": "g2/art",
		"path": [
			"g2",
			"art"
		],
		"actionData": {
			"author": "test",
			"date": "2015-11-30T21:33:21.956+00:00"
		}
	}, {
		"pathId": "g2/film",
		"path": [
			"g2",
			"film"
		],
		"actionData": {
			"author": "test",
			"date": "2015-12-07T11:23:06.985+00:00"
		}
	}]
}
```


collections/app/controllers/ImageCollectionsController.scala
```diff
@@ -0,0 +1,55 @@
...
+  import CollectionsManager.onlyLatest
+
+  val Authenticated = ControllerHelper.Authenticated
+  val dynamo = new DynamoDB(Config.awsCredentials, Config.dynamoRegion, Config.imageCollectionsTable)
+
+  def getCollections(id: String) = Authenticated.async { req =>
+    dynamo.listGet[Collection](id, "collections").map { dynamoEntry =>
+      respond(onlyLatest(dynamoEntry))
+    } recover {
+      case NoItemFound => respondNotFound("No usage rights overrides found")
```
akash1810: `No usage rights overrides found` copy + paste?

jamesgorrie: :eyes: changed

collections/app/controllers/ImageCollectionsController.scala
```diff
@@ -0,0 +1,56 @@
...
+      dynamo.listAdd(id, "collections", collection)
+        .map(publish(id))
+        .map(cols => respond(collection))
+    } getOrElse Future.successful(respondError(BadRequest, "invalid-form-data", "Invalid form data"))
+  }
+
+  def publish(id: String)(collections: List[Collection]): List[Collection] = {
+    val message = Json.obj(
+      "id" -> id,
+      "data" -> Json.toJson(onlyLatest(collections))
```
theefer: Do we want to send a diff (like we do for exports) or a full snapshot (like we do for usage and metadata edits)? I'd err towards the latter, else we need another special message for deletions and we are more at risk of concurrency issues (e.g. out of order messages).

common-lib/src/test/scala/com/gu/mediaservice/lib/aws/DynamoDBTest.scala
```diff
@@ -0,0 +1,80 @@
+package com.gu.mediaservice.lib.aws
+
+import com.amazonaws.services.dynamodbv2.document.utils.ValueMap
+import com.gu.mediaservice.model.{ActionData, Collection}
+import org.joda.time.DateTime
+import org.scalatest.{FunSpec, Matchers}
+import play.api.libs.json.{Format, JsObject, Json}
+
+class DynamoDBTest extends FunSpec with Matchers {
```
theefer: Nice to have!

jamesgorrie: I'd like to move more of the none IO stuff out to test as well.

collections/app/controllers/ImageCollectionsController.scala
jamesgorrie: At the moment we're sending over the whole collection - so no diffing, just set `image.collections = collections`.

jamesgorrie: With deleting we'll do that same, just send the new list [e.g.](https://github.com/guardian/grid/compare/image-collections...image-collection+thrall-notification?expand=1)

theefer: I must have misunderstood what `onlyLatest` does then?

jamesgorrie: As the actual saved list (in Dynamo) could be:
```JSON
{
    "data": [{
        "pathId": "g2/film",
        "path": [
            "g2",
            "film"
        ],
        "actionData": {
            "author": "test",
            "date": "2015-11-30T21:33:21.956+00:00"
        }
    }, {
        "pathId": "g2/film",
        "path": [
            "g2",
            "film"
        ],
        "actionData": {
            "author": "test",
            "date": "2015-12-07T11:23:06.985+00:00"
        }
    }]
}
```

Only latest makes sure that we only use the latest addition.
The reason for this is to keep list storage simple, and just manipulate it after. It also give us an audit trail.

I might write a test to make it more explanitory.

theefer: Oh I see. OK :+1:

jamesgorrie: [Test added](https://github.com/guardian/grid/commit/337b5ed6901d806f1012d74884e7d5c4e2fdaeb2)

theefer: Thanks!

theefer:
Looks legit :+1:

PR #1530, opened 2015-12-04
change what's logged on upload

akash1810:
We have been logging secrets :fearful:. This PR removes that and also changes the shape of what's logged for consumption by https://github.com/guardian/grid/pull/1521.

release notes:
- [ ] create new api key
- [ ] rotate keys in s3-watcher config
- [ ] check s3-watcher is still working
- [ ] delete old api key

### DIFF
--- a/s3watcher/lambda/lib/Logger.js
+++ b/s3watcher/lambda/lib/Logger.js
@@ -1,0 +1,41 @@
+const messages = {
+    DOWNLOAD: "Downloading from ingest bucket.",
+    UPLOAD: "Uploading to image-loader.",
+    DELETE: "Deleting from ingest bucket.",
+    COPY_TO_FAIL: "Copying to fail bucket.",
+    RECORD: "Recording result to Cloud Watch",
+    UPLOAD_FAIL: "Upload failed.",
+
+    LAMBDA_ERROR: "Lambda failure",
+    LAMBDA_SUCCESS: "Finished successfully."
+};
+
+const level = {
+    INFO: "INFO",
+    ERROR: "ERROR"
+};
+
+const baseMessage = function (message, level, state) {
+    return {
+        stack: "media-service",
+        app: "s3-watcher",
+        timestamp: new Date().toISOString(),
+        level: level,
+        message: message,
+        state: state
+    };
+};
+
+module.exports = {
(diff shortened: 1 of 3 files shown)


jamesgorrie:
Oops. :+1:

s3watcher/lambda/lib/Transfer.js
```diff
@@ -59,7 +59,12 @@ module.exports = {
             log("download", s3Object);
 
             return S3Helper.getS3Object(s3Object).flatMap(function(data){
-                log("upload", upload);
+                log("upload", {
+                    size: upload.size,
+                    filename: upload.params.filename,
+                    uploadedBy: upload.params.uploadedBy,
+                    stage: upload.params.stage
+                });
```
akash1810: `upload` object is from [here](https://github.com/guardian/grid/blob/master/s3watcher/lambda/lib/Upload.js#L23-L34).

akash1810:
I'm going to split out the changes to s3-watcher logging from https://github.com/guardian/grid/pull/1521 into this PR as they are very closely related. It will also make deploying simpler.

s3watcher/lambda/lib/Logger.js
```diff
@@ -0,0 +1,42 @@
...
+};
+
+module.exports = {
+    messages: messages,
+
+    log: function (stage, messageKey, state) {
+        console.log(baseMessage(stage, messageKey, level.INFO, state));
+    },
+
+    error: function (stage, state, err) {
```
jamesgorrie: Minor style question - instead of having a client call:
`Logger.log(config.stage, Logger.messages.DELETE, s3Object);`

Why not expose `logDelete()` that does this for you? It would stop the client having to know about the `messages` object.

jamesgorrie:
One tiny style question - but :+1:

s3watcher/lambda/lib/Logger.js
akash1810: I was aiming to keep `Logger` as simple as possible... though I agree the `messages` object is ugly.

akash1810:
keys rotated in TEST

akash1810:
keys rotated in PROD now too

PR #1531, opened 2015-12-04
Add drag data for multiple images

jamesgorrie:
I've stuck it in the results view as that's where we resolve the `selectedImages$`. This becomes a little weird as we can only have them as deps on the view.

This will now allow us to have clients of the drop data use the information of the selection.

__Next__
Theseus work to deserialise this into resources again.

e.g.
```JavaScript
window.addEventListener('drop', ev => {
  const t = ev.dataTransfer.getData(vndMimeTypes.get('gridImagesData'));
  console.log(t);
});
```

### DIFF
--- a/kahuna/public/js/main.js
+++ b/kahuna/public/js/main.js
@@ -39,12 +39,13 @@ var config = {
     'pandular.reAuthUri': '/login',
 
     vndMimeTypes: new Map([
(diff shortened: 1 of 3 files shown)


jamesgorrie:
@theefer as per discussion withLatestFrom > combineLatest.

theefer:
:+1:

PR #1532, opened 2015-12-04
single jscsrc file.

akash1810:
I'll be linting the cloudwatch-to-logstash lambda and need these rules again. Lets just symlink for simplicity.


akash1810:
ugh, I forgot that teamcity can't cope with symlinks with server side checkouts. Closing.

PR #1534, opened 2015-12-04
minor updates on packages

jamesgorrie:
Did a JSPM install and it update a few things. 
 I have gone through the system and everything seems to work. Happy to hold off till Monday.


theefer:
:+1: let's wait or do a proper full manual regression of the critical path.

PR #1535, opened 2015-12-04
use illustrators proper names

jamesgorrie:
So far the consensus is the credit is the illustrators name.
The actual limited name of the company is not really useful, and never shown anywhere except maybe their tax statements.

I will retrospectively go through and amend.


theefer:
:+1:

paperboyo:
But this still won't put this same name into the Byline field, right? I think it should now as Composer removes the redundancy.

PR #1536, opened 2015-12-04
adds a customised drag icon when images are selected

tsop14:
dragging

### DIFF
--- a/kahuna/public/js/search/index.js
+++ b/kahuna/public/js/search/index.js
@@ -149,7 +149,19 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
 
                     const sub = dragData$.subscribe(({ images, dt }) => {
                         if (images.size > 0) {
+                            //creates an element to use as the drag icon
+                            const dragImage = document.createElement('div');
+                                  dragImage.classList.add("dragIcon");
+
+                            const imageCount = document.createElement('span');
+                                  imageCount.classList.add("dragCount");
+
+                                  imageCount.innerHTML = images.count();
+                            dragImage.appendChild(imageCount);
+                            document.body.appendChild(dragImage);
+
                             dt.setData(vndMimeTypes.get('gridImagesData'), JSON.stringify(images));
+                            dt.setDragImage(dragImage, 0, 0);
                         }
                     });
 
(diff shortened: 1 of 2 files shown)


kahuna/public/js/search/index.js
```diff
@@ -149,7 +149,19 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
...
+                            //creates an element to use as the drag icon
+                            const dragImage = document.createElement('div');
+                                  dragImage.classList.add("dragIcon");
+
+                            const imageCount = document.createElement('span');
+                                  imageCount.classList.add("dragCount");
+
+                                  imageCount.innerHTML = images.count();
+                            dragImage.appendChild(imageCount);
+                            document.body.appendChild(dragImage);
```
theefer: Does it really need to be added to the document, or can it be a dangling element or a document fragment? (If so, does it need to be garbage collected after the drag?)

kahuna/public/js/search/index.js
```diff
@@ -149,7 +149,19 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
...
+
                             dt.setData(vndMimeTypes.get('gridImagesData'), JSON.stringify(images));
+                            dt.setDragImage(dragImage, 0, 0);
                         }
```
theefer: Will we want another placeholder when a single (non-selected) image is being dragged?

theefer:
Nice! A comment about the DOM, but looks good. It does look quite faded in the screencast, but I don't see any opacity styling so I assume that's how the browser does it?

kahuna/public/js/search/index.js
tsop14: Yes. This function is just dealing with selections so non-selected image icon settings will go somewhere else. Not sure what it should be - at the moment it is the default thumbnail.

kahuna/public/js/search/index.js
jamesgorrie: Looking at the example [here](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/setDragImage) it doesn't have to be, but I am sure I have seen it somewhere that it might.

jamesgorrie: @tsop14 [I've made it attach to the parent element so that it's automatically garbage collected](https://github.com/guardian/grid/commit/c7e78edd9c49663573b457a07d4a9e582a320795).
@theefer it seems you do need to have it attached if it's a DOM element.

jamesgorrie:
@theefer re: opacity - yip.

jamesgorrie:
@tsop14 If you're happy with my change - :+1: to this.

tsop14:
:+1:

PR #1539, opened 2015-12-07
Zero width space

NickPapacostas:
small change to fix a problem some people were experiencing when copy-pasting from emails to search. 

Zero-width-space is a valid UTF-8 character but because it's invisible can cause confusion. I know it's a particularly specific fix, but it seemed better than doing something fancy.

### DIFF
--- a/kahuna/public/js/search/index.js
+++ b/kahuna/public/js/search/index.js
@@ -35,15 +35,28 @@ export var search = angular.module('kahuna.search', [
 search.config(['$stateProvider', '$urlMatcherFactoryProvider',
                function($stateProvider, $urlMatcherFactoryProvider) {
 
-    // Query params slashes are encoded that mess up our seach query when we search for e.g. `24/7`.
-    // see: https://github.com/angular-ui/ui-router/issues/1119#issuecomment-64696060
     // TODO: Fix this
-    $urlMatcherFactoryProvider.type('Slashed', {
-        encode: val => angular.isDefined(val) ? val : undefined,
-        decode: val => angular.isDefined(val) ? val : undefined,
-        // We want to always match this type. If we match on slash, it won't update
-        // the `stateParams`.
-        is: () => true
+    $urlMatcherFactoryProvider.type('Query', {}, () => {
+        var queryFilters = {
+            // Query params slashes are encoded that mess up our seach query when we search for
+            // e.g. `24/7`.
+            // see: https://github.com/angular-ui/ui-router/issues/1119#issuecomment-64696060
+            defined: val => angular.isDefined(val) ? val : undefined,
+            // The string replacement is for removing the invisible "zero-width-space" UTF-8
+            // causing queries to fail: https://en.wikipedia.org/wiki/Zero-width_space
+            removeUtf8SpecialChars: val => val.replace(/%E2%80%8B/ig, '%20')
+        };
+        queryFilters.call = (val) => {
+            return queryFilters.removeUtf8SpecialChars(
+                queryFilters.defined(val));
+        };
+        return {
+            encode: val => queryFilters.call(val),
+            decode: val => queryFilters.call(val),
+            // We want to always match this type. If we match on slash, it won't update
+            // the `stateParams`.
+            is: () => true
+        };
     });
 
     $stateProvider.state('search', {
@@ -64,7 +77,7 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
     });
 
     $stateProvider.state('search.results', {
-        url: 'search?{query:Slashed}&ids&since&nonFree&uploadedBy&until&orderBy',
+        url: 'search?{query:Query}&ids&since&nonFree&uploadedBy&until&orderBy',
         // Non-URL parameters
         params: {
             // Routing-level property indicating whether the state has
--- a/kahuna/public/js/search/index.js
+++ b/kahuna/public/js/search/index.js
@@ -35,28 +35,15 @@ export var search = angular.module('kahuna.search', [
 search.config(['$stateProvider', '$urlMatcherFactoryProvider',
                function($stateProvider, $urlMatcherFactoryProvider) {
 
-    // TODO: Fix this
-    $urlMatcherFactoryProvider.type('Query', {}, () => {
-        var queryFilters = {
-            // Query params slashes are encoded that mess up our seach query when we search for
-            // e.g. `24/7`.
-            // see: https://github.com/angular-ui/ui-router/issues/1119#issuecomment-64696060
-            defined: val => angular.isDefined(val) ? val : undefined,
-            // The string replacement is for removing the invisible "zero-width-space" UTF-8
-            // causing queries to fail: https://en.wikipedia.org/wiki/Zero-width_space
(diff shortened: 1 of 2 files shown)


jamesgorrie:
We could do it [here](https://github.com/guardian/grid/blob/master/kahuna/public/js/search/index.js#L41-L47) to keep this controller clean?

NickPapacostas:
Ok so I realize that what i just pushed up might be substantial overkill, but i thought it might be more future-extensible or w.e 

If not, then I can either just add it to the existing matcher that was there (and maybe change it's name)
wanted to know what y'all thought

@jamesgorrie

kahuna/public/js/search/index.js
```diff
@@ -35,15 +35,28 @@ export var search = angular.module('kahuna.search', [
...
-    // see: https://github.com/angular-ui/ui-router/issues/1119#issuecomment-64696060
     // TODO: Fix this
-    $urlMatcherFactoryProvider.type('Slashed', {
-        encode: val => angular.isDefined(val) ? val : undefined,
-        decode: val => angular.isDefined(val) ? val : undefined,
-        // We want to always match this type. If we match on slash, it won't update
-        // the `stateParams`.
-        is: () => true
+    $urlMatcherFactoryProvider.type('Query', {}, () => {
+        var queryFilters = {
```
jamesgorrie: We try to use const where we can to avoid reassigning the var name.

kahuna/public/js/search/index.js
```diff
@@ -35,15 +35,28 @@ export var search = angular.module('kahuna.search', [
...
+            // Query params slashes are encoded that mess up our seach query when we search for
+            // e.g. `24/7`.
+            // see: https://github.com/angular-ui/ui-router/issues/1119#issuecomment-64696060
+            defined: val => angular.isDefined(val) ? val : undefined,
+            // The string replacement is for removing the invisible "zero-width-space" UTF-8
+            // causing queries to fail: https://en.wikipedia.org/wiki/Zero-width_space
+            removeUtf8SpecialChars: val => val.replace(/%E2%80%8B/ig, '%20')
+        };
+        queryFilters.call = (val) => {
```
jamesgorrie: we could just have this as a call function (to avoid mutating the obj). Not fussed, just a thought.

jamesgorrie:
:+1: definitely easier to extend later! Nice.

kahuna/public/js/search/index.js
NickPapacostas: good call. Still adapting :)

kahuna/public/js/search/index.js
```diff
@@ -35,15 +35,28 @@ export var search = angular.module('kahuna.search', [
...
-        is: () => true
+    $urlMatcherFactoryProvider.type('Query', {}, () => {
+        var queryFilters = {
+            // Query params slashes are encoded that mess up our seach query when we search for
+            // e.g. `24/7`.
+            // see: https://github.com/angular-ui/ui-router/issues/1119#issuecomment-64696060
+            defined: val => angular.isDefined(val) ? val : undefined,
+            // The string replacement is for removing the invisible "zero-width-space" UTF-8
+            // causing queries to fail: https://en.wikipedia.org/wiki/Zero-width_space
+            removeUtf8SpecialChars: val => val.replace(/%E2%80%8B/ig, '%20')
```
theefer: If `defined` returned `undefined`, does this risk to throw?

NickPapacostas:
@jamesgorrie we figured it out... kinda... basically "is" doesn't do what you think it does. If it looks good give me another thumbs up ^_^

theefer:
Oh we forgot to convert the replace to use a regexp with the m flag.

kahuna/public/js/search/index.js
```diff
@@ -35,15 +35,16 @@ export var search = angular.module('kahuna.search', [
...
+    const zeroWidthSpace = '\u200B';
+    function removeUtf8SpecialChars(val) {
+        return val && val.replace(zeroWidthSpace, '');
+    }
+
+    $urlMatcherFactoryProvider.type('Query', {
+        encode: val => removeUtf8SpecialChars(val),
+        decode: val => removeUtf8SpecialChars(val),
+        //call decode value that includes zero-width-space character
+        is: val => val && (val.indexOf(zeroWidthSpace) === -1)
```
jamesgorrie: What now happens to values that aren't defined?

jamesgorrie:
There seems to be a bug where, when you remove the last `char` from the query, it doesn't update the $stateParams. i.e. type `dog => backspace => backspace`. Now all you have is `d`. One more `backspace`, you still have `d` in the `qs`. I am assuming that's because we aren;t matching the `is` as `val`.

`is: angular.isDefined(val) && (val.indexOf(zeroWidthSpace) === -1)` seems to work as we aren't matching false on empty string then.

theefer:
Good catch @jamesgorrie. Good reminder to do a full manual regression of the affected feature (search), esp around corner cases like empty input etc, when fiddling with the underlying router.

NickPapacostas:
@jamesgorrie @theefer ty! Saw the bug, made the change, confirmed it worked :) 

I'll make sure to check thoroughly in the future, any chance there's a "critical path" kind of doc with stuff to check in a regression?

theefer:
Sadly no. My test is usually: search, upload an image, edit its metadata, crop it.

In this instance it'd probably just involve doing a few corner case searches.

kahuna/public/js/search/index.js
```diff
@@ -114,7 +115,7 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
                 ).
                     distinctUntilChanged(angular.identity, Immutable.is).
                     shareReplay(1);
-            }]
+            }],
```
theefer: Unnecessary (does it pass lint?)

kahuna/public/js/search/index.js
```diff
@@ -35,15 +35,16 @@ export var search = angular.module('kahuna.search', [
...
-    // TODO: Fix this
-    $urlMatcherFactoryProvider.type('Slashed', {
-        encode: val => angular.isDefined(val) ? val : undefined,
-        decode: val => angular.isDefined(val) ? val : undefined,
-        // We want to always match this type. If we match on slash, it won't update
-        // the `stateParams`.
-        is: () => true
+    const zeroWidthSpace = '\u200B';
+    function removeUtf8SpecialChars(val) {
+        return val && val.replace(zeroWidthSpace, '');
```
theefer: This only replaces the first one. Did we want to be safe and replace all zero-width space?

NickPapacostas:
ok cool ^_^ shall i take that as a thumbs up?

jamesgorrie:
Could we do a global replace as per @theefer's [comment](https://github.com/guardian/grid/pull/1539#discussion_r47078616)?

NickPapacostas:
@theefer @jamesgorrie Yep, sorry somehow missed those comments. Lint passed, regexp being used, manually tested with multiple special charz in there. Sorry for the sloppiness ^_^

jamesgorrie:
Not sloppy at all, this is the third / fourth time it's being done as it's not a very clear API.
This will hopefully be the one :clap: :+1:

theefer:
Great, don't forget to go and tell Mee-Lai!

PR #1540, opened 2015-12-07
add topic arn to collections

jamesgorrie:
I'll need this to publish to thrall.
~~I will add to prod CF too.~~ Done


akash1810:
:+1:

PR #1542, opened 2015-12-07
add update setting to elasticsearch

jamesgorrie:
Seems to be fine. While ingesting it seemed fine as it updated in a very small amount of time, and given that we retry thrall's messages, this seemed full-proofish?

But! Pausing thrall might be the things we do prior, update the settings and then unpause.

Downtime? I suppose so.


akash1810:
How would media-api react?

jamesgorrie:
No read, no write.

kenoir:
It's difficult to tell what the impact will be to the end user if this only takes a fraction of a second. My guess is that impact would be minimal. Given how onerous it is to perform a reindex (and the associated dangers of downtime) I'd say it's worth a try during a quiet period with comms sent out beforehand.

scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
```diff
@@ -143,6 +145,39 @@ object UpdateMapping extends EsScript {
...
+        indices.open(new OpenIndexRequest(imagesAlias))
+        client.close
+      }
+    }
+
+    EsClient.updateSettings
+  }
+
+  def usageError: Nothing = {
+    System.err.println("Usage: UpdateMapping <ES_HOST>")
```
theefer: Should it be `UpdateSettings` instead?

theefer:
Are we confident this is going to be fast even on the PROD index that is an order of magnitude largest than TEST?

theefer:
Could you describe what are we updating the settings to/for?

jamesgorrie:
> Are we confident this is going to be fast even on the PROD index that is an order of magnitude largest than TEST?

I need to do some more discovery - as far as I can tell from the docs close, update, and open are just one off request. The concern I have is how long it would take that change to the setting to propagate to make sure we can open again.

scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
jamesgorrie: :eyes:

jamesgorrie:
> Could you describe

This is a generic way to do it - do you mean in the PR?

theefer:
I guess I am curious what are we going to use this for (that we didn't need until now)?

PR looks :+1: apart from being scary :-)

jamesgorrie:
[We've added the hierarchy analyser](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala#L106-L111) to help search collections. We haven't been able to apply the mappings so far as we haven't updated the settings (including the analyser). So we need to update.

theefer:
OK :+1:

jamesgorrie:
Closing as the repercussions were to large in PROD.

PR #1543, opened 2015-12-08
add collections to mapping

jamesgorrie:
Waiting on #1542 to be able to use the new analyser.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
@@ -32,6 +32,8 @@ object Mappings {
 
   def withIndexName(indexName: String,  obj: JsObject) = Json.obj("index_Name" -> indexName) ++ obj
 
+  def copyTo(fieldName: String) = Json.obj("copy_to" -> fieldName)
+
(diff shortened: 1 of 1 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -105,7 +107,7 @@ object Mappings {
 
   val collectionMapping = withIndexName("collection", nonDynamicObj(
     "path" -> nonAnalysedList("collectionPath"),
-    "pathId" -> nonAnalyzedString,
+    "pathId" -> (nonAnalyzedString ++ copyTo("collections.pathHierarchy")),
```
theefer: Is it worth documenting what the `copyTo` is for (iow, I don't know what it is for :smile_cat:)?

jamesgorrie: [Done](https://github.com/guardian/grid/commit/9763f16dfb62125339dcc443f5687fb4a08a9145). Interestingly talking to @akash1810 and @NickPapacostas about using it for a mega free text search field.

theefer:
:+1:

PR #1544, opened 2015-12-08
Image collection+thrall notification

jamesgorrie:
Update ES record when collections are set.
We send over the whole collections set so no need to update.

Waiting on #1528 #1542 #1543.


thrall/app/lib/ElasticSearch.scala
```diff
@@ -131,6 +131,19 @@ object ElasticSearch extends ElasticSearchClient with ImageFields {
...
+    prepareImageUpdate(id)
+      .setScriptParams(Map(
+        "collections" -> asGroovy(collections),
+        "lastModified" -> asGroovy(JsString(currentIsoDateString))
+      ).asJava)
+      .setScript(
+        "ctx._source.collections = collections;" +
+          updateLastModifiedScript,
+        scriptType)
+      .executeAndLog(s"updating collections on image $id")
```
akash1810: minor: would logging `setting collections on image $id` be more accurate?

akash1810:
:+1:

PR #1546, opened 2015-12-09
Use image-loader import instead of download/upload in S3 watcher

theefer:
Rather than downloading and uploading the file via the Lambda, this uses the import endpoint of the loader to make the loader download the image directly from S3 using a time-limited URL.

As suggested by @kenoir.


s3watcher/lambda/lib/Logger.js
```diff
@@ -35,15 +34,11 @@ module.exports = {
     messages: messages,
 
     log: function (stage, messageKey, state) {
-        log(stage, messageKey, state)
+        log(stage, messageKey, state);
```
akash1810: :fearful: does this mean the linter didn't pick this up originally?!?!

s3watcher/lambda/lib/Transfer.js
```diff
@@ -41,27 +41,27 @@ module.exports = {
...
-                    filename: upload.params.filename,
-                    uploadedBy: upload.params.uploadedBy,
-                    stage: upload.params.stage
-                });
+            Logger.logImport(config.stage, s3Object);
 
-                return Upload.postData(upload, data.Body);
-            }).retry(5).flatMap(function(uploadResult){
-                Logger.logRecordToCloudWatch(config.stage, uploadResult);
+            const urlExpiryInSeconds = config.s3UrlExpiry;
```
akash1810: The config object (and file in S3) is changing; is it worth thinking about how we do this?

s3watcher/lambda/lib/Upload.js
```diff
@@ -20,10 +18,10 @@ module.exports = {
 
          const uploadedBy = buildUploadedBy(s3Event.path);
 
-         return  {
+         return {
              key: config.apiKey,
              url: config.baseUrl,
-             path: "/images",
+             path: "/imports",
              size: s3Event.size,
```
akash1810: is `size` applicable now?

akash1810:
couple of minor comments, but :+1:

s3watcher/lambda/lib/Logger.js
theefer: Does the linter run anywhere?

s3watcher/lambda/lib/Transfer.js
theefer: Absolutely, I forgot to mention it.

s3watcher/lambda/lib/Upload.js
theefer: I thought it might get logged somewhere, but I'm not sure..?

akash1810: I [does](https://github.com/guardian/grid/blob/master/s3watcher/lambda/lib/Transfer.js#L48), though this PR removes that.

I was thinking its more appropriate for image-loader to log this information now as it makes more sense in the context of that service.

Its minor though.

s3watcher/lambda/lib/Logger.js
akash1810: its not in the [ci script](https://github.com/guardian/grid/blob/master/s3watcher/build.sh)... oops! will raise a PR to fix.

s3watcher/lambda/lib/Transfer.js
kenoir: Perhaps we should standardise on a way of doing this for Lambdas? I prefer the include config at build time option.

kenoir:
Hopefully this will get rid of the errors we were seeing. 

Overall I prefer this approach as it makes the Lambda do even less work (pushing the possibility for error states out). I almost did this the first time round but decided against it as I didn't want to use the time limited URLs (in retrospect silly as the lambda is itself a time limited operation).

s3watcher/lambda/lib/Transfer.js
theefer: Standardising would be nice, but I'm not sure about the build-time option. Shall we chat around coffee?

theefer:
Come on, just admit it, you wanted to play with piping data through reactive streams ;-P

kenoir:
@theefer obvs

PR #1549, opened 2015-12-09
Automate DEV setup for s3watcher

theefer:
Add utility scripts and helpers to make it easy to develop on the s3watcher locally, without having to fudge things manually.

See the updated README for details.

Note: There is still an issue with having to add a flag to allow insecure TLS connections locally (the loader cert is not validating properly).


s3watcher/lambda/package.json
```diff
@@ -4,17 +4,20 @@
   "description": "",
   "main": "S3Watcher.js",
   "dependencies": {
+    "aws-sdk": "^2.2.21",
```
akash1810: should this be in dev deps? The aws sdk is already included in the lambda environment.

theefer: Ah OK that's why it was in dev deps. I'll remove it as it's now in the deps of the scripts package.json. Thanks!

theefer: Or maybe it's needed for local runs so will move back to dev deps..

kenoir:
:+1: nice

akash1810:
:+1:

Issue #1496, opened 2015-11-26
dot-properties generator doesn't generate correct usage.properties


akash1810:
dot-properties works by getting the outputs of a cloud formation stack and filling in templates with the values.

Currently in DEV there are 2 stacks: core and usage. However, the dot-properties script only looks up values in the core stack. This means the `usage.properties` file doesn't get completed properly as the values are from the usage stack.

~~Possible fixes would be:~~
~~- Rather than have two stacks for DEV, use nested stacks.~~
~~- Update dot-properties to lookup values in both stacks.~~

~~The former is better in the long term.~~

Better yet, [conditionals](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/conditions-section-structure.html).

A quick fix for now would be to add the output of [this script](https://github.com/guardian/grid/files/44998/usage.txt) to `/etc/gu/usage.properties`. (NB: file extension is `.txt` for github upload)

PR #1550, opened 2015-12-10
remove cf template for usage lambda.

akash1810:
we're not using a lambda anymore!

Doesn't fix, but certainly eliminates the need for https://github.com/guardian/grid/issues/1496.


theefer:
:+1:

PR #1552, opened 2015-12-14
crop image keyboard shortcut.

akash1810:
was playing around with this on Friday - `k` is for krop.

more of a conversation starter than wip...


kahuna/public/js/image/controller.js
```diff
@@ -54,10 +56,19 @@ image.controller('ImageCtrl', [
...
               cropKey,
               mediaCropper,
-              imageService) {
+              imageService,
+              hotkeys) {
 
         let ctrl = this;
 
+        hotkeys.bindTo($scope).add({
+            combo: 'k',
```
theefer: Why not `c`?

kahuna/public/js/crop/controller.js
```diff
@@ -1,19 +1,29 @@
...
 
     const ctrl = this;
     const body = $document[0].body;
 
-    var imageId = $stateParams.imageId;
+    const imageId = $stateParams.imageId;
+
+     hotkeys.bindTo($scope).add({
+         combo: 'esc',
+         description: 'Crop image',
```
theefer: Wrong description?

theefer:
I was hoping to abstract shortcuts away from the controllers a bit more, maybe wire with Rx, but tbh this looks fairly simple and usable. Would it be worth rewriting the ENTER shortcut on the crop screen to use the same method?

kahuna/public/js/crop/controller.js
akash1810: yep, copy/paste error.

kahuna/public/js/image/controller.js
akash1810: Was following Light Room, but they use `r` so I don't actually know where `k` came from...

akash1810:
> Would it be worth rewriting the ENTER shortcut on the crop screen to use the same method?

Done.

paperboyo:
Cool! More shortcuts, the better! But by slowly adding individual ones we may end up in a sorry state. Properly IMHO, we should first evaluate (and free up) all currently used up keyboard shortcuts on Editorial machines (there are some legacy/crazy ones blocked by Fast Scripts), then make a list of all shortcuts already there in _all_ new tools, decide which ones we would like to keep consistent _between_ all tools (like panels opening/hiding etc.). And only then it would be safe to introduce individual ones on the per-tool basis.

Oh, and watch for https://medium.com/medium-eng/the-curious-case-of-disappearing-polish-s-fa398313d4df#.nmqdn0fc8 - it’s already broken in Composer :)

jamesgorrie:
Could we not use `main.js` to store the deps? It stops us being able to get rid of that at some stage, and isn't really global.

Also - why `k` not `c` - there's a few interesting pros to mnemonic shortcuts, but `k` doesn't seem like one. I know [spacemacs loves them](https://github.com/syl20bnr/spacemacs/blob/master/doc/DOCUMENTATION.org#mnemonic).

akash1810:
> Could we not use `main.js` to store the deps?

Done.

kahuna/public/js/crop/controller.js
```diff
@@ -1,19 +1,31 @@
...
+    hotkeys.bindTo($scope)
+        .add({
+            combo: 'esc',
+            description: 'Cancel crop and return to image',
+            callback: () => $state.go('image', {imageId: ctrl.image.data.id})
+        })
+        .add({
+            combo: 'enter',
+            description: 'Create crop',
+            callback: () => ctrl.callCrop()
```
jamesgorrie: `callback: ctrl.callCrop`?

jamesgorrie:
:+1:

kahuna/public/js/crop/controller.js
```diff
@@ -102,18 +114,4 @@ crop.controller('ImageCropCtrl',
...
-             ctrl.callCrop();
-         }
-     }
-
-     //TODO find a nicer way to handle keyboard shortcuts
-     body.addEventListener('keypress', cropReturnKeyShortcut);
-
-     $scope.$on('$destroy', function(){
-         body.removeEventListener('keypress', cropReturnKeyShortcut);
-     });
```
theefer: Nice to remove this!

theefer:
Cool :+1: 

The next step could be to think of surfacing these so users can discover them, maybe as part of the tooltip on the buttons to do the same actions? Can a separate PR of course.

akash1810:
> The next step could be to think of surfacing these so users can discover them

You get this for free with this library by pressing `?`. Though it doesn't work atm, so will look at it.

paperboyo:
> so users can discover them

Researching potential conflicts with browsers/OSes ([1](https://en.wikipedia.org/wiki/Table_of_keyboard_shortcuts), [2](http://dmcritchie.mvps.org/firefox/keyboard.htm)) and general thinking behind [shortcuts in web apps](https://bugzilla.mozilla.org/show_bug.cgi?id=380637), I've learned that you get the overlay of them for Gmail when tapping `?`. Neat.

theefer:
> You get this for free with this library by pressing ?

One does not prevent the other, both seem desirable eventually?

akash1810:
> make a list of all shortcuts already there in all new tools, decide which ones we would like to keep consistent between all tools (like panels opening/hiding etc.)

Would like to compile this list of shortcuts that can and should be consistent throughout the tools. Will talk with @alastairjardine tomorrow.

paperboyo:
:-) Do we need to worry about OS/browser shortcuts? Can we/do we want to/should we override them (some/all)? I will try to compile a list of FastScripts shrotcuts on Editorial machines (I’m sure we can free up some legacy ones...).

jamesgorrie:
Most webapps now just use a single letter e.g. a,b,c,d - the only problem is, when you're in text-making mode e.g. in a search box, you need to not listen to these (gmail can sometimes catch you when you think you're typing, but you're actually deleting all your mail).

The pages we have shortcuts on don't require any typing, so this seems good.

akash1810:
> Though it doesn't work atm, so will look at it.

Fixed with https://github.com/jspm/registry/pull/683.

PR #1554, opened 2015-12-15
Make full-frame thumb draggable and selectable from modal

theefer:
I noticed that it was not possible to drag the full-frame thumbnail, or click on it to select it from the modal Grid. This fixes both.


jamesgorrie:
:+1:

PR #1555, opened 2015-12-15
Revert "Revert "add collections to image response""

jamesgorrie:
Reverts guardian/grid#1551

We now have the `collections` in the mappings.
This will store all new images to
```
{
  "image": {
    "collections": []
  }
}
```

But will also fall back to older images.


theefer:
:+1:

PR #1556, opened 2015-12-15
AspectRatio calculations for full frame crops.

NickPapacostas:
First pass, tests don't pass, I need to make it less strict.

just looking for feedback on the approach in general cus i'm a scala scrub.

Main concerns in comments

### DIFF
--- a/cropper/app/lib/AspectRatio.scala
+++ b/cropper/app/lib/AspectRatio.scala
@@ -1,0 +1,21 @@
+package lib
+
+object AspectRatio {
+  case class Ratio(friendly: Option[String], width: Int, height: Int)
+
+  def gcd(a: Int, b: Int): Int = if (b == 0) a else gcd(b, a % b)
+
+  def calculate(width: Int, height: Int) : Option[Ratio] = {
+    val knownRatios = List(
+      Ratio(Some("5:3"), 5, 3),
+      Ratio(Some("2:3"), 2, 3),
+      Ratio(Some("16:9"), 16, 9)
+    )
+    val rGcd = gcd(width, height)
+    val simplifiedWidth  = width / rGcd
+    val simplifiedHeight = height / rGcd
+    println(s"for ($width, $height) sHeight: $simplifiedHeight, sWidth: $simplifiedWidth where GCD was $rGcd")
+
+    knownRatios.find(ratio => ratio.width == simplifiedWidth && ratio.height == simplifiedHeight)
+ }
+}
--- a/cropper/app/lib/AspectRatio.scala
+++ b/cropper/app/lib/AspectRatio.scala
@@ -15,14 +15,14 @@ object AspectRatio {
   def gcd(a: Int, b: Int): Int = if (b == 0) a else gcd(b, a % b)
 
   def calculate(width: Int, height: Int, tolerance: Int = 3) : Option[Ratio] = {
-    val matchingRatios = for {
+    val matchingRatio = for {
       w <- width - tolerance until width + tolerance
       h <- height - tolerance until height + tolerance
       g = gcd(w, h)
       simplifiedWidth = w / g
       simplifiedHeight = h / g
-      ratio <- knownRatios if ratio.width == simplifiedWidth &&  ratio.height == simplifiedHeight
+      ratio <- knownRatios.find(ratio => ratio.width == simplifiedWidth &&  ratio.height == simplifiedHeight)
     } yield ratio
-    matchingRatios.headOption
+    matchingRatio.headOption
   }
 }
--- a/cropper/app/lib/AspectRatio.scala
+++ b/cropper/app/lib/AspectRatio.scala
@@ -1,21 +1,27 @@
 package lib
 
-object AspectRatio {
-  case class Ratio(friendly: Option[String], width: Int, height: Int)
+import scala.annotation.tailrec
 
+object AspectRatio {
+  case class Ratio(friendly: String, width: Int, height: Int)
+
+  @tailrec
   def gcd(a: Int, b: Int): Int = if (b == 0) a else gcd(b, a % b)
 
-  def calculate(width: Int, height: Int) : Option[Ratio] = {
+  def calculate(width: Int, height: Int, tolerance: Int = 3) : Option[Ratio] = {
     val knownRatios = List(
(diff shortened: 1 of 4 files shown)


cropper/app/lib/AspectRatio.scala
```diff
@@ -0,0 +1,21 @@
+package lib
+
+object AspectRatio {
+  case class Ratio(friendly: Option[String], width: Int, height: Int)
+
+  def gcd(a: Int, b: Int): Int = if (b == 0) a else gcd(b, a % b)
```
NickPapacostas: Write your own gcd? seems not good but didn't find a compelling alternative

cropper/app/lib/AspectRatio.scala
```diff
@@ -0,0 +1,21 @@
...
+  def calculate(width: Int, height: Int) : Option[Ratio] = {
+    val knownRatios = List(
+      Ratio(Some("5:3"), 5, 3),
+      Ratio(Some("2:3"), 2, 3),
+      Ratio(Some("16:9"), 16, 9)
+    )
+    val rGcd = gcd(width, height)
+    val simplifiedWidth  = width / rGcd
+    val simplifiedHeight = height / rGcd
+    println(s"for ($width, $height) sHeight: $simplifiedHeight, sWidth: $simplifiedWidth where GCD was $rGcd")
```
NickPapacostas: will remove

cropper/app/model/ExportRequest.scala
```diff
@@ -38,7 +39,13 @@ object ExportRequest {
   def boundsFill(dimensions: Dimensions): Bounds = Bounds(0, 0, dimensions.width, dimensions.height)
 
   def toCropSpec(cropRequest: ExportRequest, dimensions: Dimensions): CropSpec = cropRequest match {
-    case FullExportRequest(uri)          => CropSpec(uri, boundsFill(dimensions), None, FullExport)
+    case FullExportRequest(uri)          =>
+      CropSpec(
+        uri,
+        boundsFill(dimensions),
+        AspectRatio.calculate(dimensions.width, dimensions.height).flatMap(_.friendly),
```
NickPapacostas: Thought about making calculate take Dimensions ( which are just width: Int and height: Int ) but it seemed like too much. maybe i just need to embrace typing ^_^

cropper/app/lib/AspectRatio.scala
```diff
@@ -0,0 +1,21 @@
...
+  case class Ratio(friendly: Option[String], width: Int, height: Int)
+
+  def gcd(a: Int, b: Int): Int = if (b == 0) a else gcd(b, a % b)
+
+  def calculate(width: Int, height: Int) : Option[Ratio] = {
+    val knownRatios = List(
+      Ratio(Some("5:3"), 5, 3),
+      Ratio(Some("2:3"), 2, 3),
+      Ratio(Some("16:9"), 16, 9)
+    )
```
theefer: You could move that `val` outside the function so it doesn't get allocated every time, but just once.

akash1810:
nice :+1:

cropper/app/lib/AspectRatio.scala
```diff
@@ -0,0 +1,28 @@
...
+  val knownRatios = List(
+    Ratio("5:3", 5, 3),
+    Ratio("2:3", 2, 3),
+    Ratio("16:9", 16, 9)
+  )
+
+  @tailrec
+  def gcd(a: Int, b: Int): Int = if (b == 0) a else gcd(b, a % b)
+
+  def calculate(width: Int, height: Int, tolerance: Int = 3) : Option[Ratio] = {
```
jamesgorrie: Could tolerance be a percent as if the images get's smaller 3 becomes bigger.

cropper/app/lib/AspectRatio.scala
```diff
@@ -0,0 +1,28 @@
...
+    val matchingRatios = for {
+      w <- width - tolerance until width + tolerance
+      h <- height - tolerance until height + tolerance
+      g = gcd(w, h)
+      simplifiedWidth = w / g
+      simplifiedHeight = h / g
+      ratio <- knownRatios if ratio.width == simplifiedWidth &&  ratio.height == simplifiedHeight
+    } yield ratio
+    matchingRatios.headOption
```
jamesgorrie: Would we ever have more than one match? We could make this a `find` or something similar if not?

cropper/app/lib/AspectRatio.scala
```diff
@@ -0,0 +1,28 @@
+package lib
+
+import scala.annotation.tailrec
+
+object AspectRatio {
+  case class Ratio(friendly: String, width: Int, height: Int)
+
+  val knownRatios = List(
+    Ratio("5:3", 5, 3),
```
jamesgorrie: You could make friendly automatic, but we might potentially want to add things like "megasuperlandslide".

jamesgorrie:
A couple of comments, but minor opinions, not blockers.

cropper/app/lib/AspectRatio.scala
NickPapacostas: yea i thought about that a bit, the problem I saw was that the accuracy of the cropping itself has the same level of looseness at all sizes (from what i could tell) and so as the images get smaller and smaller, the tolerance will be less and less pixels which doesn't really represent what's going on. 

I wasn't sure though, I talked about it briefly with @theefer and we decided to let it ride for now. I'm torn on it though. it should probably be a percentage with a minimum or something like that.

cropper/app/lib/AspectRatio.scala
NickPapacostas: I don't think more than one match would make a lot of sense. I had it as a find earlier but headOption seems to do the same job. I'll change it to find though because it just seems nicer ^_^

cropper/app/lib/AspectRatio.scala
jamesgorrie: Leave a note and someone, some day will get to it.

cropper/app/lib/AspectRatio.scala
jamesgorrie: find seems more expressive as to what we are actually trying to achieve.

PR #1557, opened 2015-12-15
Drag to collection ui

jamesgorrie:
Simplish way of adding images to a collection. 

__TODO:__
* remove from collection
* search by collection
* show image collections on image

add-to-coll 1


kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -53,13 +54,56 @@ grCollectionsPanel.controller('GrNodeCtrl', ['collections', function(collections
...
+    ctrl.addImagesToCollection = imagesJson => {
+        const imageIds = imagesJson.map(imageJson => imageJson.id);
+        const promises = imageIds.map(id => mediaApi.find(id).then(image =>
+            image.perform('add-collection', {body: {data: ctrl.collectionPath}})));
+
+        return Promise.all(promises);
+    };
+}]);
+
+grCollectionsPanel.directive('grAddToCollection',
```
theefer: Should the name explicit "drop" to make it more obvious how it's adding?

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -53,13 +54,56 @@ grCollectionsPanel.controller('GrNodeCtrl', ['collections', function(collections
...
+        controller: 'GrAddToCollectionCtrl',
+        controllerAs: 'ctrl',
+        bindToController: true,
+        scope: {
+            collectionPath: '=grAddToCollection'
+        },
+        link: function(scope, element, _, ctrl) {
+            element.on('drop', jqEv => {
+                const ev = jqEv.originalEvent;
+                const el = element[0];
```
theefer: Note that you should be able to use jQlite helpers on `element` to add/remove CSS classes instead?

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -53,13 +54,56 @@ grCollectionsPanel.controller('GrNodeCtrl', ['collections', function(collections
 
 }]);
 
+grCollectionsPanel.controller('GrAddToCollectionCtrl',
+    ['mediaApi', '$scope', function(mediaApi, $scope) {
+
+    const ctrl = this;
+
+    ctrl.saved = false;
```
theefer: Is this used? Is there any feedback while it's adding?

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -53,13 +54,56 @@ grCollectionsPanel.controller('GrNodeCtrl', ['collections', function(collections
...
+
+    ctrl.addImagesToCollection = imagesJson => {
+        const imageIds = imagesJson.map(imageJson => imageJson.id);
+        const promises = imageIds.map(id => mediaApi.find(id).then(image =>
```
theefer: Is this looking up from the API again, rather than creating resources from the in-memory object? Not a big deal, just curious.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -53,13 +54,56 @@ grCollectionsPanel.controller('GrNodeCtrl', ['collections', function(collections
...
+            image.perform('add-collection', {body: {data: ctrl.collectionPath}})));
```
theefer: Could we move this to the collection service just to abstract out of the nitty gritty directives?

theefer:
Cool!

theefer:
Also TODO: somehow show what collections an image is in in the results? This served as a nice action feedback when adding labels too.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
jamesgorrie: Nope.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
jamesgorrie: To create the resource from the string JSON that we get from the drop data is quite tricky - I will run you through it. Happy to change it later.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
jamesgorrie: Done

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
jamesgorrie: But we should have some, I will work with the state a little more once I can figure out how. I might scrap the `ctrl` all together.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
jamesgorrie: I prefer the DOM way, but happy to change

kahuna/public/js/services/api/collections-api.js
```diff
@@ -42,12 +42,20 @@ apiServices.factory('collections', ['mediaApi', function (mediaApi) {
         });
     }
 
+    function addImagesToCollection(imageIds, collectionPath) {
+        const promises = imageIds.map(id => mediaApi.find(id).then(
+                image => image.perform('add-collection', {body: {data: collectionPath}})));
```
theefer: Happy to go with that, maybe we could put a note in saying it's inefficient (talks to API when we have all the information) and why we can't do the right thing yet?

theefer:
:+1:

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
jamesgorrie: Yip, done.

jamesgorrie:
I've shifted this to be completely in the directive following the guidelines of attribute directives shouldn't have controllers, and element directives should.

theefer:
OK, I don't have strong opinions on this, although curious if there's a rationale for that difference?

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -53,22 +54,47 @@ grCollectionsPanel.controller('GrNodeCtrl', ['collections', function(collections
...
+            const collectionPath = $parse(attrs.grDropIntoCollection)(scope);
+
+            element.on('drop', jqEv => {
+                const ev = jqEv.originalEvent;
+                const gridImagesData = ev.dataTransfer.getData(vndMimeTypes.get('gridImagesData'));
+                const gridImageData = ev.dataTransfer.getData(vndMimeTypes.get('gridImageData'));
+
+                if (gridImagesData !== '' && gridImageData !== '') {
+                    // TODO: potentially add some UI feedback on adding to collection
+                    const imagesData = gridImagesData !== "" ?
```
akash1810: does the linter complain about `""`?

jamesgorrie:
The rationale is around scope. As this is already in a directive (with controller), keeping it to an attribute directive, with no controller or own scope stops screwing up knowing what scope you're in.

It's easier to show in the code if that doesn't make sense.

PR #1560, opened 2015-12-16
update jspm

jamesgorrie:
We've cached a dirty version (0.16.17) in TC that was fixed in 18 that took our `baseUrl` into consideration in the build (thus trying to include `file:///assets/js/main.js`). 

This forces the minor updates.


akash1810:
:+1:

PR #1562, opened 2015-12-16
add collections to query syntax

jamesgorrie:
Using `~` is by no means to final case, it just allows me to demo the collections more fully by adding something and then selecting the collection.

As this is programmatic it's not a tie in.


theefer:
:+1:

PR #1564, opened 2015-12-16
use gr-icon-label

akash1810:
Looks like we missed one. To reproduce:
- Find an image that cannot be cropped
- Visit https://media.domain.co.uk/images/IMAGE_ID/crop

Before:
screen shot 2015-12-16 at 18 03 00

After:
screen shot 2015-12-16 at 18 03 14


jamesgorrie:
:+1:

PR #1565, opened 2015-12-16
crop shortcuts

akash1810:
Yay shortcuts!

Side-effect - this allows for the key combo `c` `f` `enter` for a full-frame crop.


jamesgorrie:
:+1: worth mentioning this to power users - @JonnyWeeks etc.

PR #1566, opened 2015-12-17
Panels

jamesgorrie:
Instigate by introducing the collections panel.
With more than one panel it's becoming hard to keep track of what's going on as classes are changed in a few places, which doubles.

This should hopefully give us a `panelled-content` view, which can be controlled with a service, allowing for some nice stuff including state saving.


alastairjardine:
Are you removing the click a second time to lock the panel? Why?

jamesgorrie:
@alastairjardine  - Nope - nothing will change, just the way it's constructed in the view, all should be business as usual - unless there's something panel-wise you'd like to add / remove / make better while I'm here.

Apologies, I should have mentioned it in the PR.

paperboyo:
> unless there's something panel-wise you'd like to add / remove / make better

Me? Yeah:
> removing the click a second time to lock the panel

Hahaha.

But on a serious note (the above is super-serious, too, but there may be some arguments for second click...), I would definitely keep close and lock buttons on the left of the LHS panel (as they are in the GIF), but would **move** close and lock buttons in the RHS panel **to the right**. Rationale is: if there has to be this extra click to lock at least it won't be miles away to be even more annoying. And open/close should not even require moving your mouse (one should show over the other).

Great stuff! Keyboard shortcuts!

jamesgorrie:
@paperboyo re: close buttons: noted, good rationale.

To reiterate though - the gif is there just to show the functionality exists.
Actually it's no longer there as it so far from what the UI is it's unhelpful.

alastairjardine:
@

alastairjardine:
Button click fail...ignore

jamesgorrie:
You can delete the comments if they're not useful?

alastairjardine:
@jamesgorrie Was just explaining myself for the accidental close! Feel free to delete :)

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.html
```diff
@@ -1,5 +1,4 @@
-<div class="panel panel__left panel--side"
-     ng:class="{ 'panel__left--hidden': !ctrl.isVisible }">
+<div>
```
theefer: Are you removing the `panel` class on purpose? Seems to leave the other `panel__` classes below dangling without a parent.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -15,23 +15,26 @@ grCollectionsPanel.controller('GrCollectionsPanelCtrl', [
     'panelService',
     '$rootScope',
     'collections',
-    function (panelService, $rootScope, collections) {
+    function (panelService, $rootScope, collections, userPrefs) {
```
theefer: Not getting injected?

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -15,23 +15,26 @@ grCollectionsPanel.controller('GrCollectionsPanelCtrl', [
...
+    //panelService.available(panelName, true);
+    //
+    //$rootScope.$on(
+    //    `ui:panels:${panelName}:updated`,
+    //    () => {
+    //        ctrl.isVisible = panelService.isVisible(panelName);
+    //    }
+    //);
+
+
```
theefer: Remove if not used?

kahuna/public/js/components/gr-panel/gr-panel.html
```diff
@@ -1,7 +1,4 @@
-<div class="panel panel--side panel__right image-details"
-     ng:mouseover="ctrl.metadataPanelMouseOver()"
-     ng:mouseleave="ctrl.metadataPanelMouseLeave()"
-     ng:class="{ 'panel__right--hidden': !ctrl.isVisible }">
+<div class="image-details">
```
theefer: Same comment as above re removal of `panel` class.

kahuna/public/js/user-prefs/service.js
```diff
@@ -0,0 +1,11 @@
+import angular from 'angular';
+import
```
theefer: Even if not used, probably should be valid syntax?

kahuna/public/js/components/gr-top-bar/gr-top-bar.css
```diff
@@ -2,7 +2,7 @@ gr-top-bar {
     display: block;
     height: 50px; /* faking we're in the flow */
     position: relative;
-    z-index: 3;
+    z-index: 30;
```
theefer: Any rationale for this change? Will need to check how that fits with other overlays etc.

kahuna/public/js/search/results.html
```diff
@@ -104,21 +104,9 @@
...
-                        <gr-icon-label gr-icon="lock_open">Lock info panel</gr-icon-label>
-                    </span>
-                    <span ng:if="ctrl.metadataPanelVisible && ctrl.metadataPanelLocked">
-                        <gr-icon-label gr-icon="lock">Unlock info panel</gr-icon-label>
-                    </span>
-                    <span ng:if="!ctrl.metadataPanelVisible">
-                        <gr-icon-label gr-icon="chrome_reader_mode">Show info panel</gr-icon-label>
-                    </span>
-                </a>
+                <gr-panel-button class="results-toolbar-item results-toolbar-item--right results-toolbar-item__metadata-panel-controls"
```
theefer: Confusing that the same element is both `results-toolbar-item` and child BEM element `results-toolbar-item__metadata-panel-controls`, though it looks like it was already the case so nevermind me.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.html
jamesgorrie: It's another refactoring with the class names because of the image page.
I was hoping to do that separately (i.e. move the image page over to use the same panelled view).

kahuna/public/js/search/results.js
```diff
@@ -11,6 +11,7 @@ import '../util/seq';
 import '../components/gu-lazy-table/gu-lazy-table';
 import '../downloader/downloader';
 import '../components/gr-delete-image/gr-delete-image';
+import '../components/gr-panel-button/gr-panel-button';
```
theefer: I can't find that file?

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
jamesgorrie: Yip, there's a bit of cleanup.

kahuna/public/js/user-prefs/service.js
jamesgorrie: Strange, that wasn't picked up - this service is ghost code anyway.

kahuna/public/js/components/gr-top-bar/gr-top-bar.css
jamesgorrie: I was thinking of having layers, the same way you do with tabindexes - makes it easier to add things.

I was thinking you could have:
* 1 -> everything
* 10 -> things to be just above everything
* 20 -> user controls i.e. panels
* 30 -> the top layer i.e. topbar.

kahuna/public/js/search/results.html
jamesgorrie: It is indeed the same as what it was, and given the length and complexity of the class names, I'd prefer to leave it.

jamesgorrie:
@theefer - re: `.panel` classNames - perhaps I should just refactor the whole thing out - to be honest, it would just be a lot of copy paste.

kahuna/public/js/components/gr-top-bar/gr-top-bar.css
theefer: Absolutely, sounds like the best approach. But in absence of a layer reference, we will need to just check manually that id doesn't interfere with other overlays (dnd, auth error, advanced search, etc)

kahuna/public/js/search/results.js
jamesgorrie: My bad, forgot to add it to git.

theefer:
Would be nice to get a run through of the architectural changes.

jamesgorrie:
> architectural

Strong word for CSS. Makes sense - it'll help with the cleanup that should be substantial as I was hoping this would get rid of some of the complexities.

theefer:
I meant mostly the JS to be honest.

kahuna/public/js/components/gr-panel-button/gr-panel-button.js
```diff
@@ -0,0 +1,33 @@
...
+
+panelButton.directive('grPanelButton', [function() {
+    return {
+        restrict: 'E',
+        template: template,
+        bindToController: true,
+        controller: 'GrPanelButton',
+        controllerAs: 'ctrl',
+        scope: {
+            panel$: '=grPanel'
```
theefer: Is this stream expected to emit different panels over time, or just one?

jamesgorrie: Multiple.
It's a way of avoiding a client calling `.getPanel` if it hasn't been created somewhere.

[I've added a note here](https://github.com/guardian/grid/pull/1566/files#diff-7965c78e2f73afa5a99c770c34b37dedR8).

There were two other options (with their problems):
* A promise, although I didn't know what the fail would be.
* Have them as resolves in the view - but that felt like we were coupling the panels to something not really related i.e. the panel controller wasn't really controlling itself.

jamesgorrie:
> I meant mostly the JS to be honest.

If that's the case, there wouldn't be much change at all - it's mostly `HTML` / `CSS` as the panel is actually baked into the page (we could use this as a case to component it out). 

I say leave it for another PR?

jamesgorrie:
[I've added the collections panel](https://github.com/guardian/grid/commit/adc9132a0a9cfb98789987ea50f2e78a1db3dbc2) - if that shows more, architecturally, how this works?

jamesgorrie:
@theefer [moved the panels to be a dependancy on the view](https://github.com/guardian/grid/commit/4a36302f987796721a2b19be8aed343f816752be#diff-04b33f0ec2971bf016d5d80534267132R71).

Also add lock the button (working on the hover css now)
lock-panel-com

kahuna/public/js/components/gr-panel-button/gr-panel-button.js
```diff
@@ -0,0 +1,37 @@
+import angular from 'angular';
+import template from './gr-panel-button.html!text';
+import '../../util/rx';
+
+export const panelButton = angular.module('gr.panelButton', ['util.rx']);
+
+panelButton.controller('GrPanelButton', ['$scope', 'inject$', 'subscribe$', function($scope, inject$, subscribe$) {
```
theefer: `subscribe$` no longer used I think?

kahuna/public/js/components/gr-panel/gr-panel.html
```diff
@@ -1,16 +1,8 @@
-<div class="panel panel--side panel__right image-details"
-     ng:mouseover="ctrl.metadataPanelMouseOver()"
-     ng:mouseleave="ctrl.metadataPanelMouseLeave()"
-     ng:class="{ 'panel__right--hidden': !ctrl.isVisible }">
+<div class="image-details panel§">
```
theefer: `panel§`? :-)

kahuna/public/js/components/gr-panels/gr-panels.css
```diff
@@ -0,0 +1,52 @@
...
+}
+
+
+/* Stupid hack as we haven't decoupled the search toolbar and search results */
+.search-panels .gr-panel {
+    display: block;
+    margin-top: 36px;
+}
+
+/*
```
theefer: Unclosed comment?

kahuna/public/js/components/gr-top-bar/gr-top-bar.css
theefer: Were you able to check the `z-index` of other overlays?

kahuna/public/js/search/index.js
```diff
@@ -65,6 +68,20 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
                 });
                 return {state: $dsr$.redirect.state, params};
             }]
+        },
+        controllerAs: 'ctrl',
+        controller: ['panels', function(panels) {
+            const ctrl = this;
+            ctrl.collectionsPanel = panels.collectionsPanel;
+            ctrl.metadataPanel = panels.metadataPanel;
+        }],
```
theefer: Are these needed on top of the similar code in `results.js`? I think they are actually, just checking I understand well.

kahuna/public/js/search/index.js
```diff
@@ -65,6 +68,20 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
...
+        resolve: {
+            panels: ['panelService', function(panelService) {
+               const collectionsPanel = panelService.createPanel({ hidden: true });
+               const metadataPanel = panelService.createPanel({ hidden: true });
+
+               return { collectionsPanel, metadataPanel };
+           }]
```
theefer: Feels tider imho, nice :+1:

theefer:
Just a question open re z-index, else looks :+1:

kahuna/public/js/components/gr-top-bar/gr-top-bar.css
jamesgorrie: Updated a few, I'll make a note to go through and adjust all of them.
For now it works.

kahuna/public/js/components/gr-panels/gr-panels.css
jamesgorrie: Gone

kahuna/public/js/search/index.js
jamesgorrie: [Yip](https://github.com/guardian/grid/pull/1566/files#diff-b8c6d4770b5b04294cc8c88d27ae65d3R78)

theefer:
Cool. Would be good to test with a few users before rolling out. Not sure we want to push such a change tomorrow so close to Christmas, so maybe first thing in January?

theefer:
I'm getting a double scrollbar:

image

theefer:
Not hiding the panel means it stays open and hides a whole column's worth of images, possibly in such a way that the user doesn't realise. Also makes it less clear what's the difference between opened and locked.

Do we want to have a chat about how to get this out quickly without breaking more than we fix?

paperboyo:
> Not hiding the panel means it stays open and hides a whole column's worth of images, possibly in such a way that the user doesn't realise.

Very true also currently. One solution would be for the panel to have very slightly reduced opacity. Better solution would be for the panels to never (or at least initially) be the width equal to a column of images.

theefer:
> Better solution would be for the panels to never (or at least initially) be the width equal to a column of images.

Too fiddly, as the width adapt to the width of the screen.

theefer:
> Very true also currently.

Really? It only overlays images when in "show" mode, which requires that your mouse is over the panel. Moving your mouse out of the panel hides it, revealing the hidden images again. The only way for the panel to stick is to Lock it, which reflows images.

Are you saying you sometimes get into a state where the panel is hiding images persistently?

paperboyo:
All true - it's less of a problem now. But it's there. Whenever a panel covers exactly one RHS column of images it's a problem. But I can see panel hiding on mouse-out in show mode on the above gif as well (maybe a little later)?

> Too fiddly, as the width adapt to the width of the screen.

Yeah :cry:

jamesgorrie:
I've added the scroll => hide back to not change to many moving parts from the original.
This is to avoid making it better for some and worse for others, we'll leave it the same.

theefer:
:+1:

PR #1567, opened 2015-12-21
tooltips via titip

akash1810:
Uses http://quantumui.org/appdoc/documents/home/titip.

Initially for keyboard shortcut prompts. Can be used to add rich hover state to images on search page too.

titip

### DIFF
--- a/kahuna/package.json
+++ b/kahuna/package.json
@@ -50,6 +50,16 @@
             ]
           }
         }
+      },
+      "github:quantumui/titip@397fd324d5b5cc98ee332f876a100a70f69b273e": {
+        "main": "dist/css/titip.css!",
+        "shim": {
+          "titip": {
+            "deps": [
+              "css"
+            ]
+          }
+        }
       }
     }
   },
--- a/kahuna/package.json
+++ b/kahuna/package.json
@@ -32,7 +32,7 @@
       "text": "github:systemjs/plugin-text@^0.0.2",
       "theseus": "npm:theseus@0.5.2",
       "theseus-angular": "npm:theseus-angular@^0.3.1",
-      "titip": "github:quantumui/titip@master",
+      "titip": "github:quantumui/titip@397fd324d5b5cc98ee332f876a100a70f69b273e",
       "ua-parser-js": "npm:ua-parser-js@0.7.3",
       "uri-templates": "npm:uri-templates@0.1.5"
     },
@@ -47,16 +47,6 @@
             "deps": [
               "moment",
               "./css/pikaday.css!"
-            ]
-          }
-        }
-      },
-      "github:quantumui/titip@master": {
-        "main": "dist/css/titip.css!",
-        "shim": {
-          "titip": {
-            "deps": [
-              "css"
             ]
           }
         }
--- a/kahuna/package.json
+++ b/kahuna/package.json
@@ -32,7 +32,7 @@
       "text": "github:systemjs/plugin-text@^0.0.2",
       "theseus": "npm:theseus@0.5.2",
       "theseus-angular": "npm:theseus-angular@^0.3.1",
-      "titip": "github:akash1810/titip@1.0",
+      "titip": "github:guardian/titip@1.0.0",
       "ua-parser-js": "npm:ua-parser-js@0.7.3",
       "uri-templates": "npm:uri-templates@0.1.5"
     },
--- a/kahuna/package.json
+++ b/kahuna/package.json
@@ -32,7 +32,7 @@
       "text": "github:systemjs/plugin-text@^0.0.2",
       "theseus": "npm:theseus@0.5.2",
       "theseus-angular": "npm:theseus-angular@^0.3.1",
-      "titip": "github:quantumui/titip@397fd324d5b5cc98ee332f876a100a70f69b273e",
+      "titip": "github:akash1810/titip@1.0",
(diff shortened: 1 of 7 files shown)


kahuna/package.json
```diff
@@ -49,6 +50,16 @@
             ]
           }
         }
+      },
+      "github:quantumui/titip@master": {
```
akash1810: temporary, waiting on https://github.com/jspm/registry/pull/689

jamesgorrie:
Are we risk of not integrating any other Quantum UI bits so creating (more) disparity in the UI?
Or is it customisable enough that we could keep the functionality, but extend it's look?

Just asking as we could try to integrate a lib with more functional bits that we know we can use e.g. Paper.

akash1810:
> Or is it customisable enough that we could keep the functionality, but extend it's look?

Yep - its a css only lib, which was the main reason to use it.

kahuna/package.json
```diff
@@ -32,6 +32,7 @@
       "text": "github:systemjs/plugin-text@^0.0.2",
       "theseus": "npm:theseus@0.5.2",
       "theseus-angular": "npm:theseus-angular@^0.3.1",
+      "titip": "github:quantumui/titip@master",
```
theefer: Depending on `master` isn't ideal because it makes the builds non-reproduceable, if master is updated. Could you open an issue on that project to request them to tag versions? Alternatively, is it possible to use Git hashes?

kahuna/public/js/crop/view.html
```diff
@@ -8,22 +8,22 @@
     <gr-top-bar-actions>
         <div class="top-bar-item side-padded">{{ ctrl.cropWidth() }} &times; {{ ctrl.cropHeight() }}</div>
 
-        <label class="top-bar-item clickable side-padded" title="landscape">
+        <label class="top-bar-item clickable side-padded titip-default titip-bottom" data-title="landscape [l]">
```
theefer: I'm curious if these should be capital letter to avoid confusion?

theefer:
:+1:

kahuna/package.json
akash1810: > Could you open an issue on that project to request them to tag versions?

https://github.com/quantumui/titip/issues/1

> is it possible to use Git hashes?

Yes it is. What would you advise for [jspm's registry](https://github.com/jspm/registry/pull/689)?

kahuna/public/js/crop/view.html
akash1810: Is that `landscape` -> `Landscape` or `[l]` -> `[L]`?

The latter implies needing to press `shift` to use the shortcut, which isn't correct.

kahuna/package.json
theefer: I've merged your registry PR. I don't think versions have anything to do with the registry, but we should probably reference a Git hash until they setup a version we can reference in the upstream repo.

kahuna/package.json
theefer: Merged that PR.

kahuna/public/js/crop/view.html
theefer: The latter. Does it work with `shift` or not?

AFAICS in interfaces I use, shortcut hints tend to be capitalised, e.g. `Ctrl+C`, without implying having to press `shift`.

kahuna/package.json
akash1810: Thanks. Updated with commit hash.

kahuna/public/js/crop/view.html
akash1810: > Does it work with `shift` or not?

Nope.

theefer: OK. Still think people will get it and it would avoid confusion between `1`, `l`, `i`, `|`, etc.

paperboyo: +1 for capitals

jamesgorrie: Interesting as I am used to `c` vs `C` implying shift.
This is probably not the norm though.

kahuna/package.json
```diff
@@ -49,6 +50,16 @@
             ]
           }
         }
+      },
+      "github:quantumui/titip@397fd324d5b5cc98ee332f876a100a70f69b273e": {
```
akash1810: Ugh, the titip lib only has a master release, which as @theefer pointed out, shouldn't be used as it may not re-produce reliable results.

TC fails w/out this override - could put this override in the [registry repo](https://github.com/jspm/registry/tree/master/package-overrides/github/quantumui) but it feels a bit wrong to use a commit hash version there...

kahuna/public/js/crop/view.html
akash1810: I've seen shortcuts listed in both upper (soundcloud) and lower (twitter). Happy to convert to upper and get feedback from users.

kahuna/package.json
theefer: It does, I guess we can just go for that. Worse case we can fork their repo and create a tag release if they can't pull their fingers out to push a git tag themselves.

akash1810: Forked it to my account as I don't have perms for organisation account. I've also [removed](https://github.com/quantumui/titip/pull/2) the default cursor styling.

akash1810:
Now using forked repo where we've tagged a release.

Merging.

Issue #331, opened 2015-01-07
Make label search case insensitive


jamesgorrie:
[Works](https://media.gutools.co.uk/search?query=EW:%20Eagle%20Nebula)
[Doesn't work](https://media.gutools.co.uk/search?query=EW:%20Eagle%20nebula)

PR #1568, opened 2015-12-21
Copy to sstemmer catchall

NickPapacostas:
using copy_to to create a field where we can add non analyzed (or non-standard analyzed) fields so that we can search them normally (using the standard english_s_stemmer).


NickPapacostas:
note: this requires reindexing

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -75,7 +75,8 @@ object Mappings {
     "subLocation" -> standardAnalysedString,
     "city" -> standardAnalysedString,
     "state" -> standardAnalysedString,
-    "country" -> standardAnalysedString
+    "country" -> standardAnalysedString,
+    "sStemmerCatchAll" -> sStemmerAnalysedString
```
jamesgorrie: Not sure sStemmer is the right name as it's having a few things happen to it. 
[There's a comment here about it](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/IndexSettings.scala#L14).

Could be called `englishAnalysedCatchAll`?

NickPapacostas: yup absolutely

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -60,11 +60,11 @@ object Mappings {
 
   val metadataMapping = nonDynamicObj(
     "dateTaken" -> dateFormat,
-    "description" -> sStemmerAnalysedString,
-    "byline" -> standardAnalysedString,
+    "description" -> (sStemmerAnalysedString ++ copyTo("metadata.englishAnalysedCatchAll")),
+    "byline" -> (standardAnalysedString ++ copyTo("metadata.englishAnalysedCatchAll")),
```
akash1810: How are updates handled? That is, when `byline` is updated, would `copyTo` blindly append to `englishAnalysedCatchAll` or will it replace the current byline, or recreate the entire field?

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -75,7 +75,8 @@ object Mappings {
     "subLocation" -> standardAnalysedString,
     "city" -> standardAnalysedString,
     "state" -> standardAnalysedString,
-    "country" -> standardAnalysedString
+    "country" -> standardAnalysedString,
+    "englishAnalysedCatchAll" -> sStemmerAnalysedString
```
theefer: Do we need to add any other fields to this, such as `city`, `title`, etc? Worth comparing the list with the "any field" matcher in the syntax.

NickPapacostas: @jamesgorrie @akash1810 any thoughts on this? Are there other fields you've heard of having trouble due to mismatched field analyzer / query analyzer?

If not I wonder if it's better to just have another field for byline ("bylineEnglishAnalyzed or something)?

jamesgorrie: I'm inclined to say bosh most of it in there. 
[Here's what we have at the moment](https://github.com/guardian/grid/blob/cb97663639890dd21ff940154494ea8795850194/media-api/app/lib/elasticsearch/ElasticSearch.scala#L67-L69).

I see we put the labels in there - which I think would be good too, we'd just need to make sure that happens when we [update the labels (userMetadata)](https://github.com/guardian/grid/blob/cb97663639890dd21ff940154494ea8795850194/thrall/app/lib/ElasticSearch.scala#L120).

NickPapacostas: good call, if we get labels in there it'll fix this too (i'm pretty sure >.<): https://github.com/guardian/grid/issues/331

NickPapacostas:
something I just want to note down here (as it's impacted my dev cycle) is that the "copy_to" solution only stores this new field in the inverted index (what's used for searching) and it is not possible to view what the contents of that inverted index are, so it's hard to get insight into what is actually in that field (other than looking at the results of your searches)

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -61,21 +61,22 @@ object Mappings {
   val metadataMapping = nonDynamicObj(
     "dateTaken" -> dateFormat,
     "description" -> sStemmerAnalysedString,
```
jamesgorrie: I suppose the question now is whether we want to analyse these other fields similar to the other fields so we could do cross field searching. Not really sure which side I fall on, worth asking though.

akash1810:
> it is not possible to view what the contents of that inverted index are

:cry: There might be a plugin available if desperately needed?

theefer:
Was there a conclusion on whether a single field with everything in is better or worse than having individual english-analysed versions of the fields we care about?

jamesgorrie:
I think the question was around when we added a new field e.g. `shortDescription` which would be easier. I think the conclusion to that was that both would require a reindex.

The only advantage to storing duplicated, differently analysed fields would be that we can choose programatically which ones to search. This could be useful to refine the way we free search.

NickPapacostas:
tested and working on TEST

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -120,7 +121,7 @@ object Mappings {
   val userMetadataMapping =
     nonDynamicObj(
       "archived"    -> boolean,
-      "labels"      -> nonAnalysedList("label"),
+      "labels"      -> (nonAnalysedList("label") ++ copyTo("metadata.englishAnalysedCatchAll")),
```
theefer: Do we really want to add label to this?

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -61,22 +61,23 @@ object Mappings {
   val metadataMapping = nonDynamicObj(
     "dateTaken" -> dateFormat,
     "description" -> sStemmerAnalysedString,
-    "byline" -> standardAnalysedString,
+    "byline" -> (standardAnalysedString ++ copyTo("metadata.englishAnalysedCatchAll")),
     "bylineTitle" -> standardAnalysedString,
     "title" -> sStemmerAnalysedString,
```
theefer: I think we should also add the title?

theefer: Ah ignore me, it's already the right analyser..

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -61,22 +61,23 @@ object Mappings {
...
-    "credit" -> nonAnalyzedString,
+    "credit" ->  (nonAnalyzedString ++ copyTo("metadata.englishAnalysedCatchAll")),
     "creditUri" -> nonAnalyzedString,
     "copyright" -> standardAnalysedString,
```
theefer: Should we add copyright too?

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -61,22 +61,23 @@ object Mappings {
...
     "copyrightNotice" -> standardAnalysedString,
-    "suppliersReference" -> standardAnalysedString,
+    "suppliersReference" -> (standardAnalysedString ++ copyTo("metadata.englishAnalysedCatchAll")),
+    "source" -> (nonAnalyzedString ++ copyTo("metadata.englishAnalysedCatchAll")),
     "subjects" -> nonAnalysedList("subject"),
-    "source" -> nonAnalyzedString,
     "specialInstructions" -> nonAnalyzedString,
```
theefer: Should we add specialInstructions too?

PR #1569, opened 2015-12-21
Removes usage warning and tweaks UI

kenoir:
Changes:
- Adds usage count to tab label
- Label is greyed and non clickable when no usages available

With usages:
usages

Without usages (tab is not clickable):
usage

### DIFF
--- a/kahuna/public/js/image/view.html
+++ b/kahuna/public/js/image/view.html
@@ -173,7 +173,7 @@
                    gr:track-click-data="{ tab: 'Usage' }">
                 <div class="radio-list__selection-state"></div>
                 <div class="radio-list__label-value">
-                    Usages ({{ctrl.usagesCount || "None"}})
+                    Usages ({{ctrl.usagesCount == 0 ? "None" : ctrl.usagesCount}})
                 </div>
             </label>
         </div>
--- a/kahuna/public/js/image/view.html
+++ b/kahuna/public/js/image/view.html
@@ -158,12 +158,19 @@
                    id="show-usage"
                    class="radio-list__circle"
                    name="show-usage"/>
+
             <label for="show-usage"
                    class="radio-list__label"
-                   ng:class="{'radio-list--selected': ! showMetadata}"
-                   ng:click="showMetadata = false">
+                   ng:class="
+                   {
+                    'radio-list--disabled': ctrl.usagesCount < 1,
+                    'radio-list--selected': ! showMetadata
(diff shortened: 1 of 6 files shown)


kenoir:
Pinging @alastairjardine.

alastairjardine:
Might be worth making it clickable, and putting in an informative message when there is zero usage on the resulting tab view.

It probably needs something to explain why it's not clickable otherwise.

theefer:
:+1: 

and +1 to what @alastairjardine said.

kenoir:
@alastairjardine Previously the tab said "None" when it was clickable. 

My thinking was that there should be some obvious differentiation on the image page as to whether an image has usages or not - greying and disabling the usage tab was my solution.

kenoir:
Updated wording when no usages are available. Tab still remains unavailable as I think we can convey "no usages" without having to force the user to: click and see usages empty and then click again to return to metadata view.

kahuna/public/js/image/view.html
```diff
@@ -160,14 +160,21 @@
...
+                    'radio-list--disabled': ctrl.usagesCount < 1,
+                    'radio-list--selected': ! showMetadata
+                   }"
+                   ng:click="(ctrl.usagesCount < 1) || (showMetadata = false)"
                    gr:track-click="Image panel tab"
                    gr:track-click-data="{ tab: 'Usage' }">
                 <div class="radio-list__selection-state"></div>
-                <div class="radio-list__label-value">Usage</div>
+                <div class="radio-list__label-value">
+                    Usages ({{ctrl.usagesCount || "None"}})
```
theefer: I personally favour [not using integers as booleans](https://github.com/guardian/flexible-content/wiki/JavaScript-Style-Guide#syntax) for clarity. Could use a ternary operator instead if you agree, but happy for others to chip in on this coding style.

theefer:
Minor comment on coding style, else :+1:

kahuna/public/js/image/view.html
kenoir: A ternary is more descriptive and not significantly more code. Have replaced.

akash1810:
> without having to force the user to: click and see usages empty and then click again to return to metadata view.

Could we not render the usage tab at all then when there's no usage?

kenoir:
@akash1810 I think that might not send the signal strongly enough that there are no usages, it might just make folks wonder where it's gone.

PR #1570, opened 2015-12-21
Add tracking to image tabs

jamesgorrie:
Probably worth knowing - also good for if we want to turn things off.


akash1810:
:+1:

theefer:
:+1:

PR #1571, opened 2015-12-21
Graceful fail for usage where image isn't found

kenoir:
Performs a get request before attempting an upsert
Fails gracefully when no document found to update
i.e. message is deleted, no error logged

### DIFF
--- a/thrall/app/lib/ElasticSearch.scala
+++ b/thrall/app/lib/ElasticSearch.scala
@@ -18,6 +18,7 @@ import org.joda.time.DateTime
 import scala.collection.convert.decorateAll._
 import scala.concurrent.{ExecutionContext, Future}
 
+import play.api.Logger
 
 object ImageNotDeletable extends Throwable("Image cannot be deleted")
 
@@ -92,9 +93,11 @@ object ElasticSearch extends ElasticSearchClient with ImageFields {
       }
   }
 
-  def updateImageUsages(id: String, usages: JsValue)(implicit ex: ExecutionContext): Future[UpdateResponse] =
-    prepareImageUpdate(id)
-      .setScriptParams(Map(
+  case class GracefulFail(msg: String)
+
+  def updateImageUsages(id: String, usages: JsValue)(implicit ex: ExecutionContext): Future[Any] =
+    prepareUpdateIfImageExists(id).map(requestBuilder => {
+      requestBuilder.setScriptParams(Map(
         "usages" -> asGroovy(usages),
         "lastModified" -> asGroovy(JsString(currentIsoDateString))
       ).asJava)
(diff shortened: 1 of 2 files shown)


usage/app/controllers/UsageApi.scala
```diff
@@ -8,7 +8,6 @@ import scala.util.Try
 
 import play.api.libs.json.JsError
 import play.api.Logger
-import play.api.Logger
```
akash1810: double import? could remove the additional `import play.api.mvc.Results._` too?

akash1810:
:+1:

thrall/app/lib/ElasticSearch.scala
```diff
@@ -103,6 +103,7 @@ object ElasticSearch extends ElasticSearchClient with ImageFields {
           updateLastModifiedScript,
         scriptType)
       .executeAndLog(s"updating usages on image $id")
+      .recover { case e: DocumentMissingException => {}}
```
jamesgorrie: We could - to be descriptive of having side effects, have a return of `Future[Unit]` and then match on `UpdateResponse` to return `Unit`.

Or potentially match on `UpdateResponse`and recover on `DocumentMissingException ` to return a new type of `EsActionComplete`?

jamesgorrie: Completely sidelined - just wondering.

theefer: +1 to returning unit (`()`).

theefer: We could log the info here just for sanity of debugging.

theefer:
:+1:

thrall/app/lib/ElasticSearch.scala
kenoir: @theefer `.executeAndLog(s"updating usages on image $id")` does log a descriptive error when it fails already which is why I removed the previous log statement from here as I felt it was redundant.

@jamesgorrie going to release this as is. I'd prefer not to return `Future[Any]` though so maybe you could explain your solution in person when you're around?

usage/app/controllers/UsageApi.scala
kenoir: Weird. Maybe I just needed to be really really sure.

Have removed.

thrall/app/lib/ElasticSearch.scala
jamesgorrie:
```scala
case class EsActionComplete(action: String)
//...
 def updateImageUsages(id: String, usages: JsValue)(implicit ex: ExecutionContext): Future[EsActionComplete] =
//...
.executeAndLog(s"updating usages on image $id")
.map(_ => EsActionComplete("deleted-image"))
.recover { case e: DocumentMissingException => EsActionComplete("deleted-image-not-found") }
.incrementOnFailure(failedUsagesUpdates) { case e: VersionConflictEngineException => true }
```

PR #1572, opened 2015-12-22
Use Tumbling window to allow observable completion

kenoir:
Replaces previous exception throwing and catching as I observed event
loss.

Local testing indicates this also resolves the memory leak issue (though
to be sure it will have to soak in PROD).

giphy

### DIFF
--- a/usage/app/lib/UsageRecorder.scala
+++ b/usage/app/lib/UsageRecorder.scala
@@ -21,16 +21,9 @@ case class ResetException() extends Exception
 object UsageRecorder {
   val usageSubject  = PublishSubject[UsageGroup]()
   val usageStream   = UsageStream.observable.merge(usageSubject)
-  val resetInterval = 30.seconds
 
   val subscriber = Subscriber((_:Any) => Logger.debug(s"Sent Usage Notification"))
   def subscribe  = UsageRecorder.observable.subscribe(subscriber)
-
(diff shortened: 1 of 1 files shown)


theefer:
Still confused why this is necessary, but if it works.. Is this going to cause bit spikes of updates being sent to thrall? Do we know how many messages accumulate in 30s?

Presumably this means a server can get killed while it's accumulating messages in its window. Will the new servers spun up catch, record and send the missed usages?

kenoir:
@theefer I think what's going on is that as the stream never completes it makes it hard for the JVM to garbage collect on memory used by events in that stream. 

It'll send one message per media id changed every 30 seconds, (though this should not change from the current impl which effectively "forgets" which media ids have changed every 30 seconds). 

The `Tumbling` window doesn't mean that events get batch processed every 30 seconds, just that they are bundled into 30 second chunks the events on which get processed whenever they are scheduled.

I think it is possible to lose events between deploys, though as we unsubscribe from the event stream when the app is stopped this is better than otherwise.

My feeling is that the most sensible way of preventing event loss is to separate the CAPI poller and the usage recorder which is beyond the scope of this PR. Other mechanisms would involve synchronising around some external datastore or only recording a CAPI document change after we've written corresponding usages (which would __not__ play nice with the polling approach).

theefer:
OK :+1: 

I don't think I fully understand the behaviour when a new server starts up, whether it records (overrides) all usages for N recent content items. Might be easier to just have a quick chat about it :-)

PR #1573, opened 2015-12-22
directive that wraps the titip library

akash1810:
Should have noticed/appreciated the repeated markup in https://github.com/guardian/grid/pull/1567. 

means we can:
- easily swap out titip if needed
- easily migrate to titip with global find/replace `title` -> `gr-title`


kahuna/public/js/main.js
```diff
@@ -69,7 +69,8 @@ var kahuna = angular.module('kahuna', [
 
     // directives used throughout
     'gr.imageFadeOnLoad',
-    'grIcon'
+    'grIcon',
+    'grTitle'
```
akash1810: in similar fashion to `grIcon`, global import to make it super easy to use.

kahuna/public/js/components/gr-title/gr-title.js
```diff
@@ -0,0 +1,20 @@
+import angular from 'angular';
+import 'titip';
+
+export const module = angular.module('grTitle', []);
+
+module.directive('grTitle', [function () {
```
theefer: Would "tooltip" be a little more obvious than "title"?

kahuna/public/js/components/gr-title/gr-title.js
```diff
@@ -0,0 +1,20 @@
...
+    return {
+        restrict: 'A',
+        link: function (scope, element, attrs) {
+            const theme = attrs.grTitleTheme || 'default';
+            const position = attrs.grTitlePosition || 'bottom';
```
theefer: If these aren't used, can we just hardcode the defaults to reduce the API surface?

kahuna/public/js/components/gr-title/gr-title.js
```diff
@@ -0,0 +1,20 @@
...
+
+            element.attr('data-title', attrs.grTitle)
+                .addClass(`titip-${theme}`)
+                .addClass(`titip-${position}`);
+
+            element.removeAttr('title');
```
theefer: Odd?

kahuna/public/js/components/gr-title/gr-title.js
jamesgorrie: :+1: As you mentioned that we can use these for more than just snazzy titles?

kahuna/public/js/components/gr-title/gr-title.js
jamesgorrie: Assuming so you don't get the title popup too. Could add `aria-label`if we're being pedantic.

theefer: In this case, it would be better to encourage/enforce not having both this directive and a title on the same element? Could noisily fail (throw or log error) if that were the case so we can fix the template?

kahuna/public/js/components/gr-title/gr-title.js
akash1810: > Would "tooltip" be a little more obvious than "title"?

I did that initially, but opted for grTitle as its essentially a custom styled `title` attribute.

> can use these for more than just snazzy titles

Yep! Was planning on creating a `grRichTitle` directive, which, thinking about it, isn't the best name, so maybe `tooltip` is best!

theefer: I always found the `title` attribute name confusing, so I might be biased against it :-)

kahuna/public/js/components/gr-title/gr-title.js
jamesgorrie: Yip. Less is better.

kahuna/public/js/components/gr-title/gr-title.js
akash1810: `position` is used [here](https://github.com/guardian/grid/pull/1573/files#diff-f858b9ead9e730dc77b95579832a991aR38)

Was actually debating hardcoding `theme` - will do so!

theefer: OK!

kahuna/public/js/components/gr-title/gr-title.js
akash1810: > Assuming so you don't get the title popup too. Could add `aria-label` if we're being pedantic.

Exactly.

> Could noisily fail (throw or log error) if that were the case so we can fix the template?

:fearful:

kahuna/public/js/components/gr-title/gr-title.js
akash1810: renamed to `grTooltip`

jamesgorrie:
:+1:

theefer:
:+1:

PR #1574, opened 2015-12-22
Complete request stream after response received

theefer:
Minor fix for purity, doesn't actually address or solve the timeout issue.


jamesgorrie:
:+1:

PR #1575, opened 2015-12-22
allow watching of tooltip text

akash1810:
Tooltips are hard... Fixes:

screen shot 2015-12-22 at 21 27 51

### DIFF
--- a/kahuna/public/js/components/gr-tooltip/gr-tooltip.js
+++ b/kahuna/public/js/components/gr-tooltip/gr-tooltip.js
@@ -13,7 +13,7 @@ module.directive('grTooltip', ['onValChange', function (onValChange) {
                 .addClass(`titip-default`)
                 .addClass(`titip-${position}`);
 
-            const autoUpdates = angular.isDefined(attrs.grTooltipUpdates) || false;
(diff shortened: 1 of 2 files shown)


kahuna/public/js/components/gr-tooltip/gr-tooltip.js
```diff
@@ -3,15 +3,23 @@ import 'titip';
...
         restrict: 'A',
-        link: function (scope, element, attrs) {
+        link: function ($scope, element, attrs) {
             const position = attrs.grTooltipPosition || 'bottom';
 
             element.attr('data-title', attrs.grTooltip)
                 .addClass(`titip-default`)
                 .addClass(`titip-${position}`);
+
+            const autoUpdates = angular.isDefined(attrs.grTooltipUpdates) || false;
```
theefer: `someBool || false` is surely equivalent to just `someBool`?

theefer:
:+1:

kahuna/public/js/components/gr-tooltip/gr-tooltip.js
akash1810: yep

PR #1576, opened 2015-12-23
Collections panel ui

jamesgorrie:
Waiting on #1566.
[See the smaller diff here](https://github.com/guardian/grid/compare/panels...collections-panel-ui?expand=1)

Basically as it was.
Two major things were added:
* An edit mode toggle
* A "add images to this collection" button, avoiding dnd.

collection-panel-ui


tsop14:
Is that instead of drag and drop, or as well?

theefer:
I don't think I've ever seen a checkbox used as a modal control. Would a button (maybe toggling btw Edit/Done) make more sense, ideally also reusing the "pen" edit icon we have elsewhere?

kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
```diff
@@ -19,31 +19,47 @@
         </button>
 
         <a class="node__name flex-spacer"
-           ui:sref="search.results({query: (node.data.name)})">
+           ui:sref="search.results({query: '~'+node.data.content.pathId})">
```
theefer: Will that work if it contains a space?

theefer:
Same question as @tsop14 too.

paperboyo:
+1 to @tsop14 question

+1 to what @theefer said. Right click/long-press/long-tap context menu (I know - hard)? As it is currently I would also use the pen icon - but **only** for consistency (although I hate it and consider it an early - and ugly - workaround for a problem that manifests itself here again). The problem with the pen icon showing on hover is that it will join an add to collection icon that is already there, together with a bin icon on click (where is the rename icon?), another too-similar (plus) add new node icon that is confusingly contextual to a node (not clear what level a new node will appear in) and all the other zylions of icons that we will have there in the end...

theefer:
To clarify, my suggestion is just to replace the `[ ] Edit` control with an always-visible `[pen-icon] Edit` button, which possibly toggles into a `[pen-icon] Done` button. I'm not talking about hover round pen icon buttons alongside each collection.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
```diff
@@ -19,31 +19,47 @@
...
+                ng:click="ctrl.addImagesToCollection()"
+                gr:tooltip="Add selected images to this collection">
+            <gr-icon>add_to_photos</gr-icon>
         </button>
+        <span ng:if="ctrl.saving">Saving…</span>
+    </div>
+
+    <form class="node__add-child" ng:show="active" ng:submit="ctrl.addChild(childName).then(clearForm);">
+        <input class="node__add-child-input" type="text" required ng:model="childName"
+               placeholder="olympics, pantaloons, summer2015…" />
```
tsop14: Is this placeholder a bit confusing? On labels atm you can submit 'x, y, z' to create 3 labels 'x', 'y', 'z' but on collections you would create one new collection called 'x, y, z' .

jamesgorrie:
> Is that instead of drag and drop, or as well?

Both

jamesgorrie:
Agreed with @theefer 
edit-collections-button

theefer:
Nice

paperboyo:
:) Shouldn't it say "Finish editing" instead of "Finished editing"? :pedant:

jamesgorrie:
Both actually work.
One is in the first person, spoke to oneself. i.e. "I am finished editing :bouquet:" abbreviated to "Finished editing".
"Finish editing" is a command, i.e. "Browser - I have completed my task, now finish my editing or :facepunch:".

As it is, I think it sounds friendlier.

paperboyo:
:rabbit:

kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
jamesgorrie: [Added it to query-syntax](https://github.com/guardian/grid/commit/1cf1b463a9a69213c55b870dd62b16137d06a737)

kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
jamesgorrie: Good point, I've just simplified it to a single example and prefixed it with e.g.

jamesgorrie:
A few extra additions - but ready for PR

kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
```diff
@@ -19,31 +19,48 @@
...
-           ui:sref="search.results({query: (node.data.name)})">
+           ui:sref="search.results({query: ctrl.getCollectionQuery(node.data.content.pathId)})">
             {{node.data.name}}
         </a>
 
+
         <button class="node__action clickable" type="button"
                 ng:if="ctrl.deletable"
-                ng:click="ctrl.remove()">
+                ng:show="ctrl.editing"
```
theefer: Bit odd to have both `ng:if` and `ng:show`. Should they just be unified into the `ng:if`?

kahuna/public/js/components/gr-panel-button/gr-panel-button-small.html
```diff
@@ -0,0 +1,29 @@
+<div class="fill-height flex-container"
+     ng:class="{ 'flex-container--rtl': ctrl.position == 'right' }">
+    <button class="button-ico-square clickable" type="button"
+            gr:tooltip="Hide {{ctrl.name}} panel"
+            ng:if="!ctrl.state.hidden" ng:click="ctrl.hidePanel()"
```
theefer: Can we stick to one attribute per line? otherwise it's easy to miss and harder to read.

kahuna/public/js/search-query/query-syntax.js
```diff
@@ -17,3 +18,7 @@ export function removeLabel(q, label) {
 export function removeLabels(q, labels) {
     return labels.reduce((q, curr) => removeLabel(q, curr.name), q);
 }
+
+export function     getCollection(path) {
```
theefer: Spaaaaaaaces!

kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
jamesgorrie: The logic was that if it is a show hide mechanism it should be `ng:show` i.e. show hiding children. If it was logic around whether something should exist we should use `ng:if` i.e. the delete button.

It's also to get around the fact the `ng:if` creates it's own scope.

kahuna/public/js/search-query/query-syntax.js
jamesgorrie: Space = readability no?

kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
akash1810: >  If it was logic around whether something should exist we should use ng:if i.e. the delete button.

What about [ngDisabled](https://docs.angularjs.org/api/ng/directive/ngDisabled)?

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.html
```diff
@@ -1,9 +1,24 @@
...
     <div class="panel__content">
         <div class="error" ng:if="ctrl.error">
             There was an error loading the collections.
             Please try again later.
         </div>
-        <gr-nodes class="tree tree--top" gr:nodes="ctrl.collections"></gr-nodes>
+
+        <!-- TODO: we'll probably move this out to another template once we have something more
+        final -->
+        <div class="flex-container collections-panel-heading text-small">
```
akash1810: do we want to hide this if there is an error? ie `ng:if=! ctrl.error`

kahuna/public/js/components/gr-panel-button/gr-panel-button.js
```diff
@@ -29,7 +30,24 @@ panelButton.directive('grPanelButton', [function() {
...
             panel: '=grPanel',
             position: '@grPosition',
-            name: '@grName'
+            name: '@grName',
+            icon: '@grIcon'
+        }
+    };
+}]);
+
+panelButton.directive('grPanelButtonSmall', [function() {
```
akash1810: What's the difference between `grPanelButtonSmall` and `grPanelButton`? From what I can tell the small one doesn't have button labels?

kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
jamesgorrie: @akash1810 that's to disable inputs? I think we're talking about showing / hiding content.

kahuna/public/js/components/gr-panel-button/gr-panel-button-small.html
jamesgorrie: Done

kahuna/public/js/search-query/query-syntax.js
jamesgorrie: Done

kahuna/public/js/components/gr-panel-button/gr-panel-button.js
jamesgorrie: Yes, that's it. It seems better than convoluting a template with loads of `ng:if`s.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.html
jamesgorrie: Nice catch, thanks.

kahuna/public/js/components/gr-panel-button/gr-panel-button.js
akash1810: Could it be done with css then? (though not sure if that'll be clearer or not...)

jamesgorrie: We could ng:click="addClassToMakeCssWork" - but why not keep it in the template?

kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
theefer: Not sure about ngDisabled either, since the editing mode is transforming the UI rather than enabling functionality that we otherwise want present but disabled.

Re if vs show, I don't think it's a big deal, but given how infrequently people will toggle to editing mode, it doesn't necessarily feel worth rendering extra DOM elements (and directives) for every node visible in the possibly large tree at all times. Up to you though.

theefer:
Only the minor discussion on if/show left for me, but generally :+1: :exclamation:

kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
jamesgorrie: Agreed - changed.

PR #1577, opened 2015-12-23
upload lambda artifact during deploy

akash1810:
- When a lambda is created in cloudformation, we tell it where to find its code in S3. Let this be V1.
- When we run a riffraff deploy of the lambda, we update-function-code, which updates the code the lambda uses, but not the S3 object. Let this be V2.
- Any subsequent cloudformation updates will revert the lambda to run off V1 :cry: 

This fixes that as we keep the S3 object up-to-date.

TODO:
- [x] the upload artifact deploy type puts the file here: `/<bucketName>/<targetStage>/<packageName>/<filePathAndName>` this is different from our current path in cloudformation, so we need to update the path in our template.

### DIFF
--- a/s3watcher/deploy.json
+++ b/s3watcher/deploy.json
@@ -11,12 +11,22 @@
                     "PROD": {
                         "name": "@PROD_ARN"
                     }
-                }
+                },
+                "bucket": "media-service-dist"
             }
         }
     },
     "recipes": {
         "default": {
+            "depends": [
+                "artifactUploadOnly",
+                "updateLambda"
+            ]
+        },
+        "artifactUploadOnly": {
+            "actions": ["lambda.uploadArtifacts"]
+        },
+        "updateLambda": {
             "actions": ["lambda.updateLambda"]
         }
     }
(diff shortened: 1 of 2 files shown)


theefer:
:+1: if tested

gidsg:
@akash1810 re: `/<bucketName>/<targetStage>/<packageName>/<filePathAndName>` you may control this path using the Boolean options provided by Riffraff e.g. `prefixStack` https://riffraff.gutools.co.uk/docs/magenta-lib/types#awss3

akash1810:
> you may control this path using the Boolean options provided by Riffraff e.g. `prefixStack`

Ah nice!

Our cf currently uses the path `/<bucketName>/<targetStage>/<filePathAndName>`. The default path riffraff uses is a little nicer/clearer.

akash1810:
Updated, tested and working.

s3watcher/deploy.json
```diff
@@ -13,11 +13,19 @@
                     }
                 }
             }
+        },
+        "s3-watcher": {
+            "type": "aws-s3",
+            "data": {
+                "bucket": "media-service-dist",
+                "cacheControl": "public, max-age=60",
+                "publicReadAcl": false
```
akash1810: Although secrets are obtained from the bucket that triggers the lambda, setting `publicReadAcl` to false as paranoid...

PR #1580, opened 2015-12-24
tooltips on label cancel and label save

akash1810:
Quite a big diff! The main reason is extracting a new class `overflow-hidden` so that the tooltips can show.

screen shot 2015-12-24 at 10 22 18


theefer:
Should we decide whether we want to use such tooltips on every single control? It is really better/necessary, esp when there is no keyboard shortcut to explain?

jamesgorrie:
Agreed - icons are created to be succinct and understandable. If we need a tooltip, we probably need another icon.

akash1810:
> If we need a tooltip, we probably need another icon.

Very true!

Closing.

Issue #1342, opened 2015-10-10
Illustrators lists need Publication selector


paperboyo:
Hello,

Noticed a problem when setting rights to [Observer’s David Simonds](https://github.com/guardian/grid/blob/c6687f01d6b873a3a45a2c41e78eddf9bdd4438b/common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala#L129) today.

After correctly setting the rights and reloading the page (because of https://github.com/guardian/grid/pull/1341), the Credit field was populated with the name of the illustrator and the byline field was empty which, apart from being incorrect, makes the image invalid.

What should happen instead is that the name of the illustrator should go into Byline field and the Credit should be populated with the relevant Publication similarly to how Photographers work.

Regards
Mateusz

PR #1581, opened 2015-12-24
Use illustrator as byline too, not just credit

theefer:
Fixes https://github.com/guardian/grid/issues/1342

Currently, when selecting Illustrator as Usage Rights, the `creator` name gets copied to the `credit` field only. In practice it should also populate the `byline` field, especially now that Composer won't duplicate them in its output.

cc @jamesgorrie I don't think there was any reason not to do this (anymore?) ?


theefer:
Merry :christmas_tree: @paperboyo!

:santa: :snowman: :gift:

jamesgorrie:
:+1:

PR #1583, opened 2015-12-24
Usage CAPI Poller with retry

kenoir:
Adds retrying to capi poller with exponential backoff

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/rx/SingleThreadedScheduler.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/rx/SingleThreadedScheduler.scala
@@ -1,0 +1,13 @@
+package com.gu.mediaservice.lib.rx
+
+import rx.lang.scala.schedulers.ExecutionContextScheduler
+import java.util.concurrent.Executors
+import scala.concurrent.ExecutionContext
+
+
+trait SingleThreadedScheduler {
+  private val singleThreadedExecutor = Executors.newSingleThreadExecutor()
+  val scheduler = ExecutionContextScheduler(ExecutionContext.fromExecutor(singleThreadedExecutor))
+}
+
+
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -29,7 +29,7 @@ object Build extends Build {
   val lib = project("common-lib")
     .libraryDependencies(loggingDeps ++ awsDeps ++ elasticsearchDeps ++
       playDeps ++ playWsDeps ++ scalazDeps ++ commonsIODeps ++ akkaAgentDeps ++
-      pandaDeps ++ imagingDeps ++ reactiveXDeps)
+      pandaDeps ++ imagingDeps)
     .testDependencies(scalaCheckDeps ++ scalaTestDeps)
 
   val thrall = playProject("thrall")
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -29,7 +29,7 @@ object Build extends Build {
   val lib = project("common-lib")
     .libraryDependencies(loggingDeps ++ awsDeps ++ elasticsearchDeps ++
       playDeps ++ playWsDeps ++ scalazDeps ++ commonsIODeps ++ akkaAgentDeps ++
-      pandaDeps ++ imagingDeps)
+      pandaDeps ++ imagingDeps ++ reactiveXDeps)
     .testDependencies(scalaCheckDeps ++ scalaTestDeps)
 
   val thrall = playProject("thrall")
--- a/usage/app/lib/Config.scala
+++ b/usage/app/lib/Config.scala
@@ -24,6 +24,7 @@ object Config extends CommonPlayAppProperties with CommonPlayAppConfig {
   lazy val loginUriTemplate = services.loginUriTemplate
 
   val defaultPageSize = 100
+  val defaultMaxRetries = 6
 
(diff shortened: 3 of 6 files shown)


usage/app/lib/ContentPollStream.scala
```diff
@@ -107,8 +115,28 @@ trait ContentPollStream {
...
+
+        Logger.info(fullMessage)
+
+        if (isLastAttempt) throw attempt.lastFail
+        Observable.timer(waitSeconds)
+      })
+    )
+
+  lazy val rawObservable = retryingPollObservable
+    .observeOn(scheduler)
```
kenoir: If we do not use a single threaded scheduler here it will simply try the next poll event on a different thread!

usage/app/lib/Config.scala
```diff
@@ -32,6 +33,7 @@ object Config extends CommonPlayAppProperties with CommonPlayAppConfig {
   val capiPreviewUser = properties("capi.preview.user")
   val capiPreviewPassword = properties("capi.preview.password")
   val capiPageSize = Try(properties("capi.page.size").toInt).getOrElse[Int](defaultPageSize)
+  val capiMaxRetries = Try(properties("capi.maxRetries").toInt).getOrElse[Int](defaultMaxRetries)
```
akash1810: Need to add to cloud formation template too?

project/Build.scala
```diff
@@ -29,7 +29,7 @@ object Build extends Build {
   val lib = project("common-lib")
     .libraryDependencies(loggingDeps ++ awsDeps ++ elasticsearchDeps ++
       playDeps ++ playWsDeps ++ scalazDeps ++ commonsIODeps ++ akkaAgentDeps ++
-      pandaDeps ++ imagingDeps)
+      pandaDeps ++ imagingDeps ++ reactiveXDeps)
```
akash1810: Everything to use rx? :smiley_cat:

kenoir: Created an object in common lib that provides a single threaded scheduler so this seems natural.

usage/app/lib/Config.scala
kenoir: I'm not sure if it's worth it. This allows for a default (set above) so i'm not sure if it's worth actually setting it.

This pattern allows us to override this value manually with config if we ever need to (like locally).

usage/app/lib/ContentPollStream.scala
```diff
@@ -6,11 +6,14 @@ import scala.concurrent.duration._
...
 import scala.concurrent.Future
 
+import rx.lang.scala.schedulers.ExecutionContextScheduler
+
 import com.gu.contentapi.client.GuardianContentClient
 import com.gu.contentapi.client.model.SearchResponse
 import com.gu.contentapi.client.model.v1.Content
 
 import com.gu.mediaservice.lib.aws.{DynamoDB, NoItemFound}
+import com.gu.mediaservice.lib.rx.SingleThreadedScheduler
```
jamesgorrie: Have you added this file?

kenoir: I have now.

jamesgorrie:
:+1:

common-lib/src/main/scala/com/gu/mediaservice/lib/rx/SingleThreadedScheduler.scala
```diff
@@ -0,0 +1,13 @@
+package com.gu.mediaservice.lib.rx
+
+import rx.lang.scala.schedulers.ExecutionContextScheduler
+import java.util.concurrent.Executors
+import scala.concurrent.ExecutionContext
+
+
+trait SingleThreadedScheduler {
```
theefer: Does this need to live in common (and pull in rx) if it's only used by usage?

usage/app/lib/ContentPollStream.scala
```diff
@@ -107,8 +115,28 @@ trait ContentPollStream {
...
 
   val pollInterval  = Duration(pollIntervalInSeconds, SECONDS)
-  val rawObservable =  Observable.interval(pollInterval)
-    .flatMap(_ => getContent)
+  case class PollAttempt(lastFail: Throwable, i: Int)
+
+  lazy val retryingPollObservable = Observable.interval(pollInterval)
+    .flatMap(_ => getContent).retryWhen(
+      _.zipWith(Observable.from(1 to maxRetries))((e: Throwable, i: Int) => PollAttempt(e,i))
+      .flatMap(attempt => {
```
theefer: Note: you can destructure in the flatMap to avoid having to dereference properties off the object every time:

``` scala
  flatMap { case PollAttempt(lastFail, i) =>
    // can use i and lastFail
  }
```

usage/app/lib/ContentPollStream.scala
```diff
@@ -107,8 +115,28 @@ trait ContentPollStream {
...
+        val isLastAttempt = attempt.i >= maxRetries
+        val waitSeconds = pow((attempt.i),2).second
```
theefer: Might be worth dropping a small comment explaining that this is exponential backoff in case it's not immediately obvious :-)

usage/app/lib/ContentPollStream.scala
kenoir: Nice tip.

kenoir:
@theefer addressed your comments.

PR #1584, opened 2015-12-30
updating procfile for grid_runner

NickPapacostas:
adding tildes tells sbt to run the processes in "watch" mode which keeps them open. without it the processes that were spawned in ruby would just exit successfully. I haven't found any downsides to this but lemme know if you think there might be

re: https://github.com/NickPapacostas/grid_runner


akash1810:
:+1:

Issue #1308, opened 2015-09-26
Docker-compose / Kubernetes


lucmichalski:
Hi guys,

Thanks for this great repo, is there any Kubernetes deployment files or Docker-compose to test this framework ?

Have a good week end all !

theefer:
We don't have any yet, but that'd be really interesting to help make it easier to build and deploy. Contributions welcome!

EdgeCaseBerg:
:+1: would love to see an easy way to boot this up with docker

lucmichalski:
I see much more crane or kubernetes to deploy such framework. As soon as I finish a couple of stuff, I will check that out.

Also, I would see an extension with Neo4J of Grid:
https://github.com/kbastani/spark-neo4j/blob/master/app/docker-compose.yml

akash1810:
Fixed in https://github.com/guardian/grid/pull/1586.

Issue #1090, opened 2015-08-13
link to dev-nginx is broken


mingfang:
I'm getting 404 here https://github.com/guardian/dev-nginx

akash1810:
Hi, we've been made aware of this here https://github.com/guardian/grid/issues/1077, more specifically, @theefer 's [comment](https://github.com/guardian/grid/issues/1077#issuecomment-130363247). We're working on fixing this.

nvercamm:
Hello,

can we get this repo from somewhere? I'm trying to setup your grid system

TravelTrexIT:
Hi Guys, we are hanging with our test installation of grid. Can you estimate when the dev Link will be available again?

theefer:
The dev-nginx script is really just a helper to install SSL certificates and setup a local nginx config. You should be able to run all the services without it, on ports 9001-9008 and, if needed, configure a local http server to serve these different localhost ports over different virtual host domains?

TravelTrexIT:
we are actaully missing the setup-app.rb script to perform step 6 command of the nginx setup. 
"sudo /setup-app.rb /nginx-mapping.yml" 
I thought that might be in dev-nginx. Can you give me a hint where to find it?

theefer:
I'll be looking into open sourcing the dev-nginx repo. That said, as mentioned above, it's really just a helper that writes some nginx virtual-host config files for all the domains described in the `nginx-mapping.yml` file, with some local SSL certificates, e.g. `https://api.media.local.domain.com/` => `http://localhost:9001/`. You should be able to set these up even without dev-nginx if you want to run the service now.

Sorry for the inconvenience.

kbossee:
nginx-mapping.yml wasn't very clear for me, I had to look at Services.scala to really get how the urls are composed.

Suppose your domain is "example.com", and you choose the most obvious name for the app: grid.

Then the urls become:
```
kahuna            grid.example.com
media-api         api.grid.example.com
image-loader      loader.grid.example.com
cropper           cropper.grid.example.com
metadata-editor   grid-metadata.example.com
imgops            grid-imgops.example.com
grid-usage        grid-usage.example.com
```

With these urls, you should have in your settings.ini
```
domain_root = grid.example.com
panda_domain = example.com
```

There are some checks in the code that require https.
If you need to create an ssl-certificate, I followed [this tutorial](http://articles.slicehost.com/2007/12/19/ubuntu-gutsy-self-signed-ssl-certificates-and-nginx)

[...]

PR #1586, opened 2016-01-04
docker in DEV

akash1810:
# USAGE

```sh
cd /path/to/grid/docker
docker-compose up -d
```

The `configs` container will automatically generate the configs for all the services. All that is required is a a settings file in `$HOME/.gu/grid/grid-settings.yml`. See [README](https://github.com/guardian/grid/blob/aa-docker/docker/README.md) for more info.

`ftp-watcher` and `reaper` haven't been dockerized because we don't run them in DEV. Will raise as a separate PR if needed.

NB: Default performance on osx is horrible because vboxsf :cry:. Use [docker-osx-dev](https://github.com/brikis98/docker-osx-dev).

NB: The initial launch will take a while as the ivy cache is populated. The ivy cache is shared between all containers to alleviate this somewhat.

# TODO
- [x] Investigate perf issues on osx. Use a samba container to share files between host and service containers? Or [docker-osx-dev](https://github.com/brikis98/docker-osx-dev).
- [x] Write nginx configs for each service
- [x] Script to migrate local elastic to docker elastic. Use [elasticdump](https://github.com/taskrabbit/elasticsearch-dump).
- [x] Dockerize `imgops`
- [x] Docs


akash1810:
Ready for review!

akash1810:
fixes https://github.com/guardian/grid/issues/1308
fixes https://github.com/guardian/grid/issues/1090 (this PR has an nginx conf generator)

docker/Dockerfile-imgops
```diff
@@ -0,0 +1,17 @@
+FROM ubuntu:14.04
```
akash1810: Not too sure what image is the best to use here - `debian:jessie` would work too. The main consideration/concern is the size of the base image to download.

docker/configs/generators/generators/dot_properties.py
```diff
@@ -0,0 +1,47 @@
+import os
+import logging
+from . import *
+from cloudformation import get_stack_outputs
+
+LOGGER_NAME = os.path.splitext(os.path.basename(__file__))[0]
+LOGGER = logging.getLogger(LOGGER_NAME)
+OUTPUT_DIR = '/configs/etc/gu'
```
akash1810: This path is only correct within the context of docker. The generator scripts can be used outside of docker, so this should probably be configurable...

docker/configs/generators/generators/nginx.py
```diff
@@ -0,0 +1,40 @@
+import os
```
akash1810: replaces the dependancy on `dev-nginx`

docker/configs/generators/grid-settings.yml.template
```diff
@@ -0,0 +1,30 @@
+# awscli profile name, defaulted to media-service Hint cat ~/.aws/credentials
```
akash1810: yaml file replaces [ini](https://github.com/guardian/grid/blob/master/scripts/dot-properties/settings/settings.ini.template) one because we use yaml elsewhere.

docker/docker-compose-sbt.yml
```diff
@@ -0,0 +1,10 @@
+# Base container for sbt services. In a seperate file so that a container doesn't get created for it by default.
+sbt:
+  build: .
+  dockerfile: Dockerfile-sbt
+  restart: unless-stopped
+  volumes:
+    - "./entrypoints:/entrypoints"
+  entrypoint: /entrypoints/sbt.sh
+  expose:
+    - "9000"
```
akash1810: all sbt projects are run on port 9000 in their container with the nginx container operating as a proxy.

docker/docker-compose.yml
```diff
@@ -0,0 +1,142 @@
...
+  volumes_from:
+    - configs
+  entrypoint: /entrypoints/nginx.sh
+
+# Shared volume container
+data:
+  image: java:8
+  volumes:
+    - "../:/code"
+    - "/root/.ivy2"
```
akash1810: share the ivy cache with all containers so that they don't have to populate the cache themselves

docker/docker-compose.yml
```diff
@@ -0,0 +1,142 @@
...
+
+elasticsearch:
+  build: .
+  dockerfile: Dockerfile-elastic
+  container_name: elasticsearch
+  expose:
+    - "9200"
+    - "9300"
+  ports:
+    - "9200:9200"
```
akash1810: elastic is available on `localhost:9200` as usual and also as `https://elasticsearch.foobar.co.uk`

akash1810:
Crude idle performance analysis on osx by running `docker stats $(docker ps -a -q)` (my VM is allocated 1 core and 8GB RAM).

```
CONTAINER           CPU %               MEM USAGE / LIMIT     MEM %               NET I/O               BLOCK I/O
0337c9ad827b        0.32%               245.1 MB / 8.365 GB   2.93%               358.1 kB / 407.8 kB   24.58 kB / 311.3 kB
27d66cdebf1d        4.09%               533 MB / 8.365 GB     6.37%               692.9 kB / 255.2 kB   0 B / 53.25 kB
37cbb4e043bc        2.58%               451.8 MB / 8.365 GB   5.40%               24.61 kB / 5.483 kB   0 B / 49.15 kB
750453c082be        1.15%               382.2 MB / 8.365 GB   4.57%               10.26 kB / 648 B      0 B / 40.96 kB
950212eb52c9        8.73%               533.6 MB / 8.365 GB   6.38%               5.012 MB / 5.643 MB   0 B / 49.15 kB
a7b05dd95b9b        2.89%               532.9 MB / 8.365 GB   6.37%               3.262 MB / 3.134 MB   0 B / 2.978 MB
dafa8603382a        0.00%               3.105 MB / 8.365 GB   0.04%               5.535 MB / 5.452 MB   0 B / 8.192 kB
e14d88f33022        2.70%               476.1 MB / 8.365 GB   5.69%               136.1 kB / 31.76 kB   0 B / 49.15 kB
e88e47d97f70        0.00%               0 B / 0 B             0.00%               0 B / 0 B             0 B / 0 B
e99d3c095bbb        0.00%               0 B / 0 B             0.00%               0 B / 0 B             0 B / 0 B
f24968c800b4        3.17%               505.8 MB / 8.365 GB   6.05%               193.6 kB / 77.2 kB    0 B / 49.15 kB
f5bb1f5566b0        0.00%               3.174 MB / 8.365 GB   0.04%               20.65 kB / 14.32 kB   0 B / 4.096 kB
f9abafe9cc11        4.64%               498.9 MB / 8.365 GB   5.96%               325.8 kB / 302.6 kB   0 B / 61.44 kB
```

docker/configs/generators/generators/dot_properties.py
jamesgorrie: Why not use `/etc/gu` on docker

akash1810: The `configs` container generates all the necessary configs (`.properties` files, `nginx.conf` etc.) and shares them via a `volume` directive. I'm writing all configs to `/configs/<destination-path>` for the simplicity of sharing a single volume (`/configs`). The `entrypoint` of a container then copies the config files it needs from `/configs`.

Why is sharing one generic volume simpler than sharing many specific ones? Here's a (shortened) sequence of events I actually went through:

The `configs` container generates:
  - `nginx.conf` to `/etc/nginx/nginx.conf`
  - nginx proxy server configs to `sites-enabled/<grid-service>.conf`
  - `.properties` files to `/etc/gu`

In order to share these generated files with other containers, the directive would look like:

```yaml
configs:
  image: python:2.7
  volumes:
    - "./entrypoints:/entrypoints"
    - "./configs/generators:/configs/generators"
    - "$HOME/.aws/:/root/.aws:ro"
    - "$HOME/.gu/grid/:/root/.gu/grid"
    - "/etc/gu"
    - "/etc/nginx/nginx.conf"
    - "/etc/nginx/sites-enabled"
  entrypoint: /entrypoints/configs.sh
```

But, sharing a file (`/etc/nginx/nginx.conf`) as a volume is a bit silly! So we'd have to share `/etc/nginx/` as a whole. This means the contents of `/etc/nginx` on the `configs` container has to have all the files needed for nginx to run. Which means running (or at least installing) nginx on that container :mask: !

I then added the `imgops` container to the mix, which requires its own `nginx.conf` too, but where to put it? `/etc/nginx/imgops/nginx.conf`? :mask: !!

So I settled on creating:
  - `/configs/etc/gu/*.properties`
  - `/configs/nginx/nginx.conf`
  - `/configs/nginx/sites-enabled/*.conf`
  - `/configs/imgops/nginx.conf`

And sharing `/configs`.

Couldn't each container just generate its own config files? I started off doing this and it felt really ugly and hard to understand (and maintain) and goes against the idea of a container doing only one thing.

(Hmm... maybe I should write a blog post...)

akash1810: made the output directory a required command line argument - https://github.com/guardian/grid/commit/876cdd492f03ad2f570a779e8b17381b2e0d9083

docker/docker-compose.yml
```diff
@@ -0,0 +1,141 @@
...
+    - "9300:9300"
+  volumes:
+    - "./configs/elasticsearch:/usr/share/elasticsearch/config"
```
akash1810: Might also be a good idea to mount a host directory to the `/usr/share/elasticsearch/data` volume so that the data is persisted outside of the container.

akash1810:
Added a few more docs - the todo list is complete now! :smile: 

(will squash the commits after thumbs)

docker/README.md
```diff
@@ -0,0 +1,91 @@
...
+## Requirements
+- Docker
+- Docker Compose
+
+### Additional requirements for OSX
+- [docker-osx-dev](https://github.com/brikis98/docker-osx-dev)
+
+## Setup
+- Ensure you've set up the aws-cli with a `media-service` profile (`aws configure --profile media-service`).
+- Create `$HOME/.gu/grid/grid-settings.yml` using [`grid-settings.yml.template`](./configs/generators/grid-settings.yml.template) as a template.
```
akash1810: By default, the docker-machine VM has [limited access](https://docs.docker.com/engine/userguide/dockervolumes/#mount-a-host-directory-as-a-data-volume) to the host file system, so the location changed from `/etc/hosts` to `$HOME/.gu`.

> If you are using Docker Machine on Mac or Windows, your Docker daemon has only limited access to your OS X or Windows filesystem. Docker Machine tries to auto-share your /Users (OS X) or C:\Users (Windows) directory.

> All other paths come from your virtual machine’s filesystem. For example, if you are using VirtualBox some other folder available for sharing, you need to do additional work. In the case of VirtualBox you need to make the host folder available as a shared folder in VirtualBox. Then, you can mount it using the Docker -v flag.

Alternatively, we could share `/etc/gu` from an OSX host to the Docker VM, however this seemed better as it means fewer setup steps.

Alternatively alternatively, we could use linux by default...

PR #1588, opened 2016-01-04
use viewport height instead of setting the px height

jamesgorrie:
This means we can let the browser do the working by setting it to use `vh`.

It also means we can stop using the debounce, which is causing a bug by setting the height at the wrong interval.


theefer:
:+1:

PR #1589, opened 2016-01-04
Move gr-downloader into components

theefer:
Not sure why it was at the root of `js` but feels like it belongs in `components` like the other components.


jamesgorrie:
It's before we had the component directory where everything was a component.
:+1:

PR #1591, opened 2016-01-04
track keyboard shortcuts

akash1810:


jamesgorrie:
Could we abstract this out to a `shortcuts` component that automatically tracks in case people forget?

akash1810:
> Could we abstract this out to a shortcuts component that automatically tracks in case people forget?

Done.

jamesgorrie:
:+1:

PR #1592, opened 2016-01-04
Move batch results archiving to button in action bar

theefer:
I also extracted the logic into a new component.

Note that we now have two similar components, one for batch-archiving on the results page and another one for archiving on the image page. They have slightly different semantics, which isn't ideal, but we could unify them as a separate phase, since at least they're more aligned now.

# Before

image

# After

image

### DIFF
--- a/kahuna/public/js/components/gr-archiver/gr-archiver.js
+++ b/kahuna/public/js/components/gr-archiver/gr-archiver.js
@@ -23,15 +23,14 @@ module.controller('grArchiverCtrl', [
 
         const ctrl = this;
 
-        $scope.$watchCollection(() => ctrl.images.toJS(), (images) => {
+        $scope.$watchCollection(getImageArray, (images) => {
             const allArchived = images.every(imageAccessor.isArchived);
             ctrl.archivedState = allArchived ? 'archived' : 'unarchived';
         });
 
         ctrl.archive = () => {
             ctrl.archiving = true;
-            var imageArray = Array.from(ctrl.images);
-            archiveService.batchArchive(imageArray)
+            archiveService.batchArchive(getImageArray())
                 .then(() => {
                     ctrl.archiving = false;
                 });
@@ -39,12 +38,16 @@ module.controller('grArchiverCtrl', [
 
         ctrl.unarchive = () => {
             ctrl.archiving = true;
-            var imageArray = Array.from(ctrl.images);
-            archiveService.batchUnarchive(imageArray)
+            archiveService.batchUnarchive(getImageArray())
                 .then(() => {
                     ctrl.archiving = false;
                 });
         };
+
+        function getImageArray() {
+            // Convert from Immutable Set to JS Array
+            return Array.from(ctrl.images);
+        }
     }
 ]);
 
--- a/kahuna/public/js/components/gr-archiver/gr-archiver.js
+++ b/kahuna/public/js/components/gr-archiver/gr-archiver.js
@@ -34,7 +34,7 @@ module.controller('grArchiverCtrl', [
         ctrl.archive = () => {
             ctrl.archiving = true;
             archiveService.batchArchive(getImageArray())
-                .then(() => {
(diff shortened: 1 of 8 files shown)


theefer:
The collapsed icon-only version is very confusing too now:

image

But it just highlights that we shouldn't be using the same lock icon for two completely different purposes (library and locking panels). The icon-only version is only on very narrow screens, but we already wanted to review the library icon/states, so worth doing ASAP.

jamesgorrie:
Note: [There is the bookmark icons in the MD set](https://design.google.com/icons/#ic_bookmark).

kahuna/public/js/components/gr-archiver/gr-archiver.js
```diff
@@ -0,0 +1,62 @@
...
+    'imageAccessor',
+    function ($scope,
+              archiveService,
+              imageAccessor) {
+
+        const ctrl = this;
+
+        $scope.$watchCollection(() => ctrl.images.toJS(), (images) => {
+            const allArchived = images.every(imageAccessor.isArchived);
+            ctrl.archivedState = allArchived ? 'archived' : 'unarchived';
```
jamesgorrie: What's the default of `archivedState`? Is it worth setting on the controller instantiation for clarity?

kahuna/public/js/components/gr-archiver/gr-archiver.js
```diff
@@ -0,0 +1,62 @@
...
+        });
+
+        ctrl.archive = () => {
+            ctrl.archiving = true;
+            var imageArray = Array.from(ctrl.images);
```
jamesgorrie: `const imageArray = ctrl.images.toJS()`?

jamesgorrie:
Much clearer and probably handy to reuse.
:+1:

kahuna/public/js/components/gr-archiver/gr-archiver.js
theefer: I wonder if `Array.from` may be more generic than `toJS`, as it works with any iterable structure, incl plain JS arrays. I might change the `toJS` above into an `Array.from` instead for consistency, if that works with you?

kahuna/public/js/components/gr-archiver/gr-archiver.js
theefer: Technically this watch gets executed immediately when the controller is initialised, but I imagine you're referring to making it obvious what properties exist on that controller?

I don't really like repetition but I think explicitness may be worth it in this case. Could do the same with `ctrl.archiving = false`.

theefer:
@jamesgorrie Addressed your comments and added a few improvements.

kahuna/public/js/components/gr-archiver/gr-archiver.js
jamesgorrie: I didn't mind, there's just two different ways, I was wondering how it was worried about it working on a custom iterable, but I guess that's how it works with `NodeList`s etc.

The `toJS()` does help recognise that it from `Immutable` without a typesystem though.

jamesgorrie:
Your comment does what my comment was saying about recognising it from Immutable.
:

kahuna/public/js/components/gr-archiver/gr-archiver.js
theefer: I guess I was coming at it the other way, making the directive compatible with any `Iterable` object, rather than specifically ImmutableJS. So for example if the images end up refactored into a JS array, things will keep working. Probably overly superstitious, I admit.

paperboyo:
> using the same lock icon for two completely different purposes

Oh yes! How that happened? Having a lock panel icon separate from show panel button (they used to be one) makes my usual routine of opening and immediately locking a panel even more arduous than before. :cry:

theefer:
I'm going to be looking at the three-state library icon, and the duplicate use of lock for it and panels, as the next thing I pick up.

PR #1593, opened 2016-01-04
Collections panel ui+with state

jamesgorrie:
Saves the state of the collection tree in localStorage.
Could refactor into a userPrefs module, but not sure we need that just yet.


kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
```diff
@@ -11,39 +10,56 @@
...
-           ui:sref="search.results({query: (node.data.name)})">
+           ui:sref="search.results({query: ctrl.getCollectionQuery(node.data.content.pathId)})">
             {{node.data.name}}
         </a>
 
+
         <button class="node__action clickable" type="button"
                 ng:if="ctrl.deletable"
-                ng:click="ctrl.remove()">
+                ng:show="ctrl.editing"
```
theefer: ng:if vs ng:show again?

kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
```diff
@@ -11,39 +10,56 @@
...
+        <button class="node__action clickable"
+                ng:if="!ctrl.editing && ctrl.hasImagesSelected && !ctrl.saving"
+                ng:click="ctrl.addImagesToCollection()"
+                gr:tooltip="Add selected images to this collection">
+            <gr-icon>add_to_photos</gr-icon>
         </button>
+        <span ng:if="ctrl.saving">Saving…</span>
+    </div>
+
+    <form class="node__add-child" ng:show="active" ng:submit="ctrl.addChild(childName).then(clearForm);">
```
theefer: Would `ng:if` be more conservative in not having one form per tree node in the DOM?

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -30,22 +35,136 @@ grCollectionsPanel.controller('GrCollectionsPanelCtrl', [
...
+    }
+
+
+    function setState(pathId, show) {
+        const newState = angular.extend({}, getStateFromStorage(), {[pathId]: show});
+        $window.localStorage.setItem(localStorageKey, JSON.stringify(newState));
+    }
+
+    function getState(pathId) {
+        const state = getStateFromStorage();
```
theefer: Does this mean you're going to hit `localStorage` for every single node in the tree? Could be a bit expensive. Is it worth caching the tree once and for all (and invalidate/update the cache on `setState`)?

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -30,22 +35,136 @@ grCollectionsPanel.controller('GrCollectionsPanelCtrl', [
...
+        return angular.isDefined(state[pathId]) ? state[pathId] : defaultShowState;
```
theefer: Maybe we should start using `Option` in JS :-)

kahuna/public/js/search-query/query-syntax.js
```diff
@@ -17,3 +18,7 @@ export function removeLabel(q, label) {
 export function removeLabels(q, labels) {
     return labels.reduce((q, curr) => removeLabel(q, curr.name), q);
 }
+
+export function     getCollection(path) {
```
theefer: I think this got fixed elsewhere. Does master need merged in so changes are reflected?

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -30,22 +35,136 @@ grCollectionsPanel.controller('GrCollectionsPanelCtrl', [
...
+    ctrl.getCollectionQuery = path => getCollection(path);
+
+    // TODO: move this somewhere sensible, we probably don't want an observable for each node.
+    const add$ = new Rx.Subject();
+    const pathWithImages$ =
+            add$.withLatestFrom(ctrl.selectedImages$, (path, images) => ({path, images}));
+
+    const hasImagesSelected$ = ctrl.selectedImages$.map(i => i.size > 0);
+    ctrl.addImagesToCollection = () => {
+        ctrl.saving = true;
```
theefer: Could this live in the `subscribe$` too to isolate side-effects and group the changes to `ctrl.saving` in the same code location?

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -30,22 +35,136 @@ grCollectionsPanel.controller('GrCollectionsPanelCtrl', [
...
+    const defaultShowState = true;
+
+    function getStateFromStorage() {
+        const jsonStr = $window.localStorage.getItem(localStorageKey) || '{}';
+        return JSON.parse(jsonStr);
+    }
+
+
+    function setState(pathId, show) {
+        const newState = angular.extend({}, getStateFromStorage(), {[pathId]: show});
```
theefer: As discussed, should we only store one of the states (either all `true` or all `false` values), and let the rest be the default?

theefer:
A few comments, and there are merge conflicts, but looks generally sane.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
jamesgorrie: Legacy of not merging.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
jamesgorrie: Legacy of not merging, apologies.

kahuna/public/js/search-query/query-syntax.js
jamesgorrie: Merge :rage4:

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
theefer: I had a recollection that accessing `localStorage` could be really slow. Can't find anything to back it up so I might be wrong, and this is probably OK for now, but could be worth addressing later as still feels like a bit of a waste not to cache it.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
jamesgorrie: Happy to do in another PR as not relevant here.
I'd like to turn a few other things into streams too.

jamesgorrie:
I've accessed `localStorage` a lot less and used a set as it's nicer.

I was going down the route of making it a stream (should have in the first place) but it started getting academic, so stopped.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
jamesgorrie: I suppose it's I/O technically, so could be, but it was simple to cache.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -37,13 +39,44 @@ grCollectionsPanel.controller('GrCollectionsPanelCtrl', [
     ctrl.selectedCollections = selectedCollections;
 }]);
 
+grCollectionsPanel.factory('collectionsTreeState', ['$window', function($window) {
+    // TODO: Add garbage collection to state.
+    const localStorageKey = 'collectionsTreeOpen';
+    const jsonStr = $window.localStorage.getItem(localStorageKey) || '[]';
+    const jsonArr = JSON.parse(jsonStr);
+    // A little bit of security incase this was set weirdly before.
```
theefer: If you want to be paranoid, you could surround the `JSON.parse` with `try-catch`. Up to you.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -37,13 +39,44 @@ grCollectionsPanel.controller('GrCollectionsPanelCtrl', [
...
+    // The var is use so we don't have to read from localStorage the whole time.
+    var stateCache = Array.isArray(jsonArr) ? new Set(jsonArr) : new Set();
+
+
+    function setState(pathId, show) {
+        const newState = show ? stateCache.add(pathId) : stateCache.delete(pathId);
```
theefer: AFAICS you're using ES6 Sets, not Immutable Sets, so `add` and `delete` would mutate `stateCache` too?

jamesgorrie: True, I was using Immutable, and thought I would swap back, forgetting the reason I used them in the first place was because ES6 ones are crapper.

theefer: Might as well use the ES6 mutable one inside here and then you can replace your `var` with a `const` ;-)

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
jamesgorrie: Done.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
jamesgorrie: Good call, done.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -37,13 +39,55 @@ grCollectionsPanel.controller('GrCollectionsPanelCtrl', [
...
+    const localStorageKey = 'collectionsTreeOpen';
+    const jsonStr = $window.localStorage.getItem(localStorageKey) || '[]';
+
+    // A little bit of superstition in case this was set weirdly before.
+    let jsonArr = [];
+    try {
+        let jsonArr = JSON.parse(jsonStr);
+        if (!Array.isArray(jsonArr)) {
+            throw new Error(`Invalid ${localStorageKey} json: ${jsonArray}`);
+        }
```
theefer: I'm confused by this. Why re-declare `jsonArr` with another `let`? That means the parent `jsonArr` will never be set to anything but its default?

jamesgorrie: I'm not sure if it's a bug or feature, but ES6 (at least the transpilers) scope `const`s in the `try catch` blocks respectively.

So:
```
try {
  const jsonArr = JSON.parse(jsonStr);
} catch(e) {
  const jsonArr = []
}
```
transpiles too:
```JS
try {
    var jsonArr = JSON.parse(jsonStr);
  } catch (e) {
    var jsonArr$__1 = [];
  }
```

> the parent jsonArr

I'm not sure which parent `val` you're talking about?

theefer: The `let` in `let jsonArr = JSON.parse(jsonStr);` is block-scoped (i.e. scoped to the `try`), so it locally shadows the parent `let jsonArr = [];`, which is never updated and always equal to `[]`.

[Reduced example](https://google.github.io/traceur-compiler/demo/repl.html#var%20json%20%3D%20'%7B%7D'%0A%0Afunction%20parse(json)%20%7B%0A%20%20let%20jsonArr%20%3D%20%5B%5D%3B%0A%20%20try%20%7B%0A%20%20%20%20let%20jsonArr%20%3D%20JSON.parse(json)%3B%0A%20%20%20%20if%20(!Array.isArray(jsonArr))%20%7B%0A%20%20%20%20%20%20throw%20new%20Error(%60Invalid%20%24%7BlocalStorageKey%7D%20json%3A%20%24%7BjsonArray%7D%60)%3B%0A%20%20%20%20%7D%0A%20%20%7D%20catch%20(e)%20%7B%0A%20%20%20%20console.log(%22ERROR%3A%22%2C%20e)%0A%20%20%7D%0A%20%20return%20jsonArr%3B%0A%7D%0A%0Aconsole.log(%22parsing%20%22%2C%20'%7B%7D'%2C%20%22expected%3A%20%5B%5D%2C%20actual%3A%20%22%2C%20parse('%7B%7D'))%0Aconsole.log(%22parsing%20%22%2C%20'%5B1%2C2%5D'%2C%20%22expected%3A%20%5B1%2C2%5D%2C%20actual%3A%20%22%2C%20parse('%5B1%2C2%5D'))%0Aconsole.log(%22parsing%20%22%2C%20'%7B'%2C%20%22expected%3A%20%5B%5D%2C%20actual%3A%20%22%2C%20parse('%7B'))) (open the console)

Removing the inner `let` fixes the issue, but breaks the case if the parsed JSON isn't an Array.

theefer: For the Array checking, you could just set the value to `[]` in that case (possibly logging an error). Using exceptions is best avoided for normal control flow.

jamesgorrie: Sorry I misread what you said, and thus missed my error.
Removed the unnecessary error and just set the value.
I've also updated the `localStorage` value to stop it happening again.

theefer:
:+1:

PR #1595, opened 2016-01-04
Don't render results if search failed

theefer:
This should fix the Sentry errors we've seen recently.

### DIFF
--- a/kahuna/public/js/search/results.js
+++ b/kahuna/public/js/search/results.js
@@ -154,6 +154,9 @@ results.controller('SearchResultsCtrl', [
             }
 
             return images;
+        }).catch(error => {
+            ctrl.loadingError = error;
+            return $q.reject(error);
         }).finally(() => {
             ctrl.loading = false;
         });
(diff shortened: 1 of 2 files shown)


kahuna/public/js/search/results.js
```diff
@@ -154,6 +154,9 @@ results.controller('SearchResultsCtrl', [
             }
 
             return images;
+        }).catch(error => {
+            ctrl.loadingError = error;
```
akash1810: When will this happen?

theefer: If the user doesn't have a session and gets a 401 back, or if the server is broken or down (5xx).

akash1810: Just an observation, nothing of particular interest... the [global banner message](https://github.com/guardian/grid/blob/master/kahuna/public/js/errors/http.js) should also show in these cases too.

akash1810:
:+1:

kahuna/public/js/search/results.js
theefer: Yes, I think we need to think of a way to bring errors back into their location of origin (the button triggered, the area not rendered) instead of always having global errors.

PR #1596, opened 2016-01-05
Avoid double scrollbars w meta panel

theefer:
Fixes when a selection is present and the meta panel hidden:
image

cc @jamesgorrie was there any reason why only the `x` overflow was being hidden?


tsop14:
:+1:

theefer:
Going to wait for @jamesgorrie in case I missed some reason why the CSS was the way it was.

jamesgorrie:
No reason. :+1:

PR #1599, opened 2016-01-06
check last modified before updating metadata

NickPapacostas:
adding last modified to metadata editor sqs notifications and checking them in thrall before updating

### DIFF
--- a/metadata-editor/app/controllers/EditsController.scala
+++ b/metadata-editor/app/controllers/EditsController.scala
@@ -195,7 +195,7 @@ object EditsController extends Controller with ArgoHelpers with DynamoEdits with
     val message = Json.obj(
       "id" -> id,
       "data" -> Json.toJson(edits),
-      "userMetadataLastModified" -> printDateTime(new DateTime())
+      "lastModified" -> printDateTime(new DateTime())
     )
 
     Notifications.publish(message, "update-image-user-metadata")
--- a/metadata-editor/app/controllers/EditsController.scala
+++ b/metadata-editor/app/controllers/EditsController.scala
@@ -195,7 +195,7 @@ object EditsController extends Controller with ArgoHelpers with DynamoEdits with
     val message = Json.obj(
       "id" -> id,
       "data" -> Json.toJson(edits),
-      "lastModified" -> printDateTime(new DateTime())
+      "userMetadataLastModified" -> printDateTime(new DateTime())
     )
 
     Notifications.publish(message, "update-image-user-metadata")
--- a/metadata-editor/app/controllers/EditsController.scala
+++ b/metadata-editor/app/controllers/EditsController.scala
@@ -4,6 +4,9 @@ package controllers
 import java.net.URI
 import java.net.URLDecoder.decode
 
+import com.gu.mediaservice.lib.formatting._
+import org.joda.time.DateTime
+
 import scala.concurrent.Future
 
 import play.api.data.Forms._
@@ -191,7 +194,8 @@ object EditsController extends Controller with ArgoHelpers with DynamoEdits with
     val edits = metadata.as[Edits]
     val message = Json.obj(
       "id" -> id,
-      "data" -> Json.toJson(edits)
+      "data" -> Json.toJson(edits),
+      "lastModified" -> printDateTime(new DateTime())
     )
 
     Notifications.publish(message, "update-image-user-metadata")
--- a/thrall/app/lib/ElasticSearch.scala
+++ b/thrall/app/lib/ElasticSearch.scala
@@ -131,19 +131,18 @@ object ElasticSearch extends ElasticSearchClient with ImageFields {
       .executeAndLog(s"removing exports from image $id")
       .incrementOnFailure(failedExportsUpdates) { case e: VersionConflictEngineException => true }
 
-  def applyImageMetadataOverride(id: String, metadata: JsValue, userMetadataLastModified: JsValue)(implicit ex: ExecutionContext): Future[UpdateResponse] = {
+  def applyImageMetadataOverride(id: String, metadata: JsValue, lastModified: JsValue)(implicit ex: ExecutionContext): Future[UpdateResponse] = {
     prepareImageUpdate(id)
       .setScriptParams(Map(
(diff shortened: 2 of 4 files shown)


thrall/app/lib/ElasticSearch.scala
```diff
@@ -131,19 +131,25 @@ object ElasticSearch extends ElasticSearchClient with ImageFields {
...
-        "userMetadata" -> asGroovy(metadata),
-        "lastModified" -> asGroovy(JsString(currentIsoDateString))
+        "userMetadata" -> asGroovy(userMetaData),
+        "lastModified" -> asGroovy(lastModified)
       ).asJava)
       .setScript(
-        "ctx._source.userMetadata = userMetadata;" +
-          refreshEditsScript +
-          updateLastModifiedScript,
+      s"""| if (ctx._source.lastModified && ctx._source.lastModified > lastModified) {} else {
```
kenoir: If you negate this, you could remove the `else` and `{}`.

thrall/app/lib/ThrallMessageConsumer.scala
```diff
@@ -43,7 +43,7 @@ object ThrallMessageConsumer extends MessageConsumer(
     withImageId(exports)(id => ElasticSearch.deleteImageExports(id))
 
   def updateImageUserMetadata(metadata: JsValue): Future[UpdateResponse] =
-    withImageId(metadata)(id => ElasticSearch.applyImageMetadataOverride(id, metadata \ "data"))
+    withImageId(metadata)(id => ElasticSearch.applyImageMetadataOverride(id, metadata))
```
theefer: For consistency with other methods, would it make sense to still pass the data as second argument and the date as third? Avoids having to teach `ElasticSearch` about message structure, and keeps that in this message consumer instead?

kenoir:
:+1:

thrall/app/lib/ElasticSearch.scala
theefer: Is this using the image's top-level `lastModified`? Is this safe, given that concurrent updates to  other  bits of the model (e.g. exports or usage) will also be updating the top-level `lastModified`?

thrall/app/lib/ThrallMessageConsumer.scala
NickPapacostas: yep

thrall/app/lib/ElasticSearch.scala
NickPapacostas: #logic ^_^ will do

kenoir: Isn't it using the `lastModified` set here: https://github.com/guardian/grid/pull/1599/files#diff-2454b3b50883ce82c6498f0096e91dc1R198 ?

theefer: Yes but what does it compare it to on the database record?

kenoir: Good point. Is the solution then  to change the structure of what gets written for `userMetadata` to include `lastModified` from the message as a sibling of `data` and `uri`?

kenoir: ^ and to perform the comparison on that date when writing (obvs).

metadata-editor/app/controllers/EditsController.scala
```diff
@@ -191,7 +194,8 @@ object EditsController extends Controller with ArgoHelpers with DynamoEdits with
     val edits = metadata.as[Edits]
     val message = Json.obj(
       "id" -> id,
-      "data" -> Json.toJson(edits)
+      "data" -> Json.toJson(edits),
+      "lastModified" -> printDateTime(new DateTime())
```
jamesgorrie: You can use `DateTime.now()`... just me preference.

thrall/app/lib/ElasticSearch.scala
jamesgorrie: Would we want them to be the same? Technically the metadata was updated at some stage, and now we're updating the image. This time difference could be useful?

metadata-editor/app/controllers/EditsController.scala
```diff
@@ -191,7 +194,8 @@ object EditsController extends Controller with ArgoHelpers with DynamoEdits with
     val edits = metadata.as[Edits]
     val message = Json.obj(
       "id" -> id,
-      "data" -> Json.toJson(edits)
+      "data" -> Json.toJson(edits),
+      "userMetadataLastModified" -> printDateTime(new DateTime())
```
kenoir: Can just be `lastModified`?

theefer: +1 to generic lastModified metadata in all messages.

thrall/app/lib/ThrallMessageConsumer.scala
```diff
@@ -43,7 +43,11 @@ object ThrallMessageConsumer extends MessageConsumer(
     withImageId(exports)(id => ElasticSearch.deleteImageExports(id))
 
   def updateImageUserMetadata(metadata: JsValue): Future[UpdateResponse] =
-    withImageId(metadata)(id => ElasticSearch.applyImageMetadataOverride(id, metadata \ "data"))
+    withImageId(metadata)(
+      id => ElasticSearch.applyImageMetadataOverride(
+        id, metadata \ "data", metadata \ "userMetadataLastModified"
+      )
+    )
```
theefer: Since you asked us to be anal about styling, one way to avoid multi-line statements is to use temporary variables, e.g.

``` scala
  def updateImageUserMetadata(metadata: JsValue): Future[UpdateResponse] = {
    val data = metadata \ "data"
    val lastModified = metadata \ "userMetadataLastModified"
    withImageId(metadata)(id => ElasticSearch.applyImageMetadataOverride(id, data, lastModified))
  }
```

Bit academic here, and the last line is probably still a bit long, but you get the idea.

kenoir:
:+1:

theefer:
:+1: if tested

NickPapacostas:
tested successfully! :smiley_cat:

PR #1601, opened 2016-01-06
Unify and correct archive states

theefer:
Rework archive buttons on search and image page to be more consistent throughout in the use of icons, actions available (if auto-persisted, just show disabled control with "Kept in Library"), etc.

Also move some logic into a new logic service that is more flexible than the current one-size-fits-all image-service.

recorded

Once this is out, we can review what icon would be best to represent the three states of auto-persisted, manually archived or not archived at all.

### DIFF
--- a/kahuna/public/js/components/gr-archiver/gr-archiver.html
+++ b/kahuna/public/js/components/gr-archiver/gr-archiver.html
@@ -5,7 +5,7 @@
             class="inner-clickable batch-archive__button"
             ng:click="ctrl.unarchive()"
             ng:disabled="ctrl.archiving">
-        <gr-icon-label gr-icon="lock_open">
+        <gr-icon-label gr-icon="lock_outline">
             <span ng:if="!ctrl.archiving">Remove from Library</span>
             <span ng:if="ctrl.archiving">Removing from Library…</span>
         </gr-icon-label>
@@ -16,7 +16,7 @@
             class="inner-clickable batch-archive__button"
             ng:click="ctrl.archive()"
             ng:disabled="ctrl.archiving">
-        <gr-icon-label gr-icon="lock">
+        <gr-icon-label gr-icon="lock_open">
             <span ng:if="!ctrl.archiving">Add to Library</span>
             <span ng:if="ctrl.archiving">Adding to Library…</span>
         </gr-icon-label>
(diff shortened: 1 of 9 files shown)


kahuna/public/js/components/gr-archiver/gr-archiver.html
```diff
@@ -1,11 +1,20 @@
 <div class="batch-archive"
      ng:switch="ctrl.archivedState">
+    <span ng:switch-when="kept"
+          type="button"
```
akash1810: is `type` a valid attribute for a `span`?

theefer: Nope. Thanks, will fix!

kahuna/public/js/components/gr-archiver/gr-archiver.js
```diff
@@ -58,6 +67,21 @@ module.controller('grArchiverCtrl', [
...
+        function listAllPersistenceExplanations(images) {
+            return images.
+                map(imageLogic.getPersistenceExplanation).
+                map(items => new Set(items)).
+                reduce((all, items) => all.union(items)).
+                toArray();
+        }
+
+        // Takes an array and generates a '1, 2, 3 or 4' string
+        function humanUnion(list) {
```
akash1810: nice

kahuna/public/js/services/image-logic.js
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Helpers to apply business logic on image resources.
+ */
+imageLogic.factory('imageLogic', ['imageAccessor', function(imageAccessor) {
+
+    function canBeDeleted(image) {
+        return image.getAction('delete').then(action => !! action);
+    }
+
+    function canBeArchived(image) {
```
akash1810: minor: is it worth calling this `canBePersisted` to have a common language? (ideally, would be done everywhere though :neutral_face: )

akash1810:
:+1:

kenoir:
@theefer how does this look with the meta panel locked? Does it result in two lock icons showing next to each other with different functions?

kahuna/public/js/services/image-logic.js
theefer: I thought there was a conscious use of both words to mean different things: archived = user toggle, persisted = derived state of whether the image can be deleted, with archived = true implying persisted = true?

akash1810: ah, yep :blush:

theefer:
@kenoir It's basically crap:

image

Although it's worst form is when the screen is too narrow and we fallback to icons only:

image

My plan is to replace the icons used for archive state with something else than a lock.

theefer:
Note that I've already moved the archive button into the action bar in a previous commit so this is already the case.

PR #1602, opened 2016-01-06
Show selected collection

tsop14:
Highlights collection(s) that result images belong to. It does not necessarily show all the images from a highlighted collection. 

~A: image1, image2
~B: image1, image3

search query: ~A ~B

will highlight ~A and ~B but only display image1

highlight

### DIFF
--- a/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
+++ b/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
@@ -17,8 +17,8 @@ export var grCollectionsPanel = angular.module('grCollectionsPanel', [
 ]);
 
 grCollectionsPanel.controller('GrCollectionsPanelCtrl', [
-    'collections', 'selectedImages$',
-    function (collections, selectedImages$) {
+    'collections', 'selectedImages$', 'selectedCollections',
+    function (collections, selectedImages$, selectedCollections) {
 
     const ctrl = this;
 
@@ -34,6 +34,7 @@ grCollectionsPanel.controller('GrCollectionsPanelCtrl', [
     }).catch(() => ctrl.error = true);
 
     ctrl.selectedImages$ = selectedImages$;
+    ctrl.selectedCollections = selectedCollections || [];
 }]);
 
 grCollectionsPanel.controller('GrNodeCtrl',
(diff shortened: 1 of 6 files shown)


kahuna/public/js/search-query/query-syntax.js
```diff
@@ -22,3 +26,12 @@ export function removeLabels(q, labels) {
 export function getCollection(path) {
     return createCollection(path);
 }
+
+export function getCollectionsFromQuery(q) {
+    const collections =  querySplit(q) ? querySplit(q)
```
theefer: Could save the result of `querySplit(q)` to avoid having to do it twice?

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -34,6 +34,7 @@ grCollectionsPanel.controller('GrCollectionsPanelCtrl', [
     }).catch(() => ctrl.error = true);
 
     ctrl.selectedImages$ = selectedImages$;
+    ctrl.selectedCollections = selectedCollections || [];
```
theefer: Out of curiosity, is the `|| []` ever needed?

kahuna/public/js/search-query/query-syntax.js
```diff
@@ -22,3 +26,12 @@ export function removeLabels(q, labels) {
...
+        .filter(bit => bit.charAt(0) === '~')
+        .map(path => path.replace(/('|"|~)/g, '').split('/'))
+        : '';
```
theefer: Why not return `[]` (empty list of collections) if there aren't any?

theefer:
:+1:

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
tsop14: Maybe, I put it in because sometimes it is `undefined` which then causes a type error when `.some` is called in: 

`ctrl.isSelected = ctrl.selectedCollections.some(col => {
        return angular.equals(col, ctrl.node.data.content.path);
    });`

Does that count? Or is there a better way to deal with that in `ctrl.isSelected`.

kahuna/public/js/search-query/query-syntax.js
tsop14: 👆 this fixes type error I mentioned above.

theefer:
:+1:

PR #1604, opened 2016-01-07
allow es host to be configured in .properties file

akash1810:
Finally got https://github.com/guardian/grid/pull/1586 working last night with native like perf on osx! Planning to break up https://github.com/guardian/grid/pull/1586 to make the diff smaller. ~~Although that's the main motivation, reading `es.host` from `.properties` means we can connect to other environments more easily too.~~

NB: `es.port` and `es.cluster` are static, so haven't moved them.

### DIFF
--- a/media-api/app/lib/Config.scala
+++ b/media-api/app/lib/Config.scala
@@ -24,7 +24,7 @@ object Config extends CommonPlayAppConfig with CommonPlayAppProperties {
 
   val elasticsearchHost: String =
     if (stage == "DEV")
-      string("es.host")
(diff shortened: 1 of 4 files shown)


theefer:
Not sure I understand what you mean by connecting to other environments?

akash1810:
Actually, neither do I! Was thinking the value of that property could be set to the dns name of an es machine in AWS... but the required ports are closed! So that's an incorrect side effect of this change!

theefer:
:+1:

PR #1605, opened 2016-01-07
initial step at not allowing / in collection name

jamesgorrie:
Reason:
```
path: List("guide", "this that")
pathId: "guide/this+that"
search: ~"guide/this+that"
```

```
path: List("guide", "this/that")
pathId: "guide/this%2Fthat"
search: ~"guide/this%2Fthat"
```

My suggestion is that we don't allow slashes in the collection name. Every other tree system I've looked at seems to use this, bar mac, that does a `replace("/", ":")`. 

We could be intelligent and if someone does input `this/that` we just create it as a `this` with child of `that`.

Another option would be to keep it as is and have the `media-api` encode collection queries, but then we've landed up with the encoding/decoding being across multiple systems. We would also still have the problem of searching for `List("guide", "this that", "this/that")`.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/Collection.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/Collection.scala
@@ -8,9 +8,11 @@ import play.api.libs.functional.syntax._
 
 import com.gu.mediaservice.lib.collections.CollectionsManager
 import com.gu.mediaservice.lib.argo.model.{Action, EmbeddedEntity}
+import com.gu.mediaservice.lib.net.URI.encode
 
 case class Collection(path: List[String], actionData: ActionData) {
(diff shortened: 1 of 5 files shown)


common-lib/src/main/scala/com/gu/mediaservice/model/Collection.scala
```diff
@@ -23,7 +25,7 @@ object Collection {
   implicit val formats: Format[Collection] = Format(reads, writes)
 
   def imageUri(rootUri: String, imageId: String, c: Collection) =
-    URI.create(s"$rootUri/images/$imageId/${c.pathId}")
+    URI.create(s"$rootUri/images/$imageId/${encode(c.pathId)}")
```
theefer: Does that encode slashes too?

jamesgorrie:
Closing for #1607

PR #1607, opened 2016-01-07
Slash in collection name v2

jamesgorrie:
No more encoding, and disallowing slashes for now, we can figure out what we do with the delimiter later.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
@@ -35,4 +35,7 @@ object CollectionsManager {
         col.path == collection.path && col.actionData.date.isAfter(collection.actionData.date)
       }}
     }
+
+  // We could use `ValidationNel`s here, but that's overkill
+  def isValidPathBit(s: String) = if (s.contains(delimiter)) false else true
 }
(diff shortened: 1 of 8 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
```diff
@@ -38,4 +35,7 @@ object CollectionsManager {
         col.path == collection.path && col.actionData.date.isAfter(collection.actionData.date)
       }}
     }
+
+  // We could use `ValidationNel`s here, but that's overkill
+  def isValidPathBit(s: String) = if (s.contains(delimiter)) false else true
```
theefer: Did you mean `!`?

theefer:
:+1:

common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
jamesgorrie: Nope, if it contains the delimiter, it's invalid.

theefer: `= ! s.contains(delimiter))`?

paperboyo:
Oh, cool! What about `,` that is used to separate multiple labels, `~` if it stays for collections (I like it) and `#`?

PR #1608, opened 2016-01-07
Slash in collection name v2+ui

jamesgorrie:
When you have a slash in the name:
collections-error

### DIFF
--- a/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
+++ b/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
@@ -43,7 +43,12 @@ grCollectionsPanel.controller('GrNodeCtrl',
     const ctrl = this;
     ctrl.saving = false;
     ctrl.deletable = false;
-    ctrl.addChild = childName => collections.addChildTo(ctrl.node, childName);
+    ctrl.formError = null;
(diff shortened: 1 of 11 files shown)


kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -43,7 +43,12 @@ grCollectionsPanel.controller('GrNodeCtrl',
     const ctrl = this;
     ctrl.saving = false;
     ctrl.deletable = false;
-    ctrl.addChild = childName => collections.addChildTo(ctrl.node, childName);
+    ctrl.formError = null;
+    ctrl.addChild = childName =>
```
theefer: Maybe it's just me, but I would reserve `{}`-free arrow functions to one-liners. Possibly reminds me too much of Coffee:hankey:Script :mask:

theefer:
:+1:

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
jamesgorrie: Changed, the reason I find it odd is that when we use return, the indentation is always harder to work out.

PR #1610, opened 2016-01-07
userMetadataLastModified into image root

NickPapacostas:
This is so that if we add similar fields to exports and usages (which are arrays not objects) we can be consistent


akash1810:
Not sure an export can/will be modified?

NickPapacostas:
@theefer  @akash1810  I think the idea was to put it on "exports" generally e.g. "exportsLastModified"

theefer:
:+1: and sorry for the persistent nitpicking. ES data is one of these things that is quite painful to fix after the fact :cry:

NickPapacostas:
no worries! better to get it right the first time :)

PR #1611, opened 2016-01-07
Collections from dynamo

jamesgorrie:
I'm using the same mechanism we use on each image i.e. one big collections store.

I don't think it'll get unwieldy, and I'll do some testing, but it does fix the problem that we can have more people playing with it as even 1 person has hit problems with the consistency of the previous store being in memory.

This could also be useful as Dynamo has some nifty stuff around consistent read/write.


theefer:
I hadn't realised it was going to be stored as a single row containing the whole tree/list. Not sure what problems this may cause, if any, so maybe it's OK, just a bit odd. Maybe worth a chat tomorrow w @kenoir too?

collections/app/store/CollectionsStore.scala
```diff
@@ -1,50 +1,40 @@
...
+  def remove(collectionPath: String): Future[List[Collection]] = {
+    dynamo.listGet[Collection](rootId, key) flatMap { collections =>
+      val path = CollectionsManager.uriToPath(collectionPath)
 
-        store.putData(Json.toJson(newCollectionsList))
-        Future.successful(oldCollection)
-      } getOrElse Future.failed(InvalidCollectionJson(json))
+      CollectionsManager.findIndex(path, collections) map { index =>
+        dynamo.listRemoveIndex[Collection](rootId, key, index)
+      } getOrElse Future.failed(CollectionNotFound(path))
```
theefer: This is error prone because there is no locking between the read and the write (delete), so indexes may have shifted. If we're going to use DynamoDB to make things concurrency-safe, we should look for a way to make all updates atomic.

jamesgorrie:
Closing for #1613

PR #1612, opened 2016-01-07
make collections search case insensitive

jamesgorrie:
We're trying to not make the same mistake we made with labels.

Now, we will always analyse and search in lowercase, but I have added a `description` to the collections to allow for the user to chose their desired caps.

In short, if an image is in a collection `g2/art/paint`, `G2/art/PaiNt` or any other derivative would match.

If someone recreated your collection e.g. `g2/art/PaiNT`, the description will then appear as such.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/Collection.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/Collection.scala
@@ -9,16 +9,20 @@ import play.api.libs.functional.syntax._
 import com.gu.mediaservice.lib.collections.CollectionsManager
 import com.gu.mediaservice.lib.argo.model.{Action, EmbeddedEntity}
 
-case class Collection(path: List[String], actionData: ActionData) {
+import scalaz.NonEmptyList
+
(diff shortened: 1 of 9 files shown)


common-lib/src/main/scala/com/gu/mediaservice/model/Collection.scala
```diff
@@ -32,6 +36,14 @@ object Collection {
       Action("remove", uri, "DELETE")
     ))
   }
+
+  // We use this to ensure we are creating valid `Collection`s
+  def create(path: List[String], actionData: ActionData) = {
```
akash1810: we use `build` elsewhere for these types of helper functions - not sure if there's a particular convention to follow?

akash1810:
looks good :+1:

common-lib/src/main/scala/com/gu/mediaservice/model/Collection.scala
jamesgorrie: Happy to go with `build`, it seems common across the scalasphere.

PR #1613, opened 2016-01-07
Collections from dynamo rows

jamesgorrie:
Another version of #1611. This one is probably better.

I was trying to work it out and it seems a collection is about 185 bytes (with all of DynamoDBs meta information attached), if we're excessive and say 250 bytes - that gives us 4000 before we hit 1mb.

@kenoir and I were talking about the ability to add another key on the top level collection later so we could query on those if this gets slow / impossible.


collections/app/store/CollectionsStore.scala
```diff
@@ -11,43 +11,26 @@ import scala.concurrent.ExecutionContext.Implicits.global
...
-    store.getData flatMap { json =>
-      val collectionList = json.asOpt[List[Collection]]
-      val newCollectionList = collectionList.map(CollectionsManager.add(collection, _).sortBy(_.pathId))
-
-      newCollectionList.map { collections =>
-        store.putData(Json.toJson(collections))
-        Future.successful(collection)
-      } getOrElse Future.failed(InvalidCollectionJson(json))
-    }
+    dynamo.objPut(collection.pathId, "collection", collection) map (c => c)
```
theefer: What does that `map` do?

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
```diff
@@ -183,6 +187,32 @@ class DynamoDB(credentials: AWSCredentials, region: Region, tableName: String) {
       id, s"REMOVE $key[$index]"
     ) map(j => (j \ key).as[List[T]])
 
+  def objPut[T](id: String, key:String, value: T)
+                 (implicit ex: ExecutionContext, wjs: Writes[T], rjs: Reads[T]): Future[T] = Future {
+
+    val item = new Item().withPrimaryKey(IdKey, id).withJSON(key, Json.toJson(value).toString)
+
+    val spec = new PutItemSpec().withItem(item).withReturnValues(ReturnValue.ALL_OLD)
```
theefer: Do we need `withReturnValues` if we ignore it?

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
```diff
@@ -183,6 +187,32 @@ class DynamoDB(credentials: AWSCredentials, region: Region, tableName: String) {
...
+    // all we care about is whether it completed.
+  } map (outcome => value)
+
+  def objDelete[T](id: String, key:String, value: T)
+               (implicit ex: ExecutionContext, wjs: Writes[T], rjs: Reads[T]): Future[T] = Future {
+
+    val item = new Item().withPrimaryKey(IdKey, id).withJSON(key, Json.toJson(value).toString)
+
+    val spec = new PutItemSpec().withItem(item).withReturnValues(ReturnValue.ALL_OLD)
+    table.putItem(spec)
```
theefer: Should that be a delete instead?

theefer:
Looks generally sane, and as discussed could introduce a cache on reads (to be invalidated on writes + broadcast to other servers, or with some timeout). A few comments above.

collections/app/store/CollectionsStore.scala
jamesgorrie: :question:

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
jamesgorrie: Nope. Good catch.

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
jamesgorrie: The method wasn't needed as we're using [`deleteItem`](https://github.com/guardian/grid/pull/1613/files#diff-407e9be250074544f6fc2a1f642a4727R63) instead, I suppose I just didn't finish it.

theefer:
Nice :+1:

PR #1614, opened 2016-01-08
Drop into collections fix

jamesgorrie:
A few small fixes that I noticed as single image drop wasn't working:
* Fix logic to work with single image drop data
* Make drop data with multiple images consistent with single image
* Make the saving state consistent for adding images to collections through drop or click

add-collection


kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -138,48 +138,37 @@ grCollectionsPanel.directive('grDropIntoCollection',
...
-                            element.removeClass(classComplete);
-                        }, 500);
+                        scope.dropIntoCollectionSaving = false;
                     });
                 }
-                element.removeClass(classDrag);
+                element.removeClass(classOver);
             });
 
             element.on('dragover', () => {
```
theefer: Should it be drag enter or drag over?

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -138,48 +138,37 @@ grCollectionsPanel.directive('grDropIntoCollection',
...
+                if (gridImagesData !== '' || gridImageData !== '') {
                     const imagesData = gridImagesData !== '' ?
                         JSON.parse(gridImagesData) : [JSON.parse(gridImageData)];
 
-                    const imageIds = imagesData.map(imageJson => imageJson.id);
-
-                    // TODO: Find a better way of dealing with this state than classnames
-                    element.addClass(classSaving);
+                    const imageIds = imagesData.map(imageJson => imageJson.data.id);
+                    scope.dropIntoCollectionSaving = true;
```
theefer: Watch out with scope inheritance (just investigated a Composer bug due to use of scope in a non-scope-isolated directive with Rob yesterday). What scope is this on and does it risk to leak to siblings or children also observing it?

theefer:
:+1: though a small warning re use of scope.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
jamesgorrie: I suppose dragOver would make more sense.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
jamesgorrie: It's in the scope of the element that is the directive, it's not from the parent.

It's also only controlled from this directive, and read from within the element that has this as an attribute, thus the template / element for this directive.

I can't see any leak there, unless I am being :see_no_evil:

PR #1615, opened 2016-01-08
usagesLastModified field in ES, SQS message, and thrall consumption

NickPapacostas:


common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -202,6 +202,7 @@ object Mappings {
           "uploadInfo" -> uploadInfoMapping,
           "suggestMetadataCredit" -> simpleSuggester,
           "usages" -> usagesMapping,
+          "usagesLastModified" -> dateFormat,
```
jamesgorrie: Why not keep this in the usages object?

NickPapacostas: We had a similar discussion around userMetadata, the idea is that usages and exports become arrays and we can't add fields to them. In the case of userMetadata it actually *is* an object so I had added it within (see https://github.com/guardian/grid/commit/a8944edff0998045b7b7c89de29eab83ef3afb98)

but for consistency between the three (exports, usages, and usermetadata) we opted to put it at the root level. Open to other approaches though :)

theefer:
Nice :+1:

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
jamesgorrie: Not without completely smashing the data structure... :+1:

theefer:
:+

PR #1616, opened 2016-01-11
Add source to screengrab

jamesgorrie:
We still need to think about what we do with past information without this.
There are 1995 images. This will at least stop people carry on without adding it.


theefer:
How does it look on the frontend?

jamesgorrie:
Free text field.
source-on-screengrab

theefer:
:+1: 

So existing Screengrab usage will be missing source, but if someone opens the URE for it, they will have to provide the source to save?

jamesgorrie:
Yes, that's exactly right, we could do what we did before and let people know they need to add it, which worked quite well.

This is a stopgap to stop the problem getting worse.

theefer:
OK :+1:

theefer:
Ready to ship?

PR #1617, opened 2016-01-11
freeze JSPM version

jamesgorrie:
[Seems something broke](https://github.com/jspm/jspm-cli/issues/1411). It's stopping deployment, so freezing for now.


akash1810:
:+1:

PR #1619, opened 2016-01-11
Tree rehash

jamesgorrie:
The tree logic was fundamentally flawed, especially when getting to deeper than 3 level (I was doing a crappy if test).

This has been thoroughly tested, and works in all circumstances.

This will break the client until there is a refresh, but as no-one is using it, it's fine.

@tsop14 We'll need to update your work once merged.


theefer:
If you think it works :smiley:  :+1:

PR #1620, opened 2016-01-12
delete image from collection

tsop14:
Adds ability to remove an image from a collection from the image page. 
Updates UI in same way as removing labels - polls API until results match expectations, then broadcasts an event. 

remove-image-from-collection


kahuna/public/js/services/api/collections-api.js
```diff
@@ -64,6 +66,41 @@ apiServices.factory('collections', ['mediaApi', function (mediaApi) {
...
+        return angular.equals(
+            collectionsA.sort(),
+            collectionsB.sort()
+        );
+    }
+
+    function getCollectionsIds(imageCollections) {
+        const collectionsIds = [];
+        imageCollections.data.forEach(e => collectionsIds.push(e.pathId));
+        return collectionsIds;
```
jamesgorrie: We could use a `map` here:
```JS
return imageCollections.data.map(col => col.pathId);
```

kahuna/public/js/services/image-accessor.js
```diff
@@ -52,6 +52,18 @@ imageAccessor.factory('imageAccessor', function() {
         return userMetadata.data.archived.data;
     }
 
+    function readCollections(image) {
+        return image.data.collections;
+    }
+
+    function getCollectionsIds(image) {
```
jamesgorrie: [See here](https://github.com/guardian/grid/pull/1620/files#r49436910)

kahuna/public/js/services/api/collections-api.js
```diff
@@ -64,6 +66,41 @@ apiServices.factory('collections', ['mediaApi', function (mediaApi) {
...
+    }
+
+    function untilCollectionsEqual(image, expectedCollections) {
+        return image.get().then(apiImage => {
```
jamesgorrie: I know this seems to be the de facto way of doing this - but collections don't really need to poll as they don't affect image state - so I wonder if there is another way to update the collections.

Saying that, maybe it's worth leaving and doing in one refactor when we get to it @theefer?

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
```diff
@@ -202,6 +202,27 @@
     </div>
 </div>
 
+<div class="image-info" ng:if="ctrl.image.data.collections.length > 0">
+    <dl class="image-info__group ">
+        <dt class="metadata-line__key">Collections</dt>
+        <dd class="metadata-line__info flex-container"
+            ng-repeat="collection in ctrl.image.data.collections">
+            <a ui:sref="search.results({query: '~'+collection.data.pathId})">
+                {{collection.data.path.join(' ▸ ')}}
```
jamesgorrie: Are we sticking with this for now?

tsop14: Do you mean having it shown? Or the format of something?

kahuna/public/js/services/api/collections-api.js
theefer: I'm guilty of having suggested that to @tsop14. It does affect the image state doesn't it, insofar as it requires updating the list of collections found on the image?

I'm going to look into the long-fabulated way to do this more properly without read-after-write and annoying polling, but I think this is the simplest way for now. If you have any other thoughts or suggestions, please share!

kahuna/public/js/services/api/collections-api.js
```diff
@@ -64,6 +66,41 @@ apiServices.factory('collections', ['mediaApi', function (mediaApi) {
...
+                return apiImage;
+            } else {
+                return $q.reject();
+            }
+        });
+    }
+
+    function removeImageFromCollection(collection, image) {
+        return collection.perform('remove')
+            .then(newImageCollections => apiPoll(() =>
```
theefer: Does removing one collection return the list of all the remaining collections?

tsop14: Its returns the collections resource.

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
jamesgorrie: Just the arrow - I don't mind, I just can't remember.

kahuna/public/js/services/api/collections-api.js
jamesgorrie: [It returns the new resource of collections](https://github.com/guardian/grid/blob/master/collections/app/controllers/ImageCollectionsController.scala#L54)

theefer: OK :+1:

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
```diff
@@ -125,6 +127,12 @@ module.controller('grImageMetadataCtrl', [
                 });
         };
 
+        ctrl.removeImageFromCollection = function (collection) {
+            ctrl.removingCollection = collection;
+            collections.removeImageFromCollection(collection, ctrl.image)
+            .then(() => ctrl.removingCollection = false);
```
jamesgorrie: Indent for clarity?

jamesgorrie:
:

jamesgorrie:
:+1:

tsop14:
@jamesgorrie would you mind checking this again? Thanks.

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
```diff
@@ -125,6 +136,12 @@ module.controller('grImageMetadataCtrl', [
                 });
         };
 
+        ctrl.removeImageFromCollection = function (collection) {
```
jamesgorrie: You could use `ctrl.removeImageFromCollection = collection =>  {`

jamesgorrie:
:+1:

PR #1621, opened 2016-01-12
Additional optimisations to thumbnail generation and quality

theefer:
Will comment inline.

Merges into the existing thumbnail PR.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
@@ -112,8 +112,8 @@ object ImageOperations {
       converted   = applyOutputProfile(corrected, optimised = true)
       stripped    = stripMeta(converted)
       profiled    = applyOutputProfile(stripped, optimised = true)
-      resized     = resize(profiled)(width)
-      unsharpened = unsharp(resized)(thumbUnsharpRadius, thumbUnsharpSigma, thumbUnsharpAmount, thumbUnsharpThreshold)
+      thumbnailed = thumbnail(profiled)(width)
+      unsharpened = unsharp(thumbnailed)(thumbUnsharpRadius, thumbUnsharpSigma, thumbUnsharpAmount, thumbUnsharpThreshold)
       addOutput   = addDestImage(unsharpened)(outputFile)
       _          <- runConvertCmd(addOutput)
     } yield outputFile
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
@@ -102,7 +102,6 @@ object ImageOperations {
   val thumbUnsharpRadius = 0.5d
   val thumbUnsharpSigma = 0.5d
   val thumbUnsharpAmount = 0.8d
-  val thumbUnsharpThreshold = 0.02d
   def createThumbnail(sourceFile: File, width: Int, qual: Double = 100d, tempDir: File, iccColourSpace: Option[String], colourModel: Option[String]): Future[File] = {
     for {
       outputFile <- createTempFile(s"thumb-", ".jpg", tempDir)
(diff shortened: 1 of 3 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
```diff
@@ -102,19 +102,18 @@ object ImageOperations {
...
   val thumbUnsharpSigma = 0.5d
   val thumbUnsharpAmount = 0.8d
-  val thumbUnsharpThreshold = 0.02d
   def createThumbnail(sourceFile: File, width: Int, qual: Double = 100d, tempDir: File, iccColourSpace: Option[String], colourModel: Option[String]): Future[File] = {
     for {
       outputFile <- createTempFile(s"thumb-", ".jpg", tempDir)
       cropSource  = addImage(sourceFile)
-      qualified   = quality(cropSource)(qual)
-      corrected   = correctColour(qualified)(iccColourSpace, colourModel)
+      thumbnailed = thumbnail(cropSource)(width)
```
theefer: `thumbnail` is significantly faster than `resize`, although not quite as good in terms of quality, but good enough for thumbnails.

common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
```diff
@@ -102,19 +102,18 @@ object ImageOperations {
...
+      corrected   = correctColour(thumbnailed)(iccColourSpace, colourModel)
       converted   = applyOutputProfile(corrected, optimised = true)
       stripped    = stripMeta(converted)
       profiled    = applyOutputProfile(stripped, optimised = true)
-      resized     = resize(profiled)(width)
-      unsharpened = unsharp(resized)(thumbUnsharpRadius, thumbUnsharpSigma, thumbUnsharpAmount, thumbUnsharpThreshold)
-      addOutput   = addDestImage(unsharpened)(outputFile)
+      unsharpened = unsharp(profiled)(thumbUnsharpRadius, thumbUnsharpSigma, thumbUnsharpAmount)
```
theefer: Sharpening significantly improves the look of thumbnails.

common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
theefer: Now sizing down *before* doing any of the expensive operations, especially colourspace conversion, to make them cheaper (work on less data).

image-loader/app/lib/Config.scala
```diff
@@ -26,7 +26,7 @@ object Config extends CommonPlayAppProperties with CommonPlayAppConfig {
   val tempDir: File = new File(properties.getOrElse("upload.tmp.dir", "/tmp"))
 
   val thumbWidth: Int = 256
-  val thumbQuality: Double = 75d // out of 100
+  val thumbQuality: Double = 85d // out of 100
```
theefer: Raised quality to reduce artifacts. This is offset by other optimisations, esp stripping metadata and injecting much smaller profile data, which significantly reduces the size even with raised quality.

jamesgorrie:
Out of interest - what savings are we making (filesize and speed)?

common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
jamesgorrie: `unsharped` what a silly name.

jamesgorrie:
Looks good! :+1:

theefer:
Speed-wise in terms of latency of image loader, seems very similar to the current performance:

image
(Before 10am is using the new processing from this branch, after 10am is back to the old/current processing)

Filesize I don't have aggregate numbers, but looks around 20 KB on average for the old/current system vs 12 KB for the new system, but with a much larger variance in the current system (likely because we left the profile data in, which can be large).

common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
paperboyo: > `unsharped` what a silly name.

[Mixes in a blurred original](https://en.wikipedia.org/wiki/Unsharp_masking), that's why...

PR #1623, opened 2016-01-13
add public domain usage rights

jamesgorrie:
As requested by Jo B.
Approved by Robert H and added to The Spreadsheet :pray:.
I quite like the word bequeathed.


common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
```diff
@@ -1,5 +1,6 @@
 package com.gu.mediaservice.model
 
+import ch.qos.logback.core.pattern.util.RestrictedEscapeUtil
```
akash1810: what does this do?

jamesgorrie: No idea. Gone.

akash1810:
:+1:

PR #1625, opened 2016-01-13
Css colours for collections

jamesgorrie:
I've added a CSS colour to the node and the idea of `CollectionGroup`s to the image response.

The problem is that users want to know that an image has been added to a collection, but also a hint at which collection. From chatting to others, and conversations @tsop14 has had with others, having an indication of the top level collection would be enough.

i.e. culture doesn't want to use something used by Observer.

We could potentially label them with some sort of abbreviation, similar to Google's monograms or Github's Identicons. 

As mentioned in the comment, the colour generator is crap, but it'd be good to see if adding this level of identification addresses the concerns of the users.

collection-colours

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
@@ -52,5 +52,5 @@ object CollectionsManager {
 
   def getCollectionColour(s: String) = collectionColours.get(s)
 
-  def getCssColour(path: List[String]): String = path.headOption.flatMap(getCollectionColour).getOrElse("#333333")
+  def getCssColour(path: List[String]) = path.headOption.flatMap(getCollectionColour)
 }
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
@@ -43,14 +43,14 @@ object CollectionsManager {
   // We could use `ValidationNel`s here, but that's overkill
   def isValidPathBit(s: String) = if (s.contains(delimiter)) false else true
 
-  // TODO: Find something that works off a palette, this is terrible.
-  def stringToCssRgb(s: String): String = {
-    val hash = s.hashCode
-    val r = (hash & 0xFF0000) >> 16
-    val g = (hash & 0x00FF00) >> 8
-    val b = hash & 0x0000FF
-    s"rgb($r, $g, $b)"
-  }
+  val collectionColours = Map(
+    "g2"           -> "#000000",
+    "observer"     -> "#006f94",
+    "culture"      -> "#d1008b",
+    "film & music" -> "#b1532f"
+  )
 
-  def getCssColour(path: List[String]): String = path.headOption.map(stringToCssRgb).getOrElse("#cccccc")
+  def getCollectionColour(s: String) = collectionColours.get(s)
+
+  def getCssColour(path: List[String]): String = path.headOption.flatMap(getCollectionColour).getOrElse("#cccccc")
 }
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
@@ -1,7 +1,11 @@
 package com.gu.mediaservice.lib.collections
 
+import java.awt.Color
+
 import com.gu.mediaservice.lib.net.URI.{encode, decode}
 import com.gu.mediaservice.model.Collection
+
+import scala.util.Try
 
 object CollectionsManager {
   val delimiter = "/"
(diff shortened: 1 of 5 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
```diff
@@ -38,4 +42,15 @@ object CollectionsManager {
...
   // We could use `ValidationNel`s here, but that's overkill
   def isValidPathBit(s: String) = if (s.contains(delimiter)) false else true
+
+  // TODO: Find something that works off a palette, this is terrible.
+  def stringToCssRgb(s: String): String = {
+    val hash = s.hashCode
+    val r = (hash & 0xFF0000) >> 16
+    val g = (hash & 0x00FF00) >> 8
+    val b = hash & 0x0000FF
+    s"rgb($r, $g, $b)"
```
theefer: Hardcoding colours would probably better so we can pick colours that work well together and avoid the "angry fruit salad" syndrom. Just talked to @tsop14 about it and she's going to be looking into it.

media-api/app/lib/ImageResponse.scala
```diff
@@ -230,8 +231,18 @@ object ImageResponse extends EditsResponse {
     (__ \ "usages").write[UsagesEntity]
       .contramap(usagesEntity(id, _: List[Usage])) ~
     (__ \ "collections").write[List[EmbeddedEntity[Collection]]]
-      .contramap((collections: List[Collection]) => collections.map(c => collectionsEntity(id, c)))
-  )(unlift(Image.unapply))
+      .contramap((collections: List[Collection]) => collections.map(c => collectionsEntity(id, c))) ~
+    (__ \ "collectionGroups").write[List[CollectionGroup]]
```
theefer: Not sure this is going to be necessary if we try to show a hint of the top-level collection for each leaf collection (rather than grouping per top-level collection), something @tsop14 is looking into now.

jamesgorrie: Just spoke to @tsop14 - it's be good to have a chat, as, as far as I was aware, the most important thing from a users perspective was to know that another group was using the image, not really what leaf it was in as filtering, rather than scanning, has become ichiban.

common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
jamesgorrie: Yea, this was more a fallback for if we didn't have a colour for a word. I will find out a few colours and put them in.

media-api/app/lib/ImageResponse.scala
jamesgorrie: I've updated it so the CSS colour just comes in the response now.

common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
jamesgorrie: I've updated it with 4 colours of the people using it - they're from the Guardian palette.
As mentioned, we can shift this into some configuration to keep it useful to the open.

common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
```diff
@@ -38,4 +42,15 @@ object CollectionsManager {
 
   // We could use `ValidationNel`s here, but that's overkill
   def isValidPathBit(s: String) = if (s.contains(delimiter)) false else true
+
+  val collectionColours = Map(
+    "g2"           -> "#000000",
+    "observer"     -> "#006f94",
+    "culture"      -> "#d1008b",
+    "film & music" -> "#b1532f"
```
theefer: Does it work even though it's title case in PROD?

jamesgorrie: `path` and `pathId` are always in lowercase. `description` can be in WobblyCase.
[See here for more](https://github.com/guardian/grid/pull/1612)

theefer: Oh I didn't realise or forgot. :+1:

jamesgorrie:
Updated and ready for :eyes:

common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
```diff
@@ -38,4 +42,15 @@ object CollectionsManager {
...
+  )
+
+  def getCollectionColour(s: String) = collectionColours.get(s)
+
+  def getCssColour(path: List[String]): String = path.headOption.flatMap(getCollectionColour).getOrElse("#cccccc")
```
akash1810: minor: would it make more sense for the client to set the default colour?

akash1810: though it would probably make the client side logic more complicated...

jamesgorrie: I prefer to keep the client stupid - if we make this configurable later, having it on the client would be annoying.

theefer:
:+1:

PR #1626, opened 2016-01-13
use UsageRights.all to work out free categories

jamesgorrie:
Slightly nicer way of doing it and also adds public domain, which I forgot.

### DIFF
--- a/media-api/app/lib/elasticsearch/SearchFilters.scala
+++ b/media-api/app/lib/elasticsearch/SearchFilters.scala
@@ -43,26 +43,8 @@ trait SearchFilters extends ImageFields {
   val freeFilter = filterOrFilter(freeSupplierFilter, freeUsageRightsFilter)
   val nonFreeFilter = freeFilter.map(filters.not)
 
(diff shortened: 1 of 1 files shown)


media-api/app/lib/elasticsearch/SearchFilters.scala
```diff
@@ -43,26 +43,8 @@ trait SearchFilters extends ImageFields {
...
-    CommissionedAgency.category,
-    StaffPhotographer.category,
-    ContractPhotographer.category,
-    CommissionedPhotographer.category,
-    ContractIllustrator.category,
-    CommissionedIllustrator.category,
-    Composite.category
-  )
+  lazy val freeToUseCategories: List[String] =
+    UsageRights.all.filter(ur => ur.defaultCost.contains(Free)).map(ur => ur.category)
```
akash1810: nice

akash1810:
:+1:

media-api/app/lib/elasticsearch/SearchFilters.scala
jamesgorrie: λ

PR #1627, opened 2016-01-13
fix regression for dropping mulitple images to collection

jamesgorrie:
A regression from when we were making the multi-image json the same as the single image json.


jamesgorrie:
Boken in master so merging

PR #1629, opened 2016-01-13
better collection ordering

jamesgorrie:
Obviously don't want to order by the path head, more the name of the folder.
Also learnt you can have multiple conditions with tuple instead of multiple `sortBy`s.

### DIFF
--- a/collections/app/model/Node.scala
+++ b/collections/app/model/Node.scala
@@ -3,7 +3,7 @@ package model
 import play.api.libs.functional.syntax._
 import play.api.libs.json._
 
-
+// TODO: Convert fullPath to NonEmptylist
 case class Node[T](basename: String, children: List[Node[T]], fullPath: List[String], data: Option[T])
 object Node {
   def fromList[T](list: List[T], getPath: T => List[String]): Node[T] = {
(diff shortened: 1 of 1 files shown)


jamesgorrie:
Tested, merging as breaks nothing.

collections/app/model/Node.scala
```diff
@@ -22,8 +22,7 @@ object Node {
           // use the T at this level or an empty node to hold children
           Node(currentSlug, loop(children, thisFullPath), thisFullPath, thisLevel.headOption)
         }
-        .sortBy(_.fullPath.head)
-        .sortBy(_.children.isEmpty)
+        .sortBy(node => (node.children.isEmpty, node.basename))
```
theefer: Not sure I understand this..?

jamesgorrie: [I sort of kind of do](https://joelneely.wordpress.com/2008/03/29/sorting-by-schwartzian-transform-in-scala/).

In short, `sortBy` has a nice little trick built in when you sort by a tuple.
[Here's a simpler, practical example](http://stackoverflow.com/questions/10027543/how-to-sort-a-list-in-scala-by-two-fields).

If you're wandering about the actual reason for the parameters, it's sorted by if the collections has children, and then alphabetically, similar to file system UIs. So the collapsable ones are always at the top.

Slightly ironic screengrab:
nochildren

theefer: Ah OK, wasn't sure why the No children thing. Is it sorting recursively down the tree too?

jamesgorrie: It does as we're [sorting within the loop](https://github.com/guardian/grid/blob/e15305800f96db81f9295f55dc00b5e8811b8859/collections/app/model/Node.scala#L11-L26).

theefer: Ah right cool!

PR #1630, opened 2016-01-13
add panel state to collections panel

jamesgorrie:
This seemed to be the number 1 problem with collections panel.

Instead however of making it visible all the time, which could annoy news editors etc, it's now a saveable state. This way it can work for all.

### DIFF
--- a/kahuna/public/js/services/panel.js
+++ b/kahuna/public/js/services/panel.js
@@ -56,8 +56,8 @@ panelService.factory('panelService', ['storage', function (storage) {
     // different contexts. You also might not want to have the state saved.
     function setAndSaveState($scope, name, panel) {
         const storeName = `${name}PanelState`;
-        const state = storage.getItem(storeName) || {locked: false, hidden: true};
-        const sub = panel.state$.subscribe(state => storage.setItem(storeName, state));
+        const state = storage.getJs(storeName) || {locked: false, hidden: true};
+        const sub = panel.state$.subscribe(state => storage.setJs(storeName, state));
 
         panel.setHidden(state.hidden);
         panel.setLocked(state.locked);
--- a/kahuna/public/js/services/panel.js
+++ b/kahuna/public/js/services/panel.js
@@ -1,10 +1,11 @@
 import angular from 'angular';
 import Rx from 'rx';
+import '../util/storage';
 
 
-export const panelService = angular.module('kahuna.services.panel', []);
+export const panelService = angular.module('kahuna.services.panel', ['util.storage']);
 
-panelService.factory('panelService', [function () {
+panelService.factory('panelService', ['storage', function (storage) {
     function mergeState(o, n) {
         // This is to avoid us getting into the state of (hidden && locked)
         // TODO: Error on a client asking for that state, given that we don't expose a method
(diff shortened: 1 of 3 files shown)


kahuna/public/js/services/panel.js
```diff
@@ -51,8 +52,22 @@ panelService.factory('panelService', [function () {
         return newPanel(hidden, locked);
     }
 
+    // This is decoupled from the actual panels as we need the scope, which can be different in
+    // different contexts. You also might not want to have the state saved.
+    function setAndSaveState($scope, name, panel) {
+        const storeName = `${name}PanelState`;
```
jamesgorrie: I've decided to just use a key for each so when we set, we don't have to read, update, write. It's just write.

kahuna/public/js/util/storage.js
```diff
@@ -0,0 +1,24 @@
...
+    function setItem(key, val) {
+        const storeVal = typeof val === 'string' ? val : JSON.stringify(val);
+        $window.localStorage.setItem(key, storeVal);
+    }
+
+    function getItem(key) {
+        let val = $window.localStorage.getItem(key);
+        try {
+            val = JSON.parse(val);
+        } catch (_) {}
```
theefer: This will pass through any value from LS that failed to parse as JSON. Is that intended? Feels confusing for the user of this abstraction.

jamesgorrie: I wouldn't think so, as, if you stored `string` for something, that's what you would expect back.

jamesgorrie: i.e. `JSON.parse("naughty string")` throws, whereas, strangely `JSON.parse("true")` => `true`.

jamesgorrie: `undefined` would also throw, but that would leave `val = undefined`.

theefer: But if somehow the value got corrupted and you expected an Object...?

Do we need to store strings? Would make the interface simpler and more resilient only to allow Objects (incl Arrays)?

jamesgorrie: Not being able to store a string does seems like more of a bug to me, as string is a valid JS type.
Say for instance I was storing a name? Would I have to:
```JS
const n = storage.getItem('name').name;
```
?

We could log, but then if you were expecting a string, you'd get the log every time.

Ah type-unsafety - saying that we could have:
```JS
function getItem(key, isString) {}
```?

theefer: If you really want to allow any type, then it would be nice to enforce that the type you put in is the type you get out. Could do this different ways I guess (store type hint alongside the value, different `getItem` methods to retrieve different types or type argument to `getItem`, etc).

It's all a little bit superstitious, but I find the simpler the API (you know what you put in and get out), the less things tend to go wrong. In practice it doesn't seem like you're using this API to save strings now, so why generalise it to strings?

theefer: Alternatively, why not also `JSON.stringify` strings? It works, and `JSON.parse(JSON.stringify(x)) === x`.

jamesgorrie: I've been explicit in the naming of the methods, and thrown a nicer error. If we need to extend this to deal with standard items, we can.

theefer:
:+1:

PR #1631, opened 2016-01-13
[WIP] collections with colour

tsop14:
! continues work from #1625

Displays each unique collection 'base' with colour. 
On hover displays the full collection paths. 

screen shot 2016-01-13 at 18 54 21
screen shot 2016-01-13 at 18 54 32


jamesgorrie:
I know it might create a slightly longer image element, but might it be worth having these under the image as the place they are now is already cramped with icons defining state (most there are also boolean, i.e. they have usage, they have crops etc)?

kahuna/public/js/preview/image.html
```diff
@@ -75,18 +75,20 @@
...
-                <span class="titip-content preview__in-collection__list">
-                    <li ng-repeat="collection in ctrl.image.data.collections">
-                        <a ui:sref="search.results({query: '~'+collection.data.pathId})">
-                            {{collection.data.path.join(' ▸ ')}}
+            <ul ng:repeat="collectionGroup in ctrl.image.data.collectionGroups"
+                class="bottom-bar__meta-item titip-bottom preview__collections">
+                <li class="preview__collections__collection" style="background-color: {{collectionGroup.cssColour}}">
+                    {{collectionGroup.basename}}
+                </li>
+                <ul class="titip-content preview__collections__list">
```
akash1810: Don't think a `ul` can have a `ul` as a child, it should be wrapped in an `li` ([reference](https://developer.mozilla.org/en/docs/Web/HTML/Element/ul)).

jamesgorrie:
@tsop14 I've merge my changes in and made the template work off the new response. Happy to run through it.

kahuna/public/js/preview/image.html
```diff
@@ -75,18 +75,20 @@
...
-                        </a>
-                        <br>
-                    </li>
-                </span>
-            </span>
+            <ul class="bottom-bar__meta-item titip-bottom preview__collections">
+                <li class="preview__collections__collection"
+                    ng:repeat="collection in ctrl.image.data.collections"
+                    style="background-color: {{collection.data.cssColour}}">
+                    <a ui:sref="search.results({query: '~'+collection.pathId})"
```
jamesgorrie: Could we run this through the [`querySyntax`](https://github.com/guardian/grid/blob/master/kahuna/public/js/search-query/query-syntax.js#L26-L28) helper?

theefer: Alternatively there's also a filter to help with quoting.

jamesgorrie: Filters still always scare me given how often they run... add a log into one and you'll see what I mean. As it's a static value I find `ctrl.collectionLink(collection)` works well.

theefer: Like any expression in templates, it may be re-evaluated whenever a scope is digested. I am not aware of any difference in performance between `{{ transform(value) }}` and `{{ value | transform }}`? For performance, it's more important to mark immutable expressions as one-time-bind using `::`, so either `{{:: transform(value) }}` or `{{:: value | transform }}`, which again should only run once in both cases.

I haven't checked the uiRouter code, but I suspect that in the case above, the `uiSref` attribute is merely a `&` attribute, which only gets invoked (and therefore evaluated) when triggered, in this case on click. I might be wrong though as it depends on the implementation.

I agree there's a risk of slowing down performance with expensive filters, but I think it's important we all understand why and how Angular works to avoid superstition or false optimisations.

jamesgorrie: You're right, and thanks for being [Houdini](https://en.wikipedia.org/wiki/Harry_Houdini#Debunking_spiritualists) here. I've added the filter and we're using that.

kahuna/public/js/search/query-filter.js
```diff
@@ -26,3 +28,10 @@ queryFilters.filter('queryLabelFilter', function() {
         }
     };
 });
+
+queryFilters.filter('queryCollectionFilter', function() {
+    return path => {
+        if (path[0] === 'family') {console.log('filter', path);}
```
theefer: sad

jamesgorrie: [:rage4:](https://github.com/guardian/grid/commit/a08c2197479e2bb3be4cfd3c78f4fb35f33b867a)

theefer:
:+1:

tsop14:
* Collections on same row as labels.
* Background colour determined by root.
* Text is the collection leaf. :leaves: 
* Default colour (i.e. if no colour set) is grey.
* Changes colour on hover and reveals tooltip. 
* Links to the collection

In order to display the tooltip we currently use (css) you cannot have `overflow:hidden` on a parent div.

Because I *think* it is unlikely (for now) that images will have a lot of collections, the tooltip will be useful for first time users and will reduce confusion with labels, tooltip trumps overflow. 

This means an image with lots of collections will have them run over, as in image below. They wont go over to another image tile but will stick over the edge.

Thoughts?

screen shot 2016-01-15 at 18 36 37

kahuna/public/js/preview/image.html
```diff
@@ -46,6 +46,19 @@
 
     <div class="preview__info" ng:if="! ctrl.hideInfo">
         <div class="flex-container">
+            <ul class="bottom-bar__meta-item preview__collections">
+                <li class="preview__collections__collection"
+                    ng:repeat="collection in ctrl.image.data.collections"
+                    gr-tooltip="Click to open collection: {{collection.data.path.join(' ▸ ')}}"
+                    gr-tooltip-position="top">
+                    <a ui:sref="search.results({query: '~'+collection.data.pathId})"
```
theefer: Should this use your filter helper instead?

kahuna/public/js/preview/image.html
```diff
@@ -46,6 +46,19 @@
...
+                       style="background-color: {{collection.data.cssColour}}"}
```
theefer: What happens if missing? Invalid CSS expression?

theefer:
Shame about the overflow, but I can see why it's fiddly. Seems like it's probably worth shipping with the issue and thinking about how to solve it as a future step.

A few comments, else :+1:

kahuna/public/js/preview/image.html
jamesgorrie: Yip, must have been lost in a merge.

kahuna/public/js/preview/image.html
jamesgorrie: [Now it just leaves it blank](https://github.com/guardian/grid/pull/1631/files#diff-4c4a6be0bd67994ff7b2c4d740e00078R71).

PR #1632, opened 2016-01-14
fix date picker

akash1810:
before
screen shot 2016-01-14 at 10 57 08

after
screen shot 2016-01-14 at 10 56 37

(using `flex-box` would be better...)


theefer:
:+1: 

Separately, would also be nice to fix the ugly margins below Presets and Custom date range

kenoir:
Nice.

PR #1633, opened 2016-01-14
track calendar usage

jamesgorrie:
My general feeling is that people don't really use the calendar - it'd be good to see if they do, and if so, how.

### DIFF
--- a/kahuna/public/js/components/gu-date-range/gu-date-range.html
+++ b/kahuna/public/js/components/gu-date-range/gu-date-range.html
@@ -1,5 +1,8 @@
 <div class="gu-date-range" ng:init="ctrl.showOverlay = false">
-    <div class="gu-date-range__display" ng:click="ctrl.showOverlay = !ctrl.showOverlay">
+    <button class="gu-date-range__display"
+            ng:click="ctrl.showOverlay = !ctrl.showOverlay"
+            gr:track-click="{{ctrl.trackingName}}"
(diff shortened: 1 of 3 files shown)


kahuna/public/js/components/gu-date-range/gu-date-range.html
```diff
@@ -1,5 +1,8 @@
 <div class="gu-date-range" ng:init="ctrl.showOverlay = false">
-    <div class="gu-date-range__display" ng:click="ctrl.showOverlay = !ctrl.showOverlay">
+    <button class="gu-date-range__display"
```
theefer: `button` = thanks :octopus:

theefer:
:+1:

PS: I use it :-)

paperboyo:
Anecdata: I rarely do. And whenever I try, I spend considerable time setting a simple range. Every time I have to learn again if I set the start on the left or on the right (why some months/years are unavailable in the dropdown or greyed out) and why do I have to care for it and it doesn't work both ways. There are some minor glitches here as well: https://github.com/guardian/grid/issues/672.

akash1810:
@paperboyo would date filter search syntax be helpful at all?

paperboyo:
@akash1810 Maybe it should be there one day for consistency and ease of use (chips?). But I'm okay using a picker randomly, just so I can quickly modify a URL path for now. Admittedly, I'm not a huge date ranges user. Thx!

PR #1634, opened 2016-01-14
update saving on collections panel

jamesgorrie:
We need to run this in the apply to update the template. At the moment it just hangs saying "Adding..."

### DIFF
--- a/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
+++ b/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
@@ -114,7 +114,9 @@ grCollectionsPanel.controller('GrNodeCtrl',
     };
 
     subscribe$($scope, pathWithImages$, ({path, images}) => {
-       collections.addImagesToCollection(images, path).then(() => ctrl.saving = false);
+       collections.addImagesToCollection(images, path).then(() => {
+           $scope.$apply(() => ctrl.saving = false);
+       });
     });
 
     inject$($scope, hasImagesSelected$, ctrl, 'hasImagesSelected');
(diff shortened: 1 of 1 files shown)


kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -114,7 +114,9 @@ grCollectionsPanel.controller('GrNodeCtrl',
     };
 
     subscribe$($scope, pathWithImages$, ({path, images}) => {
-       collections.addImagesToCollection(images, path).then(() => ctrl.saving = false);
+       collections.addImagesToCollection(images, path).then(() => {
+           $scope.$apply(() => ctrl.saving = false);
```
theefer: I don't understand why we need an `$apply` here, as `$http` and `$q` are supposed to do that for you?

jamesgorrie: I think, potentially it's because we're in the subscribe?

jamesgorrie: And if we're in the subscribe, potentially out of angular's digest scope.
Hypothesis, but hard to test.

jamesgorrie: P.S. It does work though.

theefer: Good point, that may be it as we don't `$apply` in the [`subscribe$` helper](https://github.com/guardian/grid/blob/6bd20d7a73a8626acd3a9d9ab7e18568d866aa3a/kahuna/public/js/util/rx.js#L30-L35), even though I would still have expected it to apply through the use of `$http` and `$q`. Might be worth trying adding the `$apply` there to fix the more general problem (though such a change would need thorough QA as it has wide implications).

theefer: Ah no I know why, it's because we use [`Promise.all`](https://github.com/guardian/grid/blob/81516b64647dd8d47d63aded85f4339836a6cd37/kahuna/public/js/services/api/collections-api.js#L64). We should instead use `$q.all`.

jamesgorrie:
Closing in favour of #1636. Thanks @theefer for noticing that this wasn't the problem.

PR #1635, opened 2016-01-14
fix slight regression in calendar button styling

jamesgorrie:
Slight regression by turning it into a button.
Back to normal (and less floats).
anytimebutton


theefer:
(Checked on smaller breakpoints?)

:+1:

jamesgorrie:
Checked.

PR #1636, opened 2016-01-14
use  > Promise to stay in the digestive system

jamesgorrie:
Use `$q` over promise to keep digesting.
In favour of #1634.

### DIFF
--- a/kahuna/public/js/services/api/collections-api.js
+++ b/kahuna/public/js/services/api/collections-api.js
@@ -1,7 +1,7 @@
 import angular from 'angular';
 import apiServices from '../api';
 
-apiServices.factory('collections', ['mediaApi', function (mediaApi) {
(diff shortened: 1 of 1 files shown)


kahuna/public/js/services/api/collections-api.js
```diff
@@ -53,7 +53,7 @@ apiServices.factory('collections', ['mediaApi', function (mediaApi) {
         const promises = imageIds.map(id => mediaApi.find(id).then(
                 image => image.perform('add-collection', {body: {data: path}})));
 
-        return Promise.all(promises);
```
theefer: There's another one below

jamesgorrie: So there ~~is~~ was.

theefer:
:+1:

PR #1637, opened 2016-01-14
No promises

theefer:
Don't use Promise, use Angular's $q instead.

Noticed a few more which needed replacing. I wonder if that's related to the Witness import being occasionally flaky. Maybe not.


jamesgorrie:
Was just about to do this, nice catch :+1:

PR #1638, opened 2016-01-14
show icon once image added to collection

tsop14:
User sees little folder icon (or whatever imagery we use in the future) once an image has been added to a new collection. 

adding-images


kahuna/public/js/services/api/collections-api.js
```diff
@@ -60,13 +60,37 @@ apiServices.factory('collections',
...
     }
 
+    function getCollectionsX(apiImageResource) {
+        return apiImageResource.data.collections.map(e => e.data.pathId);
+    }
+
+    function untilNewCollectionAppears(image, collectionAdded) {
+        return image.get().then( (apiImage) => {
+            const apiCollections = getCollectionsX(apiImage);
+            if(apiCollections.indexOf(collectionAdded.data.pathId) > -1) {
```
theefer: Nitpick mode: space after `if`.

Also: can we have `Array.prototype.include` already ;_;

kahuna/public/js/services/api/collections-api.js
```diff
@@ -60,13 +60,37 @@ apiServices.factory('collections',
...
+                $rootScope.$emit('image-updated', newImage, image);
+                return newImage;
+            })
+
         ).toJS();
 
         return $q.all(promises);
     }
 
+    function getCollectionsX(apiImageResource) {
```
theefer: Is the name deliberate?

theefer:
Minor comments, functionally :+1:

kahuna/public/js/services/api/collections-api.js
jamesgorrie: :boom:Just tried it in chrome console, it seems to work:boom:

theefer: Maybe a bit too [edgy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#Browser_compatibility), esp as Traceur doesn't seem to translate it...?

kahuna/public/js/services/api/collections-api.js
```diff
@@ -60,13 +60,37 @@ apiServices.factory('collections',
...
     function addImagesToCollection(images, path) {
+
         const promises = images.map(image =>
             image.perform('add-collection', {body: {data: path}})
+            .then(collectionAdded => apiPoll(() =>
+                 untilNewCollectionAppears(image, collectionAdded)
+            ))
+            .then(newImage => {
+                $rootScope.$emit('image-updated', newImage, image);
+                return newImage;
```
akash1810: is the returned value used?

theefer:
Is this little icon still the plan, vs the coloured leaf labels discussed yesterday?

tsop14:
> Is this little icon still the plan, vs the coloured leaf labels discussed yesterday?

Little icon is on the out but it wasn't created in this PR. Going to change to coloured leaf in another PR soon.

kahuna/public/js/services/api/collections-api.js
theefer: Makes sense not to swallow the value unexpectedly imho.

theefer:
OK cool!

PR #1639, opened 2016-01-14
First pass of refactoring Angular deps to use mod.name to avoid duplication

theefer:
Discovered this way of doing things in a sample Angular project and it feels nicer to me, using the module.name of imported models rather than treating imports purely as side-effects and duplicating the module name manually. Also stricter with each file being its own module, to make it easier to reason about.

Haven't converted the whole codebase yet, only things referred to by main.js (and fixed anything that broke).


jamesgorrie:
Nice! That'll make it work a lot better with IDEs too (I can actually see where the dep comes from).
:+1:

kahuna/public/js/services/api/loader.js
```diff
@@ -1,8 +1,11 @@
-import apiServices from '../api';
+import angular from 'angular';
+import {mediaApi} from './media-api';
 
-apiServices.factory('loaderApi',
-                    ['mediaApi',
-                     function(mediaApi) {
+export var loaderApi = angular.module('kahuna.services.api.loader', [
```
jamesgorrie: You could change these to `const`s

kahuna/public/js/crop/index.js
```diff
@@ -2,13 +2,15 @@ import angular from 'angular';
...
+import '../directives/ui-crop-box';
 
 import cropTemplate from './view.html!text';
 
 
 export var crop = angular.module('kahuna.crop', [
     'kahuna.crop.controller',
-    'gr.topBar'
+    'gr.topBar',
+    'ui.cropBox'
```
akash1810: how/why was this module working before w/out this?

kahuna/public/js/services/api/loader.js
theefer: I thought about that but stuck to `var` for consistency with the rest of the codebase. We can do a general replace throughout as a separate step.

I'm not sure what it means for an export to be `var` or `const`, and specifically what happens to the exported value if you reassign it later..

kahuna/public/js/crop/index.js
theefer: It was imported in `main.js`, I moved it to be more local.

akash1810: Ah, must have missed that. :+1: for local imports!

PR #1640, opened 2016-01-14
delete all indexes of collections in dynamo

jamesgorrie:
As we just add to the list of collections in dynamo (we can have two entries for one collection) - we need to make sure we delete all references to the path - this does that.


jamesgorrie:
@tsop14 this should stop the bug you were having of having to hit delete multiple times.

theefer:
Any way we can also prevent duplicates when adding multiple times?

theefer:
:+1:

jamesgorrie:
> Any way we can also prevent duplicates when adding multiple times?

It was a decision to just keep all of them - which can be useful with something like an audit trail.

We could:
`Read => override duplicate with new `actionData` => write`, but I didn't see the need to read before as we could just use Dynamos `list_append` syntax.

theefer:
:+1: OK as not exposed through ES

PR #1641, opened 2016-01-15
[Proof] slightly disorganised attempt at one state

jamesgorrie:
Just a proof of concept really, or experiment.
The instig/inspir-ation is the heavily requested feature of next / previous on the image.

I'll make notes inline.

The most interesting points are how we deal with app state, I [make a note of it here](https://github.com/guardian/grid/pull/1641/files#r49831719).

### DIFF
--- a/kahuna/public/js/image/index.js
+++ b/kahuna/public/js/image/index.js
@@ -1,5 +1,7 @@
 import angular from 'angular';
 
+import {indexRoute} from '../routes/index';
+
 import './controller';
 import '../search/query-filter';
 import '../imgops/service';
@@ -15,14 +17,15 @@ export var image = angular.module('kahuna.image', [
     'kahuna.search.filters.query',
     'kahuna.imgops',
     'gr.topBar',
-    'grXeditable'
+    'grXeditable',
+    indexRoute.name
 ]);
 
 
 image.config(['$stateProvider',
               function($stateProvider) {
 
-    $stateProvider.state('image', {
+    $stateProvider.state('gr.image', {
         url: '/images/:imageId?crop',
         template: imageTemplate,
         controller: 'ImageCtrl',
--- a/kahuna/public/js/routes/index.js
+++ b/kahuna/public/js/routes/index.js
@@ -1,0 +1,45 @@
+import angular from 'angular';
+import {searchQueryService} from '../search-query/service';
+
+// Use the name once that PR in merged in
+import '../services/api/media-api';
+
+export const indexRoute = angular.module('gr.routes.index', []);
+
+indexRoute.config(['$stateProvider', '$urlMatcherFactoryProvider',
+    function($stateProvider, $urlMatcherFactoryProvider) {
+
+        const zeroWidthSpace = /[\u200B]/g;
+        function removeUtf8SpecialChars(val) {
+            return angular.isDefined(val) && val.replace(zeroWidthSpace, '');
+        }
+        $urlMatcherFactoryProvider.type('Query', {
+            encode: val => removeUtf8SpecialChars(val),
+            decode: val => removeUtf8SpecialChars(val),
+            //call decode value that includes zero-width-space character
+            is: val => angular.isDefined(val) && !zeroWidthSpace.test(val)
+        });
+
+        $stateProvider.state('gr', {
+            url: '?{query:Query}&ids&since&nonFree&uploadedBy&until&orderBy',
+            abstract: true,
+            template: `<ui-view />`,
+            resolve: {
+                searchResults$: ['searchQueryService', 'mediaApi', function(searchQueryService, mediaApi) {
+                    const searchResults$ = searchQueryService.state$.flatMap(params =>
+                        Rx.Observable.fromPromise(mediaApi.search(params.get('query'), params.delete('query').toJS()))
+                    );
+
+                    return searchResults$;
+                }]
+            }
+        })
+}]);
+
+
+
+indexRoute.run(['$rootScope', '$state', function($rootScope, $state) {
+    $rootScope.$on('$stateChangeError', function (event, toState, toParams, fromState, fromParams, error) {
+        console.log(error)
+    });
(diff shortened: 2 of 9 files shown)


kahuna/public/js/image/index.js
```diff
@@ -1,5 +1,7 @@
 import angular from 'angular';
 
+import {indexRoute} from '../routes/index';
```
jamesgorrie: This isn't really the way to do routes.
My personal preference would be [to have them as a separate module](https://github.com/guardian/grid/tree/4ac0c221d20ea0143c6c4c7ae4f918b693617e59/kahuna/public/js/routes) so as not to couple them to anything.

kahuna/public/js/routes/index.js
```diff
@@ -0,0 +1,45 @@
...
+            return angular.isDefined(val) && val.replace(zeroWidthSpace, '');
+        }
+        $urlMatcherFactoryProvider.type('Query', {
+            encode: val => removeUtf8SpecialChars(val),
+            decode: val => removeUtf8SpecialChars(val),
+            //call decode value that includes zero-width-space character
+            is: val => angular.isDefined(val) && !zeroWidthSpace.test(val)
+        });
+
+        $stateProvider.state('gr', {
```
jamesgorrie: Having this one root route allows as to carry the stateParams across to other routes.
At the moment this is just between `search` and `image`.

kahuna/public/js/routes/index.js
```diff
@@ -0,0 +1,45 @@
...
+            url: '?{query:Query}&ids&since&nonFree&uploadedBy&until&orderBy',
+            abstract: true,
+            template: `<ui-view />`,
+            resolve: {
+                searchResults$: ['searchQueryService', 'mediaApi', function(searchQueryService, mediaApi) {
```
jamesgorrie: The more I work with this, the more I think that these shouldn't be in the route resolvers, but i there own state services.

The reason being - the dependency is then available to the route controller, but any other controller within that can't include it as a dependency.

A good example is the [`selectedImages$`](https://github.com/guardian/grid/blob/master/kahuna/public/js/search/index.js#L142), which [child views of `search.results`](https://github.com/guardian/grid/blob/master/kahuna/public/js/search/index.js#L161-L229) [can access](https://github.com/guardian/grid/blob/master/kahuna/public/js/search/index.js#L161-L229) but controllers within that can't, so you have to pass it down in a [chain like fashion](https://github.com/guardian/grid/blob/master/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js#L150). 

I think what would be more beneficial is have a `app-state/selected-images` service, which could then be used from across the app.




https://github.com/guardian/grid/blob/master/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js#L60-L61

kahuna/public/js/search-query/service.js
```diff
@@ -0,0 +1,25 @@
...
+    function searchParams($state, $stateParams, mapFactory) {
+        // TODO: Think about restricting these to:
+        // ["ids", "archived", "nonFree", "uploadedBy", "until", "since", "offset",
+        //  "length", "orderBy", "query"]
+        // TODO: Make this start with the $stateParams
+        const params = mapFactory();
+
+        // As we want to have the side effect or a URL change happen irrespective of whether
+        // we have subscription, we setup a `do` and create a hot observable to do so.
+        const stateChange$ = params.state$.do(params => {
```
jamesgorrie: Just an interesting way to deal with the state, so that, firstly, we don't have to use `$stateParams` anywhere, but also so that we never have to reload state.

kahuna/public/js/search/query.js
```diff
@@ -83,9 +94,13 @@ query.controller('SearchQueryCtrl',
 
     $scope.$watchCollection(() => ctrl.filter, onValChange(filter => {
         filter.uploadedBy = filter.uploadedByMe ? ctrl.user.email : undefined;
-        $state.go('search.results', filter);
+        //$state.go('gr.search.results', filter);
```
jamesgorrie: This, at the moment doesn't filter down to `results` as results requires a reload to take affect, which is massively inefficient.

kahuna/public/js/search/results.js
```diff
@@ -83,10 +86,15 @@ results.controller('SearchResultsCtrl', [
              panels,
              range,
              isReloadingPreviousSearch,
-             collectionsEnabled) {
+             collectionsEnabled,
+             searchResults$) {
 
         const ctrl = this;
 
+        const totalResults$ = searchResults$.map(results => results.total);
```
jamesgorrie: Just a test.

kahuna/public/js/search-query/service.js
```diff
@@ -0,0 +1,25 @@
+import angular from 'angular';
+import {mapFactory} from '../lib/data-structure/map-factory';
+
+export const searchQueryService = angular.module('searchQuery.service', [
+    mapFactory.name
+]);
+
+searchQueryService.factory('searchQueryService', ['$state', '$stateParams', 'mapFactory',
```
jamesgorrie: This is the other option to what is spoken about in [routes holding app state](https://github.com/guardian/grid/pull/1641/files#r49831719). Here a service holds the state, which then allows [embedded controllers to access it as a dependency](https://github.com/guardian/grid/pull/1641/files#diff-32fea343d36b21c11521e08ea0059cafR37).

kahuna/public/js/routes/index.js
jamesgorrie: I also the wonder about the testability of these dependencies?

kenoir:
@jamesgorrie going to close this but keep the branch around for reference.

paperboyo:
One state, one love. One day.

jamesgorrie:
@paperboyo we're about to start working on search for wellcometrust/wellcomecollection.org - and I am glad we have this insight.

PR #1642, opened 2016-01-15
make the button a measly button

jamesgorrie:
As noticed by @theefer

### DIFF
--- a/kahuna/public/js/components/gu-date-range/gu-date-range.html
+++ b/kahuna/public/js/components/gu-date-range/gu-date-range.html
@@ -1,5 +1,6 @@
 <div class="gu-date-range" ng:init="ctrl.showOverlay = false">
-    <button class="gu-date-range__display"
+    <button type="button"
+            class="gu-date-range__display"
             ng:click="ctrl.showOverlay = !ctrl.showOverlay"
             gr:track-click="{{ctrl.trackingName}}"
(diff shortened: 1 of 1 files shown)


kahuna/public/js/components/gu-date-range/gu-date-range.html
```diff
@@ -21,6 +21,7 @@ <h2 class="gu-date-range__overlay__title search__overlay__title">Presets</h2>
         <div class="gu-date-range__overlay__presets">
             <button type="button"
                     class="gu-date-range__overlay__preset__button button"
+                    type="button"
```
theefer: There is already a `type="button"` above.

paperboyo:
A platonic :+1:

kahuna/public/js/components/gu-date-range/gu-date-range.html
theefer: I think you meant to add this to the first `button` in this file instead.

jamesgorrie: Dammit, thanks :+1: to peer review and git history rewrite.

jamesgorrie:
@paperboyo almost right.

theefer:
:+1:

theefer:
:eyes:

PR #1645, opened 2016-01-15
HOver over to show 'add to collection' button

jamesgorrie:
Seen a few people think they're in editing mode, also removes clutter.

# Before
addtocolbefore

# After
addtocolafter

### DIFF
--- a/kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
+++ b/kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
@@ -1,4 +1,4 @@
-<div ng:init="node = ctrl.node">
+<div ng:init="node = ctrl.node" class="hover-parent">
     <div class="node__info flex-container"
          ng:class="{'node__info--selected' : ctrl.isSelected}"
          gr:drop-into-collection="node.data.fullPath">
@@ -43,7 +43,7 @@
         <div class="text-small flex-container flex-center"
              ng:if="ctrl.saving || dropIntoCollectionSaving">Adding…</div>
 
-        <button class="node__action clickable"
+        <button class="node__action clickable hover-child"
                 ng:if="!ctrl.editing && ctrl.hasImagesSelected && !ctrl.saving"
                 ng:click="ctrl.addImagesToCollection()"
                 gr:tooltip="Add selected images to this collection">
--- a/kahuna/public/stylesheets/main.css
+++ b/kahuna/public/stylesheets/main.css
@@ -385,6 +385,12 @@ button[disabled],
     color: white;
 }
 
+.hover-parent .hover-child {
+    display: none;
+}
+.hover-parent:hover .hover-child {
+    display: initial;
+}
 
 /* ==========================================================================
    Angular


tsop14:
Will first time user realise that there is the option to add like this (as well as dnd) if there is no add button before they hover over?

jamesgorrie:
Nope, but we can tell them.

tsop14:
👌

👍

theefer:
Out of interest, what happens if you click on the collection instead of the [+]]?

jamesgorrie:
When you're not in edit mode (the screengrabs above) - you navigate to the collection (expected, similar to changing the query when you have images selected).

When you're in edit mode, the same thing, potentially not expected, but seems best to leave it the same for simplicity's sake.

theefer:
OK. :+1:

Seems OK for now, but it may make sense in the future to disable the search controls as soon as you have a selection, and possibly the collection navigation too, since it doesn't feel completely expected that you'd then be losing your selection.

jamesgorrie:
> disable the search controls

Yea, we did speak about it and went against it. I personally like being able to get out of selection by just carrying on, but that's subjective.

And because UI PRs need a Google Photos reference, they're quite smart and just cover up the old controls with the now-relevant controls:
google-controls

theefer:
Yes, I really like Google Photos' solution: both makes it very clear you're in a different mode, and hide any irrelevant options in that mode. We talked about it but didn't go for that in the end because it was a little more complex.

paperboyo:
> disable the search controls as soon as you have a selection, and possibly the collection navigation too

Although it may lead to some confusion(*), I would much prefer selection surviving all searching/filtering. More specifically here, I would also prefer to be able to navigate the tree _after_ making a selection. Otherwise, if forgotten, one would have to start over (bad :cat:). 

> you're in a different mode

I don't like this idea that this is a "different mode". I don't find these options irrelevant. I often select something and then apply a filter(s) to, effectively, XOR through UI.

But then: I'm still afraid to upload anything to Eric's Photos.

(*)problems arise when some currently selected photos would fall out of the current filter... answer: they should be deselected and stay that way even if the filter widens back to "allow" them again

theefer:
> I would much prefer selection surviving all searching/filtering. 

Do you have any example of apps that allow that? I think it ends up quite confusing because some of the images you selected wouldn't be visible anymore -- are they still selected? if I revert the filter change, are they selected again?

jamesgorrie:
> I would much prefer selection surviving all searching/filtering

If I select 5 images, search, and now only 2 are selected from my last search => I now apply metadata in batch, what would you expect this to apply to? This wouldn't be very explicit, but something people have to learn as people will expect different things.

> they should be deselected and stay that way even if the filter widens back to "allow" them again

This again is not an easily expected behaviour, more learned.

theefer:
Sorry you answered my scenario in your (*). Still find it needlessly confusing or, a contrario, a find Google Photos particularly clear and simple in that regard.

jamesgorrie:
The joy of hiding the ability to search, and thus clear your selection is the only way to clear your selection is saying "clear my selection". :+1: to explicit actions.

paperboyo:
> Do you have any example of apps that allow that?

Sure thing. Use them all the time. If I know that my resultant selection should be a combination of several search/filter devices scattered all over the interface, that behaviour is the only way for me to avoid writing complex expressions into a searchbox (or making one selection through UI, copying search query, pasting into text editor, clearing selection, repeating the above and then glueing the wanted search queries back together with ORs and ANDs and pasting that in searchbox).

> clear and simple

I would say: limited. You know: for cats and stuff.

> This again is not an easily expected behaviour, more learned.

Agreed. But logical and infinitely more powerful. I can imagine a hint showing "x of y became deselected because they fall out of focus".

Modality is for phones ;-)

paperboyo:
Wanted to add one thing if I may. For me it's not about two ways of doing something where I'm attached to only one of them.

This is about one way not offering what I can easily (could be easier still - Lightroom is shit with `NOT` through UI) accomplish with only one of them. And it's that functionality I'm attached to.

If the modal way would let me achieve the same - welcome, Google!

PR #1646, opened 2016-01-15
Add usageStatus and usagePlatform media-api query params

theefer:
Allows to filter by usage information, e.g.

`/images?orderBy=-usages.lastModified&usageStatus=pending,published&usagePlatform=digital`
(note that custom ordering is already supported)

cc @kenoir


jamesgorrie:
Simple, nice :+1:

PR #1647, opened 2016-01-15
[wip]tooltips behave like title attributes

akash1810:
Hide tooltip after 1.5s rather than persisting (and possibly getting in the way!)

titip

### DIFF
--- a/kahuna/public/js/components/gr-tooltip/gr-tooltip.js
+++ b/kahuna/public/js/components/gr-tooltip/gr-tooltip.js
@@ -20,13 +20,18 @@ tooltip.directive('grTooltip', ['$timeout', 'onValChange', function ($timeout, o
             }
 
             element.bind('mouseenter', () => {
-                element.addClass('titip-default')
-                    .addClass(`titip-${position}`);
+                $scope.addTimer = $timeout(() => {
+                    element.addClass('titip-default').addClass(`titip-${position}`);
 
-                $timeout(() => {
-                    element.removeClass('titip-default')
-                        .removeClass(`titip-${position}`);
+                    $scope.removeTimer = $timeout(() => {
+                        element.removeClass('titip-default').removeClass(`titip-${position}`);
+                    }, 3500);
                 }, 1500);
+            });
+
+            $scope.$on('$destroy', function() {
+                $timeout.cancel($scope.removeTimer);
+                $timeout.cancel($scope.addTimer);
             });
         }
     };
--- a/kahuna/public/js/components/gr-tooltip/gr-tooltip.js
+++ b/kahuna/public/js/components/gr-tooltip/gr-tooltip.js
@@ -3,15 +3,13 @@ import 'titip';
 
 export const tooltip = angular.module('grTooltip', []);
 
-tooltip.directive('grTooltip', ['onValChange', function (onValChange) {
+tooltip.directive('grTooltip', ['$timeout', 'onValChange', function ($timeout, onValChange) {
     return {
         restrict: 'A',
         link: function ($scope, element, attrs) {
             const position = attrs.grTooltipPosition || 'bottom';
 
-            element.attr('data-title', attrs.grTooltip)
-                .addClass(`titip-default`)
-                .addClass(`titip-${position}`);
+            element.attr('data-title', attrs.grTooltip);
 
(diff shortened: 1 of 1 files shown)


kahuna/public/js/components/gr-tooltip/gr-tooltip.js
```diff
@@ -20,6 +18,16 @@ tooltip.directive('grTooltip', ['onValChange', function (onValChange) {
                     element.attr('data-title', newTooltip);
                 }));
             }
+
+            element.bind('mouseenter', () => {
+                element.addClass('titip-default')
```
jamesgorrie: Strange indenting...

theefer: Is it?

kahuna/public/js/components/gr-tooltip/gr-tooltip.js
```diff
@@ -20,6 +18,16 @@ tooltip.directive('grTooltip', ['onValChange', function (onValChange) {
...
+                    .addClass(`titip-${position}`);
+
+                $timeout(() => {
+                    element.removeClass('titip-default')
+                        .removeClass(`titip-${position}`);
+                }, 1500);
```
theefer: You may need to cancel the `$timeout` if the directive gets `$destroy`ed as the DOM may have been torn down, in which case not sure if the above will work or not?

jamesgorrie:
That's not actually how titles behave, the remain as long as you hover over them.
Which makes sense for long title (e.g. image description) where this behaviour would be really annoying.

A potential implementation could be similar to title of only __showing__ it after you've `mouseover`ed for 500ms or so.

kahuna/public/js/components/gr-tooltip/gr-tooltip.js
akash1810: maybe the second `.addClass` should be directly below the first, rather than starting at the last `e` of `element` (same for `removeClass` below?

akash1810:
> That's not actually how titles behave, the remain as long as you hover over them.

You're right - there is a delay on showing and a much longer delay to hide (wait for it...). This behaviour seems better... though I'll reduce the hide timeout.

title

kahuna/public/js/components/gr-tooltip/gr-tooltip.js
akash1810: good point

kahuna/public/js/components/gr-tooltip/gr-tooltip.js
```diff
@@ -20,6 +18,21 @@ tooltip.directive('grTooltip', ['onValChange', function (onValChange) {
                     element.attr('data-title', newTooltip);
                 }));
             }
+
+            element.bind('mouseenter', () => {
+                $scope.addTimer = $timeout(() => {
```
akash1810: adding this to `$scope` is a bit :mask:

jamesgorrie: Why not just set it to a `const`?

theefer: Why delay the display of the tooltip? I thought one of the main advantages of this library was to be able to show them without delay, unike `title` attributes...?

akash1810: its not available for cancelling below as its block scoped

kahuna/public/js/components/gr-tooltip/gr-tooltip.js
```diff
@@ -20,6 +18,21 @@ tooltip.directive('grTooltip', ['onValChange', function (onValChange) {
...
+                    element.addClass('titip-default').addClass(`titip-${position}`);
+
+                    $scope.removeTimer = $timeout(() => {
+                        element.removeClass('titip-default').removeClass(`titip-${position}`);
+                    }, 3500);
```
theefer: If several `mouseenter` end up fired, you will only have a reference to the last timeouts. You won't be able to cancel any previous pending timeout.

This seems to be getting a little complicated -- is there a genuine need for hiding the tooltip after a delay? :)

akash1810:
> is there a genuine need for hiding the tooltip after a delay?

closing to re-think

PR #1649, opened 2016-01-15
hide search bar with selection actions

jamesgorrie:
Just a trial to see what we think - feels like a good idea to abstract this out into a component anyway.
hidetopbar


jamesgorrie:
A different colour toolbar might be nice.

theefer:
Nice! As you said it may benefit from a stronger treatment/colour to emphasise that (1) something changed, (2) you're in a different mode. The Clear selection may also have to be emphasised somehow?

Cool stuff!

jamesgorrie:
Blue Skies:
bluesky

akash1810:
YES! :heart_eyes_cat: 

"newly crowed my ugly"?!?! harsh!

jamesgorrie:
> "newly crowed mr ugly"?!?! harsh!

Not my words, it was Getty.

paperboyo:
Votum separatum, FWIW (also discussion in https://github.com/guardian/grid/pull/1645).

You've hidden some options that (only currently, I hope) do not work with selections but clear them (presumably in search of ["explicit actions"](https://github.com/guardian/grid/pull/1645#issuecomment-171990828)), but you've left others that act similarly like e.g. Collections tree (unrelated is the fact that now I'm also two clicks away from Upload when in selection mode...). There is no way to keep all navigation options hidden when in selection mode not only because there will be no space in the toolbar soon. Take the Collections tree which is both the navigation device AND metadata modifier. Same with byline, credit and other fields (they apply, but - brilliantly - they are also links). This alleged simplicity is already falling apart. And good riddance, too :)

Take a look here (poor Finder has no invert selection command, so the usefulness seems limited):
finder

Here is Lightroom's Edit menu. I could imagine links that are metadata fields to realise this with two modifier keys (because with just one they would add/substract from the current search instead ;-) ):
selection-menu

Having all searching/filtering options available and affecting selections is indefinitely more flexible. While it's true that if one plans carefully, one could first apply search and filters and only then make a selection. But what if all that effort was only to find stuff I do not want and now I would like my selection inverted to have selected what I was looking for in the first place?

Or let's suppose I searched for a label "Super-mega-great", but I would like to see adjacent photos that these were selected from. I could select all and cancel the search. If the selection would survive that search relaxation, I could just browse and see "Super-mega-great" in their context (current date ingested sorting order notwithstanding).

And one day to realise copy/synchronise all/parts of metadata we will need something akin to active photo **within** a selection - see [from 2:06](http://tv.adobe.com/watch/the-complete-picture-with-julieanne-kost/mastering-the-mysterious-multiselect-in-lightroom/).

Shame, 'cause I like the blue colour. It calms me.

I know, I know, nec Hercules contra plures :-P

jamesgorrie:
@paperboyo I think the main thing here is whether we want to turn the results page into an image management page - which we would do by having a fancy selection tool. As much as that sounds exciting, I'm not sure that the way the users will use, or want to use, the selection tool.

The intersecting is interesting - but could definitely be achieved by creating your selection by adding images to the collection your interested in and comparing them to other selections, I can see this being massively useful when shortlisting images for a gallery.

Basically I think the selection / filtering ideas are great, but they are a little to transient, especially as the picture desks can often be collaborative - so perhaps collection management would be more useful and valuable (as well as persistent and shareable).

jamesgorrie:
In that respect, the ordering of images in a collection would also be useful, as picture editors would like to be able to see their images as a story, and how it flows.

paperboyo:
> whether we want to turn the results page into an image management page

That's a fair point and there is a balance to strike. But I do not see a contradiction in _management_ vs. _searching&finding_. Nor do I see it between having good filtering mechanisms and clever collection management. They reinforce each other and should be able to interact.

> In that respect, the ordering of images in a collection would also be useful, as picture editors would like to be able to see their images as a story, and how it flows.

User sort order? Yeah! In addition to be able to sort by any (sane) metadata ;-)

kenoir:
Late to the :tada: but this I like this.

theefer:
> In that respect, the ordering of images in a collection would also be useful, as picture editors would like to be able to see their images as a story, and how it flows.

Unless this is better left for the "shelf" concept.

jamesgorrie:
> Unless this is better left for the "shelf" concept.

My thoughts exactly.

kahuna/public/js/search/index.js
```diff
@@ -162,6 +162,77 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
...
+                    fixed="true"
+                    ng:if="ctrl.active">
+                    <gr-top-bar-nav>
+                        <button
+                            class="top-bar-item clickable side-padded"
+                            type="button"
+                            ng:click="ctrl.clearSelection()">
+                            <gr-icon-label gr-icon="clear">Clear selection</gr-icon-label>
+                        </button>
+                        <div class="top-bar-item clickable side-padded">
```
tsop14: Remove `clickable`?

theefer:
Could we remove the Home/Grid icon and the vertical dots/more menu from that overlay? doesn't feel like it belongs to these contextual actions for your selection.

As previously mentioned, the Clear button could be stronger (and we should also allow using ESC to cancel the selection), and perhaps the number of selected images too. Hover and disabled states for buttons have a weird styling.

I kind of like the Edit hint, though it feels somewhat weird as it doubles up with the tab below, and "Close" feels superfluous.

Overall it seems like a nice step forward though imho!

paperboyo:
Currently, Download works with just one image selected - here it's greyed (blueed) out then.

I would be ever-so-slightly scared of ESC cancelling a selection. A carefully crafted one could disappear if I would mistakenly double-tap ESC while editing a label for a selection. But I do not feel strongly here.

What about adding Create full frames to the RHS or is it potentially too costly? (Still think we will run out of toolbar space and then the modal state is going to fall apart... :smiling_imp:).

jamesgorrie:
> remove the Home/Grid icon and the vertical dots/more

Done - much better
no-context-button


> Clear button could be stronger... perhaps the number of selected images too.

I thought potentially too large, but the more I look at it the more it says, "you're in selection mode".
mega-buttons
> Hover and disabled states for buttons have a weird styling.

I kind of like the Edit hint, though it feels somewhat weird as it doubles up with the tab below, and "Close" feels superfluous.

Overall it seems like a nice step forward though imho!

theefer:
Cool! Not sure I understood what your conclusion was about the Edit button though?

jamesgorrie:
> conclusion

I'm lacking - all though an opinion is to remove it, but then what do we do about locking?
(Still opinion and having watched people use it) - it never seems that people lock it.

[To have a conclusion, I think we'll need to get this out](https://github.com/guardian/grid/pull/1680)

kahuna/public/js/search/index.js
```diff
@@ -180,6 +180,77 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
             }]
         },
         views: {
+            selectionActions: {
+                template: `
+                <gr-top-bar
```
akash1810: can we have a separate html file for the template

theefer: Arg, :+1:

theefer: One advantage of a separate file for HTML is that we HTML lint them.

theefer:
> I'm lacking - all though an opinion is to remove it, but then what do we do about locking?

I think we should rethink locking perhaps.

paperboyo:
I do not get what would happen when I press Edit. But then, I get even less how it might be connected to (panel?) locking (does it opens metadata panel?). And then

> rethink locking

:scream: (not because it works perfectly, which it doesn't)

jamesgorrie:
> rethink locking

Potentially making it the default. Did we actually get complaints about people saying, "my navigation has gone wonky as I opened the panel" or did we foresee it and make it more confusing?

jamesgorrie:
^ On that - I mean to say, we potentially start locked.

paperboyo:
> start locked

That's better than fine with me. But it's bigger than me :laughing:

I'm not a fan of recently introduced autohiding unlocked only on scroll, BTW, if there are not enough images to scroll it won't autohide at all. And in general it's not what my brain expects, I've learned, while playing with it.

jamesgorrie:
> I've learned, while playing with it.

:clap:

theefer:
I also keep getting confused by the not-hiding panel personally.

Just a reminder, the reason why we didn't lock by default was that the locked version may cause you to lose your position. It may not be as bad a problem as we thought, but good to remember why we did it in the first place when assessing whether to change.

theefer:
I'd favour removing the Edit button for now so this can be shipped without unrelated debates about the panel, which can be improved, but not necessarily as part of this.

jamesgorrie:
Panel button solution:
edit-topbar

paperboyo:
> not be as bad a problem as we thought

Unless you are in the middle of a long list and open a panel, right?

kahuna/public/js/search/index.js
jamesgorrie: I was going to do this once we were good on actually thinking it was a good idea.

paperboyo:
Oh! :scream: So Edit button is a shortcut to otherwise hidden/unavailable Show info panel button?! Then I'm all against it! Panels for me are the backbone of UI, they should be there all the time. Let me find some better arguments... or actual arguments :)

theefer:
+1 @paperboyo

jamesgorrie:
[@paperboyo](https://github.com/guardian/grid/commit/dbc119ffdc8921b13ecf0918e58263626a4fc99c)

theefer:
Downloading/Adding states when buttons clicked are still an ugly grey :cry:

kahuna/public/js/components/gr-top-bar/gr-top-bar.js
```diff
@@ -22,8 +22,11 @@ topBar.directive('grTopBarNav', [function() {
         transclude: true,
         // Annoying to have to hardcode root route here, but only
         // way I found to clear $stateParams from uiRouter...
-        template: `<a href="/search" class="home-link" title="Home">Home</a>
-                   <ng:transclude></ng:transclude>`
+        template: `<a href="/search" class="home-link" title="Home" ng:if="!hideHomeLink">Home</a>
```
theefer: Seems like a really weird concern to need in here.

paperboyo:
Thanks, God/@theefer!

Thing is, we don't know what will go into both of these panels, some stuff may be relevant to the selected photos(*). And as I wrote in https://github.com/guardian/grid/pull/1649#issuecomment-172212298 [there is something profoundly wrong with citing oneself, don't ya think?], even the Collections panel in the gif above already breaks the rule of not being able to navigate out of selection mode (can't think of why would that be useful, anyway).

(*)A lot already is: metadata in the panel may show common values. Some _may be_: Collections tree indicating if all or only some of the selected photos are in any visible folder (may be too much visual clutter and stupid - but proves the general point).

kahuna/public/js/components/gr-top-bar/gr-top-bar.js
jamesgorrie: How so - seems to make sense?
Show the search link, or dont?

jamesgorrie:
> Downloading/Adding states when buttons clicked are still an ugly grey :cry:

Removed

kahuna/public/js/components/gr-top-bar/gr-top-bar.js
theefer: Hide the Home link. Thought there may be a way to create a specialised top-bar-with-home composed from a standard vanilla top-bar, rather than having toggles on the top-bar. Doesn't really matter tbh.

tsop14:
>  Collections panel in the gif above already breaks the rule of not being able to navigate out of selection mode (can't think of why would that be useful, anyway).

Could use a collection-tree panel like the one proposed in #1781 which you can only select (in order to add images to) and cannot be used to navigate to a search?

paperboyo:
> a collection-tree panel (...) cannot be used to navigate to a search

I'm actually all **for** mixing this functionalities (applying, navigating). I am against, what @jamesgorrie called "explicit actions". I think it's much more useful if the Collections panel lets you to do both. This was just used (along with other devices from metadata panel) as an exemplification that this mixture already exist all over the app and for a good reason.

E.g. in the proposed GIF the sorting is hidden while in selection mode, while I find it useful to be able to sort then. But I'm just repeating what I said in https://github.com/guardian/grid/pull/1649#issuecomment-172212298. And I understand that to change our model of selections as modal to the one where all functionality works while in selection is a big undertaking.

So I would say just this: the less is hidden by the blue bar - the better :-)

theefer:
The point of the blue bar *is* to hide actions that are unlikely to be desired while dealing with a selection, because they don't behave as you'd expect -- mostly carefully built selection is lost, but also what happens to selected items no longer matched once the filters have been changed?

Unless this is solved or a lot more clear, which I suspect is rather hard from a tech/UX pov and of limited value, I'd go for preventing these actions. Could do more user research if people other than @paperboyo think they might actually want to do these things.

Anyway, you folks are in charge now :-)

kenoir:
Closing for now. I still think something like this is a good idea.

PR #1650, opened 2016-01-15
Remove excessive config call

jamesgorrie:
I can't think of why we do this here, when we do it [here](https://github.com/guardian/grid/blob/master/kahuna/public/js/main.js#L134).


theefer:
:+1:

jamesgorrie:
Going to wait till Monday is case there was a reason.

PR #1651, opened 2016-01-18
Remove edition, layoutId from print usage metadata

kenoir:
As edition & layoutId are inconsistent when coming from Textporter & InDesign plugin we need to remove them from the group and usage identifiers,

### DIFF
--- a/usage/app/model/UsageId.scala
+++ b/usage/app/model/UsageId.scala
@@ -14,11 +14,8 @@ object UsageId {
   def build(printUsageRecord: PrintUsageRecord) = buildId(List(
     Some(printUsageRecord.mediaId),
     Some(printUsageRecord.printUsageMetadata.pageNumber),
-    Some(printUsageRecord.printUsageMetadata.edition),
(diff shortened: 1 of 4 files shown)


usage/app/model/UsageId.scala
```diff
@@ -14,11 +14,8 @@ object UsageId {
   def build(printUsageRecord: PrintUsageRecord) = buildId(List(
     Some(printUsageRecord.mediaId),
     Some(printUsageRecord.printUsageMetadata.pageNumber),
-    Some(printUsageRecord.printUsageMetadata.edition),
-    Some(printUsageRecord.printUsageMetadata.layoutId),
-    Some(printUsageRecord.printUsageMetadata.issueDate),
     Some(printUsageRecord.printUsageMetadata.sectionCode),
-    Some(printUsageRecord.printUsageMetadata.size),
```
kenoir: Size is also not necessary as it may change between editions.

jamesgorrie:
:+1:

akash1810:
:+1:

PR #1654, opened 2016-01-18
dynamic tooltip position

akash1810:
if the panel is on the left, tooltips go on the right.
if the panel is on the right, tooltips go on the left.

before

screen shot 2016-01-18 at 13 47 56

after
screen shot 2016-01-18 at 13 47 44


theefer:
:+1:

PR #1655, opened 2016-01-18
add collections shortcut

jamesgorrie:
The hardest bit about getting people to test the collections is the bookmarklet.
The now updates localStorage by pressing `ctrl+c`. We'll remove this once we're happy to make it default.

### DIFF
--- a/kahuna/public/js/search/index.js
+++ b/kahuna/public/js/search/index.js
@@ -98,9 +98,9 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
                 description: 'Toggle collections',
                 callback: () => {
                     const toggled = !storage.getJs('showCollectionsPanel');
+                    const using = toggled ? 'start' : 'stop';
                     storage.setJs('showCollectionsPanel', toggled);
-                    $window.alert('Now refresh to start using collections');
-
+                    $window.alert(`Now refresh to ${using} using collections`);
                 }
             });
         }],
@@ -109,7 +109,7 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
                 // keep the shortcut keys here to stop overriding
                 return new Map([
                     ['metadataPanel', 'm'],
-                    ['showCollections', 'ctrl+c']
+                    ['showCollections', 'ctrl+alt+c']
                 ]);
             }],
             panels: ['panelService', function(panelService) {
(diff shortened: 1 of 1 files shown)


kahuna/public/js/search/index.js
```diff
@@ -88,11 +92,25 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
...
+                }
+            });
         }],
         resolve: {
             shortcutKeys: [function() {
                 // keep the shortcut keys here to stop overriding
-                return new Map([['metadataPanel', 'm']]);
+                return new Map([
+                    ['metadataPanel', 'm'],
+                    ['showCollections', 'ctrl+c']
```
theefer: Is there not a risk that this conflicts with the standard "Copy" shortcut (eg copying text)? Is the assumption that people are all on Mac and Ctrl+C != Cmd+C?

theefer:
Small concern about shortcut, else :+1:

kahuna/public/js/search/index.js
akash1810: [Alternative shortcut](https://en.wikipedia.org/wiki/Konami_Code).

kahuna/public/js/search/index.js
```diff
@@ -88,11 +92,25 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
...
                 callback: panels.metadataPanel.toggleHidden
             });
+
+            keyboardShortcut.bindTo($scope).add({
+                combo: shortcutKeys.get('showCollections'),
+                description: 'Toggle collections',
+                callback: () => {
+                    const toggled = !storage.getJs('showCollectionsPanel');
+                    storage.setJs('showCollectionsPanel', toggled);
+                    $window.alert('Now refresh to start using collections');
```
akash1810: any reason we don't do this automatically?

jamesgorrie: They might be on a search they don't want to lose. As this is hyper-temporary, I don't really mind asking.

akash1810: makes sense :+1:

kahuna/public/js/search/index.js
jamesgorrie: Any problems with `ctrl+alt+c`?

theefer:
:+1:

PR #1656, opened 2016-01-18
make collection colours visible by default

jamesgorrie:
Some of the feedback was that the root collection was more useful when marking the image.

As discussed with @theefer and @tsop14, having only that show might hinder more complex / useful workflows later. 

To get around this, we can make the collection colours more prominent and permanent until people realise and understand what they are.

I ran this past a couple picture editors and they thought it would be clear enough. Let's see.

collection-colours-stay


theefer:
:+1:

PR #1657, opened 2016-01-18
picdar-export: Allow overwrite for usage fetch and send

kenoir:
- Adds config parameter for overwrite "mode"
- Overwrite is implemented for usage fetch & send

Overwrite does not check if records exist before writing them!

### DIFF
--- a/picdar-export/src/main/scala/com.gu.mediaservice.picdarexport/lib/Config.scala
+++ b/picdar-export/src/main/scala/com.gu.mediaservice.picdarexport/lib/Config.scala
@@ -23,7 +23,7 @@ object Config {
   def picdarUsageApiUrl  = properties(s"picdar.usageapi.url")
 
   val defaultOverwrite   = false
-  def overwriteFlag      = Try(properties("overwrite.active").toBoolean).getOrElse[Boolean](defaultOverwrite)
+  def overwriteFlag      = Try(properties("overwrite.active").toBoolean).getOrElse(defaultOverwrite)
 
   val picdarDeskUrl      = properties("picdar.desk.url")
   val picdarDeskUsername = properties("picdar.desk.username")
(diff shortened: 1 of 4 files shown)


picdar-export/src/main/scala/com.gu.mediaservice.picdarexport/lib/Config.scala
```diff
@@ -19,6 +22,9 @@ object Config {
 
   def picdarUsageApiUrl  = properties(s"picdar.usageapi.url")
 
+  val defaultOverwrite   = false
+  def overwriteFlag      = Try(properties("overwrite.active").toBoolean).getOrElse[Boolean](defaultOverwrite)
```
jamesgorrie: Out of interest is the type needed here?

picdar-export/src/main/scala/com.gu.mediaservice.picdarexport/lib/db/ExportDynamoDB.scala
```diff
@@ -224,12 +224,18 @@ class ExportDynamoDB(credentials: AWSCredentials, region: Region, tableName: Str
     })
   }
 
-  def getNoUsage(dateRange: DateRange) =
-    getUrnsForNotFilledFields[AssetRef](dateRange, Set("picdarUsage"))(
+  def getNoUsage(dateRange: DateRange, overwrite: Boolean = false) = {
+    val notFilledSet: Set[String] = if(overwrite) Set("noop") else Set("picdarUsage")
```
jamesgorrie: Is `noop` human-understandable enough for this?

picdar-export/src/main/scala/com.gu.mediaservice.picdarexport/lib/Config.scala
kenoir: Nope. Copy paste fever.

picdar-export/src/main/scala/com.gu.mediaservice.picdarexport/lib/db/ExportDynamoDB.scala
kenoir: Probably not. Will add a comment.

PR #1659, opened 2016-01-18
Splits dateRange on load for picdar-export

kenoir:
Prevents "too many results" exception by splitting `DateRange` into day
chunks.


theefer:
:+1:

kenoir:
This actually causes time outs for longer date ranges from the picdar API so I think I prefer the more sensible deterministic error response picdar gives.

PR #1660, opened 2016-01-18
picdar-export fetch "get" rather than "scan"

kenoir:
Uses the get strategy to avoid the massive scan problem


theefer:
:+1: (tested and shown to work?)

kenoir:
@theefer updated to use the batchGet API when getting a list of URNs. Is much faster doing the initial check now.

theefer:
Nice! :+1:

PR #1661, opened 2016-01-19
Always append image id to filename in downloaded ZIP

theefer:
This helps with the case where two images have the same filename.

Sample filenames:
* 505635752-b62f9f3bca65746d2a0cc53757398029d64156af.jpg
* DV2212354-40ebd8e5e222a7b940ec5ac9aad394a9196034a4.jpg

(Are those too cryptic?)

Unfortunately, it doesn't seem to be possible to rename the file we download when a single image is downloaded, so that remains just the image id :cry: 

cc @paperboyo


theefer:
Changed naming format, bit nicer

image

akash1810:
:+1:

PR #1662, opened 2016-01-20
Make Conditional show in free to use

jamesgorrie:
As per @JonnyWeeks's email.
We should probably change this to the suggested `cost=free,conditional`.
I can do that once this fix is out.

### DIFF
--- a/media-api/app/lib/elasticsearch/SearchFilters.scala
+++ b/media-api/app/lib/elasticsearch/SearchFilters.scala
@@ -44,7 +44,7 @@ trait SearchFilters extends ImageFields {
   val nonFreeFilter = freeFilter.map(filters.not)
 
   lazy val freeToUseCategories: List[String] =
-    UsageRights.all.filter(ur => ur.defaultCost.contains(Free)).map(ur => ur.category)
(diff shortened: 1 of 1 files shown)


akash1810:
:+1:

media-api/app/lib/elasticsearch/SearchFilters.scala
```diff
@@ -44,7 +44,7 @@ trait SearchFilters extends ImageFields {
   val nonFreeFilter = freeFilter.map(filters.not)
 
   lazy val freeToUseCategories: List[String] =
-    UsageRights.all.filter(ur => ur.defaultCost.contains(Free)).map(ur => ur.category)
+    UsageRights.all.filter(ur => ur.defaultCost.exists(cost => cost == Free | cost == Conditional)).map(ur => ur.category)
```
theefer: Surprising to use `|` instead of `||` but it turns out that's correct too (though forces evaluation of all operands).

jamesgorrie: Thanks! #1666

PR #1663, opened 2016-01-20
use little CMS when installing graphicsmagick on a mac

jamesgorrie:
It's needed for encoding.


akash1810:
strange that this hadn't come up before... :+1:

PR #1664, opened 2016-01-20
add collections panel shortcut

jamesgorrie:
Seems logical.

### DIFF
--- a/kahuna/public/js/search/index.js
+++ b/kahuna/public/js/search/index.js
@@ -94,6 +94,12 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
             });
 
             keyboardShortcut.bindTo($scope).add({
+                combo: shortcutKeys.get('collectionsPanel'),
+                description: 'Toggle collections panel',
+                callback: panels.collectionsPanel.toggleHidden
+            });
+
+            keyboardShortcut.bindTo($scope).add({
                 combo: shortcutKeys.get('showCollections'),
                 description: 'Toggle collections',
                 callback: () => {
@@ -109,6 +115,7 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
                 // keep the shortcut keys here to stop overriding
                 return new Map([
                     ['metadataPanel', 'm'],
+                    ['collectionsPanel', 'c'],
                     ['showCollections', 'ctrl+alt+c']
                 ]);
             }],
--- a/kahuna/public/js/search/index.js
+++ b/kahuna/public/js/search/index.js
@@ -115,7 +115,7 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
                 // keep the shortcut keys here to stop overriding
                 return new Map([
                     ['metadataPanel', 'm'],
-                    ['collectionsPanel', 'c'],
+                    ['collectionsPanel', 'l'],
                     ['showCollections', 'ctrl+alt+c']
                 ]);
             }],


akash1810:
what's the behaviour when localStorage is empty?

kahuna/public/js/search/index.js
```diff
@@ -109,6 +115,7 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
                 // keep the shortcut keys here to stop overriding
                 return new Map([
                     ['metadataPanel', 'm'],
+                    ['collectionsPanel', 'c'],
```
theefer: Only concern is C is "crop" on the image page, and recent talks were hinting at possibly bringing that elsewhere, on the results page perhaps. Would be nice to have a letter that we could reuse elsewhere even on the image page for instance.

theefer: L?

jamesgorrie: B for buckets?

jamesgorrie:
@akash1810 not sure this touches localStorage?

kahuna/public/js/search/index.js
theefer: Can we put bucket (and basket and tag and route) behind us and not bring them back please? :pray:

jamesgorrie: :trollface:

akash1810:
> not sure this touches localStorage?

Sorry, I meant more what's the behaviour when the `showCollectionsPanel` is not set in localStorage - i.e. would the shortcut work if the collections panel hasn't been enabled?

jamesgorrie:
@akash1810 yip, that's all dealt with inside the panel controller.

kahuna/public/js/search/index.js
jamesgorrie: I like `l` -> changed.

jamesgorrie:
Anyone?

akash1810:
:+1:

theefer:
:+1:

paperboyo:
:-( I thought panel shortcuts for showing/hiding/locking should be independent of the features housed in these panels and be consistent among tools. It would still be possible to have shortcuts to specific features (would take one to the panel scrolled to that feature in case there are more features housed in the same panel).

That stems from the model of panels having features and not every feature having/being its own panel.

jamesgorrie:
> not every feature having/being its own panel.

This isn't something we have at the moment.
If the collections moves into a sub-panel (or something of the like) - it's still nice to have a shortcut no?

paperboyo:
Keyboard shortcuts = always nice :clap:

Keyboard shortcuts to specific features = very good!.

Occupying keyboard with specific shortcuts before deciding on the general ones = dangerous, prone to conflicts/reversals, no?

But, I may have (wrongly?) assumed that `l` - coming from co`l`lections - will gonna stay as a shortcut to the whole panel(*). If not - cool!

(*)I hope it doesn't come from `l`eft, because then I will have to rip out my `r` key and stick where `;` is now :laughing:

jamesgorrie:
> But, I may have (wrongly?) assumed that l - coming from collections - will gonna stay as a shortcut to the whole panel(*). If not - cool!

I wouldn't say so - it'll be linked to collections.

I was assuming, because `c`rop existed, this was co`l`lections.

paperboyo:
Sweet! :candy:

theefer:
@paperboyo may have a point on keyboard shortcut to feature vs panel. This would prevent us from making `l` a shortcut to *add to a collection* (current image, selection, whatnot), which seems a much more useful case for a shortcut than show/hide a panel. Something to think about before it gets used too much.

paperboyo:
> Something to think about before it gets used too much.

My main (only?) point!

A good model, I think, is to have them logical, e.g.:
Feature `X` > Modified feature `Shift + X` > Feature action `Ctrl + X` > Modified feature action `Ctrl + Alt + X`> Even more `Ctrl + Alt + Shift + X`. Not that I have _clear_ examples.

But before that should come general cross-tools UI convention shortcuts, I think.

Usage of Ctrl + Alt should be mindful of https://github.com/guardian/scribe/issues/448 on Windows, though...

PR #1665, opened 2016-01-20
add austraia to the colours

jamesgorrie:
From the palette.
[Here it is](http://www.color-hex.com/color/4a7834)


tsop14:
They are quite similar shades of green? How are we selected new colours now?

jamesgorrie:
I thought the same.
We're getting the colours from the old Guardian palette - I've gone with gold, as that's Australia's other colour.

PR #1667, opened 2016-01-20
Update excluded Getty collections using list from 2015-12-02

theefer:
I've manually checked all the collections that have been changed (added or removed), tidying up some of them to ensure the collection always appears as supplierCollection in the usageRights. Some of the renames appear to be correct, in that they now match things and didn't before.

Going to check with Robert re Barcroft, as we could exclude them more precisely now we're in new rights model land.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/config/UsageRightsConfig.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/config/UsageRightsConfig.scala
@@ -4,9 +4,9 @@ object UsageRightsConfig {
 
   val payGettySourceList = List(
     "Arnold Newman Collection",
-    // Note: This is not excluded as we have a deal with Barcroft directly,
-    // and there are only a tiny amount from Getty.
-    // "Barcroft Media",
+    // Note: Even though we have a deal directly with Barcroft, it
+    // does not apply to images sourced through Getty. Silly, I know.
+    "Barcroft Media",
     "Catwalking",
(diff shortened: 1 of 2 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/config/UsageRightsConfig.scala
```diff
@@ -2,24 +2,23 @@ package com.gu.mediaservice.lib.config
...
-  // Note: we filter exclusively on matching source, rather than combining credit=Getty and source=X
-  // this is assumed to be good enough as it's unlikely other provider will use the same source.
   val payGettySourceList = List(
-    // TODO: Remove once we are on the new UsageRights model
-    // This is not excluded as we have a deal with Barcroft,
-    // and there are only a tiny amount from Getty
-    // "Barcroft Media",
+    "Arnold Newman Collection",
+    // Note: Even though we have a deal directly with Barcroft, it
+    // does not apply to images sourced through Getty. Silly, I know.
```
jamesgorrie: You say silly, I say sensible.

jamesgorrie:
:+1:

common-lib/src/main/scala/com/gu/mediaservice/lib/config/UsageRightsConfig.scala
theefer: It's the same f:bird:ing image!

PR #1669, opened 2016-01-20
Remove tooltips from collections

jamesgorrie:
It's causing way more hassle than it's worth with the flow of the document as it only sets the pseudo elements to `visibility: hidden`.


theefer:
:+1: (worth keeping in mind when used/added elsewhere then)

jamesgorrie:
Indeed - as we spoke to @akash - the tooltip lib might not  be thr right one, we can grep for gr:tooltip & title

PR #1670, opened 2016-01-20
Don't 500 if removing usageRights when there was no metadata edits

theefer:
The item ended up being `null` causing a NPE. This makes things safer, though I want to review anything it may affect before shipping.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
@@ -229,7 +229,7 @@ class DynamoDB(credentials: AWSCredentials, region: Region, tableName: String) {
   }
 
   def asJsObject(outcome: UpdateItemOutcome): JsObject =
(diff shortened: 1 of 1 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
```diff
@@ -229,14 +228,8 @@ class DynamoDB(credentials: AWSCredentials, region: Region, tableName: String) {
     jsonWithNullAsEmptyString(Json.parse(item.toJSON)).as[JsObject] - IdKey
   }
 
-  def asJsObject(outcome: PutItemOutcome): JsObject =
```
jamesgorrie: We don't use these?

theefer: No (anymore?), otherwise it wouldn't compile and pass the tests :smile:

jamesgorrie:
:+1:

PR #1671, opened 2016-01-21
picadar-export: Allow properties file to be in same directory

kenoir:


kenoir:
Closing as this is not required.

PR #1673, opened 2016-01-21
No lowercase on collections

jamesgorrie:
`pathId` becomes lowercase, we respect case in all other instances as the only place we want case insensitivity is search.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
@@ -11,7 +11,7 @@ object CollectionsManager {
   val delimiter = "/"
 
   def stringToPath(s: String) = s.split(delimiter).toList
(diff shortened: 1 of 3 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
```diff
@@ -1,17 +1,14 @@
 package com.gu.mediaservice.lib.collections
 
-import java.awt.Color
```
akash1810: I guess this was intellij doing its thing and its not actually required?

tsop14:
+1  gRrRrrReaT

akash1810:
:+1:

common-lib/src/main/scala/com/gu/mediaservice/lib/collections/CollectionsManager.scala
jamesgorrie: Indeed - I added it a while ago - compilation would fail otherwise.

jamesgorrie:
:haircuT:

PR #1674, opened 2016-01-21
create endpoint to fix our collections

jamesgorrie:
The basic path is:
* Map over nodes, to create nodes with the right data
* Turn them into a list
* Get the list of futures from Dynamo
* Resolve and return

You have to do the whole tree at once as we rely on the relationships.

### DIFF
--- a/collections/app/controllers/CollectionsController.scala
+++ b/collections/app/controllers/CollectionsController.scala
@@ -86,8 +86,8 @@ object CollectionsController extends Controller with ArgoHelpers {
         CollectionsStore.add(correctedCollection)
       }
 
-      Future.sequence(futures) map { c =>
-        respond(c)
+      Future.sequence(futures) map { updatedCollectionsList =>
+        respond(updatedCollectionsList)
       }
     }
   }
--- a/collections/app/controllers/CollectionsController.scala
+++ b/collections/app/controllers/CollectionsController.scala
@@ -140,9 +140,8 @@ object CollectionsController extends Controller with ArgoHelpers {
     (__ \ "name").write[String] ~
     (__ \ "children").lazyWrite(Writes.seq[Node[EmbeddedEntity[Collection]]](collectionEntityWrites)) ~
     (__ \ "fullPath").write[List[String]] ~
-    (__ \ "correctPath").write[List[String]] ~
     (__ \ "data").writeNullable[EmbeddedEntity[Collection]]
-  )(node => (node.basename, node.children, node.fullPath, node.correctPath, node.data))
+  )(node => (node.basename, node.children, node.fullPath, node.data))
 
   type CollectionsEntity = Seq[EmbeddedEntity[Node[Collection]]]
   implicit def asArgo: Writes[Node[Collection]] = (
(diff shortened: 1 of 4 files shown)


collections/conf/routes
```diff
@@ -10,6 +10,7 @@ GET     /collections                                    controllers.CollectionsC
 POST    /collections                                    controllers.CollectionsController.addChildToRoot
 POST    /collections/*collection                        controllers.CollectionsController.addChildToCollection(collection: String)
 DELETE  /collections/*collection                        controllers.CollectionsController.removeCollection(collection: String)
+GET     /corrected-collections                          controllers.CollectionsController.correctedCollections
```
theefer: I know it's just a temporary route, but `POST` would be better (`GET` isn't supposed to have side-effects)

collections/app/controllers/CollectionsController.scala
```diff
@@ -127,9 +151,10 @@ object CollectionsController extends Controller with ArgoHelpers {
           // This is so we don't have to rewrite the Write[Seq[T]]
           (seq => Json.toJson(seq))).contramap(collectionsEntity) ~
       (__ \ "fullPath").write[List[String]] ~
+      (__ \ "correctPath").write[List[String]] ~
```
jamesgorrie: Should probably not be written out.

collections/conf/routes
jamesgorrie: Good catch.

jamesgorrie:
Removed JSON leaking to client as per discussion with @theefer

jamesgorrie:
Tested PROD data on local and ran in less than a second.

jamesgorrie:
New record from Dynamo looks like
```JSON
{
	"collection": {
		"actionData": {
			"author": "karin.andreasson@guardian.co.uk",
			"date": "2016-01-13T10:59:40.311+00:00"
		},
		"description": "Comics",
		"path": [
			"G2",
			"Comics"
		],
		"pathId": "g2/comics"
	},
	"id": "g2/comics"
}
```

jamesgorrie:
ES response:
```JSON
{
	"collections": [{
		"uri": "https://media-collections.local.dev-gutools.co.uk/images/e74b53fe02f5667bceff5f21d4b25f2539ccc414/G2/Comics",
		"data": {
			"path": [
				"G2",
				"Comics"
			],
			"pathId": "g2/comics",
			"description": "Comics",
			"actionData": {
				"author": "wham.crash@comics.com",
				"date": "2016-01-22T16:33:45.460+00:00"
			}
		},
		"actions": [{
			"name": "remove",
			"href": "https://media-collections.local.dev-gutools.co.uk/images/e74b53fe02f5667bceff5f21d4b25f2539ccc414/G2/Comics",
			"method": "DELETE"
		}]
	}]
}
```

jamesgorrie:
Tested search for old style path with new style collection :+1: (as we use the `pathId` that hasn't changed). This means they might show up lowercase on the image until we migrate those over.

theefer:
Cool! :+1:

jamesgorrie:
I will release and run this when there are less people using the Grid.

PR #1677, opened 2016-01-22
Don't render hidden collection sub-trees

theefer:
There is a huge performance hit of doing `$compile` for every single node in the collection tree. This mitigates that issue by avoiding to render hidden nodes (currently just hidden by CSS), so you only pay for the cost of however many nodes you have expanded.

This is a good reminder that often, and especially when the sub-DOM tree may be large, it's recommended to use `ng:if` over `ng:show`.

Sample difference. This test was done with only the Guide top-section expanded (not recursively).

# Before
screenshot-2016-01-22t10 58 37

Notice the 6s long frames (!).

The blue horizontal bars at the bottom are `parseHTML` from the [`$compile` call](https://github.com/guardian/grid/blob/master/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js#L151-L160) in the mutually recursive node tree directives.

# After
screenshot-2016-01-22t11 01 16

Still some heavy parseHTML, but much less, because we don't have to do it for the entire tree anymore, just the displayed nodes.

## Remaining issues

- [ ] Some JS errors I've witnessed (looking up `pathId` from `undefined` object)
- [ ] Still heavy performance hit due to use of `$compile` for every node. Worst case (tree fully expanded) is as bad as now. Note that this is due to a limitation in Angular making it very hard to build recursive directives.
- [ ] Results seem to reload after a while, which causes a whole re-render of the results page. Very inefficient.

Will be looking into more of that next.

### DIFF
--- a/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
+++ b/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
@@ -149,12 +149,11 @@ grCollectionsPanel.directive('grNode', ['$parse', '$compile', function($parse, $
             // We compile the template on the fly here as angular doesn't deal
             // well with recursive templates.
             $compile(`<gr-nodes
+                ng:if="ctrl.showChildren && ctrl.node.data.children.length > 0"
                 gr:selected-images="ctrl.selectedImages$"
                 gr:selected-collections="ctrl.selectedCollections"
                 gr:editing="ctrl.editing"
-                ng:show="ctrl.showChildren"
-                gr:nodes="ctrl.node.data.children"
-                ng:if="ctrl.node.data.children.length > 0"></gr-nodes>
+                gr:nodes="ctrl.node.data.children"></gr-nodes>
             `)(scope, cloned => {
                 element.find('.node__children').append(cloned);
             });


kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -149,12 +149,11 @@ grCollectionsPanel.directive('grNode', ['$parse', '$compile', function($parse, $
             // We compile the template on the fly here as angular doesn't deal
             // well with recursive templates.
             $compile(`<gr-nodes
+                ng:if="ctrl.showChildren && ctrl.node.data.children.length > 0"
```
theefer: Putting `ng:if` first as it's the "most important" one that conditions whether the others will be applied at all or not.

jamesgorrie:
>  Some JS errors I've witnessed (looking up pathId from undefined object)

This can happen if the node doesn't have data, which is possible.
i.e. your collections list is:
```Scala
List(List("g2"), List("g2", "art", "pictures"))
```

The "art" `Node` would be created, but not have content associated with it.

jamesgorrie:
> Results seem to reload after a while, which causes a whole re-render of the results page. Very inefficient.

Is it the lookup for new results?

[Something like this would stop the reload](https://github.com/guardian/grid/pull/1641/files#diff-32fea343d36b21c11521e08ea0059cafR102)

jamesgorrie:
:+1: 

:-1: To the idea of `ng:show` for UI interaction and `ng:if` for UI logic.

theefer:
> Is it the lookup for new results?

You mean for the "N new" indicator? I don't see why it'd cause a re-render?

jamesgorrie:
Yea, it does do a lookup on a timeout which is what made me think of it.

PR #1678, opened 2016-01-22
Memoize $compile'd template to reduce parseHTML overhead

theefer:
`$compile` is extremely expensive, and it was invoked for every node being rendered because it was done in the `link` function of the directive. [Learning a little bit about Angular](http://stackoverflow.com/a/15681173), we can leverage the `compile` function of the directive which is only run for every *reference* to the directive, i.e. when it appears in a given context (typically in the template of another directive or view).

It seems that memoizing the template in the `compile` phase (though still evaluating it every time in the `link` phase) vastly improves performance, by making it O(1) (technically O(r) where r is the number of times it occurs, which in practice is constant and equal to 4 currently) instead of O(n), for n the number of nodes rendered in the tree (easily ~100).

Inspired by http://stackoverflow.com/a/19172067.

# Before

screenshot-2016-01-22t11 01 16

Notice the blue horizontal bars of parseHTML.

# After

image

Much less parseHTML left.

Still too slow to my taste, but getting there.

### DIFF
--- a/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
+++ b/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
@@ -132,6 +132,12 @@ grCollectionsPanel.controller('GrNodeCtrl',
 }]);
 
 grCollectionsPanel.directive('grNode', ['$parse', '$compile', function($parse, $compile) {
+    const templateString = `<gr-nodes
+                                ng:if="ctrl.showChildren && ctrl.node.data.children.length > 0"
+                                gr:selected-images="ctrl.selectedImages$"
+                                gr:selected-collections="ctrl.selectedCollections"
+                                gr:editing="ctrl.editing"
+                                gr:nodes="ctrl.node.data.children"></gr-nodes>`;
     return {
         restrict: 'E',
         scope: {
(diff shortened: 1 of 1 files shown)


jamesgorrie:
:+1: Bonza :boom:

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -145,23 +151,27 @@ grCollectionsPanel.directive('grNode', ['$parse', '$compile', function($parse, $
...
+            let compiledTemplate;
+            return function link(scope, element) {
+                if (! compiledTemplate) {
+                    // We compile the template on the fly here as angular doesn't deal
+                    // well with recursive templates.
+                    compiledTemplate = $compile(templateString);
+                }
+
+                compiledTemplate(scope, cloned => {
+                    element.find('.node__children').append(cloned);
```
rrees: What is this bit doing?

rrees:
Amazing :+1: 

Will this pattern just work generally?

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
theefer: I'm not entirely sure where `node__children` is coming from or found, actually. I haven't changed that code.

@jamesgorrie ?

jamesgorrie: [There's a few ways of doing this - but this seemed the most sane](https://github.com/guardian/grid/blob/363f63a4fe2f3e9be3e72a7bf51d7f0582160e4b/kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html#L71).

Might be worth dropping a note in that template.

theefer: Actually I misread the directive, that template is referenced by the directive so it makes sense.

theefer:
> Will this pattern just work generally?

As discussed, it's only really applicable when `$compile` is called explicitly, which should be rare, typically in recursive directives.

rrees:
@theefer @jamesgorrie just told me that recursive directives are supported in Angular 1.5

theefer:
Oh! interesting. Too late, but nice to hear :-)

PR #1679, opened 2016-01-22
updates ui when images dragged into collection

tsop14:
The two methods (dragging and clicking on add button) both use 'addCollectionToImage' function, so checks for new collection and updates ui when images are dragged in to a collection.  

adding-images

Changes 'add a new sub-collection' icon. #perfectmatch ⚡️

screen shot 2016-01-22 at 12 02 35


kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
```diff
@@ -37,7 +37,7 @@
                 title="Add a new sub-collection"
                 ng:if="ctrl.editing"
                 ng:click="$parent.active = !$parent.active">
-            <gr-icon>add_box</gr-icon>
+            <gr-icon>create_new_folder</gr-icon>
```
theefer: What does that look like out of interest?

kahuna/public/js/services/api/collections-api.js
```diff
@@ -134,8 +137,8 @@ collectionsApi.factory('collections',
         addChildTo,
         isDeletable,
         removeFromList,
-        addImageIdsToCollection,
-        addImagesToCollection,
+        addCollectionUsingImageIds,
+        addCollectionUsingImageResources,
```
theefer: Should it be add *to* collection? Else I wrongly read it as adding (creating) a new collection.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
theefer: Ignore me, found the screenshot, doh.

theefer: If we're going for collection = folder, maybe we should replace the icon on the collection tab?

theefer:
Nice :+1:

kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
tsop14: @jamesgorrie do you think of them as folders?

PR #1680, opened 2016-01-22
add tracking to panel buttons

jamesgorrie:
Forgot to add it before.
:rage4: :rabbit:

### DIFF
--- a/kahuna/public/js/components/gr-panel-button/gr-panel-button-small.html
+++ b/kahuna/public/js/components/gr-panel-button/gr-panel-button-small.html
@@ -6,7 +6,9 @@
             ng:click="ctrl.hidePanel()"
             ng:class="{'bg-light': !ctrl.state.hidden}"
             gr:tooltip="Hide {{ctrl.name}} panel"
-            gr:tooltip-position="{{ctrl.toolTipPosition()}}">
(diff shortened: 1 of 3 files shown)


kahuna/public/js/components/gr-panel-button/gr-panel-button-small.html
```diff
@@ -6,7 +6,9 @@
             ng:click="ctrl.hidePanel()"
             ng:class="{'bg-light': !ctrl.state.hidden}"
             gr:tooltip="Hide {{ctrl.name}} panel"
-            gr:tooltip-position="{{ctrl.toolTipPosition()}}">
+            gr:tooltip-position="{{ctrl.toolTipPosition()}}"
+            gr:track-click="{{ctrl.trackingName}}"
+            gr:track-click-data="ctrl.trackingData('Hide panel')">
```
akash1810: minor: the action could just be "Hide" as we can infer its a panel from the context...

akash1810:
:+1:

kahuna/public/js/components/gr-panel-button/gr-panel-button-small.html
jamesgorrie: I thought so, but as we might have that else where, I didn't want it to clog up the data.

PR #1682, opened 2016-01-22
Slightly less verbose use of collection resources

theefer:
Needless cleanup.

### DIFF
--- a/kahuna/public/js/services/api/collections-api.js
+++ b/kahuna/public/js/services/api/collections-api.js
@@ -13,13 +13,15 @@ collectionsApi.factory('collections',
                        ['$rootScope', '$q', 'mediaApi', 'imageAccessor', 'apiPoll',
                         function ($rootScope, $q, mediaApi, imageAccessor, apiPoll) {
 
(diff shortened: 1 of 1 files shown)


kahuna/public/js/services/api/collections-api.js
```diff
@@ -13,13 +13,15 @@ collectionsApi.factory('collections',
...
+    const collectionsRoot = mediaApi.root.follow('collections');
+
     // TODO: Rx?
     let collections;
 
     function getCollections() {
+        // TODO: do we want to memoize? what if we want to reload?
         if (! collections) {
-            collections = mediaApi.root.follow('collections').get().
-                then(collectionsService => collectionsService.follow('collections').get());
```
theefer: Main thing to note is that you can do `mediaApi.root.follow('collections').follow('collections').get()` since resources are lazy that way.

jamesgorrie:
:

jamesgorrie:
:+

jamesgorrie:
:+1:

PR #1683, opened 2016-01-22
dot-properties helper script

akash1810:
- **Don't** hardcode `es.host=elasticsearch` as it only applies in the context of docker.
- `generate-dot-properties.sh` helper script that will install a python venv for you, so you don't have to.

cc @theefer


theefer:
:+1:

PR #1684, opened 2016-01-22
first pass at thrall writing to two indices with same alias

NickPapacostas:
This is so that during reindex we can add the write alias to the new and old indexes

### DIFF
--- a/thrall/app/lib/ElasticSearch.scala
+++ b/thrall/app/lib/ElasticSearch.scala
@@ -95,8 +95,8 @@ object ElasticSearch extends ElasticSearchClient with ImageFields {
   }
 
   def updateImageUsages(id: String, usages: JsValue, lastModified: JsValue)(implicit ex: ExecutionContext): List[Future[UpdateResponse]] = {
-    prepareImageUpdate(id) { request => request
-      .setScriptParams(Map(
+    prepareImageUpdate(id) { request =>
+      request.setScriptParams(Map(
         "usages" -> asGroovy(usages),
         "lastModified" -> asGroovy(lastModified)
       ).asJava)
@@ -108,7 +108,7 @@ object ElasticSearch extends ElasticSearchClient with ImageFields {
         """.stripMargin,
         scriptType)
       .executeAndLog(s"updating usages on image $id")
-      .recover { case e: DocumentMissingException => Unit }
+      .recover { case e: DocumentMissingException => new UpdateResponse }
       .incrementOnFailure(failedUsagesUpdates) { case e: VersionConflictEngineException => true }
     }
   }
@@ -175,9 +175,7 @@ def updateImageExports(id: String, exports: JsValue)(implicit ex: ExecutionConte
     }
 
   def prepareImageUpdate(id: String)(op: UpdateRequestBuilder => Future[UpdateResponse]): List[Future[UpdateResponse]] = {
-    getCurrentIndices.map( index =>
-      list.map(
-        index => {
+    getCurrentIndices.map( index => {
           val updateRequest = client.prepareUpdate(index, imageType, id)
             .setScriptLang("groovy")
           op(updateRequest)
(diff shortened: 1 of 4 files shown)


thrall/app/lib/ElasticSearch.scala
```diff
@@ -25,7 +25,7 @@ object ElasticSearch extends ElasticSearchClient with ImageFields {
 
   import com.gu.mediaservice.lib.formatting._
 
-  val imagesAlias = Config.imagesAlias
+  val imagesAlias = "writeAlias"
```
NickPapacostas: ignore this, should have been reverted

thrall/app/lib/ElasticSearch.scala
```diff
@@ -36,31 +36,33 @@ object ElasticSearch extends ElasticSearchClient with ImageFields {
 
   def currentIsoDateString = printDateTime(new DateTime())
 
-  def indexImage(id: String, image: JsValue)(implicit ex: ExecutionContext): Future[UpdateResponse] = {
-    prepareImageUpdate(id)
+  def indexImage(id: String, image: JsValue)(implicit ex: ExecutionContext): List[Future[UpdateResponse]] = {
```
jamesgorrie: Am I right in assuming, when not in reindex mode, there will only be one index?

NickPapacostas: yes i believe so. i.e. i have no reason to think otherwise

thrall/app/lib/ElasticSearch.scala
```diff
@@ -36,31 +36,33 @@ object ElasticSearch extends ElasticSearchClient with ImageFields {
...
+    prepareImageUpdate(id).map( request => request
```
jamesgorrie: We could make `prepareImageUpdate` to take a block to run on all indices.
e.g.
```Scala
prepareImageUpdate(id) { request =>
   // ...
}

// ...

def prepareImageUpdate(op: UpdateRequestBuilder => ActionResponse): UpdateRequestBuilder =
  getCurrentIndices.map(_.map (index =>
      val updateRequest = client.prepareUpdate(index, imageType, id)
        .setScriptLang("groovy")
      )
      op(request)
    ).get
```

It's a little hard to do this in GH.

thrall/app/lib/ElasticSearch.scala
```diff
@@ -165,10 +171,16 @@ object ElasticSearch extends ElasticSearchClient with ImageFields {
...
       .executeAndLog(s"setting collections on image $id")
       .incrementOnFailure(failedCollectionsUpdates) { case e: VersionConflictEngineException => true }
+    )
+
+  def prepareImageUpdate(id: String): List[UpdateRequestBuilder] =
+    getCurrentIndices.map(_.map (index =>
+      client.prepareUpdate(index, imageType, id)
+        .setScriptLang("groovy")
+      )
+    ).get
```
jamesgorrie: Having a get is going to throw on an empty Option. It might be worth thinking about where we throw, potentially in `getCurrentIndices` because that should always return something?

thrall/app/lib/ElasticSearch.scala
jamesgorrie:
```Scala
def setImageCollection(id: String, collections: JsValue)(implicit ex: ExecutionContext): List[Future[ActionResponse]] =
    prepareImageUpdate(id) { request =>
      request.setScriptParams(Map(
        "collections" -> asGroovy(collections),
        "lastModified" -> asGroovy(JsString(currentIsoDateString))
      ).asJava)
        .setScript(
          "ctx._source.collections = collections;" +
            updateLastModifiedScript,
          scriptType)
        .executeAndLog(s"setting collections on image $id")
        .incrementOnFailure(failedCollectionsUpdates) { case e: VersionConflictEngineException => true }
    }

  def prepareImageUpdate(id: String)(op: UpdateRequestBuilder => Future[ActionResponse]) =
    getCurrentIndices.map { list =>
      list.map { index =>
        val updateRequest = client.prepareUpdate(index, imageType, id)
          .setScriptLang("groovy")

        op(updateRequest)
    }}.get
```

That works.
The reason would be that we then don't have to remember to map everytime.

kenoir:
@NickPapacostas did you test/confirm the document versioning issue we talked about when writing to both indexes while performing a reindex?

kenoir:
:+1: if this works out in `test`

PR #1685, opened 2016-01-22
Return only latest collections

jamesgorrie:
We should be returning the deduped list everywhere.
The will allow the client to do a proper comparison on ES vs Collections response.


jamesgorrie:
@tsop14

theefer:
:+1:

PR #1686, opened 2016-01-22
batch remove from collection

tsop14:
Adds ability to batch remove selected images from a collection by clicking on a button with a minus sign in the collections panel. 

batch-remove

### DIFF
--- a/kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
+++ b/kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
@@ -43,15 +43,18 @@
         <div class="text-small flex-container flex-center"
              ng:if="ctrl.saving || dropIntoCollectionSaving">Adding…</div>
 
+        <div class="text-small flex-container flex-center"
+             ng:if="ctrl.removing">Removing…</div>
+
         <button class="node__action clickable hover-child"
                 title="Remove selected images from this collection"
-                ng:if="!ctrl.editing && ctrl.hasImagesSelected && !ctrl.saving"
+                ng:if="!ctrl.editing && ctrl.hasImagesSelected && !ctrl.saving && !ctrl.removing"
                 ng:click="ctrl.removeImagesFromCollection()">
             <gr-icon>indeterminate_check_box</gr-icon>
         </button>
         <button class="node__action clickable hover-child"
                 title="Add selected images to this collection"
-                ng:if="!ctrl.editing && ctrl.hasImagesSelected && !ctrl.saving"
+                ng:if="!ctrl.editing && ctrl.hasImagesSelected && !ctrl.saving && !ctrl.removing"
                 ng:click="ctrl.addImagesToCollection()">
             <gr-icon>add_to_photos</gr-icon>
         </button>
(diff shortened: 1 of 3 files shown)


kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
```diff
@@ -43,9 +43,18 @@
         <div class="text-small flex-container flex-center"
              ng:if="ctrl.saving || dropIntoCollectionSaving">Adding…</div>
 
+        <div class="text-small flex-container flex-center"
+             ng:if="ctrl.removing">Removing…</div>
+
+        <button class="node__action clickable hover-child"
+                title="Remove selected images from this collection"
+                ng:if="!ctrl.editing && ctrl.hasImagesSelected && !ctrl.saving && !ctrl.removing"
```
akash1810: This could be a function in the controller, which'll allow the template to be: `ng:if=ctrl.removeEnabled()` (or a better name).

jamesgorrie: It could just be a value?

jamesgorrie: `ctrl.removeEnabled = ctrl.hasImagesSelected && !(ctrl.editing || ctrl.hasImagesSelected || ctrl.saving || ctrl.removing)`?

akash1810: > It could just be a value?

:+1:

kahuna/public/js/services/api/collections-api.js
```diff
@@ -127,6 +127,35 @@ collectionsApi.factory('collections',
             });
     }
 
+    function findCollectionResource(image, collectionToMatch){
+        return image.data.collections.filter(collection => {
```
jamesgorrie: You could use a [`find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find)

kahuna/public/js/services/api/collections-api.js
```diff
@@ -127,6 +127,35 @@ collectionsApi.factory('collections',
...
+            return collection.data.pathId === collectionToMatch;
+        });
+    }
+
+    function getCollectionToRemove(image, collection) {
+        const filteredCollections = findCollectionResource(image, collection);
```
jamesgorrie: [If you use a find](https://github.com/guardian/grid/pull/1686/files#r50559228), the name could change too.

jamesgorrie:
Bar a few code styling things :+1:

kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
theefer: Could it? Surely it needs to be updated if any of `hasImagesSelected`, `editing` etc gets updated?

jamesgorrie: I was asking as a question - if those values change, is the digest not rerun?

kahuna/public/js/services/api/collections-api.js
theefer: Best check if supported by the browsers we support, or polyfilled by Traceur (didn't look like it last time I looked).

kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
theefer: Only expressions in templates are watched, not in the controller?

kahuna/public/js/services/api/collections-api.js
tsop14: Decided not to use 'find' in previous PR. I have given the humble filter function a honest name.

PR #1687, opened 2016-01-22
square crop

akash1810:
:black_square_button: :white_square_button: :black_square_button: :white_square_button: :black_square_button: :white_square_button: :black_square_button: :white_square_button: 

screen shot 2016-01-22 at 17 06 24

Issue #199, opened 2014-11-07
Check crop dimensions to avoid rounding noise


theefer:
For instance we sometimes end up with 599px when we would expect 600px, or 2000x1202 when it should really be 2000x1200.

Let's review the way dimensions rounding works and see if this can be tidied up.

Also affects frontend that are particularly strict on dimensions.

kenoir:
Think this might actually be resolved now.

PR #1688, opened 2016-01-22
Revert "square crop"

akash1810:
Reverts guardian/grid#1687


theefer:
When reverting, could we please explain why?

akash1810:
My bad, I explained in the parent PR.

akash1810:
Reverting as #199 is pretty obvious :cry:

screen shot 2016-01-22 at 17 29 34

PR #1690, opened 2016-01-26
symlink config generators to `scripts` directory at project root.

akash1810:
The generators are currently in the `docker` directory, which isn't the most obvious place if you're running the Grid w/out Docker.

Symlinked because there are generators that are specific to Docker (imgops, nginx, osx_hosts_file) within the [generators package](https://github.com/guardian/grid/tree/aa-config-gen-symlink/docker/configs/generators/generators), so wouldn't make sense to move the generators directory to project root...

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,10 +102,10 @@ cd cloud-formation/scripts
 ./create-dev-stack.sh
 ```
 
-### .properties files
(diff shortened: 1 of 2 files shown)


jamesgorrie:
:+1:

PR #1692, opened 2016-01-27
Handle notes and source as empty strings

kenoir:
When updating print usages we might occasionally be sent notes and
soruce fields with empty strings (which cannot be passed to DynamoDb as
it balks at empty strings in fields).

This PR updates the map generation to prevent empty string fields being
sent.


kenoir:
Tested this locally and all seems to work fine.

cc @blishen <- this should fix any usage errors caused by potentially empty fields.

akash1810:
:+1:

PR #1694, opened 2016-01-27
moves toLowerCase to manager

tsop14:
> Should this be happening inside CollectionsManager?

> Yup.

TA DA! 🌟

### DIFF
--- a/collections/app/store/CollectionsStore.scala
+++ b/collections/app/store/CollectionsStore.scala
@@ -27,7 +27,7 @@ object CollectionsStore {
   }
 
   def remove(collectionPath: List[String]): Future[Unit] = {
-    val path = CollectionsManager.pathToString(collectionPath).toLowerCase
+    val path = CollectionsManager.pathToLowerCaseString(collectionPath)
     dynamo.deleteItem(path)
   } recover {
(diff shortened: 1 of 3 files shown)


collections/app/store/CollectionsStore.scala
```diff
@@ -27,7 +27,7 @@ object CollectionsStore {
   }
 
   def remove(collectionPath: List[String]): Future[Unit] = {
-    val path = CollectionsManager.pathToString(collectionPath).toLowerCase
+    val path = CollectionsManager.pathToLowerCaseString(collectionPath)
```
jamesgorrie: Sorry to nitpick, but it could be `pathToPathId`.
It could then be [used here](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/model/Collection.scala#L11) too.

jamesgorrie:
A slight improvement if you wanted to, else :+1:

collections/app/store/CollectionsStore.scala
tsop14: With the understanding PathId is always lowercase string? Sounds good.

kenoir:
:+1:

PR #1695, opened 2016-01-28
Increases JSON body parse size

kenoir:


akash1810:
:+1:

akash1810:
what's the default, that we need to increase?

kenoir:
@akash1810 100K

kenoir:
https://www.playframework.com/documentation/2.2.x/ScalaBodyParsers

PR #1696, opened 2016-01-29
Set print usage request max size in code

kenoir:
This change sets the max request size for print usage only.

### DIFF
--- a/usage/app/lib/Config.scala
+++ b/usage/app/lib/Config.scala
@@ -25,6 +25,10 @@ object Config extends CommonPlayAppProperties with CommonPlayAppConfig {
 
   val defaultPageSize = 100
   val defaultMaxRetries = 6
+  val defaultMaxPrintRequestSizeInKb = 500
+
+  val maxPrintRequestLengthInKb = Try(properties("api.setPrint.maxLength").toInt)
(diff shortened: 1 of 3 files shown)


akash1810:
what's the advantage of this over the config file?

kenoir:
@akash1810 it sets the max request size for this route only, also i'm not convinced the config was taking effect.

usage/app/lib/Config.scala
```diff
@@ -25,6 +25,10 @@ object Config extends CommonPlayAppProperties with CommonPlayAppConfig {
 
   val defaultPageSize = 100
   val defaultMaxRetries = 6
+  val defaultMaxPrintRequestSizeInKb = 500
```
akash1810: minor: why not just do `val defaultMaxPrintRequestSizeInKb = 500 * 1024` here and get rid of `maxPrintRequestLength` above as `maxPrintRequestLength` feels like a config constant.

akash1810:
:+1:

usage/app/lib/Config.scala
kenoir: Feels weird to be doing anything except setting config here.

PR #1697, opened 2016-01-29
Picdar-Export: Tolerate empty production and publicationtext fields

kenoir:
These failures are sometimes missing in Picdar, we can tolerate their absence so make the code reflect that.


NickPapacostas:
:+1:  ^_^

PR #1698, opened 2016-01-29
Usage: Better logging on usage table update fail

kenoir:
This is an attempt to provide better visibility to dynamo update fail errors.


akash1810:
:+1:

PR #1699, opened 2016-01-29
small features for reindex and updatemapping script

NickPapacostas:
reindex script takes optional "from date", update mappings takes optional index name


scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
```diff
@@ -99,13 +104,22 @@ object Reindex extends EsScript {
...
+          val recentChangesQuery = rangeQuery("lastModified").from(fromTime.get).to(DateTime.now)
+          reindexScroll(query.setQuery(recentChangesQuery).execute.actionGet)
+        }
 
         // TODO: deleteIndex when we are confident
         client.close()
       }
     }
-    EsClient.reindex
+    if(extraArgs.isEmpty) {
```
kenoir: I'd avoid `if` statements like this wherever possible and either `foldLeft` or return a value:

```scala
val date = if(extraArgs.isEmpty) None else Some((DateTime.parse(extraArgs(0))))
EsClient.reindex(date)
```

or

```scala
EsClient.reindex(
  extraArgs.foldLeft(None)(Some(DateTime.parse(extraArgs(0))))
)
```

kenoir: Goes for the `if` statements elsewhere too.

LATaylor-guardian: A statement in the form:

val date = if(extraArgs.isEmpty) None else Some((DateTime.parse(extraArgs(0))))
EsClient.reindex(date)

is certainly more accessible to non-native Scala developers or newcomers than foldLeft.

jamesgorrie: But it's quite a good thing to find out about if you don't know it.

NickPapacostas:
tried to find a happy medium :) foldLeft didn't make it in though :disappointed:

scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
```diff
@@ -2,12 +2,19 @@ package com.gu.mediaservice.scripts
 
 import org.elasticsearch.action.admin.indices.close.CloseIndexRequest
 import org.elasticsearch.action.admin.indices.open.OpenIndexRequest
-import org.elasticsearch.action.search.SearchResponse
-import org.elasticsearch.search.sort.SortOrder
-import org.elasticsearch.index.query.QueryBuilders.matchAllQuery
+import org.elasticsearch.action.bulk.{BulkRequestBuilder}
```
akash1810: don't need braces with singular imports

scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
```diff
@@ -42,81 +49,132 @@ object MoveIndex extends EsScript {
...
+    val newIndex = EsClient.nextIndex
+
+    val from = if(args.isEmpty) None else Some(DateTime.parse(args(0)))
+    validateCurrentState(EsClient, from)
+    Await.result(reindex(from, EsClient), Duration.Inf)
+
+    def reindex(from: Option[DateTime], esClient: ElasticSearchClient) : Future[SearchResponse] = {
+      def _scroll(scroll: SearchResponse, done: Long = 0): Future[SearchResponse] = {
+        val client = esClient.client
+        val doing = done + scrollSize
```
akash1810: interesting naming :stuck_out_tongue:

scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
```diff
@@ -42,81 +49,132 @@ object MoveIndex extends EsScript {
...
+        }
+      }
 
-            bulk.execute.actionGet
-            reindexScroll(client.prepareSearchScroll(scroll.getScrollId)
-              .setScroll(scrollTime).execute.actionGet, doing)
-          } else {
+      def query(from: Option[DateTime]) : SearchRequestBuilder = {
+        val queryType = from.map(time =>
+          rangeQuery("lastModified").from(from.get).to(DateTime.now)
```
akash1810: minor: might want to validate from is always less than `DateTime.now` as its an parameter

scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
```diff
@@ -42,81 +49,132 @@ object MoveIndex extends EsScript {
...
+          .setTypes(imageType)
+          .setScroll(scrollTime)
+          .setSize(scrollSize)
+          .setQuery(queryType)
+      }
 
-        Thread.sleep(1000)
 
-        reindexScroll(query.execute.actionGet)
+      if(from.isEmpty) {
```
akash1810: its probably obvious, but I don't quite understand this `if` - can you help me?

NickPapacostas: it's not clear! maybe i should throw it in a nicely named method. 

If the user supplies a "from" time (or it's been called recursively) then don't create a new index to populate. Basically, if you're asking for a partial update, then you don't want a new index.

scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
NickPapacostas: good call!

scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
NickPapacostas: haha i'm not sure if you're referring to "doing" but i borrowed it from the old code... funny how things dont seem funny after looking at them for a month >.< how about "currentPage" or "currentBatch"

scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
akash1810: Ah! That makes sense. (could you add a comment for my simile future self? :smile:)

scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
```diff
@@ -42,81 +49,134 @@ object MoveIndex extends EsScript {
...
+      if(esClient.getCurrentIndices.isEmpty) raise("no index with the 'write' alias exists")
+      if((esClient.getCurrentIndices.length == 2) && from.isEmpty)
+        raise("there are two indices with 'write' alias, check your properties file or use http://localhost:9200/_plugin/head/ ")
+      if(from.exists(_.isAfter(DateTime.now())))
+        raise("DateTime parameter 'from' must be earlier than the current time" )
+    }
 
-        def reindexScroll(scroll: SearchResponse, done: Long = 0) {
+    val imageType = "image"
+    val scrollTime = new TimeValue(5 * 60 * 1000) // 5 minutes in milliseconds
```
kenoir: You might well be able to express this in code using the Joda lib, (http://joda-time.sourceforge.net/apidocs/org/joda/time/Period.html) but it's probably not worth it.

kenoir:
:+1:

PR #1701, opened 2016-01-29
separate read and write aliases 

NickPapacostas:
used in media-api and thrall

### DIFF
--- a/scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
+++ b/scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
@@ -13,7 +13,7 @@ object MoveIndex extends EsScript {
   def run(esHost: String, extraArgs: List[String]) {
 
     object EsClient extends ElasticSearchClient {
-      val imagesAlias = "imagesAlias"
+      val imagesAlias = "readAlias"
       val port = esPort
       val host = esHost
       val cluster = esCluster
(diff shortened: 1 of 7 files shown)


thrall/app/lib/Config.scala
```diff
@@ -23,7 +23,7 @@ object Config extends CommonPlayAppConfig {
 
   val imageBucket: String = properties("s3.image.bucket")
 
-  val imagesAlias = properties("es.index.alias")
+  val writeAlias = properties("es.index.aliases.write")
```
akash1810: are we sure thrall doesn't need the read alias too?

scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
```diff
@@ -13,7 +13,7 @@ object MoveIndex extends EsScript {
   def run(esHost: String, extraArgs: List[String]) {
 
     object EsClient extends ElasticSearchClient {
-      val imagesAlias = "imagesAlias"
+      val imagesAlias = "readAlias"
```
akash1810: can this be read from the `Config` object like [here](https://github.com/guardian/grid/pull/1701/files#diff-c8a14be53a6e8680f2e620d6c17edaadR56)

NickPapacostas: I'm not sure tbh. I couldn't get it workings and there was this comment further down in the file: 
  // FIXME: Get from config (no can do as Config is coupled to Play)

thrall/app/lib/Config.scala
NickPapacostas: I actually don't think thrall _should_ touch the read alias. in the sense that, if it's query-ing to determine what to do next, it should be doing so on the write alias shouldn't it? (<- not rhetorical ^_^) 

I double checked anyway and I don't see anywhere in Thrall where a query reads w/o writing.

akash1810:
:+1:

PR #1702, opened 2016-02-01
Set publication name from code if none given

kenoir:
Fixes a problem with usages being refused if no publication name is given.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/usage/PrintUsageMetadata.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/usage/PrintUsageMetadata.scala
@@ -41,13 +41,20 @@ case class PrintUsageMetadata(
   type LongElement = List[(String,Long)]
   type IntElement = List[(String,Int)]
 
+  val calculatedPublicationName =
+    if (publicationName.trim.isEmpty) publicationCode match {
+      case "gdn" => "Guardian"
+      case "obs" => "Observer"
+      case _ => "Unknown Publication"
+    } else publicationName
+
   def toMap = Map(
     "sectionName" -> sectionName,
     "issueDate" -> issueDate.toString,
     "pageNumber" -> pageNumber,
     "storyName" -> storyName,
     "publicationCode" -> publicationCode,
-    "publicationName" -> publicationName,
+    "publicationName" -> calculatedPublicationName,
     "sectionCode" -> sectionCode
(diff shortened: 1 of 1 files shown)


common-lib/src/main/scala/com/gu/mediaservice/model/usage/PrintUsageMetadata.scala
```diff
@@ -41,13 +41,20 @@ case class PrintUsageMetadata(
   type LongElement = List[(String,Long)]
   type IntElement = List[(String,Int)]
 
+  val calculatedPublicationName =
+    if (publicationName.trim.isEmpty) publicationCode match {
```
akash1810: `trim.isEmpty` does that mean some publication codes are whitespace? :fearful:

NickPapacostas:
*sheepish* :+1:

common-lib/src/main/scala/com/gu/mediaservice/model/usage/PrintUsageMetadata.scala
```diff
@@ -41,13 +41,20 @@ case class PrintUsageMetadata(
...
+      case "gdn" => "Guardian"
+      case "obs" => "Observer"
+      case _ => "Unknown Publication"
+    } else publicationName
```
theefer: Minor: should this also be trimmed? Or should we trim prior to storing the data?

common-lib/src/main/scala/com/gu/mediaservice/model/usage/PrintUsageMetadata.scala
kenoir: @akash1810 I think they might sometimes be whitespace at the moment, this is within our control though, @blishen is looking `t it.

kenoir:
I'm going to see if we've resolved this on the other side before merging.

common-lib/src/main/scala/com/gu/mediaservice/model/usage/PrintUsageMetadata.scala
kenoir: This is protecting specifically against the case where we try and store only whitespace in a dynamo field which borks, so I don't think it matters if we trim as long as the fields not only whitespace.

Ideally we don't do this at all.

theefer: :+1: (dynamo non-support for empty strings still boggles my mind)

kenoir:
Looks like we fixed this on the other side, so closing.

PR #1703, opened 2016-02-01
Convert ftp-watcher back to assembly packaging

theefer:
Bring back some legacy code to fix the ftp-watcher `dist` process so it generates the correct artifact with the correct mega-JAR, so we can deploy it again (in order to get rid of it).

cc @kenoir


kenoir:
:+1:

PR #1704, opened 2016-02-01
Mimics "preset" functionality from Picdar

kenoir:
Looks for IPTC category codes and applies a suppliersCategory.

**Note:** This PR requires `updateMapping` be runon ES  before deployment.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
@@ -69,7 +69,7 @@ object Mappings {
     "copyright" -> standardAnalysedString,
     "copyrightNotice" -> standardAnalysedString,
     "suppliersReference" -> standardAnalysedString,
-    "subject" -> standardAnalysedString,
+    "subjects" -> nonAnalysedList("subject"),
     "source" -> nonAnalyzedString,
     "specialInstructions" -> nonAnalyzedString,
     "keywords" -> nonAnalysedList("keyword"),
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
@@ -69,6 +69,7 @@ object Mappings {
     "copyright" -> standardAnalysedString,
     "copyrightNotice" -> standardAnalysedString,
     "suppliersReference" -> standardAnalysedString,
+    "subject" -> standardAnalysedString,
     "source" -> nonAnalyzedString,
     "specialInstructions" -> nonAnalyzedString,
     "keywords" -> nonAnalysedList("keyword"),
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
@@ -31,7 +31,7 @@ object ImageMetadataConverter {
       state               = fileMetadata.iptc.get("Province/State"),
       country             = fileMetadata.iptc.get("Country/Primary Location Name"),
       subject             = fileMetadata.iptc.get("Category")
-                              .map(Subject.create)
+                              .flatMap(Subject.create)
                               .map(_.toString)
     )
 
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
@@ -10,6 +10,21 @@ import com.gu.mediaservice.model.{FileMetadata, ImageMetadata}
 
 object ImageMetadataConverter {
 
+  def extractSubjects(fileMetadata: FileMetadata): List[String] = {
+    val supplementalCategories = fileMetadata.iptc
+      .get("Supplemental Category(s)")
+      .toList.flatMap(_.split("\\s+"))
+
+    val category = fileMetadata.iptc
+      .get("Category")
+
+    (supplementalCategories ::: category.toList)
+      .map(Subject.create)
+      .flatten
+      .map(_.toString)
+      .distinct
+  }
+
   def fromFileMetadata(fileMetadata: FileMetadata): ImageMetadata =
     ImageMetadata(
       dateTaken           = (fileMetadata.iptc.get("Date Created") flatMap parseRandomDate) orElse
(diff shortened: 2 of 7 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
```diff
@@ -7,6 +7,23 @@ import scala.util.Try
...
 
+object SupplierCategory extends Enumeration {
+  type SupplierCategory = Value
+  val Sport = Value("Sport")
+  val Entertainment = Value("Entertainment")
+  val Unknown = Value("Unknown")
+
+  // These category codes are now deprecated but still populated
+  // See: https://www.iptc.org/std/photometadata/documentation/GenericGuidelines/index.htm#!Documents/guidelineformappingcategorycodestosubjectnewscodes.htm
+  def create(category: String) = category match {
```
kenoir: Only doing the ones we know folks use for now.

theefer: Shouldn't we have loads more mappings in here?

Only raising now because it's quite an expensive process to map them.

common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
```diff
@@ -7,6 +7,23 @@ import scala.util.Try
...
+    case "S" => Sport
+    case "SPO" => Sport
+    case "E" => Entertainment
+    case "ACE" => Entertainment
+    case _ => Unknown
```
theefer: There are many more known ones. If we're sticking to a few selected ones, it may be best to filter the ones we don't recognise for now, rather than mapping them to "Unknown" IIUC?

common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
```diff
@@ -28,10 +45,12 @@ object ImageMetadataConverter {
       subLocation         = fileMetadata.iptc.get("Sub-location"),
       city                = fileMetadata.iptc.get("City"),
       state               = fileMetadata.iptc.get("Province/State"),
-      country             = fileMetadata.iptc.get("Country/Primary Location Name")
+      country             = fileMetadata.iptc.get("Country/Primary Location Name"),
+      suppliersCategory   = fileMetadata.iptc.get("Category")
```
theefer: Should we also look at Supplemental Category?

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -69,6 +69,7 @@ object Mappings {
     "copyright" -> standardAnalysedString,
     "copyrightNotice" -> standardAnalysedString,
     "suppliersReference" -> standardAnalysedString,
+    "suppliersCategory" -> standardAnalysedString,
```
theefer: Why the `suppliers` prefix?

common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
```diff
@@ -7,6 +7,23 @@ import scala.util.Try
 
 import com.gu.mediaservice.model.{FileMetadata, ImageMetadata}
 
+object SupplierCategory extends Enumeration {
+  type SupplierCategory = Value
+  val Sport = Value("Sport")
```
theefer: do we want them to be title cased?

media-api/app/lib/querysyntax/QuerySyntax.scala
```diff
@@ -37,6 +37,7 @@ class QuerySyntax(val input: ParserInput) extends Parser with ImageFields {
     "copyright" |
     "source" |
     "category" |
+    "preset" |
```
theefer: Preset is really a horrible picdarism, why not category (though that clashes with the current usage rights categories, so maybe not) or, to polyfill the future, subject maybe?

common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
kenoir: The full list is probably at that URL. 

I suspect this might be useless for anyone else except sport folks *and* is a deprecated standard. They also seem to be used as a way for folks to handle the stream of incoming images (rather than looking very far back), so we may not need to reindex old images if we're lucky.

theefer: Feels like an easy win to go beyond what they are just used to and provide more power, e.g. not just "sport" but "football", as a useful filter if you're covering a football match?

IIRC people also mentioned filtering out things like fashion, but I don't know if this was a genuine use case or a "maybe I'd do it". The point being that if they can't do it today with presents, they may not know the value of it.

media-api/app/lib/querysyntax/QuerySyntax.scala
jamesgorrie: topic?

common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
kenoir: The standards & codes for "Supplemental Category" are just a horrible eye bleeding mess which i'm not going to touch with a very long stick.

theefer: My hunch having looked at some data is that both Category and Supplemental Category are a bit of a mess, using two standards (single and three letter codes) and sometimes full words. It's not clear that the Category is reliably a different or more important category than the Supplemental Category.

My instinct would be to bundle them all into a single list -- simpler and I doubt users will ever need the power of major vs minor categories.

If it's too much faff to map many categories, it probably makes sense to start with the sport (or sport related) ones as you've done, and add more later. It would however be easier to get the ES mapping right the first time around, as an ES reindex is still not massively fun.

common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
paperboyo: > Feels like an easy win to go beyond what they are just used to and provide more power

+100, although "maybe I'd do it" ;)



> useless for anyone else except sport folks

I'd do it, I swear!



> deprecated standard

Not really, if our providers still use them. I wonder if any of the providers use _any_ of the standards that came after...

kenoir:
@theefer addressed (most of) your comments.

theefer:
:+1: Awesome, thanks for bearing with me

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -69,6 +69,7 @@ object Mappings {
     "copyright" -> standardAnalysedString,
     "copyrightNotice" -> standardAnalysedString,
     "suppliersReference" -> standardAnalysedString,
+    "subject" -> standardAnalysedString,
```
theefer: Should we turn this into a list now?

PR #1705, opened 2016-02-01
delete requests handle multiple indexes with the "write" alias

NickPapacostas:
I hadn't noticed this issue (nor is it a problem on prod) because deleting seems like it works until you refresh.

### DIFF
--- a/thrall/app/lib/ThrallMessageConsumer.scala
+++ b/thrall/app/lib/ThrallMessageConsumer.scala
@@ -52,24 +52,26 @@ object ThrallMessageConsumer extends MessageConsumer(
   def setImageCollections(collections: JsValue): Future[List[UpdateResponse]]=
     Future.sequence(withImageId(collections)(id => ElasticSearch.setImageCollection(id, collections \ "data")) )
 
-  def deleteImage(image: JsValue): Future[EsResponse] =
-    withImageId(image) { id =>
-      // if we cannot delete the image as it's "protected", succeed and delete
-      // the message anyway.
-      ElasticSearch.deleteImage(id).map {
-        case r: DeleteByQueryResponse => {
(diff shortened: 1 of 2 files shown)


thrall/app/lib/ThrallMessageConsumer.scala
```diff
@@ -52,24 +52,26 @@ object ThrallMessageConsumer extends MessageConsumer(
...
-      } recoverWith {
-        case ImageNotDeletable => {
-          Logger.info(s"Could not delete image $id")
-          Future.successful(EsResponse(s"Image cannot be deleted: $id"))
+  def deleteImage(image: JsValue): Future[List[EsResponse]] =
+    Future.sequence(
+      withImageId(image) { id =>
+        // if we cannot delete the image as it's "protected", succeed and delete
+        // the message anyway.
+        ElasticSearch.deleteImage(id).map { requests =>
```
NickPapacostas: So here now we just add an extra mapping layer, and maybe i'm missing something, but i would have thought a flat map here was what I was looking for, but it didn't seem to work.

NickPapacostas: or should i be using a for comprehension?

NickPapacostas:
Tested successfully

akash1810:
Its a shame that elastic doesn't handle this for you (deleting from multiple indexes on a single alias).

:+1:

thrall/app/lib/ThrallMessageConsumer.scala
kenoir: I've found `for` comprehensions can cloud what's happening underneath. I'm not sure why you can't `flatMap` here, what was the problem you were finding?

PR #1706, opened 2016-02-02
Keyboard shortcuts for discrete scrolling of results

theefer:
Adds the following shortcuts:

image

These shortcuts work from anywhere on the results page, including if the search field is focused. Previously, only Page Up/Down worked when the search input field was focused.

Rows and pages are scrolled discretely and the top-row is always aligned. When scrolling pages, we scroll the next bottom row (the bottom-most row that isn't fully displayed on the current page) to the top.

recorded

cc @JonnyWeeks 

Will also show to Fiona & Roger, who asked for this about a century ago.

### DIFF
--- a/kahuna/public/js/components/gu-lazy-table-shortcuts/gu-lazy-table-shortcuts.js
+++ b/kahuna/public/js/components/gu-lazy-table-shortcuts/gu-lazy-table-shortcuts.js
@@ -6,7 +6,9 @@ export var lazyTableShortcuts = angular.module('gu.lazyTableShortcuts', [
     'gr.keyboardShortcut'
 ]);
 
-lazyTableShortcuts.directive('guLazyTableShortcuts', ['keyboardShortcut', function(keyboardShortcut) {
+lazyTableShortcuts.directive('guLazyTableShortcuts',
+                             ['keyboardShortcut',
+                              function(keyboardShortcut) {
     return {
         restrict: 'EA',
         require: '^guLazyTable',
--- a/kahuna/public/js/components/gu-lazy-table-shortcuts/gu-lazy-table-shortcuts.js
+++ b/kahuna/public/js/components/gu-lazy-table-shortcuts/gu-lazy-table-shortcuts.js
@@ -47,16 +47,15 @@ lazyTableShortcuts.directive('guLazyTableShortcuts',
                     allowIn: ['INPUT'],
                     callback: invoke('scrollNextPage')
                 })
+                // Home/End not allowed in text field as useful for input navigation
                 .add({
                     combo: 'home',
                     description: 'Scroll results to the start',
-                    allowIn: ['INPUT'],
                     callback: invoke('scrollStart')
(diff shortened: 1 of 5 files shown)


kahuna/public/js/components/gu-lazy-table-shortcuts/gu-lazy-table-shortcuts.js
```diff
@@ -0,0 +1,62 @@
...
+
+export var lazyTableShortcuts = angular.module('gu.lazyTableShortcuts', [
+    'gr.keyboardShortcut'
+]);
+
+lazyTableShortcuts.directive('guLazyTableShortcuts', ['keyboardShortcut', function(keyboardShortcut) {
+    return {
+        restrict: 'EA',
+        require: '^guLazyTable',
+        link: function (scope, element, attrs, lazyTableCtrl) {
```
theefer: I quite like this pattern of implementing extra functionality by tapping into public controller methods exposed by a parent directive, which I stole from elsewhere of course.

cc @hmgibson23 would be nice for Scribe + plugins!

paperboyo:
Supea-nice! The only thing would be

> including if the search field is focused

I would exclude home/end here, which should work as in any text fields - moving to start/end of input  (selecting the text when shift added). Although, it may be more useful (if less standard) to go your way...

theefer:
We could easily exclude Home/End from the input field, I just added them for consistency. I don't mind.

Anybody else got an opinion on this?

jamesgorrie:
> Anybody else got an opinion on this?

I definitely use those a lot on a mac in text fields.

kahuna/public/js/components/gu-lazy-table/gu-lazy-table.js
```diff
@@ -201,9 +234,9 @@ lazyTable.controller('GuLazyTableCtrl', ['range', function(range) {
 }]);
 
 
-lazyTable.directive('guLazyTable', ['$window', 'observe$',
+lazyTable.directive('guLazyTable', ['$window', 'observe$', '$document',
```
jamesgorrie: Not needed? Assuming it was used before the [element](https://github.com/guardian/grid/pull/1706/files#diff-e67d7a61b905b41b4f3d88c228e1ef95R293).

theefer: Oops yes ta, that was from a WIP version, gone now.

jamesgorrie:
If my :+1: still counts, :+1:

theefer:
I removed Home/End from INPUT, thanks @jamesgorrie @paperboyo 

@jamesgorrie of course it does :sparkling_heart:

akash1810:
Didn't `home` and `end` work like this already? i.e. didn't the browser scroll to the start/end of the page already?

paperboyo:
Of course, this only pertains to the search box input, all other input fields will work as usual, of course?

PR #1708, opened 2016-02-03
SSL notes

akash1810:
Add and trust our CA to all the places to ensure SSL works.

Related - https://github.com/guardian/dev-nginx/pull/18


akash1810:
Closing as it feels more appropriate to include these notes in `dev-nginx` (which it has been).

Issue #1192, opened 2015-09-04
Respect filename when downloading from single image view


jamesgorrie:
Due to the fact that chrome (the only browser that lets you download cross domain) doesn't allow you to set the filename cross domain.

It would be good for this to work.

Issue #1707, opened 2016-02-02
single image download in FF broken


akash1810:
As noted my Mark on Guide.

Downloading a single image in firefox does not trigger a file download, it opens the image in a new tab instead.

Looking at the [docs](https://developer.mozilla.org/en/docs/Web/HTML/Element/a), a possible reason is:

> This attribute is only honored for links to resources with the same-origin.

Currently we're linking to an S3 object, on an S3 domain.

akash1810:
Fixed in #1709

paperboyo:
:gem:

PR #1709, opened 2016-02-03
set content disposition header on S3 image urls

akash1810:
Force browser to trigger a download when url is opened.

Does not affect use of url in an ``.

Fixes https://github.com/guardian/grid/issues/1192 and https://github.com/guardian/grid/issues/1707.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
@@ -1,15 +1,16 @@
 package com.gu.mediaservice.lib.aws
 
-import java.util.Date
-import java.io.{FileInputStream, File}
+import java.io.File
 import java.net.URI
-import scala.concurrent.{ExecutionContext, Future}
-import scala.collection.JavaConverters._
 
 import com.amazonaws.auth.AWSCredentials
-import com.amazonaws.services.s3.{AmazonS3Client, AmazonS3}
-import com.amazonaws.services.s3.model.{S3ObjectSummary, ObjectMetadata, PutObjectRequest, GeneratePresignedUrlRequest, ListObjectsRequest}
+import com.amazonaws.services.s3.model._
+import com.amazonaws.services.s3.{AmazonS3, AmazonS3Client}
+import com.gu.mediaservice.model.Image
 import org.joda.time.DateTime
+
+import scala.collection.JavaConverters._
+import scala.concurrent.{ExecutionContext, Future}
 import scalaz.syntax.id._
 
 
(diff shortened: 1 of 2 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
```diff
@@ -29,10 +30,19 @@ class S3(credentials: AWSCredentials) {
...
     val key: Key = url.getPath.drop(1)
-    val request = new GeneratePresignedUrlRequest(bucket, key).withExpiration(expiration.toDate)
+
+    val filename: String = image.uploadInfo.filename match {
+      case Some(f) =>   s"${f.substring(0, f.lastIndexOf('.'))} (${image.id}).jpeg"
+      case _       =>   s"${image.id}.jpeg"
+    }
+
+    val headers = new ResponseHeaderOverrides()
+    headers.setContentDisposition("attachment; filename=\"" + filename + "\"")
```
akash1810: Couldn't get string interpolation to work here, was getting:

```console
media-api_1 | [error] /code/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala:43: ')' expected but string literal found.
media-api_1 | [error]     headers.setContentDisposition(s"attachment; filename=\"$filename\"")
media-api_1 | [error]                                                                      ^
media-api_1 | [error] one error found
```

:cry:

kenoir:
:+1:

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
```diff
@@ -29,10 +30,19 @@ class S3(credentials: AWSCredentials) {
...
     new AmazonS3Client(credentials) <| (_ setEndpoint s3Endpoint)
 
-  def signUrl(bucket: Bucket, url: URI, expiration: DateTime): String = {
+  def signUrl(bucket: Bucket, url: URI, image: Image, expiration: DateTime): String = {
     // get path and remove leading `/`
     val key: Key = url.getPath.drop(1)
-    val request = new GeneratePresignedUrlRequest(bucket, key).withExpiration(expiration.toDate)
+
+    val filename: String = image.uploadInfo.filename match {
+      case Some(f) =>   s"${f.substring(0, f.lastIndexOf('.'))} (${image.id}).jpeg"
```
theefer: There is technically no guarantee that there is a `.` in a filename. A RegEx might be safer here?

theefer: Also why `jpeg` and not `jpg`?

akash1810: > A RegEx might be safer here?

Yep - https://github.com/guardian/grid/pull/1710 :crying_cat_face:

PR #1711, opened 2016-02-03
set content disposition header on S3 image urls... again

akash1810:
Reverts https://github.com/guardian/grid/pull/1710, fixing the issue with a regex.

Original PR - https://github.com/guardian/grid/pull/1709

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
@@ -1,15 +1,16 @@
 package com.gu.mediaservice.lib.aws
 
-import java.util.Date
-import java.io.{FileInputStream, File}
+import java.io.File
 import java.net.URI
-import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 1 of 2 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
```diff
@@ -29,10 +30,24 @@ class S3(credentials: AWSCredentials) {
...
     val key: Key = url.getPath.drop(1)
-    val request = new GeneratePresignedUrlRequest(bucket, key).withExpiration(expiration.toDate)
+
+    val filename: String = image.uploadInfo.filename match {
+      case Some(f)  => s"${removeExtension(f)} (${image.id}).jpg"
+      case _        => s"${image.id}.jpg"
+    }
+
+    val headers = new ResponseHeaderOverrides()
+    headers.setContentDisposition("attachment; filename=\"" + filename + "\"")
```
theefer: Note that you can use string interpolation with `s"""attachnment; filename="$filename""""`.

theefer:
:+1:

PR #1712, opened 2016-02-03
Display by date added to collection

tsop14:
- Sorts images in collection using date field.
- If in more than one collection, it takes the most recent date value. Therefore has an incorrect order for images in more than one collection. e.g. in .gif: when image added to 'Travel' position moves to top of the 'TV' collection. 
- If click clear search or click on label returns to 'newest first' sort. 

- Changes Hierarchy field from a class with value and field to an object. 

date-added-to-collection

### DIFF
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
@@ -207,7 +207,7 @@
         <dt class="metadata-line__key">Collections</dt>
         <dd class="metadata-line__info flex-container"
             ng-repeat="collection in ctrl.image.data.collections">
-            <a ui:sref="search.results({query: (collection.data.pathId | queryCollectionFilter), orderBy: 'dateAddedToCollection'})">
+            <a ui:sref="search.results({query: (collection.data.pathId | queryCollectionFilter)})">
                 {{collection.data.path.join(' ▸ ')}}
             </a>
             <span class="flex-spacer"></span>
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
@@ -207,7 +207,7 @@
         <dt class="metadata-line__key">Collections</dt>
         <dd class="metadata-line__info flex-container"
             ng-repeat="collection in ctrl.image.data.collections">
-            <a ui:sref="search.results({query: (collection.data.pathId | queryCollectionFilter)})">
+            <a ui:sref="search.results({query: (collection.data.pathId | queryCollectionFilter), orderBy: 'dateAddedToCollection'})">
                 {{collection.data.path.join(' ▸ ')}}
             </a>
             <span class="flex-spacer"></span>
--- a/kahuna/public/js/edits/labeller-compact.html
+++ b/kahuna/public/js/edits/labeller-compact.html
@@ -8,7 +8,7 @@
 
             <a class="label__value label__value--compact label__link"
                ng:if="!ctrl.disabled"
-               ui:sref="search.results({query: (label.data | queryLabelFilter)})">
+               ui:sref="search.results({query: (label.data | queryLabelFilter), orderBy: ctrl.orderBy})">
                 {{label.data}}
             </a>
         </li>
--- a/kahuna/public/js/edits/labeller-compact.html
+++ b/kahuna/public/js/edits/labeller-compact.html
@@ -8,7 +8,7 @@
 
             <a class="label__value label__value--compact label__link"
                ng:if="!ctrl.disabled"
-               ui:sref="search.results({query: (label.data | queryLabelFilter), orderBy: ctrl.orderBy})">
+               ui:sref="search.results({query: (label.data | queryLabelFilter)})">
                 {{label.data}}
             </a>
         </li>
--- a/kahuna/public/js/edits/labeller.js
+++ b/kahuna/public/js/edits/labeller.js
@@ -12,13 +12,15 @@ export var labeller = angular.module('kahuna.edits.labeller', [
 ]);
 
 labeller.controller('LabellerCtrl',
-                  ['$rootScope', '$scope', '$window', '$timeout', 'labelService', 'onValChange',
-                   function($rootScope, $scope, $window, $timeout, labelService, onValChange) {
+                  ['$rootScope', '$scope', '$window', '$timeout', '$stateParams', 'labelService', 'onValChange',
+                   function($rootScope, $scope, $window, $timeout, $stateParams, labelService, onValChange) {
     var ctrl = this;
 
     $scope.$watch(() => ctrl.image.data.userMetadata.data.labels, onValChange(newLabels => {
         ctrl.labels = newLabels;
     }));
+
+    ctrl.orderBy = $stateParams.orderBy === 'oldest' ? 'oldest' : undefined;
 
     ctrl.labels = ctrl.image.data.userMetadata.data.labels;
 
--- a/kahuna/public/js/edits/labeller.js
+++ b/kahuna/public/js/edits/labeller.js
@@ -34,8 +34,6 @@ labeller.controller('LabellerCtrl', [
         ctrl.labels = newLabels;
     }));
 
-    ctrl.orderBy = $stateParams.orderBy === 'oldest' ? 'oldest' : undefined;
-
     ctrl.labels = ctrl.image.data.userMetadata.data.labels;
 
     function saveFailed() {
--- a/kahuna/public/js/edits/labeller.js
+++ b/kahuna/public/js/edits/labeller.js
@@ -11,9 +11,23 @@ export var labeller = angular.module('kahuna.edits.labeller', [
     'kahuna.services.label'
 ]);
 
-labeller.controller('LabellerCtrl',
-                  ['$rootScope', '$scope', '$window', '$timeout', '$stateParams', 'labelService', 'onValChange',
-                   function($rootScope, $scope, $window, $timeout, $stateParams, labelService, onValChange) {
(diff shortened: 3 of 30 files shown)


theefer:
> If in more than one collection, it takes the most recent date value.

Is it worth confirming that we will be able to use "nested docs" to fix this issue before releasing this?

kahuna/public/js/edits/labeller-compact.html
```diff
@@ -8,7 +8,7 @@
 
             <a class="label__value label__value--compact label__link"
                ng:if="!ctrl.disabled"
-               ui:sref="search.results({query: (label.data | queryLabelFilter)})">
+               ui:sref="search.results({query: (label.data | queryLabelFilter), orderBy: ctrl.orderBy})">
```
theefer: Isn't this redundant (i.e. can be removed)?

theefer: Ah no, ignore me, reading on..

kahuna/public/js/search/query.js
```diff
@@ -82,6 +82,9 @@ query.controller('SearchQueryCtrl',
     }
 
     $scope.$watchCollection(() => ctrl.filter, onValChange(filter => {
+        if (filter.query.indexOf('~') === 0) {
```
theefer: This is quite simplistic (if `~` is in a word or a quoted phrase, it will incorrectly detect it as a collection filter). We should at least add a comment to that effect. Or maybe we should only detect it as a collection if the query starts with `~`?

kahuna/public/js/search/query.html
```diff
@@ -75,7 +75,7 @@
             <span class="search__modifier-toggle__text">Sort order</span>
         </button>
 
-        <ul class="search__modifier order__option" ng:class="orderShown ? 'order__option--show' : 'order__option--hide'">
+        <ul ng:hide="searchQuery.filter.orderBy === 'dateAddedToCollection'" class="search__modifier order__option" ng:class="orderShown ? 'order__option--show' : 'order__option--hide'">
```
theefer: As a rule, shall we use `ng:if` instead of `ng:show`/`ng:hide`? Tends to avoid weird bugs.

kahuna/public/js/search/query.js
```diff
@@ -95,6 +98,7 @@ query.controller('SearchQueryCtrl',
 
     function resetQueryAndFocus() {
         ctrl.filter.query = '';
+        ctrl.filter.orderBy = undefined;
```
theefer: This and the parameter in the label link seems rather brittle. There are many more links to search.results that would preserve the orderBy and possibly cause confusion. Could we find a way to programmatically clear that flag whenever we navigate to a search state that has no collection in the query?

kahuna/public/js/search/query.js
tsop14: > Or maybe we should only detect it as a collection if the query starts with ~?

I think this only checks if the first character of the query is a '~'. 

If a quote/word begins with '~'  it could still be used as a non-collection query if it was wrapped in quote marks e.g. "~said Alice".

Still, as is, it wont work for a search term "queen ~g2/sweden"

media-api/app/lib/elasticsearch/QueryBuilder.scala
```diff
@@ -33,7 +34,10 @@ class QueryBuilder(matchFields: Seq[String]) {
       case Phrase(value) => matchPhraseQuery(field, value)
       case DateRange(start, end) => rangeQuery(field).from(start.toString()).to(end.toString())
     }
-    case HierarchyField(field, value) => termQuery(field, value)
+    case HierarchyField => condition.value match {
+      case Phrase(value) => termQuery(getFieldPath("pathHierarchy"), value)
+      case _ => throw InvalidQuery("Cannot accept non-Phrase value for HierarchyField Match")
```
tsop14: `case _` should never run because we've set HierarchyField to always have a condition.value that is a `Phrase(value)`.

kahuna/public/js/search/query.js
theefer: Doh you're right, I can't read :+1:

media-api/app/lib/elasticsearch/sorts.scala
```diff
@@ -1,17 +1,44 @@
...
     for ((field, order) <- sorts) builder.addSort(field, order)
     builder
   }
 
+  def addedToCollectionTimeSort(query: List[Condition])(builder: SearchRequestBuilder): SearchRequestBuilder = {
+    //TODO get relevant values from query
+    val pathHierarchyOpt = query.map {
+      case Match(HierarchyField, Phrase(value)) => Some(value)
+      case _ => None
+    }.flatten.headOption
```
tsop14: If something is a `Match(HierarchyField, Phrase(value))` then it has to be a collection query. This is because this type is only constructed once we have identified it as a collection query. So we use it to pattern match and return the pathId it contains.

media-api/app/lib/querysyntax/QuerySyntax.scala
```diff
@@ -24,7 +24,7 @@ class QuerySyntax(val input: ParserInput) extends Parser with ImageFields {
   def ScopedMatch = rule { MatchField ~ ':' ~ MatchValue }
   def HashMatch = rule      { '#' ~ MatchValue      ~> (label      => Match(SingleField(getFieldPath("labels")), label)) }
   // All hierarchy values are lowercase as specified in the analyser
-  def CollectionRule = rule { '~' ~ ExactMatchValue ~> (collection => Match(HierarchyField(getFieldPath("pathHierarchy"), collection.string.toLowerCase), collection)) }
+  def CollectionRule = rule { '~' ~ ExactMatchValue ~> (collection => Match(HierarchyField, Phrase(collection.string.toLowerCase))) }
```
tsop14: This is a simplification to the code. Given a HierarchyField will only ever be used for collection queries (i.e. those that pass the collection rule), then we do not need to set a field path "pathHierarchy" and pass that through as we can imply it when we match on a HierarchyField [here]( https://github.com/guardian/grid/pull/1712/files#diff-690e3390cd34f6bbe4c278bd7348d7a2R37).

paperboyo:
Just some **loose** thoughts :spaghetti:

Currently, I can sort by Oldest first when in the collection - this will be impossible after this goes out, right? I would prefer sort orders to be a dropdown that would automatically change to Added order when looking at the collection but be possible to change. But then, I would like selections to survive changing sorting order as well :wink:

It may not look like that useful, until we get all these other sorting options and more (Usage!).

sort-orders

tsop14:
A button with sorting-button styling appears on collection-results page; this should make it clearer what is happening. 😑

screen shot 2016-02-04 at 18 27 37

Agree sorting by oldest/newest should be added but in a separate PR.

paperboyo:
> button with sorting-button styling appears on collection-results page

Suppa-nice!

media-api/app/lib/elasticsearch/QueryBuilder.scala
jamesgorrie: Does the compiler not know this?

kenoir: @jamesgorrie no, the base trait is `Condition` so the compiler can't tell the difference.

kenoir:
:+1: 

I think we're going to have to thank about how the client handles sort order when in other "modes" i.e. filtered by collection.

kahuna/public/js/search/index.js
```diff
@@ -267,4 +267,15 @@ search.run(['$rootScope', '$state', function($rootScope, $state) {
             $state.go('search.results', null, {reload: true});
         }
     });
+    $rootScope.$on('$stateChangeStart', (_, toState, toParams) => {
+        const collectionQuery = toParams.query && toParams.query.indexOf('~') === 0;
+        if (toState.name === 'search.results' && collectionQuery) {
+            toParams.orderBy = 'dateAddedToCollection';
+        }
+        else {
```
theefer: Shouldn't this check the `toState.name` too? Might be worth nesting the whole thing within that condition, since it's the scope required for this search controller.

kahuna/public/js/search/index.js
```diff
@@ -267,4 +267,15 @@ search.run(['$rootScope', '$state', function($rootScope, $state) {
             $state.go('search.results', null, {reload: true});
         }
     });
+    $rootScope.$on('$stateChangeStart', (_, toState, toParams) => {
```
theefer: Would be nice to add a bit of documentation here as to what this is and why.

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
```diff
@@ -207,7 +207,7 @@
         <dt class="metadata-line__key">Collections</dt>
         <dd class="metadata-line__info flex-container"
             ng-repeat="collection in ctrl.image.data.collections">
-            <a ui:sref="search.results({query: (collection.data.pathId | queryCollectionFilter)})">
+            <a ui:sref="search.results({query: (collection.data.pathId | queryCollectionFilter), orderBy: 'dateAddedToCollection'})">
```
theefer: Is this still necessary?

kahuna/public/js/edits/labeller.js
```diff
@@ -11,9 +11,23 @@ export var labeller = angular.module('kahuna.edits.labeller', [
...
+    'onValChange',
+
+   function($rootScope,
+            $scope,
+            $window,
+            $timeout,
+            $stateParams,
+            labelService,
+            onValChange) {
+
```
theefer: Change (and added `$stateParams` dependency) not used here?

theefer:
A few comments but looks good!

tsop14:
> Is it worth confirming that we will be able to use "nested docs" to fix this issue before releasing this?

That is the plan.

PR #1713, opened 2016-02-03
Use new Library icons (and refactor everything)

theefer:
This brings in the new nicer Library icons (add, remove, added, kept) from @akemitakagi to replace the confusing and overused Material Design "Lock" icons.

image
image
image
image
image
image
image
image

I've taken the opportunity to refactor a lot of the messy code around the multiple directives responsible for showing and editing Library status, removing one in the process and sharing more logic.

There are still two directives, I've left them that way as they have slightly different semantics. The one with the icon+text is used in the action bar very much as an *action button*. The other with only icon is used as a *state*, which can also be toggled on click.


akemitakagi:
Icons looks good :+1:

akash1810:
:+1:

PR #1714, opened 2016-02-03
force newline at end of genereated properies file

akash1810:
this is probably cheating a little :stuck_out_tongue_closed_eyes: 

cc @jamesgorrie


jamesgorrie:
Looks neater in configs.sh, works.

Maybe worth noting why we force the whitespace, so if someone wants to remove it they can try fix that first.

jamesgorrie:
:+1:

akash1810:
> Maybe worth noting why we force the whitespace

Done, though I can't find any documentation about the problem :neutral_face:

PR #1716, opened 2016-02-04
Revert "set content disposition header on S3 image urls... again"

akash1810:
Reverts guardian/grid#1711

D'oh, the filename in the content disposition header needs escaping! We're seeing broken images with this response from the S3 request:

```xml
<Error>
<Code>InvalidArgument</Code>
<Message>Header value cannot be represented using ISO-8859-1.</Message>
<ArgumentName>response-content-disposition</ArgumentName>
<ArgumentValue>attachment; filename="Chris Evans &amp; Father Bryan D’Arcy (46dfc37339848cc7369653c94fe83cbd8a28ae40).jpg"</ArgumentValue>
<RequestId>A134CC749F8CB3A9</RequestId>
<HostId>jqMbu71GrLmkQsX6xZ/iYusKbhWo8AdLWg41kxvMcv7rozl8L5iVo/2X7H8N4/ZrbyS37OPuMSE=</HostId>
</Error>
```

:sob:

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
@@ -1,16 +1,15 @@
 package com.gu.mediaservice.lib.aws
 
-import java.io.File
+import java.util.Date
+import java.io.{FileInputStream, File}
 import java.net.URI
+import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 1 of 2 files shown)


akash1810:
The fix would be to make the header `Content-Disposition: attachment; filename*=utf-8''<filename>` per http://tools.ietf.org/html/rfc6266#section-5

PR #1717, opened 2016-02-04
Hide ui-archiver if unarchived unless hovering

theefer:
We don't want to show it all the time, as is now the case on TEST.

### DIFF
--- a/kahuna/public/stylesheets/main.css
+++ b/kahuna/public/stylesheets/main.css
@@ -833,14 +833,15 @@ input.search-query__input {
     color: #00adee;
 }
 
-.result ui-archiver .archiver--unarchived {
+/* Hacky: Hide ui-archiver "Add to Library" (unarchived state) unless hovering */
+.result .ui-archiver--unarchived {
     display: none;
 }
 
 .result:hover .result__select,
 .result:hover .preview__fade,
 .result:hover .image-actions,
-.result:hover .archiver {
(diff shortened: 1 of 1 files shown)


jamesgorrie:
Not sure if it's that great an idea - [but I wrote this for this type of stuff](https://github.com/guardian/grid/blob/master/kahuna/public/stylesheets/main.css#L388-L393).

theefer:
The problem here, and the reason for the hack, is that the parent (the element on which we want the `:hover`) is in a separate template/directive from the child (which we want to show/hide based on the specific state of the child).

Using `hover-parent`/`hover-child` across such boundaries feels just as brittle as my hack? There is no guarantee there won't be other `hover-parent` interfering in between them, or explicit reference from one to the other to make things more explicit :cry: 

I can't really think of a way to do this better without JS (passing the state to the child, which can then deal with it explicitly), which feels overkill...

Debate/suggestions welcome though!

jamesgorrie:
>  feels just as brittle as my hack

Yup

> There is no guarantee there won't be other hover-parent interfering

It might be worth being more specific to direct children with that CSS thinking about it.

You could have something like:
``` HTML
<gr-image class="hover ">
  <!-- stuff and more stuff -->
  <div class="hover-child">
    <gr-archiver></gr-archiver>
  </div>
</gr-image>
```

This wouldn't work when it is only a part of the child component that you want to show/hide. Trying to think of a way around that.

PR #1718, opened 2016-02-04
Rename ui-archiver everywhere for consistency

theefer:
I realised the directive and classes were still called `ui-archiver` even thought I'd renamed the component to `gr-archiver-status`. This makes things consistent.

No functionality change.


jamesgorrie:
grrrr :tiger: :+1:

PR #1719, opened 2016-02-04
Persist images that have been used

theefer:
Need to do further checks to confirm behaviour, once I've figured out how to get usage running and inject some test usages.

### DIFF
--- a/kahuna/public/js/services/image-logic.js
+++ b/kahuna/public/js/services/image-logic.js
@@ -34,6 +34,8 @@ imageLogic.factory('imageLogic', ['imageAccessor', function(imageAccessor) {
             switch (reason) {
             case 'exports':
                 return 'cropped';
+            case 'usages':
+                return 'used';
             case 'persistence-identifier':
                 return 'from Picdar';
             case 'photographer-category':
--- a/media-api/app/controllers/MediaApi.scala
+++ b/media-api/app/controllers/MediaApi.scala
@@ -157,11 +157,8 @@ object MediaApi extends Controller with ArgoHelpers {
       case Some(source) =>
         val image = source.as[Image]
 
-        val hasExports = ImageResponse.imagePersistenceReasons(image).contains("exports")
-
-        if (hasExports) {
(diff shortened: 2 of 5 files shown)


akash1810:
Yay! We're closer to removing the permission layer from crop deletion!

theefer:
I don't think we will ever be able to do this safely tbh. We know crops are sometimes used outside of the normal Composer flow, in which case usage wouldn't be tracked, and we'd risk to invalidate an image that is still in use (e.g. fronts, interactives, social, etc)...

akash1810:
Oh :cry:

paperboyo:
Just a question: do you think we may ever be able to track fronts usage (and, maybe, interactives)? E.g. cutouts on fronts are fronts-only, but they travel to the apps, I think... Interactives would be harder, unless we find out _a way_ for them to use images (imgIX, tracking etc.)/

theefer:
Done some checks and fixed an error in the API where it was still possible to issue a DELETE. Should be ready to go out once +1'd.

> do you think we may ever be able to track fronts usage

It's definitely possible, there's just not been much interest so far.

> (and, maybe, interactives)?

When talking about images used by interactives outside the normal framework (e.g. they hardcode the asset URIs, instead of recording them into and reading them out of CAPI/the body), it'd likely require manual input. Would be good to assess value, and incentives (why would people take the pain to do it). Rights being an issue of course.

cc @blishen

paperboyo:
> Rights being an issue of course.

Also can sweeten the deal if that would help the use imgIX power...

akash1810:
:+1:

theefer:
Technically people could still hack in ImgIX support without images properly being in CAPI. It's very unlikely they'd do it properly though, given the complexity (master vs resizes, etc).

paperboyo:
Yeah. I can think of other gains from tracking usages more thorougly, like a future crop replacement to be able to affect fronts image replacement (also on video cards), searching for all current/historic cutouts by front or such like. I agree they may be a bit inflated/niche...

PR #1720, opened 2016-02-04
Fix persistence explanation when used for single image

theefer:
The component is used on the individual image preview page, where it doesn't make sense to talk about "the selected images" and join the reasons with "or". This tunes the logic to correctly phrase the explanation on why an image or multiple images have been persisted.


akash1810:
:+1:

PR #1721, opened 2016-02-05
Set content disposition header on S3 urls, supporting UTF-8

akash1810:
Fixes issues that caused https://github.com/guardian/grid/pull/1716

The `content-disposition` header spec supports the UTF-8 character set, however not all browsers implement it, so keep the fallback.

Tested with files named `boring.jpg` and `éxãmplę üpłøåd.jpg` in Chrome and FF with expected results!

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
@@ -57,7 +57,7 @@ class S3(credentials: AWSCredentials) {
     val key: Key = url.getPath.drop(1)
 
     // use both `filename` and `filename*` parameters for compatibility with user agents not implementing RFC 5987
-    // they'll fallback to `filename`, which will be a UTF-8 string decoded as Latin-1
+    // they'll fallback to `filename`, which will be a UTF-8 string decoded as Latin-1 - this is a rubbish string, but only rubbish browsers don't support RFC 5987 (IE8 back)
     // See http://tools.ietf.org/html/rfc6266#section-5
     val contentDisposition = s"""attachment; filename="${getContentDispositionFilename(image, CharSet.ISO8859)}"; filename*=UTF-8''${getContentDispositionFilename(image, CharSet.UTF8)}"""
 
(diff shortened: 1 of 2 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
```diff
@@ -29,10 +31,38 @@ class S3(credentials: AWSCredentials) {
...
+  }
+
+  def signUrl(bucket: Bucket, url: URI, image: Image, expiration: DateTime): String = {
     // get path and remove leading `/`
     val key: Key = url.getPath.drop(1)
-    val request = new GeneratePresignedUrlRequest(bucket, key).withExpiration(expiration.toDate)
+
+    // use both `filename` and `filename*` parameters for compatibility with user agents not implementing RFC 5987 (they'll fallback to `filename`)
+    // See http://tools.ietf.org/html/rfc6266#section-5
+    val contentDisposition = s"""attachment; filename="${getContentDispositionFilename(image, CharSet.ISO8859)}"; filename*=UTF-8''${getContentDispositionFilename(image, CharSet.UTF8)}"""
```
theefer: What are browsers not supporting RFC 5987 supposed to get? Looks like a UTF-8 string decoded as Latin-1, or am I missing something?

akash1810: > What are browsers not supporting RFC 5987 supposed to get?

They're supposed to get a valid ISO-8859-1 string.

> Looks like a UTF-8 string decoded as Latin-1

Yep.

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
```diff
@@ -29,10 +31,39 @@ class S3(credentials: AWSCredentials) {
...
+    }
+  }
+
+  def signUrl(bucket: Bucket, url: URI, image: Image, expiration: DateTime): String = {
     // get path and remove leading `/`
     val key: Key = url.getPath.drop(1)
-    val request = new GeneratePresignedUrlRequest(bucket, key).withExpiration(expiration.toDate)
+
+    // use both `filename` and `filename*` parameters for compatibility with user agents not implementing RFC 5987
+    // they'll fallback to `filename`, which will be a UTF-8 string decoded as Latin-1
```
theefer: Can we leave a note saying that yes it will be garbage, just garbage that the browsers will accept?

akash1810: done.

theefer:
:+1:

PR #1722, opened 2016-02-05
clearer "back to search" button

akash1810:
Guide requested the ability to be able to go back to a search within the embedded Grid (when used standalone, they use browser navigation).

I mentioned this button goes back to the search, which they were happy to hear! This PR makes it more obvious.

screen shot 2016-02-05 at 12 34 42

### DIFF
--- a/kahuna/public/js/image/view.html
+++ b/kahuna/public/js/image/view.html
@@ -1,7 +1,7 @@
 <gr-top-bar>
     <gr-top-bar-nav>
         <a class="top-bar-item side-padded clickable" ui:sref="search" title="Back to search">
-            <gr-icon-label gr-icon="search">Search</gr-icon-label>
+            <gr-icon-label gr-icon="youtube_searched_for">Back to search</gr-icon-label>
         </a>
     </gr-top-bar-nav>
 


kahuna/public/js/image/view.html
```diff
@@ -1,7 +1,7 @@
 <gr-top-bar>
     <gr-top-bar-nav>
-        <a class="top-bar-item side-padded clickable" ui:sref="search" title="Back to search">
-            <gr-icon-label gr-icon="search">Search</gr-icon-label>
+        <a class="top-bar-item side-padded clickable" ui:sref="search" gr:tooltip="Back to search" gr:tooltip-position="right">
+            <gr-icon-label gr-icon="youtube_searched_for">Back to search</gr-icon-label>
```
akash1810: super intuitive icon name...

kahuna/public/js/image/view.html
```diff
@@ -1,7 +1,7 @@
 <gr-top-bar>
     <gr-top-bar-nav>
-        <a class="top-bar-item side-padded clickable" ui:sref="search" title="Back to search">
-            <gr-icon-label gr-icon="search">Search</gr-icon-label>
+        <a class="top-bar-item side-padded clickable" ui:sref="search" gr:tooltip="Back to search" gr:tooltip-position="right">
```
theefer: Is there really value in a tooltip that exactly repeats the label of the button? Or is that for the corner case of narrow browsers where the button was collapsed to the icon?

akash1810: Not really - removed. The `title` attribute will still display in the corner case mentioned.

akash1810:
Tooltip removed:

screen shot 2016-02-05 at 13 19 34

paperboyo:
Clever icon! :smiley:

tsop14:
:+1:

PR #1725, opened 2016-02-08
Nested Usage & Date constraint

kenoir:
In order to allow us to query pending usage by supplier, we need to perform nested queries in Elasticsearch on Usage.

This PR allows queries like: `usages@status:pending usages@>added:2014-01-28 supplier:"Getty Images"`

To result in ES query like:

```json
{
  "bool" : {
    "must" : [ {
      "match" : {
        "usageRights.supplier" : {
          "query" : "Getty Images",
          "type" : "phrase"
        }
      }
    }, {
      "nested" : {
        "query" : {
          "bool" : {
            "must" : [ {
              "match" : {
                "usages.status" : {
                  "query" : "pending",
                  "type" : "boolean",
                  "operator" : "AND"
                }
              }
            }, {
              "range" : {
                "dateAdded" : {
                  "from" : "1970-01-01T01:00:00.000+01:00",
                  "to" : "2016-01-28T00:00:00.000Z",
                  "include_lower" : true,
                  "include_upper" : true
                }
              }
            } ]
          }
        },
        "path" : "usages"
      }
    } ]
  }
}
```

Depends: https://github.com/guardian/grid/pull/1760

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
@@ -26,7 +26,20 @@ object Mappings {
 
   val dynamicObj = Json.obj("type" -> "object", "dynamic" -> true)
 
-  def nonDynamicObj(obj: (String, JsValueWrapper)*) = Json.obj("type" -> "object", "dynamic" -> "strict", "properties" -> Json.obj(obj:_*))
+  val nestedType = Json.obj(
+    "type" -> "nested",
+    "include_in_parent" -> true
+  )
+
+  def nonDynamicObjWithAttrs(obj: (String, JsValueWrapper)*)(attrs: JsObject = Json.obj()): JsObject =
+    nonDynamicObj(obj:_*) ++ attrs
+
+  def nonDynamicObj(obj: (String, JsValueWrapper)*) =
+    Json.obj(
+      "type" -> "object",
+      "dynamic" -> "strict",
+      "properties" -> Json.obj(obj:_*)
+    )
 
   def nonAnalysedList(indexName: String) = Json.obj("type" -> "string", "index" -> "not_analyzed", "index_name" -> indexName)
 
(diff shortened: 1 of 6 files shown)


media-api/app/lib/querysyntax/QuerySyntax.scala
```diff
@@ -5,29 +5,78 @@ import org.parboiled2._
 
 import com.gu.mediaservice.lib.elasticsearch.ImageFields
```
theefer: Could the tests be updated to reflect the new syntax?

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -41,13 +51,13 @@ object Mappings {
     nonDynamicObj(
       // TODO: extract these to a configuration setting
       "picdarUrn" -> standardAnalysedString
-    )
+    )()
```
theefer: Bit :hankey: isn't it? Can `attrs` not be made into a regular trailing parameter with default value because of the varargs `*` first parameter? Is it worth defining another function similar to `nonDynamicObj` that only takes one argument so you don't have to pollute every single use of it with an extra `()` when that extra param is only used once?

media-api/app/lib/querysyntax/QuerySyntax.scala
```diff
@@ -5,29 +5,78 @@ import org.parboiled2._
...
 
   def Filter = rule {
     ScopedMatch ~> Match | HashMatch | CollectionRule |
-    DateMatch ~> Match | AtMatch |
+    DateRangeMatch ~> Match | AtMatch |
     AnyMatch
   }
 
+  def NestedMatch = rule { ParentField ~ "@" ~ NestedField ~ ':' ~ MatchValue }
+  def NestedDateMatch = rule { ParentField ~ "@" ~ DateConstraint ~ DateMatch ~> (
```
theefer: I'm a bit confused by this rule, could you explain?

kenoir: Can we reuse this outside of `Nested`

media-api/app/lib/elasticsearch/QueryBuilder.scala
```diff
@@ -42,11 +43,29 @@ class QueryBuilder(matchFields: Seq[String]) extends ImageFields {
...
+        case n: Nested => true
+        case _ => false
+      }
+
+      nested.groupBy {
+        case n: Nested => n.parentField
+        case _ => Unit
+      }.map {
+        case (parentField, nested) => nested.map {
+            case Nested(_, f, v) => println(s"$parentField : $f: $v")
```
kenoir: Needs to actually construct a query!

media-api/app/lib/querysyntax/QuerySyntax.scala
kenoir: Yup.

PR #1728, opened 2016-02-08
Adds riffraff artefact upload plugin

kenoir:
This will enable us to push artefacts from a non-Teamcity CI environment  to RiffRaff.

p.s. Jenkins

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -23,6 +23,8 @@ object Build extends Build {
     Seq(
       riffRaffPackageType := (packageZipTarball in Universal).value,
       riffRaffBuildIdentifier := env("BUILD_NUMBER").getOrElse("DEV"),
+      riffRaffPackageName := s"media-service::${name.value}",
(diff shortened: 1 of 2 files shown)


project/Build.scala
```diff
@@ -13,12 +17,22 @@ import AssemblyKeys._
...
+      riffRaffUploadManifestBucket := Option("riffraff-builds")
+    )
+
   val commonSettings =
     Seq(
       scalaVersion := "2.11.6",
       scalaVersion in ThisBuild := "2.11.6",
       organization := "com.gu",
-      fork in run := true,
+      fork in run  := true,
```
theefer: Thank God for that :+1:

theefer:
Is this the same as what TC does? Does it cause double upload when in TC, or is there some magic that prevents it?

kenoir:
@theefer we don't run the correct recipe in TC at the moment to cause this to try and upload to s3 using the plugin.

Tested and works in Jenkins.

theefer:
:+1:

PR #1729, opened 2016-02-08
Add new auth service to take over Panda-issuing role from kahuna

theefer:
This PR just adds a new auth microservice. It's not being used yet, and the next stage will be to change kahuna to auth via the auth service instead of kahuna.

Needs to be built and artifacted before https://github.com/guardian/grid-infra/pull/118 (CF change) can be applied.

### DIFF
--- a/auth/app/lib/Config.scala
+++ b/auth/app/lib/Config.scala
@@ -13,9 +13,6 @@ object Config extends CommonPlayAppConfig with CommonPlayAppProperties {
   val rootUri: String = services.authBaseUri
   val mediaApiUri: String = services.apiBaseUri
 
-  val awsCredentials: AWSCredentials =
-    new BasicAWSCredentials(properties("aws.id"), properties("aws.secret"))
-
   private lazy val corsAllowedOrigins = properties.getOrElse("cors.allowed.origins", "").split(",").toList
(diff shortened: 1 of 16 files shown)


kenoir:
:+1:

auth/app/lib/Config.scala
```diff
@@ -0,0 +1,19 @@
+package lib
+
+import com.gu.mediaservice.lib.config.{Properties, CommonPlayAppConfig, CommonPlayAppProperties}
+import com.amazonaws.auth.{BasicAWSCredentials, AWSCredentials}
+
+object Config extends CommonPlayAppConfig with CommonPlayAppProperties {
+
+  val appName = "auth"
+
+  val properties = Properties.fromPath("/etc/gu/auth.properties")
```
akash1810: do you need `panda.properties` too?

akash1810: meep, nevermind, its pulled in via the `PanDomainAuthActions` trait :smile:

akash1810:
:+1:

akash1810:
Just realised there's no update to the docker-compose file :cry:

Can add it after https://github.com/guardian/grid/pull/1730 gets merged.

PR #1730, opened 2016-02-09
upgrade to Docker 1.10 and Compose 1.6

akash1810:
Benefits:
 - With the new `context` directive, the docker-compose.yml file is moved to the more standard project root
 - Docker 1.10 uses an internal DNS, which looks more reliable than injected hosts file; migrate to `depends_on` directive to signify dependancy order
 - No need for a dedicated shared volume container - use the new top level `volumes` directive, mounting `ivy` and `sbt` on the base `sbt` container
 - Initial testing suggests networking in Docker 1.10 is a lot more reliable (likely because of the new internal DNS)
 - [More info](http://blog.docker.com/2016/02/docker-1-10/)

Drawbacks:
 - docker-osx-dev doesn't handle the new top level `volumes` directive. For example, it warns of No such file or directory as it tries to sync grid/configs-vol grid/ivy-vol grid/sbt-vol.


theefer:
:+1: I guess

PR #1731, opened 2016-02-09
update fonts and remove unused fonts

akash1810:
applies sans serif to full character set

before
screen shot 2016-02-09 at 01 31 03

after
screen shot 2016-02-09 at 01 17 24

cc @paperboyo


paperboyo:
timid :+1: (these are the correct font files and we do not use Egyptians anywhere AFAIK). Thanks!

theefer:
Files are twice as big though? :sweat:

paperboyo:
Yep. Not much we can do about it. I have one idea but it may not work/we may not wanna do it and it will take time. Will let you know when /if that happens. They should go to the cache/local storage/service worker/brain implant, though.

theefer:
We should just cache them with a long expiry probably.

PR #1732, opened 2016-02-09
Use auth service instead of kahuna Panda

theefer:
Switch to use the new `auth` service instead of `kahuna`'s Panda endpoints to (re)auth in the kahuna web app.

Release order:
1. auth
2. media-api
3. kahuna

### DIFF
--- a/kahuna/conf/routes
+++ b/kahuna/conf/routes
@@ -17,6 +17,8 @@ GET     /login                      controllers.Panda.doLogin(redirectUri: Optio
 GET     /oauthCallback              controllers.Panda.oauthCallback
 GET     /logout                     controllers.Panda.logout
 GET     /session                    controllers.Panda.session
+# Empty page to return to the same domain as the app
+GET     /ok                         controllers.Application.ok
 
 # Management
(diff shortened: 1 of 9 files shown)


theefer:
We'll want to do some further testing in TEST but this is ready to go as CNAMES are ready.

kahuna/conf/routes
```diff
@@ -17,6 +17,8 @@ GET     /login                      controllers.Panda.doLogin(redirectUri: Optio
 GET     /oauthCallback              controllers.Panda.oauthCallback
 GET     /logout                     controllers.Panda.logout
 GET     /session                    controllers.Panda.session
+# Empty page to return to the same domain as the app
+GET     /ok                         controllers.Application.ok
```
kenoir: Can we use the `/management/healthcheck` uri for this purpose?

theefer: I would rather not.

Healthcheck is for healthcheck purposes (ELB, Pingdom, other monitoring tools). It's a system-facing endpoint.

The "OK" page is a user-facing landing page.

kenoir:
:+1: Other than ^

PR #1734, opened 2016-02-09
Displays collections button and collections on image metadata panel

tsop14:
For when collections go live ⚡️

Displays the collections button to everyone.

Removes if statements which hid collection information on image view. 
Removes keyboard shortcut to hide/show collections button. 
Removes services which checks localStorage.


akash1810:
Would/should training be organised before/close after this goes out?

akash1810:
:+1:

PR #1735, opened 2016-02-09
Improved performance (avoid big serialisation, more one-time bindings and ng:if)

theefer:
Did some performance analysis again and identified that we were serialising massing blobs of JSON representing the image and the crops for each search result, which is slow and expensive. This was done in case it is needed to put in the DataTransfer object if the image is dragged. Instead I've made it lazily evaluated so we only incur the cost on the one image the user starts dragging.

Also changed more bindings to be one-time bindings if the data is never expected to change. Changed some ng:show/hide to ng:if.

The bottom line is that we have in the order of 2000-5000 watch expressions (you can get that number using ngStats or the Batarang extension), based on what the user is doing. The recommended upper limit to keep an app smooth is 2000.

It's not trivial to remove watches, so more work is needed. The collections panel alone seems to setup quite a lot of watch expressions, even when only partially expanded. Not sure if it's another cost incurred from the recursive directive, or if it's expected.

I might do more performance analysis based on the Timeline. I suspect the fade-on-load directive on images isn't helping.

I don't have reliable numbers on the improvements this PR brings, only transient numbers I witnessed while testing with tools open. Also seems to feel smoother.

We'll want to do an extensive regression test before shipping this PR.

### DIFF
--- a/kahuna/public/js/main.js
+++ b/kahuna/public/js/main.js
@@ -51,6 +51,7 @@ var config = {
     vndMimeTypes: new Map([
         ['gridImageData',  'application/vnd.mediaservice.image+json'],
         ['gridImagesData', 'application/vnd.mediaservice.images+json'],
+        ['gridCropsData',  'application/vnd.mediaservice.crops+json'],
         ['kahunaUri',      'application/vnd.mediaservice.kahuna.uri'],
         // These two are internal hacks to help us identify when we're dragging internal assets
         // They should definitely not be relied on externally.
@@ -306,7 +307,9 @@ kahuna.filter('getExtremeAssets', function() {
     };
 });
 
-// Take an image and return a drag data map of mime-type -> value
+// Take an image and return a drag data map of mime-type -> value.
+// Note: the serialisation is expensive so make sure you only evaluate
+// this filter when necessary.
 kahuna.filter('asImageDragData', ['vndMimeTypes', function(vndMimeTypes) {
     // Annoyingly cannot use Resource#getLink because it returns a
     // Promise and Angular filters are synchronous :-(
@@ -332,16 +335,20 @@ kahuna.filter('asImageDragData', ['vndMimeTypes', function(vndMimeTypes) {
     };
 }]);
 
-// Take an image and return a drag data map of mime-type -> value
-kahuna.filter('asCropsDragData', function() {
+// Take an image and return a drag data map of mime-type -> value.
+// Note: the serialisation is expensive so make sure you only evaluate
+// this filter when necessary.
+kahuna.filter('asCropsDragData', ['vndMimeTypes', function(vndMimeTypes) {
     return function(crops) {
         return {
-            'application/vnd.mediaservice.crops+json': JSON.stringify(crops)
+            [vndMimeTypes.get('gridCropsData')]: JSON.stringify(crops)
         };
     };
-});
+}]);
 
-// Take an image and return a drag data map of mime-type -> value
+// Take an image and return a drag data map of mime-type -> value.
+// Note: the serialisation is expensive so make sure you only evaluate
+// this filter when necessary.
 kahuna.filter('asImageAndCropsDragData', ['$filter',
                                           function($filter) {
     var extend = angular.extend;
@@ -412,10 +419,9 @@ kahuna.directive('uiDragData', function() {
                 // Unwrap jQuery event wrapper to access dataTransfer
                 var e = jqueryEvent.originalEvent;
 
-                // No obvious way to receive an object through an
-                // attribute without making this directive an
-                // isolate scope...
-                var dataMap = JSON.parse(attrs.uiDragData);
+                // Evaluate the attribute value to retrieve a JS object
+                // (done lazily to avoid unnecessary serialisation work)
+                var dataMap = scope.$eval(attrs.uiDragData);
                 Object.keys(dataMap).forEach(function(mimeType) {
                     e.dataTransfer.setData(mimeType, dataMap[mimeType]);
                 });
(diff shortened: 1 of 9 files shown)


kenoir:
Is there a CSS only way to do fade on load?
On Tue, 9 Feb 2016 at 17:53, Sébastien Cevey <notifications@github.com>
wrote:

> Did some performance analysis again and identified that we were
> serialising massing blobs of JSON representing the image and the crops for
> each search result, which is slow and expensive. This was done in case it
> is needed to put in the DataTransfer object if the image is dragged.
> Instead I've made it lazily evaluated so we only incur the cost on the one
> image the user starts dragging.
>
> Also changed more bindings to be one-time bindings if the data is never
> expected to change. Changed some ng:show/hide to ng:if.
>
> The bottom line is that we have in the order of 2000-5000 watch
> expressions (you can get that number using ngStats or the Batarang
> extension), based on what the user is doing. The recommended upper limit to
> keep an app smooth is 2000.
>
> It's not trivial to remove watches, so more work is needed. The
> collections panel alone seems to setup quite a lot of watch expressions,
> even when only partially expanded. Not sure if it's another cost incurred
> from the recursive directive, or if it's expected.
>
> I might do more performance analysis based on the Timeline. I suspect the
> fade-on-load directive on images isn't helping.
>
> We'll want to do an extensive regression test before shipping this PR.
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/guardian/grid/pull/1735
> Commit Summary
>
>    - Lazily evaluate drag data and drag image to avoid big serialisation
>    cost
>    - Lots more one-time binding for immutable expressions
>    - Replace more ng:if with ng:show
>
> File Changes
>
>    - *M*
>    kahuna/public/js/components/gr-collections-panel/gr-collections-panel-node.html
>    <https://github.com/guardian/grid/pull/1735/files#diff-0> (6)
>    - *M*
>    kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
>    <https://github.com/guardian/grid/pull/1735/files#diff-1> (2)
>    - *M*
>    kahuna/public/js/components/gr-panel-button/gr-panel-button-small.html
>    <https://github.com/guardian/grid/pull/1735/files#diff-2> (16)
>    - *M* kahuna/public/js/components/gu-date-range/gu-date-range.html
>    <https://github.com/guardian/grid/pull/1735/files#diff-3> (4)
>    - *M* kahuna/public/js/edits/image-editor.html
>    <https://github.com/guardian/grid/pull/1735/files#diff-4> (12)
>    - *M* kahuna/public/js/image/view.html
>    <https://github.com/guardian/grid/pull/1735/files#diff-5> (44)
>    - *M* kahuna/public/js/main.js
>    <https://github.com/guardian/grid/pull/1735/files#diff-6> (32)
>    - *M* kahuna/public/js/preview/image.html
>    <https://github.com/guardian/grid/pull/1735/files#diff-7> (10)
>    - *M* kahuna/public/js/search/results.html
>    <https://github.com/guardian/grid/pull/1735/files#diff-8> (2)
>
> Patch Links:
>
>    - https://github.com/guardian/grid/pull/1735.patch
>    - https://github.com/guardian/grid/pull/1735.diff
>
> —
> <https://github.com/guardian/grid/pull/1735>.
>

-- 

in England Number 908396

theefer:
Possibly, though in theory the directive is trying to only fade it if it's not immediately available. Not sure it's working or worth it though.

akash1810:
Feels like its best to use one time binding where ever possible? And `ng:if` over `ng:show` though don't `ng:if`s cause layout changes?

theefer:
Yes to one-time binding.

`ng:if` causes modification to the DOM tree (incl node and directive instantiation), but I don't think anything would change in terms of layout as AFAIK `display:none` elements (via `ng:show`) would not be laid out either? 

cc @OliverJAsh or am I wrong? 😕

kahuna/public/js/main.js
```diff
@@ -412,10 +419,9 @@ kahuna.directive('uiDragData', function() {
                 // Unwrap jQuery event wrapper to access dataTransfer
                 var e = jqueryEvent.originalEvent;
 
-                // No obvious way to receive an object through an
-                // attribute without making this directive an
-                // isolate scope...
-                var dataMap = JSON.parse(attrs.uiDragData);
+                // Evaluate the attribute value to retrieve a JS object
+                // (done lazily to avoid unnecessary serialisation work)
+                var dataMap = scope.$eval(attrs.uiDragData);
```
jamesgorrie: Nice.

kahuna/public/js/preview/image.html
```diff
@@ -48,13 +48,13 @@
         <div class="flex-container">
             <ul class="bottom-bar__meta-item preview__collections">
                 <li class="preview__collections__collection"
-                    ng:repeat="collection in ctrl.image.data.collections"
-                    gr-tooltip="Click to open collection: {{collection.data.path.join(' ▸ ')}}"
+                    ng:repeat="collection in ctrl.image.data.collections track by collection.data.pathId"
```
jamesgorrie: Why the `track by`?

jamesgorrie:
re: `ng:if` vs `ng:show`
Wth the `ng:if` - you loose the event handlers, which, I think, is good for performance?

About layout - it definitely repaints layout, unless browsers have got clever enough to cache layouts? That would be good.

kahuna/public/js/preview/image.html
theefer: It's recommended to always use `track by` in ngRepeat so that Angular can recycle nodes for items uniquely identified by their `collection.data.pathId`. In this instance I think it would avoid re-rendering all the collections if the list was modified (since without this, nodes would be identified by the object identity of each collection, which is entirely replaced when collections are added/removed).

I don't expect a big impact of this but I think it's good practice to use with all ngRepeat.

theefer:
> About layout - it definitely repaints layout, unless browsers have got clever enough to cache layouts? That would be good.

The question is more whether using `ng:show` (i.e. `display: none`) is more efficient in terms of layout. I don't see how it could, seeing as it's presumably omitted from the rendered tree, but I might be wrong.

There is more DOM and directive initialisation work happening with ng:if, but on the other hand, it's a lot of work you don't have to pay for *unless* the `ng:if` condition turns `true`. With `ng:show`, you pay that cost regardless of whether the element is shown or not. My unfounded suspicion was that it may be better to use `ng:show` to show/hide things quickly (e.g. on hover) and if the cost of that content is low enough. Otherwise, I suspect it's cheaper not to do the work in the first place (e.g. we shouldn't load and render the whole collection tree if it's hidden, or the complex date filter picker if it's not expanded -- we do both at the moment).

theefer:
Looks all fine (and feels a little less janky) in TEST. Will ship after lunch.

I noticed the DND between Chrome and FF is broken (Grid<->Composer), both ways, but it seems to already be the case in current `master` so not a blocker...

akash1810:
:+1:

PR #1738, opened 2016-02-10
sets nonFree value as string to match type elsewhere

tsop14:
FIXES unnecessary extra state change. 

The 'nonFree' value from the URL was a string, the value set by the radio button was a boolean; '"true"' would not match 'true'. This was causing the state to change to update to the 'new' value.


theefer:
:+1:

PR #1739, opened 2016-02-10
Re-introduce "Use auth service instead of kahuna Panda"

theefer:
The grid-infra fix should have fixed the issue, so going to try again (from the branch this time) in TEST.


theefer:
Tested on TEST including session re-auth, redirections, etc. Looks all fine.

Will ship to PROD after lunch.

PR #1744, opened 2016-02-10
Catchall mappings

NickPapacostas:
Mapping changes for adding a standard analyzed catch all field

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
@@ -61,22 +61,23 @@ object Mappings {
   val metadataMapping = nonDynamicObj(
     "dateTaken" -> dateFormat,
     "description" -> sStemmerAnalysedString,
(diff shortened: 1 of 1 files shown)


NickPapacostas:
related to: https://github.com/guardian/grid/pull/1745

kenoir:
:+1:

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -120,7 +121,7 @@ object Mappings {
   val userMetadataMapping =
     nonDynamicObj(
       "archived"    -> boolean,
-      "labels"      -> nonAnalysedList("label"),
+      "labels"      -> (nonAnalysedList("label") ++ copyTo("metadata.englishAnalysedCatchAll")),
```
theefer: Looks like I commented on another similar PR, but are we sure want to add labels to this?

theefer: See other comments on https://github.com/guardian/grid/pull/1568

PR #1746, opened 2016-02-11
editorconfig for .yml files

akash1810:
~~intellij's default is 4, grr!~~ our `*` rule makes it 4 otherwise.


NickPapacostas:
:+1:

PR #1748, opened 2016-02-11
fixes highlight open collection

tsop14:
Fixes broken open collection(s) highlight. 

Selected collections is an array of lower case strings created by parsing the query. 
Need to split the pathId string ('bear/brown') to be able to match the two. 

highlight collection

### DIFF
--- a/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
+++ b/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
@@ -142,7 +142,7 @@ grCollectionsPanel.controller('GrNodeCtrl',
     }));
 
     ctrl.isSelected = ctrl.selectedCollections.some(col => {
-        return angular.equals(col, ctrl.node.data.fullPath);
+        return angular.equals(col, ctrl.node.data.data.pathId.split('/'));
     });
 
 }]);


akash1810:
:+1:

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -142,7 +142,7 @@ grCollectionsPanel.controller('GrNodeCtrl',
     }));
 
     ctrl.isSelected = ctrl.selectedCollections.some(col => {
-        return angular.equals(col, ctrl.node.data.fullPath);
+        return angular.equals(col, ctrl.node.data.data.pathId.split('/'));
```
theefer: Out of curiosity, why is `selectedCollections` split by `/` and not just the `pathId` string? Having to split the `pathId` here adds cost to the watch expression for every single node.

tsop14:
This is because `selectedCollections` is an array of arrays of lower case strings, e.g. `["observer", "food"]`. The other ctrl.node values have maintained casing and so cannot be used. 

> Having to split the pathId here adds cost to the watch expression for every single node.

In which case calling  `.join('/')` on col should be better?

PR #1749, opened 2016-02-11
remembers scroll position in collections panel

tsop14:
Fixes: collections panel scrolling back to the top when opening a new collection / returning from an image page.

Adds directive remember-scroll-top to the collections-panel. This records the position scrolled to (using the element's scrollTop value) in localStorage, and assigns that value to the collection-panel element's scrollTop once the collections panel tree has been rendered. 

Because Zebras are important too.  

remember scroll

### DIFF
--- a/kahuna/public/js/directives/gr-remember-scroll-top.js
+++ b/kahuna/public/js/directives/gr-remember-scroll-top.js
@@ -1,0 +1,34 @@
+import angular from 'angular';
+import Rx from 'rx';
+import 'rx-dom';
+
+import {rxUtil} from '../util/rx';
+
+export const rememberScrollTop = angular.module('gr.rememberScrollTop', [
+    rxUtil.name
+]);
+
+rememberScrollTop.directive('grRememberScrollTop', ['subscribe$', function (subscribe$) {
+    const storageKey = 'gr:rememberScrollTop';
+
+    return {
+        restrict: 'A',
+        link: function(scope, element, attrs) {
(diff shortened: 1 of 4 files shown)


kahuna/public/js/directives/gr-remember-scroll-top.js
```diff
@@ -0,0 +1,34 @@
...
+
+rememberScrollTop.directive('grRememberScrollTop', ['subscribe$', function (subscribe$) {
+    const storageKey = 'gr:rememberScrollTop';
+
+    return {
+        restrict: 'A',
+        link: function(scope, element, attrs) {
+            // Only remember if attribute if truthy. This is an annoying workaround to the
+            // inability to add directives conditionally
+            if (scope.$eval(attrs.grRememberScrollTop)) {
```
akash1810: An [HTML5-esque](https://github.com/guardian/grid/blob/master/kahuna/public/js/components/gr-icon/gr-icon.js#L20) way of doing this by testing the presence of the attribute.

akash1810:
:+1:

kahuna/public/js/directives/gr-remember-scroll-top.js
tsop14: 🙈

theefer: That won't work because the attribute has to always be there, but its value evaluated, due to how it's used in the panels, and the fact AngularJS doesn't support conditionally adding directives.

PR #1750, opened 2016-02-12
clearer Collections edit button states

akash1810:
Making it more obvious what state you're in.

# Before
edit-button-before

# After
edit-button


theefer:
Cool. Is the Edit button an existing button style we use elsewhere? (The Finished style seems to be?)

akash1810:
Edit button is new. Or rather its a duplicate of `.button-cancel` with a more relevant name for the context its used in. The finished style is `.button-save` which has sensible nomenclature in this context.

tsop14:
I like it - much more obvious. Re. styling, should it be full heigh like the 'clear selection', 'add to library' et al. that appear in the top bar?  I suppose it is signalling a mode that you are in, so maybe different from them.

akash1810:
> Re. styling, should it be full heigh like the 'clear selection', 'add to library' et al. that appear in the top bar?

Tried that but a full height blue button looked a bit odd. We're also following the styling of the "new images" button here, which is blue and not full height.

tsop14:
:+1:

PR #1751, opened 2016-02-12
Preconnect to media-api and auth ahead of JS making requests to them

theefer:
Leverage the [link rel="preconnect"](https://www.w3.org/TR/resource-hints/#dfn-preconnect) draft spec implemented in Chrome and FF to "pre-warm" the connection to APIs (DNS lookup, TCP connection, TLS handshake) ahead of the JS app being downloaded, evaluated and executed to make requests against those APIs.

I'm only preconnecting to the media-api and auth API for now as these are the main two on the initial critical path for loading the search results.

I ran some WebPageTests to review the impact.

Note that both of these are on first render prior to any pre-existing DNS caching etc. They're run from NYC on a "cable connection" (the servers being in Ireland).

## Before (no preconnect)

screenshot-2016-02-12t11 53 56

## After (with preconnect)

screenshot-2016-02-12t11 54 42

Notice the red highlights in the first example. I'm not focusing on the Document Complete latency, even though it's 1.5s faster with preconnect, because there seems to be some variation in the latency of the download of the `build.js` file (1 extra second in the first example) which should not be related to this.

Interestingly, even after running several tests, we can't seem to be getting much performance improvement on the `api.media.__` requests. With preconnect, according to these results, we don't get any DNS lookup/initial connection/TLS handshake penalty, but for reason the Content Download ends up being longer. I don't know why that is, whether it's a TCP artifact, or whether the results aren't quite correct.

We do shave off about 200ms on the `media-auth.__` request though.

I'm sure there is more to it, and given the unexpectedly similar latency on `api.media.__`, it's not a massive performance win, but still saves some time on the other preconnected request, so feels worth adding anyway.

PS: there is lots and lots more we could do to optimise the waterfall and load times (better caching, bundle more files together, better caching, tree shake the build.js to try and reduce its size, did I mention better caching?), but this looks like a small easy win (4 LOCs!) and generally good practice.

---

Kudos to @igrigorik who talked about this at LDNWebPerf yesterday evening.

cc @piuccio @OliverJAsh @jamesgorrie @phamann

### DIFF
--- a/kahuna/app/controllers/Application.scala
+++ b/kahuna/app/controllers/Application.scala
@@ -12,6 +12,7 @@ object Application extends Controller {
     val returnUri = Config.rootUri + okPath
     Ok(views.html.main(
       Config.mediaApiUri,
+      Config.authUri,
       s"${Config.authUri}/login?redirectUri=$returnUri",
       Config.watUri,
       Config.sentryDsn))
--- a/kahuna/app/views/main.scala.html
+++ b/kahuna/app/views/main.scala.html
@@ -1,4 +1,5 @@
 @(mediaApiUri: String,
+  authApiUri: String,
   reauthUri: String,
   watUri: Option[String],
   sentryDsn: Option[String])
@@ -9,6 +10,10 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
     <title ui:title ui:title-suffix="the grid">the grid</title>
+
+    <!-- preconnect to core APIs ahead of JS making the requests -->
+    <link rel="preconnect" href="@mediaApiUri"/>
+    <link rel="preconnect" href="@authApiUri"/>
 
     <link rel="shortcut icon" type="image/png" href="@routes.Assets.at("images/grid-favicon-32.png")"/>
     <link rel="assets" href="@routes.Assets.at("")"/>


piuccio:
I love the randomness of these results :grin:  :+1:

theefer:
We'd probably get more consistent (and better) results if we used a CDN...

akash1810:
interesting!

:+1:

akash1810:
presumably the spec has been implemented in the browser versions on the managed machines?

theefer:
It's [not that new](http://caniuse.com/#feat=link-rel-preconnect) (Chrome 46, FF 43), and of course in the worst case it just doesn't do anything.

theefer:
If we had RUM to track time-to-usable-UI it'd be great to see the impact of shipping this, but sadly we don't...

igrigorik:
:+1:

OliverJAsh:
Nice experiment :+1:

phamann:
:+1:

andydavies:
There's a couple of things I think could be going on here…

WPT uses a Chrome extension that attaches to the page being loaded via the debug API, and then listens to the events - network, page level etc.

The waterfalls for FF and IE are generated by sniffing the traffic over the wire, so it's worth running a comparable test in FF (as it supports preconnect) to see what it's water fall looks like. (Believe the Chrome agents will eventually go down the same route)

You could capture a TCP dump, with TLS keys and decypher it in Wireshark to see what's actually happening on the wire in the Chrome case.

How many tests did you run - for something like preconnect that's going to make a difference in 10's of ms, you want to run more tests (9 on the public instance, many more if you've got access to a private instance), just the normal variation in network performance can produce oddities like content download appearing to be longer in the preconnect case?

Sometime when features like preconnect come along the Chrome debug API's don't quite reflect what's happening on the network stack or WPT doesn't interpret the output correctly and this might be the case here - you'll notice in the WPT waterfall there's no DNS lookup or TCP connection associated with the requests that preconnected and there should be.

If you're willing to share the URL (DM) I'll take a deeper look.

theefer:
@andydavies: thanks a lot for the detailed answer!

I was indeed running with Chrome. I'd ticked 1 test run, though I manually ran the test 3 times for each case (with/without preconnect). I did notice that there was no DNS lookup, TCP connection or SSL negotiation in the preconnect, which seemed to indicate something was missing.

I re-ran the test (9 runs) with FF and this time the preconnect is clearly visible much ahead of the actual request:

screenshot-2016-02-13t18 11 23

Sounds like I was seeing an artifact of how WPT instruments Chrome. Not sure what it means in terms of whether we're getting all the benefits of preconnect, but at least FF users do get it :-)

Thanks again!

andydavies:
I haven't dug through the Chrome / WPT interaction yet but…

Looking at the packet capture Chrome is resolving the hostname, creating a connection and negotiating TLS in advance - it completes this 2 seconds before it actually makes the first HTTPS request to the api.media... host.

PR #1753, opened 2016-02-12
Allow picking the date field to filter by (uploaded, taken, modified)

theefer:
image

This is done by adding a new `dateField` and `{taken,uploaded}{Since,Until}` parameters to the state (kahuna URI), and sending them along to the media-api using the already supported `{taken,uploaded}{Since,Until}` parameters.

cc @blishen

### DIFF
--- a/kahuna/public/js/search/index.js
+++ b/kahuna/public/js/search/index.js
@@ -129,7 +129,8 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
     });
 
     $stateProvider.state('search.results', {
-        url: 'search?{query:Query}&ids&since&nonFree&uploadedBy&until&orderBy&dateField&takenSince&takenUntil&modifiedSince&modifiedUntil',
+        url: 'search?{query:Query}&ids&since&nonFree&uploadedBy&until&orderBy' +
+             '&dateField&takenSince&takenUntil&modifiedSince&modifiedUntil',
         // Non-URL parameters
         params: {
             // Routing-level property indicating whether the state has
--- a/kahuna/public/js/search/index.js
+++ b/kahuna/public/js/search/index.js
@@ -129,7 +129,7 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
     });
 
     $stateProvider.state('search.results', {
-        url: 'search?{query:Query}&ids&since&nonFree&uploadedBy&until&orderBy',
+        url: 'search?{query:Query}&ids&since&nonFree&uploadedBy&until&orderBy&dateField&takenSince&takenUntil&modifiedSince&modifiedUntil',
         // Non-URL parameters
         params: {
             // Routing-level property indicating whether the state has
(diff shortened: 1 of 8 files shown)


paperboyo:
Cool! Do we know what's the percentage of images with 'dateTaken'? If you sort by 'dateTaken' what happens to the images that do not have it (does it fallback to 'modified', and vice versa)? Does any user activity in Grid change 'modified'?

theefer:
About 2M images have dateTaken  out of the 3.5M images in the db. Images without one are simply not returned when filtering by dateTaken, which kind of makes sense, but is worth noting. Have to get numbers on the dateTaken coverage for recent images... 

All user edits on metadata, usage crops etc bump last modified.

kahuna/public/js/search/index.js
```diff
@@ -129,7 +129,8 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
     });
 
     $stateProvider.state('search.results', {
-        url: 'search?{query:Query}&ids&since&nonFree&uploadedBy&until&orderBy',
+        url: 'search?{query:Query}&ids&since&nonFree&uploadedBy&until&orderBy' +
+             '&dateField&takenSince&takenUntil&modifiedSince&modifiedUntil',
```
jamesgorrie: I know you can use the date type, although I haven't done it before. i.e. `{modifiedUntil:date}`.
That applies to `boolean`s and `int`s too.

paperboyo:
> About 2M images have dateTaken out of the 3.5M images in the db

That's a much better ratio than I thought. If indeed most of new images have this field, then



> Images without one are simply not returned when filtering by dateTaken

wouldn't present a problem. :-)

> All user edits on metadata, usage crops etc bump last modified.

Super-cool! Thanks for explanation.

kahuna/public/js/search/index.js
theefer: Good to know thanks. I think I'm going to leave as a String for now because that's how everything else works and I don't really want to refactor so much more unrelated stuff as part of this change..

akash1810:
:+1:

kahuna/public/js/search/index.js
jamesgorrie: Yea, I only found it out the other day and thought it was quite nice.

akash1810: Is that what's going on with `{query:Query}` too? If so, its not restricted to native types!

jamesgorrie: Nope, you can write your own types, but some are native.
I can't find an exhaustive list though.

jamesgorrie: [More on custom types though](http://angular-ui.github.io/ui-router/site/#/api/ui.router.util.$urlMatcherFactory#methods_type).

jamesgorrie: [And here we are](https://github.com/angular-ui/ui-router/blob/master/src/params/paramTypes.ts#L21)

PR #1754, opened 2016-02-12
fixes unnecessary state change

tsop14:
by using angular equals to compare objects


theefer:
:+1:
The only risk is if we ever use it on large objects where we actually want to rely on object identity, but I did a quick scan and it doesn't look like we do today.

PR #1755, opened 2016-02-12
save defaultRestriction when present

akash1810:
add the defaultRestriction to the model to save it and show the message in the banner

before :cry: 
screen shot 2016-02-12 at 15 40 39

after :tada: 
screen shot 2016-02-12 at 15 39 28


tsop14:
🎉👍

PR #1757, opened 2016-02-12
uncheck restriction deletes the property from the model

akash1810:
Allows restriction removal by unchecking the checkbox and saving. Previously you had to delete the value in the text area then save.

Weekend do this a lot. It looks really painful! Especially if you haven't yet learnt how to get the combination just right (as shown in the before gif).

Before
restrictions-before

After
restrictions-after

### DIFF
--- a/kahuna/public/js/usage-rights/usage-rights-editor.js
+++ b/kahuna/public/js/usage-rights/usage-rights-editor.js
@@ -135,6 +135,11 @@ usageRightsEditor.controller(
         const data = ctrl.category.value ?
             angular.extend({}, ctrl.model, { category: ctrl.category.value }) : {};
 
+        // unchecking restrictions will remove restriction on save
+        if (! ctrl.showRestrictions && data.restrictions) {
+            delete data.restrictions;
+        }
+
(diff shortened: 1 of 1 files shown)


kahuna/public/js/usage-rights/usage-rights-editor.js
```diff
@@ -147,6 +147,12 @@ usageRightsEditor.controller(
 
     ctrl.cancel = () => ctrl.onCancel();
 
+    $scope.$watch(() => ctrl.showRestrictions, onValChange(restrictions => {
```
akash1810: this feels pretty hacky - happy to get an alternative!

theefer: Could we do this on Save rather than here? That way you can toggle the restriction and not loose the text; you'd only lose it after you save?

akash1810: Ah yes, much better!

Done.

theefer:
If it still works for categories that require restrictions :+1:

PR #1758, opened 2016-02-15
adds newest first sorting on collections

tsop14:
Allows users to sort by 'newest first' on collections query. 

sorting on collections

### DIFF
--- a/kahuna/public/js/search/query.js
+++ b/kahuna/public/js/search/query.js
@@ -68,6 +68,8 @@ query.controller('SearchQueryCtrl',
             ctrl.filter[key] = valOrUndefined($stateParams[key]);
         }
 
+        ctrl.collectionSearch = ctrl.filter.query ?  ctrl.filter.query.indexOf('~') === 0 : false;
+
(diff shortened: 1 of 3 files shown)


kenoir:
Does this clash with https://github.com/guardian/grid/pull/1753?

tsop14:
It might need a local merge but it doesn't step on #1753's toes.

kahuna/public/js/search/query.js
```diff
@@ -101,7 +104,7 @@ query.controller('SearchQueryCtrl',
     });
 
     function resetQueryAndFocus() {
-        ctrl.filter.query = '';
+        ctrl.filter.query = undefined;
```
akash1810: Would `delete ctrl.filter.query` be better?

kenoir:
:+1:

kahuna/public/js/search/query.js
tsop14: I don't think so as `''` matches the value set when there is an empty query param. I think deleting it would stop it matching and trigger another state.go.

theefer: I think either way works here

PR #1759, opened 2016-02-15
no longer splits collection query string

tsop14:
`getCollectionsFromQuery(q)` only used in gr-collections-panel.js - do not need to split the value in both places for comparison.


theefer:
\o/ :+1:

PR #1762, opened 2016-02-15
refactoring gr-panel -> gr-info-panel

akash1810:
With the aim of fixing the issue of the disappearing info panel, refactor to a less confusing name.

Similar refactoring for the CSS on the way.


theefer:
:+1:

paperboyo:
Is there a reason why the LHS panel rememebers state via localstorage, but not the RHS one?

akash1810:
@paperboyo no ones asked for it to ;) What state would it remember?

paperboyo:
I haven't asked for something?! I can't believe it ;-)

It should remember exactly the same states that Collection panel does, I think. No? Not this PR, obviously, just piggybacking, 'cause slightly related...

akash1810:
Ah, so locked/unlocked? And not scroll position (this won't be useful in this context).

We haven't remembered locked/unlocked thus far to give as much space to the thumbnails as possible. We could revisit if there's a demand - I'll take a look at the metrics for panel toggling.

paperboyo:
Thanks. I think consistency would be nice to have in these devices (although, I agree about the scroll position!).

akash1810:
> I'll take a look at the metrics for panel toggling.

We're not recording this :cry:.

PR #1763, opened 2016-02-16
Fix labeller to update when labels are added

theefer:
Was subtly broken in the `onValChange` due to Resource properties not being enumerable.


kenoir:
:+1:

PR #1764, opened 2016-02-16
Fix erroneous state.go which broke search vertical offset

theefer:
This was always creating an extra `state.go` when the search page is opened.


kenoir:
:+1:

PR #1765, opened 2016-02-16
removes unused css rules

akash1810:
Following on from https://github.com/guardian/grid/pull/1762, these rules are not used so lets remove them for simplicity.


theefer:
If you've checked they're not used (why did renaming some files make them no longer used?) :+1:

akash1810:
> why did renaming some files make them no longer used?

It didn't. I did a global search on the kahuna directory for these classes and the only match was in the css files. I guess they've been hanging around for a while as removing css is a little scary...

akash1810:
Double checked on TEST. Merging.

PR #1766, opened 2016-02-17
Reduce alpha of unlocked panels

kenoir:
In order that we can see what's underneath so as not to hide images.

screenshot from 2016-02-17 11 22 18

### DIFF
--- a/kahuna/public/js/components/gr-panels/gr-panels.css
+++ b/kahuna/public/js/components/gr-panels/gr-panels.css
@@ -20,22 +20,25 @@
 .gr-panel__content {
     width: 290px;
     position: fixed;
-    background: #444;
+    background: rgba(68, 68, 68, 0.9);
     overflow: auto;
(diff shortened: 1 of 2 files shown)


tsop14:
Nice! What does it look like when the panel is locked?

jamesgorrie:
> What does it look like when the panel is locked?

It would look grey as it would pick up the background from behind.

You might want to set the background to `rgba(0,0,0,.9)` (but the right values) to not knock out the text too?

theefer:
What @jamesgorrie said. Either way, would be nice for the panel not to be transparent when locked so text is not faded out and the browser doesn't have to composite layers (performance).

kenoir:
@jamesgorrie @theefer yeah yeah. Very nice.

akash1810:
Shiny!

:+1:

paperboyo:
:gem:

PR #1767, opened 2016-02-17
simplify panel height logic, controlling height with css

akash1810:
Fixes issue of disappearing info panel too!

### DIFF
--- a/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.html
+++ b/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.html
@@ -10,17 +10,17 @@
     <div class="flex-container collections-panel-heading text-small" ng:if="! ctrl.error">
         <div class="flex-spacer">Collections</div>
 
-            <button ng:class="{'button-edit': ! $parent.editing, 'button-save': $parent.editing}"
-                    ng:click="$parent.editing = !$parent.editing">
-                <gr-icon-label ng:if="! $parent.editing"
-                               gr:icon="edit">Edit</gr-icon-label>
-                <gr-icon-label ng:if="$parent.editing"
-                               gr:icon="check">Finished</gr-icon-label>
-            </button>
-        </div>
-        <gr-collection-tree gr:nodes="ctrl.collections"
-                  gr:editing="editing"
-                  gr:selected-images="ctrl.selectedImages$"
-                  gr:selected-collections="ctrl.selectedCollections">
-        </gr-collection-tree>
+        <button ng:class="{'button-edit': ! $parent.editing, 'button-save': $parent.editing}"
+                ng:click="$parent.editing = !$parent.editing">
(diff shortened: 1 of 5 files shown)


akash1810:
tested on TEST too.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.html
```diff
@@ -1,6 +1,5 @@
-<div class="panel"
-     ng:init="editing = false">
-    <div class="panel__content">
+<div ng:init="editing = false">
+    <div>
```
theefer: Do we need this `div`?

theefer:
Not quite working as is (crops the bottom of the RHS panel), maybe `min-height`? In my experience `height: 100%` never quite worked for me. @jamesgorrie may have more ideas.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.html
akash1810: nope, removed.

akash1810:
Fixed now, using `calc`.

kahuna/public/js/components/gr-panels/gr-panels.css
```diff
@@ -51,3 +48,7 @@
     display: block;
     margin-top: 36px;
 }
+
+.gr-panel-height {
+    height: calc(100vh - 86px); /* 50px search bar, 36px toolbar */
```
theefer: @jamesgorrie didn't we use `calc` before? wondering why we went to JS if so.

jamesgorrie: I think I was overcomplicating  / abstracting things - so it was obviously broken.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.html
```diff
@@ -1,16 +1,14 @@
...
-        <!-- We use $parent.editing here as ng:if creates it's own child scope -->
-        <div class="flex-container collections-panel-heading text-small" ng:if="! ctrl.error">
-            <div class="flex-spacer">Collections</div>
+    <!-- TODO: we'll probably move this out to another template once we have something more
+    final -->
+    <!-- We use $parent.editing here as ng:if creates it's own child scope -->
+    <div class="flex-container collections-panel-heading text-small" ng:if="! ctrl.error">
+        <div class="flex-spacer">Collections</div>
 
             <button ng:class="{'button-edit': ! $parent.editing, 'button-save': $parent.editing}"
```
tsop14: These don't need to be indented.

tsop14:
:+1: only indent.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.html
akash1810: good spot!

PR #1769, opened 2016-02-17
Allow batch downloading lo-rez images

kenoir:
This seems to be aprt of multiple peoples daily workflow so we should probably cater for them.

screenshot from 2016-02-22 13 23 50

Moves creating download urls to a service, uses Rx. 

This PR also replaces the download control on the image page with the shared component.

### DIFF
--- a/kahuna/public/js/components/gr-downloader/gr-downloader.js
+++ b/kahuna/public/js/components/gr-downloader/gr-downloader.js
@@ -1,9 +1,14 @@
 import angular from 'angular';
 import JSZip from 'jszip';
 import './gr-downloader.css!';
+import Rx from 'rx';
 import template from './gr-downloader.html!text';
 
-export const downloader = angular.module('gr.downloader', []);
+import '../../services/image/downloads';
+
+export const downloader = angular.module('gr.downloader', [
+    'gr.image-downloads.service'
+]);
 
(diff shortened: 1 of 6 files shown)


NickPapacostas:
:+1: tested locally and it works great. maybe someone more familiar with our angular stuff wants to look that over though? i just don't think i know enough to comment

kahuna/public/js/components/gr-downloader/gr-downloader.js
```diff
@@ -21,31 +24,46 @@ const bytesToSize = (bytes) => {
...
+    ctrl.download    = (downloadKey) => {
+
         ctrl.downloading = true;
 
-        const zip = new JSZip();
-        const imageHttp = url => $http.get(url, { responseType:'arraybuffer' });
-        const imagesAddedToZip = Array.from(ctrl.images.values()).map(image =>
-            imageHttp(image.data.source.secureUrl)
-                .then(resp => zip.file(imageName(image), resp.data))
+        const downloads$ = imageDownloadsService.download(
```
jamesgorrie: Pedant alert: worth being `imageDownloadsService.downloads$()`

kenoir: https://github.com/guardian/grid/pull/1778

PR #1770, opened 2016-02-18
use updated amis

akash1810:
Triggered by CVE-2015-7547.

Also:
- Tags built AMIs in AWS
- Uses pretty printed timestamp rather than time in milliseconds
- Builds our own imagemagick instance (previously we were using an AMI from frontend)

Post release:
- [ ] Delete old AMI and snapshots


kenoir:
:+1:

PR #1774, opened 2016-02-19
collections panel refactor

tsop14:
Creates parent directive grCollectionsTree which takes the attributes needed to construct grNodes and grNode. 

grNodes and grNode reference this parent directive for unchanging attributes - rather than setting them again in scope object. 

Adds ability to add a custom select event - for future non-navigation uses of the collections tree.  🌲

### DIFF
--- a/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
+++ b/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
@@ -67,7 +67,6 @@ grCollectionsPanel.controller('GrCollectionsPanelCtrl', [
 
     const ctrl = this;
 
-    ctrl.isVisible = false;
     ctrl.error = false;
 
     collections.getCollections().then(collections => {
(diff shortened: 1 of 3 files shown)


kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -99,75 +98,82 @@ grCollectionsPanel.controller('GrNodeCtrl',
...
     return {
         restrict: 'E',
+        require: ['grNode', '^^grCollectionTree'],
         scope: {
             node: '=grNode',
-            nodeList: '=grNodeList',
-            editing: '=grEditing',
-            selectedImages$: '=grSelectedImages',
-            selectedCollections: '=grSelectedCollections'
+            nodeList: '=grNodeList'
```
theefer: Is this not just `grCollectionTreeCtrl.nodes`?

theefer: Ah maybe not, seems to be filtered down recursively? Not sure what it represents then :)

theefer:
Great! :+1:

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
tsop14: `nodeList` is a list of 'sibling' nodes, which is used when a collection is deleted. So, think it does need to be passed down recursively.

theefer: OK makes sense (might be nice to rename for clarity at some point, but not urgent)

PR #1776, opened 2016-02-19
Style text inputs using classes, not element selector

theefer:
Minor CSS refactor to avoid styling text `input` and `textarea` directly, since this makes it hard to do anything with those without inheriting the styling. Use a class that's explicitly applied instead.

This is to help the upcoming search chips feature implementation.


jamesgorrie:
:+1: Long time coming

PR #1780, opened 2016-02-22
Structured query search using UI chips

theefer:
Use visual "chips" as part of the search to help users rely on faceted searches.

So far, only the facet fields, labels and credits are auto-completed. We could easily add more ES suggesters to do the same on things like `by`, `uploader` (!), etc.

Sample use:

recorded
recorded

Sorry, the code dump is little epic. I blame Angular :cry: 

## Minor bugs left

- [ ] Datalist dropdown offset to the right when the query is wider than the input field (horizontal scroll). Due to some tricky positioning of the dropdown I haven't managed to solve yet.
- [ ] Collections are static filters (can't be edited), which is OK, but they also can't be focused, which breaks keyboard navigation (Left/Right) to jump over these chips.
- [ ] Scroll datalist vertical position when up/down moves the active item out of view (existing issue with datalist)

## Next

- [ ] Update Advanced panel to document the feature and guide the user through using it
- [ ] Announce the feature more widely via comms to all picture editors
- [ ] In-tool tooltip to advertise feature for those who don't read comms
- [ ] Add ES suggesters for all fields we care about (e.g. `by`, `uploader` (!!), `country`, etc) to add auto-suggestion
- [ ] Separate `by` from `photographer`/`illustrator` (currently cheating by using `photographer` list instead of existing `byline`s), talk to @paperboyo if needed

### DIFF
--- a/kahuna/public/js/components/gr-chips/gr-chips.css
+++ b/kahuna/public/js/components/gr-chips/gr-chips.css
@@ -1,0 +1,64 @@
+gr-chips {
+    white-space: nowrap;
+    overflow: hidden;
+}
+
+/* Make sure the last text token is wide enough */
+gr-chips .gr-chips__chip-wrapper:last-child gr-text-chip input {
+    min-width: 2em;
+}
+
+.gr-chips__placeholder {
+    position: absolute;
+    color: #aaa;
+    -webkit-user-select: none;
+       -moz-user-select: none;
+            user-select: none;
+}
+
+gr-text-chip,
+gr-filter-chip,
+gr-static-filter-chip,
+gr-filter-chooser-chip,
+gr-filter-chip,
+gr-datalist {
+    display: inline-block;
+}
+
+gr-filter-chip button,
+gr-static-filter-chip button,
+gr-filter-chooser-chip button {
+    border: 0;
+    background: transparent;
+    padding: 0;
+}
+
+.gr-filter-chip__value,
+.gr-static-filter-chip__value,
+.gr-filter-chooser-chip__value {
+    padding: 4px 2px;
+}
+
+.gr-filter-chip__key,
+.gr-static-filter-chip__key,
+.gr-static-filter-chip__value {
+    cursor: default;
+}
+
+gr-text-chip input,
+gr-filter-chip input,
+gr-filter-chooser-chip input {
+    border: 0;
+    background: transparent;
+}
+
+gr-text-chip button:focus,
+gr-filter-chip button:focus,
+gr-static-filter-chip button:focus,
+gr-filter-chooser-chip button:focus,
+gr-text-chip input:focus,
+gr-filter-chip input:focus,
+gr-static-filter-chip input:focus,
+gr-filter-chooser-chip input:focus {
+    outline: none;
+}
--- a/kahuna/public/js/components/gr-chips/gr-chips.css
+++ b/kahuna/public/js/components/gr-chips/gr-chips.css
@@ -1,6 +1,7 @@
 gr-chips {
     white-space: nowrap;
     overflow: hidden;
+    cursor: text; /* pretend the whole thing is a text input */
 }
 
 /* Make sure the last text token is wide enough */
--- a/kahuna/public/js/components/gr-chips/gr-chips.html
+++ b/kahuna/public/js/components/gr-chips/gr-chips.html
@@ -1,0 +1,18 @@
+<div class="gr-chips__wrapper"
+     ng:click="$grChipsCtrl.wrapperClicked($event)">
+    <span ng:if="$grChipsCtrl.placeholder && $grChipsCtrl.isEmpty()"
+          class="gr-chips__placeholder"
+    >{{$grChipsCtrl.placeholder}}</span>
+    <span ng:repeat="$chip in $grChipsCtrl.items"
+          ng:switch="$chip.type"
+          class="gr-chips__chip-wrapper">
+       <gr-text-chip ng:switch-when="text"
+                     gr-chip="$chip"></gr-text-chip>
+       <gr-filter-chip ng:switch-when="filter"
+                       gr-chip="$chip"></gr-filter-chip>
+       <gr-static-filter-chip ng:switch-when="static-filter"
+                              gr-chip="$chip"></gr-static-filter-chip>
+       <gr-filter-chooser-chip ng:switch-when="filter-chooser"
+                               gr-chip="$chip"></gr-filter-chooser-chip>
+    </span>
+</div>
--- a/kahuna/public/js/components/gr-chips/gr-chips.js
+++ b/kahuna/public/js/components/gr-chips/gr-chips.js
@@ -169,6 +169,7 @@ grChips.controller('grChipsCtrl', ['$scope', function($scope) {
         // If chip to remove is focused, move focus right before or after
         const removedItem = $grChipsCtrl.items[index];
         if (removedItem === $grChipsCtrl.focusedItem) {
+            /* jshint expr: true */
             $grChipsCtrl.focusEndOfChipBefore(removedItem) ||
                 $grChipsCtrl.focusStartOfChipAfter(removedItem);
         }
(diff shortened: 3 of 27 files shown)


theefer:
I think it'd be worth revamping the "Advanced search" tab into a simple doc of using this search syntax (basically type `+`, `-` or `#` and follow the suggestions).

theefer:
Also missing is integration with date filters (you can kind of do it already using the date range picker) and the nested filters (usage, etc). Also no auto completion for collections, but they do get turned into a (immutable) chip:

image

jamesgorrie:
> I think it'd be worth revamping the "Advanced search" 

Could you make it that the first time you click in the box it gives you something with a "dismiss" button to make it more obvious as this is a great feature.

kahuna/public/js/components/gr-chips/gr-chips.js
```diff
@@ -0,0 +1,241 @@
...
+
+export const grChips = angular.module('gr.chips', [
+    grTextChip.name,
+    grFilterChip.name,
+    grStaticFilterChip.name,
+    grFilterChooserChip.name
+]);
+
+grChips.controller('grChipsCtrl', ['$scope', function($scope) {
+    const $grChipsCtrl = this;
```
jamesgorrie: Why the new naming convention?

tsop14:
My, what B✨E✨A✨U✨T✨Y! 

Does the 'selected' collection still work? It was checking node names against the text in the search box.

kahuna/public/js/components/gr-chips/gr-chips.css
```diff
@@ -0,0 +1,64 @@
...
+
+/* Make sure the last text token is wide enough */
+gr-chips .gr-chips__chip-wrapper:last-child gr-text-chip input {
+    min-width: 2em;
+}
+
+.gr-chips__placeholder {
+    position: absolute;
+    color: #aaa;
+    -webkit-user-select: none;
```
akash1810: > [This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select).

Assuming editorial browsers are happy with this selector?

kahuna/public/js/components/gr-chips/gr-chips.html
```diff
@@ -0,0 +1,18 @@
+<div class="gr-chips__wrapper"
+     ng:click="$grChipsCtrl.wrapperClicked($event)">
```
akash1810: Why `$grChipsCtrl` rather than `ctrl` (which seems to be preferred in this project)?

kahuna/public/js/components/gr-chips/gr-chips.js
```diff
@@ -0,0 +1,241 @@
...
+            $grChipsCtrl.setFocusedChip($grChipsCtrl.items[index], 0, 0);
+            return true;
+        }
+    };
+
+    $grChipsCtrl.focusStartOfFirstChip = function() {
+        const firstItem = $grChipsCtrl.items[0];
+        if (firstItem) {
+            $grChipsCtrl.setFocusedChip(firstItem, 0, 0);
+            return true;
```
akash1810: `.setFocusedChip` returns `true` or undefined, however its return value doesn't seem to be used. Could this be `return $grChipsCtrl.setFocusedChip(firstItem, 0, 0);` instead? Or could the `return` be removed?

kahuna/public/js/components/gr-chips/gr-filter-chip.html
```diff
@@ -0,0 +1,21 @@
...
+    <button type="button"
+            class="gr-filter-chip__type"
+            ng:click="$grFilterChipCtrl.toggleFilterType()"
+    >{{$chip.filterType === 'inclusion' ? '+' : '−' }}</button
+    ><span class="gr-filter-chip__key">{{$chip.key}}:</span
+    ><input type="text"
+            class="gr-filter-chip__value"
+            gr-datalist-input
+            gr-auto-width
+            ng:trim="false"
```
akash1810: Why are we not trimming the value?

kahuna/public/js/components/gr-chips/gr-chips.js
theefer: If you mean the `$` prefix, I borrowed this from angular-material coding style as I thought their code looked [nicely organised](https://github.com/angular/material/blob/master/src/components/chips/js/chipsDirective.js) and that it made the reference a little clearer in the templates and elsewhere. I'm not obsessed about it particularly so could remove the `$` if people prefer.

If you mean naming each controller differently rather than `ctrl`, this is to make things more self-explanatory when directives manipulate 2-3 different other controllers than themselves. It's also used to access the reference for the "parent" directive (e.g. `grChipsCtrl` from within a `grTextChip`) within the template, through scope inheritance, with hierarchy enforced via `require`.

kahuna/public/js/components/gr-chips/gr-chips.html
theefer: See [previous response](https://github.com/guardian/grid/pull/1780#discussion_r53762472), you need different names if you want to be able to refer to different controllers.

kahuna/public/js/components/gr-chips/gr-chips.js
theefer: I've changed it to return the return value of `$grChipsCtrl.setFocusedChip(...)`, as it's more concise, thanks.

The return value is [sometimes used](https://github.com/guardian/grid/pull/1780/files#diff-af864ba9fa711b16b26b24afdfe8af6bR172) to determine whether the focus did something or not, in order to fallback to another behaviour.

kahuna/public/js/components/gr-chips/gr-filter-chip.html
theefer: If the ngModel value is trimmed, it's impossible for `grAutoWidth` to calculate the width required for the field when you have leading or trailing spaces. There is a [comment to that effect](https://github.com/guardian/grid/pull/1780/files#diff-e04ed75395af4a024b163a60173e4d07R5) by that directive though arguably it's a little far from the code here, but I was trying to avoid polluting templates with such comments.

I tried to automatically set ngTrim in the directive but it doesn't seem to be possible :-(

paperboyo:
> We could easily add more ES suggesters

Add them all (URs)! ;)

:gem:!

kahuna/public/js/components/gr-chips/gr-chips.css
theefer: I checked and there is a draft and [good browser support](http://caniuse.com/#search=user-select) so I think it should be OK.

theefer:
> Does the 'selected' collection still work? It was checking node names against the text in the search box.

In theory the query is exactly the same as before[*], it just gets parsed into a structured before it's manipulated by the chips component, and re-serialised as a string for the rest of the app.

In practice, there are a few minor discrepancies, particularly labels and collections use `label:` and `collection:` instead of `#` and `~`. This actually breaks the the 'selected' collection (thanks!) because the helper that extracts the collection from the query doesn't account for the `collection:` syntax, though arguably it should.

I'll have a think about the best way to render a query consistently...

paperboyo:
> things like `by`

If that's indeed gonna get added (quite a list!), I volunteer to correct the first 1000 names to contain proper diacritics (if we can **teach** ES to use the corrected ones). Oh, shoe! Make it 1500 :)

theefer:
Added suggestions for agency, category and byline (only our photographers for now since we can't list all bylines).

paperboyo:
:happy_bunny:

> our photographers

Illustrators and such as well?

My suggestion about `byline` cleanup is obviously beyond the scope of this PR, but I hope it's gonna be possible one day.

theefer:
> Illustrators and such as well?

Shouldn't that be `illustrator:`? Could add that if useful.

paperboyo:
Yeah, it should, I think it might be useful, although there are not that many... I would add if not a problem.

Photographers would collate just staff and contract? Or commissioned as well (not sure how many are here and how quickly it grows). Which makes me want to see what happens if the suggestion list if too long. And how it sorts suggestions (global or user search popularity)...

Oh, I wanna see :eyes:

theefer:
@paperboyo I've now added illustrators:

recorded

paperboyo:
:beer: and then :beer:

theefer:
@tsop14 'selected' collection now fixed.

tsop14:
🙌

theefer:
> Could you make it that the first time you click in the box it gives you something with a "dismiss" button to make it more obvious as this is a great feature.

I like that idea. I think it'd be worth bedding in the feature with a few users first, and maybe doing this "New feature" built-in announcement alongside revamping the "Advanced search" (or maybe it should just be Help? or simply (?)) and sending comms about the feature.

No I'm not just trying to delay extra work until I'm gone, but I realise that due to time pressure, I haven't done enough work with users as I should have done :disappointed:

tsop14:
> alongside revamping the "Advanced search" (or maybe it should just be Help? or simply (?)) and sending comms about the feature.

Could have a '?' which clicks to reveal a clickable list of all the available search terms. When you can click on e.g. 'by' it appends empty 'by' pill to search bar for you to fill in.

theefer:
> Could have a '?' which clicks to reveal a clickable list of all the available search terms. When you can click on e.g. 'by' it appends empty 'by' pill to search bar for you to fill in.

That's exactly what I had in mind, alongside a very simple help to tell user about `+`, `-` and maybe `#`.

theefer:
@tsop14 I think I've fixed the weight height issues with the query field, changes building now..

Issue #1590, opened 2016-01-04
Collections


jamesgorrie:
Things required:
- [x] [Create collections](https://github.com/guardian/grid/pull/1576)
- [x] [Add to collections from results](https://github.com/guardian/grid/pull/1576)
- [x] Add to collections on upload
- [x] [View collections on results](https://github.com/guardian/grid/pull/1579)
- [x] View collections on image page
- [x] [Tree state saving](https://github.com/guardian/grid/pull/1593)

tsop14:
- [ ] Remove image from collection

akash1810:
- [ ] hide panels when printing

kenoir:
- [ ] Remove collection from image when collection deleted

kenoir:
This is substantially complete.

PR #1781, opened 2016-02-23
add to collections on upload

tsop14:
#1590 Adds ability to add/remove images to/from collections on upload page.

Remembers open/closed nodes and position on collections tree. 

Can batch add collections on one image to all recent uploads. 
Can remove all collections by 'adding' no collections (as per labels). 

! If the same image is upload again and it was previously added to collections, it will only display the collections once it is added to a collection. So, the collection is present in the media-collection api but not on the image's collections data in the media api and it only fetches the collections list when collections are edited.

collections on uploads

### DIFF
--- a/kahuna/public/js/edits/image-editor.js
+++ b/kahuna/public/js/edits/image-editor.js
@@ -37,10 +37,34 @@ imageEditor.controller('ImageEditorCtrl', [
 
     var ctrl = this;
 
-    ctrl.status = ctrl.image.data.valid ? 'ready' : 'invalid';
-    ctrl.saving = false;
-    ctrl.saved = false;
+    ctrl.batchApplyCollections = batchApplyCollections;
+    ctrl.batchApplyUsageRights = batchApplyUsageRights;
+    ctrl.categories;
     ctrl.error = false;
+    ctrl.image;
+    ctrl.saved = false;
+    ctrl.saving = false;
+    ctrl.showUsageRights;
+    ctrl.status = ctrl.image.data.valid ? 'ready' : 'invalid';
+    ctrl.usageRights = imageService(ctrl.image).usageRights;
+    ctrl.usageRightsCategory;
+
+
+    //TODO put collections in their own directive
+    ctrl.addCollection;
+    ctrl.addToCollection = addToCollection;
+    ctrl.batchRemoveCollections;
+    ctrl.collectionError = false;
+    ctrl.collections;
+    ctrl.confirmDelete;
+    ctrl.openCollectionTree = openCollectionTree;
+    ctrl.removeImageFromCollection = removeImageFromCollection;
+    ctrl.selectionMode = true;
+
+    // TODO: Find a way to broadcast more selectively
+    const batchApplyUsageRightsEvent = 'events:batch-apply:usage-rights';
+    const batchApplyCollectionsEvent = 'events:batch-apply:collections';
+    const batchRemoveCollectionsEvent = 'events:batch-remove:collections';
 
     const metadata = ctrl.image.data.userMetadata.data.metadata;
     const usageRights =  ctrl.image.data.userMetadata.data.usageRights;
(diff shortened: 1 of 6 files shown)


kenoir:
Nice.

kahuna/public/js/edits/image-editor.js
```diff
@@ -5,13 +5,17 @@ import './service';
...
+import {rememberScrollTop} from '../directives/gr-remember-scroll-top';
 
 
 export var imageEditor = angular.module('kahuna.edits.imageEditor', [
     'kahuna.edits.service',
     'gr.image.service',
     'kahuna.edits.usageRightsEditor',
-    'gr.archiverStatus'
+    'gr.archiverStatus',
+    collectionsApi.name,
```
kenoir: Why this way and not the name as a string?

paperboyo:
Super-useful! Am I seeing correctly that plus icon for collections is different to the plus icon for labels? Am I correct that they should be the same? :)

kahuna/public/js/edits/image-editor.js
tsop14: For funsies? To make it easier to read? Er...

tsop14:
@paperboyo 
screen shot 2016-02-24 at 11 33 21

paperboyo:
Thanks, @tsop14. Have to buy meself glasses... :eyeglasses:

tsop14:
Haha, they were different! Now they are the same though... and that's all that matters.

kahuna/public/js/edits/image-editor.js
jamesgorrie: I would argue moving towards this way as it's stops having funny-little-strings-that-can-change-at-any-time lying all over the place.

Strings are for sentences.

jamesgorrie: And JSON.

kahuna/public/js/edits/image-editor.js
```diff
@@ -115,6 +121,74 @@ imageEditor.controller('ImageEditorCtrl', [
...
+    ctrl.collectionError = false;
+
+    ctrl.selectionMode = true;
+
+    ctrl.openCollectionTree = () => {
+        ctrl.addCollection = true;
+
+        if (!ctrl.collections) {
+            collections.getCollections().then(collections => {
+                ctrl.collections = collections.data.children;
```
jamesgorrie: You could memoize this in the API, and then not have to worry about performance here.

kahuna/public/js/edits/image-editor.js
theefer: > Strings are for sentences.

:-)

kahuna/public/js/edits/image-editor.js
tsop14: Sounds good - can you vary up how often the result gets reset? What happens if you're trying to add to collection someone else recently created?

kahuna/public/js/edits/image-editor.js
```diff
@@ -21,20 +25,46 @@ imageEditor.controller('ImageEditorCtrl', [
...
+    ctrl.categories;
     ctrl.error = false;
+    ctrl.image;
+    ctrl.saved = false;
+    ctrl.saving = false;
+    ctrl.showUsageRights;
+    ctrl.status = ctrl.image.data.valid ? 'ready' : 'invalid';
+    ctrl.usageRights = imageService(ctrl.image).usageRights;
+    ctrl.usageRightsCategory;
+
```
tsop14: Read you should have all the bindables at the top, so gave it a go.

tsop14:
Colours 💜💛💚💙

screen shot 2016-02-25 at 12 34 00

kahuna/public/js/edits/image-editor.js
tsop14: > You could memoize this in the API, and then not have to worry about performance here.

This is already being done 😁

NickPapacostas:
:+1:

PR #1784, opened 2016-02-25
removes collections panel for printing

tsop14:
#1590 

Still prints 11 blank pages on chrome 😰but only one on Firefox... 

screen shot 2016-02-25 at 09 53 19


theefer:
:+1:

PR #1785, opened 2016-02-25
removes track by on ng-repeat

tsop14:
Fixes: Error: [ngRepeat:dupes] Duplicates in a repeater are not allowed. 

Fixes images with two sibling collections with the same name in different casing, not having their collections displayed on search page.


theefer:
:+1:

PR #1786, opened 2016-02-25
makes header sticky on collections panel

tsop14:
Makes edit bar sticky. As we now remember scroll position you cannot see if you are in edit mode - will be unable to add images to collections. Also prevents repetitive scrolling when adding subcollections. 

sticky collections header


akash1810:
:+1:

PR #1787, opened 2016-02-25
sets collections header z value

tsop14:
NOW
screen shot 2016-02-25 at 16 45 04

BEFORE
screen shot 2016-02-25 at 16 44 48


theefer:
:+1: 

When playing with z-index, always worth checking if this conflicts with other overlays, modals, etc.

tsop14:
Took the value from main.css notes

> 10 = things to be just above everything

PR #1788, opened 2016-02-25
[do not merge] show the link to frontend iff published

akash1810:
we could add a link to preview/viewer at a later point


NickPapacostas:
:+1:

akash1810:
This isn't actually working as only the icon doesn't get rendered, the `a` tag is still on the DOM.

A cleaner solution would be to adapt what the API returns. Currently it returns this:

```json
"data": {
  "id": "XXXX",
  "references": [
    {
      "type": "frontend",
      "uri": "http://www.theguardian.com/XXXX",
      "name": "XXXX"
    },
    {
      "type": "composer",
      "uri": "https://composer._.co.uk/content/XXXX"
     }
  ],
  "platform": "digital",
  "media": "image",
  "status": "pending"
}
```

I'm going to change the API to not return a frontend link if `status` === `pending` (with the added bonus of keeping logic out of the client).

PR #1790, opened 2016-02-26
Add issue and PR templates

akash1810:
GitHub have recently added issue and PR templates https://github.com/blog/2111-issue-and-pull-request-templates.

Although we follow Tackley's Law (TM) on creating issues, the template would act as a prompt for the details needed.

Inspired by https://github.com/guardian/frontend/pull/12028

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,17 @@
+# ISSUE
+
+## Steps to Reproduce
+
+## Actual Results (include screenshots)
+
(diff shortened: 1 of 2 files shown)


kenoir:
Reluctant really to add this. I think this is mostly common sense and is going to become annoying when what we're doing doesn't quite fit the template.

I do think there is a reason for this where there is a higher cost to failure or in larger teams where members might not be familiar with the guidelines.

paperboyo:
templates

Unless the template contains
Regards
Mateusz

PR #1791, opened 2016-02-26
adds Janus support for cf actions

akash1810:
If you're using temporary credentials, `aws iam get-user` fails so `STACK_NAME` is not set correctly.

Fail gracefully and suggest setting the envvar which, if set,  will be preferred over the result of `aws iam get-user`.


akash1810:
Should also mean people new to the team won't need permanent credentials at all! :heart: Janus :heart:

akash1810:
merging as @ShaunYearStrong is joining the team and needs this!

PR #1792, opened 2016-02-26
adding square crop to aspect ratio calculation

NickPapacostas:
seb's mention of this feature in his talk made me realize we probably hadn't added it here, 

so, true to his process vision, i just fixed it. ^_^


kenoir:
:+1:

PR #1795, opened 2016-03-01
Aggregates query

NickPapacostas:
adding new aggregations namespace in routes and functionality for structured queries along with aggregations, also added general DateHistogram aggregation endpoint to be used for usage analysis

* the reason i made the new controller and namespace is that previously suggestions was the only thing using aggregation and thus had that functionality under it, but now we have a reason to pull it out. there are probably more things that can be pulled out of suggestions but i think this is a good first pass.

No previous functionality should be effected. (which i've tested locally)

Examples!

Structured Queries with aggregates:

*before* 
https://api.media.local.dev-gutools.co.uk/images/metadata/credit

returned:
```json
[
   {
      key:"AFP/Getty Images",
      count:249
   },
   {
      key:"AP",
      count:167
   },
   {
      key:"Getty Images",
      count:162
   },
   {
      key:"Reuters",
      count:134
   }

....
```

*Now* you can add a query:
https://api.media.local.dev-gutools.co.uk/images/metadata/credit?q=usages@status:pending

and it will get parsed like it does for normal searches, and filter the aggregations accordingly

Aggregations example:

Right now the only thing under "aggregations" is the date histogram feature, which provides you with interval-based (right now just monthly) counts of date fields based on search queries (...confusing >.<)

example:
If I want monthly usages where status is pending and credit is "REX/Shutterstock" I could do:

https://api.media.local.dev-gutools.co.uk/images/aggregations/date/usages.dateAdded?q=usages@status:pending%20credit:REX/Shutterstock

and get something like this:

```json
data:[
   {
      key:"2014-07-01T00:00:00.000Z",
      count:1
   },
   {
      key:"2016-01-01T00:00:00.000Z",
      count:1
   }
]
}
```

### DIFF
--- a/media-api/app/controllers/AggregationController.scala
+++ b/media-api/app/controllers/AggregationController.scala
@@ -7,9 +7,6 @@ import play.api.mvc.{AnyContent, Request, Controller}
 
 import scala.util.Try
 
-/**
-  * Created by npapacostas on 01/03/2016.
-  */
 object AggregationController extends Controller {
   val Authenticated = Authed.action
 
--- a/media-api/app/controllers/AggregationController.scala
+++ b/media-api/app/controllers/AggregationController.scala
@@ -1,0 +1,40 @@
+package controllers
+
+import play.api.libs.concurrent.Execution.Implicits._
+import lib.elasticsearch.ElasticSearch
+import lib.querysyntax.{Parser, Condition}
+import play.api.mvc.{AnyContent, Request, Controller}
+
+import scala.util.Try
+
+/**
+  * Created by npapacostas on 01/03/2016.
+  */
+object AggregationController extends Controller {
+  val Authenticated = Authed.action
+
+  def dateHistogram(field: String, q: Option[String]) = Authenticated.async { request =>
+    ElasticSearch.dateHistogramAggregate(AggregateSearchParams(field, request))
+      .map(ElasticSearch.aggregateResponse(_))
+  }
+
+}
+
+case class AggregateSearchParams(
+                                  field: String,
+                                  q: Option[String],
+                                  structuredQuery: List[Condition])
+
+object AggregateSearchParams {
+  def parseIntFromQuery(s: String): Option[Int] = Try(s.toInt).toOption
+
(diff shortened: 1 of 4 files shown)


media-api/app/controllers/AggregationController.scala
```diff
@@ -0,0 +1,40 @@
...
+
+import play.api.libs.concurrent.Execution.Implicits._
+import lib.elasticsearch.ElasticSearch
+import lib.querysyntax.{Parser, Condition}
+import play.api.mvc.{AnyContent, Request, Controller}
+
+import scala.util.Try
+
+/**
+  * Created by npapacostas on 01/03/2016.
```
akash1810: :smiley_cat:

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -214,7 +235,12 @@ object ElasticSearch extends ElasticSearchClient with SearchFilters with ImageFi
   def completionSuggestionBuilder(name: String) = new CompletionSuggestionBuilder(name)
 
   def searchResultToAggregateResponse(response: SearchResponse, aggregateName: String) = {
-    val buckets = response.getAggregations.getAsMap.get(aggregateName).asInstanceOf[InternalTerms].getBuckets
+    val buckets = response.getAggregations
+      .getAsMap
+      .get(aggregateName)
+      .asInstanceOf[MultiBucketsAggregation]
```
akash1810: why `asInstanceOf` and not a `match`?

kenoir: It might be more concise here.

kenoir:
Can you give some example queries and responses in the description?

media-api/app/controllers/AggregationController.scala
NickPapacostas: *smh*

kenoir:
:+1:

media-api/app/lib/elasticsearch/ElasticSearch.scala
akash1810: how come?

PR #1796, opened 2016-03-02
Replace created date with publication date

kenoir:
Note: this will require a reimport of all usages from the picdar export
db!


NickPapacostas:
:+1:

PR #1797, opened 2016-03-03
Aggregate empties

NickPapacostas:
Allows you to get all elements in an interval, e.g. I want all the usages in a time period, but ES will exclude months with no data, which makes it bad for graphing. This way all months will be returned whether or not there was a usage

i thought about making this parameterizable but figured right now I'm the only one with a use case so we can leave that for later


kenoir:
:+1:

PR #1800, opened 2016-03-03
sticky edit bar on collections

tsop14:
same same but better #1786

Makes bar sticky but hides with panel. 

hidden header


NickPapacostas:
:+1:

PR #1805, opened 2016-03-10
Simplify imgops setup, use system nginx instead

ShaunYearStrong:
After having a hell of a time installing everything on El Capitan, @akash1810 and I simplified the setup a bit. It means the imgops is running on the system nginx, but also means there's no compilation issues due to El Capitan not playing nice with openssl.

### DIFF
--- a/imgops/README.md
+++ b/imgops/README.md
@@ -3,7 +3,7 @@ Local version of the imgops service
 
 ## Requirements
 * [Nginx with with image filter module](http://nginx.org/en/docs/http/ngx_http_image_filter_module.html)
-  * Linux: `sudo apt-get install nginx`
+  * Linux: `sudo apt-get install nginx nginx-extras`
   * Mac: `brew install nginx-full --with-image-filter`
 * [GD](http://libgd.github.io/)
   * Linux: `sudo apt-get install libgd-dev`
--- a/imgops/README.md
+++ b/imgops/README.md
@@ -2,6 +2,9 @@
 Local version of the imgops service
 
 ## Requirements
+* [Nginx with with image filter module](http://nginx.org/en/docs/http/ngx_http_image_filter_module.html)
+  * Linux: `sudo apt-get install nginx`
+  * Mac: `brew install nginx-full --with-image-filter`
 * [GD](http://libgd.github.io/)
   * Linux: `sudo apt-get install libgd-dev`
   * Mac:  `brew install gd`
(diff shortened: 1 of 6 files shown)


akash1810:
I think this would also make [dev-start.sh](https://github.com/guardian/grid/blob/master/imgops/dev-start.sh) redundant too, so should be safe to remove it.

imgops/README.md
```diff
@@ -2,6 +2,9 @@
 Local version of the imgops service
 
 ## Requirements
+* [Nginx with with image filter module](http://nginx.org/en/docs/http/ngx_http_image_filter_module.html)
+  * Linux: `sudo apt-get install nginx`
```
akash1810: You'd also need to install the `nginx-extras` package to get the image filter module on linux.

akash1810:
> so should be safe to remove it.

And also update the (many) [run scripts](https://github.com/guardian/grid/blob/master/docs/running.md) and documentation, removing the need to start imgopts. Also the [imgopts readme](https://github.com/guardian/grid/blob/master/imgops/README.md).

akash1810:
other than updating the run scripts, looks good!

ShaunYearStrong:
@akash1810 - Remove it from the shell scripts and readme. @NickPapacostas Would have to do it for the grid runner gem after this goes in. Wasn't sure if it needed to be removed in the docker file, with it being a container.

akash1810:
[Grid Runner](https://github.com/guardian/grid_runner) is on GH :smile: 

The Dockerfile wouldn't need updating as it doesn't use the `dev-start.sh` script. In docker land, the imgops container is an nginx container which uses this [server template](https://github.com/guardian/grid/blob/master/docker/configs/generators/generators/resources/templates/imgops/nginx.conf.template).

akash1810:
^ I can explain clearer IRL!!

akash1810:
:+1:

akash1810:
Just realised we missed the [Procfile](https://github.com/guardian/grid/blob/master/Procfile#L9). That should be the [last reference](https://github.com/guardian/grid/search?utf8=%E2%9C%93&q=dev-start.sh)!

imgops/README.md
```diff
@@ -2,6 +2,9 @@
 Local version of the imgops service
 
 ## Requirements
+* [Nginx with with image filter module](http://nginx.org/en/docs/http/ngx_http_image_filter_module.html)
+  * Linux: `sudo apt-get install nginx nginx-extras`
+  * Mac: `brew install nginx-full --with-image-filter`
```
tsop14: I had to use `brew install homebrew/nginx/nginx-full --with-image-filter` as I didn't have the homebrew nginx tap installed. Will new users have it installed?

ShaunYearStrong: Good point, I'll add the line in

akash1810: > Will new users have it installed?

No. You'd have to `brew tap homebrew/nginx-full` first then `brew install` or `brew install homebrew/nginx/nginx-full --with-image-filter` as you did.

PR #1806, opened 2016-03-10
Expose user permissions

kenoir:
Expose permissions for current user in order to provide this info to the UI client so controls can be varied according to permissions.

This is specifically to add controls around paid imagery.

Example response:

`GET /permissions`

```json
{
  "data": {
    "description": "Available permissions for current user (you)",
    "permissionsSet": {
      "user": {
        "email": "nope@guardian.co.uk",
        "firstName": "Robert",
        "lastName": "Kenny",
        "avatarUrl": "https://lh3.googleusercontent.com/nope/photo.jpg"
      },
      "permissions": [
        "editMetadata",
        "deleteImage",
        "deleteCrops",
        "bigSpender"
      ],
      "lastUpdated": 1458298190366
    }
  }
}
```

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Store.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Store.scala
@@ -68,9 +68,20 @@ object PermissionType extends Enumeration {
   val EditMetadata = Value("editMetadata")
   val DeleteImage  = Value("deleteImage")
   val DeleteCrops  = Value("deleteCrops")
+  val BigSpender   = Value("bigSpender")
 }
-
 class PermissionStore(bucket: String, credentials: AWSCredentials) extends BaseStore[PermissionType, List[String]](bucket, credentials) {
+
+  type FuturePerms = Future[Set[PermissionType.PermissionType]]
+
+  def getUserPermissions(
+    user: PandaUser
+  ): FuturePerms = store.future.map {
+    _.filter {
+      case (_, list) => list.contains(user.email.toLowerCase)
+    }.keys
+  }.map(_.toSet)
+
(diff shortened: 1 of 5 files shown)


media-api/app/controllers/PermissionsController.scala
```diff
@@ -0,0 +1,49 @@
...
+object PermissionsController extends Controller with ArgoHelpers {
+
+  val Authenticated = Authed.action
+  val permissionStore = Authed.permissionStore
+
+  def permissionsResponse(permissions: Set[PermissionType.PermissionType]) = {
+    val perms = Json.toJson(permissions.map(_.toString))
+
+    val permsData = Json.obj(
+      "description" -> "Available permissions for current user (you)",
```
akash1810: worth adding the current username/email here?

common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Store.scala
```diff
@@ -68,9 +68,20 @@ object PermissionType extends Enumeration {
   val EditMetadata = Value("editMetadata")
   val DeleteImage  = Value("deleteImage")
   val DeleteCrops  = Value("deleteCrops")
+  val BigSpender   = Value("bigSpender")
```
akash1810: might want to add this key to the [sample permissions file](https://github.com/guardian/grid/blob/master/cloud-formation/scripts/permissions.properties) too

akash1810:
The `permissionStore` is updated on an Akka schedule, would it be worth adding a `lastUpdated` property?

kenoir:
Closing for now, as we're unsure how helpful this will be.

NickPapacostas:
:+1:

PR #1808, opened 2016-03-11
Collections: Check for child nodes before delete

kenoir:
This PR adds a server side check to node deletion to ensure nodes with
children are not deleted.

This is in order to avoid problems when UI collection trees are out of
sync.

### DIFF
--- a/collections/app/controllers/CollectionsController.scala
+++ b/collections/app/controllers/CollectionsController.scala
@@ -22,6 +22,7 @@ import com.gu.mediaservice.model.{ActionData, Collection}
 import store.{CollectionsStoreError, CollectionsStore}
 
 
+case class HasChildrenError(message: String) extends Throwable
 case class InvalidPrinciple(message: String) extends Throwable
 case class AppIndex(name: String, description: String, config: Map[String, String] = Map())
 
@@ -63,6 +64,9 @@ object CollectionsController extends Controller with ArgoHelpers {
   def invalidJson(json: JsValue) =
     respondError(BadRequest, "invalid-json", s"Could not parse json: ${json.toString}")
 
+  def invalidTreeOperationError(message: String) =
+    respondError(BadRequest, "invalid-tree-operation", message)
+
   def storeError(message: String) =
     respondError(InternalServerError, "collection-store-error", message)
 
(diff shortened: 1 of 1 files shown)


kenoir:
@jamesgorrie checks for children as discussed.

collections/app/controllers/CollectionsController.scala
```diff
@@ -2,6 +2,8 @@ package controllers
 
 import java.net.URI
 
+import scala.util.Try
```
NickPapacostas: I dont think this gets used in the end

collections/app/controllers/CollectionsController.scala
```diff
@@ -126,11 +137,35 @@ object CollectionsController extends Controller with ArgoHelpers {
...
+
+      // Traverse the tree using the path
+      val maybeTree = path
+        .foldLeft[MaybeTree](Some(tree))((optBranch, nodeName) => {
+
+          optBranch.flatMap { _.children.find(_.basename == nodeName) }
+        })
+
+      // Does out target node have children?
+      !maybeTree.flatMap(_.children.headOption).isEmpty
```
NickPapacostas: intelliJ is suggesting isDefined vs !isEmpty, might be a little clearer but doesn't matter really

collections/app/controllers/CollectionsController.scala
```diff
@@ -126,11 +137,35 @@ object CollectionsController extends Controller with ArgoHelpers {
     } getOrElse Future.successful(invalidJson(req.body))
   }
 
+  type MaybeTree = Option[Node[Collection]]
+  def hasChildren(path: List[String]): Future[Boolean] =
+    allCollections.map { tree =>
```
NickPapacostas: Does it make more sense to retrieve the collection by using dynamo.get? 

it seems like this is traversing all the collections to find the selected one, but I think they're all stored individually in dynamo, so it might be faster. Could also be nice to have a collections/:id endpoint that would use the same method, but maybe that doesnt matter.

NickPapacostas:
:+1:  

no blockers, just some comments

collections/app/controllers/CollectionsController.scala
kenoir: I think the structure of the data store prevents this from being efficient. ccing @jamesgorrie to see if he thinks different.

Issue #1801, opened 2016-03-04
Mischievous Chips


tsop14:
😈🍟

- [ ] Double-clicking to select all doesn't work if you do it in the are after the word.
- [ ] When try to select text with a broad click-drag, depending on where do you start dragging (still: always within the text box area), different glitches happen in different browsers... 
- [ ] Cannot Paste (from the context menu).

paperboyo:
[] :fries: do not play well with `usages@` syntax. `usages` falls out of a chip that only says e.g. `status:` and `@` is hidden and... just have a play :)

paperboyo:
- [ ] :fries: bleed out of the searchbox on latest beta and nightly of Firefox on Win10 like so:

![chips bleed](https://cloud.githubusercontent.com/assets/6032869/13557544/5c780728-e3ea-11e5-824b-cf42220ff2c8.PNG)

NickPapacostas:
:fries: they're trying to be clever with nested queries
- [ ] query limited to size of chip when it aint a chip!
(can't reproduce, just keep an eye out)

- [ ] when I do a nested query, click on an image, and then go back to search, chips kick in incorrectly
the query in the example below was "usages@>added:2015"
![screen shot 2016-03-07 at 12 27 40](https://cloud.githubusercontent.com/assets/1441582/13569460/1208b892-e460-11e5-87da-fe9cbeeb2ca9.png)

paperboyo:
- [ ] :fries: embarrassingly reveal the lack of UI responsiveness when cut long enough:

![chips-long](https://cloud.githubusercontent.com/assets/6032869/14135680/c88e8a12-f656-11e5-9c6f-04d45d5738ce.png)

paperboyo:
- [ ] Trying to add a `#` at the beginning of the search term (sound effects are: aaarrrrgh!! and nooooo!):

![untitled](https://cloud.githubusercontent.com/assets/6032869/14210610/a65cb168-f821-11e5-8a65-9d2bfe20540c.gif)

tsop14:
Think  #1830 will fix this?

paperboyo:
Oh cool, haven't connected the dots, sorry! Can I have a play with crop, pretty please (at any point, incl. when it's in PROD ;-) ).

tsop14:
Yep, the branch will do this 👉 ![add label](https://cloud.githubusercontent.com/assets/8484757/14210787/7cf15d1e-f822-11e5-852e-8f61f1093dec.gif)

[...]

PR #1809, opened 2016-03-15
Display search chips to the left, to the left

tsop14:
chips

Once created, pushes chips to the right of the search box.
Only  has ONE text chip which stays on the left. 

#1801

### DIFF
--- a/kahuna/public/js/components/gr-chips/gr-filter-chooser-chip.js
+++ b/kahuna/public/js/components/gr-chips/gr-filter-chooser-chip.js
@@ -54,12 +54,12 @@ grFilterChooserChip.controller('grFilterChooserChipCtrl', ['$timeout', function(
             } else {
                 // Not a key, turn it back to text search
                 $grChipsCtrl.focusEndOfFirstChip();
-                chip.value = "FIELD NOT FOUND";
+                chip.value = 'FIELD NOT FOUND';
                 $grFilterChooserChipCtrl.notValid = true;
 
                 $timeout(() => {
                     $grFilterChooserChipCtrl.removeSelf();
-                    $grChipsCtrl.items[0].value += (" " + value);
+                    $grChipsCtrl.items[0].value += (' ' + value);
                 }, 800);
 
             }
--- a/kahuna/public/js/components/gr-chips/gr-filter-chooser-chip.js
+++ b/kahuna/public/js/components/gr-chips/gr-filter-chooser-chip.js
@@ -16,7 +16,7 @@ export const grFilterChooserChip = angular.module('gr.filterChooserChip', [
     autoWidth.name
 ]);
 
-grFilterChooserChip.controller('grFilterChooserChipCtrl', ['$timeout', function($timeout) {
+grFilterChooserChip.controller('grFilterChooserChipCtrl', function() {
     const $grFilterChooserChipCtrl = this;
 
     let $grChipsCtrl;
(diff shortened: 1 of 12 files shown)


NickPapacostas:
:+1: definitely nicer 

general thoughts:
- min width is better but still gets crowded fast on a small screen (potentially no solution for this)
- colorz! so we can see the chips more clearly all the way over there ^_^

ShaunYearStrong:
I agree with @NickPapacostas regarding colours, it'll mean you don't miss the chip appearing as your focus would be on the cursor at the time.

NickPapacostas:
@ShaunYearStrong  now i feel obligated to mention that the colors were @tsop14 's idea that she mentioned earlier and i unashamedly stole!

ShaunYearStrong:
Okay then, I agree with @tsop14 then, and revoke my support for you!

NickPapacostas:
:ideograph_advantage:

tsop14:
- Creates on text box on the left for free search. 
- In this search box you can select, paste and delete as normal. 
- Moves all chips to the left. 
chips deleting


- Removes the 'any' chip. 
- When you create a chip with a field that is not in the valid list, it returns it to text
chips

! You cannot select the text and the chips and start typing again.

kahuna/public/js/components/gr-chips/gr-filter-chooser-chip.js
```diff
@@ -52,19 +52,16 @@ grFilterChooserChip.controller('grFilterChooserChipCtrl', function() {
...
+                // Not a key, turn it back to text search
+                $grChipsCtrl.focusEndOfFirstChip();
+                chip.value = "FIELD NOT FOUND";
+                $grFilterChooserChipCtrl.notValid = true;
+
+                $timeout(() => {
+                    $grFilterChooserChipCtrl.removeSelf();
+                    $grChipsCtrl.items[0].value += (" " + value);
+                }, 800);
+
```
NickPapacostas: why the delay?

NickPapacostas:
:+1:  :ideograph_advantage:

kahuna/public/js/components/gr-chips/gr-filter-chooser-chip.js
tsop14: Gives the user time to read the 'FIELD NOT FOUND' message (which is inside the chip) before said chip gets removed. 
Thoughts?

tsop14:
Usages parsing fixes: 

NOW
screen shot 2016-03-21 at 17 03 26

BEFORE

screen shot 2016-03-21 at 17 02 42

Also creates one text 'chip' from all the text section of the parsed query and passes this through as the model value. Rather than 'normaliZing' in the gr-chips controller.

kahuna/public/js/search/structured-query/syntax.js
```diff
@@ -39,7 +39,27 @@ export function structureQuery(query) {
...
+
+    const cleanStruct = [textChip].concat(filterChips);
+
+    function mergeTextValues(chips){
+        return chips.filter(e => e.type === 'text')
+            .map(e => e.value)
+            .join(' ');
+    }
+
+    return cleanStruct;
```
NickPapacostas: we could have just called it "struct" :disappointed:

NickPapacostas:
:+1: :ideograph_advantage:

kahuna/public/js/search/structured-query/syntax.js
tsop14: :boar:

PR #1810, opened 2016-03-15
Adds hasRightsCategory filter

kenoir:
In order that paid image searches can be narrowed to those with a rights category (and consequently definitely pay-for imagery).


NickPapacostas:
:+1:

paperboyo:
This is very useful! There is no way to use it from within the UI, though, is there?

PR #1811, opened 2016-03-16
adding structured query tracking

NickPapacostas:
See comments for a question


kahuna/public/js/search/structured-query/structured-query.js
```diff
@@ -67,10 +72,24 @@ grStructuredQuery.directive('grStructuredQuery', ['subscribe$', function(subscri
                 const queryString = ngModelCtrl.$viewValue || '';
                 ctrl.structuredQuery = structureQuery(queryString);
             };
-
+            
             subscribe$(scope, ctrl.newQuery$, query => {
```
NickPapacostas: I wanted to add another "subscribe$" call to the ctrl.newQuery$ stream but the last one declared was always the only one that ran. So i put them in the same block, but i perhaps i missed something simple

kenoir:
:+1:

PR #1814, opened 2016-03-16
add to collection from image page

tsop14:
Makes collection overlay from uploads page into a directive.
Adds collection overlay directive to image page. 

adding image


kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
```diff
@@ -27,6 +28,7 @@ module.controller('grImageMetadataCtrl', [
     function ($rootScope,
               $scope,
               $window,
+              $timeout,
```
kenoir: Where is this used?

kenoir:
Other than ^ :+1:

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
tsop14: Good spot! Was added before made a directive. Not needed. Shoo timeout.

PR #1815, opened 2016-03-16
Allow different views of "pay" content

kenoir:
Creates and sets a default a `maybeFree` filter that shows free content and content with no
rights.

screenshot from 2016-03-17 15 43 23


NickPapacostas:
is "Free and no rights" obvious to other people? I don't quite get it, but that might just be me.

NickPapacostas:
:+1:

kenoir:
@NickPapacostas not necessarily. I might wait on clarification from @blishen

PR #1820, opened 2016-03-18
Re-writes advanced search panel info

tsop14:
Creates a directive for an example chip.
Re-writes the advance panel.
Makes more beautiful 💅

screen shot 2016-03-18 at 17 29 09

screen shot 2016-03-18 at 17 29 17

### DIFF
--- a/kahuna/public/js/directives/gr-chip-example.js
+++ b/kahuna/public/js/directives/gr-chip-example.js
@@ -1,0 +1,24 @@
+import angular from 'angular';
+
+export const grChipExample = angular.module('gr.chipExample', []);
+
+
+grChipExample.directive('grChipExample', [function(){
+    return {
+        restrict: 'E',
+        template: `<span class="gr-filter-chip">
+        <span class="gr-filter-chip__type">{{grExclude}}</span>
+        <span class="gr-filter-chip__key">{{grFilterField}}:&nbsp</span>
+        <span>{{grExampleSearch}}&nbsp</span>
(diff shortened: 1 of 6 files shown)


paperboyo:
This is amazing!

kenoir:
Much better.

NickPapacostas:
:+1:

paperboyo:
Maybe to save on some height, it could collate inclusion and exclusion:
"Type `+` to include or `-` to exclude:".

Also, for clarity, maybe "For existing filters, **click** `+` on the  left hand side to **toggle** inclusion and exclusion" would be clearer (apart from "inclusion and exclusion" part :-S).

As it never really was "Advanced", maybe we could save some space by replacing this text with an e.g. https://design.google.com/icons/#ic_info?

It says "filtering examples", it would be better to have a comprehensive list here of **all** filters. And to be able to treat usages similarly to the rest ;-).

kenoir:
I love @paperboyo's idea of replacing the "Advanced" text with the info icon.

tsop14:
> As it never really was "Advanced", maybe we could save some space by replacing this text

screen shot 2016-03-22 at 12 48 49


> be able to treat usages similarly to the rest

screen shot 2016-03-22 at 12 42 15

> As it never really was "Advanced",

Yep but have kept reference to 'advance' here to make it clear this is the same as before. 
screen shot 2016-03-22 at 12 45 42

ShaunYearStrong:
:+1:

PR #1821, opened 2016-03-21
Quick fix for usage chips

kenoir:
This PR fixes the most common use of usages in search results. Ideally
the structured query lib should know about nested queries but this will
suffice for now.


NickPapacostas:
clever! :+1:

PR #1823, opened 2016-03-22
corrects square shortcut and makes all lowercase

tsop14:
Makes all keyboard shortcuts lowercase - upper case letters do not work as shortcut. 
Corrects letter for square crop to 'v' -> 's'. 

Fixes: 
screen shot 2016-03-22 at 17 44 32


NickPapacostas:
:+1: :ideograph_advantage:

akash1810:
`l` was regarded as unclear - is it an uppercase `i` or a lowercase `l`? The answer is fairly intuitive, however there was a concern it wasn't super obvious, especially in the keyboard shortcut modal; a different font family may help??

PR #1825, opened 2016-03-23
prompt to alter search when no results

akash1810:
The embedded Grid defaults search for Video pages, setting a label search. However, its not guaranteed that any images will exist and its been (incorrectly) said that "the Grid doesn't work". This PR prompts to check your search if this is the case.

I've started creating a `clear-search` component that can be placed in next to this message, however need to share state between controllers, so need a factory and so refactoring :unamused:. I'll carry on with this on a branch, however I think this is a good first step to solving the problem.

/cc @jamesgorrie


kenoir:
:+1:

PR #1826, opened 2016-03-29
Allow negated text searches

kenoir:
The ability to do these was removed with the removal of the "any" chip.


NickPapacostas:
:+1: just noting that we're intentionally de-chipping negation, and removing the "any" chip concept

tsop14:
:+1:

PR #1828, opened 2016-03-29
"No Rights" results in an "invalid" image

kenoir:
In order to prevent use of "No Rights" images and to encourage entering better rights data on upload.

**Image page**
imagep_inv_c

**Search result**
search_inv

**Upload page**
upload_inv

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
@@ -122,7 +122,10 @@ case object NoRights
   val restrictions = None
   val name = "No Rights"
   val description =
-    "Images for which we do not have the rights to use."
+    "Images which we do not currently have the rights to use."
+
+  override val caution =
+    Some("You cannot use this image in content!")
 
   lazy val jsonVal = Json.obj()
 
--- a/kahuna/public/js/components/gr-crop-image/gr-crop-image.js
+++ b/kahuna/public/js/components/gr-crop-image/gr-crop-image.js
@@ -18,6 +18,7 @@ cropImage.controller('grCropImageCtrl', [
         }
 
         $scope.$watch(() => ctrl.image.data.metadata, onValChange(() => updateState()));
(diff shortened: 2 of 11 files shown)


paperboyo:
Worth making invalid more obvious/different from paid? E.g.

invalid_obvious
[added (ugly!) sign and a red frame around UR, haven't "updated" the tooltip]

Possible downside of having an image overlay would be making harder a use case of using a Grid as a temporary selection machine (I will upload these and decide which one(s) to use later...). For users that do that (or using lowres positionals) an easy way to search just for their invalid files (in order to delete them and not pollute the system), would be nice to have.

kenoir:
@paperboyo 
norights

paperboyo:
Mine and yours are equally bad. Maybe it should just be a pic of Nicholas (instead of Arnold)? I trust @akemitakagi.

kahuna/public/js/components/gr-metadata-validity/gr-metadata-validity.html
```diff
@@ -1,5 +1,5 @@
 <div ng:if="! image.data.valid"
     class="image-notice image-info__group validity validity--invalid validity--invalid--point-up">
     <strong>Incomplete metadata</strong>
-    Fill in description and credit before use
+    Fill in description, credit and usable rights
```
jamesgorrie: "usable"? Usage?

kenoir: "rights that make this image usable"

I am not good at this.

NickPapacostas:
👍 🉐 

one small point, if you change the rights to "no rights" on a pic that was previously croppable, the "can't crop" error only comes up after you try cropping (as opposed to the button saying "cannot crop")... 

just fyi, dont think it matters.

kenoir:
@NickPapacostas Needed to watch changes and update deliberately.

https://github.com/guardian/grid/pull/1828/commits/b6f9c10d8b3c876367d3f2cbba42807d7cde5887

NickPapacostas:
👍

PR #1829, opened 2016-03-31
Ensure we adhere strictly to aspect ratio

kenoir:
For known aspect ratios we use the "pure" ratios rather than the
possibly off by one width/height sizes from the UI cropper.


NickPapacostas:
:+1:  :ideograph_advantage:

PR #1830, opened 2016-03-31
allows you to negate existing search terms

tsop14:
Stops the creation of a chip if the character is at the beginning of a word. Lets users turn an existing search term into a negative search.

now:
negating works

fixes: 
negatin


NickPapacostas:
:+1: :ideograph_advantage:

PR #1833, opened 2016-04-01
fixes problem with collections and quotes

tsop14:
Wraps all collections in double quotes, fixes navigating to a collection with no spaces and single quote mark i.e. apostrophes e.g. "jonny's"
quotes

Does not allow adding a collection with double quotes. Informs users. 
screen shot 2016-04-01 at 14 18 34


ShaunYearStrong:
:+1:

PR #1840, opened 2016-04-06
Fix cropper width

ShaunYearStrong:
Reverts the removal of the new crop tool and fixes the width issues.

Before:
picture 45

After:
picture 44


kenoir:
:+1:

PR #1843, opened 2016-04-07
[WIP] Refresh client if out of sync with server 

kenoir:
This will force hard refresh the client if it sees a differing app
version on the server.

My suspicion is that forcing a refresh is a bit obnoxious and this
should probably just display a message when out of date.

### DIFF
--- a/kahuna/package.json
+++ b/kahuna/package.json
@@ -4,6 +4,7 @@
       "baseURL": "public"
     },
     "dependencies": {
+      "amiuptodate": "npm:amiuptodate@^0.0.3",
       "angular": "github:angular/bower-angular@1.4.3",
       "angular-animate": "github:angular/bower-angular-animate@1.4.3",
       "angular-bootstrap": "github:angular-ui/bootstrap-bower@^0.13.0",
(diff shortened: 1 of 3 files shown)


NickPapacostas:
Really nice code, I learned a lot from it.  (angular / node packaging, more observable examples to draw on, nice functional pipelining etc)

I guess my only concern would be that it's not a common enough issue to merit the potential maintenance requirements. 

:+1: :ideograph_advantage:

akash1810:
How does this work if, for example, you're in the middle of a crop? It looks like the page would refresh without your consent?

kenoir:
@akash1810 indeed it would. But if the UI is out of sync with API it could very well fail anyway.

akash1810:
Depends on which API. _If_ auto refresh does becomes painful, could display a message on screen that you're out of date, prompting to refresh, for example.

:+1:

akash1810:
Also, rather than polling, you could use [requestIdleCallback](https://developers.google.com/web/updates/2015/08/using-requestidlecallback?hl=en)

kenoir:
This is going to stay on hold for a little while until we see another use case for it.

kenoir:
Closing until we have a reason to do this.

PR #1844, opened 2016-04-07
end date range picker now goes back up to 100 years

NickPapacostas:
Some archive people requested this

https://gyazo.com/358b42b41c6da32ab856fca324e126d3


akash1810:
:+1: if the `endDate >= startDate` validation still works

NickPapacostas:
yea pikaday handles that ^_^ good call though

PR #1848, opened 2016-04-12
Upload PNGs

Reettaphant:
Allow for uploading pngs to the grid and send back data on the png type and the png size.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/FileMetadata.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/FileMetadata.scala
@@ -5,13 +5,15 @@ import play.api.libs.functional.syntax._
 
 
 case class FileMetadata(
-  iptc: Map[String, String]      = Map(),
-  exif: Map[String, String]      = Map(),
-  exifSub: Map[String, String]   = Map(),
-  xmp: Map[String, String]       = Map(),
-  icc: Map[String, String]       = Map(),
-  getty: Map[String, String]     = Map(),
-  colourModel: Option[String]    = None
+  iptc: Map[String, String]                     = Map(),
+  exif: Map[String, String]                     = Map(),
+  exifSub: Map[String, String]                  = Map(),
+  xmp: Map[String, String]                      = Map(),
+  icc: Map[String, String]                      = Map(),
+  getty: Map[String, String]                    = Map(),
+  colourModel: Option[String]                   = None,
+  colourModelInformation: Map[String, String]   = Map()
+
 )
 
 object FileMetadata {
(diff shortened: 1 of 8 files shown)


image-loader/app/lib/imaging/FileMetadataReader.scala
```diff
@@ -78,17 +81,50 @@ object FileMetadataReader {
...
+          Dimensions(width, height)
+        }
+
+        case _ => None
+
       }
-      yield Dimensions(jpegDir.getImageWidth, jpegDir.getImageHeight)
     }
 
+  def exportPngDirectory(metadata: Metadata, mimeType: Option[String]): Map[String, String] = {
```
kenoir: I don't think you should have to hand the mimeType in here. Can you move the case statement up to the calling method?

image-loader/app/lib/imaging/FileMetadataReader.scala
```diff
@@ -21,7 +22,7 @@ object FileMetadataReader {
   private implicit val ctx: ExecutionContext =
     ExecutionContext.fromExecutor(Executors.newCachedThreadPool)
 
-  def fromIPTCHeaders(image: File): Future[FileMetadata] =
+  def fromIPTCHeaders(image: File, mimeType: Option[String]): Future[FileMetadata] =
```
kenoir: Is this data were trying to extract in the IPTC headers? If not this method is misnamed, or extracting the PNG data needs to go in another method.

common-lib/src/main/scala/com/gu/mediaservice/model/FileMetadata.scala
```diff
@@ -11,7 +11,9 @@ case class FileMetadata(
   xmp: Map[String, String]       = Map(),
   icc: Map[String, String]       = Map(),
   getty: Map[String, String]     = Map(),
-  colourModel: Option[String]    = None
+  colourModel: Option[String]    = None,
+  pngData: Map[String, String]   = Map()
```
kenoir: Does this have to be specific to PNG data? Could we split the data collected into more generically named chunks?

kenoir:
Nice work. Let's show @paperboyo in TEST before we deploy.

paperboyo:
Amazing! So the metadata engine we use deals with PNG properties! Brilliant! Some should be shared between file formats (e.g. True Colour without alpha can be both some PNGs & all RGB JPGs). It also looks like we can not only extract textual metadata from PNG-specific fields, but *also* from the embedded XMP! (https://github.com/drewnoakes/metadata-extractor/issues/62#issuecomment-74349426). Good times!

kenoir:
:+1: We should confirm this behaves in test (and doesn't disrupt the jpeg workflow).

kenoir:
@Reettaphant if you're happy this works in test, feel free to push it to prod.

kahuna/public/js/services/api/media-cropper.js
```diff
@@ -38,9 +38,11 @@ cropperApi.factory('mediaCropper', ['mediaApi', '$q', function(mediaApi) {
     function canBeCropped(image) {
         // Images can only be cropped if there is a link to the crops
         // TODO: this should be an Action
+        console.log('image ', image);
```
kenoir: Can we take these out before pushing to PROD?

PR #1850, opened 2016-04-13
Adds original file metadata to image view

kenoir:
This is useful for commercial folks whose workflow is mainly downloading
images from the grid for 3rd parties.

screenshot from 2016-04-13 12 11 06

### DIFF
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
@@ -41,6 +41,37 @@ module.controller('grImageMetadataCtrl', [
 
         // Alias for convenience in view
         ctrl.metadata = ctrl.image.data.metadata;
+
+        function formatBytes(bytes) {
+            if(bytes == 0) return '0 Byte';
+
+            const k = 1000;
+            const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
+            const i = Math.floor(Math.log(bytes) / Math.log(k));
+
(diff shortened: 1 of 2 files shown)


kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
```diff
@@ -41,6 +41,37 @@ module.controller('grImageMetadataCtrl', [
 
         // Alias for convenience in view
         ctrl.metadata = ctrl.image.data.metadata;
+
+        function formatBytes(bytes) {
+            if(bytes == 0) return '0 Byte';
+
+            const k = 1000;
+            const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
```
akash1810: :smile: are `TB`+ needed?? (or even possible)

akash1810:
Filename and dimensions are already on the page. Now that we have PNG (or soon) support, `mime type` is great!

I'd be tempted to only have this metadata in a single place, rather than duplicate it...

akash1810:
Also, having dimensions on the RHS may feel a bit odd when you select a crop as the dimensions are for the master image, rather than the crop.

kenoir:
Closing as we've already got this (albeit in an easily missed place).

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
kenoir: Future proof.

paperboyo:
> `mime type` is great

Yep! But let's do some work on reconciling that, colour models of JPGs and PNGs, info about transparency, ICC colour profile into something more human readable than `mime type: image/jpeg`. I'm eager to help. Oh, and let's serve :fries: with it ;-)

PR #1860, opened 2016-04-18
Remove pay suppliers from "free" list

kenoir:


NickPapacostas:
do we still want corbis on the "free source list"?  https://github.com/guardian/grid/blob/399032ada6bd073d813bcabaf309d3b5c38b9d74/media-api/app/lib/usagerights/Config.scala#L51

kenoir:
Looks like that list isn't consumed anywhere.

PR #1861, opened 2016-04-18
Adding leases root uri to media-api index response

NickPapacostas:
So that we can build the right routes on the client.


kenoir:
:+1:

PR #1865, opened 2016-04-20
Leases ui

NickPapacostas:
this is to make the Kahuna ui for the leases app, restricting photo use if there's a lease on it.

leases-ui


leases/app/controllers/MediaLeaseController.scala
```diff
@@ -38,13 +37,14 @@ object MediaLeaseController extends Controller
...
-    "media-leases",
-    "Media leases service",
-    Map(),
-    List(Link("by-media-id", s"$rootUri/leases/media/{id}"))
-  )
-  def index = Authenticated { _ => respond(appIndex) }
+
+  val indexResponse = {
+    val appIndex = AppIndex("media-leases", "Media leases service", Map())
+    val indexLinks =  List(Link("by-media-id", s"$rootUri/leases/media/{id}"))
```
NickPapacostas: I was thinking this really belongs in "Actions" not "Links", so I tried doing that, but in a surprise twist the "{ }" in the URL is incompatible with java.net.uri which is needed for Action

The solutions I found online seemed overly clunky and made the point that actually "{}" and not allowed in URIs >.>
Open to ideas 🉐

kenoir: Came across this before, vaguely considered refactoring so that URI's were
not required. Gave up. Calling @theefer to see if he has an opinion.
On Wed, 20 Apr 2016 at 22:43, Nicholas Papacostas <notifications@github.com>
wrote:

> In leases/app/controllers/MediaLeaseController.scala
> <https://github.com/guardian/grid/pull/1865#discussion_r60493231>:
>
> > @@ -38,13 +37,14 @@ object MediaLeaseController extends Controller
> >
> >    val notFound = respondNotFound("MediaLease not found")
> >
> > -  val appIndex = AppIndex(
> > -    "media-leases",
> > -    "Media leases service",
> > -    Map(),
> > -    List(Link("by-media-id", s"$rootUri/leases/media/{id}"))
> > -  )
> > -  def index = Authenticated { _ => respond(appIndex) }
> > +
> > +  val indexResponse = {
> > +    val appIndex = AppIndex("media-leases", "Media leases service", Map())
> > +    val indexLinks =  List(Link("by-media-id", s"$rootUri/leases/media/{id}"))
>
> I was thinking this really belongs in "Actions" not "Links", so I tried
> doing that, but in a surprise twist the "{ }" in the URL is incompatible
> with java.net.uri which is needed for Action
>
> The solutions I found online seemed overly clunky and made the point that
> actually "{}" and not allowed in URIs >.>
> Open to ideas 🉐
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/guardian/grid/pull/1865/files/e13d8a565019866b6430deb059466977aa568b2e#r60493231>
>

-- 

in England Number 908396

NickPapacostas:
TODO: 

[] notes in tooltip
[] fix starting date not including today

NickPapacostas:
leases-ui

paperboyo:
Cool! Why is there a blue bar over either red or green one if I may ask?

NickPapacostas:
@paperboyo i was hoping that would be "intuitive"! >.<

blue means it's the "currently active" lease... red and green determine if it's an "allow" or "deny" lease

paperboyo:
Ah! Ok. So the exclusive leases can overlap in time?

leases
[is it allowed or denied on 2016-05-04? why only one is "active"? unless "active" means "edited now" and not "in operation"]

This is the date format Grid uses throughout the UI:

date

Maybe leases should use it too?

Also, I hope they are displayed in chronological order. Also: what about infinity? (before the big bang, after the armageddon)

In general, the wording is quite verbose (like what I write :D). I wonder if some kind of graphical representation (timeline? 😱)  would be more immediately obvious (but best to ignore this comment, I think).

kenoir:
I think at this point it's important to remember that multiple leases are unlikely (and more than 2 ever is very unlikely).

A timeline would be a good representation but way overkill here right now.

+1 to normalising dates & making it clearer which lease is current.

paperboyo:
> A timeline would be a good representation but way overkill here right now.

I agree. Disagreeing with myself at the same time.

NickPapacostas:
@paperboyo agreed on all points! I'll fix the date time format. 

Can you think of a less verbose way to show the dates? remember they're both optional

paperboyo:
> Can you think of a less verbose way to show the dates?

I'm much better at criticising than coming up with ideas...

Maybe:
Allowed: 7 Apr 2016 to 9 Apr 2016

Or:
Allowed: 7 Apr 2016 – 9 Apr 2016 [en dash]

But not really sure it's any better (could be understood quicker which would be the point here...).

NickPapacostas:
I was thinking something similar, but if there's only 1 date (a before OR an after) then it will be a bit confusing ^_^ 

maybe in those cases we just say "before" and "after"?

paperboyo:
I should think so.

leases/app/controllers/MediaLeaseController.scala
theefer: Hey guys :smile: 

It could work as a Link, but making it an Action might be nicer. I see the problem with URI, technically the field should probably allow URI templates (which is a superset of URIs), so might have to be changed back to String indeed. Note that the Actions are inspired by the Swagger spec, which [allows](http://swagger.io/specification/#pathsObject) URI templates in a similar context.

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
```diff
@@ -135,6 +135,10 @@ module.controller('grImageMetadataCtrl', [
                 .then(() => ctrl.removingCollection = false);
         };
 
+        ctrl.displayLeases = () => {
+            return (ctrl.userCanEdit) || (!ctrl.userCanEdit && ctrl.image.leases > 0);
```
Reettaphant: !ctrl.userCanEdit in the second boolean is redundant

Reettaphant:
Could the date picker disappear when its not in focus?

common-lib/src/main/scala/com/gu/mediaservice/model/leases/MediaLease.scala
```diff
@@ -8,30 +8,35 @@ import org.joda.time.DateTime
...
-  id: Option[String],
-  leasedBy: Option[String],
-  startDate: Option[DateTime] = None,
-  endDate: Option[DateTime] = None,
-  access: MediaLeaseType = AllowUseLease,
-  notes: Option[String],
-  mediaId: String,
-  createdAt: DateTime = new DateTime()
-) {
+                       id: Option[String],
```
akash1810: space, space, space 👾

kahuna/public/js/components/gr-crop-image/gr-crop-image.js
```diff
@@ -2,24 +2,30 @@ import angular from 'angular';
...
+                        ctrl.canBeCropped = allowed;
+                    });
+                }
             });
         }
 
         $scope.$watch(() => ctrl.image.data.metadata, onValChange(() => updateState()));
         $scope.$watch(() => ctrl.image.data.usageRights, onValChange(() => updateState()));
-
+        $rootScope.$on('leases-updated', () => updateState());
```
akash1810: any reason this is on the rootScope?

kahuna/public/js/components/gr-tooltip/gr-tooltip.js
```diff
@@ -3,23 +3,26 @@ import 'titip';
...
-            const position = attrs.grTooltipPosition || 'bottom';
+tooltip.directive('grTooltip', [
+    'onValChange',
+    '$interpolate',
+    function (onValChange, $interpolate) {
+        return {
+            restrict: 'A',
+            link: function ($scope, element, attrs) {
+                const position = attrs.grTooltipPosition || 'bottom';
+                const toolTipText = $interpolate(attrs.grTooltip);
```
akash1810: Why are we using interpolate?

Not saying its wrong, I've just never used it/don't really know what it does 😄

kahuna/public/js/components/gr-crop-image/gr-crop-image.js
akash1810: add `onValChange` too?

kahuna/public/js/components/gu-date-range/gu-date-range.js
```diff
@@ -69,30 +70,32 @@ guDateRange.directive('guDateRange', [function () {
...
             ctrl.guFirstDay = ctrl.guFirstDay || 0;
-
             var startInput = el.find('.gu-date-range__input__start--hidden')[0];
             var startContainer = el.find('.gu-date-range__overlay__pikaday--start')[0];
 
             var endInput = el.find('.gu-date-range__input__end--hidden')[0];
             var endContainer = el.find('.gu-date-range__overlay__pikaday--end')[0];
 
             var iso8601Format = 'YYYY-MM-DDTHH:mm:ssZ';
+            var tenYearsInSeconds = (10 * 365 * 24 * 60 * 60 * 1000);
```
akash1810: `const`?

akash1810: `* 1000` = milliseconds rather than seconds - I'd rename variable for clarity

kahuna/public/js/components/gu-date-range/gu-date-range.js
```diff
@@ -105,14 +108,11 @@ guDateRange.directive('guDateRange', [function () {
                 pikaEnd.show();
             });
 
-            $scope.$watch('pikaEndValue', function (pikaEndValue) {
-                var date = pikaEndValue && new Date(pikaEndValue);
-                pikaStart.setMaxDate(date || new Date());
```
akash1810: Does this mean you're able to select a startDate that's > endDate now?

kahuna/public/js/leases/leases.js
```diff
@@ -0,0 +1,142 @@
+import angular from 'angular';
+import template from './leases.html!text';
+import '../util/rx';
+
+import '../services/api/leases';
+import './leases.css!';
+import '../components/gr-confirm-delete/gr-confirm-delete.js';
+
+
+export var leases = angular.module('kahuna.edits.leases', [
```
akash1810: `const`?

kahuna/public/js/leases/leases.js
```diff
@@ -0,0 +1,142 @@
...
+
+        ctrl.cancel = () => ctrl.editing = false;
+
+        ctrl.save = () => {
+            if (ctrl.access === null) {
+                $window.alert('Please select an access type (Allow or Deny)');
+            } else {
+                ctrl.adding = true;
+                ctrl.newLease.mediaId = ctrl.image.data.id;
+                ctrl.newLease.createdAt = new Date();
```
akash1810: could we set this server side?

kahuna/public/js/services/api/media-cropper.js
```diff
@@ -5,63 +5,67 @@ export var cropperApi = angular.module('kahuna.services.api.cropper', [
...
-    function canBeCropped(image) {
-        // Images can only be cropped if there is a link to the crops
-        // TODO: this should be an Action
-        return image.getLink('crops').
-        then(() => true, () => false);
-    }
+        function canBeCropped(image) {
+            // Images can only be cropped if there is a link to the crops
+            // TODO: this should be an Action
+            if (image.data.source.mimeType === 'image/png') {
```
akash1810: this diff seems out of place in this PR...

kahuna/public/js/components/gr-tooltip/gr-tooltip.js
NickPapacostas: It tells angular to parse the tooltip. This was so I could put scope variables in the tooltip, like I wanted to put the lease creator or notes etc. and it was printing "{{lease.notes}}" or w.e.

kahuna/public/js/services/api/media-cropper.js
NickPapacostas: >.<

kahuna/public/js/leases/leases.js
NickPapacostas: const.

kahuna/public/js/components/gu-date-range/gu-date-range.js
NickPapacostas: No this says that you can set a start date that is after today. When searching that concept didn't really make sense. 

Do you think it's a problem to have this spill over into the search functionality? I tried splitting that behavior out but found it was a little heavy for something that's effectively harmless

kahuna/public/js/components/gu-date-range/gu-date-range.js
NickPapacostas: yup and yup

kahuna/public/js/leases/leases.js
NickPapacostas: I actually think it ignores this and does set it server side, good catch

kahuna/public/js/components/gr-tooltip/gr-tooltip.js
akash1810: Would the `grTooltipUpdates` attribute help here? For example [this usage](https://github.com/guardian/grid/blob/master/kahuna/public/js/search/results.html#L24-L31).

kahuna/public/js/components/gu-date-range/gu-date-range.js
akash1810: > Do you think it's a problem to have this spill over into the search functionality?

It feels a little odd to (potentially) search for uploads in the future, but no real strong opinion from me here, especially as it doesn't break anything. I'd release and get feedback.

NickPapacostas:
@akash1810 @ShaunYearStrong 

Mind taking one more pass? I'd like to send out comms today and merge tomorrow

kahuna/public/js/leases/leases.css
```diff
@@ -0,0 +1,106 @@
...
+.lease:hover {
+  color:white;
+}
+
+gr-leases .metadata-line__key {
+  margin-bottom: 10px
+}
+
+gr-leases gr-confirm-delete {
+  float: right;
```
akash1810: super minor: flexbox?

kahuna/public/js/leases/leases.html
```diff
@@ -0,0 +1,83 @@
...
+          gr:tooltip-position="left">
+
+      <gr-icon ng-show="!ctrl.editing" ng:class="{'spin': ctrl.adding}">add_box</gr-icon>
+      <span>
+         <span ng:show="ctrl.adding">Adding...</span>
+      </span>
+  </button>
+</dt>
+
+<form class="lease__form"
```
akash1810: Would `lease-form` be more BEM-y

kahuna/public/js/leases/leases.js
```diff
@@ -0,0 +1,149 @@
...
+                startDate: new Date(Date.now() - oneDayInMilliSeconds),
+                endDate: new Date(Date.now() + oneDayInMilliSeconds),
+                access: null
+            };
+            ctrl.access = null;
+        };
+
+
+        ctrl.leaseStatus = (lease) => {
+            const active = lease.active ? 'active ' : ' ';
```
akash1810: can this be a bool?

kahuna/public/js/leases/leases.js
```diff
@@ -0,0 +1,149 @@
...
+
+            let current = '';
+            if (ctrl.leases.current) {
+                current = ctrl.leases.current.data.id == lease.id ? 'current ' : '';
```
akash1810: similar as above - can this be a bool rather than string?

kahuna/public/js/services/api/leases.js
```diff
@@ -0,0 +1,70 @@
...
+  ]);
+
+leaseService.factory('leaseService', [
+  '$rootScope',
+  '$q',
+  'imageAccessor',
+  'mediaApi',
+  'editsService',
+  function ($rootScope, $q, imageAccessor, mediaApi, editsService) {
+    var leasesRoot;
```
akash1810: bit odd to have this here, though I may not be understanding what's going on...

kahuna/public/js/leases/leases.js
NickPapacostas: OK I use the string in an ng class. I guess it's better to do that logic in the template?

``` ng:class="active: lease.active" ``` ? and then put a helper method for "current" in the controller?

File: kahuna/public/js/services/api/leases.js (hunk 15 above)
Comment (NickPapacostas) [inline kahuna/public/js/services/api/leases.js]: yea not the prettiest... took the pattern from here: https://github.com/guardian/grid/blob/master/kahuna/public/js/services/api/media-cropper.js

open to improvements though

File: kahuna/public/js/leases/leases.html (hunk 12 above)
Comment (NickPapacostas) [inline kahuna/public/js/leases/leases.html]: I will make that change while reading 

https://css-tricks.com/bem-101/

File: kahuna/public/js/leases/leases.css [hunk 16]
```diff
@@ -0,0 +1,106 @@
...
+  background-color: lightgreen;
+}
+
+.active.denied {
+  background-color: red;
+}
+
+gr-leases .gu-date-range__overlay {
+  position: absolute;
+  z-index: 1;
```
ShaunYearStrong: Good man

kahuna/public/js/leases/leases.css
```diff
@@ -0,0 +1,106 @@
...
+  position: relative;
+}
+
+.add-lease__header {
+  display: block;
+}
+
+.lease {
+  margin: 4px;
+  padding: 3px;
```
ShaunYearStrong: A bit magic. But if that's the way it's done throughout, keep going.

kahuna/public/js/leases/leases.css
```diff
@@ -0,0 +1,106 @@
...
+  border-top-left-radius: 2px;
+  border-bottom-left-radius: 2px;
+}
+
+.lease:hover {
+  color:white;
```
ShaunYearStrong: Maybe use hex colours for consistency?

kahuna/public/js/leases/leases.css
```diff
@@ -0,0 +1,106 @@
...
+  padding: 3px;
+  margin-bottom: 7px;
+}
+
+.lease__current {
+  padding: 2px;
+
+}
+
+.current {
```
ShaunYearStrong: Is this a modifier for an existing block?/ Should be `block--current` if so to group it.

kahuna/public/js/leases/leases.js
akash1810: > I guess it's better to do that logic in the template?

Either or. I think its a naming thing - `active`, to me, just suggests boolean; `active = 'active'` seems odd. Maybe `status = 'active'`?

kahuna/public/js/components/gr-crop-image/gr-crop-image.js
```diff
@@ -2,24 +2,35 @@ import angular from 'angular';
...
 export const cropImage = angular.module('gr.cropImage', [
 
 ]);
 
 cropImage.controller('grCropImageCtrl', [
-    '$scope', 'mediaCropper', 'onValChange',
-    function ($scope, mediaCropper, onValChange) {
+    '$scope', 'mediaCropper', 'onValChange', 'leaseService', '$rootScope',
+    function ($scope, mediaCropper, onValChange, leaseService,  $rootScope) {
         let ctrl = this;
```
jamesgorrie: `const`

kahuna/public/js/components/gr-crop-image/gr-crop-image.js
```diff
@@ -2,24 +2,35 @@ import angular from 'angular';
...
 
         function updateState () {
-            mediaCropper.canBeCropped(ctrl.image).then(croppable => {
-                ctrl.canBeCropped = croppable;
+            leaseService.allowedByLease(ctrl.image).then(allowed => {
```
jamesgorrie: You might want to prefix with a verb e.g. `isAllowedByLease` to indicate that you're asking.

kahuna/public/js/components/gr-crop-image/gr-crop-image.js
```diff
@@ -2,24 +2,35 @@ import angular from 'angular';
...
+                    });
+                } else {
+                    ctrl.canBeCropped = false;
+                }
             });
         }
 
-        $scope.$watch(() => ctrl.image.data.metadata, onValChange(() => updateState()));
-        $scope.$watch(() => ctrl.image.data.usageRights, onValChange(() => updateState()));
+        $rootScope.$on('leases-updated', () => {
```
jamesgorrie: Haskellify 1 line!
```
 $rootScope.$on('leases-updated', updateState);
```

kahuna/public/js/leases/leases.html
```diff
@@ -0,0 +1,83 @@
...
+  </gu-date-range>
+
+  <input type="text" ng:model="ctrl.newLease.notes" placeholder="Notes..."/>
+
+  <div class="lease__form__buttons">
+    <button class="lease__form__buttons__button-cancel button-cancel" type="button" ng:click="ctrl.cancel()" title="Close">
+        <gr-icon-label gr-icon="close"><span ng:hide="ctrl.grSmall">Cancel</span></gr-icon-label>
+    </button>
+    <button
+        class="lease__form__buttons__button-save button-save"
```
ShaunYearStrong: I think the class here is getting a bit specific. Are they styles here specifically linked to being part of leases or can this become its own block?

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
```diff
@@ -135,6 +135,10 @@ module.controller('grImageMetadataCtrl', [
                 .then(() => ctrl.removingCollection = false);
         };
 
+        ctrl.displayLeases = () => {
```
jamesgorrie: `ctrl.displayLeases = () => ctrl.userCanEdit || ctrl.image.leases > 0);`?

kahuna/public/js/leases/leases.html
ShaunYearStrong: http://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/ Is a good one.

kahuna/public/js/components/gu-date-range/gu-date-range.js
```diff
@@ -69,30 +70,32 @@ guDateRange.directive('guDateRange', [function () {
...
-
             var startInput = el.find('.gu-date-range__input__start--hidden')[0];
             var startContainer = el.find('.gu-date-range__overlay__pikaday--start')[0];
 
             var endInput = el.find('.gu-date-range__input__end--hidden')[0];
             var endContainer = el.find('.gu-date-range__overlay__pikaday--end')[0];
 
             var iso8601Format = 'YYYY-MM-DDTHH:mm:ssZ';
+            const tenYearsInMilliseconds = (10 * 365 * 24 * 60 * 60 * 1000);
+            const tenYearsFromNow =  new Date(Date.now() + tenYearsInMilliseconds);
```
jamesgorrie: Do we not use [moment](http://momentjs.com) anywhere? If not we should, it's great.
[Looks like we install it](https://github.com/guardian/grid/blob/master/kahuna/package.json#L25).

kahuna/public/js/leases/leases.html
NickPapacostas: hahah yea... i think that's the product of some copy/pasting... very diplomatically phrased of you though... "a bit specific" >.<

kahuna/public/js/leases/leases.js
```diff
@@ -0,0 +1,149 @@
...
+    'gr.confirmDelete',
+    'util.rx'
+]);
+
+
+leases.controller(
+    'LeasesCtrl',
+    ['$window', '$q', '$scope', 'inject$', 'leaseService', '$rootScope',
+    function($window, $q, $scope, inject$, leaseService, $rootScope) {
+        let ctrl = this;
```
jamesgorrie: `const`?
You can still modify properties on the pbject.

kahuna/public/js/leases/leases.js
jamesgorrie: > do that logic in the template

Definitely
`<div ng:class="{ 'thing--active': ctrl.active }"></div>`

kahuna/public/js/services/api/leases.js
jamesgorrie: memoisation.

leases/app/controllers/MediaLeaseController.scala
```diff
@@ -23,8 +23,7 @@ import lib.{LeaseStore, ControllerHelper}
 case class AppIndex(
                      name: String,
                      description: String,
-                     config: Map[String, String] = Map(),
-                     links: List[Link] = Nil)
+                     config: Map[String, String] = Map())
```
jamesgorrie: Is this used?

jamesgorrie:
If I'm allowed I'd say 👍 to get people using it.
A couple of tidy ups, but those can go in as the it's updated.

NickPapacostas:
Thanks everyone, great feedback! I'll make those changes but consider it merge-able barring some major change.

kahuna/public/js/components/gu-date-range/gu-date-range.js
ShaunYearStrong: 👍  For moment. Used it before and it saves a lot of DateTime based headaches.

PR #1866, opened 2016-04-21
Allows cropping PNGs with type TrueColor (converts to JPEG)

Reettaphant:
Allow for cropping pngs with type TrueColor. Produce crops as jpegs.


kenoir:
:+1: Looks good. 

Can we let @paperboyo try it out in test first?

PR #1867, opened 2016-04-21
Add Staff Illustrator option

kenoir:
In order that rights & restrictions can be correctly chosen by users
this changeset adds a "Staff Illustrator" option to `metadata-editor`.

foop

Requires deployment of:

- `media-api`
- `metadata-editor`

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
@@ -121,6 +121,11 @@ object MetadataConfig {
     "Suki Dhanda"         -> "The Observer"
   )
 
+  val staffIllustrators = List(
+    "Mona Chalabi",
+    "Jan Diehm"
+  )
+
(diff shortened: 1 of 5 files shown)


ShaunYearStrong:
👍  Looks simple enough to me

paperboyo:
So much for simplification :stuck_out_tongue_winking_eye:

The choice with buttons may be easier, but it's inconsistent with the rest. It should be a dropdown (with currently, quite lonely two people, but maybe later - more). Adding more later won't create unnecessary user confusion then when the switch for dropdown would be necessary.

kenoir:
@paperboyo it's just what the control does with fewer options. It _automatically_ switches type based on the list size :boom:

paperboyo:
Clippy-like intelligence, then :smiling_imp:

kenoir:
@paperboyo :paperclip: "It looks like you're trying to imply that you don't like this change. Would you like some help with that?"

paperboyo:
Who? Me?! Never! I love all the thingz! :rabbit:

PR #1868, opened 2016-04-21
Adds X/Y input boxes on crop screen

kenoir:
In order that users don't have to pixel hunt when trying to produce exact crops.

screenshot from 2016-04-21 16 42 21


NickPapacostas:
👍 🉐

PR #1869, opened 2016-04-22
Provide information on Permissions from Auth

kenoir:
In order that we can provide this information to Central Production.

Is used by https://github.com/guardian/grid-permissions


NickPapacostas:
👍

PR #1870, opened 2016-04-26
Use Ubuntu Wily, systemd and GraphicsMagick 1.3.21

kenoir:
Uses a new Ubuntu `Wily` AMI and consequently uses `systemd` and `GraphicsMagick 1.3.21`.

Removes `logrotate` config as we are now logging to `syslog` which is already rotated.

Depends:
 - https://github.com/guardian/grid-infra/pull/236


packer/media-service/provisioning.json
```diff
@@ -1,6 +1,6 @@
 {
     "variables": {
-        "machine_images_ami": "ami-6857e51b"
+        "machine_images_ami": "ami-161b9765"
```
kenoir: Bumps us to `wily`

packer/media-service/provisioning.json
```diff
@@ -20,11 +20,6 @@
 
     "provisioners": [
     {
-        "type": "file",
```
kenoir: Logs now go to syslog, so already taken care of by logrotate.

packer/media-service/media-service-ami.sh
```diff
@@ -8,10 +8,6 @@ function new_section {
 
 set -e
 
-new_section "Moving config files into place"
-mv /tmp/media-service_logrotate /etc/logrotate.d/media-service
-sudo chown root:root /etc/logrotate.d/media-service
```
NickPapacostas: if it goes to syslog now, are we using the syslog rotator or logrotate (per your comment below)?
I know very little about this (i.e. the blog posts i just read), so maybe we just talk about it IRL

packer/media-service/provisioning.json
NickPapacostas: just confirming, this defines the new base ami to build the packer-generated local ami right? "**Am i right**?"

NickPapacostas:
👍  lez give it a shot ^_^

akash1810:
Any reason why this only affects cropper?

It feels like this change would put the packer provisioning.json in an odd, cropper specific, state?

We're also not explicitly installing a specific version of GM as part of the packer build... I guess 1.3.21 is latest?

kenoir:
This change only affects cropper as making a blanket change to all apps
would be very disruptive. Others will follow. 1.3.21 is latest for Wily.
Specifying a version sounds like a good idea.
On Tue, 26 Apr 2016 at 22:55, Akash A <notifications@github.com> wrote:

> Any reason why this only affects cropper?
>
> It feels like this change would put the packer provisioning.json in an
> odd, cropper specific, state?
>
> We're also not explicitly installing a specific version of GM as part of
> the packer build... I guess 1.3.21 is latest?
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/guardian/grid/pull/1870#issuecomment-214899866>
>

-- 

in England Number 908396

packer/media-service/media-service-ami.sh
kenoir: logrotate already has config to rotate syslog logs!

packer/media-service/provisioning.json
kenoir: Yup.

PR #1875, opened 2016-04-27
Gallery view

ShaunYearStrong:
gallery-load


kahuna/public/js/components/gu-lazy-gallery-shortcuts/gu-lazy-gallery-shortcuts.js
```diff
@@ -0,0 +1,40 @@
...
+    return {
+        restrict: 'EA',
+        require: '^guLazyGallery',
+        link: function (scope, element, attrs, lazyGalleryCtrl) {
+            console.log(lazyGalleryCtrl);
+            function invoke(fnName) {
+                return (event) => {
+                    // Must cancel any scrolling caused by the key
+                    event.preventDefault();
+                    lazyGalleryCtrl[fnName]();
```
ShaunYearStrong: This doesn't work, any help would be great.

kenoir: Do you mean that it doesn't prevent the default scrolling event?

kenoir:
@ShaunYearStrong happy to help with Observable stuff - what were you thinking needed improving specifically.

kahuna/public/js/components/gu-lazy-gallery-shortcuts/gu-lazy-gallery-shortcuts.js
ShaunYearStrong: I meant that it didn't call the function. I've since worked out it's because the names of the functions were wrong and Im a moron.

ShaunYearStrong:
@kenoir Everything? haha. Nah, I guess I'd just like things to be a bit more reactive. Also I'd like help with the lazy loading logic in terms of reactive programming. Definitely haven't quite got my head around it yet.

kahuna/public/js/components/gu-lazy-gallery-shortcuts/gu-lazy-gallery-shortcuts.js
```diff
@@ -27,13 +27,13 @@ lazyGalleryShortcuts.directive('guLazyGalleryShortcuts',
                     combo: 'left',
                     description: 'Go to the previous image',
                     allowIn: ['INPUT'],
-                    callback: invoke('previousImage')
+                    callback: invoke('previousItem')
```
paperboyo: `Item` not `Image` - very prudent: in preparation of Grid support for video ;-) Absolutely innocent ping: @akash1810.

paperboyo:
I’ve got so many thoughts on that functionality, maybe I’ll write an email. But then - it would be too long...

Even if *nothing* goes according to my sick plan :smiling_imp: - it’s already a :gem:!

kahuna/public/config.js
```diff
@@ -95,6 +95,10 @@ System.config({
     "github:jspm/nodelibs-util@0.1.0": {
       "util": "npm:util@0.10.3"
     },
+    "github:tapmodo/Jcrop@0.9.12": {
```
ShaunYearStrong: Not sure how this got in here...

kahuna/public/stylesheets/main.css
```diff
@@ -1087,6 +1093,162 @@ textarea.ng-invalid {
 }
 
 /* ==========================================================================
+   Gallery
```
akash1810: can we extract these styles to their own file?

kahuna/public/js/components/gu-lazy-gallery/gu-lazy-gallery.js
```diff
@@ -0,0 +1,144 @@
...
+
+lazyGallery.controller('GuLazyGalleryCtrl', ['$scope', 'subscribe$', function($scope, subscribe$) {
+    let ctrl = this;
+    ctrl.canGoNext = false;
+    ctrl.canGoPrev = false;
+
+    ctrl.init = function({items$, preloadedItems$, currentIndex$}) {
+        const itemsCount$ = items$.map(items => items.length).distinctUntilChanged();
+
+        const buttonCommands$ = new Rx.BehaviorSubject('galleryStart');
```
NickPapacostas: I think this is fine, but I'm wondering: if you did it the other way (where you assigned all the handlers within the creation block) could you avoid using BehaviorSubject? 

I'm asking because I was just looking at the docs and behavior subject seems to be for "[a value that changes over time](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/subjects/behaviorsubject.md)" which I don't think really applies to the buttonpress stream. 

just something to be aware of

ShaunYearStrong: Hmm true, I'll take a look. When I initialise in the creation block I can add something straight to the stream.

paperboyo:
Thinking hard how to get rid of the buttons at the bottom, so that the image can be even bigger (than just by removal of caption area thanks to info panel hook-up). Not related would be getting rid of home/end buttons. I would say it's not likely that they will be used a lot (but it's **great** that home/end keys work!).

Maybe the remaining two previous/next buttons could be moved into (already too crowded, though) bar at top... Hmm. The other solution would be to have them at the sides of the image - but I'm against because it's a long way for a mouse from one to the other then.

There is also a radical option - no visual buttons at all. But with that I would wait for proper fullscreen/everything hidden.

Just thinking out loud...

NickPapacostas:
👍  looks good. Agree with @paperboyo but all in due time. 

Nice job ^_^

NickPapacostas:
screen shot 2016-05-26 at 17 26 46


just noticed this if you scroll down

ShaunYearStrong:
@NickPapacostas Good spot! This is why we test

akash1810:
couple things I noticed:

- clicking `last` scrolls through all images
- when new images come in, should you be able to scroll left to see them? (not in the scope of this PR)
- its amazing 🎉

ShaunYearStrong:
@paperboyo @akash1810 @NickPapacostas - I've fixed the hover area and removed the first/last image logic for now, with a view to moving to a sort of circular gallery i.e. Pressing "Next" on the last image would take you to the first and vice-versa. I think this should be an improvement though, rather than holding up this going out.

kahuna/public/js/preview/image.js
```diff
@@ -87,6 +90,41 @@ image.directive('uiPreviewImage', function() {
...
+            bindToController: true,
+            link: function(scope, element, attrs, ctrl) {
+                ctrl.loading = false;
+                const image$ = new Rx.Subject();
+
+                const optimisedImage$ = image$.flatMap((image) => {
+                    return Rx.Observable.fromPromise(imgops.getFullScreenUri(image));
+                });
+
+                scope.$watch(() => ctrl.image.data.id, () => {
```
NickPapacostas: for my own knowledge: is this faster than checking the object?

NickPapacostas:
👍 🉐

kahuna/public/js/preview/image.js
ShaunYearStrong: It's because this means we're only checking if the actual image has changed, as opposed to the whole object, which also stores things like labels and the like. We don't want to update the visible image when these are added.

kahuna/public/js/components/gu-lazy-gallery/gu-lazy-gallery.html
```diff
@@ -0,0 +1,60 @@
...
+                   class="gallery-item__select__checkbox"
+                   id="id-{{::galleryCtrl.item.data.id}}-gallery-item__select__checkbox--deselect"
+                   ng:click="ctrl.select(galleryCtrl.item)" />
+            <label for="id-{{::galleryCtrl.item.data.id}}-gallery-item__select__checkbox--deselect"
+                   class="gallery-item__select__checkbox__label">
+                <gr-icon>check_box</gr-icon>
+            </label>
+        </div>
+
+        <div class="gallery-item__preview"
```
akash1810: not a blocker - would be good to confirm this is needed in this view

kahuna/public/js/components/gu-lazy-gallery/gu-lazy-gallery.html
```diff
@@ -0,0 +1,60 @@
...
+        </div>
+
+    </div>
+
+    <div class="gallery__controls">
+
+        <button class="gallery__control gallery__control--left"
+                ng:click="galleryCtrl.prevItem()">
+            <gr-icon>navigate_before</gr-icon>
+            <span class="visibly-hidden">Previous</span>
```
akash1810: is the `span` needed if we're just hiding it anyway?

kahuna/public/js/components/gu-lazy-gallery/gu-lazy-gallery.html
akash1810: maybe add some tracking?

kahuna/public/js/components/gu-lazy-gallery/gu-lazy-gallery.html
ShaunYearStrong: For accesibility reasons, yes. If a screen reader user were to tab to this arrow, they'd have no idea what it did. I could always change it to a `title` attribute instead, but this is also pretty standard.

kahuna/public/js/components/gu-lazy-gallery/gu-lazy-gallery.html
ShaunYearStrong: Will do! The additional div was needed when things were transcluded I'll see if it is now

NickPapacostas: Not sure if you meant Mixpanel tracking @akash1810 but we no longer have a valid MP account, so don't do that @ShaunYearStrong  ^_^

NickPapacostas:
commented on an outdated diff >.<  just saw the tracking point. i think this should be removed b.c our mp account is invalid at the moment @ShaunYearStrong 

Sorry for the delay on this.

kahuna/public/js/components/gu-lazy-gallery/gu-lazy-gallery.html
akash1810: Shoot forgot about that! Maybe we should move to GA at some point?

kenoir:
:+1: :+1: :+1:

PR #1876, opened 2016-04-27
hasAlpha property to colourModelInformation

Reettaphant:
- Adds an hasAlpha property to colourModelInformation to allow for more accurate detection of transparent pngs.
- This property is used to determine whether image crops should be jpegs.

### DIFF
--- a/image-loader/app/lib/imaging/FileMetadataReader.scala
+++ b/image-loader/app/lib/imaging/FileMetadataReader.scala
@@ -133,21 +133,21 @@ object FileMetadataReader {
 
     val formatter = format(source)("%A")
 
-    for {
-      output <- runIdentifyCmd(formatter)
-    } yield {
+    runIdentifyCmd(formatter).map{ hasAlpha => getColourInformation(metadata, hasAlpha.headOption) }
+      .recover { case _ => getColourInformation(metadata, None) }
+  }
 
-      val pngDir = metadata.getFirstDirectoryOfType(classOf[PngDirectory])
+  private def getColourInformation(metadata: Metadata, hasAlpha: Option[String]): Map[String, String] = {
+    val pngDir = metadata.getFirstDirectoryOfType(classOf[PngDirectory])
 
-      Map(
-        "hasAlpha" -> output.headOption,
-        "colorType" -> Option(pngDir.getDescription(PngDirectory.TAG_COLOR_TYPE)),
-        "bitsPerSample" -> Option(pngDir.getDescription(PngDirectory.TAG_BITS_PER_SAMPLE)),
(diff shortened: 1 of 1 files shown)


image-loader/app/lib/imaging/FileMetadataReader.scala
```diff
@@ -121,17 +123,28 @@ object FileMetadataReader {
...
 
-  def getColorModelInformation(metadata: Metadata): Map[String, String] = {
+  def getColorModelInformation(image: File, metadata: Metadata): Future[Map[String, String]] = {
+
+    val source = addImage(image)
+
+    val formatter = format(source)("%A")
+
+    for {
+      output <- runIdentifyCmd(formatter)
```
kenoir: Doing this makes extracting all the non `hasAlpha` attributes dependent on `runIdentifyCmd` succeeding. It might be better to allow it fail gracefully and still return valuable information.

kenoir:
:+1: apart from comments.

kenoir:
@Reettaphant can we merge this still?

Reettaphant:
@kenoir Not until I can update the AMI unfortunately.

image-loader/app/lib/imaging/FileMetadataReader.scala
```diff
@@ -125,17 +127,28 @@ object FileMetadataReader {
       }
     }
 
-  def getColorModelInformation(metadata: Metadata): Map[String, String] = {
+  def getColorModelInformation(image: File, metadata: Metadata): Future[Map[String, String]] = {
 
-      val pngDir = metadata.getFirstDirectoryOfType(classOf[PngDirectory])
+    val source = addImage(image)
+
+    val formatter = format(source)("%A")
```
paperboyo: Turns out this just detects if the transparency is _supported_, not if it there, actually. Wholly opaque images like 1e364185d42b0663fe77c9b9b0128124ea712188 or 5c6cc57cfaf8392a631d5541b133d9774a441ba1 still slip through :(

If we are using ImageMagick, we should do [this](https://www.imagemagick.org/discourse-server/viewtopic.php?t=18596#p71706). If GraphicsMagick - [that](http://stackoverflow.com/a/39912666).

Reettaphant: Oh dear. We are using GraphicsMagick unfortunately, the solution seems a bit ugly but we should fix it of course.

paperboyo: Another possibly relevant bit https://github.com/guardian/grid/blob/master/cropper/app/lib/Crops.scala#L101

paperboyo: We may also want to try
`gm -identify -format %r`
whenever it contains `Matte`, there is transparency. Tests needed, though.

PR #1879, opened 2016-04-29
Removes some JCrop related code

kenoir:
As we don't use JCrop anymore


NickPapacostas:
👍  I tried to pull this branch down and remove jcrop with jspm, but i didn't see any changes to commit when i did so.

kenoir:
We removed it from the package.json a while  ago. Just came across this in Kahuna and thought it probably should go.

NickPapacostas:
ok sounds good. i must just need to clean up my local packages

PR #1880, opened 2016-04-29
Kahuna: Use systemd

kenoir:
Depends:
- https://github.com/guardian/grid-infra/pull/240


kahuna/conf/kahuna.conf
```diff
@@ -1,28 +0,0 @@
...
-
-chdir /home/media-service
-
-# automatically restart if the process dies
-# respawn
-
-script
-  CMD="$USER_HOME/bin/$APP"
-  echo "$CMD" > $USER_HOME/logs/cmd.txt
-  $CMD > $LOGFILE 2>&1
```
NickPapacostas: i don't understand what this script at the bottom was meant to do.... basically log that the app was run?

NickPapacostas: which i'm guessing systemd does in it's own log

kenoir: It logs the command that was used to run the app and directs stderr and stdout to the logfile, in the new systemd version this just goes to syslog.

PR #1882, opened 2016-05-04
remove debug statements from DOM

NickPapacostas:
read this article yesterday: 

https://medium.com/@hackupstate/improving-angular-performance-with-1-line-of-code-a1fb814a6476#.qivz5vemt

tried to do some profiling to get some performance metrics, I saw less CPU utilisation but in extremely laboratory conditions ^_^

Still, good to do it regardless

### DIFF
--- a/kahuna/public/js/main.js
+++ b/kahuna/public/js/main.js
@@ -111,6 +111,11 @@ kahuna.config(['$urlRouterProvider',
     $urlRouterProvider.otherwise('/search');
 }]);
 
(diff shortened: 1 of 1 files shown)


ShaunYearStrong:
Can't see it hurting 👍  Just remember in the future if we ever need it for some reason.

akash1810:
would this have any impact on sentry and its stack traces?

akash1810:
would be good to implement this in the other tools too

PR #1883, opened 2016-05-05
Expose crop URLs

Reettaphant:
Expose all image crops as links to the user.

### DIFF
--- a/kahuna/public/js/components/gr-display-crops/gr-display-crops.css
+++ b/kahuna/public/js/components/gr-display-crops/gr-display-crops.css
@@ -1,0 +1,14 @@
+.gr-display-crops {
+    display: flex;
+    justify-content: space-between;
+}
+
+.gr-display-crops__crop {
+    padding-bottom: 5px;
+}
+
+.gr-display-crops__crop__list {
+    flex-direction: column;
+    justify-content: flex-start;
+}
+
--- a/kahuna/public/js/components/gr-display-crops/gr-display-crops.html
+++ b/kahuna/public/js/components/gr-display-crops/gr-display-crops.html
@@ -1,0 +1,37 @@
+<div class="image-info">
+    <dl class="image-info__group image-info__wrap" ng:if="ctrl.crops.length > 0">
+        <div class="gr-display-crops">
+            <dt class="image-info__heading image-info__wrap">Crops</dt>
+            <button
+                ng-if="!ctrl.showCrops"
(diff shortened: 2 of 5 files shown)


kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
```diff
@@ -279,3 +279,40 @@
         </dd>
     </dl>
 </div>
+<div class="image-info">
```
kenoir: Is it possible/reasonable to componentise this control, so it can be shared with the search view (only handling single image case for now)?

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
```diff
@@ -279,3 +279,40 @@
...
+    <dl class="image-info__group image-info__wrap" ng:if="ctrl.image.data.source.mimeType === 'image/png' && ctrl.image.data.exports.length > 0">
```
kenoir: I think we're ok to allow this for all images? CC @blishen

kenoir:
@Reettaphant I think you could safely deploy 40e9800 separately, decoupling it from the PNG work entirely.

Reettaphant:
@kenoir I've now separated the crop link information into a separate component, and deleted the extra commits.

kenoir:
Very nice. :+1:

kenoir:
@Reettaphant Tried this in test. Looks good. Deploy at will.

paperboyo:
Just couple of points on how it looks, if I may (leaving aside if promoting the direct use of crops, incl. 140px, is beneficial, luckily Composer won’t allow them!).

This is great to at last see collapsible sub-panels! We will need them more and more, but (all IMHO!):

- the whole width of the title area should be clickable (incl. “Crops” title) and should expand/collapse a sub-panel - not just the chevron button area
- even if we decide that the chevron should be on the right (and not left, like in Collections), Grid uses ▸ for closed and ▾ for opened (not confusing ▾ for closed and ▴ for opened) - see e.g. date picker, Collections panel
- I find it strange that Crops sub-panel shows up in both Metadata _and_ Usages panel tabs

Maybe, when (and if) Crops panel is reworked to be independent of app state - it would be a good place for the Crops sub-panel? See e.g. this highly popular cropping material (4eccf49f2cc53d7fb74c73afdd20f9e0af3fc850):

crops---which-is-which

It’s very hard to tell which crop is which...

PR #1884, opened 2016-05-05
Adds List of currently in use AMIs with descriptions

kenoir:


NickPapacostas:
helpful! should it go in grid-infra though? 👍

kenoir:
@NickPapacostas thought about it - I think it's more likely to be useful here (alongside the place we're building the AMIs).

NickPapacostas:
yea that makes sense 👍

PR #1886, opened 2016-05-06
adding pointer-events: none to make original images un-draggable

NickPapacostas:
@paperboyo mentioned that he saw some folks dragging the original image into PS, this should prevent it

Do folks not like putting this in main.css? or adding it to all the places original images are shown?


kenoir:
:-1:  This is part of the print workflow right now, so unfortunately we can't do this.

PR #1890, opened 2016-05-09
Checkered background

Reettaphant:
Checkered background for transparent images:
screen shot 2016-05-26 at 14 51 09

### DIFF
--- a/kahuna/public/js/image/view.html
+++ b/kahuna/public/js/image/view.html
@@ -169,8 +169,10 @@
 
 <div class="image-holder">
     <div class="easel">
-        <div class="easel__canvas" ng:if="!ctrl.crop">
-            <img class="easel__image"
+        <div class="easel__canvas"
+            ng:if="!ctrl.crop"
+        >
+            <img class="easel__image easel__image--checkered__background"
                  alt="preview of original image"
                  ng:src="{{:: ctrl.optimisedImageUri}}"
                  grid:track-image="ctrl.image"
@@ -190,7 +192,7 @@
                ui:drag-data="ctrl.image | asImageAndCropsDragData:ctrl.crop"
                ui:drag-image="extremeAssets.smallest | assetFile">
                 <!-- TODO: Add tracking to crop -->
-                <img class="easel__image"
+                <img class="easel__image easel__image--checkered__background"
                      alt="cropped image"
                      ng:src="{{:: extremeAssets.largest | assetFile}}"
                 />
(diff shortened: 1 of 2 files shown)


paperboyo:
Wooo - stylish! Is it on a search page (in browser) or on an image page (in viewer)? Or on both?

Often to judge quality of a cutout, it's good practice to check it against white *and* against black. If it works there - will likely work everywhere (on all our card tones).

Checkered background is useful to distinguish between cutouts and non-cutouts quickly on a search page. Not so much to judge the cutout itself.

Just saying, because I'm not sure I would suddenly want to have a background colour switcher...

Reettaphant:
Do we want to have a checkered background behind transparent images? 

@blishen @NickPapacostas

paperboyo:
Not that we should emulate anyone... Here is Photoshop:

picture 21

How hard would it be to have three options:
- Grid's dark background
- something whiteish
- checkered

Easiest would be if one could cycle between them by e.g. Shift+clicking on an image and the choice would be remembered in localstorage.

All of the above, because I can't make up my mind what would be universally most useful... Ignore at will.

NickPapacostas:
@Reettaphant re: checkered background, i think it's a good idea. But couldn't we just do it really simply for V1 (i.e. it just defaults to checkered )

Reettaphant:
@NickPapacostas @paperboyo I agree with Nick. If it looks like something useful as it is, I'd like to release it now and then, if there is a need, perhaps look into making the UI more complex later.

paperboyo:
I'm extremely malleable. I agree with the majority. Also: always great to have a reason for moaning later 😜

Maybe just something a tad lighter than black checkers? Hard to see details of a cutout otherwise:
picture 22

kahuna/public/js/image/view.html
```diff
@@ -190,7 +192,7 @@
                ui:drag-data="ctrl.image | asImageAndCropsDragData:ctrl.crop"
                ui:drag-image="extremeAssets.smallest | assetFile">
                 <!-- TODO: Add tracking to crop -->
-                <img class="easel__image"
+                <img class="easel__image checkered__background"
```
NickPapacostas: As I've recently learned thanks to Shaun and Akash, we're trying to use [BEM](https://css-tricks.com/bem-101/) for selectors. 

so i _think_ this becomes 

``` HTML
<img class="easel__image easel__image--checkered__background"
```

Reettaphant:
@paperboyo What do you think of the new colouring scheme?

paperboyo:
@Reettaphant Perfect!

akash1810:
👍 :shipit:

PR #1891, opened 2016-05-09
Excluded getty metadata

NickPapacostas:
Some providers send us the Getty xmp metadata block when they shouldn't. 

We already had a convention to handle this situation, but I had to modify it to make it 

a) case sensitive
b) added "newspix internation" because they were sending us some with that and the regex wasn't matching. 

If I didn't do it the right scala way just hollar ^____^

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
@@ -170,7 +170,9 @@ trait GettyProcessor {
 }
 
 object GettyXmpParser extends ImageProcessor with GettyProcessor {
-  val excludeFromPatterns = List("newspix international", "i-Images").map(ex => s"(?i)${ex}".r)
+  // including 'newspix internation' because they're sending us lots with that in the credit
+  val excludeFromPatterns = List("newspix international", "newspix internation", "i-images")
+    .map(ex => s"(?i)$ex".r)
 
   // Some people send over Getty XMP data, but are not affiliated with Getty
   def excludedCredit(credit: Option[String]) = {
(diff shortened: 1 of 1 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -170,10 +170,14 @@ trait GettyProcessor {
 }
 
 object GettyXmpParser extends ImageProcessor with GettyProcessor {
-  val excludeFrom = List("newspix international")
+  // including 'newspix internation' because they're sending us lots with that in the credit
+  val excludeFromPatterns = List("newspix international", "newspix internation", "i-images")
+    .map(ex => s"(?i)$ex".r)
```
akash1810: Mapping over all items results in `List((?i)newspix international, (?i)newspix internation, (?i)i-images)`. Is this desired? It means `inewspix international` gets caught, for example.

NickPapacostas: I think the (?i) is scala's version of "/i", i.e. case insensitive 

http://daily-scala.blogspot.co.uk/2010/01/regular-expression-2-rest-regex-class.html

So I think this just creates a list of 3 case insensitive regexs

akash1810: ah, yep! my bad! Might want to add start and end line markers too?

NickPapacostas: Yea I thought about it but they weren't consistently in one place in the credit, I think it's ok because we're just saying "these aren't Getty" in this block. and sorry for the delay, got home and found our internet was gone >.< thanks for the help

akash1810:
👍

PR #1892, opened 2016-05-11
adding staff illustrators

NickPapacostas:


NickPapacostas:
👍 nice job, you're improving.

common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
```diff
@@ -130,7 +130,9 @@ object MetadataConfig {
 
   val staffIllustrators = List(
     "Mona Chalabi",
-    "Jan Diehm"
+    "Jan Diehm",
+    "Nadja Popovich",
+    "Guardian US Interactive Team"
```
akash1810: this feels odd

PR #1893, opened 2016-05-11
read stack name from file

akash1810:
easier for docker as the ~/.gu/grid directory is already shared and saves having to share env vars

`GRID_STACK_NAME` env var feels a bit ugly as we prepend it with `media-service-DEV-` - cleaner to just read a value from disk and use it.

### DIFF
--- a/cloud-formation/scripts/stack-name.sh
+++ b/cloud-formation/scripts/stack-name.sh
@@ -1,20 +1,21 @@
 #!/bin/bash
 
-if [ -z "$GRID_STACK_NAME" ];
-then
+STACK_NAME_FILE="${HOME}/.gu/grid/dev_stack_name"
+
+if [ -f ${STACK_NAME_FILE} ]; then
+    export STACK_NAME=`cat ${STACK_NAME_FILE}`
+else
     USER_NAME=`aws iam get-user`
 
     if [ $? != "0" ]; then
         echo -e "\033[1;41m    FAILED!    \033[m"
         echo '`iam get-user` failed. Are you using temporary credentials?'
-        echo 'If you are using temporary credentials please set the GRID_STACK_NAME environment variable and try again:'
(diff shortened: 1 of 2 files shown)


cloud-formation/scripts/stack-name.sh
```diff
@@ -1,20 +1,21 @@
...
         echo '`iam get-user` failed. Are you using temporary credentials?'
-        echo 'If you are using temporary credentials please set the GRID_STACK_NAME environment variable and try again:'
+        echo "If you are using temporary credentials please create ${STACK_NAME_FILE} with the name of your stack"
         echo ''
-        echo "  export GRID_STACK_NAME=foo && $0"
+        echo "  echo media-service-DEV-foobar > ${STACK_NAME_FILE} && $0"
         echo ''
         exit 1
     else
         export STACK_NAME="media-service-DEV-`$USER_NAME | jq '.User.UserName' | tr -d '"' | tr [A-Z] [a-z]`"
```
NickPapacostas: Reetta and I have been having trouble running this script, we'd get this error:

``` ./stack-name.sh: line 19: {: command not found ```

I can't figure out exactly why it's happening, but I figured out that if you call ``` aws iam get-user ``` again in the else clause, it seems to fix the issue:
```bash
        TRIMMED_USER_NAME="media-service-DEV-`aws iam get-user | jq '.User.UserName' | tr -d '"' | tr [A-Z] [a-z]`"
        export STACK_NAME=$TRIMMED_USER_NAME

```

NickPapacostas: Happy to make a commit for it just wanted to run it by you here

NickPapacostas:
👍 while debugging the issue in the comment i saw it in action ^_^

cloud-formation/scripts/stack-name.sh
akash1810: don't quite understand what you mean - happy to chat IRL if helpful?

PR #1894, opened 2016-05-12
handling getty's sloppy AssetId capitalisation

NickPapacostas:
For ages we've only been catching "AssetID" and after they bothered us about this again we looked at their data again and they have been sending "AssetID" sometimes and "AssetId" other times. 

we've informed them of this, but as we're already a few years behind on using this id, we figured it'd be good to start storing it asap


Reettaphant:
👍

PR #1881, opened 2016-05-03
[WIP] Make crops into panel component

kenoir:
In order that viewing the available crops can be seperated from the
image page. With a view to:

- Generalising a method for making crops unavailable
- Providing a view of crops in the search (and gallery) view

screenshot from 2016-05-03 13-24-30

PR #1895, opened 2016-05-13
Rv crop link layout

Reettaphant:
Styles the show png crops button similarly to the show metadata button and places the crop links with the metadata so that they are not displayed when viewing usages.

### DIFF
--- a/kahuna/public/js/components/gr-display-crops/gr-display-crops.css
+++ b/kahuna/public/js/components/gr-display-crops/gr-display-crops.css
@@ -1,6 +1,6 @@
 .gr-display-crops {
-    display: flex;
-    justify-content: space-between;
+    color: #999;
+    font-size: 1.4rem;
 }
 
 .gr-display-crops__crop {
--- a/kahuna/public/js/components/gr-display-crops/gr-display-crops.html
+++ b/kahuna/public/js/components/gr-display-crops/gr-display-crops.html
@@ -1,19 +1,11 @@
 <div class="image-info">
     <dl class="image-info__group image-info__wrap" ng:if="ctrl.crops.length > 0">
         <div class="gr-display-crops">
-            <dt class="image-info__heading image-info__wrap">Crops</dt>
-            <button
-                ng-if="!ctrl.showCrops"
-                ng-click="ctrl.showCrops=true"
-            >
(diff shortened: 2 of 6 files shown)


paperboyo:
Could I ask for a screenshot for the code-blind, please?

Reettaphant:
@paperboyo Screenshots: 
screen shot 2016-05-13 at 11 22 42
screen shot 2016-05-13 at 11 22 35

paperboyo:
Just IMHO!

This is wrong, cause it's consistent with something that's wrong too 😆 

1. Show full metadata is separated from above (partial) metadata by a rule and from Collections by the same rule. The rule above is wrong, because Full metadata is a part of metadata (hence the same font weight). The below rule is correct, because metadata and Collections are two parts.

2. Show full metadata should be just called Full metadata and the unnecessary trickery of switching Show to Hide got rid of. Chevron is widely understood and enough.

3. Show crops (which should be called Crops) is a distinct item, just as Collections and Labels are. It should be the same font weight/size. (Forgetting for now that it may really belong to the crop panel).

Because of the last sentence about the crop panel - this whole comment may be easily disregarded :-)

But, more generally, I think these small Show more things should look distinctively different to the whole categories which are more like collapsible sub-panels and entities in their own right. Do we have any already? Not sure. Collections and Labels are higher order than Show full metadata, but they are not exactly like Usages (which currently is shown as tab, not vertically as a sub-panel). Crops are not part of the metadata the way Labels or Collections are. But making Crops into a new tab makes no sense. 

I would just adjust it to be the size and weight of Labels and Collections got rid of both "Show" and "hide" parts and logic and wait and see on https://github.com/guardian/grid/pull/1881...

paperboyo:
Better think twice before giving him a screenshot!

Reettaphant:
I'd rather just fix the crop layout in this PR to make it consistent with the other collapsable section. Rethinking how this information is displayed would be another issue. I think being able to collapse the crops is important because it they can be quite distracting and not everyone uses them.

paperboyo:
> I think being able to collapse the crops is important because it they can be quite distracting and not everyone uses them

Oh yes! I was never suggesting otherwise (me English bad)!

NickPapacostas:
👍

PR #1899, opened 2016-05-16
Adding photoshot to excluded getty xmp providers

NickPapacostas:
We have no identified that our supplier model is wrong in that it assumes that any photo with Getty XMP is a Getty photo, which is incorrect, we should simply acknowledge the metadata and classify them as we would otherwise. 

However, this particular bug, on this particular day (due to Cannes) is actively costing us money. So i'm going to take the pseudo short-cut for now.


akash1810:
👍

PR #1901, opened 2016-05-17
remove GettyXMPParser 

NickPapacostas:
which was marking any photo with Getty XMP metadata a Getty photo.

After talking with Mateusz it turns out that isn't really true. Getty XMP can be used by anyone and some folks may choose to append it without sending the photo through Getty. 

I've tested this and the XMP still gets in ES, we just don't mark it as a getty photo. 

If this seems too heavy handed (as there may have been a specific reason it was put in) I can alternatively just remove the "usageRights = ... " part, which was the most costly problem. 

@paperboyo 
@akash1810 
@jamesgorrie

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
@@ -17,7 +17,6 @@ object SupplierProcessors {
     BarcroftParser,
     CorbisParser,
     EpaParser,
-    GettyXmpParser,
     GettyCreditParser,
     PaParser,
     ReutersParser,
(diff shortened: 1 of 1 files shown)


paperboyo:
> as there may have been a specific reason it was put in

Yeah, even though it's a hack, it's worth knowing why was it there before removal, I would say...

One day we will have to dip into XMP anyway. Or jump right in.

NickPapacostas:
the commit it was added: 178b16101216fd51c9d6ed89d071713e4e00e0ea 

The message is (unsurprisingly): "More aggressively attribute images to Getty based on credit"

I don't know what the specific issue it was dealing with was, however, there is no circumstance where attributing an image to Getty based on it having a Getty XMP block is correct (right @paperboyo ?) 

I'm going to look now to see if there's another way we can check the XMP for something _actually_ Getty-indicating.

paperboyo:
> there is no circumstance where attributing an image to Getty based on it having a Getty XMP block is correct

Solely? No. But it may be an indicator... Without looking at the image db directly (I can't) it's hard to tell.

> I'm going to look now to see if there's another way we can check the XMP for something actually Getty-indicating.

That's much more correct approach. Remember, that we currently omit most of XMP (incl. e.g. dc:creator): https://github.com/drewnoakes/metadata-extractor/blob/4e409590cfb4a8fa50b75868902b378bd4cdf099/Source/com/drew/metadata/xmp/XmpReader.java#L208

akash1810:
I don't know enough of the history of this process to comfortably +1 it :(

pinging @jamesgorrie

PR #1905, opened 2016-05-19
[WIP] Prevent imageops from serving images that are too large

Reettaphant:


Reettaphant:
this work is now in rv-image-ops-png24

PR #1910, opened 2016-05-25
[WIP] Leases in ES / batching

NickPapacostas:
* add leases to ES
* add leases to Image model
* Leases publishing to SQS


kenoir:
Might be worth separating the UI changes from the backend.

kahuna/public/js/leases/leases.js
```diff
@@ -140,7 +128,7 @@ leases.directive('grLeases', [function() {
         bindToController: true,
         template: template,
         scope: {
-            image: '=grImage',
```
NickPapacostas: put this back so it'll work when deploying back end separately

PR #1912, opened 2016-05-27
Make sure that imageops does not serve png24s

Reettaphant:
- An optimised version of a large png is saved in a separate bucket when the image is uploaded. Imgops will serve this image instead of the original. 
- Changes in grid-infra: https://github.com/guardian/grid-infra/pull/243

### DIFF
--- a/cloud-formation/dev-template.json
+++ b/cloud-formation/dev-template.json
@@ -155,27 +155,6 @@
                 "Bucket": { "Ref": "ThumbBucket" }
             }
         },
-        "OptimisedPngBucket": {
-            "Type": "AWS::S3::Bucket"
-        },
-        "OptimisedPngBucketPolicy": {
-            "Type": "AWS::S3::BucketPolicy",
-            "Properties": {
-                "PolicyDocument": {
-                    "Id": "OptimisedPngBucketPolicy",
-                    "Statement": [ {
-                        "Sid": "OptimisedPngBucketWriteAccess",
-                        "Action": [ "s3:PutObject", "s3:GetObject", "s3:DeleteObject" ],
-                        "Effect": "Allow",
-                        "Resource": { "Fn::Join": [ "",
-                            [ "arn:aws:s3:::", { "Ref": "OptimisedPngBucket" } , "/*" ] ]
-                        },
-                        "Principal": { "AWS": { "Fn::GetAtt": [ "User", "Arn" ] } }
-                    } ]
-                },
-                "Bucket": { "Ref": "OptimisedPngBucket" }
-            }
-        },
 
         "KeyBucket": {
             "Type": "AWS::S3::Bucket"
@@ -719,9 +698,6 @@
         },
         "ThumbBucket": {
             "Value": { "Ref": "ThumbBucket" }
-        },
-        "OptimisedPngBucket": {
-            "Value": { "Ref": "OptimisedPngBucket" }
         },
         "KeyBucket": {
             "Value": { "Ref": "KeyBucket" }
--- a/cloud-formation/dev-template.json
+++ b/cloud-formation/dev-template.json
@@ -155,6 +155,27 @@
                 "Bucket": { "Ref": "ThumbBucket" }
             }
         },
+        "OptimisedPngBucket": {
+            "Type": "AWS::S3::Bucket"
+        },
+        "OptimisedPngBucketPolicy": {
+            "Type": "AWS::S3::BucketPolicy",
+            "Properties": {
+                "PolicyDocument": {
+                    "Id": "OptimisedPngBucketPolicy",
+                    "Statement": [ {
+                        "Sid": "OptimisedPngBucketWriteAccess",
+                        "Action": [ "s3:PutObject", "s3:GetObject", "s3:DeleteObject" ],
+                        "Effect": "Allow",
+                        "Resource": { "Fn::Join": [ "",
+                            [ "arn:aws:s3:::", { "Ref": "OptimisedPngBucket" } , "/*" ] ]
+                        },
+                        "Principal": { "AWS": { "Fn::GetAtt": [ "User", "Arn" ] } }
+                    } ]
+                },
+                "Bucket": { "Ref": "OptimisedPngBucket" }
+            }
+        },
 
         "KeyBucket": {
             "Type": "AWS::S3::Bucket"
@@ -698,6 +719,9 @@
         },
         "ThumbBucket": {
             "Value": { "Ref": "ThumbBucket" }
+        },
+        "OptimisedPngBucket": {
+            "Value": { "Ref": "OptimisedPngBucket" }
         },
         "KeyBucket": {
             "Value": { "Ref": "KeyBucket" }
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/ImageIngestOperations.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/ImageIngestOperations.scala
@@ -4,7 +4,7 @@ import java.io.File
 import scala.concurrent.Future
 import com.amazonaws.auth.AWSCredentials
 
-class ImageIngestOperations(imageBucket: String, thumbnailBucket: String, credentials: AWSCredentials)
+class ImageIngestOperations(imageBucket: String, thumbnailBucket: String, pngBucket: String, credentials: AWSCredentials)
   extends S3ImageStorage(credentials) {
 
   def storeOriginal(id: String, file: File, mimeType: Option[String], meta: Map[String, String] = Map.empty) =
@@ -13,8 +13,12 @@ class ImageIngestOperations(imageBucket: String, thumbnailBucket: String, creden
   def storeThumbnail(id: String, file: File, mimeType: Option[String]) =
     storeImage(thumbnailBucket, fileKeyFromId(id), file, mimeType)
 
+  def storeOptimisedPng(id: String, file: File) =
+    storeImage(pngBucket, fileKeyFromId(id), file, Some("image/png"))
+
   def deleteOriginal(id: String) = deleteImage(imageBucket, fileKeyFromId(id))
   def deleteThumbnail(id: String) = deleteImage(thumbnailBucket, fileKeyFromId(id))
+  def deletePng(id: String) = deleteImage(pngBucket, fileKeyFromId(id))
 
   def fileKeyFromId(id: String): String = id.take(6).mkString("/") + "/" + id
 }
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/ImageIngestOperations.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/ImageIngestOperations.scala
@@ -4,7 +4,7 @@ import java.io.File
 import scala.concurrent.Future
 import com.amazonaws.auth.AWSCredentials
 
-class ImageIngestOperations(imageBucket: String, thumbnailBucket: String, pngBucket: String, credentials: AWSCredentials)
+class ImageIngestOperations(imageBucket: String, thumbnailBucket: String, credentials: AWSCredentials)
   extends S3ImageStorage(credentials) {
 
   def storeOriginal(id: String, file: File, mimeType: Option[String], meta: Map[String, String] = Map.empty) =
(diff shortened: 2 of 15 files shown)


cloud-formation/dev-template.json
```diff
@@ -155,6 +155,27 @@
...
             }
         },
+        "OptimisedPngBucket": {
+            "Type": "AWS::S3::Bucket"
+        },
+        "OptimisedPngBucketPolicy": {
+            "Type": "AWS::S3::BucketPolicy",
+            "Properties": {
+                "PolicyDocument": {
+                    "Id": "MyPolicy",
```
NickPapacostas: OptimisedPngBucketPolicy maybe? ;)

image-loader/app/model/ImageUpload.scala
```diff
@@ -28,57 +29,92 @@ case object ImageUpload {
...
+              .getOrElse(Future.successful(None))
+            else
+              Future(None)
+
+        val thumbDimensionsFuture = FileMetadataReader.dimensions(thumb, Some("image/jpeg"))
+
+        val pngDimensionsFuture = if (uploadIsPng24)
+            FileMetadataReader.dimensions(pngOption.get, Some("image/png"))
+          else
+            Future(None)
```
NickPapacostas: Need another Scala person's feedback, but i think this can just be "None", it will get overlooked when used in the "for" comprehension below. 

@akash1810 is that about right? This is for my own understanding, not a blocker

image-loader/app/model/ImageUpload.scala
```diff
@@ -28,57 +29,92 @@ case object ImageUpload {
...
+            originalUsageRights = processedImage.usageRights
+          )
+        }
+          yield {
+            ImageUpload(uploadRequest, finalImage)
+          }
       }
     }
-  }
+  }.flatMap(f => f)
```
NickPapacostas: This seems shady, I'm surprised you need it, maybe it's because the "yield" in a "for" comprehension calls "map" whereas the arrows call "flatMap", so perhaps adding another flattening before the yield... not sure though, happy to sit with you and try it out.

imgops/imgops.template.conf
```diff
@@ -13,6 +13,19 @@ server {
...
 
+    location /optimised {
+        # We're using Google here as a resolver
+        # as Amazon's 172.16.0.23 is very very slow
+        resolver 8.8.8.8 8.8.4.4;
+        set $pngBucket {{OPTIMISED_PNG_BUCKET}};
+        # We might need to review what we do with bigger images
+        image_filter_buffer 50M;
+        image_filter_interlace on;
+        image_filter_jpeg_quality $arg_q;
```
NickPapacostas: So do we return jpegs from image ops? or is this line irrelevant?

kahuna/public/js/imgops/service.js
```diff
@@ -23,8 +23,18 @@ imgops.factory('imgops', ['$window', function($window) {
     }
 
     function getOptimisedUri(image, options) {
-        return image.follow('optimised', options).getUri().catch(() => {
-            return image.source.secureUrl || image.source.file;
+        return image.follow('fileMetadata').get().then(metadata => {
+            //TODO: add check for alpha once isAlpha property in metadata is working
+            if (metadata.data.colourModelInformation && metadata.data.colourModelInformation.colorType &&
```
NickPapacostas: maybe helper-ize this conditional as it's a bit verbose ^_^

NickPapacostas:
Looks good! the only thing I think is a potential blocker is that .flatMap at the end of the method I commented on.

cloud-formation/dev-template.json
```diff
@@ -155,6 +155,27 @@
                 "Bucket": { "Ref": "ThumbBucket" }
             }
         },
+        "OptimisedPngBucket": {
```
akash1810: why not use the existing imageBucket?

akash1810: you'll need to make the same change in grid-infra's cloud formation

common-lib/src/main/scala/com/gu/mediaservice/lib/ImageIngestOperations.scala
```diff
@@ -13,8 +13,18 @@ class ImageIngestOperations(imageBucket: String, thumbnailBucket: String, creden
   def storeThumbnail(id: String, file: File, mimeType: Option[String]) =
     storeImage(thumbnailBucket, fileKeyFromId(id), file, mimeType)
 
+  def storeOptimisedPng(id: String, file: File) = {
+    val key = fileKeyFromId(id)
```
akash1810: can remove `key` as not used?

common-lib/src/main/scala/com/gu/mediaservice/lib/ImageIngestOperations.scala
```diff
@@ -13,8 +13,18 @@ class ImageIngestOperations(imageBucket: String, thumbnailBucket: String, creden
...
+    storeImage(pngBucket, optimisedPngKeyFromId(id), file, Some("image/png"))
+  }
+
   def deleteOriginal(id: String) = deleteImage(imageBucket, fileKeyFromId(id))
   def deleteThumbnail(id: String) = deleteImage(thumbnailBucket, fileKeyFromId(id))
+  def deletePng(id: String) = deleteImage(pngBucket, fileKeyFromId(id))
```
akash1810: should `fileKeyFromId(id)` be `optimisedPngKeyFromId(id)` as that's how its stored above?

image-loader/app/model/ImageUpload.scala
```diff
@@ -28,57 +29,92 @@ case object ImageUpload {
...
+      val sourceDimensionsFuture = FileMetadataReader.dimensions(uploadedFile, uploadRequest.mimeType)
+
+      val thumbFuture = for {
+        fileMetadata <- fileMetadataFuture
+        colourModel <- colourModelFuture
+        iccColourSpace = FileMetadataHelper.normalisedIccColourSpace(fileMetadata)
+        thumb <- ImageOperations.createThumbnail(uploadedFile, thumbWidth, thumbQuality, tempDir, iccColourSpace, colourModel)
+      } yield thumb
+
+      val pngOption = if (uploadIsPng24) {
```
akash1810: minor: would `optimisedPngFile` be a clearer variable name?

image-loader/app/model/ImageUpload.scala
akash1810: don't think so as that'll return an `Object` rather than a `Future[Option[Dimensions]]`

imgops/imgops.template.conf
akash1810: similar to above, you'll a corresponding change in the grid-infra repo

image-loader/app/model/ImageUpload.scala
```diff
@@ -119,4 +160,13 @@ case object ImageUpload {
     )
   }
 
+  private def isPng24(mimeType: Option[String], fileMetadata: FileMetadata): Boolean =
+    mimeType match {
+      case Some("image/png") => {
+        val colourType = fileMetadata.colourModelInformation.get("colorType").getOrElse("")
```
akash1810: can just be `val colourType = fileMetadata.colourModelInformation.getOrElse("colorType", "")`

image-loader/app/model/ImageUpload.scala
akash1810: we're doing `if (uploadIsPng24)` a few times, might be worth extracting to an object?

cloud-formation/dev-template.json
Reettaphant: @akash1810 I thought it would be cleaner to have the optimised pngs in a separate bucket because they exist fora particular purpose.

image-loader/app/model/ImageUpload.scala
Reettaphant: @akash1810 I'm not sure what you mean by extracting to an object, should the boolean value be inside an object?

cloud-formation/dev-template.json
kenoir: Couldn't you simply prefix the pngs in the existing bucket?

Reettaphant: Yes, I could do that if it makes things simpler.

image-loader/app/model/ImageUpload.scala
akash1810: I was more thinking about encapsulating the `isPng` check within an object. Something like:

```scala
case class PngThing(optimisedFile: Option[File], dimensions: Future[Option[Dimensions]])

case object PngThing {
  private def isPng(file: File) = true
  
  def build (file: File) = {
    if (isPng(file)) {
      val optimisedFile = {
        val optimsedFile = tempDir.getAbsolutePath() + "/optimisedpng" + ".png"
        Seq("pngquant", "--quality", "1-85", file.getAbsolutePath(), "--output", optimsedFile)
        new File(optimsedFile)
      }

      val dimensions: Future[Option[Dimensions]] = FileMetadataReader.dimensions(optimisedFile, Some("image/png"))
      
      PngThing(Some(optimisedFile), dimensions)
    }
    else {
      PngThing(None, Future(None))
    }
  }
}

```

then usage becomes

```scala
val pngThing = PngThing.build(file)
pngThing.optimised
```

not a blocker, just a thought 😄

image-loader/app/model/ImageUpload.scala
akash1810: apologies as you've already changed this!

Is the use of empty string necessary? As it looks like we're using empty string for `None` which feels odd. Can it be replaced with a pattern match? Though I may be misunderstanding the code!

Something like:

```scala
mimeType match {
  case Some("image/png") -> {
    fileMetadata.colourModelInformation.get("colourType") match {
      case Some("True Colour") => true
      case Some("True Colour with Alpha") => true
      case _ => false
    }
  }
  case _ => false
}
```

image-loader/app/model/ImageUpload.scala
Reettaphant: Yes, that makes sense. I'll update that!

Reettaphant:
@kenoir @NickPapacostas I've now updated this to read from the image bucket only.

docker/configs/generators/generators/resources/templates/dot-properties/media-api.properties.template
```diff
@@ -3,6 +3,7 @@ aws.id={{AwsId}}
 aws.secret={{AwsSecret}}
 s3.image.bucket={{ImageBucket}}
 s3.thumb.bucket={{ThumbBucket}}
+s3.png.bucket={{OptimisedPngBucket}}
```
akash1810: is this (and in `thrall.properties.template` below) still needed?

image-loader/app/model/ImageUpload.scala
```diff
@@ -28,56 +75,73 @@ case object ImageUpload {
...
+          colourModel <- colourModelFuture
+          fullFileMetadata = fileMetadata.copy(colourModel = colourModel)
+
+          metadata = ImageMetadataConverter.fromFileMetadata(fullFileMetadata)
+          cleanMetadata = metadataCleaners.clean(metadata)
+
+          sourceAsset = Asset.fromS3Object(s3Source, sourceDimensions)
+          thumbAsset = Asset.fromS3Object(s3Thumb, thumbDimensions)
+
+          pngAsset = if (optimisedPng.isPng24)
```
akash1810: could `pngAsset` be a property of `optimisedPng`?

kahuna/public/js/imgops/service.js
```diff
@@ -33,4 +42,9 @@ imgops.factory('imgops', ['$window', function($window) {
         getLowResUri
     };
 
+    function isPng24(metadata) {
+        return metadata.data.colourModelInformation && metadata.data.colourModelInformation.colorType &&
+            (metadata.data.colourModelInformation.colorType === 'True Color' || metadata.data.colourModelInformation.colorType === 'True Color with Alpha');
```
akash1810: minor: could regex this

media-api/app/lib/ImageResponse.scala
```diff
@@ -137,18 +159,24 @@ object ImageResponse extends EditsResponse {
...
 
-  def imageLinks(id: String, secureUrl: String, withWritePermission: Boolean, valid: Boolean) = {
+  def imageLinks(id: String, secureUrl: String, securePngUrl: Option[String], withWritePermission: Boolean, valid: Boolean) = {
     val cropLink = Link("crops", s"${Config.cropperUri}/crops/$id")
     val editLink = Link("edits", s"${Config.metadataUri}/metadata/$id")
     val optimisedLink = Link("optimised", makeImgopsUri(new URI(secureUrl)))
+
+    val optimisedPngLink = securePngUrl match {
+      case Some(secureUrl) => Link("optimisedPng", makeImgopsUri(new URI(secureUrl)))
+      case _ => Link("", "")
```
akash1810: this feels a bit odd, could we make `optimisedPngLink` an `Option` instead?

akash1810:
couple of minor comments, but none are blockers, lets release and get feedback 👍

image-loader/app/model/ImageUpload.scala
Reettaphant: Getting a pngAsset could be method but not a property I think because calculating it requires knowing the source dimensions and s3PngOption, futures that are resolved just a few lines above this (although I guess calculating the s3PngOption future could be moved to pngAsset). The method would still require the isPng24 check which is annoying.

akash1810: > method but not a property

Yep! My bad.

> The method would still require the isPng24 check which is annoying.

You could make use of the check that's done in the `build` method by passing in the dimension future.

All that said, this is just to encapsulate logic into an object and is a micro optimisations and isn't a blocker to shipping! 😄

Issue #1907, opened 2016-05-25
Error generating dot_properties


comeback-techie:
I'm trying to setup grid app locally on my ubuntu 14.04 vm. The following are the steps I followed:
Installed/Cloned/Downloaded
java, sbt, git, nginx, exiftool, jq, awscli, pip, guardian grid

Configured
java
- set JAVA_HOME
- added JAVA_HOME to environment

awscli 
- generated aws key and secret key by creating IAM user Sarma using
  AWS credentials of Kiran Lanka
- aws configure profile media-service, 
- exported it to .profile,
- set it as default aws profile 

nginx
- created host hydlocal.in in hosts file
- created sub-domains grid.hydlocal.in and api.grid.hydlocal.in
- generated self-signed local ssl certificates for these domains
- added conf file under sites-available to add server, port, ssl and
  location for the subdomain 

Elasticsearch
- ran elasticsearch installer
- start elasticsearch

Cloud Formation
- created dev stack that resulted in:
  - creating new user under IAM service with media-service-DEV-user
  - creating new dev stack under cloudformation service with 
    media-service-DEV-
  - creating s3 buckets required for Grid app with their respective
    policies:
         - Image, ImageOrigin, ImageCollection
         - Collection, Key, Leases, Thumb, Topic
         - Indexed Image Metadata, S3Watcher, Usage

Now, when trying to generate the dot_properties as per README, I am getting below error. 

root@brns-VirtualBox:~/grid/grid/docker/configs/generators# python -m generators.dot_properties /etc/gu
2016-05-25 12:09:53,305 dot_properties INFO Start
2016-05-25 12:09:53,325 cloudformation INFO Using AWS profile media-service
2016-05-25 12:09:53,489 botocore.credentials INFO Found credentials in shared credentials file: ~/.aws/credentials
2016-05-25 12:09:53,660 boto3.resources.action INFO Calling iam:get_user with {}
2016-05-25 12:09:53,668 botocore.vendored.requests.packages.urllib3.connectionpool INFO Starting new HTTPS connection (1): iam.amazonaws.com
2016-05-25 12:09:54,956 cloudformation INFO Using stack media-service-DEV-Sarma
2016-05-25 12:09:55,030 botocore.vendored.requests.packages.urllib3.connectionpool INFO Starting new HTTPS connection (1): cloudformation.us-west-2.amazonaws.com
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 162, in _run_module_as_main
[...]

PR #1913, opened 2016-05-28
clearer readme instructions

akash1810:
`/etc/gu` is owned by sudo meaning you have to run the generators as sudo too.

For simplicity, `chown` it...


akash1810:
in response to #1907

NickPapacostas:
👍

PR #1917, opened 2016-06-01
add period to filename to add extension

akash1810:
also use es6 string interpolation

after a conversation with @NickPapacostas around image download filenames, I noticed we weren't adding a `.` character to the filename, so it never got an extension 😛


NickPapacostas:
👍 deploying to test now. Should solve Nicola's issue immediately

NickPapacostas:
merging and deploying. ty @akash1810 !!

Issue #1117, opened 2015-08-20
Support XMP and metadata reconcilliation


paperboyo:
[EDIT: Thanks for reopening. This issue is way too long and first comments just document me learning about what's wrong, through a specific case (still occurs). It gets more general in https://github.com/guardian/grid/issues/1117#issuecomment-138691921 where I realise we are not really reading XMP metadata (maybe, treat [this](https://github.com/guardian/grid/issues/1117#issuecomment-143583140) as a first comment). At the end there are some examples of how more robust could we be. I think, would we ever want to fix it, we should take a look at why metadata-extractor [omits loads of XMP](https://github.com/drewnoakes/metadata-extractor/blob/master/Source/com/drew/metadata/xmp/XmpDirectory.java#L82-L107) and then decide if we want to do reconciliation ourselves or should we, maybe, consider contributing to metadata-extractor directly. Or use Adobe's XMPToolkit. I may be wrong, as always!]

Hello,

Happened with just one image, but I suspect it has something to do with the new usage rights categories (never happened to me before). If not, and if it's an outlier - apologies and, please, close.

Steps:
1. Download 4fb55b5c7d85075e91d3e6727024635b374bffbf for which meta reads "By: Christopher Thomond, Credit: The Guardian"
2. Modify, so that the image is not recognised as existing (I've opened in PS, colour corrected, saved as JPG, with "_modified" added to the filename)
3. Upload to Grid.

Here is a screenshot of resulting image's metadata:
![usage_rights-reupload](https://cloud.githubusercontent.com/assets/6032869/9384339/d32c768c-4747-11e5-8860-02e086ebd507.jpg)

Ideally, all metadata should survive such a ride.

Regards
Mateusz

akash1810:
Can't reproduce on TEST

<img width="910" alt="screen shot 2015-09-07 at 22 02 03" src="https://cloud.githubusercontent.com/assets/836140/9722806/2afb33cc-55ac-11e5-8c1e-9da186603c3f.png">

paperboyo:
I can reproduce in both environments (the same image). I open it in Photoshop, modify and resave. It MAY BE that there is something wrong with this specific image metadata that hoses it by open/save in Ps. Or, something may be wrong with Ps metadata engine (unlikely).

Here is a Ps metadata dump from the original image:
```
[...]

PR #1918, opened 2016-06-02
Adding special instructions to metadata on upload

NickPapacostas:
This was requested by Jonny and seemed reasonable. 

here's what Jonny said:

> 1a. Is it possible for you to make the 'special instructions' metadata field visible at all times (including when uploading) so that we can safely place notes onto images without risk of them being published? At present all we can do is write special notes into the caption and hope that subs spot them and that they don't get published, which seems needlessly risky.

### DIFF
--- a/kahuna/public/js/upload/jobs/required-metadata-editor.html
+++ b/kahuna/public/js/upload/jobs/required-metadata-editor.html
@@ -92,6 +92,27 @@
                 ng:click="ctrl.batchApplyMetadata('copyright')"
             >⇔</button>
         </label>
+
+        <label class="job-info--editor__field flex-center">
+            <div class="job-info--editor__label text-small">Special Instructions</div>
+            <input
+                type="text"
+                name="special-instructions"
+                class="text-input job-info--editor__input"
+                ng:model="ctrl.metadata.specialInstructions"
+                ng:change="ctrl.save()"
+                ng:model-options="{updateOn: 'default blur', debounce: { default: ctrl.saveOnTime, blur: 0 }}"
+                ng:class="{ 'job-info--editor__input--with-batch': ctrl.withBatch }"
+            />
+
+            <button
+                class="job-editor__apply-to-all"
+                title="Apply these instructions to all your current uploads"
+                type="button"
+                ng:if="ctrl.withBatch"
+                ng:click="ctrl.batchApplyMetadata('specialInstructions')"
+            >⇔</button>
+        </label>
     </div>
     <!-- Angular doesn't submit a form without a submit element, bonza!
     see: https://docs.angularjs.org/api/ng/directive/form#submitting-a-form-and-preventing-the-default-action -->
--- a/kahuna/public/js/upload/jobs/required-metadata-editor.js
+++ b/kahuna/public/js/upload/jobs/required-metadata-editor.js
@@ -74,6 +74,7 @@ jobs.controller('RequiredMetadataEditorCtrl',
             byline: originalMetadata.byline,
             credit: originalMetadata.credit,
             copyright: originalMetadata.copyright,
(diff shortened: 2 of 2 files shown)


paperboyo:
Will this be conditional like Copyright (appear only if present)?

Doesn't the purpose of this overlap with our Restrictions functionality? Maybe Restrictions should just be decoupled from (and available for all) Usage Rights? There are also leases now that provide a much more strict control over what that is trying to accomplish... Shouldn't they be enough?

The reason I'm saying this is two-fold:
- there is gonna be another field on an Upload page and them being non-configurable/collapsible and some appearing conditionally (Copyright), the upload flow is less predictable (e.g. a number of tabs it gets to go from one field to the other changes)
- this field, differently to Restrictions, but similarly to Copyright, is a part of standard metadata schema, so usual problems apply: we do not read XMP, we do not update pics on download etc

NickPapacostas:
>Will this be conditional like Copyright (appear only if present)?

No this would be present all the time (that was Jonny's specific request at least). 

I think you're right that it overlaps with Restrictions, and that their coupling to Usage Rights might not be correct. As you said, the difference is that this comes from the standard schema. When you say 

>  we do not read XMP

do you mean that if a photo has "special instructions" metadata we won't capture that? I tried an example and it seemed to work. 

> we do not update pics on download

I didn't know that ^___^ but it makes sense. 
1) I'm not sure that's so bad (is it so bad?) and 
2) Since we already do that with other fields, I don't think that's a reason not to do it. 


To your point about leases, you're right, but it seems like special instructions are much softer restrictions than leases and serve a different purpose. from what I can tell at least

paperboyo:
> No this would be present all the time

Current thinking for Copyright seems to be: show it if it's there, but don't let users introduce it. Here, it looks like Jonny would like to introduce Special instructions where there were none. This would be a third device to perform a similar function (Restrictions and Usage Rights in general, Leases and Special instructions).

> do you mean that if a photo has "special instructions" metadata we won't capture that

We won't but only when it's in XMP but nowhere else (or if it's a PNG). So: not often :-)

> I didn't know that ^___^ but it makes sense.
> 1) I'm not sure that's so bad (is it so bad?) and
> 2) Since we already do that with other fields, I don't think that's a reason not to do it.

Yeah, https://github.com/guardian/grid/issues/823.
1. Not so bad. A bit bad. Sometimes (downloading to fix and reuploading) a major PITA.
2. You are right. I'm not sure if it's a bigger problem if we invent our own fields or use (but do not update, so in a way - mistreat) standard schema. What I'm sure is that one day it would be good to spend some time studying different schemas, our own model and decide how much of the standards we want to use, where do we want to introduce our own schema, because we have specific needs not served by others.

And then to try to be a good citizen. Because even though we are Consumer, Creator and Changer according to [MWG model](https://github.com/guardian/grid/issues/1117#issuecomment-143841313), we act as if we only consume (and then, only Tesco Basics)... 

> To your point about leases, you're right, but it seems like special instructions are much softer restrictions than leases and serve a different purpose

Ok. Softer than Leases. Harder than Restrictions, though? You can see what I'm getting at ;-)

I have no strong opinions here one way or another, just not sure adding this (and should be added to the info panel as well, of course) will not muddy things...

kenoir:
I'm reluctant to add a field that doesn't have an explicit (and unique) purpose. 

What did Jonny ask for again? Might be worth adding his request to the PR description to aid the discussion.

Re downloads not having changes - we probably want to have some way to version images before doing this.

NickPapacostas:
@kenoir good call, I've updated it. 

I don't feel strongly about it and since he did I figured it was pretty low cost. Open either way though.

kenoir:
@NickPapacostas sorry for :snail:. What Jonny is asking for sounds fine to me :+1:

PR #1922, opened 2016-06-06
Gallery performance

ShaunYearStrong:
Couple of quick tweaks. The debounce works to an extent, but a better solution is still needed for people scrolling quickly through images.

### DIFF
--- a/kahuna/public/js/preview/image.js
+++ b/kahuna/public/js/preview/image.js
@@ -111,7 +111,7 @@ image.directive('uiPreviewImageLarge', ['observe$', 'inject$', 'imgops',
 
                 const optimisedImage$ = image$.flatMap((image) => {
                     return Rx.Observable.fromPromise(imgops.getFullScreenUri(image));
-                });
+                }).debounce(5);
 
                 scope.$watch(() => ctrl.image.data.id, () => {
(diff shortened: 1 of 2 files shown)


kahuna/public/js/preview/image.js
```diff
@@ -111,7 +111,7 @@ image.directive('uiPreviewImageLarge', ['observe$', 'inject$', 'imgops',
 
                 const optimisedImage$ = image$.flatMap((image) => {
                     return Rx.Observable.fromPromise(imgops.getFullScreenUri(image));
-                });
+                }).debounce(5);
```
NickPapacostas: is this 5ms? Seems a little low ^_^

ShaunYearStrong: According to [the docs](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/debounce.md) it is, buuuut I tried higher and it drops too many. Even at 10 it won't do they next image along. Any ideas?

NickPapacostas: none :(

kenoir:
@ShaunYearStrong this ready to be reviewed?

PR #1923, opened 2016-06-06
Removing photographers per Jo's request:

NickPapacostas:
>I think the issue started with this old list - it contains several people who were not staff photographers:
>Marcus Mays
Richard Blake
James Mann
Emma Baddeley
James Michelson
Millie Burton 

>If we have images by any of these people on the grid, please make them chargeable,

I've checked and they're all chargeable. My only concern is around picdar importing @kenoir do you think it's an issue?


ShaunYearStrong:
👍

PR #1926, opened 2016-06-13
Use the Apache Commons HttpClient lib in Cropper

kenoir:
... to [talk to Media API](https://github.com/guardian/grid/blob/master/cropper/app/controllers/Application.scala#L143).

There is a bug in Play WS that prevents requests made to secure
endpoints using SNI from handshaking. Cropper is now behind an [SNI](https://en.wikipedia.org/wiki/Server_Name_Indication)
secured endpoint and is consequently suffering from this bug:

https://groups.google.com/forum/#!topic/play-framework/T7ZhclgAAMU

This PR moves the request functionality in Cropper to use Apache Commons
HttpClient library which supports SNI as of version 4.3.2, see:

https://issues.apache.org/jira/browse/HTTPCLIENT-1119

This PR is required to move to using Cloudfront for all user facing REST
APIs (in order that out-of-region request times are sped up, and network
reliability increased).

**This has been tested in TEST**

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -64,7 +64,7 @@ object Dependencies {
 
   val commonsNetDeps = Seq(
     "commons-net" % "commons-net" % "3.3",
-    "org.apache.httpcomponents" % "httpclient" % "4.3.1"
+    "org.apache.httpcomponents" % "httpclient" % "4.5.2"
   )
 
   val commonsIODeps = Seq("commons-io" % "commons-io" % "2.4")
--- a/cropper/app/controllers/Application.scala
+++ b/cropper/app/controllers/Application.scala
@@ -139,17 +139,51 @@ object Application extends Controller with ArgoHelpers {
   // TODO: lame, parse into URI object and compare host instead
   def isMediaApiUri(uri: String): Boolean = uri.startsWith(Config.apiUri)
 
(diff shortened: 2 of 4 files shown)


project/Dependencies.scala
```diff
@@ -64,7 +64,7 @@ object Dependencies {
 
   val commonsNetDeps = Seq(
     "commons-net" % "commons-net" % "3.3",
-    "org.apache.httpcomponents" % "httpclient" % "4.3.1"
+    "org.apache.httpcomponents" % "httpclient" % "4.5.2"
```
jamesgorrie: Why commons if it's cropper only?

kenoir: It's also used in the scripts package.

kenoir:
Worth also mentioning that another approach (as discussed with @sihil) is to specify the API endpoint in config rather than reading it from the API response. 

I've chosen to do this as it seemed cleaner than doing string substitution to fix the API URI and more in keeping with @theefer's ideas in https://github.com/argo-rest/spec.

theefer:
Gosh, is SNI still not fixed in latest Play? :sob:

NickPapacostas:
👍

kenoir:
@theefer I think it's fixed in the latest version. We should upgrade.

PR #1930, opened 2016-06-15
Service changes for Leases in ES

kenoir:


leases/app/lib/LeaseStore.scala
```diff
@@ -37,8 +39,8 @@ object LeaseStore {
   private def uuid = Some(UUID.randomUUID().toString)
   private def key(id: String) = UniqueKey(KeyEquals('id, id))
 
-  def put(lease: MediaLease) = Scanamo.put(client)(tableName)(lease.copy(id=uuid))
-  def delete(id: String) = Scanamo.delete(client)(tableName)(key(id))
+  def put(lease: MediaLease) = Future { Scanamo.put[MediaLease](client)(tableName)(lease.copy(id=uuid))  }
+  def delete(id: String) = Future { Scanamo.delete(client)(tableName)(key(id)) }
```
NickPapacostas: I added these so I could use the results as promises (i.e. chain subsequent actions), is this the right way to do it? 

I noticed the other methods in here returned mappable objects but these didn't so I wasn't sure.

media-api/app/lib/ImageExtras.scala
```diff
@@ -24,10 +24,16 @@ object ImageExtras {
...
   def validityMap(image: Image): Map[String, Boolean] = Map(
-    "no_rights"           -> !hasRights(image.usageRights),
-    "missing_credit"              -> !hasCredit(image.metadata),
-    "missing_description"         -> !hasDescription(image.metadata)
+    "no_rights"            -> !hasRights(image.usageRights),
+    "missing_credit"       -> !hasCredit(image.metadata),
+    "missing_description"  -> !hasDescription(image.metadata)
+  )
+
+  def validityOverrides(image: Image): Map[String, Boolean] = Map(
```
NickPapacostas: Open to other / better names/ways of doing this

media-api/app/lib/ImageResponse.scala
```diff
@@ -159,32 +138,21 @@ object ImageResponse extends EditsResponse {
...
-    }
-
-    val baseLinksWithOptimised = optimisedPngLink match {
-      case Some(link) => link :: baseLinks
-      case None => baseLinks
+      List(optimisedLink, imageLink, usageLink, leasesLink)
     }
 
-    if (valid) (cropLink :: baseLinksWithOptimised) else baseLinksWithOptimised
+    if (valid) (cropLink :: baseLinks) else baseLinks
```
NickPapacostas: I think these shouldn't be here

thrall/app/lib/ElasticSearch.scala
```diff
@@ -117,7 +117,23 @@ object ElasticSearch extends ElasticSearchClient with ImageFields {
...
+    prepareImageUpdate(id){ request =>
+      request.setScriptParams( Map(
+        "leaseByMedia" -> asGroovy(leaseByMedia),
+        "lastModified" -> asGroovy(lastModified)
+      ).asJava)
+        .setScript(
+          replaceLeasesScript +
+            updateLastModifiedScript,
+          scriptType)
+      .executeAndLog(s"updating leases on image $id with: $leaseByMedia")
```
NickPapacostas: this is noisy (prints out the whole media lease) 

It was helpful for me but maybe not in the long term?

NickPapacostas:
I think some of the optimised stuff is getting overridden here. 

Otherwise I just asked had a few conceptual comments, no blockers

leases/app/lib/LeaseStore.scala
kenoir: I think Scanamo now supports async requests, but this is ok for now.

thrall/app/lib/ElasticSearch.scala
kenoir: Probably fine for an audit trail to begin with.

NickPapacostas:
👍

PR #1931, opened 2016-06-16
Move Usage from CAPI Polling to Crier Kinesis Stream

kenoir:
In order to more accurately and reliably monitor image usage.


usage/app/lib/KinesisStreamReader.scala
```diff
@@ -0,0 +1,57 @@
...
+import java.util.UUID
+
+import com.amazonaws.auth._
+import com.amazonaws.auth.profile.ProfileCredentialsProvider
+import com.amazonaws.internal.StaticCredentialsProvider
+import com.amazonaws.services.kinesis.clientlibrary.interfaces.{IRecordProcessor, IRecordProcessorFactory}
+import com.amazonaws.services.kinesis.clientlibrary.lib.worker.{InitialPositionInStream, Worker, KinesisClientLibConfiguration}
+import com.gu.logback.appender.kinesis.helpers.CustomCredentialsProviderChain
+
+class KinesisStreamReader {
```
kenoir: As this is not a general Kinesis stream reader impl, might be better to give it a more specific name (like `CrierStreamReader`).

usage/app/Global.scala
```diff
@@ -17,6 +17,10 @@ object Global extends WithFilters(CorsFilter, RequestLoggingFilter, new GzipFilt
   }
 
   override def onStart(app: Application) {
+    val crierReader = new KinesisStreamReader()
+
+    crierReader.start()
```
kenoir: As we are starting the worker before we've subscribed to the event stream, what's happening to events in the gap? Are they being queued for when the subscription starts or lost?

kenoir:
Some whitespace stuff to fix I think, not really important.

What do you think the behaviour of this will be if there are multiple instances reading from the Kinesis stream, (like during a deployment)?

usage/app/lib/EventProcessor.scala
```diff
@@ -0,0 +1,100 @@
...
+      Logger.debug(s"Initialized an event processor for shard $shardId")
+    }
+
+    override def processRecords(records: JList[Record], checkpointer: IRecordProcessorCheckpointer): Unit = {
+
+      records.asScala.map { record =>
+
+        val buffer: Array[Byte] = record.getData.array()
+
+        for {
```
kenoir: Does this need to be a `for` comprehension?

usage/app/lib/EventProcessor.scala
```diff
@@ -0,0 +1,100 @@
...
+
+    private def processEvent(event: Event): Unit = {
+
+      val dateTime: DateTime = new DateTime(event.dateTime)
+
+      event.eventType match {
+        case EventType.Update => {
+
+          event.payload match {
+            case Some(content: EventPayload.Content) => {
```
kenoir: Do we know what kind of content is available in this stream? Fronts? Galleries?

usage/app/lib/KinesisStreamReader.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  protected val eventProcessorFactory = new IRecordProcessorFactory {
+    override def createProcessor(): IRecordProcessor =
+      new CrierEventProcessor()
+  }
+
+  lazy val worker = new Worker(eventProcessorFactory, config)
+
+  private lazy val workerThread =
+    new Thread(worker, s"${getClass.getSimpleName}-$workerId")
```
kenoir: Is a `Thread` the best way to do this? What about an Akka Actor?

usage/app/lib/UsageStream.scala
```diff
@@ -9,16 +10,22 @@ import com.gu.mediaservice.model.{PendingUsageStatus, PublishedUsageStatus}
 import model._
 
 object UsageStream {
-  val contentStream = MergedContentStream.observable
 
-  val observable = contentStream.flatMap((container: ContentContainer) => {
-    val usageGroupOption = UsageGroup
+  val contentStream: Observable[ContentContainer] = MergedContentStream.observable
+
+  //TODO: restore this back to flatmap when ready
```
kenoir: Remove comment?

kenoir:
Nice work. A few questions to answer but it'll be good to remove the polling.

kenoir:
Is there a way to test this against the existing polling method to ensure it's doing the same thing?

usage/app/lib/EventProcessor.scala
LATaylor-guardian: We don't provide fronts or galleries, only content https://github.com/guardian/content-api-models/blob/master/models/src/main/thrift/content/v1.thrift#L1206

and 

See: https://github.com/guardian/content-api-models/blob/master/models/src/main/thrift/events/crier/event.thrift#L46

We supply update, retrievableUpdate (when the event is too large for Kinesis) and delete events.

kenoir: @LATaylor-guardian Thanks Luke. Any plans to include other content types?

LATaylor-guardian: We don't at the moment as no one has expressed a need for them, but are happy to consider others if requested. Something like Fronts doesn't exists within CAPI so would not be something we personally would provide.

Reettaphant:
@kenoir Multiple instances can coordinate reading from the stream: http://docs.aws.amazon.com/streams/latest/dev/developing-consumers-with-kcl.html

usage/app/Global.scala
Reettaphant: They are being queued as explained here: http://reactivex.io/documentation/subject.html

kenoir:
@Reettaphant can you rebase / merge master here?

Reettaphant:
@kenoir yes, I've done that now.

kenoir:
:+1:

PR #1935, opened 2016-06-17
Batch leases front end

NickPapacostas:
batch-leases-search-view
leases-upload-page

TODO:
- Jumpiness on search page and uploads
- audit requests being made, some may be excessive
- [X] upload page styles generally
- [X] batch apply on uploads page?


kahuna/public/js/leases/leases.js
```diff
@@ -31,60 +47,81 @@ leases.controller(
...
                     )
                     .finally(() => {
-                        ctrl.editing = false;
-                        ctrl.adding = false;
                         ctrl.resetLeaseForm();
                 });
             }
         };
 
+        function addLease(newLease) {
```
kenoir: When called from a batch event we can't distinguish between the image that originated the update and the ones to which it _should_ apply. Consequently the lease is double applied to the originating image.

kahuna/public/js/leases/leases.js
```diff
@@ -31,60 +47,81 @@ leases.controller(
...
+
+        // These events allow this control to work as a hybrid on the upload page
+        const batchAddLeasesEvent = 'events:batch-apply:add-labels';
+        const batchRemoveLeasesEvent = 'events:batch-apply:remove-labels';
+
+        if (Boolean(ctrl.withBatch)) {
+            $scope.$on(batchAddLeasesEvent,
+                    (e, leases) => leases.map(lease => addLease(lease)));
+            $scope.$on(batchRemoveLeasesEvent,
+                    () => console.log("batch-remove-leases"));
```
kenoir: We need to add a remove all leases for an image method to the lease service I think.

kahuna/public/js/leases/leases.js
kenoir: Perhaps the batch apply should remove and re-add all leases (so multiple leases would be sent)?

kenoir:
Two things strike me from working on this PR:
- We need a better way to tell if leases have been updated when sending the `updated-leases` event, it works when adding or removing a single leases but if we want to be able to make multiple changes where the count remains the same *it wont work*.
- The upload page batch methods are a damn mess. This PR introduces **another** way of doing a batch update when one already exists. Needs looking at.

kenoir:
:+1: Release after referendum.

kenoir:
This PR now includes changes to media-api and leases - both apps will need to be released first before Kahuna.

NickPapacostas:
> make multiple changes where the count remains the same it wont work.

How can the count remain the same? Maybe if there's latency and you make two changes back to back? 
Regardless, I agree it's not the best way, I thought it would be a good proxy so we didn't have to compare the lease objects themselves. 

As for the batch uploads, I gave it a try and it works well so I say we go with your approach.  Ultimately I think it will be tough to unify label and lease batching from the uploads page because their models are different (labels aren't their own 1-1 objects with images from what I can tell). So I'm not sure we'll be able to do any better.

kahuna/public/js/components/gr-panels/gr-panels.css
```diff
@@ -18,7 +18,6 @@
     width: 290px;
     position: fixed;
     background: rgba(68, 68, 68, 0.9);
-    overflow: auto;
```
ShaunYearStrong: On long panels this may allow things to overflow out of the page. I'm not sure there's a good solution for this as you can't have it hidden in one direction and visible in the other. Just something someone may point out.

kahuna/public/js/leases/leases.css
```diff
@@ -93,14 +111,41 @@ gr-leases .gu-date-range__display {
...
+gr-leases.leases-flex {
+  display: flex;
+  flex-direction: row;
+}
+
+gr-leases.leases-flex ul {
+  display:flex;
+  flex-direction: row;
+  line-height: 1em;
+  font-size: .75em;
```
ShaunYearStrong: Mixing between `rem` and `em` is bad. tru to use just one, probably `rem` as it's far easier to predict what the result will be.

kahuna/public/js/leases/leases.css
```diff
@@ -43,15 +57,11 @@ gr-leases .metadata-line__key {
   margin-bottom: 10px
 }
 
-gr-leases gr-confirm-delete {
-  float: right;
-}
-
-.active.allowed {
+.allowed {
   background-color: lightgreen;
```
ShaunYearStrong: Use hex for consistency

PR #1938, opened 2016-06-27
make leases trigger persistence

NickPapacostas:
I noticed some images with leases have been removed by reaper. 

How do we feel about this, should having a lease (expired or current) trigger persistence? 

I was tempted to say if theres a "current" lease don't delete it, but there were a few expired leases with useful notes and some leases might be in the future and therefore not current.


kenoir:
> How do we feel about this, should having a lease (expired or current) trigger persistence? 

Yes.

:+1:

PR #1943, opened 2016-07-01
Usages stream

Reettaphant:
To try to make usages work on prod this pr: 

- does not merge preview/live streams together to avoid indexing errors
- varies the app name according to stage


kenoir:
:+1:

PR #1946, opened 2016-07-05
[WIP] Fix memory leak

Reettaphant:


Reettaphant:
@kenoir It's been running fine on test for quite a while. The limit of 20 is a random number, I'm not sure how to set the limit.

PR #1948, opened 2016-07-06
Usage improvements

Reettaphant:
- Use publish subject instead of replay subject
- wait for thrift deserialization before processing event to save memory
- process only one usage per image and content


kenoir:
:+1:

Reettaphant:
@kenoir This is now finished

Issue #280, opened 2014-12-04
Add css pre-processor


jamesgorrie:
The CSS is becoming WET so we should look at DRYing it.
- [ ] decide on pre-processor

paperboyo:
Closing as it’s not gonna be needed in CSS Level 78, so at the time we would get to it. 😜

PR #1958, opened 2016-07-15
[WIP] Switching from CSS to SCSS

shaundillon:
Because maintaining CSS is a faff.

To do:
- [x] Clean up anything the reset mucked up
- [x] Define and parameterise colours
- [x] "" spacing
- [x] "" font sizes
- [ ] Refactor the separate controller/directive CSS into something more generic
- [ ] Get the runner scripts to watch for changes and recompile


kahuna/public/config.js
```diff
@@ -63,6 +64,12 @@ System.config({
     "github:jspm/nodelibs-buffer@0.1.0": {
       "buffer": "npm:buffer@3.5.3"
     },
+    "github:jspm/nodelibs-constants@0.1.0": {
```
ShaunYearStrong: I have no idea where most of this came from.

kahuna/public/stylesheets/modules/_typography.scss
```diff
@@ -0,0 +1,14 @@
+/** =============================================
+
+_TYPOGRAPHY.SCSS - Type styles and sizes
+
+============================================= **/
+
+$baseFontFamily: "Guardian Agate Sans Web", Georgia;
```
paperboyo: Georgie is a serif font. Better fallbacks for Agate would be
`"Lucida Sans Unicode", "Lucida Grande", sans-serif`
I think.

paperboyo:
Cool! We could, maybe, use this opportunity to introduce proper italics fonts. It will add extra weight to the app (hopefully, just once, as they should be cached), but should prevent ugliness like e.g.:

picture 38

(they should be full charset, too - "original")

kahuna/public/stylesheets/modules/_typography.scss
ShaunYearStrong: Cool, noted!

paperboyo:
Fixes: https://github.com/guardian/grid/issues/280. Or it would.

shaundillon:
Closing due to inactivity

PR #1960, opened 2016-07-19
Use InstanceProfileCredentialsProvider where available.

kenoir:
In order that we can remove hard coded pan-domain credentials from the instances.

Depends on https://github.com/guardian/grid-infra/pull/263.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Panda.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Panda.scala
@@ -1,7 +1,7 @@
 package com.gu.mediaservice.lib.auth
 
-import com.amazonaws.auth.BasicAWSCredentials
-import com.amazonaws.internal.StaticCredentialsProvider
+import com.amazonaws.auth.{InstanceProfileCredentialsProvider, AWSCredentialsProviderChain, BasicAWSCredentials, AWSCredentials}
+import com.amazonaws.auth.profile.ProfileCredentialsProvider
 import com.gu.pandomainauth.action.AuthActions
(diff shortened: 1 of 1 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Panda.scala
```diff
@@ -20,11 +20,10 @@ trait PanDomainAuthActions extends AuthActions {
...
 
   override lazy val domain: String = properties("panda.domain")
-  lazy val awsKeyId                = properties("panda.aws.key")
-  lazy val awsSecretAccessKey      = properties("panda.aws.secret")
 
-  override def awsCredentialsProvider = new StaticCredentialsProvider(
-    new BasicAWSCredentials(awsKeyId, awsSecretAccessKey)
+  override def awsCredentialsProvider = new AWSCredentialsProviderChain(
+    new ProfileCredentialsProvider("media-service"),
+    new InstanceProfileCredentialsProvider
```
akash1810: does the ordering matter at all?

akash1810:
👍

common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Panda.scala
philwills: It does in theory, but in practise you're unlikely to have both profile credentials and instance profile credentials.

kenoir: Yup. http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/AWSCredentialsProviderChain.html

sihil:
👍

PR #1963, opened 2016-07-25
[WIP] Use publication date from BlockElement rather than LastModified on Content.

kenoir:
In order to more accurately record publication date for an image.

### DIFF
--- a/usage/app/Global.scala
+++ b/usage/app/Global.scala
@@ -25,12 +25,12 @@ object Global extends WithFilters(CorsFilter, RequestLoggingFilter, new GzipFilt
     UsageApi.keyStore.scheduleUpdates(Akka.system(app).scheduler)
 
     // Eval subscription to start stream
-    previewSubscription
+//    previewSubscription
     liveSubscription
   }
 
   override def onStop(app: Application) {
-    previewSubscription.unsubscribe
+//    previewSubscription.unsubscribe
     liveSubscription.unsubscribe
   }
 }
--- a/usage/app/model/MediaUsage.scala
+++ b/usage/app/model/MediaUsage.scala
@@ -68,20 +68,19 @@ object MediaUsage {
       Try { item.getLong("date_removed") }.toOption.map(new DateTime(_))
     )
 
-  def build(elementWrapper: ElementWrapper, contentWrapper: ContentWrapper) = {
-    val usageId = UsageId.build(elementWrapper, contentWrapper)
-    val usageMetadata = UsageMetadataBuilder.build(contentWrapper.content)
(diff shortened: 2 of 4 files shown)


usage/app/Global.scala
```diff
@@ -25,12 +25,12 @@ object Global extends WithFilters(CorsFilter, RequestLoggingFilter, new GzipFilt
     UsageApi.keyStore.scheduleUpdates(Akka.system(app).scheduler)
 
     // Eval subscription to start stream
-    previewSubscription
+//    previewSubscription
```
kenoir: Remove these - left in for testing locally.

usage/app/model/MediaUsage.scala
```diff
@@ -68,20 +68,19 @@ object MediaUsage {
       Try { item.getLong("date_removed") }.toOption.map(new DateTime(_))
     )
 
-  def build(elementWrapper: ElementWrapper, contentWrapper: ContentWrapper) = {
-    val usageId = UsageId.build(elementWrapper, contentWrapper)
-    val usageMetadata = UsageMetadataBuilder.build(contentWrapper.content)
+  def build(elementWrapper: ImageElementWrapper, usageMetadata: DigitalUsageMetadata, groupId: String) = {
```
kenoir: Now we're being more specific about what we pass in too.

usage/app/model/UsageGroup.scala
```diff
@@ -1,11 +1,12 @@
 package model
 
 import com.gu.contentapi.client.model.v1.Content
-import com.gu.contentapi.client.model.v1.{ElementType, Element}
+import com.gu.contentapi.client.model.v1.{ElementType, Element, BlockElement, CapiDateTime, ImageElementFields}
```
kenoir: Some of these can be removed.

usage/app/model/UsageGroup.scala
```diff
@@ -25,10 +25,23 @@ object UsageGroup {
...
+      lastModified
+    ).map(contentWrapper => {
+        val group = UsageGroup(
+          createUsages(contentWrapper).toSet,
+          contentWrapper.id,
+          status,
+          lastModified
+        )
+
+        if (group.usages.isEmpty) {
```
kenoir: Seems weird to be putting this in an option, can it be handled higher up?

kenoir:
Sadly the block publication date is being set incorrectly by Composer (it seems to practically be the creation date). So this will have to hold until that problem is fixed.

kenoir:
Closing until the block publication date is fixed elsewhere.

PR #1964, opened 2016-07-27
Adds a date limit for recording usages

kenoir:
from firstPublicationDate on Content.

This is intended to prevent us incorrectly recording usages on old
content that the Usages app has not seen before.

### DIFF
--- a/usage/app/lib/Config.scala
+++ b/usage/app/lib/Config.scala
@@ -31,6 +31,7 @@ object Config extends CommonPlayAppProperties with CommonPlayAppConfig {
   val defaultPageSize = 100
   val defaultMaxRetries = 6
   val defaultMaxPrintRequestSizeInKb = 500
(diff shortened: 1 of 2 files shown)


usage/app/lib/Config.scala
```diff
@@ -31,6 +31,7 @@ object Config extends CommonPlayAppProperties with CommonPlayAppConfig {
   val defaultPageSize = 100
   val defaultMaxRetries = 6
   val defaultMaxPrintRequestSizeInKb = 500
+  val defaultDateLimit = "2016-01-01T00:00:00+00:00"
```
akash1810: why a string and not a date object?

kenoir: Thought about it - i'd rather keep the defaults to primitives (as that's roughly what we'll be getting out of the properties files).

akash1810:
👍

PR #1965, opened 2016-08-01
Split auth Store into seperate files

kenoir:
To stop this file becoming too big

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/BaseStore.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/BaseStore.scala
@@ -1,0 +1,54 @@
+package com.gu.mediaservice.lib
+
+import org.joda.time.DateTime
+
+import akka.actor.Scheduler
+import akka.agent.Agent
+
+import scala.concurrent.duration._
+
+import org.slf4j.LoggerFactory
+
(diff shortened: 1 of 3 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/BaseStore.scala
```diff
@@ -0,0 +1,54 @@
...
+import scala.concurrent.duration._
+
+import org.slf4j.LoggerFactory
+
+import com.amazonaws.auth.AWSCredentials
+import com.amazonaws.AmazonServiceException
+
+import org.apache.commons.io.IOUtils
+
+import _root_.play.api.libs.json._
```
akash1810: any reason for `_root`?

akash1810:
👍

common-lib/src/main/scala/com/gu/mediaservice/lib/BaseStore.scala
kenoir: There is a `play` package in `common-lib` at the same level so it's necessary to be explicit here.

akash1810: oh 😞

PR #1967, opened 2016-08-01
Replace lastSeen date from search with client date

kenoir:
In order to fix a problem with weighted search where some images are not
displayed.

This should hopefully also simplify the update polling logic.

### DIFF
--- a/kahuna/public/js/search/results.js
+++ b/kahuna/public/js/search/results.js
@@ -157,11 +157,7 @@ results.controller('SearchResultsCtrl', [
             // results), unless we're reloading a previous search in
             // which case we reuse the previous time too
 
-            // FIXME: the resolution of uploadTime is seconds, which could
-            // not be enough to avoid multiple images sharing an
-            // uploadTime and issues with duplicate results in the
-            // queried set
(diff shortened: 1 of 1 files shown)


ShaunYearStrong:
👍

kahuna/public/js/search/results.js
```diff
@@ -157,11 +157,7 @@ results.controller('SearchResultsCtrl', [
             // results), unless we're reloading a previous search in
             // which case we reuse the previous time too
 
-            // FIXME: the resolution of uploadTime is seconds, which could
-            // not be enough to avoid multiple images sharing an
-            // uploadTime and issues with duplicate results in the
-            // queried set
-            const latestTime = images.data[0] && images.data[0].data.uploadTime;
+            const latestTime = moment().toISOString();
```
akash1810: too lazy to check... has `moment` been imported further up?

akash1810:
👍

kahuna/public/js/search/results.js
kenoir: oh yeah

PR #1970, opened 2016-08-08
Adds usage reindex from CAPI endpoint

kenoir:
This adds a REST endpoint which when called gets the content by id from CAPI, re-extracts images and then overwrites any existing usages with the resulting usages.

### DIFF
--- a/usage/app/controllers/UsageApi.scala
+++ b/usage/app/controllers/UsageApi.scala
@@ -6,6 +6,8 @@ import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 import scala.util.Try
 
+import org.joda.time.DateTime
+
 import play.api.libs.json.JsError
 import play.api.Logger
 import play.api.mvc.Controller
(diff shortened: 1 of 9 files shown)


usage/app/controllers/UsageApi.scala
```diff
@@ -86,6 +91,34 @@ object UsageApi extends Controller with ArgoHelpers {
...
+          container.map(LiveCrierContentStream.observable.onNext(_))
+        }
+        case _ => Unit
+      }
+    })
+
+    result
+      .map(_ => Accepted)
+      .recover { case error: Exception => {
+        Logger.error("UsageApi reindex for for content failed!", error)
```
jennysivapalan: Will this log the content ID? Might help for debugging and rechecking usage later?

usage/app/lib/UsageRecorder.scala
```diff
@@ -77,8 +77,13 @@ object UsageRecorder {
...
       val deletes = (dbUsageGroup.usages -- usageGroup.usages).map(UsageTable.delete)
-      val creates = (usageGroup.usages -- dbUsageGroup.usages).map(UsageTable.create)
-      val updates = (usageGroup.usages & dbUsageGroup.usages).map(UsageTable.update)
+      val creates = (if(usageGroup.isReindex) { usageGroup.usages } else {(usageGroup.usages -- dbUsageGroup.usages)})
+        .map(UsageTable.create)
+      val updates = (if(usageGroup.isReindex) { Set() } else {usageGroup.usages & dbUsageGroup.usages})
+        .map(UsageTable.update)
+
+      //val creates = (usageGroup.usages -- dbUsageGroup.usages).map(UsageTable.create)
+      //val updates = (usageGroup.usages & dbUsageGroup.usages).map(UsageTable.update)
```
jennysivapalan: Commented lines no longer needed?

jennysivapalan:
👍

PR #1971, opened 2016-08-09
Make restricted use invalid without lease

kenoir:
These changes make images without restrictions "invalid" for cropping, previously restrictions overrode validity making them croppable.

restrict_is_invalid


NickPapacostas:
👍  i feel like this one might get people more upset than paid use requiring a lease ^_^

kenoir:
@NickPapacostas maybe - this was requested by one of the picture editors. Currently adding restrictions allows cropping though which I think is counter-intuitive.

kenoir:
@blishen we happy with this?

PR #1972, opened 2016-08-09
Add link to image page for invalid image on upload

kenoir:
In order to make all uploaded images editable from the image page, this
change makes the "invalid" text into a clickable link so that users can visit
the full image page from the upload page.

### DIFF
--- a/kahuna/public/js/edits/image-editor.html
+++ b/kahuna/public/js/edits/image-editor.html
@@ -64,11 +64,12 @@
                    ng:switch-when="ready"
                    ui:sref="image({imageId: ctrl.image.data.id})">View image ▸</a>
 
-                <span class="result-editor__status status status--invalid"
-                      ng:switch-when="invalid">
+                <a class="result-editor__status status status--invalid"
+                   ng:switch-when="invalid"
+                   ui:sref="image({imageId: ctrl.image.data.id})">
                         Invalid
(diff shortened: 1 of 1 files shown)


kahuna/public/js/edits/image-editor.html
```diff
@@ -64,11 +64,11 @@
                    ng:switch-when="ready"
                    ui:sref="image({imageId: ctrl.image.data.id})">View image ▸</a>
 
-                <span class="result-editor__status status status--invalid"
+                <a class="result-editor__status status status--invalid"
                       ng:switch-when="invalid">
                         Invalid
-                    <gr-icon title="You need to add both a credit and description.">help</gr-icon>
-                </span>
+                    <gr-icon title="This image will be uncroppable">help</gr-icon>
```
akash1810: worth keeping the reason for uncroppableness in?

kenoir: The reasons are visible on the image page - there are now so many possible reasons that trying to squeeze an "invalid reasons list" into the UI here seems like a separate job.

The error message before was incorrect - in that the reasons it suggests for validity may not be true.

akash1810: Ah ok 👍

akash1810:
👍

PR #1975, opened 2016-08-09
Remove compulsary restrictions on Witness & Obits

kenoir:
Removes compulsary restrictions on GuardianWitness & Obits images


clloyd:
:+1:

PR #1978, opened 2016-08-10
Usage postfix

Reettaphant:
- Fix fetching usage postfix in dev environment

### DIFF
--- a/usage/app/lib/Config.scala
+++ b/usage/app/lib/Config.scala
@@ -46,6 +46,8 @@ object Config extends CommonPlayAppProperties with CommonPlayAppConfig {
 
   val usageDateLimit = Try(properties("usage.dateLimit")).getOrElse(defaultDateLimit)
 
+  val userName = properties("user.name")
+
   val topicArn = properties("sns.topic.arn")
   val composerBaseUrl = properties("composer.baseUrl")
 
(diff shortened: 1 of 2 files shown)


Reettaphant:
@kenoir This also requires allowing the user to perform iam:GetUser. Should be added to the janus list of things on janus that gives the permissions required to run usages, or to the dev cloudformation template?

usage/app/lib/Config.scala
```diff
@@ -80,7 +80,7 @@ object Config extends CommonPlayAppProperties with CommonPlayAppConfig {
   )
 
   private val iamClient: AmazonIdentityManagement =
-    new AmazonIdentityManagementClient(credentialsProvider)
+    new AmazonIdentityManagementClient(awsCredentials)
```
kenoir: Why not use `credentialsProvider`?

Reettaphant: That creates a security token is invalid exception. I assumed it was a problem with using credentials from janus to perform the request.

Reettaphant:
@kenoir I had a chat with Adam and it would be possible to create bespoke grid permissions on janus that would list individually all permissions used to run grid. However, he also pointed out that cloudformation credentials can be used to create dynamo tables the first time usages is run and the user name could be stored in the properties file. Doing it this way would avoid doing the work in janus.

kenoir:
:+1:

PR #1979, opened 2016-08-10
Replace "Invalid" with "Unusable"

kenoir:
Which is more descriptive of the actual state of the image, also adds a title hint to use leases.

Image page:
unusable_image

Upload page:
unusable_upload


akash1810:
👍

PR #1981, opened 2016-08-11
display usage statuses correctly

Reettaphant:
Usages was not picking usage events from the live stream stream. Deleted usages were only read from the preview stream, and that's why they were not displayed.


kenoir:
:+1:

PR #1982, opened 2016-08-11
place date picker relative to calling button

ShaunYearStrong:
Fixes Johnny's problem with the data picker overlay positioning

picture 97
picture 98

### DIFF
--- a/kahuna/public/js/leases/leases.css
+++ b/kahuna/public/js/leases/leases.css
@@ -67,9 +67,10 @@ gr-leases .metadata-line__key {
 }
 
 gr-leases .gu-date-range__overlay {
-  position: fixed;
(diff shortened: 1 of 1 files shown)


kenoir:
:+1:

kenoir:
Reverted as this causes:

screenshot from 2016-08-11 14-05-41

ShaunYearStrong:
@kenoir Ooops, right you are. Need to look at that z-index.

ShaunYearStrong:
@kenoir - I've fixed this problem *but* Removing the overflow means that when the panel contents go over a certain height, they become inaccessible. 

I'm not sure there's a great solution here to be honest, aside from calculating the date picker's position with JS on init.

PR #1983, opened 2016-08-12
Having edit permissions makes images valid

kenoir:
- Adds a validityOverride with users with write perms to images.
- Modifies the UI to show invalidReasons whatever the valid state.

invalid-croppable


kahuna/public/js/components/gr-crop-image/gr-crop-image.js
```diff
@@ -14,16 +14,8 @@ cropImage.controller('grCropImageCtrl', [
         let ctrl = this;
 
         function updateState () {
-            leaseService.allowedByLease(ctrl.image).then(allowed => {
```
kenoir: @NickPapacostas think we forgot to change this after leases ended up in ES.

PR #1987, opened 2016-08-16
Auth logs to elk stack and log error code when auth fails

jennysivapalan:
@kenoir I've followed the previous implementation https://github.com/guardian/grid/pull/1134 and it looks like auth should be cloudformed to push logs https://github.com/guardian/grid-infra/pull/119/files

I haven't got this running this locally as I haven't got my properties set up correctly yet. Perhaps could use test environment to check?

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/argo/ArgoHelpers.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/argo/ArgoHelpers.scala
@@ -55,7 +55,7 @@ trait ArgoHelpers extends Results {
 
   // TODO: find a nicer way to serialise ErrorResponse[Nothing] without this hack
   def respondError(errorStatus: Status, errorKey: String, errorMessage: String, links: List[Link] = Nil): Result = {
-    Logger.warn(s"Responding with error status $errorStatus, $errorMessage")
+    Logger.warn(s"Responding with error status ${errorStatus.header.status}, $errorMessage")
     val response = ErrorResponse[Int](
       errorKey     = errorKey,
       errorMessage = errorMessage,
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/argo/ArgoHelpers.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/argo/ArgoHelpers.scala
@@ -3,6 +3,7 @@ package com.gu.mediaservice.lib.argo
 import java.net.URI
 import play.api.libs.json.{Json, Writes}
 import play.api.mvc.{Results, Result}
+import play.api.Logger
 
 import com.gu.mediaservice.lib.argo.model._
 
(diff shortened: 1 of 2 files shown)


kenoir:
The TEST environment is currently configured to push logs to CODE ELK - which I think is broken. You might have a hard time testing this locally as you'll need to provide the app with permissions to push to a kinesis stream you have visibility of.

kenoir:
:+1:

common-lib/src/main/scala/com/gu/mediaservice/lib/argo/ArgoHelpers.scala
```diff
@@ -54,6 +55,7 @@ trait ArgoHelpers extends Results {
 
   // TODO: find a nicer way to serialise ErrorResponse[Nothing] without this hack
   def respondError(errorStatus: Status, errorKey: String, errorMessage: String, links: List[Link] = Nil): Result = {
+    Logger.warn(s"Responding with error status $errorStatus, $errorMessage")
```
kenoir: Worth noting that this will change the behaviour of all Grid applications using the logging framework.

kenoir:
Looks good in TEST:

From instance logs:
```
Aug 16 10:01:48 ip-172-31-11-117 auth[1024]: 2016-08-16 10:01:48,091 [New I/O worker #2] WARN  GUUID= application:112 - Responding with error status Result(401, Map()), Not authenticated
Aug 16 10:01:48 ip-172-31-11-117 auth[1024]: 2016-08-16 10:01:48,104 [New I/O worker #2] WARN  GUUID= application:112 - Responding with error status Result(419, Map()), Session expired, required to log in again
Aug 16 10:01:48 ip-172-31-11-117 auth[1024]: 2016-08-16 10:01:48,111 [New I/O worker #2] WARN  GUUID= application:112 - Responding with error status Result(403, Map()), Not authorized
Aug 16 10:01:48 ip-172-31-11-117 auth[1024]: 2016-08-16 10:01:48,127 [New I/O worker #2] WARN  GUUID= application:112 - Responding with error status Result(401, Map()), Invalid API key provided
```

common-lib/src/main/scala/com/gu/mediaservice/lib/argo/ArgoHelpers.scala
kenoir: Maybe `${errorStatus.code}` might be clearer in the logs.

jennysivapalan: Ah thanks - I will update it

jennysivapalan:
Great thanks @kenoir - I'll update to just the code now.

common-lib/src/main/scala/com/gu/mediaservice/lib/argo/ArgoHelpers.scala
jennysivapalan: Looks like it needs to `${errorStatus.header.status}` - will double check on TEST it's doing the right thing.

kenoir:
:+1:

PR #1988, opened 2016-08-16
Use version 0.1.6 of Pandular (increase timeout)

kenoir:
Uses the most recent version of Pandular with a configurable timeout.
Sets tge timetout to 7000 millis to (hopefully) better cater for users
with extremely high latency connections (e.g. AUS).


clloyd:
👍

PR #1989, opened 2016-08-16
Make missing metadata images always invalid

kenoir:
... regardless of leasing status.

screenshot from 2016-08-16 14-57-21

### DIFF
--- a/media-api/app/lib/ImageExtras.scala
+++ b/media-api/app/lib/ImageExtras.scala
@@ -40,14 +40,16 @@ object ImageExtras {
   import scala.concurrent.duration._
   import com.gu.mediaservice.lib.FeatureToggle
 
-  def validityMap(image: Image): Map[String, Boolean] = Map(
-    "paid_image"           -> Costing.isPay(image.usageRights),
-    "conditional_paid"     -> Costing.isConditional(image.usageRights),
-    "no_rights"            -> !hasRights(image.usageRights),
-    "missing_credit"       -> !hasCredit(image.metadata),
-    "missing_description"  -> !hasDescription(image.metadata),
-    "current_deny_lease"   -> hasCurrentDenyLease(image.leases),
-    "over_quota"           -> UsageQuota.isOverQuota(image.usageRights)
+  case class ValidityCheck(valid: Boolean, overrideable: Boolean)
+
+  def validityMap(image: Image): Map[String, ValidityCheck] = Map(
+    "paid_image"           -> ValidityCheck(Costing.isPay(image.usageRights), false),
+    "conditional_paid"     -> ValidityCheck(Costing.isConditional(image.usageRights), false),
+    "no_rights"            -> ValidityCheck(!hasRights(image.usageRights), true),
+    "missing_credit"       -> ValidityCheck(!hasCredit(image.metadata), true),
+    "missing_description"  -> ValidityCheck(!hasDescription(image.metadata), true),
+    "current_deny_lease"   -> ValidityCheck(hasCurrentDenyLease(image.leases), true),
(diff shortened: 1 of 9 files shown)


media-api/app/lib/ImageExtras.scala
```diff
@@ -40,29 +40,46 @@ object ImageExtras {
...
+    !optToBool(validityMap.find(_._2.valid == true))
+
+  def invalidityIsOverridden(
+    validityMap: Map[String, ValidityCheck],
+    validityOverrides: Map[String, Boolean]
+  ): Boolean = {
+      optToBool(validityOverrides.find(_._2 == true)) &&
+      !optToBool(validityMap.filter(_._2.valid == true)
+        .find(_._2.overrideable == false))
+  }
```
jennysivapalan: This is a bit hard to follow, is it worth a unit test to help with the logic?

media-api/app/lib/ImageExtras.scala
```diff
@@ -40,29 +40,46 @@ object ImageExtras {
...
+
+  def validityMap(image: Image): Map[String, ValidityCheck] = Map(
+    "paid_image"           -> ValidityCheck(Costing.isPay(image.usageRights), true),
+    "conditional_paid"     -> ValidityCheck(Costing.isConditional(image.usageRights), true),
+    "no_rights"            -> ValidityCheck(!hasRights(image.usageRights), true),
+    "missing_credit"       -> ValidityCheck(!hasCredit(image.metadata), false),
+    "missing_description"  -> ValidityCheck(!hasDescription(image.metadata), false),
+    "current_deny_lease"   -> ValidityCheck(hasCurrentDenyLease(image.leases), true),
+    "over_quota"           -> ValidityCheck(UsageQuota.isOverQuota(image.usageRights), true)
   )
```
jennysivapalan: Why do some things get an override?

media-api/app/lib/ImageExtras.scala
kenoir: My instinct is adding a test is adding _more_ complicated code to test something already complicated. A test might help catch changes in behaviour if someone comes along and breaks it by accident.

I'd prefer to make the code clearer (and probably more verbose) to help with both things.

media-api/app/lib/ImageExtras.scala
kenoir: The logic is:
- Images can be invalid -> Invalid images can't be cropped
- Images can be deemed invalid by a set of checks
- Image validity can be overridden (to valid) if any one of a set of checks is met
- Some of the set of checks that mark an image as invalid cannot be overridden

media-api/app/lib/ImageExtras.scala
```diff
@@ -13,56 +13,59 @@ object ImageExtras {
...
+      "conditional_paid"     -> createCheck(Costing.isConditional(image.usageRights)),
+      "no_rights"            -> createCheck(!hasRights(image.usageRights)),
+      "missing_credit"       -> createCheck(!hasCredit(image.metadata), false),
+      "missing_description"  -> createCheck(!hasDescription(image.metadata), false),
+      "current_deny_lease"   -> createCheck(hasCurrentDenyLease(image.leases)),
+      "over_quota"           -> createCheck(UsageQuota.isOverQuota(image.usageRights))
+    )
+  }
 
   def validityOverrides(image: Image, withWritePermission: Boolean): Map[String, Boolean] = Map(
```
jennysivapalan: I think `validityOverrides` should move up to be above where `shouldOverride` is defined then it makes it a bit easier to follow the logic of createCheck method. Or it should be inlined with `shouldOverride` (as that is a val so called just once).

jennysivapalan:
Definitely much simpler to follow 👍

PR #1993, opened 2016-09-14
Adds an option to switch off Crier stream reading

kenoir:
Leaving only the API active (in order to reduce contention between the
two functions in production).

This is also in order to run the historic usage reingest in a timely fashion.


kenoir:
Related to: https://github.com/guardian/grid/pull/1993

PR #2000, opened 2016-09-26
Use permissions API

kenoir:
- For consistency across tools
- To make updating permissions easier

This change requires the following dependencies to be released first:
- [x] Permissions App able to CRUD perms: https://github.com/guardian/permissions/pull/65
- [x] Apps can access perms S3 bucket: https://github.com/guardian/grid-infra/pull/267
- [x] Update to BucketPolicy for access to originating permissions bucket (no PR)
- [x] Batch copy existing permissions to Permissions API: https://github.com/guardian/permissions/pull/72

This change should only affect the following services:
- Media API
- Cropper


common-lib/src/main/scala/com/gu/mediaservice/lib/auth/PermissionsHandler.scala
```diff
@@ -1,25 +1,31 @@
...
+  def hasPermission(user: Principal, permission: Permission) : Future[Boolean] = {
     user match {
       case u: PandaUser => {
-        permissionStore.hasPermission(permission, u.email)
+        Permissions.get(permission)(PermissionsUser(u.email)).map {
+          case PermissionGranted => true
+          case PermissionDenied => false
+
+        // fail open
+        } recover { case  _ => true }
```
jennysivapalan: Just to check - if the call fail then we'll grant access? Assumption is we trust the user?
We should log the recover?

kenoir: Yes, it should fail open this side. I'll add logging here.

PR #2004, opened 2016-10-13
Add extra info to readme, make stack-name script work on os x

philmcmahon:
Hopefully should help the next person.

### DIFF
--- a/cloud-formation/scripts/stack-name.sh
+++ b/cloud-formation/scripts/stack-name.sh
@@ -16,6 +16,7 @@ else
         echo ''
         exit 1
     else
-        export STACK_NAME="media-service-DEV-`$USER_NAME | jq '.User.UserName' | tr -d '"' | tr [A-Z] [a-z]`"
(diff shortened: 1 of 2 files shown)


NickPapacostas:
@akash1810 i remember there was some complexity about this but i can't remember what >.< I thought we didn't need permanent credz

akash1810:
> I thought we didn't need permanent credz

Sort of. You can't call IAM methods over the CLI with Janus credentials. You can, however in the web console.

Might be better to remove these scripts entirely in favour of "use the cfn json template to create a stack" to avoid confusion?

akash1810:
> Might be better to remove these scripts entirely in favour of "use the cfn json template to create a stack" to avoid confusion?

https://github.com/guardian/grid/pull/1893 made some effort to do this; if you create a file in `~/.gu/grid/dev_stack_name` an IAM call won't be made.

philmcmahon:
I guess we could make the cfn script create /grid/dev_stack_name (if it doesn't already) and then I can update my note to just say 'make sure dev_stack_name exists'

PR #2006, opened 2016-10-27
Implement X-Frame-Options to prevent embedding

kenoir:
... in unwanted sites.

In order to prevent click-jacking attacks.

This change should continue to allow embedding in composer.

screenshot from 2016-10-27 12-18-33


jennysivapalan:
👍

kenoir:
Turns out the Content Security Policy added by default was too strict as we seem to use inline scripts. I've relaxed that filter, and trying again in test.

clloyd:
This has broken firefox, I think it needs to be https://composer.gutools.co.uk

PR #2009, opened 2016-10-28
Update running.md

fredex42:
adding info @Reettaphant  told me :)

### DIFF
--- a/docs/running.md
+++ b/docs/running.md
@@ -1,10 +1,8 @@
 # Running services
 
 Grid consists of many micro-services. There are a few different ways to start them.
-Note that some services don't compile until you hit them for the first time, once it's running you could try:
-
-for x in 900{1..9}; do curl http://localhost:$x; done
-for x in 901{0..11}; do curl http://localhost:$x; done
+Note that some services don't compile until you hit them for the first time.
+There is a script included that will do this for you called **the_pingler.sh**
 
 ## Manually
 From the **project root**:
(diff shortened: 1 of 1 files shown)


docs/running.md
```diff
@@ -1,6 +1,10 @@
 # Running services
 
 Grid consists of many micro-services. There are a few different ways to start them.
+Note that some services don't compile until you hit them for the first time, once it's running you could try:
+
+for x in 900{1..9}; do curl http://localhost:$x; done
+for x in 901{0..11}; do curl http://localhost:$x; done
```
Reettaphant: I'd remove these two lines here, just because you don't need all the services running most of the time. I've had loads of problems with sbt when running grid, and I think the more services get complied the harder it is. Maybe it might be useful to add a separate note for thrall because that's what people are using mostly?

fredex42: hmmm, maybe you're right.  I've since discovered the Pingler which is a slightly nicer interface to this, was going to update this line to reflect.  Maybe the best advice would be, run the Pingler and use grid_runner to only start up the services you need, or to start them in a specific order?

btw, I have not had concurrency issues with sbt in grid, but maybe that's just because I wasn't doing too much work on it.

Reettaphant: I think running all services with grid runner is the simplest way of getting grid running rather than starting services manually. Usually its just a matter of getting thrall to compile after that which is not a big issue.

NickPapacostas:
👍

PR #2010, opened 2016-10-31
Recent usage warning

fredex42:
show warning if there are "recent" usages. "recent" is defined in imageUsageController

@clloyd @ShaunYearStrong

### DIFF
--- a/kahuna/public/config.js
+++ b/kahuna/public/config.js
@@ -95,10 +95,6 @@ System.config({
     "github:jspm/nodelibs-util@0.1.0": {
       "util": "npm:util@0.10.3"
     },
-    "github:tapmodo/Jcrop@0.9.12": {
-      "css": "github:systemjs/plugin-css@0.1.13",
-      "jquery": "npm:jquery@3.1.1"
-    },
     "github:vitalets/angular-xeditable@0.1.9": {
       "angular": "github:angular/bower-angular@1.4.3",
       "css": "github:systemjs/plugin-css@0.1.13"
--- a/kahuna/public/config.js
+++ b/kahuna/public/config.js
@@ -95,6 +95,10 @@ System.config({
     "github:jspm/nodelibs-util@0.1.0": {
       "util": "npm:util@0.10.3"
     },
+    "github:tapmodo/Jcrop@0.9.12": {
+      "css": "github:systemjs/plugin-css@0.1.13",
+      "jquery": "npm:jquery@3.1.1"
+    },
     "github:vitalets/angular-xeditable@0.1.9": {
       "angular": "github:angular/bower-angular@1.4.3",
       "css": "github:systemjs/plugin-css@0.1.13"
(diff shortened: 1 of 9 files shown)


kahuna/public/js/components/gr-image-usage/gr-image-usage.js
```diff
@@ -58,12 +58,19 @@ module.directive('grImageUsage', [function() {
...
+    'imageUsagesService',
+    function (imageUsagesService) {
     const ctrl = this;
 
     ctrl.formatTimestamp = (timestamp) => {
         return moment(timestamp).fromNow();
     };
+
+    ctrl.isRecent = (timestamp) => {
+        var nowtime = new Date();
```
clloyd: This file seems to be using ES6 const/let. I think this should probably be `const` to match style

usage/app/lib/Config.scala
```diff
@@ -85,7 +85,7 @@ object Config extends CommonPlayAppProperties with CommonPlayAppConfig {
       .withRegion(Regions.EU_WEST_1)
 
   val postfix = if (stage == "DEV") {
-    iamClient.getUser().getUser().getUserName()
+    "b0rked"
```
clloyd: Do we want to commit this?

kahuna/public/js/components/gr-image-usage/gr-image-usage.js
```diff
@@ -58,12 +58,19 @@ module.directive('grImageUsage', [function() {
...
+        return moment(timestamp).isBetween(moment(nowtime).subtract(imageUsagesService.recentTime,'days'),moment(nowtime));
```
clloyd: Can we use isAfter like below, I think it's easier to understand what's going on

kahuna/public/js/components/gr-image-usage/gr-image-usage.js
kenoir: :+1:

usage/app/lib/Config.scala
kenoir: I think that was one of @Reettaphant's changes. Probably best to check with her what's going on here (if this isn't working for you and is the reason you've made this subsititution).

kahuna/public/config.js
```diff
@@ -95,6 +95,10 @@ System.config({
     "github:jspm/nodelibs-util@0.1.0": {
       "util": "npm:util@0.10.3"
     },
+    "github:tapmodo/Jcrop@0.9.12": {
```
kenoir: What's this for?

kenoir:
Looks good apart from comments.

usage/app/lib/Config.scala
fredex42: the reason i included it was that iamClient.getUser().getUser().getUserName() was causing an AWS permissions error when loading.  I agree that we could possibly change the string "b0rked" to something a little less irreverant, but I think we still need a static string value here unfortunately

fredex42: @Reettaphant do you have any preference?

kahuna/public/config.js
fredex42: I don't remember adding that manually, but I think it was a requirement somewhere along the line. I can remove it and see if it breaks :)

fredex42:
gridwarning

fredex42:
@kenoir - https://github.com/guardian/grid/pull/2011

PR #2011, opened 2016-10-31
fixing cryptic amazon 403 when starting up usages with temp credentials

fredex42:
I don't fully understand what usages is doing with these values, so I've just reduced the fatal error to a warning for the time being.  This does work fine on my system though, @kenoir @Reettaphant please could you advise if this is likely to break anything.

thanks

also @clloyd

### DIFF
--- a/usage/app/lib/Config.scala
+++ b/usage/app/lib/Config.scala
@@ -85,7 +85,13 @@ object Config extends CommonPlayAppProperties with CommonPlayAppConfig {
       .withRegion(Regions.EU_WEST_1)
 
   val postfix = if (stage == "DEV") {
-    iamClient.getUser().getUser().getUserName()
+    try {
+      iamClient.getUser().getUser().getUserName()
+    } catch {
+      case e:com.amazonaws.AmazonServiceException=>
+        Logger.warn("Unable to determine current IAM user, probably because you're using temp credentials.  Usage may not be able to determine the live/preview app names")
+        "tempcredentials"
+    }
   } else {
     stage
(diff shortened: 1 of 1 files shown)


fredex42:
@Reettaphant @kenoir what are the stack name parameters (that the postfix goes into) used for, out of interest?  Presumably they should match up with what you deployed from cloudformation, but there does not seem to be an automated way of doing that if you're using temp credentials.  Maybe we should add it as a configuration parameter?

Reettaphant:
I'm not sure what you mean. The postfix just gets attached to the stack name. I don't think we can get any unique names from the cloudformation if the plan is to create a shared dev stack for grid.

fredex42:
@Reettaphant @clloyd then maybe the best option is to simply name the "shared" dev stack media-service-livex-DEV and remove this check entirely?  Then if you want to run on a specific one for any reason you can alter this line?

usage/app/lib/Config.scala
```diff
@@ -85,7 +85,13 @@ object Config extends CommonPlayAppProperties with CommonPlayAppConfig {
       .withRegion(Regions.EU_WEST_1)
 
   val postfix = if (stage == "DEV") {
-    iamClient.getUser().getUser().getUserName()
+    try {
+      iamClient.getUser().getUser().getUserName()
```
akash1810: janus creds don't allow iam calls... not sure how this has been working 😕

clloyd: They will soon!

PR #2012, opened 2016-11-02
Recent usage indicator on index page

fredex42:
- turning the usage indicator red if it's recent on the index page as well
- modifying tool tip to explain why the indicator is red

### DIFF
--- a/kahuna/public/js/preview/image.js
+++ b/kahuna/public/js/preview/image.js
@@ -60,8 +60,6 @@ image.controller('uiPreviewImageCtrl', [
         imageUsagesService.getUsages(ctrl.image).hasDigitalUsages$;
 
     const usageListRecent$ = imageUsagesService.getUsages(ctrl.image).usageListAfter$(imageUsagesService.recentTime);
-
-        console.log("usageListRecent$: " + usageListRecent$)
     $scope.$on('$destroy', function() {
         freeUpdateListener();
     });
--- a/kahuna/public/js/preview/image.js
+++ b/kahuna/public/js/preview/image.js
@@ -59,10 +59,14 @@ image.controller('uiPreviewImageCtrl', [
     const hasDigitalUsages$ =
         imageUsagesService.getUsages(ctrl.image).hasDigitalUsages$;
 
+    const usageListRecent$ = imageUsagesService.getUsages(ctrl.image).usageListAfter$(imageUsagesService.recentTime);
+
+        console.log("usageListRecent$: " + usageListRecent$)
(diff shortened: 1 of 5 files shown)


fredex42:
redfront

kahuna/public/js/preview/image.js
```diff
@@ -59,10 +59,14 @@ image.controller('uiPreviewImageCtrl', [
     const hasDigitalUsages$ =
         imageUsagesService.getUsages(ctrl.image).hasDigitalUsages$;
 
+    const usageListRecent$ = imageUsagesService.getUsages(ctrl.image).usageListAfter$(imageUsagesService.recentTime);
+
+        console.log("usageListRecent$: " + usageListRecent$)
```
clloyd: Can we remove the console.log

kahuna/public/js/services/image/usages.js
```diff
@@ -42,6 +42,17 @@ imageUsagesService.factory('imageUsagesService', [function() {
             const hasPlatformUsages = (platform) =>
                 filterByPlatform(platform).every((group) => !group.isEmpty());
 
+            const usageListAfter$ = (since) => usages$.map((usagesList) => {
+                    const nowtime = new Date();
+                    return usagesList.filter((usage) => {
+                        // console.log("evaluating for " + usage);
```
clloyd: Again can we remove comments/console.logs

usage/app/lib/Config.scala
```diff
@@ -85,7 +85,7 @@ object Config extends CommonPlayAppProperties with CommonPlayAppConfig {
       .withRegion(Regions.EU_WEST_1)
 
   val postfix = if (stage == "DEV") {
-    iamClient.getUser().getUser().getUserName()
+    "ropnert"
```
clloyd: Not sure what we decided for this, but shouldn't be in this PR

kahuna/public/js/preview/image.js
fredex42: doh! will do

usage/app/lib/Config.scala
fredex42: agreed!

fredex42: again needed to put something temporary in so that it would work. discussion is in seperate PR.

fredex42:
now release :)

paperboyo:
Hello! This is not working correctly (is out-of-sync with Usage indicators on a picture page). If there is no way to fix it, I would get rid of it as safe images are marked as unsafe...

PR #2013, opened 2016-11-03
Adds Time & Life Picture to the Getty Collection exclusion list.

kenoir:
In order that those images show as pay for.

screenshot from 2016-11-03 13-49-06


akash1810:
👍

PR #2014, opened 2016-11-10
Fix usages count on Index page

clloyd:
This fixes the issues caused by #2012, the usages count is now based on `.count()` rather than length. Also refactored the usages service to provide a stream on recent usages, which is used by both `controller` and `preview/image.js`

cc @kenoir @paperboyo


NickPapacostas:
👍

PR #2016, opened 2016-11-15
WIP Auditing

NickPapacostas:
Adds messages to the auditing kinesis stream for image deletion. As is this does not include deletions by Reaper. The media api checks if a panda user made the request if yes it passes their email along which thrall then checks for the email to determine if it should also go to auditing.

The only reason it's WIP is because right now the "message" in kibana is "null" >.<

### DIFF
--- a/media-api/app/controllers/MediaApi.scala
+++ b/media-api/app/controllers/MediaApi.scala
@@ -187,7 +187,12 @@ object MediaApi extends Controller with ArgoHelpers {
         if (imageCanBeDeleted) {
           canUserDeleteImage(request, source) map { canDelete =>
             if (canDelete) {
-              Notifications.publish(Json.obj("id" -> id), "delete-image")
+              val email = request.user match {
+                case user: PandaUser => user.email
+                case _ => ""
+              }
+
+              Notifications.publish(Json.obj("id" -> id, "userEmail" -> email), "delete-image")
               Accepted
             } else {
               ImageDeleteForbidden
--- a/thrall/app/lib/Config.scala
+++ b/thrall/app/lib/Config.scala
@@ -43,4 +43,8 @@ object Config extends CommonPlayAppConfig {
   val healthyMessageRate = properties("sqs.message.min.frequency").toInt
 
   val dynamoTopicArn: String = properties("indexed.image.sns.topic.arn")
+
+  val auditingStreamName: String = properties("auditing.streamName")
+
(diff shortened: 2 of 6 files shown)


media-api/app/controllers/MediaApi.scala
```diff
@@ -187,7 +187,12 @@ object MediaApi extends Controller with ArgoHelpers {
         if (imageCanBeDeleted) {
           canUserDeleteImage(request, source) map { canDelete =>
             if (canDelete) {
-              Notifications.publish(Json.obj("id" -> id), "delete-image")
+              val email = request.user match {
+                case user: PandaUser => user.email
+                case _ => ""
+              }
+
+              Notifications.publish(Json.obj("id" -> id, "userEmail" -> email), "delete-image")
```
akash1810: iirc api keys have an identifier too - might be worth extracting that here

thrall/app/lib/Config.scala
```diff
@@ -43,4 +43,8 @@ object Config extends CommonPlayAppConfig {
   val healthyMessageRate = properties("sqs.message.min.frequency").toInt
 
   val dynamoTopicArn: String = properties("indexed.image.sns.topic.arn")
+
+  val auditingStreamName: String = properties("auditing.streamName")
```
akash1810: can we update the [dot properties template](https://github.com/guardian/grid/blob/master/docker/configs/generators/generators/resources/templates/dot-properties/thrall.properties.template) too?

thrall/app/lib/Notifications.scala
```diff
@@ -1,5 +1,57 @@
...
+import play.api.libs.json.JsValue
 
 object DynamoNotifications extends SNS(Config.awsCredentials, Config.dynamoTopicArn)
+
+
+object Auditing {
+  private lazy val auditingCredentialsProvider = Config.auditingKinesisWriteRole.map(new STSAssumeRoleSessionCredentialsProvider(_, "media-service"))
+
+  val region =  Region getRegion Regions.EU_WEST_1
+  val client = region.createClient(classOf[AmazonKinesisClient], auditingCredentialsProvider.getOrElse(new ProfileCredentialsProvider("cmsFronts")), null)
```
akash1810: > `new ProfileCredentialsProvider("cmsFronts")`

will this work within EC2?

akash1810:
Closing as we're moving towards killing the Auditing service.

PR #2018, opened 2016-11-23
tweaks after we got setup with the dev stack

paulmr:
This PR includes changes to the documentation and a couple of the setup scripts which handle a few of the gotchas that we (me and @jharewinton ) encountered while getting the grid up and running locally.

As we are not grid experts, we're happy to hear comments on any of these proposals and remove some of the changes if they aren't a good idea.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,12 @@ You will need to install:
 * Nginx
 * [GraphicsMagick](http://www.graphicsmagick.org/)
 `sudo apt-get install graphicsmagick` or `brew install graphicsmagick --with-little-cms2`.
+
+If you get a permissions issue when running
+``` brew install graphicsmagick --with-little-cms2 ```
+you can install little cms (``` brew install little-cms2 ``` )
+and then compile graphicsmagick from source, it will automatically pick up on the little cms installation
+
 * [awscli](https://aws.amazon.com/cli/)
 * [jq](https://stedolan.github.io/jq/)
 * [exiftool](http://www.sno.phy.queensu.ca/~phil/exiftool/)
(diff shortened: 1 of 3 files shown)


kenoir:
:+1: 

As an aside, we should probably be providing a VM image or container for Grid development with some description of the provisioning process (Dockerfile, Packerfile etc.) as that would make life easier (rather than trying to describe the setup process for multiple targets as here.

NickPapacostas:
@kenoir 

Agreed about the container. Last time I spoke to @akash1810 he mentioned wanting to rip out the docker stuff

README.md
```diff
@@ -82,27 +88,11 @@ Start Elasticsearch from the `elasticsearch` directory:
...
-```sh
-cd cloud-formation/scripts
-./create-dev-stack.sh
-```
-
-** The create-dev-stack script will use your default aws profile **
 
 ### Generate .properties files
 
+Firstly you will need to follow the guide in the dev-utils folder of this [repo](https://github.com/guardian/grid-infra). This will give you
```
akash1810: this is a private repo...

paulmr:
agreed @kenoir, with the only caveat that it should still be simple to replace individual components with those from your source directory for easier development (after having worked with the pluto image which runs well but is hard to modify).

If it ran via docker by compiling from a shared local fs that would be ideal I think.

akash1810:
> mentioned wanting to rip out the docker stuff

Yeah! Or at least update it to the latest version of docker... when's the next hack day? 😝

README.md
paulmr: indeed, which is why we figured it was safe to link to it, however, perhaps we should actually be more explicit about that in the README? (e.g. "Guardian devs can use this private repo to get help with their local setup").

PR #2019, opened 2016-12-02
Remove all integration tests.

susiecoleman:
I tried to run the integration tests, they are not functional and not normally used to testing. Discussed with @kenoir and decided that it was sensible to remove them to avoid future confusion.


kenoir:
Probably need to delete a bit here: https://github.com/guardian/grid/blob/e29eeb8af50b752a838deebe40f932daa5a404d8/project/Build.scala#L138 too.

susiecoleman:
@kenoir thanks removed this reference. I don't think there's any others...

kenoir:
:+1:

PR #2022, opened 2016-12-14
Track image usage in media atoms

susiecoleman:
Before: The usage service listened to the crier stream, extracted image elements and recorded the usage
Now: The usage service still records image elements but it also looks at the atoms on the crier stream to see if they are media atoms and if they have a poster image. If the atom has a poster image this usage is recorded.


usage/app/model/UsageGroup.scala
```diff
@@ -50,74 +46,143 @@ object UsageGroup {
...
+      atomId <- getImageId(atom)
+      if atomId != None
+    } yield atom
+  }
+
+  private def getImageId(atom: Atom): Option[String] = {
+    try {
+      val posterImage = atom.data.asInstanceOf[AtomData.Media].media.posterImage
+      posterImage match {
+        case Some(image) => Some(image.mediaId.replace("https://api.media.gutools.co.uk/images/", ""))
```
akash1810: can we use the [shared config value](https://github.com/guardian/grid/blob/f4060db088b257ad2f17d909f6b93a67818ca2e1/common-lib/src/main/scala/com/gu/mediaservice/lib/config/Services.scala#L24) here?

susiecoleman: Not sure... Definitely nicer to pull it out but because the media atom atom maker always uses PROD grid even in CODE I'm not sure if this is the right value to use?

akash1810: If CODE Grid Usages uses CODE capi it should be ok as CODE MAM uses CODE capi irrc

kenoir: Even if it should always use PROD I would push this value out to config as it's fairly easy to do (even if you hard code this in the app config or services) and is less surprising as all env config will be together.

usage/app/model/MediaUsage.scala
```diff
@@ -96,4 +96,21 @@ object MediaUsage {
     None,
     printUsage.dateAdded
   )
+
+  def build(atomWrapper: AtomWrapper, contentWrapper: ContentWrapper): MediaUsage = {
```
kenoir: Does it make sense to modify `ContentWrapper` so that it can contain both "content" and "atoms"?

kenoir:
Nice work :)

usage/app/model/MediaUsage.scala
susiecoleman: Created a more general MediaWrapper so I'm now passing less things around

usage/app/model/UsageGroup.scala
susiecoleman: Fixed

PR #2025, opened 2017-01-05
Set more flexible content security policy to prevent (and allow) iFraming

clloyd:
This replaces the older Frame options which wasn't flexible enough to allow embeds from all gutools domains with the newer CSP allowing wildcards.


clloyd:
@akash1810 test on TEST and seems to be working... (blocking on non-gutools and allowing on gutools)

akash1810:
I don't know much about CSP... I assume nested iframes are ok? As in the Grid inside MAM inside Pluto?

clloyd:
@akash1810 regarding nested iframes... all the pages in the chain need to be "allowed"

PR #2029, opened 2017-01-27
use secureUrl for crop list links

akash1810:
because https > http!


akash1810:
requested from Tim Hughes

I suspect a better solution would be to add a grid image picker to the app he is using, but this is useful nonetheless.

akash1810:
@ShaunYearStrong I believe its the Angular media atom app as he is part of Labs.

PR #2033, opened 2017-04-05
use media-service profile

akash1810:
use the media-service profile when running the s3watcher lambda in DEV


akash1810:
I don't think we'll see this problem within the lambda itself as it runs with its own [role](https://github.com/guardian/grid-infra/blob/master/platform/cloud-formation/media-service.json#L5042-L5056) (equivalent to an EC2 instance profile) so we wouldn't need to explicitly set the profile in the lambda execution environment.

PR #2035, opened 2017-04-05
[Health] Replace jscs with eslint in s3watcher lambda

gtrufitt:
Quick update to the s3watcher lambda to use eslint (config taken from the [pluto ingestion lambda](https://github.com/guardian/media-atom-maker/pull/356), let me know if it should be different?!) from jscs.


akash1810:
@gtrufitt is this ready to go?

gtrufitt:
Should be!

PR #2041, opened 2017-04-11
S3 elk logger

akash1810:
Updates S3Watcher to push logs directly onto the Kinesis stream that the ELK stack is consuming.

This allows us to remove the [cloud watch logs to logstash](https://github.com/guardian/grid/tree/master/cloudwatch-logs-to-logstash) lambda which doesn't work under node 6 (because of the nasty [eval](https://github.com/guardian/grid/blob/master/cloudwatch-logs-to-logstash/lambda/index.js#L19)).

It also uses the [elk-kinesis-logger](https://github.com/guardian/elk-kinesis-logger) module.


s3watcher/lambda/S3Watcher.js
```diff
@@ -3,35 +3,48 @@ const Rx = require('rx');
...
+        stack: EnvironmentConfig.stack,
+        app: EnvironmentConfig.app,
+        roleArn: EnvironmentConfig.loggingRoleArn,
+        streamName: EnvironmentConfig.loggingStream
+    };
 
-    const transfer = lambda.config.map(function(config){
-        return Transfer.init(lambda.s3Event, config);
-    });
+    console.log(JSON.stringify(props));
```
gtrufitt: Was this for testing or do you actually want to log these?

s3watcher/lambda/lib/Transfer.js
```diff
@@ -23,25 +22,24 @@ module.exports = {
...
-            Logger.logDelete(config.stage, s3Object);
-
+            logger.log('Deleting from ingest bucket.', s3Object);
             return S3Helper.deleteS3Object(s3Object);
         };
 
         const failGraceful = function(e) {
-            Logger.logImportFail(config.stage, e);
-            Logger.logCopyToFailBucket(config.stage, s3CopyObject);
+            logger.log('Import failed.', e);
```
gtrufitt: Super minor and not as part of this PR but maybe you could (possibly...) `return this` in the `log` method in the logger module so that you can chain? You could then be all like 

```
logger
  .log('Import failed.', e)
  .log('Copying to fail bucket.', e);
```

Which is slightly nicer?

s3watcher/lambda/S3Watcher.js
akash1810: for debugging to work out why the STS.AssumeRole call was failing

s3watcher/lambda/lib/Transfer.js
akash1810: 😍 method chaining is so nice! Want to raise a PR in the logger module? (I can do it tomorrow if not).

gtrufitt: Sure!

gtrufitt: Oke doke, I bumped the [version to `0.1.0`](https://www.npmjs.com/package/elk-kinesis-logger), I'm happy to update and implement this after you've merged though to avoid confusion

akash1810:
Follow up PRs:
- https://github.com/guardian/grid/pull/2042/files
- https://github.com/guardian/grid/pull/2042/files

akash1810:
I think this lambda can be further improved with ES6 classes, but wanted to keep this PR small.

akash1810:
[Upgrade lambdas running old versions of node](https://trello.com/c/qhNHKZHD/75-upgrade-lambdas-running-old-versions-of-node)

PR #2042, opened 2017-04-12
remove redundant lambda

akash1810:
S3Watcher will be pushing to ELK directly, rather than through this proxy.

Depends on https://github.com/guardian/grid/pull/2041, https://github.com/guardian/grid-infra/pull/288


akash1810:
[Upgrade lambdas running old versions of node](https://trello.com/c/qhNHKZHD/75-upgrade-lambdas-running-old-versions-of-node)

PR #2046, opened 2017-04-24
Parse RCS usage mails directly

mbarton:
The last days usage is emailed to the Grid as a CSV attachment (via SES). This was previously parsed using a lambda with the results placed as JSON in the config bucket and then periodically read by the Grid itself.

This PR removes the lambda and moves the parsing logic into the Grid itself. The `UsageStore` periodically updates by finding the latest object in a new bucket (where the mails will be received directly) and parsing the CSV directly from the mail.

Both the `UsageStore` and `QuotaStore` are now non-optional in the code-base and always constructed, even in DEV. An example mail has been placed in both the DEV bucket for testing purposes. The TEST and PROD buckets will receive new mails, only retaining them for 7 days to minimise the impact of listing the objects.

### DIFF
--- a/docker/configs/generators/generators/resources/templates/dot-properties/media-api.properties.template
+++ b/docker/configs/generators/generators/resources/templates/dot-properties/media-api.properties.template
@@ -11,3 +11,4 @@ s3.usagemail.bucket={{UsageMailBucket}}
 persistence.identifier=picdarUrn
 mixpanel.token={{mixpanel_token}}
 es.index.aliases.read=readAlias
+quota.store.key=rcs-quota.json
--- a/docker/configs/generators/generators/resources/templates/dot-properties/media-api.properties.template
+++ b/docker/configs/generators/generators/resources/templates/dot-properties/media-api.properties.template
@@ -7,6 +7,7 @@ auth.keystore.bucket={{KeyBucket}}
 sns.topic.arn={{SnsTopicArn}}
 cors.allowed.origins={{cors}}
 s3.config.bucket={{ConfigBucket}}
+s3.usagemail.bucket={{UsageMailBucket}}
 persistence.identifier=picdarUrn
 mixpanel.token={{mixpanel_token}}
 es.index.aliases.read=readAlias
--- a/media-api/app/lib/UsageStore.scala
+++ b/media-api/app/lib/UsageStore.scala
@@ -1,19 +1,20 @@
-package com.gu.mediaservice.lib.usage
+package lib
 
 import java.io.InputStream
 import java.util.Properties
 import javax.mail.Session
 import javax.mail.internet.{MimeBodyPart, MimeMultipart}
 
-import scala.concurrent.Future
-import org.joda.time.DateTime
-import play.api.libs.concurrent.Execution.Implicits._
-import play.api.libs.json._
-import play.api.libs.functional.syntax._
 import com.amazonaws.auth.AWSCredentials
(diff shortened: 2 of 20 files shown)


docker/configs/generators/generators/resources/templates/dot-properties/media-api.properties.template
```diff
@@ -7,6 +7,8 @@ auth.keystore.bucket={{KeyBucket}}
 sns.topic.arn={{SnsTopicArn}}
 cors.allowed.origins={{cors}}
 s3.config.bucket={{ConfigBucket}}
+s3.usagemail.bucket={{UsageMailBucket}}
 persistence.identifier=picdarUrn
 mixpanel.token={{mixpanel_token}}
 es.index.aliases.read=readAlias
+quota.store.key=rcs-quota.json
```
akash1810: not sure we still need this property...

media-api/app/lib/UsageStore.scala
```diff
@@ -59,29 +59,67 @@ object StoreAccess {
...
+  def csvParser(list: List[String]): List[SupplierUsageSummary] = {
+    def stripQuotes(s: String): String = s.stripSuffix("\"").stripPrefix("\"")
+
+    list.headOption match {
+      case Some(head) if head.split(',').toList.length == 2 =>
+        list.tail.map { line =>
+          val lineList = line.split(',')
+          val (supplier, count) = (
+            stripQuotes(lineList.head),
+            stripQuotes(lineList.tail.head)
```
akash1810: we know its a two column csv, `.tail.head` could be `.lift(2)`

akash1810: we could also check the values of the header row, to further ensure its the right csv we're looking for

media-api/test/lib/UsageStoreTest.scala
```diff
@@ -0,0 +1,21 @@
...
+import org.scalatest.{FunSpec, Matchers}
+
+class UsageStoreTest extends FunSpec with Matchers {
+  describe("Usage Store") {
+    it("should parse RCS usage emails") {
+      val stream = getClass.getResourceAsStream("/example.mail")
+
+      val lines = UsageStore.extractEmail(stream)
+
+      lines.head should be ("\"Cpro Name\",\"Id\"")
```
akash1810: `Cpro Name`?

docker/configs/generators/generators/resources/templates/dot-properties/media-api.properties.template
mbarton: vs hard-coding the name?

akash1810: vs not having it at all. From the looks of it, this prop was used to identify what S3 key to read; if were reading the latest S3 key now we should be able to get rid of this prop. (Though I could be reading the code incorrectly... 😉 )

mbarton: We're reading the latest for the usage store. This is for the quota store

akash1810: d'oh! my bad

media-api/test/lib/UsageStoreTest.scala
mbarton: That's what's in the mails!

PR #2052, opened 2017-05-03
Update & link not working (proposal)

kant:


akash1810:
Hi @kant, we've [updated the readme](https://github.com/guardian/grid/blob/master/docs/01-dev-setup.md#nginx) a little now to make the use of `dev-nginx` clearer. Do you think this PR is still needed?

PR #2060, opened 2017-06-09
upgrade angular and angular based packages

shaundillon:
This is tested and appears to be good on TEST.

There's a new warning introduced when adding to `selection` as it's an immutable iterable, but RxJS uses `length` when `immutableJS` recommends use of `size`. That said, as long as we don't upgrade immtable or RxJS (much bigger changes) then this is fine.


gtrufitt:
Do we want to upgrade the `rx-angular` package for the rx bindings too or is that too risky? Would seem the next jump would be 1.0.0 which might have performance/stability benefits?

shaundillon:
@gtrufitt - I didn't notice that one. I'll check compatibility ad see if it's just a quick bump. If so I'll add that too

PR #2064, opened 2017-07-04
Docs updates

Reettaphant:
Update docs
- fix broken link to dot properties
- run scripts from the folder they are in
- add a note about accepting certificates

We specify the url that the app will be running on. This is specific to the organisation that is running it but maybe we could add the guardian link somewhere visible?

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -44,3 +44,10 @@ on GNU Linux, it may be something like
 
 ### Grid-runner
 - When using grid-runner to run grid, individual services have to sometimes be restarted when switching branches
+
+### Accepting certificates
+- You have to manually accept the certificates for the urls for the different grid services.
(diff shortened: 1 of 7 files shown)


TROUBLESHOOTING.md
```diff
@@ -44,3 +44,10 @@ on GNU Linux, it may be something like
 
 ### Grid-runner
 - When using grid-runner to run grid, individual services have to sometimes be restarted when switching branches
+
+### Accepting certificates
+- You have to manually accept the certificates for the urls for the different grid services.
```
akash1810: as this is a problem with the cert from dev-nginx, maybe this note should be there instead?

Reettaphant: I think it is useful to mention it here. Because grid has so many services, this needs to be done multiple times and because of the number of errors, the original problem can be difficult to find even if you are aware of the issue.

PR #2065, opened 2017-07-04
move to teamcity

akash1810:
Moves Grid off Jenkins and into Teamcity because:
- the Jenkins instance is about to be recycled by AWS
- makes Grid CI more consistent with other Tools

What's been kept:
- one build per app
- conditional build triggers; `auth` will get triggered if there are changes to `auth/**`, `common-lib/**` or `project/**`, for example

What's changed:
- build identifiers in RiffRaff; they are now prefixed with `media-service::teamcity::`
- you can now see the status of each build a PR triggers in github

Things to note:
- the play projects have been templated in teamcity, meaning we can change them all by changing the template which is faster than doing each individually


akash1810:
[Move Grid to Teamcity](https://trello.com/c/JoEYH3d8/141-move-grid-to-teamcity)

akash1810:
depends on #2068

PR #2067, opened 2017-07-05
don't leak the domain

akash1810:
instead, use the [`apiBaseUri`](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/config/Services.scala#L24).


akash1810:
will merge and deploy once we've fixed usage deploys

akash1810:
depends on #2068

PR #2068, opened 2017-07-05
fix riff-raff.yaml

akash1810:
Riffraff uses the package name to determine where to upload the artefact in S3. That is, it was uploading to `/media-service-dist/upload/TEST/app.zip` rather than `/media-service-dist/TEST/usage/app.zip`

Tested and working on TEST 🎉


akash1810:
[Grid usage app on TEST failing on redeploys due to not being able to find config: capi.pollIntervalSeconds. Investigate as this could cause issues for PROD too.](https://trello.com/c/9lXcDVfo/74-grid-usage-app-on-test-failing-on-redeploys-due-to-not-being-able-to-find-config-capipollintervalseconds-investigate-as-this-cou)

akash1810:
see https://github.com/guardian/riff-raff/issues/446

PR #2072, opened 2017-07-19
allow for specifying crop type in url

Reettaphant:
- Adds a new crop type parameter to session storage
- During this session, only the crop size specified by the parameter is allowed
- Used by media atom maker to help users to select the correct crop size

### DIFF
--- a/kahuna/public/js/crop/controller.js
+++ b/kahuna/public/js/crop/controller.js
@@ -6,14 +6,18 @@ var crop = angular.module('kahuna.crop.controller', ['gr.keyboardShortcut']);
 
 crop.controller('ImageCropCtrl',
                 ['$scope', '$rootScope', '$stateParams', '$state',
-                 '$filter', '$document', 'mediaApi', 'mediaCropper',
+                 '$filter', '$document', '$window', 'mediaApi', 'mediaCropper',
                  'image', 'optimisedImageUri', 'keyboardShortcut',
                  function($scope, $rootScope, $stateParams, $state,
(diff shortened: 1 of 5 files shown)


kahuna/public/js/crop/controller.js
```diff
@@ -6,14 +6,18 @@ var crop = angular.module('kahuna.crop.controller', ['gr.keyboardShortcut']);
...
                  'image', 'optimisedImageUri', 'keyboardShortcut',
                  function($scope, $rootScope, $stateParams, $state,
-                          $filter, $document, mediaApi, mediaCropper,
+                          $filter, $document, $window, mediaApi, mediaCropper,
                           image, optimisedImageUri, keyboardShortcut) {
 
     const ctrl = this;
     const imageId = $stateParams.imageId;
+    if ($stateParams.cropType) {
+        $window.sessionStorage.setItem('cropType', $stateParams.cropType);
```
akash1810: can we use [storage.js](https://github.com/guardian/grid/blob/master/kahuna/public/js/util/storage.js) for this?

kahuna/public/js/util/storage.js
```diff
@@ -3,12 +3,17 @@ import angular from 'angular';
 export const storage  = angular.module('util.storage', []);
 
 storage.factory('storage', ['$window', function($window) {
-    function setJs(key, val) {
+    function setJs(key, val, toSessionStorage) {
```
akash1810: minor: could default `toSessionStorage` to `false` to be explicit

PR #2074, opened 2017-07-31
happy linter

akash1810:
This fixes the latency issues Aus was having. The linter was failing which was failing the js build and thus we weren't serving a js bundle. This makes the linter happy again!

TODO
- [x] deploy to TEST


akash1810:
we should fix CI too, but that's a different bit of work...

PR #2075, opened 2017-08-02
place datepicker next to button

shaundillon:
# Before

Off to the right, off the bottom of the page with no ability to scroll
picture 254

# After

Always below the button
picture 255


kahuna/public/js/leases/leases.css
```diff
@@ -66,10 +66,13 @@ gr-leases .metadata-line__key {
   background-color: red;
 }
 
+.image-details .gu-date-range__overlay {
```
shaundillon: The pickers in the sidebar still need this

PR #2081, opened 2017-09-15
Keep an exception, because it is useful

sihil:


sihil:
Hmmm. Can I still merge this @akash1810 ?

sihil:
(I've just rebased against master)

PR #2085, opened 2017-09-27
Add avalon to list of Getty credit exclusions

shaundillon:
Thanks to @susiecoleman.

Adding Avalon to this ever-expanding list. There's a good case for abstracting this away but this'll work for now.


shaundillon:
[Images being misattributed to Getty](https://trello.com/c/ExBiJyEW/182-images-being-misattributed-to-getty)

shaundillon:
@akash1810 Isn't [this](https://github.com/guardian/grid/blob/master/common-lib/src/test/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessorsTest.scala#L237) a test for that? I could just change it to check avalon

akash1810:
implicitly, yes! 👍

paperboyo:
> There's a good case for abstracting this away

Better to fix it properly that to abstract a hack. If I understood you correctly, sorry if not.

PR #2086, opened 2017-09-27
remove unused settings

akash1810:
we're on riffraff.yaml everywhere now and this val isn't used


akash1810:
[Auto update AMIs for remaining Grid services](https://trello.com/c/7KcptGQg/199-auto-update-amis-for-remaining-grid-services)

PR #2093, opened 2017-10-09
amigo remaining services

akash1810:


akash1810:
tested on TEST, deploying to PROD

.editorconfig
```diff
@@ -14,3 +14,6 @@ indent_size = 2
 
 [*.yml]
 indent_size = 2
+
+[*.yaml]
```
gtrufitt: No biggie, for future reference this could be `*.{yml,yaml}`  above

akash1810: https://github.com/guardian/grid/pull/2094/files 😄

gtrufitt: 💯

Issue #1381, opened 2015-10-19
User not valid vor media-service


kbossee:
I think I got most of the setup correctly to run this locally, but now I'm getting stuck on logging in.

I've got media-api and kahuna running.

When I open https://grid.local.be/ I get a 401, which redirects me to google.
After I log in, I get a 403 on https://grid.local.be/oauthCallback?...
The logs don't tell me much, just "user not valid for media-service"

```
2015-10-19 16:47:06,252 [play-akka.actor.default-dispatcher-20] DEBUG GUUID= application:74 - user not authed against grid.local.be, authing
2015-10-19 16:47:06,256 [play-akka.actor.default-dispatcher-20] INFO  GUUID= request:93 - 0:0:0:0:0:0:0:1 - "GET /login?redirectUri=https%3A%2F%2Fgrid.local.be%2Fsearch HTTP/1.0" 303 20 "https://grid.local.be/search" 4ms
2015-10-19 16:47:06,516 [ForkJoinPool-4-worker-3] DEBUG GUUID= application:74 - fresh user login
2015-10-19 16:47:06,516 [ForkJoinPool-4-worker-3] INFO  GUUID= application:93 - user kristof.bossee@gmail.com not valid for media-service
2015-10-19 16:47:06,517 [play-akka.actor.default-dispatcher-21] INFO  GUUID= request:93 - 127.0.0.1 - "GET /oauthCallback?state=n5ba3b3gkcqc35k558arfi9d9d&code=4/8S-8AJwRazM7M7eeT2uFg63apnL-8EaW7I04djQIjB4&authuser=0&session_state=51e55abc56ce9a93fb50068c11329e1090faf230..500f&prompt=none HTTP/1.0" 403 20 "https://grid.local.be/search" 180ms
```

In the pan-domain-authentication project I read something about 403, and CORS so I added that to my settings.ini but that didn't help.

```
[output]
; Directory to write files to (defaults to ./output). Generated files need to be put in /etc/gu/
directory = /etc/gu/

[aws]
; Cloudformation stackname. Hint aws cloudformation list-stacks | grep "media-service-"
stack-name = media-service-kbossee

[properties]
; Domain root of your site, e.g. media.foobar.co.uk
domain_root = grid.local.be

; Token used to access mixpanel. See https://mixpanel.com/help/questions/articles/where-can-i-find-my-project-token
mixpanel_token =

; Comma separated list of CORS domains.
cors = grid.local.be,api.grid.local.be

[...]

PR #2096, opened 2017-11-01
Configurable oauth domain 

noandrea:
Add
panda_oauth_domain: guardian.co.uk
panda_oauth_multifactor_enable: true

properties to configure the google authentication domain 

see guardian/grid#1381

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Panda.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Panda.scala
@@ -16,12 +16,9 @@ trait PanDomainAuthActions extends AuthActions {
     val oauthDomain:String = properties.getOrElse("panda.oauth.domain", "guardian.co.uk")
     val oauthDomainMultiFactorEnabled:Boolean = Try(properties("panda.oauth.multifactor.enable").toBoolean).getOrElse(true)
     // check if the user email domain is the one configured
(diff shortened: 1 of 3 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Panda.scala
```diff
@@ -1,18 +1,27 @@
...
-    (authedUser.user.emailDomain == "guardian.co.uk") && authedUser.multiFactor
+    val oauthDomain:String = properties.getOrElse("panda.oauth.domain", "guardian.co.uk")
+    val oauthDomainMultiFactorEnabled:Boolean = Try(properties("panda.oauth.multifactor.enable").toBoolean).getOrElse(true)
+    // check if the user email domain is the one configured
+    var authorized:Boolean = (authedUser.user.emailDomain == oauthDomain)
+    // if authorized check if multifactor is to be evaluated
+    if(authorized && oauthDomainMultiFactorEnabled) {
+      authorized = authorized && authedUser.multiFactor
+    }
+    return authorized
```
akash1810: would prefer to avoid the mutation, could we do something like:

```scala
val isAuthorized: Boolean = authedUser.user.emailDomain == oauthDomain
if (oauthDomainMultiFactorEnabled) isAuthorized && authedUser.multiFactor else isAuthorized
```

PR #2097, opened 2017-11-08
Fail build when JSHint fails

shaundillon:
https://trello.com/c/n8iIloIf/165-grid-build-doesnt-fail-when-js-linter-fails


shaundillon:
[Grid build doesn't fail when js linter fails](https://trello.com/c/n8iIloIf/165-grid-build-doesnt-fail-when-js-linter-fails)

PR #2100, opened 2017-11-15
restrict the crops user is able to pick

Reettaphant:
When you are cropping an image for mam in the grid, the grid only allows you crop images that are the correct crop type (crop type gets stored in session storage). 

This restriction does not stop people from navigating away from the crop page and choosing the wrong crop size from already existing crops. This pr makes sure that this does not happen by only allowing choosing crops of the correct type. 

When choosing a composer image: 

screen shot 2017-11-15 at 15 05 07

### DIFF
--- a/kahuna/public/js/image/view.html
+++ b/kahuna/public/js/image/view.html
@@ -84,29 +84,22 @@
                             ng:class="{'image-crop--selected': crop == ctrl.crop}">
                             <a draggable="true"
                                ng:init="extremeAssets = (crop | getExtremeAssets)"
+                               ng-if="ctrl.allowCropSelection(crop)"
                                ng:click="ctrl.cropSelected(crop)"
                                ui:sref="{crop: crop.id}"
                                ui:drag-data="ctrl.image | asImageAndCropsDragData:crop"
                                ui:drag-image="extremeAssets.smallest | assetFile">
 
-                                <img class="image-crop__image"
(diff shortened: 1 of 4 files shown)


paperboyo:
❤️ 

Is it going to be clear to users why some types/crops are unavailable, esp. if there are none yet of the allowed type? Is it worth adding some messaging (through tooltips on unallowed types/crops or a yellow banner or...)?

Reettaphant:
Good point. I guess we shouldn't assume people will know. I've added a message in a tooltip.

kahuna/public/js/image/view.html
```diff
@@ -84,29 +84,24 @@
                             ng:class="{'image-crop--selected': crop == ctrl.crop}">
                             <a draggable="true"
                                ng:init="extremeAssets = (crop | getExtremeAssets)"
+                               ng-if="ctrl.allowCropSelection(crop)"
```
akash1810: minor: an `ng-switch` may be more obvious here due to the resulting indentation.

kahuna/public/stylesheets/main.css
```diff
@@ -1811,6 +1811,10 @@ FIXME: what to do with touch devices
     border: solid 2px #00adee;
 }
 
+.image-crop--disabled {
+  opacity: 0.5;
```
akash1810: Should we also add `cursor: not-allowed`?

PR #2102, opened 2017-11-28
distinguish staff photographers

akash1810:
add a blue border around staff images to promote usage of them over agency images

image

a border was chosen over a [badge](https://material.io/icons/#ic_verified_user) as @paperboyo and I were concerned about badge overload


shaundillon:
a11y-wise: There should be some text marking these as staff photos too. Even if it's hidden or just the link title.

akash1810:
@shaundillon good point, added.

paperboyo:
❤️ Seconded, @shaundillon, but I think it can be useful being user-facing. Maybe info on hover? Most indicators have this, one can learn the app this way.

akash1810:
tooltip

PR #2109, opened 2018-01-19
USA TODAY Sports

mbarton:
USA TODAY sports


paperboyo:
Thx! There is [this thing](https://github.com/guardian/grid-batcher/) nobody know how to use to apply the change retroactively, too.

PR #2110, opened 2018-02-01
display save button on top

akash1810:
per Katherine Purvis's email


kahuna/public/stylesheets/main.css
```diff
@@ -195,6 +195,8 @@ button[disabled],
 .button-save {
     background: #00adee;
     color: white;
+    // above the edit button of a field below
+    z-index: 10;
```
akash1810: see z-index rules defined here - https://github.com/guardian/grid/blob/master/kahuna/public/stylesheets/main.css#L8

PR #2114, opened 2018-03-02
Remove preview capi

tomrf1:
`PreviewContentApi` is not used anywhere, so no need to migrate to the new preview endpoint


ci/play-build.sh
```diff
@@ -27,8 +27,7 @@ pushd ..
 
 echo "Building $PROJECT"
 
-java -Xmx1024m \
```
tomrf1: teamcity build was running out of heap space

PR #2123, opened 2018-04-20
remove jQuery

shaundillon:
Does as it says on the tin. Removes the need for jQuery. The other changes are due to restrictions in JQLite that comes with angular. Here's [the API ref](https://docs.angularjs.org/api/ng/function/angular.element#angularjs-s-jqlite) to clarify

### DIFF
--- a/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
+++ b/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
@@ -193,10 +193,7 @@ grCollectionsPanel.directive('grNode', ['$parse', '$compile', function($parse, $
 
                 compiledTemplate(scope, cloned => {
                     const container = element.find('gu-template-container');
-                    // JQLite doesn't have modern array methods :(
-                    for (let i = 0; i < cloned.length; i++) {
-                      container.append(cloned[i]);
-                    }
+                    Array.from(cloned).forEach(clone => container.append(clone));
                 });
 
                 grNodeCtrl.init(grCollectionTreeCtrl);
--- a/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
+++ b/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
@@ -194,7 +194,7 @@ grCollectionsPanel.directive('grNode', ['$parse', '$compile', function($parse, $
                 compiledTemplate(scope, cloned => {
                     const container = element.find('gu-template-container');
                     // JQLite doesn't have modern array methods :(
-                    for(let i=0; i < cloned.length; i++) {
+                    for (let i = 0; i < cloned.length; i++) {
                       container.append(cloned[i]);
                     }
                 });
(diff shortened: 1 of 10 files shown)


kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -192,7 +192,11 @@ grCollectionsPanel.directive('grNode', ['$parse', '$compile', function($parse, $
                 }
 
                 compiledTemplate(scope, cloned => {
-                    element.find('.node__children').append(cloned);
+                    const container = element.find('gu-template-container');
+                    // JQLite doesn't have modern array methods :(
+                    for (let i = 0; i < cloned.length; i++) {
```
akash1810: could do `Array.from(cloned).forEach(foo => bar)`?

kahuna/public/js/components/gu-date-range/gu-date-range.js
```diff
@@ -53,126 +53,125 @@ guDateRange.directive('guDateRange', [function () {
...
-            $scope.save = setCustomRange;
-            $scope.setPresetDate = setPresetDate;
-            $scope.clearStart = clearStart;
-            $scope.clearEnd = clearEnd;
-
-            resetView();
+          ctrl.guDateFormat = ctrl.guDateFormat || 'DD-MMM-YYYY';
+          ctrl.guAnyTimeText = ctrl.guAnyTimeText || 'anytime';
+          ctrl.guFirstDay = ctrl.guFirstDay || 0;
+          var startInput = el[0].querySelectorAll('.gu-date-range__input__start--hidden')[0];
```
akash1810: minor: worth creating a variable for `el[0]` for DRYness?

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
shaundillon: Good point!

kahuna/public/js/components/gu-date-range/gu-date-range.js
shaundillon: Yep, sounds good to me. It's a pattern used in the main.js file too.

kahuna/public/js/components/gu-date-range/gu-date-range.js
```diff
@@ -53,126 +53,130 @@ guDateRange.directive('guDateRange', [function () {
...
-            $scope.save = setCustomRange;
-            $scope.setPresetDate = setPresetDate;
-            $scope.clearStart = clearStart;
-            $scope.clearEnd = clearEnd;
-
-            resetView();
+          ctrl.guDateFormat = ctrl.guDateFormat || 'DD-MMM-YYYY';
+          ctrl.guAnyTimeText = ctrl.guAnyTimeText || 'anytime';
+          ctrl.guFirstDay = ctrl.guFirstDay || 0;
+          const originalEl = el[0];
```
shaundillon: This is where the actual change starts, the rest is autoindent

PR #2132, opened 2018-05-08
run elasticsearch in docker in DEV

akash1810:
Also:
- add `dev-start.sh` script to quickly start ES and the Play apps
- update docs


dev-start.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
```
LATaylor-guardian: Can this script check that docker is installed on the dev's machine and advise how to install it if it is not?

E.g

```
type docker >& /dev/null || ( echo 'docker required' && exit 1 )
```

It could even go further and if it does not find it install docker itself.

LATaylor-guardian:
Apart from my comment 👍

dev-start.sh
akash1810: 👍 Updated now. We're now checking the requirements listed [here](https://github.com/guardian/grid/blob/master/docs/02.01-dev-setup.md#requirements)

PR #2133, opened 2018-05-09
run imgops in docker

akash1810:
Imgops is the image resizer service. Currently you have to manually start it, which is easy to forget. Its a simple nginx server, so lets just run it in docker instead to simplfy the dev process.


.gitignore
```diff
@@ -10,4 +10,5 @@ data
 .DS_Store
 *.py[cod]
 .venv
-kahuna/npm-debug.log
\ No newline at end of file
+kahuna/npm-debug.log
+imgops/dev/nginx.conf
```
akash1810: this file is generated in `dev-start.sh`

get-stack-resource.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
akash1810: this is actually a `git mv` plus some changes... not sure why GH thinks its a new file...

imgops/README.md
```diff
@@ -1,4 +1,4 @@
-# Img Opts for the Grid
+# ImgOps for the Grid
 Local version of the imgops service
 
 ## Requirements
```
shaundillon: We should add Docker to the requirements

akash1810: https://github.com/guardian/grid/blob/master/dev-start.sh#L27 😄

akash1810: updated the setup docs now too.

get-stack-resource.sh
```diff
@@ -0,0 +1,24 @@
...
+    echo "Please specify a stack resource."
+    exit 1
+fi
+
+if [ -z ${STACK_NAME} ];
+then
+    STACK_NAME=media-service-DEV
+fi
+
+JQ_FILTER="jq '.StackResources[] | select(.LogicalResourceId == \"$STACK_RESOURCE\") | .PhysicalResourceId'"
```
LATaylor-guardian: Does this script check that `jq` is installed and if not install it?

PR #2137, opened 2018-05-14
update dev docs

akash1810:
to reflect that we use docker, eliminating a few dependencies and setup steps


docs/02.01-dev-setup.md
```diff
@@ -4,24 +4,19 @@
 Ensure you have the following installed:
 - sbt
 - Java 8
+- npm
```
shaundillon: Might as well say Node, webpack needs it anyway

PR #2138, opened 2018-05-15
Introduce Tiers in Media API

marialivia16:
Introducing the concept of tiers with 2 possible values: 
- `Internal`: access to everything.
- `External`: access to `GET /images/*` endpoints. You only get back staff photographer images.

TODO:
- [x] Add tests
- [ ] Store keys encrypted in DynamoDB (separate PR)


common-lib/src/main/scala/com/gu/mediaservice/lib/auth/ApiKey.scala
```diff
@@ -0,0 +1,32 @@
...
+  }
+}
+
+case class ApiKey(value: String, tier: Tier)
+object ApiKey extends ArgoHelpers {
+  val unauthorizedResult = respondError(Forbidden, "forbidden", "Unauthorized - the API key is not allowed to perform this operation", List.empty)
+
+  def apply(content: String): ApiKey = {
+    val rows = content.split("\n")
+    val name = rows.head
```
LATaylor-guardian: If content is an empty String the call to `.head` will error.

media-api/app/controllers/SearchParams.scala
```diff
@@ -0,0 +1,161 @@
...
+
+object PayType extends Enumeration {
+  type PayType = Value
+  val Free = Value("free")
+  val MaybeFree = Value("maybe-free")
+  val All = Value("all")
+  val Pay = Value("pay")
+
+  def create(s: String) = s match {
+    case "free" => Some(Free)
```
LATaylor-guardian: Is it worth pulling these Strings out into private object members given we're relying on them being the same?

media-api/app/controllers/SearchParams.scala
```diff
@@ -0,0 +1,161 @@
...
+      commaSep("usageStatus"),
+      commaSep("usagePlatform"),
+      request.user.apiKey.tier
+    )
+  }
+
+
+  def toStringMap(searchParams: SearchParams): Map[String, String] =
+    Map(
+      "q"                 -> searchParams.query,
```
LATaylor-guardian: Same with these search param Strings.

LATaylor-guardian:
Great work @marialivia16!

media-api/app/controllers/SearchParams.scala
marialivia16: I'm going to leave this file as it is because it's not directly related to this work, I've just moved the SearchParam classes out of the controller file.

common-lib/src/main/scala/com/gu/mediaservice/lib/auth/ApiKey.scala
```diff
@@ -0,0 +1,32 @@
...
+
+  def apply(content: String): ApiKey = {
+    val rows = content.split("\n")
+    val name = rows.headOption.getOrElse("")
+    val tier = rows.tail.headOption.map(Tier(_)).getOrElse(Internal)
+    ApiKey(name, tier)
+  }
+
+  def hasAccess(apiKey: ApiKey, request: Request[Any]): Boolean = apiKey.tier match {
+    case External if request.method != "GET" || !request.path.startsWith("/images") => false //TODO: Also check domain to be api.media
```
akash1810: might be able to use [this val](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/config/Services.scala#L7)

build.sbt
```diff
@@ -126,4 +127,15 @@ def playProject(projectName: String, port: Int): Project =
...
       )
-    ))
\ No newline at end of file
+    ))
+
+val testSettings = Seq(
+  testOptions in Test += Tests.Setup(_ => {
+
+    println(s"Launching docker container with ES")
+    s"docker-compose up -d".!
```
akash1810: This'll run all the services defined [`docker-compose.yml`](https://github.com/guardian/grid/blob/master/docker-compose.yml). I don't think we'd need the imgops container in CI, so we could run `docker-compose up -d elasticsearch` instead to only start an elasticsearch container.

common-lib/src/main/scala/com/gu/mediaservice/lib/auth/ApiKey.scala
```diff
@@ -0,0 +1,32 @@
...
+case object External extends Tier
+object Tier {
+  def apply(value: String): Tier = value.toLowerCase match {
+    case "internal" => Internal
+    case "external" => External
+    case _ => External
+  }
+}
+
+case class ApiKey(value: String, tier: Tier)
```
akash1810: are we losing the name of the api key now?

marialivia16: `value` is set to the first line in the S3 file. Bad naming... I can change it to `name` if it makes more sense.

marialivia16:
[Support Read Only API access in the Grid for syndication partners.](https://trello.com/c/vcfbwcDj/14-support-read-only-api-access-in-the-grid-for-syndication-partners)

PR #2144, opened 2018-05-16
Disable CSRF as it is blocking print usage updates

mbarton:
```
[CSRF] Check failed because no token found in headers for /usages/print
```

We should investigate why CSRF was not just letting the request though (a la cURL)


mbarton:
Turned out this might have been a red herring. Closing in favour of #2145 for now

PR #2145, opened 2018-05-16
Kick off the usage streams

mbarton:
#2124 unfortunately removed the code that kicks off the usage streams meaning that both digital and print usages have not been recorded 😞 

This PR ports across the old code to start them (from [Global.scala](https://github.com/guardian/grid/blob/ca156b993007235fe8b3e1dd1d87b02d76809e10/usage/app/Global.scala)).

TODO:
- [x] test in CODE
- [x] test in PROD

### DIFF
--- a/usage/app/lib/UsageRecorder.scala
+++ b/usage/app/lib/UsageRecorder.scala
@@ -4,7 +4,7 @@ import model._
 import play.api.Logger
 import play.api.libs.json._
 import rx.lang.scala.subjects.PublishSubject
-import rx.lang.scala.{Observable, Subscriber}
+import rx.lang.scala.{Observable, Subscriber, Subscription}
 
 case class ResetException() extends Exception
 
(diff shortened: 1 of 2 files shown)


usage/app/lib/UsageRecorder.scala
```diff
@@ -14,8 +14,9 @@ class UsageRecorder(usageMetrics: UsageMetrics, usageTable: UsageTable, usageStr
   val liveUsageStream: Observable[UsageGroup] = usageStream.liveObservable.merge(usageSubject)
 
   val subscriber = Subscriber((_:Any) => Logger.debug(s"Sent Usage Notification"))
-  def subscribeToPreview  = previewObservable.subscribe(subscriber)
-  def subscribeToLive  = liveObservable.subscribe(subscriber)
+
+  var subscribeToPreview: Option[Subscription] = None
```
LATaylor-guardian: Why do these need to be `var`?

mbarton: We assign to them in the `start` method so we can call `unsubscribe` in the `stop` method.

I don't think there's a reason why we wouldn't just start them when they are constructed but the code prior to the upgrade kicked them off manually in `Global.scala` so I was keeping to the spirit of that code

LATaylor-guardian:
👍

PR #2148, opened 2018-05-18
log to disk in addition to stdout

akash1810:
Now that we're running all projects in a single command, filtering the logs from a specific app is impossible. Logging to disk allows us to see app specific logs under the logs directory of the app.


akash1810:
updated to only log to disk in DEV

PR #2151, opened 2018-05-22
Add missing url to Content Security Policy Directive.

LATaylor-guardian:


mbarton:
For posterity this is because a re-auth is done by iframing in Google which then redirects back to the Grids own OK end-point. Hence Kahuna needs to allow an iframe of itself!

PR #2152, opened 2018-05-22
Add an demo endpoint for Eyevine

marialivia16:
Adding a `/images/example` endpoint for demo purposes that always return the same image.

### DIFF
--- a/media-api/app/controllers/MediaApi.scala
+++ b/media-api/app/controllers/MediaApi.scala
@@ -4,7 +4,7 @@ import java.net.URI
 
 import com.gu.editorial.permissions.client.Permission
 import com.gu.mediaservice.lib.argo._
-import com.gu.mediaservice.lib.argo.model._
+import com.gu.mediaservice.lib.argo.model.{Action, _}
 import com.gu.mediaservice.lib.auth.Authentication.{AuthenticatedService, PandaUser, Principal}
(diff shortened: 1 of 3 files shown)


marialivia16:
[Create a demo endpoint for Eyevine to hit](https://trello.com/c/NeCD0g54/42-create-a-demo-endpoint-for-eyevine-to-hit)

media-api/conf/example-image.json
```diff
@@ -0,0 +1,160 @@
+{
```
akash1810: any reason why we're not fetching this from elasticsearch?

marialivia16: The idea was to have the same image in all environments without having to worry whether it exists in ES or whether the ids are different for example.

LATaylor-guardian:
👍 If this is representative of what we expect Eyevine to receive then let's get this merged and over to them so they can begin the work on their side 🚢

media-api/conf/example-image.json
akash1810: makes sense 👍 however, eyevine would need to be able to download the image to fully test the process their end. Could we return data for a real image, with links to our cdn etc?

marialivia16: Will do, makes sense.

PR #2155, opened 2018-05-24
Log API keys to graph usage

marialivia16:
This PR aims at auditing the api key we have, in order to delete the unused ones and attribute them the correct tier.


common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Authentication.scala
```diff
@@ -52,6 +54,9 @@ class Authentication(config: CommonConfig, actorSystem: ActorSystem,
       case Some(key) =>
         keyStore.lookupIdentity(key) match {
           case Some(apiKey) =>
+            val marker: Marker = MarkerFactory.getMarker(key)
+            val mc: MarkerContext = MarkerContext(marker)
+            Logger.info(s"Using api key $key with name ${apiKey.name} and tier ${apiKey.tier}")(mc)
```
akash1810: Lets not log the full key 🙈. Would the name be enough?

marialivia16: 👍 I got too excited by this, name will be enough, just a bit harder to associate with a key.

Issue #418, opened 2015-02-04
Paid for images not available when you start search from a paid one


paperboyo:
Hello,

May be expected behaviour, if yes - then disregard, please.

Steps to reproduce:
- go to Grid
- uncheck 'free to use'
- click on any photo with a 'pay' indicator
- click on any metadata link, like a photographer's name

Result:
The results shown thereafter exclude paid for photos (as per default Grid state), so that even the photo you were searching from is not displayed.

Expected result:
Switch 'free tu use' filter OFF when searching from a paid photograph. The use case would be - these kind of searches happen often when one wants to see more photos from the same shoot.

Regards
Mateusz

PR #2156, opened 2018-05-25
Allows editors to see paid for images by default

shaundillon:
https://trello.com/c/TkHOfzRs/49-allow-picture-editors-to-see-paid-for-and-free-images-by-default

What it says up there ^ Requires work in Permissions too


auth/app/auth/AuthController.scala
```diff
@@ -36,14 +39,23 @@ class AuthController(auth: Authentication, config: AuthConfig,
     val firstName = user.firstName
     val lastName = user.lastName
 
+    val showPaidFuture: Future[Boolean] = hasPermission(PandaUser(request.user), Permissions.ShowPaid)
+
+
+    val showPaid: Boolean = Await.result(showPaidFuture, 100.millis)
```
shaundillon: I'm pretty sure this is gross and horrible but I'm not sure how else I can do it, and the result _does_ rely on it after all

paperboyo:
Possibly see also https://github.com/guardian/grid/issues/418 (?).

shaundillon:
@paperboyo That's related but the fix is different, that requires maintaining the search query between pages.

auth/app/auth/AuthController.scala
mbarton: Use `auth.AuthAction.async` in `def index` below and then restructure the index response to map over the future:

```scala
hasPermission(PandaUser(request.user), Permissions.ShowPaid).map { canSeePaid =>
   respond(
      ... blah blah ...
   )
}
```

tbh we could just copy the whole lot into the method, no need for it to be val

kahuna/public/js/search/query.js
```diff
@@ -162,6 +162,8 @@ query.controller('SearchQueryCtrl',
     mediaApi.getSession().then(session => {
         ctrl.user = session.user;
         ctrl.filter.uploadedByMe = ctrl.uploadedBy === ctrl.user.email;
+        ctrl.filter.nonFree = session.user.permissions.showPaid ?
+          session.user.permissions.showPaid : undefined;
```
akash1810: could this just be `!!session.user.permissions.showPaid`?

PR #2157, opened 2018-05-29
Fix silly cast to BoxedUnit exception

mbarton:
An upgrade to Scanamo changed the return type of `deleteItem` from `Unit` to an AWS SDK `DeleteItemOutcome`, hence the following call to `mapTo[Unit]` failed.

This meant users deleting a collection were shown an error message even though the delete worked just fine. The solution is to simply avoid `mapTo` and the compiler works everything out.


akash1810:
[Grid: Cannot delete collections](https://trello.com/c/NYmtqdUn/415-grid-cannot-delete-collections)

PR #2160, opened 2018-05-30
add ability to create collections on upload

shaundillon:
https://trello.com/c/wihWF6ON/55-image-upload-process-allow-collection-creation-on-upload

Far simpler than I imaged 🎉 

## Before
collectionbefore

## After
collectionafter

### DIFF
--- a/kahuna/public/js/components/gr-collection-overlay/gr-collection-overlay.css
+++ b/kahuna/public/js/components/gr-collection-overlay/gr-collection-overlay.css
@@ -40,6 +40,13 @@
     width: 365px;
     background: #444;
     z-index: 10;
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+}
+
+.collection-overlay__collections__heading-text {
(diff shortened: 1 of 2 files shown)


kahuna/public/js/components/gr-collection-overlay/gr-collection-overlay.css
```diff
@@ -40,6 +40,13 @@
     width: 365px;
     background: #444;
     z-index: 10;
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+}
+
+.collection-overlay__collections__heading-text {
+  margin: 0 10px 0 0;
```
akash1810: super minor: `margin-left` to be slightly less verbose

shaundillon: It's a `<p>` tag which comes with default browser styles since we're not using a rest 😢

shaundillon:
Tested on CODE

PR #2163, opened 2018-06-05
Save Syndication Rights

marialivia16:
- Updating mappings to include the syndication rights fields
- Thrall picks up rights coming from the RCS feeds and saves them in ES (also updates the `lastModified` date)
- Showing new fields in the media API


akash1810:
Great work!

PR #2167, opened 2018-06-06
include angular-hotkeys css to render the modal

akash1810:
We're using the angular-hotkeys module for keyboard shortcuts. We need to include their css for the help modal to render correctly (it currently doesn't render).

I suspect this broke when we moved to webpack. See https://github.com/guardian/grid/pull/1552#issuecomment-165063633.

Not entirely convinced this is webpack best practice either. Maybe [resolve](https://webpack.js.org/configuration/resolve/#resolve-alias) is better? cc @RichieAHB 

# After
Pressing `shift + ?` shows:
image


akash1810:
Discussed with @RichieAHB and @jonathonherbert IRL, we've come to agree that using resolve doesn't provide the explicitness of this version. Equally, the preference towards either solution aren't strong enough to change this.

Merging.

PR #2170, opened 2018-06-08
Update mappings process

marialivia16:


scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
```diff
@@ -211,6 +207,38 @@ object UpdateMapping extends EsScript {
   }
 }
 
+object GetMapping extends EsScript {
+
+  def run(esHost: String, extraArgs: List[String]) {
+    // TODO: add the ability to update a section of the mapping
```
akash1810: copy pasted TODO?

scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
```diff
@@ -211,6 +207,38 @@ object UpdateMapping extends EsScript {
...
+
+        client.close()
+      }
+    }
+
+    EsClient.getMappings(extraArgs.headOption)
+  }
+
+  def usageError: Nothing = {
+    System.err.println("Usage: UpdateMapping <ES_HOST>")
```
akash1810: should this be `GetMapping`?

scripts/README.md
```diff
@@ -21,14 +21,19 @@ Optionally takes a DateTime string as a second argument. Will perform reindex fo
 ### UpdateMapping
 When you add a mapping e.g. You add a new field to the [image mapping](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala#L73)
 you should add the mapping with this script as we are using [`strict`](http://www.elasticsearch.org/guide/en/elasticsearch/guide/current/dynamic-mapping.html)
-mappings (you cannot just add things willy nilly).
+mappings (you cannot just add things willy nilly). Updating mappings is done in 2 steps:
```
LATaylor-guardian: Can we have a single script do this rather than two seperate actions?

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchClient.scala
```diff
@@ -24,7 +24,7 @@ trait ElasticSearchClient {
   private lazy val settings: Settings =
     ImmutableSettings.settingsBuilder
       .put("cluster.name", cluster)
-      .put("client.transport.sniff", true)
+      .put("client.transport.sniff", false)
```
LATaylor-guardian: Is this something that should only be applied when we are updating mappings?

scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
```diff
@@ -211,6 +207,38 @@ object UpdateMapping extends EsScript {
...
+      val cluster = esCluster
+
+      def getMappings(specifiedIndex: Option[String]) {
+        val index = getCurrentAlias.getOrElse(imagesAlias)
+        println(s"getting mapping on index: $index")
+        val result = client.admin.indices
+          .prepareGetMappings(index)
+          .execute.actionGet
+
+        println(result.getMappings.get(index).get(imageType).getSourceAsMap)
```
LATaylor-guardian: Could the code inside this println throw an exception  if the ES request above fails?

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchClient.scala
marialivia16: It doesn't look like it's affecting the other scripts.

PR #2171, opened 2018-06-11
Encode path when delete

marialivia16:
We have a collection that contains a `+` character in the path which means it can't be deleted.

The reason is because the `+` will be decoded into a space and when we look for it in DynamoDb it won't match anything.

e.g:
Given collection `Sport/gsport/AL+Cricket` with the DynamoDb entry `sport/gsport/al+cricket` => it'll be transformed into `sport/gsport/al cricket` => nothing will be deleted as it doesn't exist.


akash1810:
👍

PR #2172, opened 2018-06-11
Replace mixpanel with Google analytics

shaundillon:
This replaces mixpanel with the same kind of analytics integration used in our other apps


kahuna/app/KahunaComponents.scala
```diff
@@ -33,15 +33,15 @@ object KahunaSecurityConfig {
 
     val frameSources = s"frame-src ${config.services.authBaseUri} ${config.services.kahunaBaseUri} https://accounts.google.com"
     val frameAncestors = s"frame-ancestors ${config.services.toolsDomains.map(domain => s"*.$domain").mkString(" ")}"
-    val connectSources = s"connect-src ${(services :+ config.imageOrigin).mkString(" ")} 'self'"
+    val connectSources = s"connect-src ${(services :+ config.imageOrigin).mkString(" ")} 'self' www.google-analytics.com"
 
-    val imageSources = s"img-src data:* ${config.services.imgopsBaseUri} https://${config.fullOrigin} https://${config.thumbOrigin} ${config.cropOrigin} 'self'"
+    val imageSources = s"img-src data:* ${config.services.imgopsBaseUri} https://${config.fullOrigin} https://${config.thumbOrigin} ${config.cropOrigin} www.google-analytics.com 'self'"
```
akash1810: Why does GA need `img-src`? Tracking pixels?

kahuna/app/views/main.scala.html
```diff
@@ -2,7 +2,8 @@
   authApiUri: String,
   reauthUri: String,
   sentryDsn: Option[String],
-  sessionId: String)
+  sessionId: String,
+  googleTrackingId: String)
```
akash1810: Can we make this an `Option[String]`? That way we can avoid the empty string comparison below.

kahuna/public/js/analytics/analytics.js
```diff
@@ -0,0 +1,49 @@
...
+            ((...args) => (window.ga.q = window.ga.q || []).push(args));
+
+        const { ga } = window;
+        ga("create", gaId, "auto");
+        ga("set", "transport", "beacon");
+        ga("send", "pageview");
+
+        return ga;
+    }
+    return (...args) => window.debugGA && console.log(...args);
```
akash1810: Is the linter happy with `console.log`? It shouldn't be!

PR #2174, opened 2018-06-11
show syndication status in search results

akash1810:
display an icon if we have acquired syndication rights (the bottom image in screenshot)

image

### DIFF
--- a/kahuna/public/js/services/image-logic.js
+++ b/kahuna/public/js/services/image-logic.js
@@ -61,12 +61,25 @@ imageLogic.factory('imageLogic', ['imageAccessor', function(imageAccessor) {
         });
     }
 
+    function isSyndicated(image) {
+        const syndicationRights = image.data.syndicationRights;
+
+        const hasSyndicationRights = !!syndicationRights;
+
+        if (!hasSyndicationRights) {
+            return false;
+        }
+
+        return syndicationRights.rights.find(_ => _.acquired === true);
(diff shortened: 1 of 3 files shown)


kahuna/public/js/services/image-logic.js
```diff
@@ -61,12 +61,25 @@ imageLogic.factory('imageLogic', ['imageAccessor', function(imageAccessor) {
...
+    function isSyndicated(image) {
+        const syndicationRights = image.data.syndicationRights;
+
+        const hasSyndicationRights = !!syndicationRights;
+
+        if (!hasSyndicationRights) {
+            return false;
+        }
+
+        return syndicationRights.rights.find(_ => _.acquired === true);
```
akash1810: `rights` is a list, `isSyndicated` is true if at least one of the rights has been acquired

LATaylor-guardian:
Have we checked with users that they are clear what this icon is and means?

LATaylor-guardian:
👍

akash1810:
> Have we checked with users that they are clear what this icon is and means?

No. However, this feature is dependant on the RCS feed polling and is also a blocker for other feature work:
- Surfacing rights acquired images via the API
- Surfacing rights acquired images via a query param

Given this field won't get populated until we start polling RCS, I was planning on shipping the above and then iterate on the design later as we should then have real examples to demo/discuss.

PR #2175, opened 2018-06-12
add `hasRightsAcquired` filter to media api

akash1810:
`hasRightsAcquired` is an optional parameter:
- `true` returns images that have rights acquired
- `false` returns images that do not have rights acquired

Note: setting this value will only return images that have been processed by the RCS feed.


media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -99,12 +99,28 @@ class ElasticSearch(config: MediaApiConfig, searchFilters: SearchFilters, mediaA
...
+      ++ simpleCostFilter
+      ++ costFilter
+      ++ hasExports
+      ++ hasIdentifier
+      ++ missingIdentifier
+      ++ dateFilter
+      ++ usageFilter
+      ++ hasRightsCategory
+      ++ searchFilters.tierFilter(params.tier)
+      ++ rightsAcquiredFilter
```
akash1810: this line is the main change here... the other changes are new lines

LATaylor-guardian:
👍

PR #2177, opened 2018-06-12
Encode space as %20

marialivia16:
Both the `+` character and the space character reached the controller as `+` creating confusion when decoding.

This ensures space is encoded as `%20` and `+` as `%2B`.


LATaylor-guardian:
Nice 👍

LATaylor-guardian:
Nice 👍

Issue #1854, opened 2016-04-13
Determine if "viewed up to" is used


kenoir:
And remove it if not.

akash1810:
GA stats indicates it is being used. Closing this issue.

paperboyo:
Hmm, in the last month it’s 0 uses on some days and up to ~20 uses on other days. Could this be one user or two? I would love to see this feature being used, because it never made any sense to me!

PR #2181, opened 2018-06-18
Track suspected barely/never used features

shaundillon:
It's strongly suspected that the 'viewed up to' (https://github.com/guardian/grid/issues/1854) and 'Show Preview' actions are never used.

In the interest of thoroughness, I've added tracking to these actions so we can make a decision about killing them off.

Blocked by #2172


akash1810:
Last time we tried to kill `viewed up to` a few people complained! Using data to drive this is a god way to go!

Issue #1508, opened 2015-11-27
Better reliability and performance with selections


paperboyo:
Hello,

Something you probably are aware of, just wanted to register it as an issue(s):
1. Selections are unreliable when you haven't _seen_ all images in between the first and last. To test, select an image, drag a slider quickly (but not too far, see pt. 2) and shift+select the last image. Then slowly drag past all the images in the selection and notice that, actually, not all are selected. Deselected are the ones that are just showing up now in front of you.
2. Selections/deselections are very slow. Here are some poor-man's stats:

Latest Chrome on MacPro: 140 images: 40sec and one dialog box (unresponsive site). Deselection 1s.
Latest Firefox on MacPro: 258 images: **5min 20sec** and **24** dialog boxes (unresponsive script). Deselection 3s.

I do think it's not a priority. Especially, no. 2. But no. 1 may bring a very unpleasant surprises. Even more so, if they get unnoticed (like with batch metadata editing).

Regards
Mateusz

NickPapacostas:
![](i.imgur.com/IZG0gNr.gif)

paperboyo:
3. Some operations on selections are unreliable, e.g. downloading a ZIP (contains just a random subset of selected photos), adding to library (“Failed to add to library. Please try again.”).

paperboyo:
This is much better, thanks mainly to https://github.com/guardian/grid/pull/2905 and https://github.com/guardian/grid/pull/2917. Next step would require using bulk APIs…

PR #2183, opened 2018-06-19
Batch full frame crop

shaundillon:
This moves us a step closer to 1 click galleries

## Improvements
- Detect large selections (Separate PR)

fullframe2

### DIFF
--- a/kahuna/public/js/main.js
+++ b/kahuna/public/js/main.js
@@ -4,8 +4,6 @@ import angular from 'angular';
 import 'angular-ui-router';
 import {heal} from 'pandular';
 
-import uriTemplates from 'uri-templates';
-
 import {cropperApi} from './services/api/media-cropper';
 import {editsApi}   from './services/api/edits-api';
 import {loaderApi}  from './services/api/loader';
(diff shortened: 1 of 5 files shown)


paperboyo:
Woooo! 👏 

Does it:
- hide button if all images in a selection already have full frames? (but still shows it when at least one doesn’t)
- hide progress indicator (...) when all crops are done?

Here I’m not sure: because Clear selection has no keyboard shortcut (should be ESC?), and because Create full frames is so close to it – should Create full frames be a two-step button for safety? (just thinking about someone using it by mistake on a [large selection](https://github.com/guardian/grid/issues/1508).

Overall: bombastic and indeed a (big) step closer.

shaundillon:
> Does it:

>hide button if all images in a selection already have full frames? (but still shows it when at least one doesn’t)
Nope, good shout. I'll see if I can add it to this PR.

>hide progress indicator (...) when all crops are done?
Yep!

paperboyo:
> Detect large selections (Separate PR)

Limit would be entirely arbitrary, but given the [very real and dangerous pains](https://github.com/guardian/grid/issues/1508), maybe it should apply to **all** selections/operations?

Currently if thumbs haven’t shown themselves while making a selection, images won be selected! And then, operations done on that large selection won’t only be not applied to these missing images but also, often, fail partially. This leaves the systems in a state where one cannot know which part of the selection the operation was applied to (e.g. a downloaded zip will contain just a random selection of images from your initial selection; much more problematic for batch metadata operations).

shaundillon:
@paperboyo I could add an arbitrary limit to this, but applying it to *all* selections is a much bigger job

paperboyo:
Alrighty, no worries!

kahuna/public/js/main.js
```diff
@@ -220,10 +218,24 @@ kahuna.run(['$rootScope', 'httpErrors',
             function($rootScope, httpErrors) {
 
     $rootScope.$on('events:error:unauthorised', () =>
-        $rootScope.$emit('track:event', 'Authentication', null, 'Error', null, { 'Error code': httpErrors.unauthorised.errorCode }));
+        $rootScope.$emit(
```
akash1810: minor: this could be its own pr...

shaundillon: Linting wouldn't pass without it :(.

It's a separate commit at least, so it should be easy to unpick if needed?

akash1810: Sorry, I hadn't realised it was a whitespace change! Ignore me entirely!

PR #2186, opened 2018-06-21
add `hasCrops` as a query string to Grid

akash1810:
The API already has a `hasExports` query, so this just exposes that to Kahuna... with a more obvious name...


shaundillon:
Yay Linting!

PR #2187, opened 2018-06-21
🍟🍟🍟🍟🍟🍟 hasCrops chip 🍟🍟🍟🍟🍟

akash1810:
image

### DIFF
--- a/kahuna/public/js/search/structured-query/query-suggestions.js
+++ b/kahuna/public/js/search/structured-query/query-suggestions.js
@@ -28,7 +28,8 @@ export const filterFields = [
     'usages@<added',
     'usages@>added',
     'usages@platform',
-    'usages@status'
+    'usages@status',
+    'hasCrops'
 ];
 // TODO: add date fields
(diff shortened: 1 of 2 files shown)


akash1810:
Ideally, the presence of `hasCrops` would indicate `true`, similar to html attributes, but that's a bigger change...

shaundillon:
It looks weird in camelcase

paperboyo:
Oh, oh, oh... true/false just repeats +/- so is redundant.

Would it not be betterer to has a chip `has:` with a value `crops`?
We would then one day extend it to cover other values, like e.g. `leases`, `icc profile`, `transparency` etc.

akash1810:
> Would it not be betterer to has a chip has: with a value crops?

Completely agree, but that's a bigger bit of work, which we can do afterwards.

> Oh, oh, oh... true/false just repeats +/- so is redundant.

👍 https://github.com/guardian/grid/pull/2187#issuecomment-399160862

PR #2180, opened 2018-06-15
Metadata-extractor lib upgrade

marialivia16:
This upgrades the metadata-extractor lib and saves all the new xmp metadata.

As discussed with @paperboyo and @akash1810 

NEEDS TESTING!

PR #2190, opened 2018-06-25
introduce `has` query to api

akash1810:
Support the ability to query the api for images that have a field, for example:
- `has:crops` for images that have been cropped
- `has:rightsSyndication` for images that have RCS information
- `has:fileMetadata.xmp` for images that have xmp metadata

NB A cropped image is saved with `exports` in ES. We allow this field to be queried as `crops` for simplicity.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ImageFields.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ImageFields.scala
@@ -49,6 +49,7 @@ trait ImageFields {
     case f if editsFields.contains(f)       => editsField(f)
     case f if collectionsFields.contains(f) => collectionsField(f)
     case f if usagesFields.contains(f)      => usagesField(f)
+    case "crops" => "exports"
     case f => f
   }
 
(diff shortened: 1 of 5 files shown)


paperboyo:
💎!
Can this go all the way, like
`has:fileMetadata.xmp:photoshop:SupplementalCategories[2]:"Housing"` in [here](https://api.media.test.dev-gutools.co.uk/images/5b227ffee05b7b6c5ea31457c4bae34ca8aa6c1a/fileMetadata)?
(`[2]` is an array indicator, so maybe should be dropped?)

akash1810:
> Can this go all the way, like
`has:fileMetadata.xmp:photoshop:SupplementalCategories[2]`

I'd suspect it would struggle with a `:` delimiter. Confirming is difficult because these fields are only extracted in #2180... Will investigate.

paperboyo:
> I'd suspect it would struggle with a : delimiter.

Oh, I don’t care about `:` specifically! This is what is there when you look, but maybe for consistency we prefer `.`?



> I think we'd need to adjust the way we analyse these fields if we wanted to do this, this feels out of scope for this PR.

Cool. We want (I want), but not now, will keep it at the back of my head. Pretty swollen back of my head :)

akash1810:
> Oh, I don’t care about `:` specifically! This is what is there when you look, but maybe for consistency we prefer `.`?

We're just extract the xmp block as is atm. If we translating it into a more structured form, I'd argue we should add it to `metadata` so its indexed and searchable like other fields.

paperboyo:
For the record, discussed IRL, proper reconcilliation and adding to 1st order metadata all metadata is desirable but out of scope for now.

First, we should better reconcile the fields we already are bringing over, byline, credit, subjects, keywords etc. This will fix things like [truncated bylines](https://github.com/guardian/grid/issues/1117#issue-102139203) or [poor, misspelled location](https://github.com/guardian/grid/issues/1117#issuecomment-149366878). Only afterwards we should think about bringing more data over to 1st order.

1st order sounds a bit starwarsy...

PR #2193, opened 2018-06-26
Add Shoot ID

shaundillon:
Adds "Shoot ID" to images, allowing them to be grouped by photoshoot

shootidimagepage
shootidpanel
shootidupload


kahuna/.htmllintrc
```diff
@@ -28,7 +28,7 @@
     // our identation is a bit more complex than this, so don't check it
     "indent-width": false,
     "label-req-for": true,
-    "line-end-style": "lf",
+    "line-end-style": false,
```
shaundillon: I added this because git was auto CLRF-ing my files. I did some reading and apparently this is recommended if you have users on different systems.

shaundillon:
@akash1810 I'll put it on TEST now. Need to check over it properly as the grid doesn't run properly on my home computer

shaundillon:
@paperboyo I agree re: the ordering. It's just that the current format of those fields and how they're edited best matched Shoot ID.

akash1810:
Closing as this is replaced in #2241, #2242 and #2243 to fit the new `album` model introduced in #2222.

PR #2195, opened 2018-06-27
class anti forgery mismatch as 4XX

akash1810:
Currently, as we're not catching this exception explicitly, it's bubbling up as a 5XX and tripping our alarms. A 4XX response is more reasonable.


LATaylor-guardian:
👍

PR #2204, opened 2018-07-02
add `croppedBy` search chip

akash1810:
To find images that have bee cropped by a specific person

Not overly pleased with the camel cased `croppedBy` in the UI, but supporting `cropped by` isn't trivial, so let's get something shipped in the mean time.

Also, `croppedBy:me` would be really nice to save you typing your email address... but again, not trivial.


paperboyo:
👏 

> croppedBy:me would be really nice

+1, but only as a general `me` helper working among more 🍟

PR #2209, opened 2018-07-03
log image downloads

akash1810:
Depends on https://github.com/guardian/grid/pull/2208.

Currently, the API returns an S3 signed URL which can be used to download an image. Although this works, we don't know who is downloading what.

This PR wraps this S3 signed URL in a new `/download` route which logs and writes to CloudWatch Metrics and then redirects.

A CW Metric can have a maximum of 10 dimensions, so group non Syndication tier users by tier and then split each Syndication user.

Example graph that results:
image


akash1810:
https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/GeneratePresignedUrlRequest.html#addRequestParameter-java.lang.String-java.lang.String- may be a better solution.

PR #2215, opened 2018-07-09
Schedule updates for quota and usage stores

mbarton:
The calls to update the usage and quota stores were removed, probably in #2124.

This PR adds them back and simplifies the code by removing a class that was only there to hold references to the stores.


mbarton:
Tested in TEST

PR #1115, opened 2015-08-20
Unify save/cancel button styles.

akash1810:
Adding labels to the save and cancel buttons of the usage rights editor form.

Also removing some duplicated css :tada:.

button

PR #2219, opened 2018-07-11
edit metadata inline in kahuna

akash1810:
We seem to have introduced a regression somewhere as we were previously doing this (see the gif in https://github.com/guardian/grid/pull/1115)

Also add some spacing between elements because its all too tight atm.

# Before
before

# After
after

### DIFF
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
@@ -42,6 +42,7 @@
                 <dt class="metadata-line metadata-line__key">Title</dt>
                 <dd class="image-info__title metadata-line__info"
                     editable-text="ctrl.metadata.title"
+                    ng:hide="titleEditForm.$visible"
                     onbeforesave="ctrl.updateMetadataField('title', $data)"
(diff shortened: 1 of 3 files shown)


kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
```diff
@@ -143,10 +148,16 @@
 
             <dt ng:if="ctrl.hasLocationInformation" class="image-info__group--dl__key metadata-line__key image-info__wrap">Location</dt>
             <dd ng:if="ctrl.hasLocationInformation" class="image-info__group--dl__value metadata-line__info">
-                        <span ng:repeat="prop in ['subLocation', 'city', 'state', 'country']" ng:if="ctrl.metadata[prop]">
-                            <span class="metadata-line__info">
-                                <a ui:sref="search.results({query: (ctrl.metadata[prop] | queryFilter:ctrl.locationFieldMap[prop])})">{{ctrl.metadata[prop]}}</a></span><span ng:if="! $last">,</span>
-                        </span>
+                <span ng:repeat="prop in ['subLocation', 'city', 'state', 'country']" ng:if="ctrl.metadata[prop]">
```
akash1810: whitespace changes

PR #2220, opened 2018-07-12
Usage: upgrade Kinesis Client Library

marialivia16:
The Kinesis Client Library version we were using was not compatible with AWS SDK 2.11.


LATaylor-guardian:
👍

PR #2224, opened 2018-07-13
Updates to the logback file

marialivia16:
- Only use `application.home` from `logback.xml` when running in DEV.
- Only log to file.

### DIFF
--- a/common-lib/src/main/resources/logback.xml
+++ b/common-lib/src/main/resources/logback.xml
@@ -2,11 +2,20 @@
 
     <conversionRule conversionWord="coloredLevel" converterClass="play.api.libs.logback.ColoredLevel" />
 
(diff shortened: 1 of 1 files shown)


common-lib/src/main/resources/logback.xml
```diff
@@ -2,11 +2,20 @@
 
     <conversionRule conversionWord="coloredLevel" converterClass="play.api.libs.logback.ColoredLevel" />
 
+    <if condition='property("STAGE").contains("DEV")'>
```
LATaylor-guardian: Haven't seen this before - where does property come from and how is it able to access `STAGE`?

marialivia16: It's defined here: https://github.com/guardian/grid/blob/master/.sbtopts#L3

PR #2228, opened 2018-07-18
persistable collections

akash1810:
Add configuration for a list of persistable collections. Images in one of these collections will be retained forever.

The primary use-case is for Archive. Currently, they're adding images to the `GNM Archive` collection and adding the `gnmarchive` label and finally manually archive the images. A lot of work! Automating the persistence feels like a quick win.

This implementation isn't brilliant as its putting collection knowledge inside the media-api rather than the collections api. I've done this as the cheapest thing to do.

The correct way would be to update the collections model to include a flag and let this flag be set client-side. However that's quite a bit more work because collections are denormalised. If we get requests to persist more and more collections, then we could revisit this (maybe all collections should be persisted by default?).

### DIFF
--- a/media-api/app/lib/ImageResponse.scala
+++ b/media-api/app/lib/ImageResponse.scala
@@ -52,6 +52,14 @@ class ImageResponse(config: MediaApiConfig, s3Client: S3Client, usageQuota: Usag
   def hasLeases(image: Image) =
     image.leases.leases.nonEmpty
 
+  def isInPersistedCollection(image: Image): Boolean = {
+    // list of the first element of each collection's `path`, i.e all the root collections
+    val collectionPaths: List[String] = image.collections.map(_.path.head)
+
+    // is image in at least one persisted collection?
+    (collectionPaths diff config.persistedRootCollections).length < collectionPaths.length
+  }
+
   def isPhotographerCategory[T <: UsageRights](usageRights: T) =
     usageRights match {
       case _:Photographer => true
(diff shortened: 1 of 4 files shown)


media-api/app/lib/elasticsearch/SearchFilters.scala
```diff
@@ -66,7 +66,8 @@ class SearchFilters(config: MediaApiConfig) extends ImageFields {
     filters.bool.must(filters.existsOrMissing("usages", exists = true)),
     filters.exists(NonEmptyList(identifierField(config.persistenceIdentifier))),
     filters.bool.must(filters.boolTerm(editsField("archived"), value = true)),
-    filters.bool.must(filters.terms(usageRightsField("category"), persistedCategories))
+    filters.bool.must(filters.terms(usageRightsField("category"), persistedCategories)),
+    filters.bool.must(filters.terms(collectionsField("path"), config.persistedRootCollections.toNel.get))
```
akash1810: `.toNel.get` is safe as `config.persistedRootCollections` will always have a value

media-api/app/lib/ImageResponse.scala
```diff
@@ -52,6 +52,14 @@ class ImageResponse(config: MediaApiConfig, s3Client: S3Client, usageQuota: Usag
   def hasLeases(image: Image) =
     image.leases.leases.nonEmpty
 
+  def isInPersistedCollection(image: Image): Boolean = {
+    // list of the first element of each collection's `path`, i.e all the root collections
+    val collectionPaths: List[String] = image.collections.map(_.path.head)
```
LATaylor-guardian: This is using `.head`.

It would be safer to rewrite this as: 

```
val collectionPaths: List[String] = image.collections.flatMap(_.path.headOption)
```

akash1810: 👍 https://github.com/guardian/grid/pull/2234

PR #2233, opened 2018-07-21
the start of more structured logging with markers 

akash1810:
Introduce `GridLogger` to log with markers and move some logging to it.

### DIFF
--- a/usage/app/lib/UsageConfig.scala
+++ b/usage/app/lib/UsageConfig.scala
@@ -62,9 +62,9 @@ class UsageConfig(override val configuration: Configuration) extends CommonConfi
 
   private val iamClient: AmazonIdentityManagement = withAWSCredentials(AmazonIdentityManagementClientBuilder.standard()).build()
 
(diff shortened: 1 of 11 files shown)


usage/app/lib/UsageConfig.scala
```diff
@@ -62,9 +62,9 @@ class UsageConfig(override val configuration: Configuration) extends CommonConfi
 
   private val iamClient: AmazonIdentityManagement = withAWSCredentials(AmazonIdentityManagementClientBuilder.standard()).build()
 
-  val postfix: String = if (stage == "DEV") {
+  val postfix: String = if (isDev) {
     try {
-      iamClient.getUser().getUser().getUserName()
+      iamClient.getUser.getUser.getUserName
```
LATaylor-guardian: At least some of these brackets (i.e the one denoting the side effect of an http request) are needed.

LATaylor-guardian:
👍

PR #2237, opened 2018-07-23
Syndication lease UI

marialivia16:
screen shot 2018-07-23 at 11 37 49
screen shot 2018-07-23 at 11 38 03
screen shot 2018-07-23 at 12 35 54


paperboyo:
Minor: in the last screenshot the UI text should read “Allow syndication” instead of “allow-syndication”, I think.

kahuna/public/js/leases/leases.js
```diff
@@ -49,14 +50,26 @@ leases.controller('LeasesCtrl', [
                 ctrl.adding = true;
                 ctrl.newLease.createdAt = new Date();
                 ctrl.newLease.access = ctrl.access;
+                if (!ctrl.showCalendar && ctrl.access === 'deny-syndication') {
+                    ctrl.newLease.startDate = null;
+                    ctrl.newLease.endDate = null;
+                }
+
+                let syndLeases = ctrl.leases.leases.filter((l) =>
```
LATaylor-guardian: Can `ctrl.leases` ever be null or undefined here? If so this will error and probably needs a guard.

kahuna/public/js/leases/leases.js
```diff
@@ -49,14 +50,26 @@ leases.controller('LeasesCtrl', [
...
+                    l.access.endsWith('-syndication')
+                );
 
-                leaseService.batchAdd(ctrl.newLease, ctrl.leases, ctrl.images)
-                    .catch(() =>
-                        alertFailed('Something went wrong when saving, please try again.')
+                if (ctrl.access.endsWith('-syndication') && syndLeases.length > 0) {
```
LATaylor-guardian: As above.

LATaylor-guardian:
👍

kahuna/public/js/leases/leases.html
```diff
@@ -31,10 +31,24 @@
...
   </select>
+
+  <div ng:if="ctrl.access === 'deny-syndication'">
+    <input type="checkbox"
+           ng:model="ctrl.showCalendar" />
+    Expires
+  </div>
+
   <gu-date-range class="lease__date"
+     ng:if="ctrl.access !== 'allow-syndication' && !(ctrl.access === 'deny-syndication' && ctrl.showCalendar === false)"
```
akash1810: minor: its better to push this down to the controller to keep the template simple

e.g.
```js
ctrl.foo = ctrl.access !== 'allow-syndication' && !(ctrl.access === 'deny-syndication' && ctrl.showCalendar === false)
```

```html
<gu-date-range ng:if="ctrl.foo">
```

kahuna/public/js/leases/leases.js
marialivia16: It seems to always be set, we never check, see: https://github.com/guardian/grid/blob/a21301fbffad7005f525bb9a456c0459867a18c9/kahuna/public/js/leases/leases.js#L87

kahuna/public/js/leases/leases.css
```diff
@@ -133,8 +138,13 @@ gr-leases .gu-date-range__display {
 gr-leases.leases-flex {
   display: flex;
   flex-direction: row;
+  width: 100%;
 }
 
+.result-editor__field-container__leases {
```
akash1810: minor: could we [reuse existing](https://github.com/guardian/grid/blob/master/kahuna/public/stylesheets/main.css#L327) classes?

kahuna/public/js/leases/leases.css
```diff
@@ -4,14 +4,19 @@
     border: 1px solid #999;
     padding: 5px;
     box-sizing: border-box;
+    width: 100%;
```
akash1810: minor: could we [reuse existing](https://github.com/guardian/grid/blob/master/kahuna/public/stylesheets/main.css#L310) classes?

PR #2238, opened 2018-07-23
run album suggestion script iff there is an album

akash1810:
related to https://github.com/guardian/grid/pull/2236


LATaylor-guardian:
👍

PR #2242, opened 2018-07-24
allow multiple images to be added to album via multi-selection

akash1810:
extends https://github.com/guardian/grid/pull/2241 to the search page

multi-album


paperboyo:
I wonder what should the behaviour/warnings be if one’s selection contains >1 image that already is a member of another album?

PR #2243, opened 2018-07-25
allow albums to be set on uploads page

akash1810:
album-on-upload


akash1810:
> The order (and groupings) of metadata should be consistent between pages.

👍 👍 👍  YES!!! Agreed and done.

image

paperboyo:
Thanks, @akash1810! Should we add a dimmed placeholder text explaining what the field is in for? Me Englisch wery sloow, but maybe:

> Use **only** to identify multiple images belonging to the same group/photoshoot.

PR #2244, opened 2018-07-26
add album search

akash1810:
Search via a link, similar to other metadata fields:

album-link

Search via chip with suggestion:

album-chip


LATaylor-guardian:
👍

PR #2248, opened 2018-07-27
Front usages

Reettaphant:
Add fronts usages to usages - the fronts tool will posts to details of the usages to the fronts usages endpoint. 
screen shot 2018-07-27 at 14 00 49

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
@@ -207,6 +207,12 @@ object Mappings {
       "composerUrl" -> nonAnalyzedString
     )
 
+  val frontUsageMetadata =
+    nonDynamicObj(
+      "addedBy" -> nonAnalyzedString,
+      "front" -> nonAnalyzedString
+    )
+
   val usagesMapping =
     nonDynamicObjWithAttrs(
       "id"           -> nonAnalyzedString,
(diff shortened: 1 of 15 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -207,6 +207,12 @@ object Mappings {
       "composerUrl" -> nonAnalyzedString
     )
 
+  val frontUsageMetadata =
+    nonDynamicObj(
+      "addedBy" -> nonAnalyzedString,
```
akash1810: Would a datetime be useful too?

common-lib/src/main/scala/com/gu/mediaservice/model/usage/UsageStatus.scala
```diff
@@ -7,13 +7,16 @@ trait UsageStatus {
   override def toString = this match {
     case _:PendingUsageStatus => "pending"
     case _:PublishedUsageStatus => "published"
+    case _:FrontUsageStatus => "front_usage"
```
akash1810: Not sure this is makes sense, semantically. We almost need a `platform` or similar in the model.

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
Reettaphant: The time the image was used does get stored - just not in the metadata. It's handled like it for all other usages.

common-lib/src/main/scala/com/gu/mediaservice/model/usage/UsageStatus.scala
Reettaphant: I understand what you mean, this doesn't really make sense but since we don't distinguish between pending and published usages in fronts usages it sort of makes sense, the status is not pending and it's not published because it is a different type of usage.

paperboyo: So the search chip would be `usages@status:fronts` like `usages@status:published/pending`?
 as opposed to `usages@platform:print/digital`? Hmm… a little confusing indeed, but as Reetta said front overrides do not really have a `pending`/`published` status as such…

PR #2250, opened 2018-07-27
(temp) remove album field from uploads page

akash1810:
We've not yet shown this feature to picture editors and have been getting questions about it as some users are confusing it with the description field (muscle memory!).

Remove the field from the uploads page until we've talked with pic eds (next week).

### DIFF
--- a/kahuna/public/js/edits/image-editor.html
+++ b/kahuna/public/js/edits/image-editor.html
@@ -119,17 +119,6 @@
             </div>
         </div>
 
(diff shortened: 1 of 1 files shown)


akash1810:
> do we need to remove it from the other places as well?

I opted against it on the assumption that they're more easily distinguishable as they're in their own section, as opposed to the uploads page where muscle memory says description is under rights.

I may be being too lazy though...

PR #2252, opened 2018-07-27
be more possessive

akash1810:
my uploads, not yours.

Also remove the, redundantly duplicated, link from the drop-down.


kahuna/public/js/search/view.html
```diff
@@ -9,8 +9,8 @@
            gr:track-click="Your recent uploads button"
```
akash1810: not changing this your to my to avoid splitting the tracking

PR #2254, opened 2018-07-30
Apply syndication rights to images from album

marialivia16:


thrall/app/lib/ElasticSearch.scala
```diff
@@ -218,21 +220,39 @@ class ElasticSearch(config: ThrallConfig, metrics: ThrallMetrics) extends Elasti
...
     }
 
-  def prepareImageUpdate(id: String)(op: UpdateRequestBuilder => Future[UpdateResponse]): List[Future[UpdateResponse]] = {
-    prepareForMultipleIndexes( index => {
-          val updateRequest = client.prepareUpdate(index, imageType, id)
-            .setScriptLang("groovy")
-          op(updateRequest)
-        }
-    )
+  def getImageAlbum(id: String)(implicit ex: ExecutionContext): Future[Option[String]] =
```
akash1810: could we return a `Future[Option[Album]]` as a start to stricter typing?

thrall/app/lib/ThrallMessageConsumer.scala
```diff
@@ -82,8 +81,34 @@ class ThrallMessageConsumer(config: ThrallConfig, es: ElasticSearch, thrallMetri
   def deleteAllUsages(usage: JsValue) =
     Future.sequence( withImageId(usage)(id => es.deleteAllImageUsages(id)) )
 
-  def updateRcsRights(rights: JsValue) =
-    Future.sequence( withImageId(rights)(id => es.updateImageSyndicationRights(id, rights \ "data")) )
+  def applyRcsRights(rightsWithId: JsValue)=
+    Future.sequence(withImageId(rightsWithId) { id =>
+      println(s"Apply rcs rights: $id")
```
akash1810: move to log statement?

thrall/app/lib/ElasticSearch.scala
```diff
@@ -218,21 +220,39 @@ class ElasticSearch(config: ThrallConfig, metrics: ThrallMetrics) extends Elasti
...
+      .executeAndLog(s"get album for image $id")
+      .map{res =>
+        res.sourceOpt.map(json => (json \ "userMetadata" \ "album" \ "title").as[String])}
+
+  def getImageIdsInAlbum(albumNameOpt: Option[String])(implicit ex: ExecutionContext): Future[List[String]] = albumNameOpt match {
+    case Some(albumName) =>
+      val q = filteredQuery(
+        matchQuery("userMetadata.album.title", albumName),
+        andFilter(
+          missingOrEmptyFilter("syndicationRights"))
```
akash1810: Shouldn't we overwrite the syndication rights info everywhere based on a RCS update?

marialivia16: That's what I initially thought. But I think we need a bit more logic in that case as at the moment I'd be overriding everything including the `acquired` field. I don't think we should change a `no` to a `yes`. I'll have a think how to only apply the rights properties.

marialivia16:
Closing this for now, as we are going to move this logic outside the Grid.

PR #2259, opened 2018-07-31
Move album field to beneath label field

akash1810:
Also renames it, within the ui, to photoshoot. This is after talking with Joanna and Josy.

# Image preview page
image

# Info panel on search
image

# Uploads page
image


kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
```diff
@@ -297,6 +288,17 @@
     </dl>
 </div>
 
+<div class="image-info">
+    <div class="image-info__group">
+        <dl class="image-info__wrap">
+            <dt class="metadata-line metadata-line__key">Photoshoot</dt>
+            <dd class="metadata-line__info">
+                <gr-album images="[ctrl.image]"></gr-album>
```
franziskas: is it worth changing the tag `<gr-album>` for consistency?

franziskas: See also https://github.com/guardian/grid/search?q=album&unscoped_q=album

akash1810: Yeah, that's definitely the correct thing to do. We'd need to also update the ES mappings, which I think should be fine...

akash1810: Done. Given the numbers, I think we can manually migrate any existing images with an `album`, moving them to `photoshoot`.

PR #2261, opened 2018-07-31
make UsageStatus optional

akash1810:
This is to support Fronts usage reporting as we don't know the publish state from this Tool.

I think we need to understand why this is before merging this; as we're rebuilding Fronts, we may have an opportunity to address this? cc @Reettaphant

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/usage/Usage.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/usage/Usage.scala
@@ -9,7 +9,7 @@ case class Usage(
   references: List[UsageReference],
   platform: String,
   media: String,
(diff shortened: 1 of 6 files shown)


akash1810:
cc @Reettaphant

paperboyo:
As overrides (fronts usages) are transient, this is an opportunity to look at supporting Historical Usage (was used, but articcle taken down, was used as override on fronts, but not any more, was pending for p.45, but finally outputted on p.56 etc). Maybe.

akash1810:
@paperboyo we model a [`removed` usage](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/model/usage/UsageStatus.scala#L17) which feels like a good fit for this problem.

Issue #2179, opened 2018-06-14
Leases batch applying woes


paperboyo:
- applied Leases are invisible in the Browser (on a search page)
- Leases can be batch applied in Browser, through Metadata Panel, but they are subsequently not visible in this panel (Leases area empty!), one has to go into Viewer (image view) to see them
- thus, Leases cannot be batch-removed in Browser

Slightly related to https://github.com/guardian/grid/issues/1906.

PR #2262, opened 2018-07-31
fix leases in info panel on search page

akash1810:
Fixes https://github.com/guardian/grid/issues/2179.

`ctrl.images` is a `Set` which doesn't have a `length` property 🤦‍♂️

Multi leases may need some attention for the syndication workflow, but I think that's a separate PR.

# Single image selected
image

# Multiple images selected
image

### DIFF
--- a/kahuna/public/js/leases/leases.js
+++ b/kahuna/public/js/leases/leases.js
@@ -197,12 +197,12 @@ leases.controller('LeasesCtrl', [
             ctrl.updateLeases();
         });
 
-        $scope.$watch(() => ctrl.images.length, () => {
+        $scope.$watchCollection(() => Array.from(ctrl.images), () => {
+            ctrl.totalImages = ctrl.images.size;
             ctrl.updateLeases();
         });
 
         ctrl.resetLeaseForm();
-        ctrl.updateLeases();
 }]);
 
 leases.directive('grLeases', [function() {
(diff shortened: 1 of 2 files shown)


akash1810:
fixes https://github.com/guardian/grid/issues/2179

kahuna/public/js/leases/leases.js
```diff
@@ -197,12 +197,12 @@ leases.controller('LeasesCtrl', [
...
         });
 
-        $scope.$watch(() => ctrl.images.length, () => {
+        $scope.$watchCollection(() => Array.from(ctrl.images), () => {
+            ctrl.totalImages = ctrl.images.size;
             ctrl.updateLeases();
         });
 
         ctrl.resetLeaseForm();
-        ctrl.updateLeases();
```
franziskas: did you mean to remove this line?

akash1810: Sorry, should have added a comment to the PR 😊. Yep, this deletion was intentional.

`$scope.$watchCollection` will ... watch ... `ctrl.images` and when it changes, call `ctrl.updateLeases`.

Explicitly calling `updateLeases` again here amounts to it being called twice on component mount*, so isn't necessary.

⏩ 🚗 🐎 

*not sure what the Angular equivalent of React mounting is...

PR #2263, opened 2018-08-01
DigitalUsageMetadata as a trait for extensibility

akash1810:
The motivation behind this is to allow for front usage metadata and syndication usage metadata without creating more options on the `Usage` model.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/usage/DigitalUsageMetadata.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/usage/DigitalUsageMetadata.scala
@@ -2,23 +2,41 @@ package com.gu.mediaservice.model
 
 import play.api.libs.json._
 
+sealed trait DigitalUsageMetadata {
+  val url: String
+  val title: String
+  def toMap: Map[String, String]
+}
 
-case class DigitalUsageMetadata(
-  webTitle: String,
+case class ArticleUsageMetadata (
   webUrl: String,
+  webTitle: String,
   sectionId: String,
   composerUrl: Option[String] = None
-) {
-  val placeholderWebTitle = "No title given"
-  val dynamoSafeWebTitle = if (webTitle.isEmpty) placeholderWebTitle else webTitle
+) extends DigitalUsageMetadata {
+  override val url: String = webUrl
+  override val title: String = webTitle
 
-  def toMap = Map(
+  private val placeholderWebTitle = "No title given"
+  private val dynamoSafeWebTitle = if(webTitle.isEmpty) placeholderWebTitle else webTitle
+
+  override def toMap: Map[String, String] = Map(
     "webTitle" -> dynamoSafeWebTitle,
     "webUrl" -> webUrl,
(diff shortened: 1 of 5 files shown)


common-lib/src/main/scala/com/gu/mediaservice/model/usage/DigitalUsageMetadata.scala
```diff
@@ -1,24 +1,44 @@
 package com.gu.mediaservice.model
 
+import java.net.URI
 import play.api.libs.json._
+import com.gu.mediaservice.syntax._
 
+sealed trait DigitalUsageMetadata {
+  val url: URI
```
Reettaphant: What does the url refer to?

akash1810: Its a link to the resource that has used an image, e.g composer article, syndication partner's website, front url.

This is a renaming of `webUrl` from before. It feels a bit redundant as `UsageReference` also has it...

common-lib/src/main/scala/com/gu/mediaservice/syntax/PlayJsonSyntax.scala
```diff
@@ -1,14 +1,33 @@
...
+      } match {
+        case Success(uri) => JsSuccess(uri)
+        case Failure(_) => JsError(s"Could not parse $uriInJson as valid URI")
+      }
+      case _ => JsError("URI as String expected")
+    }
+  }
 }
 
 object PlayJsonSyntax extends PlayJsonSyntax
```
marialivia16: I haven't seen this approach before. Why not make it an `object` from the start or leave it a `trait` and add it where used?

common-lib/src/main/scala/com/gu/mediaservice/model/usage/DigitalUsageMetadata.scala
akash1810: A quick investigation shows we generate `UsageReferences`, so we can afford to lose them and just use `DigitalUsageMetadata `. I'd suggest that as a separate PR though.

common-lib/src/main/scala/com/gu/mediaservice/syntax/PlayJsonSyntax.scala
akash1810: Not entirely sure tbh. It could probably do with a review in the future.

common-lib/src/main/scala/com/gu/mediaservice/model/usage/DigitalUsageMetadata.scala
paperboyo: Could lead to a container on front, e.g. https://www.theguardian.com/uk#opinion or to a specific story editing mode in new frontstool that will allow for deep-linking those.

Of course, because of a transient nature of image overrides and because the lack of discovery is still frontstool best security mechanism – I just wanted to mention the above for completeness and pure pedantry.

PR #2275, opened 2018-08-06
the syndication tier should only return RCS yes and Editorial yes images

akash1810:


media-api/app/lib/elasticsearch/SearchFilters.scala
```diff
@@ -86,7 +86,8 @@ class SearchFilters(config: MediaApiConfig) extends ImageFields {
   def syndicationRightsAcquiredFilter(): FilterBuilder = {
     filters.and(
       rightsAcquiredFilter(isAcquired = true),
-      filters.date(field = "syndicationRights.published", None, Some(DateTime.now)).get
+      filters.date(field = "syndicationRights.published", None, Some(DateTime.now)).get,
+      filters.term(field = "leases.leases.access", term = AllowSyndicationLease.name)
```
akash1810: specifying the field path like this isn't particularly elegant...

media-api/test/lib/ElasticSearchTest.scala
```diff
@@ -46,21 +60,20 @@ class ElasticSearchTest extends FunSpec with BeforeAndAfterAll with Matchers wit
...
     }
   }
 
-  override def afterAll {
-    Await.ready(cleanTestUserImages(), 2.seconds)
-    DateTimeUtils.setCurrentMillisSystem()
-  }
+  override def afterAll():Unit = DateTimeUtils.setCurrentMillisSystem()
+
+  override def afterEach(): Unit = Await.ready(deleteImages(), 5.seconds)
```
akash1810: *sigh* tests were failing stating the number of images returned was greater than `images.size`. This is impossible, so delete everything after each test to be safe.

akash1810:
now with passing tests! 🎉

media-api/test/lib/ElasticSearchTest.scala
franziskas: @akash1810 is still looks like more is returned that necessary
https://teamcity.gutools.co.uk/viewLog.html?buildId=159713&buildTypeId=EditorialTools_Grid_MediaApi

Maybe something isn't deleted properly?

There is a technique where you make a different name for each test run and only search for that, not sure how applicable that is.

akash1810: Just ran the tests again and ✅ 

img

build.sbt
```diff
@@ -138,7 +138,8 @@ val testSettings = Seq(
   testOptions in Test += Tests.Setup(_ => {
 
     println(s"Launching docker container with ES")
-    s"docker-compose up -d elasticsearch".!
+    s"docker-compose --file docker-compose-test.yml --project-name grid-test down".!
```
akash1810: Deleting the docker stack appears to be key to getting a green build. An alternative could be to assign a random `project-name` each time, but that may result in random stacks remaining if not cleaned up properly.

PR #2278, opened 2018-08-08
extract upsert logic into class

akash1810:
the start of copying rights...


akash1810:
replaced with https://github.com/guardian/grid/pull/2279

PR #2279, opened 2018-08-09
copy syndication rights with new rcs image

akash1810:
On receipt of a `upsert-rcs-rights` message for image A, save the rights.

If image A is in a photoshoot B:
- find all images in photoshoot B that have inferred rights and is not image A
- send a `refresh-inferred-rights` message for each of these images

On receipt of a `refresh-inferred-rights` message, save the rights.


thrall/app/lib/ThrallMessageConsumer.scala
```diff
@@ -83,9 +94,62 @@ class ThrallMessageConsumer(config: ThrallConfig, es: ElasticSearch, thrallMetri
...
+        GridLogger.error(msg)
+        Future.failed(SNSBodyParseError(msg))
+      },
+      rightsFromRcs => withImageId(rights) { id =>
+        es.getImage(id) map {
+          case Some(image) => {
+            GridLogger.info(s"Upserting syndication rights from RCS feed", image.id)
+            es.updateImageSyndicationRights(id, Some(rightsFromRcs))
+
+            image.userMetadata.map{edits => {
```
marialivia16: No need for the second pair of `{}` here.

thrall/app/lib/ThrallMessageConsumer.scala
```diff
@@ -83,9 +94,62 @@ class ThrallMessageConsumer(config: ThrallConfig, es: ElasticSearch, thrallMetri
...
+              edits.photoshoot match {
+                case Some(photoshoot) if !rightsFromRcs.isInferred => {
```
marialivia16: Maybe the naming is not great here as while reading it I was expecting `rightsFromRcs` to never be inferred. Maybe `syndicationRights` works better in this case?

thrall/app/lib/ThrallMessageConsumer.scala
```diff
@@ -83,9 +94,62 @@ class ThrallMessageConsumer(config: ThrallConfig, es: ElasticSearch, thrallMetri
...
+    )
+  }
+
+  private def initiateInferredRightsRefresh(images: List[Image], rightsFromRcs: SyndicationRights) = {
+    GridLogger.info(s"initiating refresh of inferred rights for ${images.size} images")
+    val inferredRights = rightsFromRcs.copy(published = None)
+
+    images.foreach(image => {
+      val updatedImage = image.copy(syndicationRights = Some(inferredRights))
+      thrallNotifications.publish(Json.toJson(updatedImage), subject = "refresh-inferred-rights")
```
marialivia16: This sends the entire image while we only need a small subset of the fields. Can we send only the id and the rights instead? (like we do for `upsert-rcs-rights`)

thrall/app/lib/ThrallMessageConsumer.scala
```diff
@@ -83,9 +94,62 @@ class ThrallMessageConsumer(config: ThrallConfig, es: ElasticSearch, thrallMetri
...
+          case _ => {
+            GridLogger.info(s"image $id not found")
+            None
+          }
+        }
+      }
+    )
+  }
+
+  private def initiateInferredRightsRefresh(images: List[Image], rightsFromRcs: SyndicationRights) = {
```
marialivia16: This introduces a lot of logic in `ThrallMessageConsumer`, it would be nice pull this out and have smaller steps. It would also make it easier to read if instead of the nested `upsertSyndicationRights` logic we'd have a for comprehension. But it's up to you how much time you want to spend on this :)

thrall/app/lib/ThrallMessageConsumer.scala
akash1810: Yes! Means we can reuse more code! PR updated now.

PR #2281, opened 2018-08-10
update inferred rights when updating an image's photoshoot

akash1810:
When an image moves between photoshoots, we need to be sure any inferred syndication rights are up-to-date in the old and new photoshoot.

Image has inferred rights:
- if we're removing it from a photoshoot, remove its inferred rights
- if we're moving it into a photoshoot, recalculate its inferred rights:
  - if the photoshoot has images with explicit rights, use the latest
  - else remove the inferred rights

Image has explicit rights:
- if the image was in a photoshoot, refresh the inferred rights for images in that photoshoot:
  - use the latest explicit rights in the photoshoot that's not this image. If none, clear inferred rights across the photoshoot
- if we're moving into a photoshoot and this image's rights is the latest, copy the rights across the photoshoot
- if we're moving into a photoshoot and this image's rights is not the latest, do nothing (TODO record these as a metric to know how often it happens)

### DIFF
--- a/thrall/app/lib/ThrallMessageConsumer.scala
+++ b/thrall/app/lib/ThrallMessageConsumer.scala
@@ -15,7 +15,7 @@ class ThrallMessageConsumer(
   thrallMetrics: ThrallMetrics,
   store: ThrallStore,
   dynamoNotifications: DynamoNotifications,
-  thrallNotifications: ThrallNotifications
+  syndicationNotifications: SyndicationNotifications
 )(implicit ec: ExecutionContext) extends MessageConsumer(
   config.queueUrl,
   config.awsEndpoint,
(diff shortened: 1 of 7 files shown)


thrall/app/lib/ThrallMessageConsumer.scala
```diff
@@ -115,16 +118,131 @@ class ThrallMessageConsumer(
...
+  private def deleteInferredRights(message: JsValue) = {
+    Future.sequence(
+      withImageId(message) { id =>
+        GridLogger.info("deleting inferred rights", id)
+        es.deleteSyndicationRights(id)
+      }
+    )
+  }
+
+  private def notInferred(image: Image, upcomingEdits: Edits) = {
```
marialivia16: This is long and has a lots of duplication, maybe we can improve by having smaller reusable steps?

akash1810:
> Personally, I'd take all the logic out of ThrallMessageConsumer - especially all the private functions. They add a lot of complexity that is not needed in this class.

👍refactored a little for better organisation.

PR #2285, opened 2018-08-14
add syndicationStatus param

akash1810:
to filter by syndication status:
- `sent`: rcs yes, editorial yes, syndication usage
- `queued`: rcs yes, editorial yes, no syndication usage
- `blocked`: rcs yes, editorial no
- `review`: rcs yes, no editorial response

This is a query string value and not a chip, mainly because syndication states isn't useful for the majority of Grid users initially. The query string can be bookmarked. We can add a chip if/when necessary.

### DIFF
--- a/media-api/app/model/SyndicationStatus.scala
+++ b/media-api/app/model/SyndicationStatus.scala
@@ -1,0 +1,32 @@
+package model
+
+import play.api.libs.json._
+
+sealed trait SyndicationStatus {
+  override def toString: String = this match {
+    case SentForSyndication => "sent"
+    case QueuedForSyndication => "queued"
+    case BlockedForSyndication => "blocked"
+    case AwaitingReviewForSyndication => "review"
+  }
+}
+
+object SyndicationStatus {
+  def apply(status: String): SyndicationStatus = status.toLowerCase match {
+    case "sent" => SentForSyndication
+    case "queued" => QueuedForSyndication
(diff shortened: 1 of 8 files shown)


media-api/app/model/SyndicationStatus.scala
```diff
@@ -0,0 +1,32 @@
+package model
+
+import play.api.libs.json._
+
+sealed trait SyndicationStatus {
```
marialivia16: This is very verbose, we can probably improve using https://github.com/lloydmeta/enumeratum

Also, happy to leave it as it is for now 😃

akash1810: Hmm, we do this already for `payType` and it didn't feel that nice to me...

akash1810: can't remember why though 🤷‍♂️

marialivia16: Fair enough. 

I quite like this: https://github.com/guardian/workflow/blob/ae0185d35181fa613fc2169b42406de1c17ebb11/common-lib/src/main/scala/models/Status.scala

akash1810: d'oh it was its [import in `ElasticSearch.scala`](https://github.com/guardian/grid/blob/master/media-api/app/lib/elasticsearch/ElasticSearch.scala#L8) which can be solved by using the `model` package

PR #2287, opened 2018-08-15
log endpoint to measure usage

akash1810:
The imports endpoint in image-loader currently accepts any url from the internet, which is only a little scary! This endpoint should only be used by S3 Watcher, but log out who is using it to be sure.

Ultimately, this endpoint will be rewritten to accept an S3 object path rather than a URL. This:
- will reduce the amount of logic in S3 Watcher
- be more secure as only images from the S3 Watcher inject bucket will be accepted
- enable reindexing to reuse this route

### DIFF
--- a/image-loader/app/controllers/ImageLoaderController.scala
+++ b/image-loader/app/controllers/ImageLoaderController.scala
@@ -7,6 +7,7 @@ import com.gu.mediaservice.lib.argo.ArgoHelpers
 import com.gu.mediaservice.lib.argo.model.Link
 import com.gu.mediaservice.lib.auth.Authentication.Principal
 import com.gu.mediaservice.lib.auth._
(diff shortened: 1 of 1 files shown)


image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -49,6 +50,7 @@ class ImageLoaderController(auth: Authentication, downloader: Downloader, store:
 
   def importImage(uri: String, uploadedBy: Option[String], identifiers: Option[String], uploadTime: Option[String], filename: Option[String]) =
     auth.async { request =>
+      GridLogger.info(s"request to import an image", request.user.apiKey)
```
akash1810: this'll log `request.user.apiKey` as a marker for friendly log searching

PR #2291, opened 2018-08-17
delay syndication until a specific date

akash1810:
`allow-syndication` can have a start date. Image will be sent after this date.
`deny-syndication` can have an end date. Image will require reclassification after this date.

We couldn't re-use `gu-date-range` because we only need a single date. So `gu-date` has been created.

Additionally, we can now refactor `gu-date-range` to be a lot simpler as it can embed two `gu-date`s. That's for another PR though.

# allow-syndication lease
allow-syndication

# deny-syndication lease
deny-syndication

# cropping leases
To demonstrate no regressions 🤗 

crop-lease

### DIFF
--- a/kahuna/public/js/components/gu-date/gu-date.js
+++ b/kahuna/public/js/components/gu-date/gu-date.js
@@ -1,0 +1,102 @@
+import angular from 'angular';
+import moment from 'moment';
+import Pikaday from 'pikaday';
+import 'pikaday/css/pikaday.css';
+
+import template from './gu-date.html';
+import './gu-date.css';
+
+const ISO8601_FORMAT = 'YYYY-MM-DDTHH:mm:ssZ';
+const DISPLAY_FORMAT = 'DD MMM YYYY';
+const TEN_YEARS_MILLIS = (10 * 365 * 24 * 60 * 60 * 1000);
+const START_OF_WEEK = 1; // Monday
+
+export const guDate = angular.module('gu.date', []);
+
+guDate.directive('guDate', [function () {
+    return {
+        template: template,
+        scope: {
+            label: '@',
+            date: '=',
+            minDate: '=?',
+            maxDate: '=?'
+        },
+        link: function($scope, el) {
+            if (angular.isDefined($scope.minDate) && angular.isDefined($scope.maxDate)) {
+                throw 'gu-date can have either a minDate or a maxDate. Not both.';
+            }
+
+            function getDateISOString (value) {
+                return angular.isDefined(value)
+                    ? moment(value).toISOString()
+                    : undefined;
+            }
+
+            function getDisplayValue(value) {
+                return angular.isDefined(value)
+                    ? moment(value).format(DISPLAY_FORMAT)
+                    : 'anytime';
+            }
+
+            const tenYearsFromNow = new Date(Date.now() + TEN_YEARS_MILLIS);
+
+            const root = el[0];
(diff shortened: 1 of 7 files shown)


kahuna/public/js/components/gu-date/gu-date.js
```diff
@@ -0,0 +1,102 @@
...
+import moment from 'moment';
+import Pikaday from 'pikaday';
+import 'pikaday/css/pikaday.css';
+
+import template from './gu-date.html';
+import './gu-date.css';
+
+const ISO8601_FORMAT = 'YYYY-MM-DDTHH:mm:ssZ';
+const DISPLAY_FORMAT = 'DD MMM YYYY';
+const TEN_YEARS_MILLIS = (10 * 365 * 24 * 60 * 60 * 1000);
```
franziskas: this seams a little simplistic (since years are a bit longer) but I assume you don't need this to be precise

kahuna/public/js/components/gu-date/gu-date.js
```diff
@@ -0,0 +1,102 @@
...
+            const pika = new Pikaday({
+                field: input,
+                container: container,
+                bound: false,
+                minDate: $scope.minDate && new Date($scope.minDate),
+                maxDate: tenYearsFromNow,
+                yearRange: 100,
+                firstDay: START_OF_WEEK,
+                format: ISO8601_FORMAT,
+                keyboardInput: false
```
franziskas: I assume we don't want to enable keyboard input?

kahuna/public/js/components/gu-date/gu-date.js
```diff
@@ -0,0 +1,102 @@
...
+                pika.gotoToday();
+            };
+
+            $scope.closeOverlay = () => $scope.showingOverlay = false;
+
+            if (angular.isDefined($scope.minDate)) {
+                $scope.dateRounder = (date) => moment(date).endOf('day').toDate();
+
+                $scope.$watch('minDate', value => {
+                    const dateValue = value ? new Date(value) : new Date();
```
franziskas: should the minimum date always default to today? Or am I missing something?

kahuna/public/js/components/gu-date/gu-date.js
```diff
@@ -0,0 +1,102 @@
...
+                    pika.setMinDate(dateValue);
+                });
+            }
+
+            if (angular.isDefined($scope.maxDate)) {
+                $scope.dateRounder = (date) => moment(date).startOf('day').toDate();
+
+                $scope.$watch('maxDate', value => {
+                    const dateValue = value ? new Date(value) : new Date();
```
franziskas: same as above for maximum date

kahuna/public/js/components/gu-date/gu-date.js
akash1810: This is to limit the date selection in the UI. The variable name isn't that intention revealing however...

image

kahuna/public/js/components/gu-date/gu-date.js
akash1810: Correct. Pikaday binds to the document rather than the form element, which breaks keyboard navigation elsewhere. More information here - https://github.com/guardian/grid/pull/2217.

kahuna/public/js/components/gu-date/gu-date.js
akash1810: `value` will be falsy if it has been cleared. In this scenario, I think today is the only option...?

PR #2293, opened 2018-08-17
Jr/clean dates in metadata

aug24:


common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
```diff
@@ -77,6 +79,17 @@ object ImageMetadataConverter {
 
   private def safeParsing[A](parse: => A): Option[A] = Try(parse).toOption
 
+  private def cleanDateFormat = DateTimeFormat.forPattern("yyyy-MM-dd'T'HH:mm:ss.SSSZZ")
+  def cleanDate(dirtyDate: String, fieldName: String = "none"): String = parseRandomDate(dirtyDate) match {
+    case Some(cleanDate) => cleanDateFormat.print(cleanDate)
+    case None => {
+      println(s"I give up: $dirtyDate")
+      Logger.info(s"Unable to parse date $dirtyDate from field $fieldName")
```
akash1810: could we log the image id here too?

common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
```diff
@@ -77,6 +79,17 @@ object ImageMetadataConverter {
 
   private def safeParsing[A](parse: => A): Option[A] = Try(parse).toOption
 
+  private def cleanDateFormat = DateTimeFormat.forPattern("yyyy-MM-dd'T'HH:mm:ss.SSSZZ")
+  def cleanDate(dirtyDate: String, fieldName: String = "none"): String = parseRandomDate(dirtyDate) match {
+    case Some(cleanDate) => cleanDateFormat.print(cleanDate)
+    case None => {
+      println(s"I give up: $dirtyDate")
```
akash1810: `println` not needed

PR #2294, opened 2018-08-17
Create Thrall lambda

aug24:


mbarton:
Closing now we're on ES6. We ended up leaving Thrall as an application but this PR will still be useful for reference in the future

PR #2295, opened 2018-08-17
add gu-date-range-x directive

akash1810:
The aim is for this to, eventually, replace the current `gu-date-range` by being more componentised.

`gu-date-range` is very complicated as it has presets baked into the component which are not needed in all cases. It could benefit from component composition... next PR!

# Before
image

# After
image

### DIFF
--- a/kahuna/public/js/components/gu-date/gu-date-range-x.html
+++ b/kahuna/public/js/components/gu-date/gu-date-range-x.html
@@ -1,0 +1,4 @@
+<div>
+    <gu-date class="full-width" label="from" date="start" max-date="end"></gu-date>
+    <gu-date class="full-width" label="to" date="end" min-date="start"></gu-date>
+</div>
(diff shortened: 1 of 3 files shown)


franziskas:
Could this put additional work on users having to click twice to change both dates?

kahuna/public/js/components/gu-date/gu-date-range-x.html
```diff
@@ -0,0 +1,4 @@
+<div>
+    <gu-date class="full-width" label="from" date="start" max-date="end"></gu-date>
```
franziskas: is the max date of start the end? Getting confused now, let's look at this on Monday please :)

akash1810: > is the max date of start the end?

Yes. Possibly not very intention revealing though. The idea here is in a date range, the start date should not exceed the end date, else its an invalid range.

akash1810:
> Could this put additional work on users having to click twice to change both dates?

I thought about this too! Its still four clicks:

Previously:
- Click to open picker
- Click to select start
- Click to select end
- Click to apply filter

Now:
- Click to open start
- Click to set start
- Click to open end
- Click to set end

PR #2299, opened 2018-08-21
add `isInferred` to SyndicationRights model

akash1810:
The motivation behind this is to copy the publish date across to inferred images as we only want to send images to syndication partners if they've been published.

Previously, we were using the absence of `published` as an indicator for inferred rights and to know which images to propagate rights to. With the above rule, we would not be sending all the images we should be!


akash1810:
requires https://github.com/guardian/rcs-poller-lambda/pull/9

PR #2301, opened 2018-08-22
reduce quality of master crop to 95

akash1810:
to reduce filesize whilst maintaining a high quality

It’s impossible to make sure that we not undercompress (like now) or overcompress, without resorting to varying compression depending on quality estimate through -format: %Q on image per image basis.

Very loose investigation confirms suspicions that agency images are of varying quality 75–99 (AP being the best).

As our master crops are later downsized hugely by the frontends and compressed to the bone, we should get away with quite a bit lower than 100 without absolutely any adverse effect for the readers.

If we want to be (too) safe, we can go 95 (roughly halves the filesize from 100).

### DIFF
--- a/cropper/app/lib/Crops.scala
+++ b/cropper/app/lib/Crops.scala
@@ -23,6 +23,9 @@ class Crops(config: CropperConfig, store: CropStore, imageOperations: ImageOpera
   import scala.concurrent.ExecutionContext.Implicits.global
   import Files._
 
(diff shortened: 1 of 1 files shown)


paperboyo:
For posterity, here is example showing 100 and 95 quality impact at 200% zoom. Not showing, that is, especially considering most originals arrive compressed at 75–85 and we downsize them heavily:

test

PR #2302, opened 2018-08-23
syndication icon traffic light

akash1810:
Add `syndicationStatus` to API response to create a traffic light in the search page. Putting this in the API keeps the UI simple.

The previous version of this icon doesn't make sense anymore as it signalled that rights had been acquired but nothing about if its been sent at all.

A traffic light allows you to quickly understand what's going on.

- `sent` is green
- `queued` is yellow
- `blocked` is red
- `review` is white

For example:
orange

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/Image.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/Image.scala
@@ -1,6 +1,6 @@
 package com.gu.mediaservice.model
 
-import com.gu.mediaservice.model.usage.Usage
+import com.gu.mediaservice.model.usage.{SyndicationUsage, Usage}
 import org.joda.time.DateTime
 import play.api.libs.functional.syntax._
 import play.api.libs.json._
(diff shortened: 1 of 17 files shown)


kahuna/public/js/preview/image.html
```diff
@@ -121,10 +121,8 @@
                 <gr-icon>today</gr-icon>
             </div>
 
-            <div class="bottom-bar__action"
-                 ng-if="ctrl.states.isSyndicated"
-                 title="Image is available to syndication partners">
-                <gr-icon>monetization_on</gr-icon>
+            <div class="bottom-bar__action" ng-if="ctrl.states.syndicationStatus !== 'unsuitable'">
```
akash1810: `syndicationStatus === 'unsuitable'` when there is no RCS information for the image (anything that has not been used).

Don't show the icon in this case, to keep the UI from getting too busy as there are more unused images than used!

common-lib/src/main/scala/com/gu/mediaservice/model/Image.scala
```diff
@@ -29,6 +29,29 @@ case class Image(
...
+      UnsuitableForSyndication
+    } else {
+      val hasSyndicationUsage = usages.exists(_.platform == SyndicationUsage)
+
+      if (hasSyndicationUsage) {
+        SentForSyndication
+      } else {
+        val allowSyndicationLease = leases.leases.find(_.access == AllowSyndicationLease)
+        val denySyndicationLease = leases.leases.find(_.access == DenySyndicationLease)
+
```
marialivia16: Isn't the case that an image can only have one syndication lease? If so, maybe we can simplify this a bit. 
e.g:
```
leases.leases.find(_.isSyndication).map(_.access) match {
    case Some(AllowSyndicationLease) => QueuedForSyndication
    case Some(DenySyndicationLease) => BlockedForSyndication
   case None => AwaitingReviewForSyndication
}
```

akash1810: Sadly, that rule's only enforced client-side atm 😞

PR #2307, opened 2018-08-29
don't show images for syndication review before training finished

akash1810:
Syndication starts on 23 August 2018 as that's when training had been completed. Don't show images uploaded prior to this date to keep the review queue manageable for Editorial by not showing past images that RCS has told us about (~5k images).


akash1810:
> I'd move the start date to config

👍 will do separately.

PR #2308, opened 2018-08-29
syndicationStartDate in config

akash1810:
following #2307


tonytw1:
Effects Elastic6 as well.

akash1810:
> Effects Elastic6 as well.

Rebased and updated both ES clients now.

PR #2313, opened 2018-09-12
Read api keys from DynamoDb

marialivia16:
This moves the api keys from S3 to DynamoDB.

Before merging we have to:
- [ ] Create DynamoDB tables
- [ ] Migrate all keys from S3 -> DynamoDB
- [ ] Update config to point at the tables


tonytw1:
Agree with this one. When porting to containers the API key bucket and it's format felt weird - a Dynamo table feels more natural.

The scanning the bucket / table for a key is abit of a surprise; an external developer would probably be expecting to provide the calling service with a configured key rather than having access to the underlying key store from a client deployable.

stephengeller:
Can this be closed?

PR #2318, opened 2018-09-17
allow credit to have no slash

akash1810:
Update BylineCreditReorganise cleaner, allowing credit to be a single element.

This allows the byline to be `foo/bar` and credit to be `foo`. This was previously failing as `credit` had no tail.

Ideally this cleaner would produce `byline:None` and `credit:foo/bar` however it would require a bit of a refactor to support an `Option[String]` type...maybe later?

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/ByLineCreditReorganise.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/ByLineCreditReorganise.scala
@@ -29,7 +29,7 @@ object ByLineCreditReorganise extends MetadataCleaner {
 
         // if we have a split, and the first split is the same, remove if from credit
         // and use it as byline, the rest is the credit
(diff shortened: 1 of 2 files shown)


currysoup:
I'm interested in everyones opinion on if we should just first of all confirm the`byline` and `credit` field are both defined and identical before doing any of this. Just splitting on slashes and hoping we don't have `Nil`s where we don't expect them seems like we could end up chasing edge cases forever and ever.

For example a byline: `Sam Cutler / Mateusz Karpow` and credit `Sam Cutler / The Guardian` would silently drop the Karpow credit in the current code.

Rather than assume slashes are always separators for organisational hierarchy we could maybe just say that this cleaner is *only* for when the byline and credit match and both have slashes. More strict rules about when to apply the cleaner means fewer chasing edge cases, I hope.

franziskas:
@currysoup @akash1810 maybe you can have a chat about this and decide? :)

PR #2319, opened 2018-09-17
Quotas page

mbarton:
Visualisation how far through quota we are used to be available as a "wat" page powered by https://github.com/guardian/grid-quotas. Unfortunately when we did the mighty Play 2.6 upgrade (#2124) the new CORS policy broke that page.

Not everyone who may wish to run the Grid in the future will have a wat page so I have moved the functionality into a new page (`/quotas`):

It's a beaut. It's also accessible from the three-dot menu:

I would have liked to do the page as a simple Twirl template but unfortunately the microservice architecture means Kahuna does not have access to the raw quota. As such I had to write some new JS to fetch the quota and render the UI. In theory `media-api` could serve the UI page directly but that feels even shonkier than some new JS.

The old page had some fancy bars which I have not ported across since they are non-essential and it took ages just to get this simple UI done.


mbarton:
[Grid: Grid Quotas blank](https://trello.com/c/16yqVKXg/491-grid-grid-quotas-blank)

kahuna/app/controllers/KahunaController.scala
```diff
@@ -24,6 +24,10 @@ class KahunaController(config: KahunaConfig, override val controllerComponents:
     ))
   }
 
+  def quotas = Action { req =>
```
akash1810: should we throw auth in here?

kahuna/public/js/quotas.js
```diff
@@ -0,0 +1,32 @@
...
+}
+
+function renderSupplier(name, entry) {
+    const colour = entry.exceeded ? "red" : "green";
+
+    const usage = entry.usage ? entry.usage.count : "unknown";
+    const quota = entry.quota ? entry.quota.count : "unknown";
+
+    return `
+        <h3 style="background-color: ${colour}">${name}</h3>
```
akash1810: I think semantically these should be list items.

kahuna/app/controllers/KahunaController.scala
mbarton: Ah that's a good point. I thought Kahuna didn't have access to auth actions at all but it does after the refactor! I may be able to do the whole thing as a twirl template then

PR #2320, opened 2018-09-18
Proposal: Add byline renamer for commonly misspelled bylines

currysoup:
Turns out or good pal Mat has been maintaining a list of commonly misspelled and manually fixing them.

Shall we wrap this into a metadata cleaner?


common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/PhotographerRenamer.scala
```diff
@@ -0,0 +1,140 @@
+package com.gu.mediaservice.lib.cleanup
+
+import com.gu.mediaservice.model.ImageMetadata
+
+object PhotographerRenamer extends MetadataCleaner {
```
franziskas: I can haz 1 test plz???

paperboyo:
💎!

currysoup:
Yeah! If it was cached locally for performance and had a little page for adding new ones that would be so neat!

paperboyo:
💯 to both comments. Most cleaners should be in the hands of (power)users TBH. Benefits would be mutual ;-)

Until then, expect me to be adding to this list regularly…

PR #2324, opened 2018-09-26
change media-api healthcheck to not hit es

akash1810:
If es is slow (because aws networking!), the healthcheck fails and the instance cycles. A new instance is launched in its place, and the same thing happens. And then a new instance is launched and...

Change the healthcheck to not depend on es.


marialivia16:
Oh you need to update the AppController..

PR #2325, opened 2018-09-26
Start tests for usage

aug24:
All projects should have tests, right?


mbarton:
Closing now we're on ES6. This PR will still be useful for reference however should we want to make Thrall a lambda again in the future

PR #2327, opened 2018-09-28
Update gr-description-warning.html

akash1810:
suggestion of updated copy from Jonny


mchv:
👍

paperboyo:
Comma after `Ideally`. Or if we are to be “pushy” – no “Ideally” at all.

PR #2328, opened 2018-10-11
don't syndicate illustrations

akash1810:
requested by Joanna and Nick.


paperboyo:
Hmm, was thinking about it, too. You’re right, RCS should be responsible. But if we decide to do it here, *only* three Rights categories should ever be considered for syndication IMHO:
- `staff-photographer`
- `contract-photographer`
- `commissioned-photographer`
No?

akash1810:
Agreed. However whilst some images continue to be mis-catergorised as `handout` I don't think we can use that rule.

paperboyo:
Yeah. And as Bylines: `search?query=-~%22cutouts%2Fcircular%20panelist%20bylines%22%20category:Bylines`. Which should really be called _Website permanent assets, stay away_.

Tough luck?

PR #2330, opened 2018-10-16
Update ImageCollectionsController.scala

akash1810:
`encodePlus` the incoming collection string to correctly escape


akash1810:
relates to https://github.com/guardian/grid/pull/2171

PR #2332, opened 2018-10-18
Spell MacLeod correctly

paperboyo:
After noticing inconsistent spelling, contacting Murdo himself and the Style Police, an executive decision was taken (not by me), to spell it MacLeod.

There is [this test](https://github.com/guardian/grid/blob/mk-macleod-spelling/common-lib/src/test/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessorsTest.scala#L38), which not only looks odd now, but also worries me a bit. Do we need exception [in here](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/CapitaliseProperty.scala)?
~[**UPDATE**: yeah, sorry, test fails, what does the monkey do now?]~

I will spend the rest of my days fixing past usages, cause images aren’t atoms.

- [x] Tested and working correctly on TEST (looks like `CapitaliseProperty` is not applied for Usage-derived values :)


franziskas:
@paperboyo can we have a chat? It looks good to me but I'd like to understand if there are open questions on your part

PR #2334, opened 2018-10-21
Add to PhotographerRenamer dictionary

paperboyo:
This update is after combing through first 2k most published names (thanks, @currysoup!). Apart from spelling, it contains:
- ~Guardian style `First M. Last` -> `First M Last` pattern~
- ~straight quotes to curly quotes~
- disambiguation: `Dave M Benett`=`David M Benett` and an extremely hard to track case of `Eduardo Munoz` and `Eduardo Munoz Alvarez` both being `Eduardo Muñoz` (unless two of them hang around same places and despite never using same agencies for the same events, cross-post on another’s insta ;)

~First two things may one day be automated~ (watch this space!), in which case I’ll remove them from the dictionary. As to the third one… I believe someone already tried numbers, but even I wouldn’t go that far!


paperboyo:
I have now removed all cases for fullstops and quotes (and updated comment above).

PR #2335, opened 2018-10-23
Add style guide byline cleaners

currysoup:
Often our photographers will submit their names in a way that doesn't conform to our style guide. 

This adds a single cleaner that applies three of the major causes of style guide violations in photographer bylines. Curly quotes rather than straight quotes, removing the full stop after an initial and squishing together pairs of initials.

### DIFF
--- a/common-lib/src/test/scala/com/gu/mediaservice/lib/cleanup/GuardianStyleBylineTest.scala
+++ b/common-lib/src/test/scala/com/gu/mediaservice/lib/cleanup/GuardianStyleBylineTest.scala
@@ -1,0 +1,48 @@
+package com.gu.mediaservice.lib.cleanup
+
+import org.scalatest.{FunSpec, Matchers}
+
+class GuardianStyleBylineTest extends FunSpec with Matchers with MetadataHelper {
+  it("should replace straight quotes with curly quotes") {
+    val metadata = createImageMetadata("byline" -> "Sam O'neill")
+    val cleanedMetadata = GuardianStyle.clean(metadata)
+
+    cleanedMetadata.byline should be (Some("Sam O’neill"))
+  }
+
+  it("should remove dots in initials") {
+    val metadata = createImageMetadata("byline" -> "First M. Last")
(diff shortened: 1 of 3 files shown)


paperboyo:
Amazlang! For the record, see: [initials](https://www.theguardian.com/guardian-observer-style-guide-i) and [quotation marks](https://www.theguardian.com/guardian-observer-style-guide-q).

currysoup:
Kicked the builds - seem to be OK now!

common-lib/src/test/scala/com/gu/mediaservice/lib/cleanup/GuardianStyleBylineTest.scala
```diff
@@ -0,0 +1,48 @@
...
+
+    cleanedMetadata.byline should be (Some("First AB"))
+  }
+
+  it("should remove dots in initials and insert spaces in unusual cases") {
+    val metadata = createImageMetadata("byline" -> "Ishara S.kodikara")
+    val cleanedMetadata = GuardianStyleByline.clean(metadata)
+
+    // NOTE: The capitalisation cleaner should handle this becoming Title Case
+    cleanedMetadata.byline should be (Some("Ishara S kodikara"))
```
akash1810: is it worth having a test that chains these two tests together to be explicit?

currysoup: Yeah I think this is something that could certainly be added to make sure we don't get any weird behaviour!

paperboyo:
Nobody expected Spanish Inquisition:
`M.A.PUSHPA KUMARA` changed to `Ma Pushpa Kumara`
not that we _have to_ fix everybody’s _every_ mistake… I can add it to the dictionary, but it seems to be spelled all over the place, too.

currysoup:
Ugh actually yeah we'll need to put the initial squishier after the capitaliser other wise it will turn `CP Scott` into `Cp Scott`.

currysoup:
Will do that tomorrow - should take 10 seconds!

PR #2336, opened 2018-10-25
reorder initial squish to avoid capitalizer ruining everything

currysoup:
Having the initials squished before the capitaliser caused the following bad behaviour:

`Original → Dots removed → Squish    → Capitalise`
`C. P. Scott → C P Scott         → CP Scott → Cp Scott`

This is bad, reodereding the initial squish to be after the title casing prevents this.


currysoup:
Tested on TEST!

PR #2339, opened 2018-10-30
Improve leases reliability

marialivia16:
When adding a lease to a batch selection some were randomly failing to sync in ES.
That's because we are saving the lease in Dynamo, query Dynamo and send the result to ES. 

To make it more reliable this PR sends messages to ES without relying on Dynamo to have the right answer. This introduces 2 new thrall messages: one that appends a lease and one that removes a lease from the list.

PR also includes refactoring of the `current` lease.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/leases/LeasesByMedia.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/leases/LeasesByMedia.scala
@@ -9,22 +9,20 @@ import JodaWrites._
 
 case class LeasesByMedia(
   leases: List[MediaLease],
-  lastModified: Option[DateTime],
-  current: Option[MediaLease]
+  lastModified: Option[DateTime]
 )
-case object LeasesByMedia {
+
+object LeasesByMedia {
   implicit val reader : Reads[LeasesByMedia] = (__ \ "leases").read[List[MediaLease]].map(LeasesByMedia.build)
 
   implicit val writer = new Writes[LeasesByMedia] {
     def writes(leaseByMedia: LeasesByMedia) = {
       LeasesByMedia.toJson(
         Json.toJson(leaseByMedia.leases),
-        Json.toJson(leaseByMedia.current),
         Json.toJson(leaseByMedia.lastModified.map(lm => Json.toJson(lm)))
       )
     }
   }
-
 
   def build(leases: List[MediaLease]): LeasesByMedia = {
     def sortLease(a: MediaLease, b: MediaLease) =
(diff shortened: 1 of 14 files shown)


marialivia16:
[Grid: Make Syndication lease batches easier](https://trello.com/c/6a3gp9vl/568-grid-make-syndication-lease-batches-easier)

kahuna/public/js/leases/leases.js
```diff
@@ -183,9 +183,8 @@ leases.controller('LeasesCtrl', [
         ctrl.leaseStatus = (lease) => {
             const active = lease.active ? 'active ' : ' ';
 
-            const current = ctrl.leases.current
-                .filter((lease) => lease !== null)
-                .find(l => l.id == lease.id) ? 'current ' : '';
+            // Current only makes sense for use leases
+            const current = (lease.active && lease.access.match(/-use/i)) ? 'current' : '';
```
mbarton: Maybe add a field to the lease to indicate this rather than regex matching?

leases/app/lib/LeaseStore.scala
```diff
@@ -23,7 +23,7 @@ class LeaseStore(config: LeasesConfig) extends DynamoDB(config, config.leasesTab
   def getForMedia(id: String): List[MediaLease] = Scanamo.queryIndex[MediaLease](client)(tableName, "mediaId")('mediaId -> id).flatMap(_.toOption)
 
   def put(lease: MediaLease) =
-    ScanamoAsync.put[MediaLease](client)(tableName)(lease.copy(id=Some(UUID.randomUUID().toString)))
```
mbarton: Nice one on pushing this up out of the DB 👍

common-lib/src/main/scala/com/gu/mediaservice/model/leases/LeasesByMedia.scala
```diff
@@ -0,0 +1,46 @@
...
+      )
+    }
+  }
+
+  def build(leases: List[MediaLease]): LeasesByMedia = {
+    def sortLease(a: MediaLease, b: MediaLease) =
+      a.createdAt.isAfter(b.createdAt)
+
+    val sortedLeases = leases
+      .sortWith(sortLease)
```
mbarton: Minor: I'm not sure from reading this which order it is sorting in (ascending or descending). I think it's descending?

Is it clearer with sortBy?

```scala
val lastModified = leases.sortBy(_.createdAt).reverse.headOption
```

common-lib/src/main/scala/com/gu/mediaservice/model/leases/MediaLease.scala
```diff
@@ -55,17 +53,13 @@ case class MediaLease(
     case _ => this
   }
 
-  private def withValidEndDateField: MediaLease = if (access == AllowSyndicationLease) {
-    this.copy(endDate = None) // an allow-syndication cannot end
```
mbarton: Minor: I actually prefer having things on newlines just because I can reliably put a breakpoint on them. Not a big deal though

marialivia16: Already moved to a new line :)

kahuna/public/js/leases/leases.js
marialivia16: Do you mean, have a `current` boolean field for the lease?

mbarton: Yeah something like that

mbarton: A flag meaning `isUse`

PR #2341, opened 2018-10-31
More leases cleaning up

marialivia16:
Third time is lucky 🍀

### DIFF
--- a/kahuna/public/js/preview/image.html
+++ b/kahuna/public/js/preview/image.html
@@ -117,10 +117,6 @@
                                 readonly="ctrl.selectionMode">
             </gr-archiver-status>
 
-            <div class="bottom-bar__action bottom-bar__action--lease" ng:if="ctrl.leaseStatus(ctrl.image)">
-                <gr-icon>today</gr-icon>
-            </div>
-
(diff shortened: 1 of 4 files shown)


kahuna/public/js/services/api/leases.js
```diff
@@ -113,17 +113,6 @@ leaseService.factory('leaseService', [
       return getLeasesRoot().follow('by-media-id', {id: image.data.id}).get();
     }
 
-    function allowedByLease(image) {
```
mbarton: I'm guessing this is just unused?

kahuna/public/js/preview/image.html
```diff
@@ -117,10 +117,6 @@
                                 readonly="ctrl.selectionMode">
             </gr-archiver-status>
 
-            <div class="bottom-bar__action bottom-bar__action--lease" ng:if="ctrl.leaseStatus(ctrl.image)">
```
mbarton: Lets quickly catch up IRL about what this icon used to do

mbarton: Agreed IRL this is dead code

PR #2342, opened 2018-11-02
Fix inferring of syndication rights in photoshoots

marialivia16:
There were issues when moving images with syndication rights from no album to an album because we were using a for comprehension to chain operations.

Also issues when an image in an album is the first one to receive syndication rights. In this case we were querying for other images in the album that have inferred rights, when we actually want to get all the other images that have either inferred right or null syndication rights.


marialivia16:
[Issues with inferring rights in syndication photoshoots](https://trello.com/c/tAmlUjMK/572-issues-with-inferring-rights-in-syndication-photoshoots)

Issue #2221, opened 2018-07-13
Perform metadata reconcilliation


paperboyo:
We should observe [metadata digests](https://wwwimages2.adobe.com/content/dam/acom/en/devnet/xmp/pdfs/XMP%20SDK%20Release%20cc-2016-08/XMPSpecificationPart3.pdf#page=53&zoom=auto,-263,758), deal with arrays, support [newest schemas](http://www.iptc.org/std/photometadata/specification/IPTC-PhotoMetadata#iptc-extension-schema-1-4-specifications).

We started reading all XMP metadata (yay!). We do some [date reconcilliation](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala#L29). ~We should perform reconcilliation for other fields in order to be able to have better quality data (spelling, location granularity, non-truncated fields and more).~ Done in https://github.com/guardian/grid/pull/2344.

Useful links at the bottom of [here](https://github.com/guardian/grid/issues/1117#issuecomment-149366878).

paperboyo:
Another example cropped up (82c98035d7dd8f1b50dbdf4f544b9fe08d8fe11a), where reading iptc:caption/abstract removes curly quotes (and, incorrectly, assumes them to be – illegal – Unicode for lack of charset definition), while reading xmp:dc:description would give us correct data.

paperboyo:
Another example is 3790aa49805edf47cc0fe915e24a54f9082e87c4. `iptc.Credit` (as it goes over 32 char limit) [contains](https://github.com/guardian/grid/issues/532) `"(Credit too long, see caption)"`, while `xmp.photoshop:Credit` correctly records `"Getty Images for Hong Kong Tourism Board"`.

Fixes https://github.com/guardian/grid/issues/532.

paperboyo:
Another (famous) example: location in `iptc` is truncated to `iptc.Sub-location:"Royal Court Theatre, Sloane Squa"` while it’s fine in `xmp.Iptc4xmpCore:Location | "Royal Court Theatre, Sloane Square"`.

paperboyo:
We should still 

> We should observe metadata digests, deal with arrays, support newest schemas.

but closing this for now as it’s mostly done.

PR #2344, opened 2018-11-02
Prefer XMP to IPTC

paperboyo:
So this is a first foray into https://github.com/guardian/grid/issues/2221. It’s wrong for many reasons, but it’s less wrong than what we do currently.

In future we should observe [metadata digests](https://wwwimages2.adobe.com/content/dam/acom/en/devnet/xmp/pdfs/XMP%20SDK%20Release%20cc-2016-08/XMPSpecificationPart3.pdf#page=53&zoom=auto,-263,758), deal with arrays, support [newest schemas](http://www.iptc.org/std/photometadata/specification/IPTC-PhotoMetadata#iptc-extension-schema-1-4-specifications).

But this here should already help us read more correct data.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
@@ -35,7 +35,7 @@ object ImageMetadataConverter {
                             fileMetadata.exif.get("Image Description"),
       credit              = fileMetadata.xmp.get("photoshop:Credit") orElse
                             fileMetadata.iptc.get("Credit"),
-      // FIXME: Have a way of dealing with arrays, like [1] here. Not much of a practical problem in our corpus, but still..
+      // FIXME: Have a way of dealing with arrays, like [1] here.
       byline              = fileMetadata.xmp.get("dc:creator[1]") orElse
                             fileMetadata.iptc.get("By-line") orElse
                             fileMetadata.exif.get("Artist"),
@@ -45,12 +45,10 @@ object ImageMetadataConverter {
                             fileMetadata.iptc.get("Headline"),
       copyrightNotice     = fileMetadata.xmp.get("dc:Rights") orElse
                             fileMetadata.iptc.get("Copyright Notice"),
-      // FIXME: why default to copyrightNotice again?
-      // I don't get what you mean, Mate. This makes no sense: our copyright and copyrightNotice fields
-      // read from equivalent fields in a stupid order. Not fixing, cause wrong logic is wide-spread.
+      // FIXME: our copyright and copyrightNotice fields should be one field (they read from equivalent fields).
       copyright           = fileMetadata.exif.get("Copyright") orElse
                             fileMetadata.iptc.get("Copyright Notice"),
-      // Another conflation of two separate fields, based on bad habits of our suppliers. We shouldn't do that:
+      // Here we combine two separate fields, based on bad habits of our suppliers.
       suppliersReference  = fileMetadata.xmp.get("photoshop:TransmissionReference") orElse
                             fileMetadata.iptc.get("Original Transmission Reference") orElse
                             fileMetadata.xmp.get("dc:title[1]") orElse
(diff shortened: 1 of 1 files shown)


paperboyo:
Tested on TEST. Fixes (not properly, but practically) all examples in https://github.com/guardian/grid/issues/2221.

common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
```diff
@@ -91,11 +115,11 @@ object ImageMetadataConverter {
 
 }
 
-// TODO: add category
+// TODO: add category - this is done, I think
```
currysoup: Where? If this is done then just remove the TODO

common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
```diff
@@ -30,22 +30,46 @@ object ImageMetadataConverter {
...
-      byline              = fileMetadata.iptc.get("By-line"),
-      bylineTitle         = fileMetadata.iptc.get("By-line Title"),
-      title               = fileMetadata.iptc.get("Headline"),
-      copyrightNotice     = fileMetadata.iptc.get("Copyright Notice"),
+      description         = fileMetadata.xmp.get("dc:description[1]") orElse
+                            fileMetadata.iptc.get("Caption/Abstract") orElse
+                            fileMetadata.exif.get("Image Description"),
+      credit              = fileMetadata.xmp.get("photoshop:Credit") orElse
+                            fileMetadata.iptc.get("Credit"),
+      // FIXME: Have a way of dealing with arrays, like [1] here. Not much of a practical problem in our corpus, but still..
```
currysoup: If we can find some examples of `xmp` arrays then we could come up with some logic around which element of the array we should pick. I suspect our problem will be if there are multiple people in the byline they might appear in different array elements and as such we'll just drop everyone except the first with this code.

common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
```diff
@@ -30,22 +30,46 @@ object ImageMetadataConverter {
...
+                            fileMetadata.iptc.get("Headline"),
+      copyrightNotice     = fileMetadata.xmp.get("dc:Rights") orElse
+                            fileMetadata.iptc.get("Copyright Notice"),
       // FIXME: why default to copyrightNotice again?
-      copyright           = fileMetadata.exif.get("Copyright") orElse fileMetadata.iptc.get("Copyright Notice"),
-      suppliersReference  = fileMetadata.iptc.get("Original Transmission Reference") orElse fileMetadata.iptc.get("Object Name"),
-      source              = fileMetadata.iptc.get("Source"),
-      specialInstructions = fileMetadata.iptc.get("Special Instructions"),
+      // I don't get what you mean, Mate. This makes no sense: our copyright and copyrightNotice fields
+      // read from equivalent fields in a stupid order. Not fixing, cause wrong logic is wide-spread.
```
currysoup: This comment is confusing to me - perhaps be a bit more descriptive than conversational. Something like "Picture providers often will put copyright information in the copyright notice field so we have to default to that in order to be safe"?

common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
paperboyo: There are some arrayed fields in XMP (and structured ones, too). I have checked and taking just the first here is quite safe given the quality of data. But it is not correct and may bite us one day. Writing it properly (incl. deduping of repeats) is beyond my skillz. Let’s hope we do it one day.

common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
paperboyo: Here we invented two fields for one field (expressed in different schemas). We should just call it `copyrightNotice` and replace all uses of `copyright` with it. But this was too scary for me.

PR #2345, opened 2018-11-05
Refactoring syndication rights inference logic

marialivia16:
Following the issues we've seen with the inference of syndication rights in albums, I took some time to rethink the logic (see image below). Previously every update of an image was a message sent to thrall, which caused a lot of inconsistencies, resulting in some images in albums missing their inferred rights. 

syndcation_rights_logic


mbarton:
The refactoring looks great, much much clearer.

As discussed IRL lets throw it on TEST and see how it does, especially with removing images from a shoot and the resulting loss of rights inference on the other images in the shoot.

We should also look to see if we can remove the `setRefresh(true)` calls on the queries, especially for [applyImageMetadataOverride](https://github.com/guardian/grid/blob/73e1d6c08c3b9301c83d55264f6306d8ea3b46d1/thrall/app/lib/ElasticSearch.scala#L249) since that will be called for things other than syndication. Since ElasticSearch is eventually-consistent we should avoid to much dependence on `setRefresh` that tries to pretend otherwise.

common-lib/src/main/scala/com/gu/mediaservice/model/Image.scala
```diff
@@ -30,6 +30,8 @@ case class Image(
 ) {
   def rcsPublishDate: Option[DateTime] = syndicationRights.flatMap(_.published)
 
+  def isInferred: Boolean = syndicationRights.forall(_.isInferred == true)
```
mbarton: Nitpicking but `isInferred` might be a confusing name on the whole `Image`. Maybe `hasInferredRights`?

thrall/test/syndication/SyndicationRightsOpsTest.scala
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    "Images in a photoshoot" - {
+      "correctly infer rights in the photoshoot when an image with syndication rights is added" in {
+        val photoshootTitle = Photoshoot(s"photoshoot-${UUID.randomUUID()}")
+
+        withPhotoshoot(photoshootTitle) { images =>
+          withImage(imageWithSyndRights) { imageWithRights =>
+            whenReady(syndRightsOps.upsertOrRefreshRights(image = imageWithRights, previousPhotoshootOpt = None, currentPhotoshootOpt = Some(photoshootTitle))) { _ =>
+              Thread.sleep(5000)
```
mbarton: As discussed IRL, lets try to remove the sleeps in the test

common-lib/src/main/scala/com/gu/mediaservice/model/Image.scala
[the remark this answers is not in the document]
akash1810: the comment came back!

PR #2347, opened 2018-11-08
Set shoot query limit to 200

RichieAHB:
Prior to this, when querying a photo shoot only 10 images were being returned. This meant that updating inferred syndication rights would only happen to 10 images when an image _with_ rights was moved to that album. As discussed IRL with @marialivia16 200 seems like a good upper bound ... for now.


LATaylor-guardian:
👍

PR #2350, opened 2018-11-20
Use the current check to choose validation message

Reettaphant:
Fixes a bug described in this trello card: https://trello.com/c/KIMlXl6f/518-grid-invalidreasons-doesnt-account-for-allow-leases

When we were choosing which error messages to display we were using the `invalid` field which is not the check we need to do to determine the validity of the image.


akash1810:
👍 great work!

PR #2351, opened 2018-11-20
Remove old static IAM credentials from dev template

mbarton:
We've been using Janus profile credentials (`media-service`) since the Play 2.6 upgrade (#2124)


akash1810:
🎉 do we need to update any readmes?

mbarton:
I don't think it was even mentioned before (oops!).

I've added it to https://github.com/guardian/grid/blob/mbarton/delete-static-iam/docs/02.01-dev-setup.md

mbarton:
@akash1810 ok to merge?

PR #2353, opened 2018-11-20
Move play code out of common

mbarton:
This will allow us to turn Thrall into a lambda that shares code with the normal grid apps. @aug24 started this work but found that the size of the lambda artifact was too big, mainly due to `common-lib` depending on the whole Play Framework. He then got a POC working in #2294 by copy-pasting code. This PR is the start of putting that work into production, sharing the code directly.

It's a big PR unfortunately but it should be quite straight forward to follow. I've described my approach below and have also posted comments inline to explain decisions.

Unfortunately there is code in `common-lib` that depends on Play Results ([ArgoHelpers](https://github.com/guardian/grid/blob/mbarton/remove-play-from-common/play-common-lib/src/main/scala/com/gu/mediaservice/lib/argo/ArgoHelpers.scala) for example). It would make for a much bigger PR if we rewrote that logic in each service so *for the moment* I have created a new project `play-common-lib` for these. The imaging helper classes (eg [ImageMagick](https://github.com/guardian/grid/blob/mbarton/remove-play-from-common/play-common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/im4jwrapper/ImageMagick.scala), [ExifTool](https://github.com/guardian/grid/blob/mbarton/remove-play-from-common/play-common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/im4jwrapper/ExifTool.scala)) are also there since they are shared between `cropper` and `image-loader`.

I've also removed [box](https://github.com/guardian/box) and Akka from the common library. Where we had common code for scheduling I have moved the Akka `scheduleUpdate` calls to the apps themselves. The [ThrallMessageConsumer](https://github.com/guardian/grid/blob/mbarton/remove-play-from-common/thrall/app/lib/ThrallMessageConsumer.scala) used to create a new `ActorSystem` and then only use it once for an initial scheduling call, passing in a separate executor anyway. I've simplified this to just use the executor directly.

I've also removed some configuration helpers (that depended on the Configuration class in Play). Instead we just read from the properties directly.


mbarton:
We have decided to postpone turning Thrall into a lambda since it complicates running the Grid locally. I will close this PR and look to re-open it once we have completed the ES6 migration.

PR #2354, opened 2018-11-21
Turn on GC logging for ElasticSearch

sihil:
The interesting changes are in [the last commit](https://github.com/guardian/grid/commit/c210b1a21fde899165d74b1a3452c47d6733a3cd).

In here we modify the `elasticsearch.in.sh` script used for customisations to tweak the way GC logs are stored. We then change the upstart script to enable the logging to work. Finally we change the build script to create a tarball with the replacement `elasticsearch.in.sh` script.

This is inspired by [changes made by Ophan](https://github.com/guardian/ophan/blob/master/elasticsearch/conf-override/usr/share/elasticsearch/bin/elasticsearch.in.sh#L74).

In addition this migrates the build from legacy riff-raff to modern riff-raff and I've set up a build in TeamCity (presumably it's not had one since the migration to TeamCity). The deploy used to actually rotate the ES cluster, but I've disabled this for the time being.


mbarton:
The ES6 cloudformation etc is in [editorial-tools-platform](https://github.com/guardian/editorial-tools-platform/tree/master/grid/elasticsearch). I've added a TODO to ensure we've enabled GC logs on them as well as part of the migration work

PR #2356, opened 2018-11-21
Request query/latency logging

mbarton:
Add `duration` as a separate field in the logs. Allows us for filter for requests that have taken longer than a given amount which will help investigate user reported slowness.


common-lib/src/main/scala/com/gu/mediaservice/lib/play/RequestLoggingFilter.scala
```diff
@@ -5,27 +5,39 @@ import akka.stream.Materializer
...
     val originIp = rh.headers.get("X-Forwarded-For").getOrElse(rh.remoteAddress)
     val referer = rh.headers.get("Referer").getOrElse("")
     result onComplete {
       case Success(response) =>
         val length = response.header.headers.getOrElse("Content-Length", 0)
-        logger.info(s"""$originIp - "${rh.method} ${rh.uri} ${rh.version}" ${response.header.status} $length "$referer" ${stopWatch.elapsed}ms""")
+
+        val elapsed = System.currentTimeMillis() - start
+        val markers = MarkerContext(appendEntries(Map(
+          "duration" -> elapsed
```
sihil: You want to take this opportunity to pull anything else out here e.g. status code?

mbarton: I've added more fields

PR #2357, opened 2018-11-22
Depend on smallest possible set of AWS libraries

sihil:
I got bored of waiting for the entire AWS world to download so thought I'd improve this. Hopefully this will also improve build and deploy sizes and times.


franziskas:
Might be worth adding a comment so that no-one well meaning tries to revert this again?

mbarton:
Nice one. I'd also done this in https://github.com/guardian/grid/pull/2353 but happy to merge this first

sihil:
I think separate dependencies is the most widely used pattern now. Not using it here was the exception so I don't feel the need to document it.

sihil:
Weirdly this hasn't actually built. Suspect the trigger patterns in teamcity are wrong. I'll look at that before merging.

PR #2363, opened 2018-12-04
Revert "Ability to pick illustrator publication"

mbarton:
We are seeing `Error: invalid-form-data`. Reverting pending an investigation.

Reverts guardian/grid#2362.


mbarton:
Thanks @akash1810. @paperboyo is just having a quick look before I deploy

PR #2364, opened 2018-12-05
Removing unused persistence identifier configuration field.

tonytw1:
This field does not appear to be used. The code attempts the deletion and catches failures rather than making a decision before deleting.


mbarton:
@tonytw1 good to merge this?

tonytw1:
Planning to merge this on Tuesday, so I can observe that happens next.

PR #2367, opened 2018-12-06
Rename create index method

tonytw1:
Makes it explicit that this call only makes a specific type of index (one with image mappings).


common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchClient.scala
```diff
@@ -56,10 +56,10 @@ trait ElasticSearchClient {
     val indexExists = client.admin.indices.prepareExists(index)
                         .execute.actionGet.isExists
 
-    if (!indexExists) createIndex(index)
+    if (!indexExists) createImageIndex(index)
```
franziskas: should the `!indexExists` be renamed as well? Should the `deleteIndex` method be renamed?

PR #2368, opened 2018-12-06
Enable optional decompression

regiskuckaertz:
Reads off the Kinesis stream and tries to decompress the payload. Crier will soon be using ZStandard to compress messages dispatched on the stream; this changes makes sure it is able to handle both compressed and uncompressed events.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -47,6 +47,7 @@ lazy val commonLib = project("common-lib").settings(
     "com.amazonaws" % "aws-java-sdk-kinesis" % awsSdkVersion,
     "org.elasticsearch" % "elasticsearch" % "1.7.6",
     "com.gu" %% "box" % "0.2.0",
+    "com.gu" %% "thrift-serializer" % "3.0.0",
     "org.scalaz.stream" %% "scalaz-stream" % "0.8.6",
     "com.drewnoakes" % "metadata-extractor" % "2.11.0",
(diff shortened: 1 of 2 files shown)


regiskuckaertz:
Yes, better safe than sorry :-) How can I do that?

mbarton:
I would deploy the `usage` service to your branch to the `TEST` stage (the Grid uses `TEST` instead of `CODE`). Then create an article in Composer CODE, add an image and check that the usage appears correctly next to the image in the Grid.

I'm more than happy to pair with you as well

regiskuckaertz:
No that's fine, I'll do it next week ✌️

regiskuckaertz:
ha, I just tried deploying but it failed. Is it possible the CF is outdated there?

regiskuckaertz:
I just need to learn to read. `TEST` is _not_ `CODE` 😆

mbarton:
Haha @regiskuckaertz you are not the first and will not be the last :)

Happy to pair on it as well

franziskas:
@regiskuckaertz I didn't dare to ask if you made the same mistake that I always make 🤦‍♀️

regiskuckaertz:
The story of my life 😅

regiskuckaertz:
Works!

screen shot 2018-12-10 at 16 54 50

(I'm not good with titles 😉 )

mbarton:
Thanks @regiskuckaertz!

PR #2369, opened 2018-12-07
Update SBT dependencies

mbarton:
Update libraries identified as vulnerable by Snyk.

The bulk of the code changes are due to the update from `editorial-permissions-client` from `0.8` to `2.0`. The new version returns results directly rather than wrapped in futures.

I've also updated the healthcheck for the services that use permissions. This will make sure that permissions have been downloaded before the service is marked as healthy in the load balancer. Practically this will make little difference for how we use permissions now since they just "add" to a normal users permission but will be required in the future if we move to using permissions as the definition for which users can access the Grid entirely.


mchv:
@mbarton I have added 3 more update to your changes.

I think you can as well easily update `play` to `2.6.20` and `sbt` to `1.2.8`

mbarton:
Thanks @mchv I have made those changes

PR #2373, opened 2018-12-12
Elastic6 work in progress

tonytw1:
For ongoing review; do not merge.


common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch6/ElasticSearchClient.scala
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  def ensureIndexExists(index: String): Unit = {
+    Logger.info("Checking index exists…")
+
+    val eventualIndexExistsResponse: Future[Response[IndexExistsResponse]] = client.execute {
+      indexExists(index)
+    }
+
+    val indexExistsResponse = Await.result(eventualIndexExistsResponse, tenSeconds)
```
mbarton: I'd make this method return a `Future` and push the `Await` all the way up out of the client into the top-level calling code, which here I think is `MediaApiComponents` and `ThrallComponents`

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch6/ElasticSearchClient.scala
```diff
@@ -0,0 +1,120 @@
...
+  def deleteIndex(index: String) {
+    ???
+  }
+
+  def getCurrentAlias: Option[String] = {
+    ensureIndexExists(initialImagesIndex)
+    None // TODO
+  }
+
+  // TODO do not understand why this step is required; can't you just index to the alias and let elastic sort it out?
```
mbarton: Agreed, I'm not sure we get much value from having aliases compared to just having a parameterised index name in configuration

thrall/app/lib/ElasticSearch6.scala
```diff
@@ -0,0 +1,511 @@
...
+  lazy val cluster = config("es6.cluster")
+
+  lazy val shards = config.elasticsearch6Shards
+  lazy val replicas = config.elasticsearch6Replicas
+
+  @Deprecated
+  lazy val clientTransportSniff = false
+
+  def indexImage(id: String, image: JsValue)(implicit ex: ExecutionContext): List[Future[ElasticSearchUpdateResponse]] = {
+    // TODO doesn't match the legacy functionality
```
mbarton: How does it not match the old functionality?

thrall/app/lib/ThrallConfig.scala
```diff
@@ -28,6 +28,11 @@ class ThrallConfig(override val configuration: Configuration) extends CommonConf
         "App"   -> Seq(elasticsearchApp)
       ))
 
+
+  lazy val elasticsearch6Host: String = properties.getOrElse("es6.host", "localhost")     // TODO how to discover the 6 cluster in AWS
```
mbarton: We'll have a load balancer in front of the ES cluster so could do a describe by tag?

thrall/app/lib/ThrallConfig.scala
```diff
@@ -28,6 +28,11 @@ class ThrallConfig(override val configuration: Configuration) extends CommonConf
...
+  lazy val elasticsearch6Shards: Int = if (isDev) 1 else properties.getOrElse("es6.shards", "1").toInt
```
mbarton: Maybe best not to have these default in PROD but fail if there's no config?

tonytw1: Yep; sounds sensible.

build.sbt
```diff
@@ -153,7 +156,7 @@ val testSettings = Seq(
     s"docker-compose --file docker-compose-test.yml --project-name grid-test up -d".!
 
     // This is needed to ensure docker has had enough time to start up
-    Thread.sleep(5000)
+    Thread.sleep(30000)
```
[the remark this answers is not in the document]
tonytw1: Thanks for the suggestion. I'll look into this as that long sleep isn't great.

thrall/app/lib/ElasticSearchRouter.scala
```diff
@@ -0,0 +1,57 @@
...
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class ElasticSearchRouter(primary: ElasticSearchVersion, secondary: ElasticSearchVersion) extends ElasticSearchVersion {
+
+  val primaryThenSecondary = Seq(primary, secondary)
+
+  override def indexImage(id: String, image: JsValue)(implicit ex: ExecutionContext): List[Future[ElasticSearchUpdateResponse]] = primaryThenSecondary.map{ i =>
+      i.indexImage(id, image)
+    }.head
```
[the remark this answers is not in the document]
tonytw1: +1

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch6/ElasticSearchClient.scala
tonytw1: Agree that returning the Future to the calling code is sensible choose to try to match the existing Elastic1 behaviour which appears to be a no return and blocking.
The desire was to limit changes to the calling application but I'm ok to go either way on this.

thrall/app/lib/ElasticSearch6.scala
tonytw1: This implementation is lifted from the lambda branch. It's behaviour appears different to my untrained eye but it could be the same logic with a refactor applied; equally it could be that the lambda branch devs spotted a generalisation which I've missed.

Resolve by porting the Elastic1 implementation of indexImage.

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -290,7 +284,6 @@ object Mappings {
           "usagesLastModified" -> dateFormat,
           "leases" -> leasesMapping,
           "collections" -> collectionMapping,
-          "suggestMetadataCredit" -> simpleSuggester
```
tonytw1: This was a duplicate; the same line appears above in the same block.

PR #2375, opened 2018-12-13
Improve logic to convert true colour PNGs without transparency to JPEG

paperboyo:
Improves current logic which I’m not sure even works.
Would be nice to merge #2371 first, as currently logic there provides false positives (see comments referenced there for details).

For the future, we should improve it further, so that it converts to JPEGs not _images_ but _crops_ where there is no transparency.

- [ ] Test on TEST

_this is another occurrence of @itsibitzi impersonating me, so that he can hide his code behind my ignorance!_


paperboyo:
> Improves current logic which I’m not sure even works.

Current logic do work (and is improved with https://github.com/guardian/grid/pull/2371). This, even though it reads more correct, doesn’t, so: closing (with a heavy heart).

Issue #2374, opened 2018-12-13
./cloud-formation/scripts/post-dev-stack-creation.sh: line 18: md5hash: command not found


CaerphillyMediaLtd:
Why does it say md5hash there? As far as I know that's not a valid command on any OS?

CaerphillyMediaLtd:
I also get all other kinds of errors:

```
./cloud-formation/scripts/post-dev-stack-creation.sh media-service-DEV
Uploading config to stack media-service-DEV
./cloud-formation/scripts/post-dev-stack-creation.sh: line 18: md5hash: command not found
Traceback (most recent call last):
  File "/usr/bin/aws", line 27, in <module>
    sys.exit(main())
  File "/usr/bin/aws", line 23, in main
    return awscli.clidriver.main()
  File "/usr/lib/python2.7/site-packages/awscli/clidriver.py", line 59, in main
    rc = driver.main()
  File "/usr/lib/python2.7/site-packages/awscli/clidriver.py", line 193, in main
    command_table = self._get_command_table()
  File "/usr/lib/python2.7/site-packages/awscli/clidriver.py", line 102, in _get_command_table
    self._command_table = self._build_command_table()
  File "/usr/lib/python2.7/site-packages/awscli/clidriver.py", line 122, in _build_command_table
    command_object=self)
  File "/usr/lib/python2.7/site-packages/botocore/session.py", line 645, in emit
    return self._events.emit(event_name, **kwargs)
  File "/usr/lib/python2.7/site-packages/botocore/hooks.py", line 356, in emit
    return self._emitter.emit(aliased_event_name, **kwargs)
  File "/usr/lib/python2.7/site-packages/botocore/hooks.py", line 228, in emit
    return self._emit(event_name, kwargs)
  File "/usr/lib/python2.7/site-packages/botocore/hooks.py", line 211, in _emit
    response = handler(**kwargs)
  File "/usr/lib/python2.7/site-packages/awscli/customizations/preview.py", line 69, in mark_as_preview
    service_name=original_command.service_model.service_name,
  File "/usr/lib/python2.7/site-packages/awscli/clidriver.py", line 318, in service_model
    return self._get_service_model()
  File "/usr/lib/python2.7/site-packages/awscli/clidriver.py", line 335, in _get_service_model
    api_version = self.session.get_config_variable('api_versions').get(
  File "/usr/lib/python2.7/site-packages/botocore/session.py", line 207, in get_config_variable
    logical_name)
[...]

PR #2376, opened 2018-12-14
fix post stack script

akash1810:
We've introduced a regression. [Previously](https://github.com/guardian/grid/commit/dd3fce2c61c35312e6ae8a33de751ba0b823be2a#diff-bf3ab15dc8a1ee8a37466f5bf987288e), we had aliased md5hash to be `md5sum` on linux and `md5` on osx.

As the [docs suggest osx users install the brew formula md5Sum](https://github.com/guardian/grid/blob/master/docs/02.01-dev-setup.md#optional-requirements), we should use the `md5sum` across platforms.


akash1810:
Fixes #2374

PR #2379, opened 2019-01-09
show syndication rights on image page

akash1810:
This is to signal if information has been obtained from RCS for an individual image.

It's quite dirty and ugly, but we can address that later. Hopefully this is enough to help diagnose issues that have been raised.

Example screenshot:
image


paperboyo:
A thought: if this describes all syndication statuses, then maybe putting their $ icons to the left of text would help users familiarise themselves with what they mean in browser, where the text hides behind tooltip?

akash1810:
This component is currently to indicate if RCS has told us anything about an image, rather than the syndication status. Although related, the syndication status has more states (eligible, reviewable, queued, sent).

Maybe a better way of showing this information is a progress bar/table with four sections?

| Criteria  | Status  |
|---|---|
| RCS knows about image |  Yes |
| RCS says its syndicatable | Yes  |
| Syndication lease given  | Yes |
| Image has been sent to syndication partner | Queued |

PR #2382, opened 2019-01-14
Appease Snyk, for it is an angry god

mbarton:
Snyk... what's up my friend? You're always so mad and sometimes you hide it. Like showing completely green on #2369 and then popping up a bunch of new warnings on master.

I get you're just doing your job and warnings about library vulnerabilities are important for us to keep on top of but the way you're going about it is just kind of... difficult you know?

Hopefully this pointless change to build.sbt will convince you that we are packaging up-to-date and non-vulnerable libraries. If not, we should chat some time.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -32,7 +32,7 @@ lazy val commonLib = project("common-lib").settings(
     "com.typesafe.play" %% "play" % "2.6.20", ws,
     "com.typesafe.play" %% "play-json-joda" % "2.6.9",
     "com.typesafe.play" %% "filters-helpers" % "2.6.20",
-    "com.gu" %% "pan-domain-auth-play_2-6" % "0.7.1",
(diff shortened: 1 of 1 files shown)


mbarton:
OK so a lot of this madness is caused because [Panda does try to bump Jackson](https://github.com/guardian/pan-domain-authentication/blob/master/project/Dependencies.scala#L41) but only to version `2.9.2` which is not new enough to fix the tonne of Snyk warnings. For that we need to be on `2.9.8`.

Since the Apache HTTP client warning could also be fixed transitively (across all projects) by updating Panda we could do worse that bump that as well.

To stop my head hurting I tried to add panda to our Snyk Dashboard but it DOESN'T WORK since it is still on a Scala file build, not build.sbt!

So I suspect the approach is to sort out Panda and then get the library updates for free. Unfortunately right now that means reverting or releasing my changes, so we should take a read on that as a team.

PR #2391, opened 2019-01-24
Tweak the inferring syndication rights logic

marialivia16:
When getting the latest syndication rights make sure the new syndication rights that have been passed through. The rights on the image are not necessary the latest ones. As described in `SyndicationRightsOps.scala`, this logic is needed because ES is eventually consistent.

In addition, I've added more logging to help debugging in the future.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/Image.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/Image.scala
@@ -30,7 +30,7 @@ case class Image(
   userMetadataLastModified: Option[DateTime] = None) {
   def rcsPublishDate: Option[DateTime] = syndicationRights.flatMap(_.published)
 
-  def hasInferredSyndicationRights: Boolean = syndicationRights.forall(_.isInferred == true)
+  def hasInferredSyndicationRights: Boolean = syndicationRights.forall(_.isInferred)
 
   def syndicationStatus: SyndicationStatus = {
     val isRightsAcquired: Boolean = syndicationRights.exists(_.isRightsAcquired)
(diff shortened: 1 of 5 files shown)


common-lib/src/main/scala/com/gu/mediaservice/model/Image.scala
```diff
@@ -30,7 +30,7 @@ case class Image(
 ) {
   def rcsPublishDate: Option[DateTime] = syndicationRights.flatMap(_.published)
 
-  def hasInferredSyndicationRights: Boolean = syndicationRights.forall(_.isInferred == true)
+  def hasInferredSyndicationRights: Boolean = syndicationRights.forall(_.isInferred)
```
mchv: Nice little fix

mchv:
@marialivia16 I think it will be worth adding a test in `SyndicationRightsOpsTest` to avoid regression on this

marialivia16:
@marialivia16 We already have a test covering this case, but it only fails in production because ES is eventually consistent. 

The way this is supposed to work is ([here](https://github.com/guardian/grid/pull/2345#issue-228289779) is a more elaborate diagram of the logic):
1. given an image from a photoshoot and some syndication rights
2. update the syndication rights for that image
3. fetch the latest syndication rights from the album
4. update all the images in the album to have the latest syndication rights from step 3

What happens sometimes is that when we do step 3, the update from step 2 hasn't propagated yet. That is why we're adding the current rights manually to the result from step 3. 

I hope this makes some sense, happy to chat about it if it doesn't.

mchv:
@marialivia16 I think you can reproduce this error in your test, by tweaking the way you `pause` (i.e `Thread.sleep`) and even have a test where you explicitly not propagate the update of syndication rights for that image (`step 2`)  and still do step3 and ensure you have the right results at the end.

Adding tests for eventual consistency is really powerful and could allow to find more similar issues.

mchv:
i.e what happen if you remove the `Thread.sleep(5000)`, do you expect the tests to fail or succeed?

marialivia16:
Makes sense, I've already found an issue, I'll write some more tests.

PR #2394, opened 2019-01-24
Upgrade Panda (configurable settings and OAuth providers)

mbarton:
The latest version of Panda (0.18) supports non-Google OAuth providers, including AWS Cognito. It also removes an incorrectly hard-coded check that users emails end with `guardian.co.uk`. https://github.com/guardian/pan-domain-authentication/pull/45.

This PR adds support for these to the Grid and allows the user email domain check in `validateUser` to be fully customised. If none of the new settings are present, it falls back to the old Guardian-specific defaults. The multi-factor check in `validateUser` is only performed if the multi-factor checker itself is enabled (via Pandas own configuration).

New settings:

`panda.userDomain` - the domain to user in the `email.endsWith` check in validate user
`panda.system` - the name of the system in the panda cookie
`panda.bucketName` - the bucket where the panda settings file is stored
`panda.settingsFileKey` - the key for the settings file

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Authentication.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Authentication.scala
@@ -10,6 +10,7 @@ import com.gu.mediaservice.lib.logging.GridLogger
 import com.gu.pandomainauth.PanDomainAuthSettingsRefresher
 import com.gu.pandomainauth.action.{AuthActions, UserRequest}
 import com.gu.pandomainauth.model.{AuthenticatedUser, User}
+import com.gu.pandomainauth.service.Google2FAGroupChecker
 import play.api.libs.ws.WSClient
 import play.api.mvc.Security.AuthenticatedRequest
 import play.api.mvc._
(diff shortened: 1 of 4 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Authentication.scala
```diff
@@ -63,15 +66,19 @@ class Authentication(config: CommonConfig, actorSystem: ActorSystem,
   }
 
   final override def validateUser(authedUser: AuthenticatedUser): Boolean = {
-    authedUser.user.email.endsWith("@guardian.co.uk") && authedUser.multiFactor
+    val isValidDomain = authedUser.user.email.endsWith("@" + userValidationEmailDomain)
+    val passesMultifactor = if(multifactorChecker.isEmpty) { authedUser.multiFactor } else { true }
```
tonytw1: Is the multifactorChecker.isEmpty boolean inverted?
ie. if there is a MFA checker, then check if the user has MFA else pass them.
Does this currently read is there is no MFA checker then check their MFA?

mbarton: Yes, great spot! Thanks @tonytw1, I will add this to the long list of times I've used the wrong `isEmpty/nonEmpty` and also write a test

PR #2398, opened 2019-01-29
Media API Elastic6 toggle

tonytw1:
Media API can route requests to ES1 or ES6 based on the presences of a feature toggle cookie.


media-api/app/controllers/AggregationController.scala
```diff
@@ -11,6 +11,8 @@ class AggregationController(auth: Authentication, elasticSearch: ElasticSearchVe
   extends BaseController with AggregateResponses {
 
   def dateHistogram(field: String, q: Option[String]) = auth.async { request =>
+    implicit val r = request
```
mbarton: I think you can inline this above:

```scala
  def dateHistogram(field: String, q: Option[String]) = auth.async { implicit request =>
```

media-api/app/lib/UsageQuota.scala
```diff
@@ -37,7 +40,7 @@ class UsageQuota(config: MediaApiConfig, elasticSearch: ElasticSearchVersion, sc
       waitMillis.millis)
   }.toOption.exists(_.exceeded) && FeatureToggle.get("usage-quota-ui")
 
-  def usageStatusForImage(id: String): Future[UsageStatus] = for {
+  def usageStatusForImage(id: String)(implicit request: AuthenticatedRequest[AnyContent, Principal]): Future[UsageStatus] = for {
```
mbarton: Any reason for this to be an implicit parameter rather than an explicit one?

tonytw1: Minimising the change and preserving the explicit signature which was already there as this request awareness is a temporary thing.

media-api/app/controllers/AggregationController.scala
tonytw1: Probably; if none of the downstream code already references the explicit request; hence I'm providing both.

PR #2400, opened 2019-01-29
Request correctly sized imgops preview image in Firefox

paperboyo:
Because, unlike Chrome, [Firefox differs](https://bugzilla.mozilla.org/show_bug.cgi?id=1292571) `screen.width` with zoom, we need to [treat it differently](https://stackoverflow.com/questions/30557404/get-screen-resolution-independent-of-zoom-in-firefox/34368907#34368907) to correctly request screen-sized preview image.

Thanks @mbarton and @RichieAHB (who actually wrote all this).

- [x] test on TEST

### DIFF
--- a/kahuna/public/js/imgops/service.js
+++ b/kahuna/public/js/imgops/service.js
@@ -7,10 +7,18 @@ imgops.factory('imgops', ['$window', function($window) {
 
     const lowResMaxWidth  = 800;
     const lowResMaxHeight = 800;
-
+    const isFF = !!$window.navigator.userAgent.match(/firefox/i);
 
     function getFullScreenUri(image) {
-        const { width: w, height: h } = $window.screen;
+        let { width: w, height: h } = $window.screen;
+        if (isFF) {
+            const zoom = $window.devicePixelRatio;
+            if (zoom !== 1) {
+                h = Math.round(h * zoom);
+                w = Math.round(w * zoom);
+            }
+        }
+
         return getOptimisedUri(image, { w, h, q: quality });
     }
 


RichieAHB:
Worth noting we wrote this in the GitHub editor so would be keen to make sure this was tested on CODE to make sure I didn't make any booboos.

RichieAHB:
Additionally I think this will serve large images on retina displays in Firefox (`devicePixelRatio` is `2` on retina displays) - this might not be a problem but Chrome won't get this special treatment.

paperboyo:
@RichieAHB 

> this will serve large images on retina displays in Firefox

Shit! You’re right! I just checked on 27in Retina iMac and both browsers report 2560px `screen.width` but `devicePixelRatio` of 2 at 100% zoom and up it when zooming in. `screen.width` stays put in Chrome and differs on zoom in Firefox.

[This is ridiculous!](https://github.com/w3c/csswg-drafts/issues/399) Forgetting performance considerations where we should maybe restrict image dimensions over certain limit and just wanting to serve 1:1 pixels to displays on different browsers – not sure what can we do apart from introducing even more absurd logic whereby if it’s ≥`2` we would strip `2` and just take the rest?!

Thoughts?

paperboyo:
OK. Tested on TEST, on Retinas too. This improves behaviour in Firefox on all dprs. Doesn’t change behaviour in Chrome on all dprs. Makes the browsers behave the same on non-HiDPI displays. Makes browsers widely different on HiDPI displays (Firefox gets Retina-ready images, Chrome gets half the physical resolution).
Above is true on a Mac, PC may deal with HiDPI differently (and OS has a much more granular scaling controls).

Until browsers vendors/spec writers agree, not much we can do, unless I’m wrong ¯\_(ツ)_/¯

PR #2403, opened 2019-01-30
add missing config from generation

akash1810:
We've added/changed the config in the app, however we've not kept the [config generator](https://github.com/guardian/grid/tree/master/scripts/generate-dot-properties) up to date. This adds the missing properties to the generator.

### DIFF
--- a/scripts/generate-dot-properties/config.json5
+++ b/scripts/generate-dot-properties/config.json5
@@ -44,5 +44,11 @@
 
   composer: {
     url: ''
+  },
+
+  google: {
+    tracking: {
+      id: ''
+    }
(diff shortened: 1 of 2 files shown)


akash1810:
> @akash1810 needs a little bit more context please :)

Sorry! Updated.

RichieAHB:
Does this include a default for `app.name` in `usages`?

akash1810:
> Does this include a default for `app.name` in `usages`?

it does now 😄

RichieAHB:
Nice - might be worth getting @marialivia16 and @tonytw1 to have a look at this too.

PR #2404, opened 2019-01-31
Bump panda-session to fix grid login issues (hopefully)!

RichieAHB:
This change bumps `panda-session` to `0.1.6`. We don't explicitly depend on `panda-session` but it's used by `pandular` (an angular helper that wraps `panda-session`), which actually depend on `0.1.4`. Because of `npm` we can force this to depend on `0.1.6` by explicitly declaring it here.

## The fix
So we figured out that two `419` requests would both try to reauth and each get differeny ant-forgery tokens - something that I think [Ophan had a problem with](https://github.com/guardian/ophan/pull/2622). We managed to fix this by making sure that any reauth requests (managed through loading an iframe in the page) would look for existing reauths and just wait for those instead of spawning a new request.

We implemented this and it worked! However, @mbarton then managed to [find the PR where this had already been done](https://github.com/guardian/panda-session/pull/1) in `panda-session`, it just hadn't been updated by `pandular` 😢 

## TL;DR
We believe this will stop the **BIG RED MESSAGE**!

We'll deploy this and see if that helps and then following that we can bump `panda-session` in `pandular` to do this _The Right Way_™️ ...

### DIFF
--- a/kahuna/package-lock.json
+++ b/kahuna/package-lock.json
@@ -3568,15 +3568,6 @@
           "dev": true,
           "optional": true
         },
-        "string_decoder": {
-          "version": "1.1.1",
-          "bundled": true,
-          "dev": true,
-          "optional": true,
-          "requires": {
-            "safe-buffer": "5.1.1"
-          }
-        },
         "string-width": {
           "version": "1.0.2",
           "bundled": true,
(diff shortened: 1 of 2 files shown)


akash1810:
👍

PR #2406, opened 2019-02-05
Run docker integration tests using DockerKit

mbarton:
I am trying to construct a single TeamCity build for the entire Grid and was hitting problems where we couldn't run all the tests in the project one after the other.

The issue seemed to be that SBT was running the teardown code (ie `docker-compose down`) whilst other tests were still running. We have parallel test execution disabled already so that would not help.

My solution is to run the ES containers individually for each test using [DockerKit](https://github.com/whisklabs/docker-it-scala). This does add some time overall but the build is ~6 minutes which I think is acceptable.

I was able to remove the ES6 configuration file entirely and just inject config as environment variables. This didn't seem to work for ES1 so I've kept the config mapped as a volume.

TODO:

- [x] investigate [build hanging](https://teamcity.gutools.co.uk/viewLog.html?buildId=259730&buildTypeId=EditorialTools_Grid_UnifiedGridBuild&tab=buildLog) with Guava preconditions error


mbarton:
[Use single TeamCity build and dotcom-style riff-raff.yaml for the Grid](https://trello.com/c/GZfBtr2L/20-use-single-teamcity-build-and-dotcom-style-riff-raffyaml-for-the-grid)

PR #2407, opened 2019-02-06
Batch cropping feedback

Reettaphant:
Adds better feedback for batch cropping. 

- Disable the crop button if all images in selection have full frame crops
- Tooltip explaining why button is disabled
- Crop icon spins when cropping is in progress

screen shot 2019-02-06 at 11 46 19

### DIFF
--- a/kahuna/public/js/components/gr-batch-export-original-images/gr-batch-export-original-images.js
+++ b/kahuna/public/js/components/gr-batch-export-original-images/gr-batch-export-original-images.js
@@ -9,7 +9,18 @@ batchExportOriginalImages.controller('grBatchExportOriginalImagesCtrl', [
     function($scope, $rootScope, $state, mediaCropper) {
         let ctrl = this;
 
+        const checkForFullCrops = () => ctrl.images.every(
+            image => image.data.exports.some(
+              crop => crop.specification.type === 'full'
+            )
+        );
+
+        $scope.$watch('ctrl.images', () => {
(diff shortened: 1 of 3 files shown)


paperboyo:
👏 
Lacking screenshots/anigifs, is the spinning the same as the spinning on the cropping page?

Reettaphant:
@paperboyo Sorry, forgot to attach the visual aids. The spinning is the same.

paperboyo:
Thanks! Minor: in the tooltip, I would replace `full crops` with `full frames`.

kahuna/public/js/components/gr-batch-export-original-images/gr-batch-export-original-images.js
```diff
@@ -9,6 +9,16 @@ batchExportOriginalImages.controller('grBatchExportOriginalImagesCtrl', [
     function($scope, $rootScope, $state, mediaCropper) {
         let ctrl = this;
 
+        const checkForFullCrops = () => ctrl.images.every(
+            image => image.data.exports.some(
```
akash1810: Not blocking getting this out as we can raise in a further PR, however, should we push this into the [imageAccessor](https://github.com/guardian/grid/blob/master/kahuna/public/js/services/image-accessor.js) to keep knowledge of the shape of the API in one place?

Reettaphant: Yes we should, I've made a card for it.

PR #2410, opened 2019-02-07
Fixes persistence of images

vlbee:
Images that were uploaded more than 20 days ago and had user edits but were not used were not getting persisted. This fixes persistence filters such that any image with edit changes (userMetadata) will now also be persisted. 

Unit tests have been added


paperboyo:
This is very, very sensible and long overdue! Thank you!

Will this persist images uploaded, without any edits? Does editing userMetadata include: adding a label, adding to Collection, adding Lease, adding to Photoshoot, changing Usage Rights?

vlbee:
@paperboyo We did not change any of the other persistence criteria so changes to these (label, collection, leases, photoshop, usage rights) will still cause images  to be persisted. UserMetadata changes cover user edits to things like the image credit, etc that were previously not covered by the other criteria.

paperboyo:
Replying to myself after @vlbee kindly showed me [a thing](https://github.com/guardian/grid/blob/master/media-api/app/lib/elasticsearch/impls/elasticsearch1/SearchFilters.scala#L62). The three things that will still allow for images to drop from the system are:
- labels
- leases
- uploads not otherwise touched
and, obviously, images otherwise not persisted added to non-presisted Collection.

PR #2411, opened 2019-02-07
we're not syndicating staff illustrators

akash1810:
remove them from the review list

related to https://github.com/guardian/grid/pull/2409 after getting feedback from Joanna


paperboyo:
👍 (cleaner queue = better)

PR #2414, opened 2019-02-07
Add “Daily Edit” to list of persisted collections

paperboyo:
Just because I always wanted to do this! Also, see “Per-collection auto-archiving in Grid” email from 7 Apr 2016 (sic!).

### DIFF
--- a/media-api/app/lib/MediaApiConfig.scala
+++ b/media-api/app/lib/MediaApiConfig.scala
@@ -87,7 +87,7 @@ class MediaApiConfig(override val configuration: Configuration) extends CommonCo
 
   lazy val persistedRootCollections: List[String] = properties.get("persistence.collections") match {
     case Some(collections) => collections.split(',').toList
-    case None => List("GNM Archive")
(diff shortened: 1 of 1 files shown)


media-api/app/lib/MediaApiConfig.scala
```diff
@@ -87,7 +87,7 @@ class MediaApiConfig(override val configuration: Configuration) extends CommonCo
 
   lazy val persistedRootCollections: List[String] = properties.get("persistence.collections") match {
     case Some(collections) => collections.split(',').toList
-    case None => List("GNM Archive")
+    case None => List("GNM Archive", "Daily Edit")
```
akash1810: We're reading persisted collections from config about (line 88). This line is the defaults should config be empty. I think it would be better to add this new collection to configuration.

paperboyo: Happy to do this, but couldn’t find where do we store the config…

paperboyo:
We updated PROD and TEST configs, closing.

PR #2418, opened 2019-02-11
allow the world to get objects in the image origin bucket

akash1810:
the one where crops go

### DIFF
--- a/cloud-formation/dev-template.yaml
+++ b/cloud-formation/dev-template.yaml
@@ -28,6 +28,20 @@ Resources:
         IndexDocument: index.html
         ErrorDocument: error.html
 
+  ImageOriginBucketPolicy:
+    Type: AWS::S3::BucketPolicy
+    Properties:
+      Bucket: !Ref 'ImageOriginBucket'
+      PolicyDocument:
+        Id: PublicImageRead
+        Statement:
+          - Action:
+              - s3:GetObject
+            Effect: Allow
+            Resource: !Sub 'arn:aws:s3:::${ImageOriginBucket}/*'
+            Principal:
+              AWS: '*'
+
   ConfigBucket:
     Type: AWS::S3::Bucket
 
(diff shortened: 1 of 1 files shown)


mbarton:
I'm not against this since it matches PROD behaviour but would people be suprised that their local dev crops were visible to the world?

akash1810:
> I'm not against this since it matches PROD behaviour but would people be suprised that their local dev crops were visible to the world?

Yeah, its not great, I agree. We could add a condition for [`IpAddress`](https://docs.aws.amazon.com/AmazonS3/latest/dev/example-bucket-policies.html#example-bucket-policies-use-case-3) or [`Referrer`](https://docs.aws.amazon.com/AmazonS3/latest/dev/example-bucket-policies.html#example-bucket-policies-use-case-4) to limit it to the office or Grid DEV? Not keen on the former as it wouldn't work for the community...

mbarton:
I guess we could parameterise them in the cloud formation?

RichieAHB:
That is a very good point @mbarton

akash1810:
Updated in https://github.com/guardian/grid/pull/2418/commits/535dd31bc617e2a163f701a898580d4f67f65bfd. Not especially keen on it though as we diverge across stacks now...

Setting the referrer header will mean other Tools don't be able to link to DEV images, which isn't ideal.

tonytw1:
Assuming that 'Image origin' referrs to the 'crops' bucket.

The UI accesses images in the thumbnail and full images buckets using authenticated S3 calls. UI references to the crop thumbnails are not authenticated because the crops bucket is intended to be public docroot.

If the UI authenticated it's crop bucket calls, even though the crop bucket is probably public in most deployments, would that resolve this while still accommodating the private crops bucket use case?

mbarton:
Closing since the original problem was fixed by changing config so it's no longer required

PR #2421, opened 2019-02-11
Send messages for multiple crops

RichieAHB:
Prior to this PR, when batch cropping an image in an iframe, the parent window would receive a `message` that contained the `image` and `crop` of the first image where `createFullCrop` resolved first.

This doesn't _seem_ useful, you can't guarantee which of these images which actually come through and the assumption is that if you're batch cropping you probably want _all_ of them.

This PR changes batch cropping to post a different type of `message` when batch cropping happens, posting an array of `{ image, crop }` rather than just an `{ image, crop }`.

Unfortunately, our messages don't have any form of `type` property added to discriminate between them but given consumers could (and do) receive all sorts of messages in the `message` event listener we can be pretty that they discriminate somehow and that this change won't break any consumers.

From the tools I can find (Media Atom Maker, Fronts, Composer and others) they all seems to validate a grid message by doing something like `message.image && message.crops && message.image.data` or similar.

The new message has a different shape so these checks should now fail meaning batch cropping won't randomly insert one image and instead consumers will have to add more code to handle batching explicitly.

### DIFF
--- a/kahuna/public/js/components/gr-batch-export-original-images/gr-batch-export-original-images.js
+++ b/kahuna/public/js/components/gr-batch-export-original-images/gr-batch-export-original-images.js
@@ -38,22 +38,21 @@ batchExportOriginalImages.controller('grBatchExportOriginalImagesCtrl', [
         };
 
         function cropImages() {
-            ctrl.cropping = true;
-            ctrl.needsConfirmation = false;
-            const cropImages = ctrl.images.map(image => {
-              mediaCropper.createFullCrop(image).then(crop => {
-                  //Global notification of action
-                  $rootScope.$emit('events:crop-created', {
-                      image: image,
-                      crop: crop
-                  });
-              });
-            });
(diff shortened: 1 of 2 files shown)


akash1810:
Lets also run this by @blishen for InCopy

paperboyo:
InCopy shouldn’t be affected as one doesn’t drag in crops, but images. But! It’s definitely worth checking with the Boss to see how hard would be to add dragging in multiple images to InDesign!

RichieAHB:
Good shout @akash1810 will ship it to CODE and then I can test on Composer LOCAL using CODE Grid - right?!

akash1810:
> Good shout @akash1810 will ship it to CODE and then I can test on Composer LOCAL using CODE Grid - right?!

Yep. We can also edit Composer's config at `/etc/gu/flexible-composerbackend.properties` to get it to talk to a local instance of Grid (incase Grid TEST is busy).

PR #2423, opened 2019-02-13
Attempting to account for nested query usage.dateAdded.

tonytw1:
All previous nested types where copied to the root in Elastic1, possibly to make the index match the query DSL?


tonytw1:
As far as I can tell queries against the usages.dateAdded field are actually executing against the 'dateAdded' field in ES1. This is working because the usages field has a copy to parent mapping. 

This copy to parent is not supported in ES6; hence the need to query on 'usages.dateAdded'

There's tension here between the UI query parsing DSL concept of 'Field' and elastic search 'fields'.
This seemed like the least disruptive way in and the usagesFields / metadataFields mappings could be how the original developers intended for us to resolve these elastic search nested fields.

The potential regression is that this changes where ES1 looks for dateAdded but in theory 'usage.dateAdded' and 'dateAdded' should already be copies of each other.

PR #2424, opened 2019-02-14
refactor SyndicationFilter

akash1810:
~~Refactors `SyndicationFilter` for brevity, clarity and correctness. Also adds more test images to exhaust the elasticsearch queries. This is after we've had some images returning in the review queue even though they've had a lease added to them. Looking at the previous queries, this is because the date ended query was a bit naive in that it was looking at any lease's end date rather than just a syndication lease.~~

So this doesn't do what's stated above. It's merely a refactor. To demonstrate this fact, I've commented out some test data that represents an image getting stuck in review: an image with two leases, an expired allowed use lease and an active deny syndication lease.


tonytw1:
Any feelings on the scalaz.NonEmptyList in filters?

akash1810:
> Any feelings on the scalaz.NonEmptyList in filters?

Not a big fan, we should remove it but its everywhere and will make the diff here too noisy.

akash1810:
closing in favour of #2459

PR #2432, opened 2019-02-15
default cropping leases end date to two days from now

akash1810:
request from Jonny

### DIFF
--- a/kahuna/public/js/leases/leases.js
+++ b/kahuna/public/js/leases/leases.js
@@ -225,15 +225,14 @@ leases.controller('LeasesCtrl', [
             ctrl.updateLeases();
         });
 
-        $scope.$watch(() => ctrl.access, onValChange(selectedLeaseType => {
-            const isCroppingLease = ['allow-use', 'deny-use'].includes(selectedLeaseType);
+        function getDefaultExpiryDate(leaseType) {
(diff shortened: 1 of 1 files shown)


paperboyo:
@akash1810 We have an 👍  from Joanna.

akash1810:
this isn't limited to only cropping leases and it should be 😢 not as simple a change as I'd hoped!

kahuna/public/js/leases/leases.js
```diff
@@ -223,6 +225,17 @@ leases.controller('LeasesCtrl', [
             ctrl.updateLeases();
         });
 
+        $scope.$watch(() => ctrl.access, onValChange(selectedLeaseType => {
+            const isCroppingLease = ['allow-use', 'deny-use'].includes(selectedLeaseType);
+
+            if (isCroppingLease) {
+                // cropping leases should expire in 2 days, by default
```
franziskas: could this comment me a method name?

akash1810: Sure! Something like this?

```js
function getDefaultLeaseExpiry() { return moment().add(2, 'days').startOf('day').toDate();}
```

franziskas:
Looks great :)

PR #2436, opened 2019-02-15
Correctly set last modified date when leases are ... modified

jonathonherbert:
Previously, the leases last modified date was the most recent creation date of the image's leases. This PR corrects that behaviour to set lastModified when leases are added or removed.


tonytw1:
Needs to to applied to Elastic6. I'll try to pick it up.

tonytw1:
ps. I was wondering about this when I ported it; thanks for clarifying the desired behaviour.

jonathonherbert:
Thanks for picking this up and adding the test, @tonytw1!

PR #2437, opened 2019-02-15
Add ability to batch add syndication leases when syndication leases are already present on images

jonathonherbert:
This allows users to batch add syndication to leases to collections of images when syndication leases are already present.

If a syndication is lease is present, the user is prompted with a warning, as continuing will remove any previous syndication leases with the incoming syndication lease.

This wasn't possible before -- the user would be confronted with a stern dialog.

batch-lease

### DIFF
--- a/kahuna/public/js/leases/leases.js
+++ b/kahuna/public/js/leases/leases.js
@@ -82,8 +82,10 @@ leases.controller('LeasesCtrl', [
 
                 if (incomingLeaseIsSyndication && syndLeases.length > 0) {
                   const confirmText = ctrl.images.size > 1
-                    ? "One or more of the selected images have syndication leases. These will be overwritten. Do you wish to proceed?"
-                    : "This image already has syndication rights. They will be overwritten. Do you wish to proceed?";
+                    ? "One or more of the selected images have syndication leases. " +
+                      "These will be overwritten. Do you wish to proceed?"
+                    : "This image already has syndication rights. They will be overwritten. " +
+                      "Do you wish to proceed?";
                   const shouldApplyLeases = $window.confirm(confirmText);
                   if (!shouldApplyLeases) {
                     return;
--- a/kahuna/public/js/leases/leases.js
+++ b/kahuna/public/js/leases/leases.js
@@ -79,10 +79,13 @@ leases.controller('LeasesCtrl', [
                 );
 
                 if (incomingLeaseIsSyndication && syndLeases.length > 0) {
-                    const shouldApplyLeases = $window.confirm("One or more of the selected images have syndication leases. These will be overwritten. Do you wish to proceed?");
-                    if (!shouldApplyLeases) {
-                      return;
-                    }
+                  const confirmText = ctrl.images.size > 1
+                    ? "One or more of the selected images have syndication leases. These will be overwritten. Do you wish to proceed?"
+                    : "This image already has syndication rights. They will be overwritten. Do you wish to proceed?";
+                  const shouldApplyLeases = $window.confirm(confirmText);
+                  if (!shouldApplyLeases) {
+                    return;
+                  }
                 }
                 const leasesToAdd = incomingLeaseIsSyndication
                   ? ctrl.leases.leases.filter(_ => !_.access.endsWith('-syndication'))
--- a/kahuna/public/js/leases/leases.js
+++ b/kahuna/public/js/leases/leases.js
@@ -75,7 +75,7 @@ leases.controller('LeasesCtrl', [
                     ctrl.newLease.notes = noteWithClause.join(', ');
                 }
 
-                const incomingLeaseIsSyndication = ctrl.access.endsWith('-syndication');
+                const incomingLeaseIsSyndication = leaseService.isLeaseSyndication(ctrl);
                 const syndLeases = ctrl.leases.leases.filter((l) =>
                     l.access.endsWith('-syndication')
                 );
@@ -84,7 +84,7 @@ leases.controller('LeasesCtrl', [
                   const confirmText = ctrl.images.size > 1
                     ? "One or more of the selected images have syndication leases. " +
                       "These will be overwritten. Do you wish to proceed?"
-                    : "This image already has syndication rights. They will be overwritten. " +
+                    : "This image already has a syndication lease. It will be overwritten. " +
                       "Do you wish to proceed?";
                   const shouldApplyLeases = $window.confirm(confirmText);
                   if (!shouldApplyLeases) {
--- a/kahuna/public/js/leases/leases.js
+++ b/kahuna/public/js/leases/leases.js
@@ -24,6 +24,7 @@ leases.controller('LeasesCtrl', [
     '$timeout',
     'inject$',
     'leaseService',
+    'imageAccessor',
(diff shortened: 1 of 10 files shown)


paperboyo:
> the user would be confronted with a stern dialog

Am I right that the user is not presented with similar dialog when overwriting (existing) cropping leases? And, in general, when performing other, destructive, batch operations.

I would be for consistency of comms/warnings here. A good (better than current) display of existing leases making up a selection would be preferable to disruptive dialog which will quickly become a nuisance, IMHO.

cc/ @harpreetpurewal

kahuna/public/js/leases/leases.js
```diff
@@ -71,21 +75,30 @@ leases.controller('LeasesCtrl', [
...
-                            alertFailed('Something went wrong when saving, please try again.')
-                    )
-                    .finally(() => {
-                        ctrl.resetLeaseForm();
-                    });
+                if (incomingLeaseIsSyndication && syndLeases.length > 0) {
+                  const confirmText = ctrl.images.size > 1
+                    ? "One or more of the selected images have syndication leases. " +
+                      "These will be overwritten. Do you wish to proceed?"
+                    : "This image already has syndication rights. They will be overwritten. " +
```
akash1810: Slight wording change here: `This image already has syndication rights.` -> `This image already has a syndication lease.`

kahuna/public/js/services/api/leases.js
```diff
@@ -81,54 +58,73 @@ leaseService.factory('leaseService', [
...
-      return $q.all(images.map((image) => image.get().then( (apiImage) => {
-        const apiLeases = imageAccessor.readLeases(apiImage);
-        if (apiLeases.leases.length !== originalLeaseCount) {
+    function untilLeasesChange(images) {
+      const imagesArray = images.toArray ? images.toArray() : images;
+      return $q.all(imagesArray.map(image => {
+        return image.get().then(apiImage => {
+          const apiLeases = imageAccessor.readLeases(apiImage);
+          const leases = imageAccessor.readLeases(image);
+          if (apiLeases.lastModified !== leases.lastModified) {
```
akash1810: minor: would requiring `apiLeases.lastModified` to be after `leases.lastModified` be better here?

kahuna/public/js/leases/leases.js
```diff
@@ -71,21 +75,30 @@ leases.controller('LeasesCtrl', [
                     ctrl.newLease.notes = noteWithClause.join(', ');
                 }
 
-                let syndLeases = ctrl.leases.leases.filter((l) =>
+                const incomingLeaseIsSyndication = ctrl.access.endsWith('-syndication');
+                const syndLeases = ctrl.leases.leases.filter((l) =>
```
akash1810: minor: we could reuse [`isLeaseSyndication`](https://github.com/guardian/grid/pull/2437/files#diff-1260212b742426509a02c95e2a8f6332R137) to be DRYer?

leases/app/controllers/MediaLeaseController.scala
```diff
@@ -37,17 +37,38 @@ class MediaLeaseController(auth: Authentication, store: LeaseStore, config: Leas
...
+    mediaLease.prepareForSave.copy(id = Some(UUID.randomUUID().toString), leasedBy = userId)
+
   private def addLease(mediaLease: MediaLease, userId: Option[String]) = {
-    val lease = mediaLease.prepareForSave.copy(id = Some(UUID.randomUUID().toString), leasedBy = userId)
-    store.put(lease).map { _ =>
-      notifications.sendAddLease(lease)
+    val lease = prepareLeaseForSave(mediaLease, userId)
+    lease.isSyndication match {
+      case true =>
+        val mediaId = mediaLease.mediaId
```
akash1810: minor: we could use a for comprehension here to be ever so slightly more readable, but only just.

leases/app/lib/LeaseNotifier.scala
```diff
@@ -2,10 +2,36 @@ package lib
 
 import com.gu.mediaservice.lib.aws.{MessageSender, UpdateMessage}
 import com.gu.mediaservice.lib.formatting._
-import com.gu.mediaservice.model.{LeaseNotice, LeasesByMedia, MediaLease}
+import com.gu.mediaservice.model.{LeasesByMedia, MediaLease}
+
 import org.joda.time.DateTime
 import play.api.libs.json._
 
+case class LeaseNotice(mediaId: String, leaseByMedia: JsValue) {
```
akash1810: I think this is a bad rebase as this case class has been moved to `common-lib`?

thrall/test/lib/ElasticSearchTestBase.scala
```diff
@@ -301,16 +314,23 @@ trait ElasticSearchTestBase extends FreeSpec with Matchers with Fixtures with Be
 
       "removing a lease should update the leases last modified time" in {
         val lease = model.MediaLease(id = Some(UUID.randomUUID().toString), leasedBy = None, notes = Some("A test lease"), mediaId = UUID.randomUUID().toString)
+        val timeBeforeEdit = DateTime.now
         val id = UUID.randomUUID().toString
-        val image = createImageForSyndication(id = UUID.randomUUID().toString, true, Some(DateTime.now()), lease = Some(lease))
+        val image = createImageForSyndication(
+          id = UUID.randomUUID().toString,
```
akash1810: is this meant to be the `id` val above?

kahuna/public/js/leases/leases.js
jonathonherbert: 👍

kahuna/public/js/leases/leases.js
jonathonherbert: 👍

kahuna/public/js/services/api/leases.js
jonathonherbert: yes -- in fact, this had been written in one of my previous implementations. I'll make the update 👍

leases/app/lib/LeaseNotifier.scala
jonathonherbert: Good catch!

thrall/test/lib/ElasticSearchTestBase.scala
jonathonherbert: I hadn't touched this, but looking at it, yes I think so! I'll clean it up.

jonathonherbert:
@paperboyo yes, I take both of these points. I think we probably do want a warning for a destructive operation, but we could definitely display it in a more idiomatic way. Is this a blocker? We've run out of time for this feature as we're concentrating on Fronts for the last few weeks of the quarter 😢

akash1810:
> Am I right that the user is not presented with similar dialog when overwriting (existing) cropping leases?

You are right. But for the wrong reasons 😄. Its not possible to overwrite cropping leases as you can have multiple.

Batch applying a cropping lease is batch add in the traditional sense.

Batch applying a syndication lease is replace or add.

> I would be for consistency of comms/warnings here.

👍 There's slight precedence for the UX in this PR by way of the delete crops/usages warning that's displayed.

> Is this a blocker?

I don't think so. We can always tighten the UX in further PRs if necessary.

PR #2440, opened 2019-02-18
Refactor. Move url building concerns from the configuration to the co…

tonytw1:
…ntrollers, as not strictly a config concern.


tonytw1:
Not approved. Deleting.

PR #2444, opened 2019-02-20
Image inserts don't have an explict id which is abit of a logging bli…

tonytw1:
…nd spot; pick up image.id if available.

### DIFF
--- a/thrall/app/lib/kinesis/ThrallEventConsumer.scala
+++ b/thrall/app/lib/kinesis/ThrallEventConsumer.scala
@@ -45,7 +45,8 @@ class ThrallEventConsumer(es: ElasticSearchVersion,
 
       val updateMessage = Json.parse(message).as[UpdateMessage] // TODO validation
       val timestamp = r.getApproximateArrivalTimestamp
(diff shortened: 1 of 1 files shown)


thrall/app/lib/kinesis/ThrallEventConsumer.scala
```diff
@@ -45,7 +45,8 @@ class ThrallEventConsumer(es: ElasticSearchVersion,
 
       val updateMessage = Json.parse(message).as[UpdateMessage] // TODO validation
       val timestamp = r.getApproximateArrivalTimestamp
-      Logger.info("Got update message (" + timestamp + "): " + updateMessage.subject + "/" + updateMessage.id)
+      val idForLogging = Seq(updateMessage.id, updateMessage.image.map(_.id)).flatten
+      Logger.info("Got update message (" + timestamp + "): " + updateMessage.subject + "/" + idForLogging.mkString(" "))
```
franziskas: Would it make sense to include this part in line 48? `idForLogging.mkString(" ")`

PR #2459, opened 2019-02-22
add a failing test

akash1810:
add a commented out test to document the scenario of "images getting stuck in the syndication review queue"

this is in place of https://github.com/guardian/grid/pull/2424


franziskas:
@akash1810 is there a way to have the test not commented out but ignored so that it is refactoring-safe?

PR #2461, opened 2019-02-22
Batch edits

Reettaphant:
Allow for prepending/appending to image descriptions when batch editing images which have different descriptions.
screen shot 2019-02-22 at 14 28 14

### DIFF
--- a/kahuna/public/js/components/gr-info-panel/gr-info-panel.html
+++ b/kahuna/public/js/components/gr-info-panel/gr-info-panel.html
@@ -112,13 +112,12 @@
                     <div class="metadata-line__info" ng:class="{'image-info--multiple': ctrl.hasMultipleValues(ctrl.rawMetadata.description)}"
                          editable-textarea="ctrl.metadata.description"
                          ng:hide="descriptionEditForm.$visible"
-                         onbeforesave="ctrl.updateMetadataField('description', $data)"
+                         onbeforesave="ctrl.updateMetadataField('description', $data, ctrl.descriptionOption)"
                          e:msd-elastic
                          e:form="descriptionEditForm"
                          e:ng-class="{'image-info__editor--error': $error,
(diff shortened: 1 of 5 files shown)


paperboyo:
💎 

I would, maybe, spread them radios to fill the whole width to prevent accidental clicks?

kahuna/public/js/components/gr-info-panel/gr-info-panel.html
```diff
@@ -143,11 +136,41 @@
...
+                      >
+                        <div
+                         editable-radiolist="ctrl.descriptionOption"
+                         e-ng-options="option.value as option.text for option in ctrl.descriptionOptions"
+                         e-name="descriptionOption"
+                    >
+                        </div>
+                    </div>
+                    <div ng-if="ctrl.userCanEdit && descriptionEditForm.$visible" class="editable-buttons">
+                        <button class="button-cancel" type="button" ng-click="descriptionEditForm.$cancel()">
```
akash1810: can we add a comment around why we've re-implemented the buttons?

kahuna/public/js/edits/service.js
```diff
@@ -286,8 +287,32 @@ service.factory('editsService',
...
+           (
+             descriptionOption !== descriptionEditOptions.append.value &&
+             descriptionOption !== descriptionEditOptions.prepend.value
+            )
+        ) {
+          return value;
+        }
+
+        if (descriptionOption === descriptionEditOptions.append.value) {
+          return imageAccessor.readMetadata(image).description + ' ' + value;
```
akash1810: should we trim value here?

paperboyo:
Couple of small things:
- Shift+Enter doesn’t allow for entering new line after prepended and before appended text (subsequent soft returns do work)
- Usual committing edits by Cmd+Enter doesn’t work

franziskas:
@paperboyo please raise a new health task for those

PR #2463, opened 2019-02-25
allow cropType to be cleared

akash1810:
Setting the `cropType` query string will set a value in session storage, which is later read to freeze the crop type available. This is useful in embedded Grid when the ratio matters, for example a video thumbnail should be a 16:9 ratio.

Allow session storage to be cleared if `cropType === 'all'`. That is, we're happy to crop to any ratio.

### DIFF
--- a/kahuna/public/js/search/index.js
+++ b/kahuna/public/js/search/index.js
@@ -4,7 +4,7 @@ import Rx from 'rx';
 import 'rx-dom';
 import Immutable from 'immutable';
 import {getCollectionsFromQuery} from '../search-query/query-syntax';
-
+import {cropUtil} from '../util/crop';
 import './query';
 import './results';
 import '../preview/image';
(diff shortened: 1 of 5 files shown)


kahuna/public/js/search/index.js
```diff
@@ -82,7 +82,9 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
             const ctrl = this;
 
             if ($stateParams.cropType) {
-                storage.setJs('cropType', $stateParams.cropType, true);
+                $stateParams.cropType === 'all'
```
franziskas: Is there any way to avoid duplicating this code in three places?

akash1810: 😢 I thought this would go unnoticed. Yeah, we probably should right?

akash1810: Done.

PR #2465, opened 2019-02-26
add delete usages component

akash1810:
We have an endpoint in the usages app to delete all usages for an image. This creates a button to expose the endpoint to users via Kahuna.

The component will only appear if you have `delete_crops` permission.

delete-usages

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,11 +3,8 @@ root = true
 
 [*]
 indent_style = space
-indent_size = 4
+indent_size = 2
 end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
-
-*.{scala,yml,yaml}
-indent_size = 2
--- a/kahuna/public/js/util/string.js
+++ b/kahuna/public/js/util/string.js
@@ -8,3 +8,8 @@ string.value('humanJoin', function humanJoin(list, joiner) {
     const lastTwo = list.slice(-2).join(` ${joiner} `);
     return allButLastTwo.concat(lastTwo).join(', ');
 });
+
+string.value('stripMargin', function(template, ...args) {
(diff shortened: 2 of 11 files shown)


paperboyo:
Oh! I’m not sure about this! Very not sure! @blishen?

akash1810:
@paperboyo this is to empower CP to perform image deletions rather than rely on developers. See thread `[bug] Deleting an image from the grid issue/Kept in Library/Taken down status`.

paperboyo:
So it should be controlled by (the much more restrictive) delete_crops permission and not (much wider) edit_metadata, no? I’m less scared, but still a bit scared.

.editorconfig
```diff
@@ -3,11 +3,8 @@ root = true
 
 [*]
 indent_style = space
-indent_size = 4
+indent_size = 2
```
akash1810: because 4 space indents are H U G E !

https://github.com/guardian/grid/pull/2465/files?w=1 may be friendlier to review

kahuna/public/js/util/string.js
```diff
@@ -8,3 +8,8 @@ string.value('humanJoin', function humanJoin(list, joiner) {
     const lastTwo = list.slice(-2).join(` ${joiner} `);
     return allButLastTwo.concat(lastTwo).join(', ');
 });
+
+string.value('stripMargin', function(template, ...args) {
+  const result = template.reduce((acc, part, i) => acc + args[i - 1] + part);
```
akash1810: stolen from [here](https://github.com/guardian/grid/blob/master/scripts/generate-dot-properties/service-config.js#L1-L13)

jonathonherbert:
Walked this through with @akash1810 and it looks good! 👍

PR #2468, opened 2019-02-26
Normalise new lines when read from database

sihil:
The use of `\r` line endings instead of `\n` causes problems with presentation and behaviour of kahuna. The aim of this PR is to eliminate the issue of having different newline types stored in ES by normalising image metadata as it is retrieved from the database.

I've implemented this in the JSON reader implementation so as to be reasonably clean and normalise the data as soon as possible after leaving the DB layer. This is possibly contentious as it might be better to do this a little later and make the implementation clearer. As far as I'm aware there is no precedent for this but I'd be glad to be pointed to any examples.

This is an alternative approach to #2470 and only one should be merged (the one with the most approvals??)

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/formatting/package.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/formatting/package.scala
@@ -5,6 +5,8 @@ import scala.util.Try
 import org.joda.time.DateTime
 import org.joda.time.format.ISODateTimeFormat
 import org.joda.time.format.DateTimeFormatterBuilder
+import scalaz.Functor
+import scalaz.std.option._
 
 package object formatting {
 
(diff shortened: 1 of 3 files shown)


tonytw1:
Is there ab argument for putting this on the Writes so it happens closer to the edge of the application?
Is the scalaz (and additional vocabulary requirement) strictly necessary?
Are there any meta data read / write round trips inside the application which aren't presentation related which could be impacted by this format change?

paperboyo:
I wouldn’t know, right, but all the other cleanup operations seem to happen here: https://github.com/guardian/grid/tree/master/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup. Why not this one?

sihil:
The cleanup there modifies data before it hits the database. Good for anything that might impact search, but not so great for anything else as it makes it harder to implement a change to a cleaner.

sihil:
Closed in preference for #2470

PR #2470, opened 2019-02-26
Normalise newlines when preparing image responses for API

sihil:
The use of `\r` line endings instead of `\n` causes problems with presentation and behaviour of kahuna. The aim of this PR is to eliminate the issue of having different newline types stored in ES by normalising image metadata as it is prepared to be sent to the client.

I've implemented this in `ImageResponse` class so as to leave the change until it's about to leave the server. This localises the change and should reduce the possibility of introducing any unintended consequences. As far as I'm aware there is no precedent for this but I'd be glad to be pointed to any examples.

This is an alternative approach to #2468 and only one should be merged (the one with the most approvals??)

### DIFF
--- a/media-api/app/lib/ImageResponse.scala
+++ b/media-api/app/lib/ImageResponse.scala
@@ -8,6 +8,7 @@ import com.gu.mediaservice.lib.auth.{Internal, Tier}
 import com.gu.mediaservice.lib.collections.CollectionsManager
 import com.gu.mediaservice.model._
 import com.gu.mediaservice.model.usage._
+import com.softwaremill.quicklens._
 import lib.usagerights.CostCalculator
 import org.joda.time.DateTime
 import play.api.Logger
@@ -312,7 +313,7 @@ class ImageResponse(config: MediaApiConfig, s3Client: S3Client, usageQuota: Usag
     (__ \ "fileMetadata").write[FileMetadataEntity]
       .contramap(fileMetadataEntity(id, expandFileMetaData, _: FileMetadata)) ~
     (__ \ "userMetadata").writeNullable[Edits] ~
-    (__ \ "metadata").write[ImageMetadata] ~
+    (__ \ "metadata").write[ImageMetadata](ImageResponse.newlineNormalisingImageMetadataWriter) ~
     (__ \ "originalMetadata").write[ImageMetadata] ~
     (__ \ "usageRights").write[UsageRights] ~
     (__ \ "originalUsageRights").write[UsageRights] ~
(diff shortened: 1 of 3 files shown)


media-api/app/lib/ImageResponse.scala
```diff
@@ -363,6 +364,22 @@ class ImageResponse(config: MediaApiConfig, s3Client: S3Client, usageQuota: Usag
...
+
+  def normaliseNewLines(imageMetadata: ImageMetadata): ImageMetadata = imageMetadata.modifyAll(
+    _.description,
+    _.copyrightNotice,
+    _.copyright,
+    _.specialInstructions,
+    _.suppliersReference
+  ).using(_.map(ImageResponse.normaliseNewLines))
+
+  def normaliseNewLines(string: String): String = string.replaceAll("(\\r|\\n|\\r\\n)+", "\n")
```
mchv: using `String.replaceAll()` will compile the regular expression **each time** you call it, as implementation is 
```java
public String replaceAll(String regex, String replacement) {
        return Pattern.compile(regex).matcher(this ).replaceAll(
             replacement);
}
```
One can use a pattern instead
```scala
private val pattern = """(\r|\n|\r\n)+""".r
def normaliseNewLines(string: String): String = pattern.replaceAllIn(string, "\n")
```

I think using `"""` ease readability as well

sihil: I considered this as I wrote it but concluded it was premature optimisation and that the compiler or runtime would optimise this away given that it knows the parameter is a constant. I'll run some performance comparison tests tomorrow to see what the impact is given your concern.

tonytw1:
+1 ImageResponse looks like a good home for this presentation concern.

media-api/app/lib/ImageResponse.scala
mchv: I am quite confident neither the compiler nor runtime will optimise this, as non using a pattern is a known performance bottleneck, for instance see [3.5 precompiled pattern performance benchmark](https://www.baeldung.com/java-string-performance). You will find as well the same advice (including using triple quotes for readability) [on the Scala documentation](https://www.scala-lang.org/api/2.12.8/scala/util/matching/Regex.html)
```
An instance of Regex represents a compiled regular expression pattern.
Since compilation is expensive, frequently used Regexes should be constructed once,
outside of loops and perhaps in a companion object.
```

Change as well does not impact readability, so could be better qualified as a best practice rather than a premature optimisation.

sihil: I ran some numbers. It shaves around 15% off the clock time but will also be reducing object allocation (although that's likely to be insignificant given the context).

sihil: I've continued my testing and have also written a version in fastparse. It's about 4 times faster. But also hard to read so I'm not going to use it.

PR #2473, opened 2019-02-27
Improve logic to convert true colour PNGs without transparency to jpegs

sihil:
Two components to this PR. The first is removing a piece of code that attempted to do some optimisation to certain types of PNGs. The second is to re-introduce the logic from #2375 that decides whether the crops should be JPEG or PNGs depending on a predicted quality/size trade-off. 

In our case:
 - if the PNG has transparency then it should always be a PNG as the transparency is not available in JPEG
 - If the PNG is not true colour then we assume it is a graphic that should be retained as a PNG

I've added a few tests to document the intent of the function in question as it wasn't super clear to me when working on it.

### TODO:
- [x] Test on TEST


paperboyo:
> code looks good overall, is it worth adding a test with some image files too?

Just in case it’s useful:

`it("should return JPEG when the input type isn't PNG")`
a02963c081e0b8af4abefb59215ca80d0dfe1f9f

`it("should return PNG when the input type is PNG and it has alpha")`
0c8b8316cc217de125ced757cc0579014664bb60

`it("should return PNG when the input type is PNG and it has alpha even if it is True Color")`
8fb7125b2d8598cda4277a7690897b24870f0e88

`it("should return PNG when the input type is PNG and it is NOT true color (a graphic)`
29a4848de33fe9468723cb12bf90cfada7def74f

`it("should return JPEG when the input type is PNG and it is true color")`
this *should add* that it has to have `hasAlpha:false` too
e5f7259daf26d496c538a9b817bc18540941e72f

sihil:
Depends what we are trying to test @akash1810 - I'd suggest that we only test the logic table.

sihil:
Having chatted with @paperboyo - it's suggested that we write some image based tests as a separate concern.

sihil:
If we wanted to do tests we'd need relatively small image files so we don't bloat the repo.

paperboyo:
> If we wanted to do tests we'd need relatively small image files so we don't bloat the repo.

Good point :facepalm:. I think I have seen PNG test files somewhere of every flavour allowed by the spec. Can’t find now, but will have a look…

akash1810:
> If we wanted to do tests we'd need relatively small image files so we don't bloat the repo.

Could we use exiftool to force the metadata onto some image files?

paperboyo:
> Could we use exiftool to force the metadata onto some image files?

The characteristic we are looking for here is derived from the image makup, not merely metadata. _I think._

[Here are some](http://www.libpng.org/pub/png/pngsuite.html) comprehensive (and tiny) image files. Lots of them.

sihil:
Have you loaded any of them into the grid to see if there is a useful set?

paperboyo:
> Have you loaded any of them into the grid to see if there is a useful set?

TEST: ?query=%23pngsuite&nonFree=true

I think the “Basic format test files” are enough. I don’t think we need also testing eg. ICC profile extraction or `sRGB` chunk support, cause we would be effectively testing GraphicsMagick here.

Not sure if this suite contains any images that have more than one transparency entry in their palettes. We do support these, though.

For now:

| File                 | Desc               | `colorType`  | `bitsPerSample`  | GM’s `format %r` | `hasAlpha` |
| --------          |:-------------:| -----:| -----:| -----:| -----:|
| basn0g01     | black & white | `Greyscale` | 1 | `Bilevel` | `false` |
| basi0g02     | 2 bit (4 level) grayscale      |   2 | `Grayscale` | `false` |

_… I would fill that, but am struggling with passing multiple files to gm. plus, Grid won’t let me search for arbitrary values in `fileMetadata`, so too much clicking for last day before holidays…_

This PR should be merged. We may think about tests later ;-)

sihil:
@paperboyo @akash1810 How about these as a starting point?

sihil:
I've improved the test file names. @paperboyo and I have realised that the `hasAlpha` doesn't populate unless `gm` is installed. Mid term an error should be escalated as in PROD missing `gm` should be fatal. For the sake of this PR though, I don't want to change the tests to rely on the presence of `gm`.

paperboyo:
Been a while… should we merge it? Not sure about the tests, but doing ’em properly is probably never happen. I mean, not soon.

PR #2480, opened 2019-02-28
Remove the SameSite restriction on the Play session cookie

mbarton:
Even post #2404, we are still seeing occasional big angry red login error boxes when running the Grid.

From analysing the network requests, we can see that the errors is "missing anti-forgery token", caused because the `PLAY_SESSION` cookie is not set when we are redirected back from the Google OAuth endpoint.

We think this is because the cookie has `SameSite=Lax`, which if you have a [pessismistic read of the specification](https://tools.ietf.org/html/draft-west-first-party-cookies-07#section-4.1.1) should prevent the cookie from being included. [This Firefox blog post](https://blog.mozilla.org/security/2018/04/24/same-site-cookies-in-firefox-60/) also mentions the behaviour (at the bottom).

Chrome however does include it, Firefox does not. Since we require it to be included this PR disables the `SameSite` part of setting the cookie for `PLAY_SESSION` in the auth service.

This should have a limited security impact right now, since this cookie is only used during the auth dance with the seperate Panda cookie used to actually login. However it may impact us in the future if we decide to put more stuff in the session.

We were unable to reproduce locally but @RichieAHB made a [cool cut-down node app](https://glitch.com/edit/#!/receptive-brownie) that appears to prove our hypothesis. The intention is to throw this branch onto test after a night of inactivity to see if it makes a difference.

### DIFF
--- a/auth/app/auth/AuthComponents.scala
+++ b/auth/app/auth/AuthComponents.scala
@@ -3,13 +3,25 @@ package auth
 import com.gu.mediaservice.lib.management.ManagementWithPermissions
 import com.gu.mediaservice.lib.play.GridComponents
 import play.api.ApplicationLoader.Context
+import play.api.{Configuration, Environment}
+import play.api.http.HttpConfiguration
 import router.Routes
 
 class AuthComponents(context: Context) extends GridComponents(context) {
   final override lazy val config = new AuthConfig(configuration)
+  final override lazy val httpConfiguration = AuthHttpConfig(configuration, context.environment)
 
   val controller = new AuthController(auth, config, controllerComponents)
   val permissionsAwareManagement = new ManagementWithPermissions(controllerComponents, controller)
 
   override val router = new Routes(httpErrorHandler, controller, permissionsAwareManagement)
+}
+
+object AuthHttpConfig {
+  def apply(playConfig: Configuration, environment: Environment): HttpConfiguration = {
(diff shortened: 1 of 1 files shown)


mbarton:
[Grid - User getting frequently logged out esp AUS.](https://trello.com/c/1ubI2Doa/512-grid-user-getting-frequently-logged-out-esp-aus)

akash1810:
I might be misunderstanding this, however given the description, I'd have expected to see this behaviour all the time, rather than periodically?

mbarton:
> I'd have expected to see this behaviour all the time

We think this is triggered when the frontend gets a 419 from the API and attempts to re-auth using the iframe (using [panda-session](https://github.com/guardian/panda-session))

RichieAHB:
Specifically the iframe seems to be the issue when setting SameSite to "Lax".

mbarton:
Finally managed to test this and it does seem to work, at least from the network inspector view all the network requests are clean `200` or `303`.

That said, the Grid UI still seems to hang for 5 seconds and then redirect for an interactive login anyway! ARGTGGGGHGGHGHGHG

mbarton:
Merging since it seems to make things better. Further investigation will follow as and when we have time

mbarton:
Merging since it seems to make things better. Further investigation will follow as and when we have time

PR #2481, opened 2019-02-28
use generic radio button component for description edit form

akash1810:
Best reviewed [w/out whitespace](https://github.com/guardian/grid/pull/2481/files?w=1) as I've changed the indentation of a couple of files 😄

Updates the description edit form to use the generic `gr-radio-list` component. This results in more consistent styling.

Also makes the edit mode more generic, defaulting the mode to `overwrite`. This just removes field specific knowledge out of `editsService` factory.

# Before

image

# After

image


paperboyo:
Violet? 😆 
But, seriously, I think we need delineation between unchosen options! A lil’ faint line would do!

akash1810:
> But, seriously, I think we need delineation between unchosen options! A lil’ faint line would do!

Styling updated here - https://github.com/guardian/grid/pull/2482

PR #2482, opened 2019-03-01
update gr-radio-list style

akash1810:
- add a border around each option
- ensure the label is always Title case with a `text-transform`

# Before
image

# After
image


akash1810:
this is a relatively light change, merging w/out a review

img

PR #2485, opened 2019-03-01
Use WS for the Cropper call to media API. The SNI issue which Apache …

tonytw1:
…http components was been used to work around is likely fixed in Play 2.6.


tonytw1:
Not deleting as useful and approved. Tests ok in TEST.

PR #2489, opened 2019-03-06
Updated display crops to use the file url from media api instead of secureUri

jamie-pitts:
`exports` (ie public crops) in the media api response has both a `file` and `secureUrl`. The secure field is redundant for a crop that is already public so the UI should not depend on it


mbarton:
Thanks for the contribution Jamie! I've added a quick description of the issue

PR #2490, opened 2019-03-06
Fix replace leases action in ES6

jonathonherbert:
The replace leases action in ES6 was broken. The type that eventually made it to the Painless script didn't align with what that script expected, and requests were failing as a result. This PR should fix that.

In addition to this, the lastModified value was being set incorrectly -- this is set as now() just before we run the ES script.

Finally, it also provides better type safety in ElasticSearchVersion, for comfort 🛏

### DIFF
--- a/leases/app/lib/LeaseNotifier.scala
+++ b/leases/app/lib/LeaseNotifier.scala
@@ -2,34 +2,9 @@ package lib
 
 import com.gu.mediaservice.lib.aws.{MessageSender, UpdateMessage}
 import com.gu.mediaservice.lib.formatting._
-import com.gu.mediaservice.model.{LeasesByMedia, MediaLease}
+import com.gu.mediaservice.model.{LeaseNotice, LeasesByMedia, MediaLease}
 import org.joda.time.DateTime
 import play.api.libs.json._
-
-case class LeaseNotice(mediaId: String, leaseByMedia: JsValue) {
-  def toJson = Json.obj(
-    "id" -> mediaId,
-    "data" -> leaseByMedia,
-    "lastModified" -> printDateTime(DateTime.now())
-  )
-}
-
-object LeaseNotice {
-  import JodaWrites._
-
-  implicit val writer = new Writes[LeasesByMedia] {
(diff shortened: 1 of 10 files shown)


tonytw1:
Agreed. The missing media id on that message type would cause that. Pardon.

leases/app/lib/LeaseNotifier.scala
```diff
@@ -39,7 +14,7 @@ class LeaseNotifier(config: LeasesConfig, store: LeaseStore) extends MessageSend
   def sendReindexLeases(mediaId: String) = {
     val replaceImageLeases = "replace-image-leases"
     val leases = store.getForMedia(mediaId)
-    val updateMessage = UpdateMessage(subject = replaceImageLeases, leases = Some(leases) )
+    val updateMessage = UpdateMessage(subject = replaceImageLeases, leases = Some(leases), id = Some(mediaId) )
```
akash1810: https://github.com/guardian/grid/pull/2491 is very much inspired by this.

thrall/test/lib/ElasticSearchTestBase.scala
```diff
@@ -316,18 +316,27 @@ trait ElasticSearchTestBase extends FreeSpec with Matchers with Fixtures with Be
...
 
-        Await.result(Future.sequence(ES.replaceImageLeases(id, JsDefined(Json.toJson(updatedLeases)), asJsLookup(DateTime.now))), fiveSeconds)
+        Await.result(Future.sequence(ES.replaceImageLeases(id, updatedLeases)), fiveSeconds)
 
-        reloadedImage(id).get.leases.leases.size shouldBe 2
-        reloadedImage(id).get.leases.leases.head.notes shouldBe Some("An updated lease")
+        val newLeases = reloadedImage(id).get.leases
+        newLeases.leases.size shouldBe 2
+        newLeases.leases.head.notes shouldBe Some("An updated lease")
+        newLeases.lastModified.get.isAfter(timeBeforeEdit) shouldBe true
```
akash1810: 👍

PR #2491, opened 2019-03-06
refactor UpdateMessage

akash1810:
Make `id` required as it should always be present. Also remove `withImageId` as a result too.

Easier to review here - https://github.com/guardian/grid/pull/2491/files?w=1

### DIFF
--- a/thrall/app/lib/kinesis/ThrallEventConsumer.scala
+++ b/thrall/app/lib/kinesis/ThrallEventConsumer.scala
@@ -13,6 +13,7 @@ import com.gu.mediaservice.model.usage.UsageNotice
 import lib._
 import play.api.Logger
 import play.api.libs.json.{JodaReads, Json}
+import scalaz.syntax.id._
 
 import scala.concurrent.duration.{Duration, SECONDS}
 import scala.concurrent.{Await, ExecutionContext, Future}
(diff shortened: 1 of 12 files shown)


thrall/app/lib/kinesis/ThrallEventConsumer.scala
```diff
@@ -48,26 +49,30 @@ class ThrallEventConsumer(es: ElasticSearchVersion,
...
-              Logger.info("Completed processing of update message: " + messageLogMessage)
-            }.recover {
-              case e: Throwable =>
-                Logger.error("Failed to process update message; message will be ignored: " + ("Got update message: " + messageLogMessage), e)
-            }
+          val maybeUpdateMessage = Json.parse(message).validate[UpdateMessage] <| logParseErrors |> (_.asOpt)
 
-            Await.ready(eventuallyAppliedUpdate, ThirtySeconds)
-          }
+          maybeUpdateMessage match {
```
akash1810: fail fast - validate incoming message against `UpdateMessage`

tonytw1:
Seem to recall the 'heartbeat' message been the sticking point. Is it compatible with the mandatory id or can it been deleted?

akash1810:
@tonytw1 great point and you're absolutely right! I think we can change the shape of the heartbeat message to include an id to address this.

akash1810:
On second thoughts, we can be better still and deserialise into the original case classes rather than a generic `UpdateMessage`. Closing this PR and will raise another to that effect.

PR #2492, opened 2019-03-08
update setup

akash1810:
Fixes a couple of scripts.

- removes `setup.sh` and moving its behaviour into `dev-start.sh`
- update `dev-start.sh` to build a initial version of the client side app
- update `fetch-config.sh` to create and own `/etc/gu` if it doesn't exist or isn't writable by the current user
- adds a `Brewfile` for brew bundle goodness


fetch-config.sh
```diff
@@ -2,10 +2,33 @@
...
     --region eu-west-1
+}
+
+if [[ -w ${DOWNLOAD_DIR} ]]; then
+  downloadConfig
+  echo -e "${GREEN}Done ${NC}"
+else
+  echo "Cannot write to ${DOWNLOAD_DIR}. It either doesn't exist or is not owned by $(whoami)."
+
+  echo "Creating ${DOWNLOAD_DIR} and making user $(whoami) the owner"
```
mbarton: Maybe worth adding a message that this will prompt you for your sudo password

akash1810: Done.

PR #2503, opened 2019-04-02
add more markers to reaper logs

akash1810:
to make searching the logs in the ELK stack slightler easier

### DIFF
--- a/reaper/index.js
+++ b/reaper/index.js
@@ -27,11 +27,15 @@ exports.handler = (event, context, callback) => {
                         fetch(image.uri, { headers, method: 'DELETE' })
                         .then(res => {
                             if(res.status >= 200 && res.status < 300) {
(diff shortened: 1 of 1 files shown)


reaper/index.js
```diff
@@ -27,11 +27,15 @@ exports.handler = (event, context, callback) => {
                         fetch(image.uri, { headers, method: 'DELETE' })
                         .then(res => {
                             if(res.status >= 200 && res.status < 300) {
-                                logger.log( `${res.status}: successfully deleted picture ${imageId}`);
+                                logger.log( `${res.status}: successfully deleted picture ${imageId}`, {
```
akash1810: this is how we add arbitrary markers: https://github.com/guardian/elk-kinesis-logger/blob/master/examples/advanced-example.js#L17-L19

PR #2504, opened 2019-04-02
Each deployable hs it's own duplicated application.conf rather than t…

tonytw1:
…hem been supplied via the common-lib jar.

Moves the conf of these Play applications closer to expected normal making it more obvious to external developer where this conf if coming from.
Makes it obvious where the dependancy on /etc/gu is coming from.
Allows us to start stripping unused config from deployables which don't use all of these shared properties.
One master config file per deployable is a step towards been able to confif these applications to run in containers.


mbarton:
I've copied this work into #2510

PR #2505, opened 2019-04-02
Config properties private

tonytw1:


mbarton:
I've copied this work into #2510

PR #2508, opened 2019-04-04
use nested query for Usages

akash1810:
Usages is a [nested field](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch6/Mappings.scala#L245) in ES6, therefore we need to use a nested query against that field - https://www.elastic.co/guide/en/elasticsearch/guide/current/nested-query.html.

This PR fixes the persisted filter (including tests). `has:usages` is not fixed, however; we can do this as a separate PR.

This also fixes the recent spate of Media API 5XX alarms. This was caused by:
- reaper asking media-api for 100 images to delete
- media-api, incorrectly, returning images that have usages
- reaper issuing a DELETE request for each image
- media-api returns non 200 as the image cannot be deleted
- alarm rings

This PR addresses the issue as media-api will not return images that have usages in reaper's first request.

### DIFF
--- a/media-api/test/lib/elasticsearch/ElasticSearchTestBase.scala
+++ b/media-api/test/lib/elasticsearch/ElasticSearchTestBase.scala
@@ -2,7 +2,7 @@ package lib.elasticsearch
 
 import java.util.UUID
 
-import com.gu.mediaservice.model.{Edits, FileMetadata, ImageMetadata, Handout, StaffPhotographer}
+import com.gu.mediaservice.model.{Edits, FileMetadata, Handout, ImageMetadata, StaffPhotographer}
 import org.joda.time.DateTime
 import org.scalatest.concurrent.PatienceConfiguration.{Interval, Timeout}
 import org.scalatest.concurrent.ScalaFutures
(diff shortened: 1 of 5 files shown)


media-api/test/lib/elasticsearch/ElasticSearchTestBase.scala
```diff
@@ -20,8 +20,21 @@ class ElasticSearchTestBase extends FunSpec with BeforeAndAfterAll with Matchers
     createImage(UUID.randomUUID().toString, Handout(), usages = List(createDigitalUsage())),
 
     // with user metadata
-    createImage(id = "test-image-13-edited", Handout()).copy(userMetadata = Some(Edits(metadata = ImageMetadata(credit = Some("author")))), uploadTime = DateTime.now.minusMonths(1)),
-    createImage(id = "test-image-14-unedited", Handout()).copy(uploadTime = DateTime.now.minusMonths(1)),
+    createImage(id = "persisted-because-edited", Handout()).copy(
+      userMetadata = Some(Edits(
+        metadata = ImageMetadata(credit = Some("author")))
+      ),
+      uploadTime = DateTime.now.minusMonths(1)
```
franziskas: would it be worth extracting the upload time to a constant that clarifies that it's an irrelevant date in the past / a specific date in the past?

akash1810: 👍 went a bit further and created a helper function [`createImageUploadedInThePast`](https://github.com/guardian/grid/pull/2508/files#diff-ad7c3b07522d48cba07d084c72fd1ec3R58).

akash1810:
The number of `405`s experienced by reaper is has dropped now! 🎉 

image

PR #2510, opened 2019-04-08
Refactor configuration

mbarton:
Moves all configuration out to an `application.conf` file in each service. This should help new Grid users as it will be straight forward to see what configuration is required for each application.

Thanks to @tonytw1 who inspired much of this work with his spike on running the Grid in containers.

Objects and clients are instantiated as close as possible to each `XXXComponents` file and passed downwards rather than instantiated from config at call site.

This also removes Guardian-specific concepts like `Stack`, `Stage` etc in favour of everything being configuration.


kenoir:
Epic refactor! This would be genuinely useful for 3rd parties wanting to run The Grid - not to mention the other advantages of containerisation.

 I'd love to be able to throw a Grid instance up at https://github.com/wellcometrust/ 

/cc @jamesgorrie

mbarton:
Closing as we still need to identify the minimum changes required to make a push-button Grid for external users

PR #2512, opened 2019-04-08
Remove cropper api key

mbarton:
Cropper uses an API key to access Media API - obtaining metadata about an image before cropping. It must be created before a new deployment of the Grid is used.

This PR removes the API key, passing through the credentials from the cropping request (API key or Panda cookie), removing the need for the API key.

I've added a new field to `RequestLoggingFilter`: `originalSystem`. This will allow us to distinguish between requests from a real client and inter-service requests done on behalf of a client.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Authentication.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Authentication.scala
@@ -46,7 +46,7 @@ class Authentication(config: CommonConfig, actorSystem: ActorSystem,
 
   override def invokeBlock[A](request: Request[A], block: Authentication.Request[A] => Future[Result]): Future[Result] = {
     // Try to auth by API key, and failing that, with Panda
-    request.headers.get(headerKey) match {
+    request.headers.get(Authentication.apiKeyHeaderName) match {
       case Some(key) =>
         keyStore.lookupIdentity(key) match {
           case Some(apiKey) =>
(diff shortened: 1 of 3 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Authentication.scala
```diff
@@ -69,6 +68,19 @@ class Authentication(config: CommonConfig, actorSystem: ActorSystem,
     Authentication.validateUser(authedUser, userValidationEmailDomain, multifactorChecker)
   }
 
+  def getOnBehalfOfPrincipal(principal: Principal, originalRequest: Request[_]): OnBehalfOfPrincipal = principal match {
+    case service: AuthenticatedService =>
+      OnBehalfOfService(service)
+
+    case user: PandaUser =>
+      val cookieName = panDomainSettings.settings.cookieSettings.cookieName
```
akash1810: all the settings!

common-lib/src/main/scala/com/gu/mediaservice/lib/play/RequestLoggingFilter.scala
```diff
@@ -33,11 +34,13 @@ class RequestLoggingFilter(override val mat: Materializer)(implicit ec: Executio
   private def logSuccess(request: RequestHeader, response: Result, duration: Long): Unit = {
     val originIp = request.headers.get("X-Forwarded-For").getOrElse(request.remoteAddress)
     val referer = request.headers.get("Referer").getOrElse("")
+    val originalService = request.headers.get(Authentication.originalServiceHeaderName).getOrElse("")
     val length = response.header.headers.getOrElse("Content-Length", 0)
 
     val markers = MarkerContext(appendEntries(Map(
       "origin" -> originIp,
       "referrer" -> referer,
+      "originalService" -> originalService,
```
akash1810: Could we only add this marker if there is an `originalService`? That way we can still use `exists` in ELK when searching logs.

mbarton: Yep makes sense, I've updated the code

PR #2513, opened 2019-04-09
Remove hard-coded name for the Thrall Kinesis stream

mbarton:
At the moment you can't have more than one stream per account which makes it hard to test setting up a new stack from scratch.

This change will remove/rename the existing stream and is thus dependent on #2510 before merging, as that PR allows us to inject the stream name in config.


mbarton:
Covered by https://github.com/guardian/grid/pull/2537

PR #2524, opened 2019-04-16
don't write to thrall sns

akash1810:
Part of the effort to turn off the elasticsearch 1 cluster.

Stop writing to the SNS topic for Thrall to consume into the elasticsearch 1 cluster. We continue to write to Kinesis, which Thrall consumes into the elasticsearch 6 cluster.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/ThrallMessageSender.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/ThrallMessageSender.scala
@@ -7,8 +7,8 @@ import org.joda.time.DateTime
 import play.api.libs.json.JsValue
 
 // TODO MRB: replace this with the simple Kinesis class once we migrate off SNS
-class MessageSender(config: CommonConfig, snsTopicArn: String) {
-  private val legacySns = new SNS(config, snsTopicArn)
+class ThrallMessageSender(config: CommonConfig, thrallSnsTopic: String) {
+  private val legacySns = new SNS(config, thrallSnsTopic)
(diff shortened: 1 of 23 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/aws/ThrallMessageSender.scala
```diff
@@ -4,16 +4,13 @@ import com.gu.mediaservice.lib.config.CommonConfig
 import com.gu.mediaservice.model._
 import com.gu.mediaservice.model.usage.UsageNotice
 import org.joda.time.DateTime
-import play.api.libs.json.JsValue
 
 // TODO MRB: replace this with the simple Kinesis class once we migrate off SNS
-class MessageSender(config: CommonConfig, snsTopicArn: String) {
-  private val legacySns = new SNS(config, snsTopicArn)
+class ThrallMessageSender(config: CommonConfig) {
```
akash1810: renamed to be explicit

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/ThrallMessageSender.scala
```diff
@@ -4,16 +4,13 @@ import com.gu.mediaservice.lib.config.CommonConfig
...
   private val kinesis = new Kinesis(config, config.thrallKinesisStream)
 
   // TODO deprecate the message JsValue input in favour of the more structured update message
```
marialivia16: Is this comment still needed now?

PR #2528, opened 2019-04-17
Shonkily restrict batch editing photoshoots

mbarton:
They currently have O(N!) behaviour and can bring down the Grid!

21432b593b56b8a52cacf8b1c7f735a2

### DIFF
--- a/kahuna/public/js/components/gr-photoshoot/gr-photoshoot.html
+++ b/kahuna/public/js/components/gr-photoshoot/gr-photoshoot.html
@@ -6,7 +6,7 @@
     </button>
 
     <div ng:hide="!ctrl.editingDisabled">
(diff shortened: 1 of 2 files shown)


paperboyo:
How about adding `Editing photoshoots is *currently* not possible` to create incentive for developers to fix it? ;-)

I wonder: if it’s indeed not adding pics to photoshoot but the subsequent propagation of RCS rights to the said shoot, maybe we don’t have to do anything to upload page? (and maybe this here could be showing only if selection contains any RCS info; not worth redoing, though!)

mbarton:
Yep I think we're fine to leave the upload page since people will have to edit the photoshoot of each image manually

mbarton:
[Restrict batch editing of photoshoots to 10 images](https://trello.com/c/pTkNG9Fv/21-restrict-batch-editing-of-photoshoots-to-10-images)

mbarton:
I went with `We are working to improve this`

PR #2530, opened 2019-04-17
es6 config is no longer optional

akash1810:
## What does this change?
This PR makes the ES* config easier to reason about by removing Option types.

We've been running both clusters for a while and there is no scenario where we would only be running es1.

Update the config generator to add es6 config to the .properties files too.

Ultimately, this is to make the removal of the ES1 cluster easier.

Fewer headaches!

n/a

## Who should look at this?
**YOU**

## Tested?
- [x] locally
- [x] on TEST

### DIFF
--- a/media-api/app/lib/MediaApiConfig.scala
+++ b/media-api/app/lib/MediaApiConfig.scala
@@ -42,18 +42,13 @@ class MediaApiConfig(override val configuration: Configuration) extends CommonCo
         "App"   -> Seq(elasticsearchApp)
       ))
 
-  lazy val elasticsearchPort: Option[Int] = properties.get("es.port").map(_.toInt)
-  lazy val elasticsearchCluster: Option[String] = properties.get("es.cluster")
+  lazy val elasticsearchPort: Int = properties("es.port").toInt
+  lazy val elasticsearchCluster: String = properties("es.cluster")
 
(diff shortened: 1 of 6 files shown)


media-api/app/lib/MediaApiConfig.scala
```diff
@@ -42,19 +42,14 @@ class MediaApiConfig(override val configuration: Configuration) extends CommonCo
...
-
-  lazy val elasticsearch6Host: Option[String] =  {
-    if (isDev)
-      Some(properties.getOrElse("es6.host", "localhost"))
-    else
-      properties.get("es6.host")
-  }
-  lazy val elasticsearch6Port: Option[Int] = properties.get("es6.port").map(_.toInt)
-  lazy val elasticsearch6Cluster: Option[String] = properties.get("es6.cluster")
-  lazy val elasticsearch6Shards = Some(if (isDev) 1 else properties("es6.shards").toInt)
```
akash1810: none of this `isDev` stuff now, just obey config 😄

scripts/generate-dot-properties/config.json5
```diff
@@ -17,6 +17,14 @@
     key: ''
   },
 
+  es6: {
```
akash1810: these are the defaults for DEV

PR #2533, opened 2019-04-18
Improve updating rights logic on photoshoot actions

marialivia16:
What is happening at the moment is that for any update on a photoshoot, we are looking for syndication rights to infer and we send an update to ES for all the other images in the shoot no matter if any rights have been found. Which means that for a photoshoot of `n` images, we send `n*n` requests. 💥 

Following this change thrall should stop trying to make no-op actions. Looking at Kibana, my assumption is that these kind of requests caused the outage from a couple of days ago. When someone decided to organise a few sets of images into photoshoots, ES went crazy. 

Find the issue graph:
Screenshot 2019-04-18 at 13 52 42

It also looks like this branch compiles 2 times faster than the master branch 🏃

### DIFF
--- a/thrall/test/resources/logback-test.xml
+++ b/thrall/test/resources/logback-test.xml
@@ -8,11 +8,8 @@
         </encoder>
     </appender>
 
-    <logger name="play" level="INFO" />
-    <logger name="application" level="DEBUG" />
-
-    <logger name="com.amazonaws">
-        <level value="ERROR" />
-    </logger>
+  <root level="INFO">
+    <appender-ref ref="CONSOLE"/>
+  </root>
 
 </configuration>
--- a/thrall/test/resources/logback-test.xml
+++ b/thrall/test/resources/logback-test.xml
@@ -8,8 +8,11 @@
         </encoder>
     </appender>
 
-  <root level="INFO">
-    <appender-ref ref="CONSOLE"/>
-  </root>
+<!--  <logger name="application" level="INFO" />-->
+<!--  <logger name="play" level="INFO" />-->
+
+<!--  <root level="INFO">-->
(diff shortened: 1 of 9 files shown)


marialivia16:
[Avoid "null" updates to rights when editing a photoshoot](https://trello.com/c/vzxulw0Z/26-red)

marialivia16:
I'm running this branch on TEST and the logs look sane to me. I'll try to do the same ops without it and see the difference.

marialivia16:
I did some testing on a sequential set of actions as seen below and the number of `updateImageSyndicationRights` calls to ES has been reduced dramatically as part of this work:
Screenshot 2019-04-18 at 15 25 30

thrall/test/resources/logback-test.xml
```diff
@@ -8,11 +8,11 @@
...
 
-    <logger name="play" level="INFO" />
-    <logger name="application" level="DEBUG" />
+<!--  <logger name="application" level="INFO" />-->
+<!--  <logger name="play" level="INFO" />-->
 
-    <logger name="com.amazonaws">
-        <level value="ERROR" />
-    </logger>
+<!--  <root level="INFO">-->
```
mbarton: Should we leave this commented out?

marialivia16: I think it's sometimes useful locally when debugging, otherwise it's too noisy to uncomment. Happy to uncomment if you think the test logs are helpful.

mbarton: Nope that's fine just checking for typos 👍

akash1810:
I'm late to the party, however this is absolutely brilliant.

img

PR #2534, opened 2019-04-23
Merge all the Scala app riff-raff.yaml files

mbarton:
Move to a single Riff-Raff deploy for all Scala services, following the same model as [frontend](https://github.com/guardian/frontend/blob/master/riff-raff.yaml).

Requires https://github.com/guardian/editorial-tools-platform/pull/268 for the updated AMI parameters.

It's still possible to deploy single services if required by clicking "Preview" in Riff-Raff.

I have left reaper and s3-watcher with their own `riff-raff.yaml` files for now, since they use [node-riffraff-artefact](http://github.com/guardian/node-riffraff-artefact) and it would be non-trivial to replicate that in the SBT build.


mbarton:
[Use single TeamCity build and dotcom-style riff-raff.yaml for the Grid](https://trello.com/c/GZfBtr2L/20-use-single-teamcity-build-and-dotcom-style-riff-raffyaml-for-the-grid)

riff-raff.yaml
```diff
@@ -0,0 +1,102 @@
...
+    type: ami-cloudformation-parameter
+    parameters:
+      cloudFormationStackByTags: false
+      cloudFormationStackName: media-service
+      prependStackToCloudFormationStackName: false
+      amiTags:
+        BuiltBy: amigo
+        AmigoStage: PROD
+        Recipe: grid-xenial
+      amiParameter: ImagingAmiId
```
sihil: I think both `ami-update` and `imaging-ami-update` will try to run simultaneously? Might be worth merging them into a single block to avoid confusion:
```yaml
  ami-update:
    type: ami-cloudformation-parameter
    parameters:
      cloudFormationStackByTags: false
      cloudFormationStackName: media-service
      prependStackToCloudFormationStackName: false
      amiParametersToTags:
        AmiId:
          BuiltBy: amigo
          AmigoStage: PROD
          Recipe: editorial-tools-xenial-java8
        ImagingAmiId:
          BuiltBy: amigo
          AmigoStage: PROD
          Recipe: grid-xenial
```

sihil: In fact doing this would simplify the file as the two templates are only needed because of the two different AMI updates.

PR #2535, opened 2019-04-23
Rename Witness to Readers Photos

paperboyo:
There is plenty of code we could remove since the death of n0tice, instead here is just cosmetics. 

image

image

Yeah, I know it should be `Readers’ Photos`, but wouldn’t play nice with the URL and stuff.

- [x] tested on TEST

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
@@ -261,7 +261,7 @@ final case class GuardianWitness(restrictions: Option[String] = None) extends Us
   val defaultCost = GuardianWitness.defaultCost
 }
 object GuardianWitness extends UsageRightsSpec {
-  val category = "guardian-readers-photos"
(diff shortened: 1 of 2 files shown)


akash1810:
I think we'd want to release this close to the time when we're able to bulk update existing images.

paperboyo:
> I think we'd want to release this close to the time when we're able to bulk update existing images.

👍 2035? ;-)

akash1810:
Have just come up against an issue with this change, which I should have thought about earlier, sorry! This is causing serialisation issues, previous images will fail to serialise as the `guardian-witness` won't be a recognised value.

We'd need to add another `UsageRightsSpec` rather than update an existing one.

paperboyo:
@akash1810 I just changed the category name back

mbarton:
This is blocked as it would require a re-index. Not closing as we may want to do so in the future if we build an easier way of reindexing

sihil:
Closing as @paperboyo doesn't consider it that important.

PR #2536, opened 2019-04-24
Consolidate AMI updates and simplify

sihil:
A few suggested improvements on @mbarton's work in #2534 

The key one is consolidating the AMI update tasks into one. There is currently no dependency between the two which will mean they try to run simultaneously and I'm pretty sure that'll be non-deterministic. I think it has worked so far due to there being no new AMIs.

Once this is done, the whole config file can be further simplified.

### DIFF
--- a/riff-raff.yaml
+++ b/riff-raff.yaml
@@ -12,19 +12,19 @@ templates:
     dependencies:
       - ami-update
 
-  imaging-autoscaling:
-    type: autoscaling
-    parameters:
-      bucket: media-service-dist
-    dependencies:
-      - imaging-ami-update
-
   usage-autoscaling:
     type: autoscaling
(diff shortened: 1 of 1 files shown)


mbarton:
Ah good catch, I hadn't thought they would race to update the same cloudformation. The AMIs are different so I guess we would consolidate around the one that has graphicsmagick etc installed? It would lead to extra unused stuff in the AMI for most of the services but that won't be harmful and would be fixed if we ever went to containers

sihil:
It still updates both AMIs, but in the same deploy unit. It now uses the poorly documented `amiParametersToTags`.

sihil:
This has deployed successfully in TEST so I'm going to merge it.

PR #2537, opened 2019-04-24
Feature/configurable hardcoded properties

jamie-pitts:
This should be non-breaking, all properties should default to what they previously were if they have not been set.

Three main properties have been made configurable:
- Thrall kinesis stream name, this can be set in the relevant /etc/gu/ props files with the property key `thrall.kinesis.stream.name`
- Elastic search protocol (i.e. http or https) this defaults to http, and can be set in the /etc/gu/ props files (thrall + api) with the property key `es6.protocol` 
- The host name prefix for the services. E.g. <prefix>domain-root. This is set in the .conf file (as it is required by all the components) an example configuration is as follows:
```
hosts {
  kahunaPrefix: ""
  apiPrefix: "api."
  loaderPrefix: "loader."
  cropperPrefix: "cropper."
  metadataPrefix: "metadata."
  imgopsPrefix: "imgops."
  usagePrefix: "usage."
  collectionsPrefix: "collections."
  leasesPrefix: "leases."
  authPrefix: "auth."
}
```

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
@@ -46,7 +46,7 @@ trait CommonConfig {
   val isProd: Boolean = stage == "PROD"
   val isDev: Boolean = stage == "DEV"
 
-  final val thrallKinesisStream = s"$stackName-thrall-$stage"
+  final val thrallKinesisStream = properties.getOrElse("thrall.kinesis.stream.name", s"$stackName-thrall-$stage")
 
   // Note: had to make these lazy to avoid init order problems ;_;
   lazy val domainRoot: String = properties("domain.root")
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
@@ -50,13 +50,27 @@ trait CommonConfig {
 
   // Note: had to make these lazy to avoid init order problems ;_;
   lazy val domainRoot: String = properties("domain.root")
-  lazy val services = new Services(domainRoot, isProd)
+  lazy val rootAppName: String = properties.getOrElse("app.name.root", "media")
+  lazy val serviceHosts = ServiceHosts(stringDefault("hosts.kahunaPrefix", s"$rootAppName."),
+                                       stringDefault("hosts.apiPrefix", s"api.$rootAppName."),
+                                       stringDefault("hosts.loaderPrefix", s"loader.$rootAppName."),
+                                       stringDefault("hosts.cropperPrefix", s"cropper.$rootAppName."),
+                                       stringDefault("hosts.metadataPrefix", s"$rootAppName-metadata."),
+                                       stringDefault("hosts.imgopsPrefix", s"$rootAppName-imgops."),
+                                       stringDefault("hosts.usagePrefix", s"$rootAppName-usage."),
+                                       stringDefault("hosts.collectionsPrefix", s"$rootAppName-collections."),
+                                       stringDefault("hosts.leasesPrefix", s"$rootAppName-leases."),
+                                       stringDefault("hosts.authPrefix", s"$rootAppName-auth."))
+  lazy val services = new Services(domainRoot, isProd, serviceHosts)
 
   final def apply(key: String): String =
     string(key)
 
   final def string(key: String): String =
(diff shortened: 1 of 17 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch6/ElasticSearchClient.scala
```diff
@@ -20,6 +20,8 @@ trait ElasticSearchClient {
 
   def port: Int
 
+  def protocol: String
```
mbarton: Shall we just combine all three into a single `url` parameter for ES6? Definitely would be a breaking change though so maybe best as a separate PR

mbarton:
I'll let @akash1810 have a look before merging

jamie-pitts:
@akash1810 I've pushed some changes with your suggestions - making the kinesis stream property now required (no default), and instead of changing the `es6.protocol` property a boolean, changing the properties to a single `es6.url` property like @mbarton suggested. This is now a breaking change as these two properties will need to be set

common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
```diff
@@ -46,18 +46,32 @@ trait CommonConfig {
   val isProd: Boolean = stage == "PROD"
   val isDev: Boolean = stage == "DEV"
 
-  final val thrallKinesisStream = s"$stackName-thrall-$stage"
+  final val thrallKinesisStream = properties("thrall.kinesis.stream.name")
```
akash1810: We'd want this to be a `lazy val` rather than a `final`.

The reasoning is a bit silly... `lazy` means it's only executed when accessed. As the [test creates a ThrallConfig](https://github.com/guardian/grid/blob/master/thrall/test/lib/ElasticSearch6Test.scala#L13), `final` will mean the test is trying to find this property in the config file `/etc/gu/thrall.properties` which it won't find as the file doesn't exist and so the build fails. It's silly because the test isn't even using `ThrallConfig` for anything, its just part of the class that's being tested.

Ideally we'd restructure that part of the code, however that probably belongs in its own PR. Moving to `lazy` allows us to be... lazy.

akash1810: If we rebase off guardian/master, we can run `scripts/ci/grid.sh` to witness what CI is doing.

jamie-pitts: Ah that makes sense. I have created a PR to merge into this branch with the fix (I've lost push access to this branch i think due to the rebase)

mbarton: I've approved and merged the fix

mbarton: We should consider adding something that allows us to approve CI to run on branches. [Akka has a good model](https://github.com/akka/akka/pull/26770#issuecomment-484316945) where one of the core team can comment on a PR to have it built and tested

jamie-pitts: Thanks! I'm a bit confused as to what's happened but the branch seems to have moved to here now: https://github.com/guardian/grid/tree/bbc_feature/configurable-hardcoded-properties so I am assuming we should close this PR, and open a PR there?

jamie-pitts:
Closed, see https://github.com/guardian/grid/pull/2546 for the updated PR of this

common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
akash1810: > We'd want this to be a `lazy val` rather than a `final`.
> 
> The reasoning is a bit silly... `lazy` means it's only executed when accessed. As the [test creates a ThrallConfig](https://github.com/guardian/grid/blob/master/thrall/test/lib/ElasticSearch6Test.scala#L13), `final` will mean the test is trying to find this property in the config file `/etc/gu/thrall.properties` which it won't find as the file doesn't exist and so the build fails. It's silly because the test isn't even using `ThrallConfig` for anything, its just part of the class that's being tested.
> 
> Ideally we'd restructure that part of the code, however that probably belongs in its own PR. Moving to `lazy` allows us to be... lazy.

https://github.com/guardian/grid/pull/2548 should help resolve this.

Will look at rebasing #2546 once that's gone in.

PR #2540, opened 2019-04-25
rationalise scripts into a single script

akash1810:
Removes `kahuna/dist.sh` and `kahuna/setup.sh` as they are simply calling `npm run` and we can do that directly.

Creates `scripts/ci/grid.sh` to clean and build the client side and server side apps and riffRaffUpload. This single script means `scripts/ci/kahuna-client-side-build.sh` isn't necessary anymore.

Running a script rather than using teamcity's sbt plugin is necessary as the plugin always returns with exit code 0... even on error 😱 .


akash1810:
cc @jamie-pitts in case you're relying on these scripts

PR #2542, opened 2019-04-26
Add issue and PR templates

akash1810:
GitHub have recently added issue and PR templates https://github.com/blog/2111-issue-and-pull-request-templates.

Although we follow Tackley's Law (TM) on creating issues, the template would act as a prompt for the details needed.


paperboyo:
> Tackley's Law (TM) on creating issues

Oh, I’m sure I follow it without actually knowing about it. Could we paste it here for posterity? (or provide a link in case it’s explicit lyrics?)

PR #2546, opened 2019-04-30
Bbc feature/configurable hardcoded properties

jamie-pitts:
## What does this change?
Four main properties have been made configurable:

Thrall kinesis stream name, this can be set in the relevant /etc/gu/ props files with the property key `thrall.kinesis.stream.name`

Elastic search url, this should be the full url of the ES cluster including the port number, e.g. `http://localhost:9200`, and can be set in the /etc/gu/ props files (thrall + api) with the property key `es6.url`. Note: `es6.host` and `es6.port` have been removed with this change.

The host name prefix for the services. E.g. domain-root. This is set in the .conf file (as it is required by all the components) an example configuration is as follows:
```hosts {
  kahunaPrefix: ""
  apiPrefix: "api."
  loaderPrefix: "loader."
  cropperPrefix: "cropper."
  metadataPrefix: "metadata."
  imgopsPrefix: "imgops."
  usagePrefix: "usage."
  collectionsPrefix: "collections."
  leasesPrefix: "leases."
  authPrefix: "auth."
}
```

The bucket for where the permissions file is located. This can be set with the key `permissions.bucket` in the relevant property files. It looks like it is used in the api, auth, and cropper services.

## Tested?
- [x] locally
- [x] on TEST

_Additionally in our deployed env_

**Breaking change, es6.url and thrall.kinesis.stream.name both need setting.**


akash1810:
I've tested this on our pre-PROD environment and cannot see any issues.

Not sure where the last two commits have come from, can we clean this branch up before merging as those commits are already on master.

akash1810:
> Not sure where the last two commits have come from, can we clean this branch up before merging as those commits are already on master.

Oh hang on, https://github.com/guardian/grid/pull/2544/files was a PR into this branch rather than master, so the permissions bucket change is not on master and this PR is ok to merge.

PR #2549, opened 2019-05-02
Move Getty parser to the beginning

paperboyo:
Moves Getty parser to the beginning. This image (455c859312d6e386a78ee2455f4523f7032378d1) is from Bloomberg but the assumption to test by this PR is that because it contains a GettyGIFT block (and we, erroneously, assume it means – it’s from Getty), Bloomberg categorisation is being overridden by the Getty parser further in the file.

- [ ] test on TEST


paperboyo:
Closing as the order is controlled somewhere else :facepalm:.

PR #2556, opened 2019-05-03
remove thrall sqs message consumption

akash1810:
## What does this change?
As we're no longer writing to the Thrall SNS topic (#2524), Thrall has nothing to consume from the SQS queue. This PR removes the code that reads from SQS and also the code that writes to ES1.

Thrall has no need to know about ES1 now!

We're also now able to remove the abstract classes, however that will be another PR to reduce the amount of change here.

Thrall continues to report healthy as a result of #2555.

img

## Tested?
- [ ] locally
- [ ] on TEST


akash1810:
https://github.com/guardian/grid/pull/2556#issuecomment-490478097 is solved here https://github.com/guardian/grid/pull/2561

PR #2559, opened 2019-05-09
update manifest endpoints to return buildinfo as json

akash1810:
## What does this change?
Updates `/management/manifest` to return the buildinfo as json using the [sbt-buildinfo plugin](https://github.com/sbt/sbt-buildinfo).

Somewhere in the Play upgrade we stopped using https://github.com/guardian/sbt-version-info-plugin which meant the `/management/manifest` endpoints are returning stale content. This PR replaces that stale content.

`/management/manifest` is a handy way to see what version of the app is running as it includes the last commit id. Hopefully helping to solve https://github.com/guardian/grid/pull/2556#issuecomment-490478097.

Helping to solve https://github.com/guardian/grid/pull/2556#issuecomment-490478097.

image

## Who should look at this?

@guardian/digital-cms 

## Tested?
- [x] locally
- [x] on TEST

### DIFF
--- a/auth/app/auth/AuthComponents.scala
+++ b/auth/app/auth/AuthComponents.scala
@@ -11,8 +11,10 @@ class AuthComponents(context: Context) extends GridComponents(context) {
   final override lazy val config = new AuthConfig(configuration)
   final override lazy val httpConfiguration = AuthHttpConfig(configuration, context.environment)
 
+  final override val buildInfo = utils.buildinfo.BuildInfo
+
   val controller = new AuthController(auth, config, controllerComponents)
-  val permissionsAwareManagement = new ManagementWithPermissions(controllerComponents, controller)
+  val permissionsAwareManagement = new ManagementWithPermissions(controllerComponents, controller, buildInfo)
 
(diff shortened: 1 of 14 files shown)


RichieAHB:
Only question is did anything rely on the previous manifest? Otherwise it looks good 👍

akash1810:
> Only question is did anything rely on the previous manifest? Otherwise it looks good 👍

I don't think so. If they did, there's been no noise about it being broken for a while.

sihil:
If we rely on this it would be great to introduce a test so that we don't regress it again.

akash1810:
> If we rely on this it would be great to introduce a test so that we don't regress it again.

I think we can cover this by adding prout?

rtyley:
> I think we can cover this by adding prout?

Prout will tell you when your pull-request is out, and do it based on the manifest url you provide - it will issue 'OVERDUE' warnings if a merged pull request isn't seen in the manifest after a time-out period ytou specify.

https://www.theguardian.com/info/developer-blog/2015/feb/03/prout-is-your-pull-request-out

PR #2565, opened 2019-05-10
remove ES1 from DEV

akash1810:
## What does this change?
Removes ES1 cluster from DEV environment (via `docker-compose.yml`) and also moves ES6 cluster to the standard `9200` port.

Forget about ES1.

img

## Who should look at this?

@guardian/digital-cms

## Tested?
- [x] locally
- [ ] on TEST


kenoir:
Begs the question _why_ we needed 2 ES clusters for testing ever.

akash1810:
> Begs the question _why_ we needed 2 ES clusters for testing ever.

We were dual stacking an ES1 cluster and an ES6 cluster during an ES upgrade to avoid downtime. Now that it's finished, we've dropped the ES1 cluster.

PR #2568, opened 2019-05-13
added initial support for tiff image types

jamie-pitts:
## What does this change?
Allows for uploading of tiff images, this has been safeguarded by having the optional configurable property `transcoded.mime.types` in the image-loader .properties which expects a comma separated list of image types (thinking for the future), so for now this can be set to just `image/tiff`. 

The main idea of this implementation is to keep the tiff image and use it for crops, but to also convert it into a png and then optimise it (using the png optimisation that was already there) for displaying the image to the user. Additionally crops are created as either jpeg or png depending on if the image has transparency (using similar rules as was already present).

Metadata extraction seems to be working as expected - there is no TiffDirectory to be used, but from research it seems the EXIFD0 directory is always present so is safe to use. I haven't looked into the colour space/metadata too much, and so some more work could potentially be done there in the future.

Tiff images can be successfully uploaded and viewed within the grid. Also that they can then be downloaded back again

## Tested?
- [x] locally
- [ ] on TEST
- [x] bbc cloud deployed env

---

Happy to arrange a meet up if you would prefer to discuss this in person/discuss different approaches :)


cropper/app/lib/Crops.scala
```diff
@@ -108,15 +108,15 @@ class Crops(config: CropperConfig, store: CropStore, imageOperations: ImageOpera
...
-    *  - If the PNG has transparency then it should always be a PNG as the transparency is not available in JPEG
-    *  - If the PNG is not true colour then we assume it is a graphic that should be retained as a PNG
+    *  - If the image has transparency then it should always be a PNG as the transparency is not available in JPEG
+    *  - If the image is not true colour then we assume it is a graphic that should be retained as a PNG
     */
   def cropType(mediaType: String, colourType: String, hasAlpha: Boolean): MimeType = {
-    val isPng = mediaType == "image/png"
-    val isGraphic = !colourType.startsWith("True Color")
-    val retainAsPng = hasAlpha || isGraphic
+    val isJpeg = mediaType == "image/jpeg"
```
kenoir: Might be an idea to make `mediaType` a sum type i.e.

```scala
sealed trait MediaType {
  val id: String
}

case object ImageJpeg extends MediaType {
  override val id: String = "image/jpeg"
}

//etc
```

image-loader/app/lib/imaging/FileMetadataReader.scala
```diff
@@ -150,34 +150,54 @@ object FileMetadataReader {
           Dimensions(width, height)
         }
 
+        case Some("image/tiff") => for {
+          exifDir <- Option(metadata.getFirstDirectoryOfType(classOf[ExifIFD0Directory]))
```
kenoir: Blech, this whole metadata lib needs a Scala wrapper.

Consider starting to push out interactions with that lib _elsewhere_, or even give it the typeclass treatment and abstract implementation away entirely.

image-loader/app/lib/imaging/MimeTypeDetection.scala
```diff
@@ -2,34 +2,12 @@ package lib.imaging
 
 import java.io.File
 
+import org.apache.tika.Tika
```
kenoir: 👏

paperboyo:
This is 💎!

PR #2571, opened 2019-05-20
Feature/configurable test properties

felixcs1:
Hi! I'm Felix a graduate software engineer working with Jamie Pitts on the grid for a few months. 

## What does this change?
Our CI process (jenkins pipeline) has limited memory and was failing to create the Elasticsearch docker image for Thrall and Media-Api tests. To work around this we have created a `test.properties` file that determines if Docker is used for the tests. 

The file is completely optional and values are defaulted if it's not present.

This file can be used for other configurable test properties in the future.

To test locally, create a `/etc/gu/test.properties` files with the contents
```
es6.test.url=http://localhost:9200
es6.useDocker=false
```
run elasticsearch locally (or locally run an ES docker image) and then run thrall and Api tests.

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/thrall/test/syndication/SyndicationRightsOpsTestsBase.scala
+++ b/thrall/test/syndication/SyndicationRightsOpsTestsBase.scala
@@ -15,8 +15,12 @@ import org.scalatest.{BeforeAndAfterAll, FreeSpec, Matchers}
 import play.api.libs.json.Json
 
 import scala.concurrent.duration._
+import scala.util.Properties
 
 trait SyndicationRightsOpsTestsBase extends FreeSpec with Matchers with Fixtures with BeforeAndAfterAll with ScalaFutures with DockerKit with DockerTestKit with DockerKitSpotify {
+
+  val useEsDocker = Properties.envOrElse("ES6_USE_DOCKER", "true").toBoolean
+  val es6TestUrl = Properties.envOrElse("ES6_TEST_URL", "http://localhost:9200")
 
   def ES: ElasticSearchVersion
   def esContainer: Option[DockerContainer]
--- a/media-api/test/lib/elasticsearch/ElasticSearchTestBase.scala
+++ b/media-api/test/lib/elasticsearch/ElasticSearchTestBase.scala
@@ -13,11 +13,15 @@ import org.scalatest.time.{Milliseconds, Seconds, Span}
 import org.scalatest.{BeforeAndAfterAll, FunSpec, Matchers}
 
 import scala.concurrent.duration._
+import scala.util.Properties
 
 trait ElasticSearchTestBase extends FunSpec with BeforeAndAfterAll with Matchers with ScalaFutures with Fixtures with DockerKit with DockerTestKit with DockerKitSpotify {
 
   val interval = Interval(Span(100, Milliseconds))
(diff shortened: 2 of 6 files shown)


thrall/test/syndication/SyndicationRightsOpsTestsBase.scala
```diff
@@ -18,6 +18,10 @@ import scala.concurrent.duration._
 
 trait SyndicationRightsOpsTestsBase extends FreeSpec with Matchers with Fixtures with BeforeAndAfterAll with ScalaFutures with DockerKit with DockerTestKit with DockerKitSpotify {
 
+  val testProps = getTestProperties
+  val es6TestUrl = testProps.getOrElse("es6.test.url", "http://localhost:9206")
```
akash1810: As of https://github.com/guardian/grid/pull/2565 we've removed the ES1 cluster completely, that is we can use the more standard `9200` port here.

kenoir:
> Interesting. This raises the question of whether or not launching an ES cluster in CI should be done in code or by CI itself. I think I'm leaning toward the latter; our CI currently runs [this script](https://github.com/guardian/grid/blob/master/scripts/ci/grid.sh) and I think it would be sensible to launch ES there.

IMHO build/test processes should be as agnostic of CI environment as possible (to enable portability and reproducibility of the testing env - you want your local env to be the same as test so those tests are valid).

Docker is a good way of achieving that.

As to whether it makes sense to reduce the memory footprint of tests using a _real_ ES cluster, it absolutely does. I'd lean towards abstracting the interface to ES away so as to be able to replace it with fakes in tests wherever possible. 

Then you only need a real ES cluster in end-to-end and perf testing (which you can do separately).

felixcs1:
> * Are you able to share the spec of the instance of Jenkins you're running in CI?

We have multiple Jenkins slave nodes running as a t2.medium, each node has 4GB of memory and can have multiple jobs running. We use declarative Jenkins pipelines, with each stage running inside a docker container. We found that the sbt compile/testing process takes about 2-3GB (Per service) and then ES takes more on top of that. As multiple jobs can be running on the same node this causes memory problems. We also weren't so keen on having the code launching docker inside docker running on Jenkins. Instead we launch a separate light weight docker ES image that is then linked to the image running the ES tests.

> * How do you ensure CI writes a file to `/etc/gu/test.properties` prior to running the sbt tests?

The `test.properties` file is built into the docker image running the ES tests.  

> * Do you ever change `es6.test.url` away from the default? If not, maybe a simpler solution here is to use environment variables? (Simpler as it removes file io.)

No we don't ever change `es6.test.url`, happy to look into using env vars if you feel this is preferable.

thrall/test/syndication/SyndicationRightsOpsTestsBase.scala
felixcs1: Sorry must have missed that when rebasing

felixcs1:
I have updated this PR to use environment variables as suggested. As it happens we DO change the es6 test url, on our CI pipeline we give it `http://<es-docker-container-name>:9200` which is needed for the way we are linking our testing container to our es container

PR #2574, opened 2019-05-28
Show progress for batch operations

mbarton:
Show numerical progress in the top bar when batch operations are applied.

2f3ebfd99d2daf5e94e656cbc45305d5

This will give people more confidence that something is still happening during a long running batch operation and give us visibility on performance improvements if/when we move more batch operations to the server.

Since each component (and each service) implements batch as it sees fit, I have added the tracking using events on the root scope.

I also made the progress appear above the existing cost overview to avoid the controls below jumping up and down when you edit them.

It can track edits to more than one field at once, although the UI tends to lock up anyway when you try and do that.

TODO:
- [x] prompt when refreshing the page if an operation is still in progress
- [x] work out why sometimes the info panel disappears once operations complete
- [x] where to put progress (can't see during full frame crop)
- [x] change color to white (match buttons)

### DIFF
--- a/kahuna/public/js/util/batch-tracking.js
+++ b/kahuna/public/js/util/batch-tracking.js
@@ -1,0 +1,18 @@
+// TODO MRB: invoke function lazily, does it improve UI jank?
+export function trackAll($rootScope, key, input, fn) {
+    let completed = 0;
+    $rootScope.$broadcast("events:batch-operations:start", { key, completed: 0, total: input.size ? input.size : input.length });
+
+    return Promise.all(input.map(item => fn(item).then(intermediateResult => {
+        completed++;
+        $rootScope.$broadcast("events:batch-operations:progress", { key, completed });
+
+        return intermediateResult;
+    }))).then(finalResult => {
+        $rootScope.$broadcast("events:batch-operations:complete", { key });
+        return finalResult;
+    }).catch(error => {
+        $rootScope.$broadcast("events:batch-operations:complete", { key });
+        throw error;
+    });
+}
--- a/kahuna/public/js/util/batch-tracking.js
+++ b/kahuna/public/js/util/batch-tracking.js
@@ -1,7 +1,8 @@
 // TODO MRB: invoke function lazily, does it improve UI jank?
 export function trackAll($rootScope, key, input, fn) {
     let completed = 0;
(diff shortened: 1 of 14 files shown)


paperboyo:
This is 💎!

> I also made the progress appear above the existing cost overview to avoid the controls below jumping up and down when you edit them.

Good. The downside: one may be editing a field so far away from the top that this won’t be visible. Let’s have a think. But even if we won’t (think) it’s brilliant already!

mbarton:
> The downside: one may be editing a field so far away from the top that this won’t be visible

Definitely. It's also not visible when you click the full-frame crop button which is another reason to put it somewhere else

paperboyo:
> The downside: one may be editing a field so far away from the top that this won’t be visible

Actually, I may have been wrong: while the updating happens, the long fields (eg. a super-long description) is collapse into _Multiple descriptions_, so all is good.

kahuna/public/js/util/batch-tracking.js
```diff
@@ -0,0 +1,18 @@
+// TODO MRB: invoke function lazily, does it improve UI jank?
+export function trackAll($rootScope, key, input, fn) {
```
akash1810: This could be a [factory](https://docs.angularjs.org/guide/providers#factory-recipe) to avoid passing `$rootScope` around.

paperboyo:
Looking great, greater by the minute!

Before the reshuffling of buttons in subsequent PRs, could we, as part of this one, make sure that the progress bar “button” does not elongate when progress goes from single to double digit (and, possibly triple, cos quadruple will kill the browser currently)?

dfsd
[notice the jump from 9 to 10]

Thx!

BTW, shitty UI mockups from an amateur of reshuffling for posterity and future PRs. Reasoning:
- less buttons jump around and keep to their places, less confusion and better habit forming
- selection action (Cancel selection could be called Deselect), should be further apart from “dangerous” and proper actions (like eg. Crop full frame).
- the reshuffling, ideally, would make some tiles that contain numbers fixed width so that changing the numbers won’t cause button movements (unless reserving every tile for millions would look ridiculous)

image2

image

paperboyo:
Tiny: I know I spoke highly of the shadow behind the text, but too early. Maybe we should match the rest of white text on blue?
image

PR #2576, opened 2019-05-29
Don't require credit to process bylines

itsibitzi:
We can handle a missing credit but not a missing byline so we shouldn't require both of them in the metadata cleaner

Also had to modify how the tests are run to deal with the fact there might not be a credit after the processor has run (e.g. when there's a byline such as `Sam Cutler` and no credit)

Hopefully the tests make sense?

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/BylineCreditReorganise.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/BylineCreditReorganise.scala
@@ -18,11 +18,9 @@ object BylineCreditReorganise extends MetadataCleaner {
     )
   }
 
-  def bothExist(byline: Field, credit: Field): Option[(String, String)] =
-    for(b <- byline; c <- credit) yield (b, c)
-
   def removeBylineFromCredit(bylineField: Field, creditField: Field) =
-    bothExist(bylineField, creditField).map { case (byline, credit) =>
+    bylineField.map { byline =>
+      val credit = creditField.getOrElse("")
       val bylineParts = byline.split("/").filter(!_.isEmpty)
       val creditParts = credit.split("/").filter(!_.isEmpty)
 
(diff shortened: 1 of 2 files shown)


paperboyo:
Just a comment to document current state of `BylineCreditReorganise` cleaner.

It only applies to Grid’s Byline and Credit fields.

Byline is composed from [these](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala#L39) fileMetadata fields, Credit – from [these](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala#L36).

_Historical bit: In [IPTC schema](https://www.iptc.org/std/photometadata/specification/IPTC-PhotoMetadata#credit-line) byline+source=credit (eg. Byline: `John Doe`; Source: `Intermediary/Agency`; Credit: `John Doe/Intermediary/Agency`). But because many of our suppliers use Source field for the name of the intermediary only and include Agency only in Credit, Grid decided to use Credit to come up with a Source. And called it Credit. Confusing, I know, and should be changed, but hey._

`BylineCreditReorganise` cleaner splits everything by `/` (but not by ` for ` or by ` via `) and:
- removes multiple spaces and spaces around slashes (`   John Doe / Intermediary` -> `John Doe/Agency`
- leaves Byline empty if it was empty on arrival, so eg. Credit: `John Doe/Agency` is not turned into Byline: `John Doe`, Credit: `Agency`. This is because some providers leave Byline empty and have no Byline in Credit (Credit: `Allstar/MGM`)
- if the Byline contains `/`, everything after the first `/` is removed from Byline and is prepended to the Credit (even if Credit is empty), deduplication is applied (eg. Byline: `John Doe/Intermediary/Agency`, Credit: `John Doe/Intermediary/Agency` becomes Byline: `John Doe`, Credit: `Intermediary/Agency`

Other cleaners are then applied in [this order](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/MetadataCleaner.scala#L15).

PR #2577, opened 2019-05-31
Add to PhotographerRenamer dictionary #10

paperboyo:
Thx!


paperboyo:
Thanks! One day we will all be dead…

PR #2579, opened 2019-06-04
Use imagemagick for tiff image operations

mbarton:
Use imagemagick for calls to `convert` imaging operations. It produces correct colours in the resulting crops where graphicsmagick does not.


mbarton:
cc @jamie-pitts @felixcs1. I'll look to merge this once the original TIFF support PR is merged (#2568)

PR #2592, opened 2019-06-21
add is search

akash1810:
## What does this change?
As we cannot perform `or` searches, the `is` query is a handy shortcut for pre-saved queries. For example, `is:staff` represents a staff, commissioned or contract photographer.

We can perform more complex queries with ease.

n/a

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/media-api/app/lib/elasticsearch/impls/elasticsearch6/IsQueryFilter.scala
+++ b/media-api/app/lib/elasticsearch/impls/elasticsearch6/IsQueryFilter.scala
@@ -1,0 +1,28 @@
+package lib.elasticsearch.impls.elasticsearch6
+
+import com.gu.mediaservice.lib.ImageFields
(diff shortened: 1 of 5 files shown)


akash1810:
similar to https://github.com/guardian/grid/pull/2190

paperboyo:
💎 
Just a thought: open to suggestions!

`is:staff`: all 6 categories
`is:staff photo`: 3 photography (sub)categories
`is:staff illustration`: 3 illustration (sub)categories

I know that nobody will be using the latter, but WDYT?

akash1810:
> 💎
> Just a thought: open to suggestions!
> 
> `is:staff`: all 6 categories
> `is:staff photo`: 3 photography (sub)categories
> `is:staff illustration`: 3 illustration (sub)categories
> 
> I know that nobody will be using the latter, but WDYT?

Sounds sensible, let's do that as a different PR though?

PR #2593, opened 2019-06-21
add `is` chip to kahuna

akash1810:
## What does this change?
Adds `is` chip to the chips.

Requires https://github.com/guardian/grid/pull/2592.

We can perform more powerful search

image

If we can't serialise into a `IsQueryFilter`:
image

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/kahuna/public/js/search/structured-query/query-suggestions.js
+++ b/kahuna/public/js/search/structured-query/query-suggestions.js
@@ -34,7 +34,8 @@ export const filterFields = [
     'croppedBy',
     'filename',
     'photoshoot',
-    'leasedBy'
+    'leasedBy',
+    'is'
 ].sort();
 // TODO: add date fields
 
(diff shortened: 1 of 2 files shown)


akash1810:
similar to https://github.com/guardian/grid/pull/2191

kahuna/public/js/search/structured-query/query-suggestions.js
```diff
@@ -60,6 +61,10 @@ const subjects = [
     'weather'
 ];
 
+const isSearch = [
+  'staff'
```
akash1810: this should probably be returned by an endpoint, to be more scalable...

mbarton: Agreed although at least it's isolated here. If we do make an API for it I suggest we follow the data model @itsibitzi defined for the new PFI search component: https://github.com/guardian/threads/tree/master/src/components/search. That way switching it in will be easier

PR #2596, opened 2019-06-24
additional is searches

akash1810:
## What does this change?
Per email thread with Jonny and Jo:
- renames `is:staff` to `is:gnm-owned-photo`
- adds two new `is` searches:
  - `is:gnm-owned-illustration`
  - `is:gnm-owned`

Ideally this would also take the form of a checkbox/select box...

More useful search.

image

img

## Tested?
- [x] locally
- [x] on TEST

### DIFF
--- a/kahuna/public/js/search/structured-query/query-suggestions.js
+++ b/kahuna/public/js/search/structured-query/query-suggestions.js
@@ -62,6 +62,8 @@ const subjects = [
 ];
 
 const isSearch = [
+  'GNM-owned-photo',
+  'GNM-owned-illustration',
   'GNM-owned'
 ];
 
--- a/kahuna/public/js/search/structured-query/query-suggestions.js
+++ b/kahuna/public/js/search/structured-query/query-suggestions.js
@@ -62,7 +62,7 @@ const subjects = [
 ];
 
 const isSearch = [
-  'staff'
+  'GNM-owned'
 ];
 
 querySuggestions.factory('querySuggestions', ['mediaApi', 'editsApi', function(mediaApi, editsApi) {
--- a/media-api/test/lib/elasticsearch/ConditionFixtures.scala
+++ b/media-api/test/lib/elasticsearch/ConditionFixtures.scala
@@ -15,9 +15,11 @@ trait ConditionFixtures {
   val dateMatchCondition = Match(SingleField("adatefield"), DateRange(dateRangeStart, dateRangeEnd))
 
   val hasFieldCondition = Match(HasField, HasValue("foo"))
+
   val isOwnedPhotoCondition = Match(IsField, IsValue(IsOwnedPhotograph.toString))
   val isOwnedIllustrationCondition = Match(IsField, IsValue(IsOwnedIllustration.toString))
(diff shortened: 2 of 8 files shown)


mbarton:
Is there a wording we can use that is not Guardian-specific? Or maybe a config parameter for the organisation name or to turn the feature off altogether.

akash1810:
👍 to pushing this to config. Something like `company.name`... For a different PR, however.

kahuna/public/js/search/structured-query/query-suggestions.js
```diff
@@ -62,7 +62,9 @@ const subjects = [
 ];
 
 const isSearch = [
-  'staff'
+  'GNM-owned-photo',
```
akash1810: regarding this being Guardian specific and not generic enough - I think it would be better to model this as

```javascript
const isSearch = [{key: 'GNM Owned Photo', value: 'owned-photo', ...}];
```

That is, keep Guardian specific terminology in the rendering tier; Kahuna will make an api request passing `owned-photo` as the value.

This should also mean the UI becomes more human readable with the removal of kebab cased values.

paperboyo: > This should also mean the UI becomes more human readable with the removal of kebab cased values.

On the one hand, this is welcomed, although the correct spelling would be `GNM-owned photo` (to follow our style guide for adjectival hyphenation).

On the other hand, there is some value in URL reflecting the values in UI one-to-one. And we have precedents here:

image

mbarton: > keep Guardian specific terminology in the rendering tier

Not sure I follow, would the `company.name` parameter be in Kahuna config or Kahuna would have the GNM name hard-coded?

akash1810: Kahuna would have `company.name` in config. What I meant by:

> keep Guardian specific terminology in the rendering tier

is that the server will have no knowledge of `GNM`, meaning the client will serialise `${company.name} Owned Photo` into the correct value.

mbarton:
LGTM 👍

media-api/test/lib/elasticsearch/ConditionFixtures.scala
```diff
@@ -14,6 +15,12 @@ trait ConditionFixtures {
   val dateMatchCondition = Match(SingleField("adatefield"), DateRange(dateRangeStart, dateRangeEnd))
 
   val hasFieldCondition = Match(HasField, HasValue("foo"))
+
+  val isOwnedPhotoCondition = Match(IsField, IsValue(IsOwnedPhotograph.toString))
+  val isOwnedIllustrationCondition = Match(IsField, IsValue(IsOwnedIllustration.toString))
+  val isOwnedImageCondition = Match(IsField, IsValue(IsOwnedImage.toString))
+  val isInvalidCondition = Match(IsField, IsValue("a-random-string"))
```
paperboyo: Entirely not sure if this, but something prevents typing-ahead to select values in `is:` chip. Not a biggie, but just wanted to highlight, @akash1810.

PR #2597, opened 2019-06-25
tweak search placeholder

akash1810:
## What does this change?
Adjust phrasing to suggest that chips are available (brilliant suggestion by @sihil)

Also adjusts the colour to make it more readable

Similar to https://github.com/guardian/grid/pull/2290

making things more obvious

Before
image

After
image

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/kahuna/public/js/components/gr-chips/gr-chips.css
+++ b/kahuna/public/js/components/gr-chips/gr-chips.css
@@ -84,3 +84,7 @@ gr-static-filter-chip,
 gr-filter-chip {
     margin-right: 5px;
 }
+
+gr-text-chip input::placeholder {
+  color: #888;
+}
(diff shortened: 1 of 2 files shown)


sihil:
Do we track searches? Can we plot number of normal and advanced searches (perhaps excluding @paperboyo which would throw the results) in Kibana?

akash1810:
> Do we track searches? Can we plot number of normal and advanced searches (perhaps excluding @paperboyo which would throw the results) in Kibana?

We don't and probably should. We could add the search as markers into the [log line](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch6/ElasticSearch6Executions.scala#L36).

PR #2475, opened 2019-02-27
Add includeOverQuota option to search

mbarton:
Filters out images that are over quota by inserting the equivalent of a `-supplier:XXX` chip for each supplier that is over quota.

PR #2598, opened 2019-06-25
is over quota

akash1810:
## What does this change?
Adds is:over-quota search. Unfortunately this means passing a `UsageStore` really far down in order to get store updates, which isn't especially clean... if only scala could pass by reference! (`IsQueryFilter`'s apply could take an anonymous function rather than a `UsageStore`).

UPDATE: it's a lot cleaner with anonymous function and the tests are easier to write too!

https://github.com/guardian/grid/pull/2475 take II.

We can search for images that are over/under quota.

n/a

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/media-api/app/MediaApiComponents.scala
+++ b/media-api/app/MediaApiComponents.scala
@@ -26,13 +26,13 @@ class MediaApiComponents(context: Context) extends GridComponents(context) {
     replicas = config.elasticsearch6Replicas
   )
 
-  val elasticSearch = new lib.elasticsearch.impls.elasticsearch6.ElasticSearch(config, mediaApiMetrics, es6Config)
-  elasticSearch.ensureAliasAssigned()
-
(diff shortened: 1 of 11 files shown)


akash1810:
This new query will return all images from over quota agencies. I suspect there is more value in returning images still under quota (ie the inverse) as they represent images that can still be used. This can be achieved with a negation chip.

Will confirm with Jo and others what the exact use case is.

mbarton:
Maybe remove the UI changes from this PR and keep the query? Then we can wire it into the frontend in a later PR however we like

akash1810:
> Maybe remove the UI changes from this PR and keep the query? Then we can wire it into the frontend in a later PR however we like

Have made it it's own PR (https://github.com/guardian/grid/pull/2601). Nothing stopping it going out as a chip and then iterate on a more friendly and approachable UX.

PR #2599, opened 2019-06-28
introduce gr-top-bar-row

akash1810:
## What does this change?
As a way to introduce more search shortcuts to the search page's top bar, we need the idea of rows where the second row will house the shortcuts.

No shortcuts added in this PR, just adding the extra space.

Provides more space to add search shortcuts.

image

## Tested?
- [x] locally
- [x] on TEST

### DIFF
--- a/kahuna/public/js/search/view.html
+++ b/kahuna/public/js/search/view.html
@@ -1,5 +1,6 @@
 <!-- TODO: rename this file and split it into it's components -->
-<gr-top-bar fixed="true">
+<gr-top-bar fixed="true" class="search-bar">
+  <gr-top-bar-row>
     <gr-top-bar-nav>
         <search-query></search-query>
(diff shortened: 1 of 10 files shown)


kahuna/public/js/search/view.html
```diff
@@ -14,6 +15,11 @@
         </a>
         <file-uploader class="top-bar-item side-padded"></file-uploader>
     </gr-top-bar-actions>
+  </gr-top-bar-row>
+
+  <gr-top-bar-row small="true">
+
```
akash1810: new search components will go here

akash1810:
Worth noting that this PR shouldn't be merged until we actually start to make use of the extra space - otherwise we're taking up 25px of real estate for no reason...

mbarton:
Closing for now. We can revive it if required in the future

PR #2603, opened 2019-07-03
updating setup... again

akash1810:
## What does this change?
Separates installation of dependencies from running the app.

Previously, `dev-start.sh` would check that dependencies are installed. This isn't necessary after the first time - that is dependency installation is a one off process that is now in a `setup.sh` script.

Also moves all scripts into the `scripts` directory.

Running `dev-start.sh` is faster!

n/a

## Tested?
- [x] locally
- [ ] on TEST


Brewfile
```diff
@@ -1,13 +1,15 @@
+cask "AdoptOpenJDK/openjdk/adoptopenjdk8"
```
akash1810: making Brewfile fully exhaustive

PR #2607, opened 2019-07-04
invert over-quota to under-quota

akash1810:
## What does this change?
Simpler query language.

It is more common to ask "what images can I use (under-quota)" rather than "what images can I not use (over-quota)".

Whilst the former could be expressed as not(over-quota), it is by far the more common query so should be the default; to get over-quota we now do not(under-quota).

Relates to #2605.

Less confusion?

img

## Who should look at this?

@guardian/digital-cms

## Tested?
- [x] locally
- [ ] on TEST


paperboyo:
Thank you for going an extra mile!

giphy-downsized

PR #2608, opened 2019-07-08
Use source field to figure out some more suppliers

paperboyo:
After a [good change](https://github.com/guardian/grid/pull/2575) to cleaners we have been missing some images from ~[Alamy](https://media.gutools.co.uk/search?query=-%22alamy%20live%20news%22%20alamy%20-has:usageRights&since=2019-05-28T23:00:00.000Z&nonFree=true&until=2019-07-08T22:59:59.999Z)~ and [PA](https://media.gutools.co.uk/search?query=%22pa%20wire%22%20uploader:pa%20-has:usageRights&since=2019-05-28T23:00:00.000Z&nonFree=true&until=2019-07-08T22:59:59.999Z), because the whole cleaned `credit` can suddenly contain token(s) moved from `byline`.

~Here, for Alamy, we add matching by `source`.~ For PA, we do the same, but I had to remove the `credit` cleaner of `[different spellings]` > `PA`, because eg. cleaned `BBC/PA` could then be turned into just `PA`. There are no new pics with these `credit`s to clean, but would we ever re-ingest, we will suffer. This cleaner could be reinstated, if it would operate on just these tokens, instead of whole `credit`.

Ideally, we would not be cleaning in [several](https://github.com/guardian/grid/blob/a5ec2de3427f5fbf5f4e73f1e8b968f445e76a16/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala#L100) [places](https://github.com/guardian/grid/tree/master/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup) and matching half-blind.

- [x] tested on TEST

Opinions welcomed, @guardian/digital-cms. Thanks!

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
@@ -69,7 +69,7 @@ object ActionImagesParser extends ImageProcessor {
 }
 
 object AlamyParser extends ImageProcessor {
-  def apply(image: Image): Image = image.metadata.credit match {
+  def apply(image: Image): Image = (image.metadata.credit, image.metadata.source) match {
     case Some("Alamy") | Some("Alamy Stock Photo") => image.copy(
       usageRights = Agencies.get("alamy")
     )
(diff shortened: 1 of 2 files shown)


paperboyo:
> Happy to pair.

Great, lemme know!

PR #2610, opened 2019-07-29
Extract hardcoded links in HTML to API

felixcs1:
## What does this change?
Makes some guardian specific hardcoded links in HTML now come from media-api.

Currently some guardian specific links to google docs and forms are hardcoded into HTML, namely: 
- 'Having trouble picking usage rights?' link
- The feedback form from the menu option in the top right 
- The 'email grid team' in the invalid session error message
- The invalid session help doc

We would like these to be configurable for use in the BBC. This is a non breaking change which hard codes the links into media-api, which are then requested by the front end and placed into the HTML. 

The next stage would be to make the links come into media-api from a config file, which would be a breaking change so thought it best to make this PR first before progressing. 

The links listed above are still making their way to the ui.

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/kahuna/public/js/common/user-actions.html
+++ b/kahuna/public/js/common/user-actions.html
@@ -6,7 +6,7 @@
     </button>
     <ul class="drop-menu__items drop-menu__items--right" ng:if="showUserActions">
         <li><a href="/quotas" target="_blank">Quotas</a></li>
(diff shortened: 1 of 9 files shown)


akash1810:
Closing as superseded by https://github.com/guardian/grid/pull/2610.

PR #2611, opened 2019-08-05
extract hard coded guardian specific links into config (kahuna props)

jamie-pitts:
## What does this change?

This allows for the hardcoded links (e.g. support/feedback form) to be set in the kahuna properties file (/etc/gu/kahuna.props), so that it can be easily changed and differ depending on who the grid is being used by.

By default if the properties are not set then it will not cause any problems/exceptions, just the links will be empty and will just link to themselves (no major problems to the user).

The following needs to set in `/etc/gu/kahuna.properties` if you wish for the links to work:
```
links.feedbackForm=
links.usageRightsHelp=
links.invalidSessionHelp=
links.supportEmail=
```

e.g. for the Guardian:
```
links.feedbackForm=https://goo.gl/forms/OO7CrMt9BRobi1j63
links.usageRightsHelp=https://docs.google.com/a/guardian.co.uk/document/d/1owx57raHRncnSU-voyg1b21yZBcrkM1ttka7CDQLYOs/pub
links.invalidSessionHelp=https://docs.google.com/a/guardian.co.uk/document/d/1qlQhHM2NX3xWBIREB2O8D7S4ZtN0sYb7IDV2leE4Ubc/edit?usp=sharing
links.supportEmail=mailto:thegrid.dev@theguardian.com
```

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/kahuna/app/views/main.scala.html
+++ b/kahuna/app/views/main.scala.html
@@ -3,7 +3,11 @@
   reauthUri: String,
   sentryDsn: Option[String],
   sessionId: String,
-  googleTrackingId: Option[String])
+  googleTrackingId: Option[String],
+  feedbackForm: Option[String],
+  usageRightsHelp: Option[String],
+  invalidSessionHelp: Option[String],
+  supportEmail: Option[String])
 <!DOCTYPE html>
 <html>
   <head>
@@ -33,7 +37,11 @@
     <script>
         this._clientConfig = {
           sessionId: "@Html(sessionId)",
-          googleTrackingId: "@Html(googleTrackingId.getOrElse(""))"
+          googleTrackingId: "@Html(googleTrackingId.getOrElse(""))",
+          feedbackForm: "@Html(feedbackForm.getOrElse(""))",
+          usageRightsHelp: "@Html(usageRightsHelp.getOrElse(""))",
+          invalidSessionHelp: "@Html(invalidSessionHelp.getOrElse(""))",
+          supportEmail: "@Html(supportEmail.getOrElse(""))"
         }
     </script>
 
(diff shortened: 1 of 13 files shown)


kahuna/app/views/main.scala.html
```diff
@@ -33,7 +37,11 @@
     <script>
         this._clientConfig = {
           sessionId: "@Html(sessionId)",
-          googleTrackingId: "@Html(googleTrackingId.getOrElse(""))"
+          googleTrackingId: "@Html(googleTrackingId.getOrElse(""))",
+          feedbackForm: "@Html(feedbackForm.getOrElse(""))",
```
akash1810: minor: should these new config keys have a `Link` suffix to match how they are used clientside?

jamie-pitts: Good point, i've changed this now

kahuna/public/js/errors/global.html
```diff
@@ -16,7 +16,7 @@
 
         <div>
             We are looking into this but if the problem persists please
-            <a href="mailto:thegrid.dev@theguardian.com"
+            <a href="{{ctrl.supportEmailLink}}"
```
akash1810: Would something like this make config simpler? That is, remove `mailto` from the config. Should also mean this property can be used elsewhere if necessary.

(I'm guessing at the syntax here, as don't have an IDE to help 😄 )

```suggestion
            <a href="{{`mailto:${ctrl.supportEmail}`}}"
```

akash1810: oh, we do this in a few places... feel free to leave as is and we can change later if it becomes an issue.

PR #2615, opened 2019-08-16
Use GitHub Actions for CI

akash1810:
## What does this change?
Add GitHub Action to build on push. This is so external contributions can be built in a transparent way (our CI service is IP locked). This does everything TeamCity does except upload a deployable to RiffRaff.

There are two Jobs that run in parallel:
- CI for JavaScript apps (Kahuna, S3Watcher, Reaper)
- CI for scala apps

Notes:
- Launches elasticsearch outside the integration tests (making use of #2571)
- Defines custom Dockerfile for scala build with imaging libraries needed for tests and SBT

Observations on GitHub Actions:
- Dependencies are not cached between builds, resulting in a build time of roughly 15 minutes
- Pushes trigger really quickly
- Force pushes can be a little slow to trigger (about 30 seconds)

External contributions have CI performed on them in a transparent way.

Just see the [Checks](https://github.com/guardian/grid/pull/2615/checks).

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,8 +32,6 @@ jobs:
     services:
       elasticsearch:
         image: elasticsearch:6.5.4
-        ports:
-        - 9200:9200
         # Wait for elasticsearch to report healthy before continuing.
         # see https://github.com/actions/example-services/blob/master/.github/workflows/postgres-service.yml#L28
         options: -e "discovery.type=single-node" --expose 9200 --health-cmd "curl localhost:9200/_cluster/health" --health-interval 10s --health-timeout 5s --health-retries 10
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,48 @@
+name: Grid CI
+on: [push]
+jobs:
+  JSBuild:
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        node-version: [8.6.0]
+    steps:
+      - uses: actions/checkout@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,44 @@
...
+        working-directory: ./reaper
+        run: |
+          npm install
+          npm run riffraff-artefact
+  ScalaBuild:
+    runs-on: ubuntu-18.04
+    services:
+      elasticsearch:
+        image: elasticsearch:6.5.4
+        options: -e "discovery.type=single-node" --expose 9200 --health-cmd "curl localhost:9200/_cluster/health" --health-interval 10s --health-timeout 5s --health-retries 10
```
akash1810: Wait for elasticsearch to report healthy before continuing. Inspired by [this](https://github.com/actions/example-services/blob/master/.github/workflows/postgres-service.yml#L28) as it's not noted in the [docs](https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idservices).

akash1810:
Regarding the failing tests, I think its highlighting a gap in our date logic as the UTC date in these tests are incorrect and are passing their test - https://github.com/guardian/grid/blob/3fdb1d8d4746da09af622c4c2ab9103206fc65fd/common-lib/src/test/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverterTest.scala#L83-L99

akash1810:
Checks are green!

img

Merging.

PR #2617, opened 2019-08-19
Revert "Correctly identify Bloomberg uploads"

paperboyo:
Reverts guardian/grid#2515. Bloomberg isn’t to be it’s own thing any more. Requires:
- [x] removing rights from [these](https://media.gutools.co.uk/search?query=supplier:Bloomberg&nonFree=true) (now they are [here](https://media.gutools.co.uk/search?query=%23usedToBeBloombergNotViaGetty&nonFree=true))
- [x] removing Bloomberg (via Getty) from [a list](https://github.com/guardian/grid/blob/revert-2515-mk-add-bloomberg/common-lib/src/main/scala/com/gu/mediaservice/lib/config/UsageRightsConfig.scala#L16) of Paid Getty collections

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
@@ -17,7 +17,6 @@ object SupplierProcessors {
     AllStarParser,
     ApParser,
     BarcroftParser,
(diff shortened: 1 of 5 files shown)


paperboyo:
> Do we need to update any existing images?

I just added two more since I did 93 yesterday. I’m a bit scared one will come in just before deploy and Grid will die. Will deploy to TEST first (removed images there too).

PR #2621, opened 2019-08-21
Api request cloudwatch metrics [MERGED AS #2956]

jamie-pitts:
## What does this change?

This sends CloudWatch metrics for each received request per service (ignoring healthcheck requests). The idea being that it will allow for us to see how many requests per min each service is receiving, if there are any failed requests, and the average length of time each request is taking. The way the cloudwatch metrics are sent has also been slightly modified so that they get aggregated first into a statistic set, this is a lot more useful for count metrics when deployed on multiple instances as it then allows for you to see the average over the amount of instances (e.g. 3 instances each get 100 requests each, you will see an average of 100).

A configurable property has been added `metrics.request.enabled=false` (defaults to false if not present) which can be toggled on a service by service basis if metrics are only desired for a particular service.

Not 100% sure if this is useful for you guys, but thought I'd create a PR none the less!

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
@@ -51,6 +51,11 @@ trait CommonConfig {
 
   lazy val thrallKinesisStream = properties("thrall.kinesis.stream.name")
 
+  lazy val requestMetricsEnabled: Boolean = properties.getOrElse("metrics.request.enabled", "false").toLowerCase match {
+    case "true" => true
+    case _ => false
+  }
+
   // Note: had to make these lazy to avoid init order problems ;_;
   lazy val domainRoot: String = properties("domain.root")
   lazy val rootAppName: String = properties.getOrElse("app.name.root", "media")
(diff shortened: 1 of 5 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
```diff
@@ -50,6 +50,11 @@ trait CommonConfig {
 
   lazy val thrallKinesisStream = properties("thrall.kinesis.stream.name")
 
+  lazy val requestMetricsEnabled: Boolean = properties.getOrElse("metrics.request.enabled", "false").toLowerCase match {
```
akash1810: we can remove the `match` here in favour of

```suggestion
  lazy val requestMetricsEnabled: Boolean = properties.getOrElse("metrics.request.enabled", "false").toLowerCase == "true"
```

common-lib/src/main/scala/com/gu/mediaservice/lib/metrics/CloudWatchMetrics.scala
```diff
@@ -29,7 +28,7 @@ abstract class CloudWatchMetrics(namespace: String, config: CommonConfig) {
   /** The maximum number of data points to report in one batch.
     * (Each batch costs 1 HTTP request to CloudWatch)
     */
-  val maxChunkSize: Int = 20
+  val maxChunkSize: Int = Int.MaxValue
```
akash1810: what's the reason for this change?

mbarton:
Hey @jamie-pitts, sorry this has got stuck open for so long. We will rebase it and test in our environment and hopefully merge soon

AWare:
Merged as https://github.com/guardian/grid/pull/2956

PR #2625, opened 2019-08-29
DateTime adjustments

akash1810:
## What does this change?
Move to storing computed `fileMetadata` date fields in the `ISODateTimeFormat` as they are easier to parse. From `Sun Aug 25 07:13:27.000 UTC 2019` to `2019-08-25T07:13:27.000Z`.

Here is a sample image on [PROD](https://api.media.gutools.co.uk/images/c35f33340c78a75b09c33c46e68c4c0af211449f/fileMetadata) and [TEST](https://api.media.test.dev-gutools.co.uk/images/c35f33340c78a75b09c33c46e68c4c0af211449f/fileMetadata).

Update the date parsing logic - `getParser` does not obey the timezone, that is `BST` was getting parsed into the local timezone. This is incorrect and means we're now attempting the parsing ourselves.

Finally, `BST` maps to [multiple zones](https://24timezones.com/time-zone/bst):
- British Summer Time
- Bangladesh Standard Time
- Bougainville Standard Time

We're hardcoding to be British Summer Time with `.withZone(DateTimeZone.forOffsetHours(1))`.

Tests can be run from any timezone.

Related https://github.com/guardian/grid/pull/2615

n/a

## Who should look at this?

@guardian/digital-cms

## Tested?
- [x] locally
- [x] on TEST

### DIFF
--- a/common-lib/src/test/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverterTest.scala
+++ b/common-lib/src/test/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverterTest.scala
@@ -83,19 +83,19 @@ class ImageMetadataConverterTest extends FunSpec with Matchers {
   it("should populate the dateTaken field of ImageMetadata from EXIF Date/Time Original Composite (Mon Jun 18 01:23:45 BST 2018)") {
     val fileMetadata = FileMetadata(iptc = Map(), exif = Map(), exifSub = Map("Date/Time Original Composite" -> "Mon Jun 18 01:23:45 BST 2018"), xmp = Map())
     val imageMetadata = ImageMetadataConverter.fromFileMetadata(fileMetadata)
-    imageMetadata.dateTaken should be (Some(parseDate("2018-06-18T01:23:45")))
+    imageMetadata.dateTaken should be (Some(parseDate("2018-06-18T00:23:45Z")))
   }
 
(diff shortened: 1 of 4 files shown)


common-lib/src/test/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverterTest.scala
```diff
@@ -83,19 +83,19 @@ class ImageMetadataConverterTest extends FunSpec with Matchers {
   it("should populate the dateTaken field of ImageMetadata from EXIF Date/Time Original Composite (Mon Jun 18 01:23:45 BST 2018)") {
     val fileMetadata = FileMetadata(iptc = Map(), exif = Map(), exifSub = Map("Date/Time Original Composite" -> "Mon Jun 18 01:23:45 BST 2018"), xmp = Map())
     val imageMetadata = ImageMetadataConverter.fromFileMetadata(fileMetadata)
-    imageMetadata.dateTaken should be (Some(parseDate("2018-06-18T01:23:45")))
+    imageMetadata.dateTaken should be (Some(parseDate("2018-06-18T00:23:45Z")))
```
akash1810: This test and the two below were always incorrect because the `BST` timezone difference wasn't being accounted for as the UTC version is in fact an hour previous.

ajwl:
Interesting! Someone with more knowledge of Grid should probably approve but switch to the Z format makes sense

PR #2627, opened 2019-08-29
Fail fast when image not found.

akash1810:
## What does this change?
Previously, we were always trying to [parse the response from elasticsearch into an Image and logging at error level when it failed.](https://github.com/guardian/grid/blob/2be91f98f143dd9bddeab594506efe65c2f9a1dd/media-api/app/lib/elasticsearch/impls/elasticsearch6/ElasticSearch.scala#L230-L237)

This meant that 404s produce [noisy logs](https://logs.gutools.co.uk/s/editorial-tools/goto/fda0f179f818fc1555700e07c1a2f149) relating to json parsing failure which isn't helpful.

Now, we only try to parse the response to an Image if the request succedded (that is, we recieved a hit from elasticsearch).

Less noise in the logs.

What we're trying to solve:
image

## Tested?
- [x] locally
- [ ] on TEST


ajwl:
👍 makes sense

PR #2626, opened 2019-08-29
read all xmp directories

akash1810:
## What does this change?
Rather than just `getFirstDirectoryOfType`, read all xmp directories using `getDirectoriesOfType` and iterating over the result.

## How can success be measured?
More metadata extraction.

## Screenshots (if applicable)
n/a

## Who should look at this?

@guardian/digital-cms

## Tested?
- [x] locally
- [ ] on TEST

PR #2631, opened 2019-09-03
log xmp metadata extracted from image

akash1810:
## What does this change?
In an effort to diagnose https://github.com/guardian/grid/pull/2626, add logging around what XMP metadata has been read from an image.

Best reviewed [w/out whitespace](https://github.com/guardian/grid/pull/2631/files?w=1)

Logs. Beautiful logs!

n/a

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/image-loader/app/lib/imaging/FileMetadataReader.scala
+++ b/image-loader/app/lib/imaging/FileMetadataReader.scala
@@ -14,8 +14,10 @@ import com.drew.metadata.{Directory, Metadata}
 import com.gu.mediaservice.lib.imaging.im4jwrapper.ImageMagick._
 import com.gu.mediaservice.lib.metadata.ImageMetadataConverter
 import com.gu.mediaservice.model.{Dimensions, FileMetadata}
+import net.logstash.logback.marker.{LogstashMarker, Markers}
 import org.joda.time.{DateTime, DateTimeZone}
 import org.joda.time.format.ISODateTimeFormat
+import play.api.Logger
 
 import scala.collection.JavaConverters._
 import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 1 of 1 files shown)


image-loader/app/lib/imaging/FileMetadataReader.scala
```diff
@@ -91,29 +93,40 @@ object FileMetadataReader {
       }
     } getOrElse Map()
 
+  private def toLogstashFriendlyMarkers(data: Map[String, String]): LogstashMarker = {
+    val cleaned = data.collect {
+      case (key, value) => key.replaceAll("[\\[\\]:/]", "") -> value
```
akash1810: It looks like logstash can't parse `[`, `]`, `:` and `/` characters into marker keys 😢 so just remove them... its better than removing the marker entirely.

paperboyo: Can we replace them with something(s)? No probs if not, would just help searching for eg. parent tags…

jonathonherbert: just double checking as I'd like to use this pattern -- is there a reason this method is colocated with this class? I'm thinking of where to locate it in other projects, possibly in some logging utils object

akash1810: > Can we replace them with something(s)? No probs if not, would just help searching for eg. parent tags…

This is only temporary to help understand the xmp extraction issue.

>  is there a reason this method is colocated with this class?

This is only really needed in this one place at the moment and didn't want to optimise too early. Pushing it into a logging utils object would be best.

mkuzdowicz: Yeah some logging utils object will make sense for me as well, it will be easier to add tests to this utility function later if it will be needed

PR #2634, opened 2019-09-05
Always read xmp metadata in the Getty schema into the GettyImagesGIFT namespace

akash1810:
## What does this change?
An `XMPMetaFactory` has a [cache of previously seen schemas and their prefix](https://github.com/drewnoakes/adobe-xmp-core/blob/master/com/adobe/internal/xmp/XMPMetaFactory.java#L40). This means if there are two xmp blocks using the same schema, but a different prefix, the [XMPMeta from the second xmp block](https://github.com/drewnoakes/metadata-extractor/blob/master/Source/com/drew/metadata/xmp/XmpReader.java#L170) will use the prefix from the first block as it doesn't new up the XMPMetaFactory.

That is, let:
  - Image A namespace Getty schema (http://xmp.gettyimages.com/gift/1.0/) with `prefix0`
  - Image B namespace Getty schema (http://xmp.gettyimages.com/gift/1.0/) with `GettyImagesGIFT`

If we process Image A first, the `XMPSchemaRegistry` will say the Getty namespace is prefixed with `prefix0`.

When we process Image B, we'll see it uses the Getty namespace with the `GettyImagesGIFT` prefix, but that the namespace is in the `XMPSchemaRegistry` from Image A, so we'll set the prefix to `prefix0`.

Conversely, if we process Image B first, the `XMPSchemaRegistry` cache will be in the desired state and Image A will be ingested correctly. That is, it's pot luck!

As a workaround, register the Getty prefix as `GettyImagesGIFT` early to force Getty metadata to use the correct prefix. [This is what ExifTool does.](https://github.com/exiftool/exiftool/blob/3339862c31076f9db30270b3965ac1c49ee0687a/lib/Image/ExifTool/XMP.pm#L184)

This PR also [adds failing tests in first commit](https://github.com/guardian/grid/pull/2634/checks?sha=c9e6143d3679e5e741f7b307f80bd6282c0fa610#step:4:4784) to demonstrate the issue. These tests fail because the contents of the `XMPSchemaRegistry` depends on the order in which files are read. The data in the expected objects are what is in the jpeg files, but the xmp reader doesn't obey it.

- Always reading XMP metadata on the Getty namespace to a `GettyImagesGIFT` namespace, regardless of the prefix used in the file.
- Identifying all Getty images as Getty (as we use the presence of the `GettyImagesGIFT` prefix as an identifier of a Getty image).
- Makes https://github.com/guardian/grid/pull/2626 more predicable.

n/a

## Who should look at this?

@guardian/digital-cms

## Tested?
- [ ] locally
- [ ] on TEST


paperboyo:
💎 💎 💎

mchv:
Great work @akash1810 🎖

akash1810:
> Great work @akash1810 🎖

Not all my work, relied heavily on @sihil 's excellent pairing!

PR #2644, opened 2019-09-20
Allow Kinesis messages to be compressed, but don't...

akash1810:
## What does this change?
Since https://github.com/guardian/grid/pull/2634 we're extracting more metadata from images. However, there is a limit to the size of a Kinesis record and sometimes we're exceeding that limit. Add the ability to write compressed to Kinesis.

This is part 1 and 2 of a migration:
- **Continue writing to Kinesis uncompressed**
- **Thrall reads off Kinesis in both formats**
- Write to Kinesis compressed
- Don't allow uncompressed bytes to be written to Kinesis
- Thrall reads off Kinesis only in compressed format

In order to tell if a message read off Kinesis is compressed, we add a marking byte to the head of the byte array - if the marker is there, we decompress and continue to serialise into `T` and then process the message, if marker is absent, we continue to serialise into `T` and then process the message.

no-op

**Current behaviour** 😢 
image

**Once migration complete** 😂 
image

## Who should look at this?

@guardian/digital-cms

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/json/JsonByteArrayUtil.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/json/JsonByteArrayUtil.scala
@@ -1,0 +1,55 @@
+package com.gu.mediaservice.lib.json
+
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream}
+import java.nio.charset.StandardCharsets
+import java.util.zip.{GZIPInputStream, GZIPOutputStream}
+
+import play.api.libs.json._
+
+import scala.io.Source.fromInputStream
+
+object JsonByteArrayUtil extends PlayJsonHelpers {
+  private val compressionMarkerByte: Byte = 0x00.toByte
+
+  private def compress(bytes: Array[Byte]): Array[Byte] = {
+    val outputStream = new ByteArrayOutputStream()
+    val zipOutputStream = new GZIPOutputStream(outputStream)
+    zipOutputStream.write(bytes)
+    zipOutputStream.close()
+    outputStream.close()
+    val compressedBytes = outputStream.toByteArray
+    compressionMarkerByte +: compressedBytes
+  }
+
+  private def decompress(bytes: Array[Byte]): Array[Byte] = {
+    val bytesWithoutCompressionMarker = bytes.tail
+    val byteStream = new ByteArrayInputStream(bytesWithoutCompressionMarker)
+    val inputStream = new GZIPInputStream(byteStream)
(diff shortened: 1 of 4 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/json/JsonByteArrayUtil.scala
```diff
@@ -0,0 +1,55 @@
...
+  def hasCompressionMarker(bytes: Array[Byte]) = bytes.head == compressionMarkerByte
+
+  def toByteArray[T](obj: T, withCompression: Boolean = false)(implicit writes: Writes[T]): Array[Byte] = {
+    val bytes = Json.toBytes(Json.toJson(obj))
+    if (withCompression) compress(bytes) else bytes
+  }
+
+  def fromByteArray[T](bytes: Array[Byte])(implicit reads: Reads[T]): Option[T] = {
+    val string = new String(
+      if (hasCompressionMarker(bytes)) decompress(bytes) else bytes,
```
jonathonherbert: 👍

common-lib/src/main/scala/com/gu/mediaservice/lib/json/JsonByteArrayUtil.scala
```diff
@@ -0,0 +1,55 @@
...
+
+import play.api.libs.json._
+
+import scala.io.Source.fromInputStream
+
+object JsonByteArrayUtil extends PlayJsonHelpers {
+  private val compressionMarkerByte: Byte = 0x00.toByte
+
+  private def compress(bytes: Array[Byte]): Array[Byte] = {
+    val outputStream = new ByteArrayOutputStream()
```
jonathonherbert: I feel like I _should_ be a bit intimidated by this adding an extra byte stuff, but ... this is really clear 👍

akash1810: Heavily inspired by [Tag Manager](https://github.com/guardian/tagmanager/blob/4d78a1e5ab3c1588ea89116639cb39bc19443e0d/app/services/ThriftSerializer.scala) regarding using the first byte to indicate what kind of thing we have.

common-lib/src/main/scala/com/gu/mediaservice/lib/json/JsonByteArrayUtil.scala
akash1810: Worth noting this gets simplified in the other PRs https://github.com/guardian/grid/pull/2645 and https://github.com/guardian/grid/pull/2646.

PR #2648, opened 2019-09-25
Revert "Always read off Kinesis compressed."

akash1810:
Reverts guardian/grid#2646

A few images have got a lot of xmp metadata, which is being indexed into individual fields. This is putting strain on elasticsearch.

Reverting will mean we drop these messages before they even get to thrall, so elasticsearch is happier. The impact is some images will not upload.


blishen:
Looks good!

PR #2652, opened 2019-09-27
make iframe ancestors and cors allowed origins stage unaware

akash1810:
## What does this change?
Pushing these values to config means we don't leak domains in code and also provides more flexibility.

Now, rather than hardcoding a list of subdomains, read from config a list of fully qualified domains that are allowed to make cors requests. Additionally, pushing iframe ancestors to config means we don't have to allow *.domainroot if we don't want to.

n/a

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/config/Services.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/config/Services.scala
@@ -5,7 +5,7 @@ case class ServiceHosts(kahunaPrefix: String, apiPrefix: String, loaderPrefix: S
                         usagePrefix: String, collectionsPrefix: String, leasesPrefix: String,
                         authPrefix: String)
 
-class Services(val domainRoot: String, isProd: Boolean, hosts: ServiceHosts) {
+class Services(val domainRoot: String, isProd: Boolean, hosts: ServiceHosts, corsAllowedOrigins: Set[String]) {
   val kahunaHost: String      = s"${hosts.kahunaPrefix}$domainRoot"
   val apiHost: String         = s"${hosts.apiPrefix}$domainRoot"
   val loaderHost: String      = s"${hosts.loaderPrefix}$domainRoot"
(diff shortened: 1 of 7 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/config/Services.scala
```diff
@@ -30,28 +30,9 @@ class Services(val domainRoot: String, isProd: Boolean, hosts: ServiceHosts) {
...
-      domainRoot.replace("test", "code")
-    )
-  }
-
-  // TODO move to config
-  val corsAllowedTools: Set[String] = toolsDomains.foldLeft(Set[String]()) {(acc, domain) => {
-    val tools = Set(
-      baseUri(s"composer.$domain"),
-      baseUri(s"fronts.$domain")
-    )
```
akash1810: I'm not entirely convinced Composer and Fronts need CORS access to Grid either as they proxy though the server IIRC.

kahuna/app/KahunaComponents.scala
```diff
@@ -35,7 +35,7 @@ object KahunaSecurityConfig {
     )
 
     val frameSources = s"frame-src ${config.services.authBaseUri} ${config.services.kahunaBaseUri} https://accounts.google.com"
-    val frameAncestors = s"frame-ancestors ${config.services.toolsDomains.map(domain => s"*.$domain").mkString(" ")}"
+    val frameAncestors = s"frame-ancestors ${config.services.frameAncestors.mkString(" ")}"
```
jonathonherbert: Just for my own understanding -- is the renaming here just a matter of clarifying what this config param is used for?

akash1810: Yes, you're correct;`toolsDomains` is very Guardian-y.

common-lib/src/main/scala/com/gu/mediaservice/lib/config/Services.scala
akash1810: Update: Fronts and Composer allow you to paste in a URL to a Grid crop and it'll make a request to media-api, so CORS access is needed for these tools.

PR #2656, opened 2019-10-04
Feature/kahuna download from api

jamie-pitts:
## What does this change?
The download (+ bulk download) button in the client will now download images via the api download link. This allows for metrics to be gathered about which images have been downloaded and who downloaded which images.

Additionally a new endpoint has been added to the api which allows for optimised images to be downloaded in a similar way to the original images (this also redirects you to imgops).

As the download link now calls the API, the users auth cookie needs to be present, and so the `withCredentials` flag was added. This caused a slight problem with nginx as it had a wildcard for the origin, so this was fixed by adding a couple of headers you will see in the PR - I don't think this should have any adverse effects but not an expert on this!

### DIFF
--- a/imgops/dev/nginx.conf.template
+++ b/imgops/dev/nginx.conf.template
@@ -22,6 +22,10 @@ http {
     }
 
     location / {
+      # Set proper CORS headers
+      more_set_headers 'Access-Control-Allow-Origin: $http_origin';
+      more_set_headers 'Access-Control-Allow-Credentials: true';
+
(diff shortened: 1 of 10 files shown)


mbarton:
Thanks for the contribution @jamie-pitts! We've tested it in our environment and it's good to 🚢

jamie-pitts:
> Thanks for the contribution @jamie-pitts! We've tested it in our environment and it's good to 🚢

Awesome, glad to hear! 🎉

PR #2663, opened 2019-10-21
Test-revert "store the original file name alongside the original image in aws as metadata"

paperboyo:
Allows me to test reverting guardian/grid#2660 to see if it could have affected uploads of files with filenames with diacritics (because I can’t read code :-S).

Do not merge!

### DIFF
--- a/image-loader/app/model/ImageUpload.scala
+++ b/image-loader/app/model/ImageUpload.scala
@@ -194,24 +194,15 @@ class ImageUploadOps(store: ImageLoaderStore, config: ImageLoaderConfig, imageOp
     })
   }
 
-  def storeSource(uploadRequest: UploadRequest) = {
-    val baseMeta = Map(
+  def storeSource(uploadRequest: UploadRequest) = store.storeOriginal(
+    uploadRequest.imageId,
+    uploadRequest.tempFile,
+    uploadRequest.mimeType,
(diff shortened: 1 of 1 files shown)


paperboyo:
Reverting seems to work (currently on TEST), so it looks like this might broke >ASCII characters in filename (cc @jamie-pitts):

image
[Revert left; current PROD right]

I think we should revert, what do you think, @guardian/digital-cms?

akash1810:
From https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#UserMetadata AWS have the following limitation on custom S3 object metadata:

> Each key-value pair must conform to US-ASCII

I'd vote to [remove the non-ascii characters](https://stackoverflow.com/questions/8519669/replace-non-ascii-character-from-string) from the filename rather than revert this PR.

mbarton:
Maybe URL encode the filename?

jamie-pitts:
Crap sorry about this, will look at adding a fix

jamie-pitts:
Have created a PR here with the fix suggested by @mbarton: https://github.com/guardian/grid/pull/2665

paperboyo:
Closing as https://github.com/guardian/grid/pull/2665 is the proper fix. Thanks so much, @jamie-pitts!

PR #2670, opened 2019-10-29
Feature/download to usages

jamie-pitts:
This probably isn't wanted but thought I'd create a PR just for completeness sake. Requires both https://github.com/guardian/grid/pull/2664 and https://github.com/guardian/grid/pull/2656 to be merged first (ES script fix, and downloading from API).

## What does this change?

Usages has a new 'downloads' type which records who has downloaded what and when. This is in a similar way to how the other usages are set up, and takes an API call to mark an image as downloaded (along with who downloaded it and a time stamp). 

Media API has been updated so when downloads occur it will send a async request over to usages to mark it as download. This is a best effort call and no error handling is in place, as I don't think we should be too bothered if this call fails or not.

Screenshot 2019-10-29 at 12 03 22

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch6/ElasticSearchClient.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch6/ElasticSearchClient.scala
@@ -129,6 +129,17 @@ trait ElasticSearchClient {
     Logger.info("Got alias action response: " + aliasActionResponse)
   }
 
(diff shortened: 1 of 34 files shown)


jonathonherbert:
@akash1810 given #2977 can this now be closed?

akash1810:
This branch needs rebasing before merging. This has been done in https://github.com/guardian/grid/pull/2977 as the original branch for this PR comes from outside the organisation and is write protected.

https://github.com/guardian/grid/pull/2977 has now been shipped so, closing this PR as a result.

PR #2674, opened 2019-11-01
add some client side logging for collection issue

akash1810:
## What does this change?
There's a scenario where a node in the collection tree doesn't have a pathId; this causes nodes beyond this one to not render.

Whilst we do not understand how we enter this state, add some logging to help identify the tainted node so we can correct it manually.

We can manually fix things a bit faster.

## Tested?
- [ ] locally
- [ ] on TEST


paperboyo:
👍 [not a sign of code-review]

PR #2675, opened 2019-11-01
Using local kinesis in dev mode

mkuzdowicz:
## What does this change?
Give possibility to use local kinesis in dev mode, such that more then one person can work on GRID at the same time

run grid locally and adding images will work on multiple machines at the same office

## Tested?
- [x] locally
- [x] on TEST

### DIFF
--- a/dev-start.sh
+++ b/dev-start.sh
@@ -112,7 +112,7 @@ setupLocalKinesis() {
   # dummy env variables for local kinesis
   export AWS_ACCESS_KEY_ID=foobar
   export AWS_SECRET_ACCESS_KEY=foobar
-  export AWS_REGION=eu-west-2
+  export AWS_REGION=eu-west-1
   echo 'creating local kinesis streams'
   # ignore stream already exists error
   set +e
--- a/dev-start.sh
+++ b/dev-start.sh
@@ -109,16 +109,12 @@ setupLocalKinesis() {
   # java sdk use CBOR protocol
   # which does not work with localstack kinesis which use kinesislite
   export AWS_CBOR_DISABLE=true
-  # dummy env variables for local kinesis
-  export AWS_ACCESS_KEY_ID=foobar
-  export AWS_SECRET_ACCESS_KEY=foobar
-  export AWS_REGION=eu-west-1
   echo 'creating local kinesis streams'
   # ignore stream already exists error
   set +e
   stream_name='media-service-DEV-ThrallMessageQueue-1N0T2UXYNUIC9'
-  aws --endpoint-url=http://localhost:4568 kinesis create-stream --shard-count 1 --stream-name "${stream_name}"
-  aws --endpoint-url=http://localhost:4568 kinesis --region=eu-west-2 list-streams
+  aws --profile media-service --endpoint-url=http://localhost:4568 kinesis create-stream --shard-count 1 --stream-name "${stream_name}"
+  aws --profile media-service --endpoint-url=http://localhost:4568 kinesis --region=eu-west-2 list-streams
 }
 
 main() {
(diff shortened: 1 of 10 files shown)


docs/local-kinesis.md
```diff
@@ -0,0 +1,19 @@
...
+
+command you will run app with local mode, it will use local instance of kinesis provided by [localstack](https://github.com/localstack/localstack) on http://localhost:4568
+
+it is setup in `docker-compose.yml` file
+
+if you want to cleanup your local docker environment run 
+
+     docker-compose down -v --remove-orphans
+
+if you want to do a lockup what local kinesis contains run the following command:
```
jonathonherbert:
```suggestion
if you want to do a lookup what local kinesis contains run the following command:
```

dev-start.sh
```diff
@@ -104,12 +104,30 @@ changeNodeVersion() {
   fi
 }
 
+setupLocalKinesis() {
+  sleep 3s
+  # java sdk use CBOR protocol
+  # which does not wokr wih localstack kinesis which use kinesislite
```
jonathonherbert:
```suggestion
  # which does not work with localstack kinesis -- localstack uses kinesislite
```

dev-start.sh
```diff
@@ -104,12 +104,30 @@ changeNodeVersion() {
...
+  export AWS_CBOR_DISABLE=true
+  # dummy env variables for local kinesis
+  export AWS_ACCESS_KEY_ID=foobar
+  export AWS_SECRET_ACCESS_KEY=foobar
+  export AWS_REGION=eu-west-2
+  echo 'cretaing local kinesis streams'
```
jonathonherbert:
```suggestion
  echo 'creating local kinesis streams'
```

dev-start.sh
```diff
@@ -104,12 +104,30 @@ changeNodeVersion() {
   fi
 }
 
+setupLocalKinesis() {
+  sleep 3s
+  # java sdk use CBOR protocol
+  # which does not work with localstack kinesis which use kinesislite
+  export AWS_CBOR_DISABLE=true
+  # dummy env variables for local kinesis
+  export AWS_ACCESS_KEY_ID=foobar
```
akash1810: could we use the `--profile` cli flag rather than env vars here?

dev-start.sh
```diff
@@ -104,12 +104,30 @@ changeNodeVersion() {
...
+  export AWS_SECRET_ACCESS_KEY=foobar
+  export AWS_REGION=eu-west-2
```
akash1810: why a different region from the rest of the app (eu-west-1)?

dev-start.sh
```diff
@@ -104,12 +104,30 @@ changeNodeVersion() {
...
+  echo 'creating local kinesis streams'
+  # ignore stream already exists error
+  set +e
+  stream_name='media-service-DEV-ThrallMessageQueue-1N0T2UXYNUIC9'
+  aws --endpoint-url=http://localhost:4568 kinesis create-stream --shard-count 1 --stream-name "${stream_name}"
```
akash1810: does `create-stream` behave gracefully if the stream already exists?

dev-start.sh
mkuzdowicz: it does not matter it is localhost just for AWS cli this variable need to exists with anything, will change to the same for consistency

dev-start.sh
mkuzdowicz: it does not matter :) those variables need to exists with whatever
This is more for AWS cli rather then for localstack on localhost so actual kinesis

dev-start.sh
mkuzdowicz: `set +e` solves that then it will just fail silently (stream already exists...) and move on

dev-start.sh
mkuzdowicz: done

dev-start.sh
mkuzdowicz: done

PR #2681, opened 2019-11-20
leases stuck on updating fix

mkuzdowicz:
## What does this change?

Bug fix for leases stuck on updating 

trello card: https://trello.com/c/FIoSr7L1/157-bug-small-leases-stuck-on-updating-and-never-resolves-until-page-refresh

- kahuna have unit tests setup
- after adding lease frontend do not end-up with updating... but it display added leases

## Tested?
- [x] on TEST
- [x] unit tests

### DIFF
--- a/kahuna/public/js/services/api/leases-helper.js
+++ b/kahuna/public/js/services/api/leases-helper.js
@@ -1,0 +1,33 @@
+import moment from 'moment';
+
+function readLeases(image) {
+  return image.data.leases.data;
+}
+
+export function getApiImageAndApiLeasesIfUpdated(image, apiImage) {
+    const apiLeases = readLeases(apiImage);
+    const leases = readLeases(image);
+    const apiImageAndApiLeases = {image: apiImage, leases: apiLeases};
+    const isNewlyCreated = leases.lastModified === null;
+    if (isNewlyCreated) {
+      const apiImageLeasesAreCreated = apiLeases.lastModified !== null;
+      if (apiImageLeasesAreCreated) {
+        return apiImageAndApiLeases;
+      } else {
+        return undefined;
+      }
+    } else {
+      const apiImageLeasesAreUpdated = function() {
+        const currentLastModified = moment(apiLeases.lastModified);
+        const previousLastModified = moment(leases.lastModified);
+        return currentLastModified.isAfter(previousLastModified);
+      };
+      if (apiImageLeasesAreUpdated()) {
+        return apiImageAndApiLeases;
+      } else {
+        return undefined;
+      }
+    }
+  };
+
+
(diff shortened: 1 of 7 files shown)


kahuna/public/js/services/api/leases-helper.js
```diff
@@ -0,0 +1,33 @@
...
+function readLeases(image) {
+  return image.data.leases.data;
+}
+
+export function getApiImageAndApiLeasesIfUpdated(image, apiImage) {
+    const apiLeases = readLeases(apiImage);
+    const leases = readLeases(image);
+    const apiImageAndApiLeases = {image: apiImage, leases: apiLeases};
+    const isNewlyCreated = leases.lastModified === null;
+    if (isNewlyCreated) {
```
jonathonherbert: No pattern matching/guards in JS :( One day ...

kahuna/public/js/services/api/leases-helper.spec.js
```diff
@@ -0,0 +1,151 @@
...
+            }
+          }
+        }
+      };
+
+      const actual = getApiImageAndApiLeasesIfUpdated(clientSideImage, apiImageWithUpToDateLease);
+
+      const expected = {image: apiImageWithUpToDateLease, leases: apiImageWithUpToDateLease.data.leases.data};
+
+      expect(actual).toEqual(expected);
```
jonathonherbert: Good to see these tests, thanks for setting them up, it's very good for the project 👍

kahuna/public/js/services/api/leases.js
```diff
@@ -112,13 +112,13 @@ leaseService.factory('leaseService', [
       const imagesArray = images.toArray ? images.toArray() : images;
       return $q.all(imagesArray.map(image => {
         return image.get().then(apiImage => {
-          const apiLeases = imageAccessor.readLeases(apiImage);
-          const leases = imageAccessor.readLeases(image);
-          const currentLastModified = moment(apiLeases.lastModified);
-          const previousLastModified = moment(leases.lastModified);
-          if (currentLastModified.isAfter(previousLastModified)) {
-            return { image: apiImage, leases: apiLeases };
+          const apiImageAndApiLeases = getApiImageAndApiLeasesIfUpdated(image, apiImage);
```
jonathonherbert: This is now much clearer behind a function call, excellent 👍

mkuzdowicz: thanks :)

kahuna/public/js/services/api/leases-helper.spec.js
mkuzdowicz: i am happy about that as well :)

kahuna/public/js/services/api/leases-helper.js
mkuzdowicz: would be usefull to move that to TypeScript one day

PR #2688, opened 2019-12-02
Use new node riffraff artifact in reaper

AWare:
## What does this change?

This replaces the legacy node-riffraff-artefact with the new one.

This will need the teamcity build of the reaper to run deploy.

By allowing us to bundle the lambda and not ship our dev-deps to aws.

## Tested?
- [ ] locally
- [ ] on TEST


sihil:
Nice to have, but since reaper isn't even running doesn't gain us anything just now.

PR #2689, opened 2019-12-03
remove elk logger from reaper now we ship logs by default

AWare:
this removes the elk logger from the reaper as this happens automatically

### DIFF
--- a/reaper/index.js
+++ b/reaper/index.js
@@ -1,55 +1,57 @@
-const fetch = require('node-fetch');
-const ELKKinesisLogger = require('@guardian/elk-kinesis-logger');
-const CredentialsConfig = require('./CredentialsConfig');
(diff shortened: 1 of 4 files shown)


reaper/index.js
```diff
@@ -1,55 +1,57 @@
-const fetch = require('node-fetch');
-const ELKKinesisLogger = require('@guardian/elk-kinesis-logger');
-const CredentialsConfig = require('./CredentialsConfig');
+const fetch = require("node-fetch");
+const ELKKinesisLogger = require("@guardian/elk-kinesis-logger");
```
akash1810:
```suggestion

```

akash1810:
Can we add a link to the readme for the cloudwatch log management lambdas that does the log shipping now?

Issue #2395, opened 2019-01-25
TIFF support


paperboyo:
This issue is a (very) loose checklist for TIFF image support. Feel free to add/modify/delete/flush down the drain.

- [ ] new MIME type
- [ ] since our libraries should themselves deal with different flavours of TIFF correctly, we can support most provided we channel [transparent ones](https://github.com/guardian/grid/blob/master/image-loader/app/lib/imaging/FileMetadataReader.scala#L171) through current PNG route and flat ones through current JPEG route
- [ ] [adding](https://github.com/guardian/grid/blob/master/image-loader/app/lib/imaging/FileMetadataReader.scala#L13) [TIFF directory](https://github.com/drewnoakes/metadata-extractor/tree/master/Source/com/drew/metadata/tiff) to metadata
- [ ] collecting [colourModel](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala#L146) information ([bit depth](https://github.com/guardian/grid/blob/master/image-loader/app/lib/imaging/FileMetadataReader.scala#L176), hasAlpha (transparency, really) etc.
- [ ] possibly collecting extra information without explicit control over these elements (just info about them): layers, paths etc
- [ ] possibly creating [prerenders](https://github.com/guardian/grid/blob/master/image-loader/app/model/ImageUpload.scala#L37) so that they can load faster _via_ imgOps

As some of the above operations are currently format-specific, adding TIFF provide a good opportunity to generalise them and bring some order to the galaxy.

mbarton:
For reference, the work done to add PNG support:

https://github.com/guardian/grid/pull/1848
https://github.com/guardian/grid/pull/1912 (if we wanted to optimise the tiffs for consumption)
https://github.com/guardian/grid/pull/1959

paperboyo:
Instead of my monumental #1959, I would add:
https://github.com/guardian/grid/pull/2375 (transparency sniffing leaky, follows https://github.com/guardian/grid/pull/1876)
https://github.com/guardian/grid/pull/1878 (transparency route)
https://github.com/guardian/grid/pull/1866 (no transparency > JPEG route)
https://github.com/guardian/grid/pull/1857 (format-specific metadata)

mbarton:
https://github.com/guardian/grid/pull/2568

paperboyo:
>  since our libraries should themselves deal with different flavours of TIFF correctly

[...]

PR #2690, opened 2019-12-05
Add promise queueing to try and make batch operations faster

AWare:
## What does this change?

This will only allow two batch operations or image uploads to run concurrently. I've used this pattern a few times to fix similar problems so am hopeful it'll help things.

Increased performance when running batch jobs.

## Who should look at this?
@akash1810 @paperboyo @mbarton 
Could also do with a hand spinning up grid locally or on test to test this. 

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/kahuna/public/js/upload/jobs/upload-jobs.js
+++ b/kahuna/public/js/upload/jobs/upload-jobs.js
@@ -1,5 +1,5 @@
 import angular from "angular";
-import { pAll } from "p-all";
+import pAll from "p-all";
 import template from "./upload-jobs.html";
 import "../../preview/image";
 import "../../components/gr-delete-image/gr-delete-image";
--- a/kahuna/public/js/upload/jobs/upload-jobs.js
+++ b/kahuna/public/js/upload/jobs/upload-jobs.js
@@ -1,4 +1,5 @@
 import angular from "angular";
+import { pAll } from "p-all";
 import template from "./upload-jobs.html";
 import "../../preview/image";
 import "../../components/gr-delete-image/gr-delete-image";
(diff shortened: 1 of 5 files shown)


paperboyo:
I’ll let Akash and Micheal have proper opinions, but just a quation from a village idiot: which batch operations does this apply to? Uploads? Metadata editing? Deletions? Only through UI or mediaAPI too?

AWare:
Just in the UI, and afaict, all of the above as I changed the helper that appears used in all of those places. JS is very good at queuing a far larger number of promises concurrently than it's able to deliver on.

paperboyo:
Looking forward to test batch operations on TEST, then 👏 .

kahuna/public/js/upload/jobs/upload-jobs.js
```diff
@@ -1,178 +1,196 @@
...
+          error => {
             const reason = error.body && error.body.errorKey;
 
-            const message = reason === 'unsupported-type' ?
-                'The Grid only supports JPG and PNG images.' +
-                ' Please convert the image and try again.' :
-                error.body && error.body.errorMessage || 'unknown';
+            const message =
+              reason === "unsupported-type"
+                ? "The Grid only supports JPG and PNG images." +
```
mbarton: Unrelated but does this message need changing now TIFF support has been added? cc @paperboyo

paperboyo: Thanks for review, @mbarton! Yeah, also seen it, but TIFF support hasn’t been advertised yet. Reasoning (not too strong, I admit) being outstanding [work in this](https://github.com/guardian/grid/issues/2395).

AWare:
@mbarton is there a solution to the linter noise?

PR #2692, opened 2019-12-09
Feature/delete versioned

jamie-pitts:
## What does this change?
We use a versioned S3 bucket so that we can have CRR enabled. We found that deleting images doesn't actually delete the image (and so save storage space), instead it places a delete marker on top of the image. This PR is too permanently delete the image instead of place the delete marker. It is configurable so that you can say if the bucket is versioned or not (default is not, keep things how they are currently).

This of course isn't the only way of tackling this problem - you can add a lifecycle rule to your bucket to delete old versions and clear delete markers after a certain period of time. But as this has been done with a config param, then that could also be done and this disabled.

Bucket size + number of objects should decrease over time once enabled

## Tested?
- [ x ] locally
- [ x ] on TEST


mbarton:
I've rased https://github.com/guardian/grid/pull/2998 to resolve the conflicts and get this merged

PR #2693, opened 2019-12-09
Add an exponential backoff to the apiPoll function.

AWare:
## What does this change?

While [this](https://github.com/guardian/grid/pull/2690) was probably sensible, testing with @paperboyo didn't show any improvement. 

This PR adds an [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) with jitter to the `apiPoll` function. This increases the delay to 2^(attempt) plus a little random noise. 

Testing this against an upload of 30 images reduced the http request total from ~800 to ~500

Slightly more performant batch jobs.

## Who should look at this?

@akash1810 @mbarton @paperboyo 

## Tested?
- [X] locally
- [x] on TEST

### DIFF
--- a/kahuna/public/js/util/async.js
+++ b/kahuna/public/js/util/async.js
@@ -70,7 +70,6 @@ async.factory('poll',
         const withBackoff = () => {
             const jitter = Math.floor(Math.random() * JITTER);
             const wait = pollEvery * BACKOFF ** (i++) + jitter;
-            console.log(wait);
             return wait;
         };
 
--- a/kahuna/public/js/util/async.js
+++ b/kahuna/public/js/util/async.js
@@ -70,8 +70,8 @@ async.factory('poll',
           let i = 0;
           return () => {
             const jitter = Math.floor(Math.random() * JITTER);
-            const wait = pollEvery * BACKOFF ** (i++) + jitter
-            console.log(wait)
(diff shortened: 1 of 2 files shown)


kahuna/public/js/util/async.js
```diff
@@ -6,94 +6,107 @@ export var async = angular.module('util.async', []);
...
 
         // Returns the result of promise or a rejected timeout
         // promise, whichever happens first
         function withTimeout(promise) {
-            return race([promise, timeout]);
+          return race([promise, timeout]);
         }
 
+
+        const withBackoff = (()=>{
```
akash1810: Does this need to be an IIFE?

kahuna/public/js/util/async.js
```diff
@@ -6,94 +6,107 @@ export var async = angular.module('util.async', []);
...
+          const JITTER = 100;
+          const BACKOFF = 2;
+          let i = 0;
+          return () => {
+            const jitter = Math.floor(Math.random() * JITTER);
+            const wait = pollEvery * BACKOFF ** (i++) + jitter;
+            console.log(wait);
```
akash1810:
```suggestion
```

kahuna/public/js/util/async.js
AWare: can probably bring the closed over vars into parent scope

PR #2695, opened 2019-12-11
Applying nested fields for fileMetadata

mkuzdowicz:
## Tested?
- [ ] locally
- [ ] on TEST


common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch6/Mappings.scala
```diff
@@ -48,7 +52,16 @@ object Mappings {
         assetMapping("optimisedPng"),
         userMetadataMapping("userMetadata"),
         dateField("userMetadataLastModified"),
-        dynamicObj("fileMetadata"),
+        objectField("fileMetadata").fields(
+          NestedField("iptc").dynamic(true),
```
mbarton: @mkuzdowicz I think this would also lead to large numbers of fields in the index and the same stress on ES we are already seeing.

In Giant we model metadata as a nested field of `key->value` pairs, limiting the number of fields to two.

PR #2696, opened 2019-12-17
improve uploader by deleting cached copy of image from browser

AWare:
## What does this change?

When an image is uploaded, a URL to it is held by the browser using window.createObjectURL. This could be to blame for poor performance of the upload. 

This revokes the object url for uploaded images when the upload succeeds, this seems to improve memory use post-upload.

Alleviating the firefox upload perf issues @paperboyo is experiencing. 

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/kahuna/public/js/upload/manager.js
+++ b/kahuna/public/js/upload/manager.js
@@ -10,8 +10,6 @@ upload.factory('uploadManager',
 
     function createJobItem(file) {
         var request = fileUploader.upload(file);
-        // TODO: find out where we can revoke these
-        // see: https://developer.mozilla.org/en-US/docs/Web/API/URL.revokeObjectURL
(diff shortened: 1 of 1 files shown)


kahuna/public/js/upload/manager.js
```diff
@@ -39,7 +37,13 @@ upload.factory('uploadManager',
 
         // once all `jobItems` in a job are complete, remove it
         // TODO: potentially move these to a `completeJobs` `Set`
-        $q.all(promises).finally(() => jobs.delete(job));
+        $q.all(promises).finally(() => {
+          jobs.delete(job);
+          job.map(jobItem => {
+            console.log(jobItem);
```
akash1810:
```suggestion

```

PR #2709, opened 2019-12-20
update media-api healthcheck

akash1810:
## What does this change?
Add a request to elasticsearch in media-api's healthcheck. The media-api is only useful iff it can reach elasticsearch, making it's healthcheck fail will mean the ELB will recover faster.

media-api's healthcheck reports healthy

n/a

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch6/ElasticSearchClient.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch6/ElasticSearchClient.scala
@@ -11,7 +11,7 @@ import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration._
 import scala.concurrent.{Await, Future}
 
-trait ElasticSearchClient {
+trait ElasticSearchClient extends ElasticSearch6Executions {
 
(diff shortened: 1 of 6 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch6/ElasticSearchClient.scala
```diff
@@ -53,6 +53,11 @@ trait ElasticSearchClient {
     }
   }
 
+  def healthCheck(): Future[Boolean] = {
+    val request = search(imagesAlias) limit 0
+    executeAndLog(request, "Health check").map { _ => true}.recover { case _ => false}
+  }
+
```
mkuzdowicz: maybe limit 1 ?

akash1810: Discussed over chat: `limit` here is the maximum to return, as we're primarily interested in knowing if we can query ES (rather than ES's response for an image) `limit 0` should be fine.

This PR is also mainly moving existing code around and doesn't reason about it's quality too much. That is, let's change this in it's own PR if needed.

PR #2713, opened 2020-01-02
Don't blow up when normalising whitespace chars when many are consecutive

jonathonherbert:
## What does this change?

The regex `(\r|\n|\r\n)+` can produce a stack overflow when many whitespace chars are consecutive.

By changing the capturing group to the character group `[\r\n]+` our hypothesis is we avoid having to maintain the match data as the group captures more characters. In testing, this regex can handle strings many orders of magnitude longer than the previous expression.

### Some context

This is an attempt to solve an odd stack overflow error that surfaced in our logs. Because the error occurred during a recursion, we're left with little information about its provenance --

```
java.lang.StackOverflowError: null
	at java.util.regex.Pattern$BmpCharProperty.match(Pattern.java:3812)
	at java.util.regex.Pattern$Branch.match(Pattern.java:4618)
	at java.util.regex.Pattern$GroupHead.match(Pattern.java:4672)
	at java.util.regex.Pattern$Loop.match(Pattern.java:4799)
	at java.util.regex.Pattern$GroupTail.match(Pattern.java:4731)
	at java.util.regex.Pattern$BranchConn.match(Pattern.java:4582)
	at java.util.regex.Pattern$BmpCharProperty.match(Pattern.java:3812)
```

In the absence of more information, we've found a few cases where this problem could occur in our application code, and we're patching them.

Query for an image with a large number of consecutive newlines in the description. The server should no longer stop. (Caveat -- we couldn't find one via search.)

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/media-api/test/lib/ImageResponseTest.scala
+++ b/media-api/test/lib/ImageResponseTest.scala
@@ -18,6 +18,12 @@ class ImageResponseTest extends FunSpec with Matchers {
     normalisedText shouldBe "Here is some text\nthat spans across\nmultiple lines\n"
   }
 
+  it("not cause a stack overflow when many consecutive newline characters are present") {
+    val text = "\n\r\n\n\n\r\r\r\n" * 1000
+    val normalisedText = ImageResponse.normaliseNewLines(text)
+    normalisedText shouldBe "\n"
+  }
+
   it("should not touch \\n linebreaks") {
     val text = "Here is some text\nthat spans across\nmultiple lines\n"
     val normalisedText = ImageResponse.normaliseNewLines(text)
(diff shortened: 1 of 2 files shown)


media-api/test/lib/ImageResponseTest.scala
```diff
@@ -18,6 +18,12 @@ class ImageResponseTest extends FunSpec with Matchers {
     normalisedText shouldBe "Here is some text\nthat spans across\nmultiple lines\n"
   }
 
+  it("not cause a stack overflow when many consecutive newline characters are present") {
+    val text = "\n\r\n\n\n\r\r\r\n" * 1000
+    val normalisedText = ImageResponse.normaliseNewLines(text)
+    normalisedText shouldBe "\n"
```
mkuzdowicz: lest have different names for `normaliseNewLines`

as we have 2 functions now that have the same name

PR #2716, opened 2020-01-06
specialInstructions in free-text search and as a chip

paperboyo:
## What does this change?
Adds `specialInstructions` to free-text search and adds `specialInstructions` chip.

Users can search `specialInstructions`.

## Tested?
- [ ] locally
- [ ] on TEST


paperboyo:
Hmm, so as is, this won’t make free-text search search within `specialInstructions`. Nor will it allow for partial searches within them (only full contents). So: pretty useless. Will need help, probably…

sihil:
@akash1810 does the code deal with adding new fields to an existing mapping?

paperboyo:
Closed in favour of https://github.com/guardian/grid/pull/3036.

PR #2719, opened 2020-01-10
ImageIngestOperations: simplifying uploaded image s3 key creation

mkuzdowicz:
## What does this change?

ImageIngestOperations: simplifying uploaded image s3 key creation


paperboyo:
I think we can close this. WDYT, @akash1810?

PR #2721, opened 2020-01-20
be more graceful in the handling of a gunzip exception

akash1810:
## What does this change?
Gunzip exception is raised when the message isn't compressed. Rather than throwing this exception, catch it, log it and return None for a more graceful code path.

No more `java.util.zip.ZipException: Not in GZIP format` errors in the logs.

n/a

## Who should look at this?

@guardian/digital-cms

## Tested?
- [x] locally
- [ ] on TEST


akash1810:
relates to https://github.com/guardian/rcs-poller-lambda/pull/27

PR #2723, opened 2020-01-22
Endpoint to project image json entry generated by `image-loader` for Elastic-search by giving image id that exists in S3

mkuzdowicz:
## What does this change?
Endpoint to project image json entry generated by `image-loader` for Elastic-search by giving image id that exists in S3

This PR gives endpoint that will show image projection on what only `image-loader` will produce

further rehydration will be added in next PRs

access: 

https://loader.media.local.dev-gutools.co.uk/images/project/:imageId

example image in test https://loader.media.test.dev-gutools.co.uk/images/project/cc8ddd67a7b5dfd3dea3af4bb4d1c4afaa18ec13

and see the image JSON projection that should be similar to Elastic-search entry

## Tested?
- [x] locally
- [x] on TEST


mkuzdowicz:
> This change is taking an id of an image, doing a lookup in S3 and if it exists extracts the metadata and responds in the shape of `Image`?
> 
> Could we update the upload route instead with a `dry-run` feature? That is, an endpoint where we `POST` a file with `dry-run=true` and it'll do a normal upload except for the part of copying the file to S3 or raising a message for thrall and when `dry-run=false` a standard upload is performed.
> 
> The benefit here is not relying on the pre-existence of the image in Grid.

yes that is exactly what the feature is, in terms of how the endpoint looks like i am open

That is one of the first steps as next one would be add stuff from dynamo to the image

sihil:
@akash1810 My understanding that the point of this work was to be able to reindex existing content. Whilst I can kind of see the benefit of having an upload with dry-run I'm not sure what concrete problem that would be solving?

mkuzdowicz:
> @akash1810 My understanding that the point of this work was to be able to reindex existing content. Whilst I can kind of see the benefit of having an upload with dry-run I'm not sure what concrete problem that would be solving?

yes i agree this work is meant for images that exists in s3 it is not for Dry run,
But it does not stop us to add Dry run in the future but it think it is out fo scope

What do you think @akash1810 ?

akash1810:
The dry run was more for testing any changes we make to metadata logic and accepting an image in the request's payload would remove the need to S3 lookup. That is, we can ask Grid: "how would you store this image".

Thinking about it a little more, this version makes the most sense, it's probably less flexible as it requires the image to exist, but we don't have a use case for anything else atm.

mkuzdowicz:
> The dry run was more for testing any changes we make to metadata logic and accepting an image in the request's payload would remove the need to S3 lookup. That is, we can ask Grid: "how would you store this image".
> 
> Thinking about it a little more, this version makes the most sense, it's probably less flexible as it requires the image to exist, but we don't have a use case for anything else atm.

yeah but dry run is a bit different feature which we can add later by reusing the same function created hereto project the image without storing it in S3`ImageUploadProjector.projectImage`

But for `re-ingestion/reindex` feature you need to assume that it exists in s3 and you need the s3 lookup because s3 object have additional data like: `x-amz-meta-file_name`, `x-amz-meta-upload_time`, `x-amz-meta-uploaded_by`

image-loader/app/model/ImageUploadProjector.scala
```diff
@@ -0,0 +1,99 @@
...
+}
+
+class ImageUploadProjector(config: ImageUploadOpsCfg, imageOps: ImageOperations)
+                          (implicit val ec: ExecutionContext) {
+
+  private val imageUploadProjectionOps = new ImageUploadProjectionOps(config, imageOps)
+
+  def projectImage(srcFileDigest: DigestedFile, uploadedBy: String, uploadedTime: DateTime, uploadFileName: Option[String]): Future[Image] = {
+    val DigestedFile(tempFile_, id_) = srcFileDigest
+    // identifiers_ to rehydrate
```
akash1810: minor: can we add a TODO comment here?

mkuzdowicz: done

PR #2724, opened 2020-01-23
Add Image Counter Lambda

stephengeller:
## What does this change?
This adds a lambda that gets the current image count from the media service API and posts it to CloudWatch, so that we can have monitoring and alerting on the image count.

Whether, when deployed, we can execute the lambda and see the metric show up in CloudWatch.

## Who should look at this?

@guardian/digital-cms 

## Tested?
- [ ] locally
- [X] on TEST


image-counter-lambda-4621cc3c156f19193fcc
```diff
@@ -0,0 +1 @@
+image-counter-lambda
```
akash1810: This looks like an API key 😱

image-counter-lambda/src/handler.ts
```diff
@@ -0,0 +1,54 @@
...
+
+interface MediaAPICredentials {
+  baseUrl: string;
+  "X-Gu-Media-Key": string;
+}
+
+const getImageCount = async (
+  credentials: MediaAPICredentials
+): Promise<number> => {
+  const response = await fetch(credentials.baseUrl + "/images", {
```
akash1810: We could use the `length` querystring to control the response size. That is, `length=0` will yield a smaller response than the default which is `length=10`.

scripts/ci/image-counter-lambda-build.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+apt-get update
+apt-get install zip
```
akash1810: why is this needed?

image-counter-lambda-4621cc3c156f19193fcc
stephengeller: Sorry. It's been rotated out now.

image-counter-lambda/src/getCredentials.ts
```diff
@@ -0,0 +1,17 @@
...
+async function getCredentials() {
+  const s3 = new AWS.S3();
+  const params = {
+    Bucket: "grid-conf",
+    Key: `${process.env.STAGE}/image-counter-lambda/conf.json`
+  };
+
+  const data = await s3.getObject(params).promise();
+  const buffer = data.Body;
+  const body = buffer ? buffer.toString("utf8") : "";
```
jonathonherbert: NB: bonus new Typescript 3.7 feature -- in place of the ternary expression we should be able to write `buffer.toString?.("utf8") || ""` [thanks to optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining) 😁

scripts/ci/image-counter-lambda-build.sh
stephengeller: We are running this CI script from inside a node docker image that doesn't have zip, and riffraff-artefact requires it. I personally think this is a bit cleaner than messing around with nvm stuff, as the image takes care of the node version by virtue of using `node:10`.

stephengeller:
@akash1810 what did you have in mind for the logging? Not sure what you mean by strong structure, but keen to get some better logging in here for sure.

akash1810:
> @akash1810 what did you have in mind for the logging? Not sure what you mean by strong structure, but keen to get some better logging in here for sure.

If we log a object, then we will create distinct fields in ELK. For example:

```js
const logDetail = {
  count: 100
};

console.log("retrieved count", logDetail)
```

This will create a `count` field in ELK which we can use to graph on.

media-api/conf/routes
```diff
@@ -35,6 +35,7 @@ GET     /usage/quotas/:id                             controllers.UsageControlle
 # Management
 GET     /management/healthcheck                       com.gu.mediaservice.lib.management.ElasticSearchHealthCheck.healthCheck
 GET     /management/manifest                          com.gu.mediaservice.lib.management.ManagementWithPermissions.manifest
+GET     /management/imageCounts                       com.gu.mediaservice.lib.management.ElasticSearchHealthCheck.imageCounts
```
akash1810: Could we split the additional route into another PR? This is because the order of deployment isn't completely deterministic; if the lambda was to ship first it'll get a 404 until media-api had deployed too.

stephengeller: [This has been moved here](https://github.com/guardian/grid/pull/2735)

common-lib/src/main/scala/com/gu/mediaservice/lib/management/Management.scala
```diff
@@ -63,4 +66,20 @@ class ElasticSearchHealthCheck(override val controllerComponents: ControllerComp
       }
     }
   }
+
+  def imageCounts: Action[AnyContent] = Action.async {
+    elasticsearch.countImages().map {
+      case _ @(catCount: CatCountResponse, searchResponse: SearchResponse, indexStats: IndexStatsResponse) =>
+        val json: String =
+          s"""{
+             |  "catCount": ${catCount.count},
```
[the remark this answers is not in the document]
stephengeller: > > @akash1810 what did you have in mind for the logging? Not sure what you mean by strong structure, but keen to get some better logging in here for sure.
> 
> If we log a object, then we will create distinct fields in ELK. For example:
> 
> ```js
> const logDetail = {
>   count: 100
> };
> 
> console.log("retrieved count", logDetail)
> ```
> 
> This will create a `count` field in ELK which we can use to graph on.

This has now been added, take a look at the lambda function on line 48.

scripts/ci/image-counter-lambda-build.sh
AWare: I just saw this, and this is fixed by using node-riffraff-artifact

PR #2727, opened 2020-01-27
[thrall] Catch timeouts and log with message details 

AWare:
## What does this change?
This adds a catch for `TimeoutException` in Thrall which logs the offending error message.

No more `Exception during process record block` with `TimeoutException` and instead some nice new log messages with the message data.

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/thrall/app/lib/kinesis/ThrallEventConsumer.scala
+++ b/thrall/app/lib/kinesis/ThrallEventConsumer.scala
@@ -15,7 +15,7 @@ import play.api.Logger
 import play.api.libs.json.{JodaReads, Json}
 
 import scala.concurrent.duration.{Duration, SECONDS}
-import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.concurrent.{Await, ExecutionContext, Future, TimeoutException}
 
 class ThrallEventConsumer(es: ElasticSearch6,
                           thrallMetrics: ThrallMetrics,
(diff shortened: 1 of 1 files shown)


thrall/app/lib/kinesis/ThrallEventConsumer.scala
```diff
@@ -58,15 +58,21 @@ class ThrallEventConsumer(es: ElasticSearch6,
...
                   Logger.error(s"Failed to process ${updateMessage.subject} message; message will be ignored:", e)(updateMessage.toLogMarker)
               }
-
-              Await.ready(eventuallyAppliedUpdate, ThirtySeconds)
+              try {
+                Await.ready(eventuallyAppliedUpdate, ThirtySeconds)
+              } catch {
+                case e: TimeoutException =>
+                  Logger.error(
+                    s"Timeout while processing ${updateMessage.subject} message; message will be ignored:",
```
akash1810: minor: Could we include the duration we waiting for in the log too to save us remembering/looking up what the value is?

(`ThirtySeconds` is a badly named variable, `Timeout` is more descriptive imo)

AWare: yeah, that's sensible

AWare: good catch

PR #2728, opened 2020-01-27
Thrall dead letter queue

AWare:
## What does this change?

This adds a dead letter queue sender to thrall so that dropped messages can be picked up.

The queue will fill up (we don't yet have a strategy to pick them back up).

## Tested?
- [X] locally
- [ ] on TEST

*NB this will need a conf change*


AWare:
We do need some retry strategy, as we're getting timeout exceptions. But unless we can stop out of order messages when we replay, the DLQ isn't going to work. The UI gives good feedback about update messages though, so we might be okay just replaying create and delete here.

PR #2731, opened 2020-01-28
Aa mk/admin tool project (local run only for now)

mkuzdowicz:
## What does this change?
Introducing new admin tools project for re-ingestion and image projection

This PR introduce only image projection
and have local run only for now

run the project locally by `./dev-start.sh` 
access 

https://admin-tools.media.local.dev-gutools.co.uk/images/{id}/project

to see the image projection

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/dev-start.sh
+++ b/dev-start.sh
@@ -115,7 +115,6 @@ setupLocalKinesis() {
   aws --profile media-service --region=eu-west-1 --endpoint-url=http://localhost:4568 kinesis list-streams
 }
 
-
 main() {
     checkRequirements
     checkNodeVersion
--- a/dev-start.sh
+++ b/dev-start.sh
@@ -115,10 +115,6 @@ setupLocalKinesis() {
   aws --profile media-service --region=eu-west-1 --endpoint-url=http://localhost:4568 kinesis list-streams
 }
 
-setupEnvVarsForAdminToolsAPI(){
-  export media_service_domain_root='local.dev-gutools.co.uk'
-  export media_service_api_key='dev-'
-}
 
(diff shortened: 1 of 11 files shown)


mkuzdowicz:
> This looks good, but I probably can't approve as I worked on it.
> 
> One question about using env-vars; are we happy to pass secrets (the api key) like this?

I added this like that because i was thinking that lambda will take that as env vars, but we can use secrets manager instead

- that will change a lot anyway as this PR is only for local run
and next work will bring restrcuture

mkuzdowicz:
> This looks good, but I probably can't approve as I worked on it.
> 
> One question about using envvars; are we happy to pass secrets (the api key) like this?

i commented that already before

But i guess as we use play only for local run i am happy to even hardcode it in this PR as for local run it does not matter

and we can get rid of env vars for local run

mkuzdowicz:
> > This looks good, but I probably can't approve as I worked on it.
> > One question about using envvars; are we happy to pass secrets (the api key) like this?
> 
> i commented that already before
> 
> But i guess as we use play only for local run i am happy to even hardcode it in this PR as for local run it does not matter
> 
> and we can get rid of env vars for local run

env vars removed

dev-start.sh
```diff
@@ -115,6 +115,7 @@ setupLocalKinesis() {
   aws --profile media-service --region=eu-west-1 --endpoint-url=http://localhost:4568 kinesis list-streams
 }
 
+
```
aug24: Remove redundant changed whitespace?

PR #2736, opened 2020-01-30
Upversion Elastic4S-http

aug24:
## What does this change?

Bump the version of elastic4s to a version which supports elasticsearch 7.

If media-api et al can parse elasticsearch responses when a node running 7.5.2 is in the cluster, then this has succeeded.

N/A

## Who should look at this?

Anyone.

## Tested?
- [ X] locally
- [ ] on TEST


jonathonherbert:
Closed in favour of #2844.

PR #2742, opened 2020-01-31
Create sum type for MimeType

akash1810:
## What does this change?
Removes the magic strings of `image/jpeg`, etc in favour of a sum type to get more help from the compiler.

no-op

n/a

## Tested?
- [ ] locally
- [ ] on TEST


common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
```diff
@@ -79,38 +78,41 @@ class ImageOperations(playPath: String) {
...
-      case "image/jpeg" => resizedFile
+      new File(optimisedImageName)
     }
+    case Jpeg => resizedFile
+
+    // This should never happen as we only ever crop as PNG or JPEG. See `Crops.cropType` and `CropsTest`
+    // TODO We should create a `CroppingMimeType` to enforce this at the type level.
+    //  However we'd need to change the `Asset` model as source image and crop use this model
+    //  and a source can legally be a `Tiff`. It's not a small change...
+    case Tiff => resizedFile
```
akash1810: This is a bit contentious, hoping the comment is enough for now.

common-lib/src/main/scala/com/gu/mediaservice/model/MimeType.scala
```diff
@@ -0,0 +1,41 @@
...
+class UnsupportedMimeTypeException extends Exception
+
+sealed trait MimeType {
+  def name: String = this match {
+    case Jpeg => "image/jpeg"
+    case Png => "image/png"
+    case Tiff => "image/tiff"
+  }
+
+  def fileExtension: String = name.split('/').reverse.head
```
paperboyo: Not that it matters much on output, but we currently produce eg.  a `jpg` extension. If that’s ever (to be) used on input, it would better take into account different spellings, though.

paperboyo: Sorry, ignore, saw [line 37](https://github.com/guardian/grid/pull/2742/files/2fbd965cc640d9965dffb1471bebc6c3e9a000eb#diff-214fa5d666a63a722162df9e56a4968dR37).

common-lib/src/main/scala/com/gu/mediaservice/lib/ImageStorage.scala
```diff
@@ -4,11 +4,11 @@ import java.util.concurrent.Executors
 import java.io.File
 import java.net.URI
 
-import scala.concurrent.{Future, ExecutionContext}
+import scala.concurrent.{ExecutionContext, Future}
```
AWare: thanks, intellij

common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
AWare: Should we log/throw if we get into this illegal state?

common-lib/src/main/scala/com/gu/mediaservice/model/MimeType.scala
AWare: So this would only recognise `.jpg` as a JPEG then?

cropper/app/lib/Crops.scala
```diff
@@ -111,15 +109,14 @@ object Crops {
...
     val outputAsPng = hasAlpha || isGraphic
 
-    if ( !isJpeg && outputAsPng ) {
-      ImageOperations.Png
-    } else {
-      ImageOperations.Jpeg
+    mediaType match {
+      case Png if outputAsPng => Png
+      case Tiff if outputAsPng => Png
+      case _ => Jpeg
```
AWare: 👏🏼👏🏼👏🏼
really love a good match

common-lib/src/main/scala/com/gu/mediaservice/model/MimeType.scala
paperboyo: 1. I hope we don’t do recognition by extension (we have Tika and metadata-extractor has some shiffing too), but wanted to double-check.
2. I thought this is about what extension we add when producing JPEGs.

akash1810: Correct, `fileExtension` is only used for the output of a crop in S3, not for representing mimetype.

common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
akash1810: Good idea!

PR #2746, opened 2020-02-03
Attempt to retry thrall message processing.

AWare:
## What does this change?

This adds a retry to the thrall message processor.
(NB looking for opinions especially on my use of the implicit marker context)

Failed messages get retried. 

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/thrall/app/lib/RetryHandler.scala
+++ b/thrall/app/lib/RetryHandler.scala
@@ -1,0 +1,60 @@
+package lib
+
+
+import akka.actor.ActorSystem
+import akka.pattern.{after, retry}
+import play.api.{Logger, MarkerContext}
+
+import scala.concurrent.{ExecutionContext, Future, TimeoutException}
+import scala.concurrent.duration.FiniteDuration
+import scala.util.{Failure, Success}
+
+object RetryHandler {
+  def handleWithRetryAndTimeout[T](f: () => Future[T],
+                                   retries: Int,
+                                   timeout: FiniteDuration,
+                                   delay: FiniteDuration
+                                  )(implicit actorSystem: ActorSystem,
+                                    executionContext: ExecutionContext,
+                                    mc: MarkerContext
+                                  ): () => Future[T] = {
+    def logFailures[T](f: () => Future[T])(
+      implicit executionContext: ExecutionContext, mc: MarkerContext
+    ): () => Future[T] = {
+      () => {
+        f().transform {
+          case Success(x) => Success(x)
+          case Failure(t: TimeoutException) => {
+            Logger.error("Failed with timeout. Will retry")
+            Failure(t)
+          }
+          case Failure(exception) => {
+            Logger.error("Failed with exception.", exception)
+            Failure(exception)
+          }
+        }
+      }
+    }
+
+    def handleWithTimeout[T](f: () => Future[T], attemptTimeout: FiniteDuration): () => Future[T] = () => {
+      val timeout = after(attemptTimeout, using = actorSystem.scheduler)(Future.failed(
+        new TimeoutException(s"Timeout of $attemptTimeout reached.")
+      ))
+      Future.firstCompletedOf(Seq(timeout, f()))
+    }
+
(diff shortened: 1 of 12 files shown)


thrall/app/lib/RetryHandler.scala
```diff
@@ -0,0 +1,40 @@
...
+      if (count > 0 ) {
+        Logger.info(s"Attempt $count of $retries")
+      } else {}
+      count = count + 1
+      f()
+    }
+    retry(attempt, retries, delay)
+  }
+
+  def handleWithRetryAndTimeout[T](f: () => Future[T], retries: Int, timeout: FiniteDuration)(implicit actorSystem: ActorSystem, executionContext: ExecutionContext, mc: MarkerContext): () => Future[T] = {
```
jonathonherbert: After going over the code with @AWare, the logic here makes sense given the kinds of failures we are expecting, but it'd be good to document the strategy we are using here, and why we're using it, for other developers.

thrall/app/lib/kinesis/ThrallEventConsumer.scala
```diff
@@ -74,8 +78,10 @@ class ThrallEventConsumer(es: ElasticSearch6,
   }
 
   private def processUpdateMessage(updateMessage: UpdateMessage) = {
+    implicit val mc: MarkerContext = updateMessage.toLogMarker
+
     messageProcessor.chooseProcessor(updateMessage).map { messageProcessor =>
-      Try(Await.ready(messageProcessor.apply(updateMessage), Timeout))
+      Try(Await.ready(RetryHandler.handleWithRetryAndTimeout(() => messageProcessor.apply(updateMessage),2, Timeout).apply(), Timeout))
```
jonathonherbert: This is a chunky nugget of code! Would it be possible to split it up to make it more readable? 😁 🍗

thrall/app/lib/RetryHandler.scala
```diff
@@ -0,0 +1,40 @@
...
+    Future.firstCompletedOf(Seq(timeout, f()))
+  }
+
+  def handleWithRetry[T](f: () => Future[T], retries: Int, delay: FiniteDuration)(implicit actorSystem: ActorSystem, executionContext: ExecutionContext,mc: MarkerContext): () => Future[T] = () => {
+    implicit val scheduler = actorSystem.scheduler
+    var count = 0
+    def attempt = () => {
+      if (count > 0 ) {
+        Logger.info(s"Attempt $count of $retries")
+      } else {}
```
jonathonherbert: Is this `else` necessary?

thrall/app/lib/RetryHandler.scala
```diff
@@ -0,0 +1,40 @@
...
+import play.api.{Logger, MarkerContext}
+
+import scala.concurrent.{ExecutionContext, Future, TimeoutException}
+import scala.concurrent.duration.FiniteDuration
+
+object RetryHandler {
+  def handleWithTimeout[T](f: () => Future[T], attemptTimeout: FiniteDuration)(implicit actorSystem: ActorSystem, executionContext: ExecutionContext,mc: MarkerContext): () => Future[T] = () => {
+
+    val timeout = after(attemptTimeout, using = actorSystem.scheduler)(Future.failed(
+      (() => {
```
jonathonherbert: We should be able to do this with a block, rather than an IIFE.

thrall/app/lib/kinesis/ThrallEventConsumer.scala
AWare: I've pulled it over some lines and commented them, not sure how much we can split them

AWare:
I've got some logic which logs in case of out of order return (which shouldn't be possible) and tidied some of the rest up.

AWare:
@akash1810 I have moved the running into its own function and put a test around it to ensure correct ordering.

AWare:
So I think I had a flaky test, after rerunning the tests the bulk insert test failed with 
image
which is from 
image
and all I can really think of is that somehow the test infrastructure in github isn't getting torn down after runs and is causing this test to erroneously fail.

PR #2747, opened 2020-02-03
Add _elasticsearch endpoint

jonathonherbert:
## What does this change?

This adds a `GET /images/:id/_elasticsearch` endpoint, which returns the raw output from ElasticSearch.

Useful when we'd like to examine the contents of ES directly, for example to checking reingestion dry runs against existing data.

### Why  _elasticsearch?

`_` indicates the method is something that isn't intended for public consumption. (It's also an Elasticsearch convention!) `elasticsearch` is as unambiguous as we can be 👍 

Tested on CODE.

Make a request to `GET /images/:id/_elasticsearch`. The result (in the `data` envelope) should be identical to the response from ES.

## Who should look at this?

@akash1810 
@mkuzdowicz 

## Tested?
- [x] locally
- [x] on TEST

### DIFF
--- a/media-api/app/controllers/MediaApi.scala
+++ b/media-api/app/controllers/MediaApi.scala
@@ -138,33 +138,6 @@ class MediaApi(
     }
   }
 
-  def getImageResponseFromES(id: String, request: Authentication.Request[AnyContent]): Future[Option[(Image, JsValue, List[Link], List[Action])]] = {
-    implicit val r: Authentication.Request[AnyContent] = request
-
-    val include = getIncludedFromParams(request)
-
-    elasticSearch.getImageById(id) map {
-      case Some(source) if hasPermission(request, source) =>
-        val writePermission = canUserWriteMetadata(request, source)
-        val deleteImagePermission = canUserDeleteImage(request, source)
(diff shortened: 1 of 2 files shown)


media-api/app/controllers/MediaApi.scala
```diff
@@ -120,28 +120,20 @@ class MediaApi(
...
+    }
+  }
 
+  /**
+    * Get the raw response from ElasticSearch.
+    */
+  def getImageRaw(id: String) = auth.async { request =>
+    getImageResponseFromES(id, request) map {
+      case Some((source, _, imageLinks, imageActions)) =>
+        respond(source, imageLinks, imageActions)
```
akash1810: Do we want the `imageLinks` and `imageActions` on the raw endpoint if it's designed to be an ES proxy? That is, I'm not sure the consumer of this endpoint will be using this data, so not sure its needed.

jonathonherbert: Yes, I added this for consistency with other endpoints -- I considered taking another approach but the Grid docs state 'exposed as hypermedia APIs (argo)'. If you think it's not necessary we can get rid.

PR #2748, opened 2020-02-03
proxy the api key from Gateway to the Grid APIs

akash1810:
## What does this change?
We want to have authentication on the admin-tools project. This could be achieved in a few ways:
- Like other Grid services, polling the S3 bucket where API keys live
- Acting as a proxy

The former brings a lot of Play dependencies into the lambda so we opted for the second. This change means:
- we fail fast of the api key header is empty
- if the api key is provided, we make a call to the leases api, if the response is a 200 we know the key is valid

API Gateway seems to lowercase headers, so we're doing a lowercase comparison, which is a bit gross!

admin-tools lambda has auth

n/a

## Who should look at this?

img

## Tested?
- [x] locally
- [x] on TEST

### DIFF
--- a/admin-tools/lambda/src/main/scala/com/gu/mediaservice/LambdaHandler.scala
+++ b/admin-tools/lambda/src/main/scala/com/gu/mediaservice/LambdaHandler.scala
@@ -2,16 +2,25 @@ package com.gu.mediaservice
 
 import com.amazonaws.services.lambda.runtime.Context
 import com.amazonaws.services.lambda.runtime.events.{APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent}
+import com.gu.mediaservice.lib.auth.Authentication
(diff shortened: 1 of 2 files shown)


admin-tools/lambda/src/main/scala/com/gu/mediaservice/LambdaHandler.scala
```diff
@@ -20,36 +29,50 @@ class LambdaHandler {
...
-      case _ =>
-        val emptyRes = Json.obj("message" -> s"image with id=$mediaId not-found").toString
-        println(s"image not projected \n $emptyRes")
-        new APIGatewayProxyResponseEvent()
-          .withStatusCode(404)
-          .withHeaders(Map("content-type" -> "application/json").asJava)
-          .withBody(emptyRes)
+    val headers = event.getHeaders.asScala.toMap
+
+    println(s"headers are $headers")
```
jonathonherbert: They sure are!

PR #2751, opened 2020-02-04
handle issue with http headers case sensitivity and refactoring

mkuzdowicz:
## What does this change?
- handle issue with http headers case sensitivity
- refactoring

clients like curl or API gateway may lowerCases custom header names

## Tested?
- [x] locally
- [x] on TEST

### DIFF
--- a/admin-tools/lambda/src/main/scala/com/gu/mediaservice/LambdaHandler.scala
+++ b/admin-tools/lambda/src/main/scala/com/gu/mediaservice/LambdaHandler.scala
@@ -13,14 +13,6 @@ import scala.concurrent.duration.Duration
 import scala.concurrent.{Await, Future}
 
 class LambdaHandler {
-  private def getUnauthorisedResponse = {
-    val res = Json.obj("message" -> s"missing or invalid api key header").toString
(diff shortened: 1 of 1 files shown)


admin-tools/lambda/src/main/scala/com/gu/mediaservice/LambdaHandler.scala
```diff
@@ -31,48 +23,63 @@ class LambdaHandler {
...
+    val res = Json.obj("message" -> s"missing or invalid api key header").toString
+
+    new APIGatewayProxyResponseEvent()
+      .withStatusCode(401)
+      .withHeaders(Map("content-type" -> "application/json").asJava)
+      .withBody(res)
+  }
+
+  private def getAuthKeyFrom(headers: Map[String, String]) = {
+    // clients like curl or API gateway may lowerCases custom header names, yay!
```
jonathonherbert: Maaaan

mkuzdowicz: :)

PR #2753, opened 2020-02-04
Create sum type for MimeType, take II

akash1810:
## What does this change?
https://github.com/guardian/grid/pull/2742 take II with an addition to support crops created in the early years of Grid (~2016) which state mime type w/out an 'image/' prefix.

There are quite a few of these images, so also adding a TODO to fix them later on.

```console
curl -XPOST 'http://localhost:9200/_search' -d '{
 "query": {
  "term": {
   "exports.assets.mimeType": "jpg"
  }
 }
}'
```

Total `jpg`: 567868
Total `png`: 9231

no-op

n/a

## Tested?
- [x] locally
- [x] on TEST


sihil:
I think this is a good thing in general. I'm not sure how it helps us deal with the current grid issue or which card this relates to - can we perhaps start linking to trello card in PRs? Also `jpg` is the pre-1998 windows extension so perhaps we can look to using `jpeg`?

akash1810:
Agreed. Not directly related to current Grid issue, just a general improvement. Happy to park for now.

I'm not sure if downstream services expect `jpg`, so reluctant to change in this PR.

image-loader/app/lib/imaging/MimeTypeDetection.scala
```diff
@@ -2,12 +2,18 @@ package lib.imaging
...
+
+import scala.util.{Failure, Success, Try}
 
 object MimeTypeDetection {
   val tika = new Tika()
 
-  def guessMimeType(file: File): Option[String] =
-    Option(tika.detect(file))
-
+  def guessMimeType(file: File): Either[UnsupportedMimeTypeException, MimeType] = Try(MimeType(tika.detect(file))) match {
```
akash1810, line 14: This allows us to fail fast on the `Left`.

akash1810:
Related to https://github.com/guardian/grid/pull/2895

common-lib/src/main/scala/com/gu/mediaservice/model/MimeType.scala
```diff
@@ -0,0 +1,53 @@
...
+object MimeType {
+  def apply(value: String): MimeType = value.toLowerCase match {
+    case "image/jpeg" => Jpeg
+    case "image/png" => Png
+    case "image/tiff" => Tiff
+
+    // Support crops created in the early years of Grid (~2016) which state mime type w/out an 'image/' prefix
+    // TODO correct these values in a reindex
+    case "jpg" => {
+      Logger.info("Encountered legacy mime type representation")
```
jonathonherbert, line 29: Is it useful to log out the type here as well? (It may well be not useful at all, I'm just wondering if we'd like to know numbers etc.)

image-loader/app/model/ImageUpload.scala
```diff
@@ -161,30 +161,29 @@ class Uploader(val store: ImageLoaderStore,
...
-    )
-
-    // Abort early if unsupported mime-type
-    val supportedMimeType = config.supportedMimeTypes.exists(guessedMimeType.contains(_))
-    if (!supportedMimeType)
-      Future.failed(new UnsupportedMimeTypeException(uploadRequest, guessedMimeType.getOrElse("Not Provided")))
-
-    Future.successful(uploadRequest)
 
+    MimeTypeDetection.guessMimeType(tempFile_) match {
```
jonathonherbert, line 165: Yes, this is nicer 😁

common-lib/src/main/scala/com/gu/mediaservice/model/MimeType.scala
```diff
@@ -0,0 +1,53 @@
+package com.gu.mediaservice.model
+
+import play.api.Logger
+import play.api.libs.json._
+
+class UnsupportedMimeTypeException(val mimeType: String) extends Exception
+
+sealed trait MimeType {
+  def name: String = this match {
+    case Jpeg => "image/jpeg"
```
jonathonherbert, line 10: This is nice, the types are clearer across the PR 👍

common-lib/src/main/scala/com/gu/mediaservice/model/MimeType.scala
akash1810: Added. I've not made it a marker or anything as we should be able to run a bulk update query on ES to correct things.

akash1810:
Fully tested on TEST and in locally using a PROD snapshot, so should not need reverting like last time 🤞 . Will merge tomorrow.

PR #2754, opened 2020-02-05
making async requests management better by bridge from OkHttp to futures

mkuzdowicz:
## What does this change?
making async requests management better by bridge from OkHttp to futures

## Tested?
- [x] locally
- [x] on TEST


mkuzdowicz:
> Looks good -- is there a reason we need to block on certain requests?

yes api key validation need to be done first
and image-loader projection need to be executed first and after that e can feed the rest like collections, crops and so on

PR #2757, opened 2020-02-05
grid-batch-index-input-generator lambda

mkuzdowicz:
## What does this change?
Introducing grid-batch-index-input-generator lambda code

Idea

lambda trigged in schedule every given interval
lambda calls dynamo table that store image ids to reindex and state at which the reindex was proceed
flow of lambda code

get given batch of ids from dynamo (for example 10) 
set state if ids to 1 which means in_progress
create images projection josn blobs
if some items projection was not possible to create
updateStateToNotFoundImages
put them to s3
send message to kinesis:
```
{
"subject" : "batch-index",
        "bulkIndexRequest": {
               "bucket": ...,
                "key": ...
       }
}
```
updateStateToFinished
if the process failed
resetItemsState

s3 files were created and notification message was send to kineses

Full e2e test: missing image was reindexed

## Tested?
- [x] locally
- [x] on TEST


.gitignore
```diff
@@ -17,3 +17,5 @@ imgops/dev/nginx.conf
 .bloop
 .metals
 /stress-test/test-files/
+
+LOGS_LOCATION_IS_UNDEFINED
```
akash1810: 😂

admin-tools/lambda/src/main/scala/com/gu/mediaservice/BatchIndexHandlerAwsFunctions.scala
```diff
@@ -0,0 +1,88 @@
...
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClient
+import com.amazonaws.services.dynamodbv2.document.{Table, DynamoDB => AwsDynamoDB}
+import com.amazonaws.services.kinesis.model.PutRecordRequest
+import com.amazonaws.services.kinesis.{AmazonKinesis, AmazonKinesisClientBuilder}
+import com.amazonaws.services.s3.AmazonS3ClientBuilder
+import com.amazonaws.services.s3.model.ObjectMetadata
+import com.gu.mediaservice.lib.json.JsonByteArrayUtil
+import play.api.libs.json.JsValue
+
+class BatchIndexHandlerAwsFunctions(cfg: BatchIndexHandlerConfig) {
```
akash1810: Could we reuse the AWS helpers in [`common-lib`](https://github.com/guardian/grid/tree/master/common-lib/src/main/scala/com/gu/mediaservice/lib/aws)?

admin-tools/lambda/src/main/scala/com/gu/mediaservice/BatchIndexLocalHandler.scala
```diff
@@ -0,0 +1,36 @@
...
+    batchSize = 1,
+    kinesisEndpoint = Some("http://localhost:4568"),
+    awsCreds = Some(awsCredentials)
+  )
+
+  private val batchIndex = BatchIndexHandler(cfg)
+
+  // ids that exists on your local grid
+  val mediaIds = List(
+    "011db3facffd4ef3d3e07eabbd3c07bb41cbf819",
```
akash1810: Can this be pushed to input rather than hard coded?

admin-tools/lambda/src/main/scala/com/gu/mediaservice/BatchIndexLocalHandler.scala
```diff
@@ -0,0 +1,36 @@
...
+import com.amazonaws.auth.profile.ProfileCredentialsProvider
+
+import scala.concurrent.ExecutionContext.Implicits.global
+
+object BatchIndexLocalHandler extends App {
+
+  lazy val awsCredentials = new ProfileCredentialsProvider("media-service")
+
+  private val cfg = BatchIndexHandlerConfig(
+    apiKey = "dev-",
```
akash1810: Can this be pushed to config rather than hard coded?

admin-tools/lib/src/main/scala/com/gu/mediaservice/ImagesBatchProjection.scala
```diff
@@ -0,0 +1,38 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+
+object ImagesBatchProjection {
+  def apply(apiKey: String, domainRoot: String): ImagesBatchProjection =
+    new ImagesBatchProjection(apiKey, domainRoot)
+}
+
+case class ImageIdMaybeBlobEntry(id: String, blob: Option[String])
+
+case class ImageIdBlobEntry(id: String, blob: String)
```
akash1810: Why do we need this and `ImageIdMaybeBlobEntry`?

admin-tools/lambda/src/main/scala/com/gu/mediaservice/BatchIndexHandler.scala
```diff
@@ -0,0 +1,155 @@
...
+    val state = 1
+    println(s"updating items state to in progress")
+    updateItemsState(ids, state)
+    state
+  }
+
+  // used to track images that were not projected successfully
+  def updateStateToNotFoundImages(notFoundIds: List[String]): Option[Int] = {
+    if (notFoundIds.isEmpty) None else {
+      val state = 2
```
akash1810: Can we create a sum type for `state` to be explicit of their meaning?

admin-tools/lambda/src/main/scala/com/gu/mediaservice/BatchIndexHandlerAwsFunctions.scala
```diff
@@ -0,0 +1,88 @@
...
+
+  import cfg._
+
+  def putToS3(imageBlobs: List[String]): String = {
+    val fileContent = imageBlobs.mkString("\n")
+    val key = s"batch-index/${UUID.randomUUID().toString}.json"
+    val metadata = new ObjectMetadata
+    metadata.setContentType("application/json")
+    val res = s3client.putObject(batchIndexBucket, key, new ByteArrayInputStream(fileContent.getBytes), metadata)
+    val path = s"s3://$batchIndexBucket/$key"
```
akash1810: I think the message just needs to be the key as `thrall` should get the bucket from config.

.gitignore
mkuzdowicz: logstash complaints :D

admin-tools/lambda/src/main/scala/com/gu/mediaservice/BatchIndexHandlerAwsFunctions.scala
mkuzdowicz: i will prefer to provide that in message rather that introducing another config to be honest

admin-tools/lambda/src/main/scala/com/gu/mediaservice/BatchIndexLocalHandler.scala
mkuzdowicz: this is BatchIndexLocalHandler for local run only like for dev i will move that to dev pakcage

admin-tools/lambda/src/main/scala/com/gu/mediaservice/BatchIndexHandlerAwsFunctions.scala
akash1810: We also have this [message publisher](https://github.com/guardian/grid/blob/f8b6c7d63de84fa9813db537d034322aaa2751de/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/ThrallMessageSender.scala#L16).

mkuzdowicz: i was looking into that but they have a bit different use case, the use well with EC2
were you have creds chain profile -> instance profile for lambda the best solution is just standard client without any creds as it will use execution role, but i can double check

admin-tools/lambda/src/main/scala/com/gu/mediaservice/BatchIndexLocalHandler.scala
mkuzdowicz: it is the same approach as we sued for projection api inside play

admin-tools/lib/src/main/scala/com/gu/mediaservice/ImagesBatchProjection.scala
mkuzdowicz: to be able to track which images were successfully projected

ImageIdMaybeBlobEntry will give you id and maybe something
and then it will be mapped to lis of ImageIdBlobEntry (success) and list of String (failed projection id) in `BatchIndexHandler.partitionToSuccessAndNotFound `function

such that we can record state in Dynamo
1 -> in progress
2 -> not found
3 -> all done

mkuzdowicz: i didn't want to use pure tuples

admin-tools/lambda/src/main/scala/com/gu/mediaservice/BatchIndexLocalHandler.scala
mkuzdowicz: done

admin-tools/lambda/src/main/scala/com/gu/mediaservice/BatchIndexHandler.scala
mkuzdowicz: ok

admin-tools/lambda/src/main/scala/com/gu/mediaservice/BatchIndexLocalHandler.scala
mkuzdowicz: done

admin-tools/lambda/src/main/scala/com/gu/mediaservice/BatchIndexHandler.scala
mkuzdowicz: done

admin-tools/lambda/src/main/scala/com/gu/mediaservice/BatchIndexHandlerAwsFunctions.scala
mkuzdowicz: or we can use case class for S3Location(bucket: String, key: String) if thrall s3 client needs that

mkuzdowicz: i mention that in your PR: https://github.com/guardian/grid/pull/2767/files#diff-9e12ba48c7d2af2248094256c01c3d36R100

mkuzdowicz: refined

admin-tools/lib/src/main/scala/com/gu/mediaservice/ImagesBatchProjection.scala
```diff
@@ -0,0 +1,40 @@
+package com.gu.mediaservice
+
+import com.gu.mediaservice.lib.config.{ServiceHosts, Services}
+import com.gu.mediaservice.model.Image
+
+import scala.concurrent.duration.Duration
+import scala.concurrent.{Await, ExecutionContext, Future}
+
+object ImagesBatchProjection {
+  def apply(apiKey: String, domainRoot: String, timeout: Duration): ImagesBatchProjection =
```
akash1810: This feels a bit redundant as it has the same constructor as the class, so `new ImagesBatchProjection(args)` has the same effect - we're trading `new` for an `apply`?

admin-tools/lib/src/main/scala/com/gu/mediaservice/ImagesBatchProjection.scala
```diff
@@ -0,0 +1,40 @@
...
+    new ImagesBatchProjection(apiKey, domainRoot, timeout)
+}
+
+class ImagesBatchProjection(apiKey: String, domainRoot: String, timeout: Duration) {
+
+  private val ImageProjector = createImageProjector
+
+  def getMaybeImagesProjectionBlobs(mediaIds: List[String])(implicit ec: ExecutionContext): List[Either[Image, String]] = {
```
akash1810: Would `getImageProjections` be more intention revealing here?

mkuzdowicz: or `getImagesProjection` ?

admin-tools/lib/src/main/scala/com/gu/mediaservice/ImagesBatchProjection.scala
mkuzdowicz: in lambda it does not make difference so i can roll back to new

admin-tools/lib/src/main/scala/com/gu/mediaservice/ImagesBatchProjection.scala
mkuzdowicz: done

PR #2759, opened 2020-02-06
Introduce a sum type of messages for better strictness across the message boundary

jonathonherbert:
## What does this change?

introduces a reindex message to the list of processed message types as a way of introducing a sum type for message processing.

We can migrate other message types to this new methodology in a similar way.

## Why?

Rather than extending the existing `UpdateMessage` class, this adds a new sum type in a `ThrallMessage` namespace.

This isn't necessary for this operation -- we could just add another line to our `chooseProcessor` function in `MessageProcessor`. Why do it?

- it opens a door to better typesafety for message producers and consumers at the boundaries of the queue
- as a result it can simplify downstream code (e.g. handlers in `MessageProcessor`, which at the moment processes JSON structures case-by-case to ensure they're well-formed).

Issuing a correctly formed `reindex` message to the Kinesis queue (see `ThrallMessageSender`) should result in an image being reindexed.

## Who should look at this?

@akash1810 
@mkuzdowicz 

## Tested?
- [ ] locally
- [ ] on TEST


common-lib/src/main/scala/com/gu/mediaservice/lib/aws/ThrallMessageSender.scala
```diff
@@ -1,33 +1,73 @@
...
+  }
 }
 
 // TODO add RequestID
+
+sealed abstract class ThrallMessage extends Product with Serializable {
+  def getId: String
+  val subject: String
+
+  def toLogMarker: LogstashMarker = {
```
AWare: I think this might be worth a slight revisit soon, as there's places that log id without the marker- and places where we could log a few other markers and it's not easy.

PR #2765, opened 2020-02-07
add bulk insert ability to thrall

akash1810:
## What does this change?
Updates the elasticsearch client in thrall with the ability bulk insert images. We're not linked to the message processor yet, so this is just a preliminary PR to add the tests.

Thrall can bulk insert images.

n/a

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/thrall/app/lib/elasticsearch/ElasticSearch.scala
+++ b/thrall/app/lib/elasticsearch/ElasticSearch.scala
@@ -27,6 +27,20 @@ class ElasticSearch(config: ElasticSearchConfig, metrics: Option[ThrallMetrics])
   lazy val shards = config.shards
   lazy val replicas = config.replicas
 
+  def bulkInsert(images: Seq[Image])(implicit ex: ExecutionContext): List[Future[ElasticSearchBulkUpdateResponse]] = {
+    val request = bulk {
+      images.map(img => {
+        indexInto(imagesAlias, Mappings.dummyType)
+          .id(img.id)
+          .source(Json.stringify(Json.toJson(img)))
+      })
+    }
+
+    val response = executeAndLog(request, s"Bulk inserting ${images.length} images")
+
+    List(response.map(_ => ElasticSearchBulkUpdateResponse()))
+  }
+
   def indexImage(id: String, image: JsValue)(implicit ex: ExecutionContext): List[Future[ElasticSearchUpdateResponse]] = {
     val painlessSource = loadPainless(
       // If there are old identifiers, then merge any new identifiers into old and use the merged results as the new identifiers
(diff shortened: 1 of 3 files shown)


thrall/test/lib/elasticsearch/ElasticSearchTest.scala
```diff
@@ -15,6 +19,30 @@ class ElasticSearchTest extends ElasticSearchTestBase {
...
+            userMetadata = Some(Edits(labels = List("foo", "bar"), metadata = ImageMetadata(description = Some("my description"))))
+          )
+
+          val imageTwo = createImage("superman", StaffPhotographer("Clark Kent", "Kent Farm")).copy(
+            usages = List(usage())
+          )
+
+          val images: List[Image] = List(imageOne, imageTwo)
+
+          def eventualMatchAllSearchResponse: Future[Response[SearchResponse]] = ES.client.execute(ElasticDsl.search(elasticSearchConfig.alias).size(images.length * 10))
```
akash1810: Would be nice to re-use `ES.countImages` here, but couldn't get the test to pass. This implementation is inspired by a similar scenario in [`media-api`](https://github.com/guardian/grid/blob/32d06a9db2ff068c48e3a785c09102842d13b5cd/media-api/test/lib/elasticsearch/ElasticSearchTest.scala#L69)

thrall/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -27,6 +27,20 @@ class ElasticSearch(config: ElasticSearchConfig, metrics: Option[ThrallMetrics])
...
 
+  def bulkInsert(images: Seq[Image])(implicit ex: ExecutionContext): List[Future[ElasticSearchBulkUpdateResponse]] = {
+    val request = bulk {
+      images.map(img => {
+        indexInto(imagesAlias, Mappings.dummyType)
+          .id(img.id)
+          .source(Json.stringify(Json.toJson(img)))
+      })
+    }
+
```
mkuzdowicz: Suggestion to consider:

according to contract between lambda that `produces and notify`
the blob read from s3 will be json string with `\n` delimiter
so i think we do not need to serialize and deserialize that again

Unless you think that it will improve contract between services

akash1810: The intention here is for `bulkInsert` to only accept `Seq[Image]` as input for type safety.

The contract between the services will be upheld in the [message processor](https://github.com/guardian/grid/blob/master/thrall/app/lib/kinesis/MessageProcessor.scala#L22). On receipt of a message, `thrall` will get the JSON blob and attempt to parse it as `Seq[Image]` and call `bulkInsert` if parsing is successful, if unsuccessful, log and fail. This provides a nice separation of concern as the ES class doesn't care about validating the S3 JSON. Reading the JSON from S3 will be done in a different PR.

mkuzdowicz: yeah i think performance implication will be very small so i am ok with that

PR #2766, opened 2020-02-07
use alias when querying for stats

akash1810:
## What does this change?
Using the alias means we don't need to rely on the index name as this will change if/when we reindex the entire cluster.

Should be a no-op.

n/a

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch6/ElasticSearchClient.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch6/ElasticSearchClient.scala
@@ -64,9 +64,9 @@ trait ElasticSearchClient extends ElasticSearch6Executions {
 
 
   def countImages(): Future[ElasticSearchImageCounts] = {
-    val queryCatCount = catCount("images") // document count only of index including live documents, not deleted documents which have not yet been removed by the merge process
-    val queryImageSearch = search("images") limit 0 // hits that match the query defined in the request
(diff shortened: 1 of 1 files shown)


paperboyo:
👍 (where’s my Review button?!)

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch6/ElasticSearchClient.scala
```diff
@@ -64,9 +64,9 @@ trait ElasticSearchClient extends ElasticSearch6Executions {
 
 
   def countImages(): Future[ElasticSearchImageCounts] = {
-    val queryCatCount = catCount("images") // document count only of index including live documents, not deleted documents which have not yet been removed by the merge process
-    val queryImageSearch = search("images") limit 0 // hits that match the query defined in the request
-    val queryStats = indexStats("images") // total accumulated values of an index for both primary and replica shards
+    val queryCatCount = catCount(imagesAlias) // document count only of index including live documents, not deleted documents which have not yet been removed by the merge process
+    val queryImageSearch = search(imagesAlias) limit 0 // hits that match the query defined in the request
+    val queryStats = indexStats(imagesAlias) // total accumulated values of an index for both primary and replica shards
```
akash1810: We're not able to get stats for an alias directly. We'd first need to lookup which index the alias points to, then get the stats for that index.

Closing this PR for now as it's not worth the effort.

PR #2767, opened 2020-02-07
Add `batch-index` message processing to thrall

akash1810:
## What does this change?
This message will have a `s3Path` property which is a path to an object in S3. We'll take that file, read as string and parse it as `List[Image]` and then bulk insert into elasticsearch (see https://github.com/guardian/grid/pull/2765).

If we're unable to parse the file to the correct type, we throw.

We can process a `batch-index` message.

n/a

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/BaseStore.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/BaseStore.scala
@@ -2,10 +2,7 @@ package com.gu.mediaservice.lib
 
 import java.io.InputStream
 
-import _root_.play.api.Logger
 import akka.actor.Scheduler
-import com.amazonaws.AmazonServiceException
-import com.amazonaws.util.IOUtils
 import com.gu.Box
 import com.gu.mediaservice.lib.aws.S3
 import com.gu.mediaservice.lib.config.CommonConfig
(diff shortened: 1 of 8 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/BaseStore.scala
```diff
@@ -25,19 +22,7 @@ abstract class BaseStore[TStoreKey, TStoreVal](bucket: String, config: CommonCon
   protected val store: Box[Map[TStoreKey, TStoreVal]] = Box(Map.empty)
   protected val lastUpdated: Box[DateTime] = Box(DateTime.now())
 
-  protected def getS3Object(key: String): Option[String] = {
```
akash1810: This moves to the more generic `S3` class - https://github.com/guardian/grid/pull/2767/files#diff-9e12ba48c7d2af2248094256c01c3d36R100

thrall/app/lib/ThrallConfig.scala
```diff
@@ -25,4 +25,5 @@ class ThrallConfig(override val configuration: Configuration) extends CommonConf
 
   lazy val from: Option[DateTime] = properties.get("rewind.from").map(ISODateTimeFormat.dateTime.parseDateTime)
 
+  lazy val bulkIndexBucket: String = properties("s3.bulk.index.bucket")
```
akash1810: Thrall now has a new config to read from `/etc/gu/thrall.properties`.

mkuzdowicz: do we need that ? s3Path in message should be full

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/ThrallMessageSender.scala
```diff
@@ -40,7 +40,8 @@ case class UpdateMessage(
   crops: Option[Seq[Crop]] = None,
   mediaLease: Option[MediaLease] = None,
   leases: Option[Seq[MediaLease]] = None,
-  syndicationRights: Option[SyndicationRights] = None
+  syndicationRights: Option[SyndicationRights] = None,
+  s3Path: Option[String] = None
 ) {
```
mkuzdowicz: s3Path could be either `Option[String]` or case class for bucket and key it thrall s3 client needs that, i prefer not to have that bucket in config as it is better to decouple that

PR #2768, opened 2020-02-10
test for mateusz

AWare:
change the polling for grid uploads to see if it changes anything
@paperboyo


paperboyo:
This didn’t help. Thanks! The only other thing to test would be https://github.com/guardian/grid/pull/2696 (no idea if it makes sense, but can’t remember any other PRs to this functionality that could affect large files upload, possibly some other part of the forest has changed or, maybe, browsers?)

PR #2770, opened 2020-02-11
Add reindex route to admin-tools

jonathonherbert:
## What does this change?

Adds the option to reingest images when projecting them. Reingestion is a destructive reindex – it'll stomp on anything that is already there in ES.

Intended to be used with the upcoming CLI tool.

Call `POST admin-tools.{local|test}.dev-gutools.co.uk/images/project/:id?reingest=true` for an image that is present everywhere but Elasticsearch.  You should receive a 204 response. The image should reappear in ES with other data (e.g. leases, collections) restored.

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/admin-tools/lambda/src/main/scala/com/gu/mediaservice/ImageProjectionLambdaHandler.scala
+++ b/admin-tools/lambda/src/main/scala/com/gu/mediaservice/ImageProjectionLambdaHandler.scala
@@ -48,14 +48,11 @@ class ImageProjectionLambdaHandler extends LazyLogging {
     val merger = new ImageDataMerger(config)
     val result: FullImageProjectionResult = merger.getMergedImageData(mediaId.asInstanceOf[String])
     result match {
-      case FullImageProjectionSuccess(mayBeImage) =>
-        mayBeImage match {
-          case Some(img) =>
-            if (reingest) { putToKinesis(img, streamName) }
-            getSuccessResponse(img)
-          case _ =>
-            getNotFoundResponse(mediaId)
-        }
+      case FullImageProjectionSuccess(Some(image)) =>
+        if (reingest) { putToKinesis(image, streamName) }
+        getSuccessResponse(image)
+      case FullImageProjectionSuccess(None) =>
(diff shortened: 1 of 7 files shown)


admin-tools/lambda/src/main/scala/com/gu/mediaservice/ImageProjectionLambdaHandler.scala
```diff
@@ -3,53 +3,72 @@ package com.gu.mediaservice
...
+  private val streamName = sys.env("KINESIS_STREAM")
+  // if we want to release the load from main grid image-loader we can pass a dedicated endpoint
+  private val imageLoaderEndpoint = sys.env.get("IMAGE_LOADER_ENDPOINT")
 
   def handleRequest(event: APIGatewayProxyRequestEvent, context: Context): APIGatewayProxyResponseEvent = {
-
-    logger.info(s"handleImageProjection event: $event")
-
     val mediaId = event.getPath.stripPrefix("/images/projection/")
+    val shouldReingest = if (event.getQueryStringParameters != null) {
```
jonathonherbert: If there's a better way of handling this `null` case, _please_ let me know 😁

admin-tools/lambda/src/main/scala/com/gu/mediaservice/ImageProjectionLambdaHandler.scala
```diff
@@ -3,53 +3,69 @@ package com.gu.mediaservice
...
     }
   }
 
+  private def projectImage(mediaId: String, config: ImageDataMergerConfig, reingest: Boolean) = {
+    logger.info(s"starting handleImageProjection for mediaId=$mediaId, reingest=$reingest")
+    logger.info(s"with config: $config")
+
+    val merger = new ImageDataMerger(config)
+    val result: FullImageProjectionResult = merger.getMergedImageData(mediaId.asInstanceOf[String])
+    result match {
```
jonathonherbert: I flattened this out – hopefully it's a little more readable.

PR #2771, opened 2020-02-11
Add script to help scanning multiple dynamo tables for keys

jonathonherbert:
## What does this change?

Adds a script to help when scanning multiple dynamo tables for keys. We've needed to do this to figure out what was present in our dynamo tables relative to our ElasticSearch index.

As an example – the command
```
node scan-dynamo-table-for-pk.js table1,table2 outputFile.txt table1key,table2Key
```
Will scan table1 and table2 for keys table1key and table2key respectively, writing their values (delimited by \n) to outputFile.txt.

## Tested?
- [x] locally
- [ ] on TEST


jonathonherbert:
Hah! Yes, that might take longer 👍

PR #2772, opened 2020-02-11
Remove the complex gubbins from the file upload

AWare:
## What does this change?

This stops wrapping uploaded images in an array, which reduces the amount of memory our javascript is requesting- and allows the browser to manage the upload. thanks @mbarton 

This also corrects the method call which frees the object url.

We've managed to upload 250mb images, but 700mb is above the limit as image-loader seems to have some issues handling really big files

Able to upload big images.

## Who should look at this?
@akash1810 

## Tested?
- [X] locally
- [X] on TEST

### DIFF
--- a/kahuna/package-lock.json
+++ b/kahuna/package-lock.json
@@ -7287,520 +7287,390 @@
           "dependencies": {
             "abbrev": {
               "version": "1.1.1",
-              "bundled": true,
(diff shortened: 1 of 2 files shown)


paperboyo:
👏 Thanks, Gentlemen!

PR #2773, opened 2020-02-11
Add script to upload files with lists of image IDs to dynamo

jonathonherbert:
## What does this change?

Adds a script to upload files with lists of image IDs to dynamo.

Example usage:

```
node upload-ids-to-dynamo.js tableName fileName.txt
```

Run the script. You should see a table full of image IDs with the progress_state set to `0`.

## Tested?
- [x] locally
- [ ] on TEST


scripts/reindex-images/scan-dynamo-table-for-keys.js
```diff
@@ -0,0 +1,87 @@
+const AWS = require("aws-sdk");
```
akash1810: is this the same as https://github.com/guardian/grid/pull/2771?

jonathonherbert: Sorry, needs to be diffed from #2771, please review that or I can recreate the PR and use #2771 as base (as I should have done when creating this PR)

jonathonherbert: Oh wait, you can make the change whilst the PR is open! C'est extra!

PR #2775, opened 2020-02-12
calling projection lambda via api gateway instead of projection function directly

mkuzdowicz:
## What does this change?
calling projection lambda via api gateway instead of projection function directly

everything still works as before

## Tested?
- [x] locally
- [x] on TEST

### DIFF
--- a/admin-tools/lib/src/main/scala/com/gu/mediaservice/BatchIndexHandler.scala
+++ b/admin-tools/lib/src/main/scala/com/gu/mediaservice/BatchIndexHandler.scala
@@ -18,7 +18,7 @@ import scala.util.{Failure, Success, Try}
 
 case class BatchIndexHandlerConfig(
                                     apiKey: String,
-                                    domainRoot: String,
+                                    projectionEndpoint: String,
                                     batchIndexBucket: String,
                                     kinesisStreamName: String,
                                     dynamoTableName: String,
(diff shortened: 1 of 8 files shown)


admin-tools/lib/src/main/scala/com/gu/mediaservice/BatchIndexHandler.scala
```diff
@@ -39,8 +39,9 @@ class BatchIndexHandler(cfg: BatchIndexHandlerConfig) {
   private val GetIdsTimeout = new FiniteDuration(GetIdsTimoutInMins, TimeUnit.MINUTES)
   private val GlobalTimeout = new FiniteDuration(GlobalTimoutInMins, TimeUnit.MINUTES)
   private val ImagesProjectionTimeout = new FiniteDuration(ProjectionTimoutInMins, TimeUnit.MINUTES)
+  private val gridClient = new GridClient()
```
mkuzdowicz: without new
  private val gridClient = GridClient()
and pass maxIdleConnections

admin-tools/scripts/upload_jar.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+stage="$1"
+
+aws s3 cp admin-tools/lambda/target/scala-2.12/admin-tools-lambda.jar \
```
akash1810: This will change as soon as I get the riffraff integration working 😄

PR #2667, opened 2019-10-24
Feature/add people in image metadata

jamie-pitts:
Note: This PR requires https://github.com/guardian/grid/pull/2664 to be merged first.
**As this adds a new mapping into ES the `UpdateMapping` command will need to be run against the ES index first.** (this is just adding a new field and not editing existing ones, so relatively safe/easy)

## What does this change?

This adds a new field `peopleInImage` to the metadata of the image. It extracts from the XMP fields: `Iptc4xmpExt:PersonInImage` and `GettyImagesGIFT:Personality` - more can easily be added if/when found. 

The following is an example of the metadata section in the api response after uploading an image which includes the person in image metadata:
```
  "metadata": {
    "dateTaken": "2019-10-24T15:09:53.460Z",
    "description": "TOKYO, JAPAN - OCTOBER 24: Kieran Read, Scott Barrett, Aaron Smith, Codie Taylor and Shannon Frizell of the All Blacks stretch during a New Zealand All Blacks training session at Tatsuminomori Seaside Park on October 24, 2019 in Tokyo, Japan. (Photo by Hannah Peters/Getty Images)",
    "credit": "Getty Images",
    "byline": "Hannah Peters",
    "bylineTitle": "Staff",
    "title": "New Zealand Media Access",
    "copyrightNotice": "2019 Getty Images",
    "copyright": "2019 Getty Images",
    "suppliersReference": "775424562",
    "source": "Getty Images AsiaPac",
    "specialInstructions": "All use in books (including autobiographies) or magazines, which are entirely dedicated to any and/or all RWCL competitions and/or Matches requires separate written approval from Rugby World Cup Limited.",
    "keywords": [
      "sport",
      "rugby",
      "rugby union",
      "rugby union world cup",
      "rugby union world cup 2019",
      "all blacks - rugby team",
      "bestof",
      "topix"
    ],
    "city": "Tokyo",
    "country": "Japan",
    "subjects": [
      "sport"
    ],
    "peopleInImage": [
      "Codie Taylor",
      "Aaron Smith",
      "Shannon Frizell",
      "Kieran Read",
      "Scott Barrett"
    ]
  }
```

This will be the first PR in a series of PRs to add functionality to search for people who are in an image, and to display who is in an image in the metadata section of the UI.

PR #2778, opened 2020-02-12
Remove redundant copyrightNotice field

paperboyo:
NB: this does not require an immediate remapping, as we'll no longer be passing the `copyrightNotice` field into Elasticsearch; you'd need to remap before a reindex, though, to ensure that index is removed on existing images.

## What does this change?
This removes copyrightNotice field as this is an [equivalent of copyright](https://www.iptc.org/std/photometadata/specification/IPTC-PhotoMetadata#copyright-notice). It also reconciles copyright correctly (XMP>IPTC>EXIF). It also makes copyright cleaner only apply to byline and credit fields as copyright is the field that holds a statement of rights and we shouldn’t mess around with its contents just because it can be inconsistent, even more so, because we don’t use it anywhere that would require it to be cleaned or consistent.

No useless fields.

## Tested?
- [ ] locally
- [x] on TEST


jennygrahamjones:
> This requires remapping!
Does this mean there is some action required in addition to merging the PR?

paperboyo:
> Does this mean there is some action required in addition to merging the PR?

That’s my understanding, reading the note at the beginning of https://github.com/guardian/grid/pull/2667. Similarly, [this one](https://github.com/guardian/grid/pull/2716) may need one too (analyzers change). Will discuss and won’t merge myself!

paperboyo:
[[mappings change] copyright and copyrightNotice should be one field](https://trello.com/c/aguHJcul/1632-mappings-change-copyright-and-copyrightnotice-should-be-one-field)

akash1810:
This change drops a property from documents in ES. I think it should be safe to merge this w/out remapping. We should test on TEST first though!

jonathonherbert (approved):
Been through the problem and code w/ @paperboyo and all looks good. Presume this is tested in CODE, Mat, v. happy to test if not.

paperboyo:
> Presume this is tested in CODE

Yep. Just sent an email, if we don’t hear anything, will merge. Thx!

PR #2780, opened 2020-02-13
add riffraff to admin-tools lambdas

akash1810:
## What does this change?
Add RiffRaff to the admin-tools lambdas to bring the deployment in line with other apps and we can also start having CD on it too.

Easier deployment strategy.

n/a

## Who should look at this?

@mkuzdowicz 
@jonathonherbert 

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/admin-tools/lambda/riff-raff.yaml
+++ b/admin-tools/lambda/riff-raff.yaml
@@ -1,0 +1,12 @@
+stacks: [media-service]
+regions: [eu-west-1]
+deployments:
+  admin-tools-lambda:
+    type: aws-lambda
+    parameters:
+      bucket: media-service-dist
+      fileName: admin-tools-lambda.jar
+      prefixStack: false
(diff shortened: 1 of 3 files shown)


admin-tools/lambda/riff-raff.yaml
```diff
@@ -0,0 +1,12 @@
...
+deployments:
+  admin-tools-lambda:
+    type: aws-lambda
+    parameters:
+      bucket: media-service-dist
+      fileName: admin-tools-lambda.jar
+      prefixStack: false
+      functionNames:
+        - "grid-admin-tools-lambda-"
+        - "grid-batch-index-input-generator-lambda-"
```
mkuzdowicz: now it is
`admin-tools-image-projection-lambda-{stage}`
`admin-tools-image-batch-index-lambda-{stage}`

PR #2783, opened 2020-02-13
Remove Redundant Bylines and Credits

itsibitzi:
## What does this change?

Sometimes the byline and credits contain "redundant" values. A few of these have been found in the wild, values such as `HANDOUT`, `stringer`, `Uncredited`, etc. etc.

These should be removed!

In the case of the byline field these can be removed completely but it's not valid for there to be no credit field. In the case of a credit having on or more of these redundant values we either keep the redundant value (if there's one) or choose the right-most.

Added tests with both`"/"` and `" / "` as the token separator

## Tested?
- [x] locally
- [x] on TEST

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/RedundantTokenRemover.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/RedundantTokenRemover.scala
@@ -1,0 +1,41 @@
+package com.gu.mediaservice.lib.cleanup
+import com.gu.mediaservice.model.ImageMetadata
+
+object RedundantTokenRemover extends MetadataCleaner {
+  val toRemove = List(
+    "Handout",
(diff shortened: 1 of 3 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/RedundantTokenRemover.scala
```diff
@@ -0,0 +1,41 @@
...
+      } else {
+        Some(c)
+      }
+    }.map(_.trim),
+  )
+
+  def removeHandoutTokens(text: String): String = {
+    text.split("/").filter { tok =>
+      val trimmedToken = tok.trim
+      !toRemove.contains(trimmedToken)
```
akash1810: nit: should we `!toRemove.contains(trimmedToken.toLower())` and then the `toRemove` val can be smaller?

PR #2787, opened 2020-02-14
validApiKey() function refactor

mkuzdowicz:
## What does this change?
validApiKey() function refactor

## Tested?
- [ ] locally
- [x] on TEST


admin-tools/lib/src/main/scala/com/gu/mediaservice/ImagesBatchProjection.scala
```diff
@@ -11,7 +11,7 @@ class ImagesBatchProjection(apiKey: String, domainRoot: String, timeout: Duratio
 
   def validApiKey(projectionEndpoint: String) = {
     val projectionUrl = new URL(s"$projectionEndpoint/not-exists")
-    gridClient.makeGetRequestSync(projectionUrl, apiKey).statusCode == 404
+    gridClient.makeGetRequestSync(projectionUrl, apiKey).statusCode != 401
```
akash1810: Could we do `!= 200` to catch all the `4XX` cases?

akash1810: Oh, I've just looked at the URL we're requesting, and 404 makes more sense!

PR #2794, opened 2020-02-14
A retry mechanism for thrall.

AWare:
just this in a new pr https://github.com/guardian/grid/pull/2746

### DIFF
--- a/thrall/app/lib/RetryHandler.scala
+++ b/thrall/app/lib/RetryHandler.scala
@@ -1,0 +1,60 @@
+package lib
+
+
+import akka.actor.ActorSystem
+import akka.pattern.{after, retry}
+import play.api.{Logger, MarkerContext}
+
+import scala.concurrent.{ExecutionContext, Future, TimeoutException}
+import scala.concurrent.duration.FiniteDuration
+import scala.util.{Failure, Success}
+
+object RetryHandler {
+  def handleWithRetryAndTimeout[T](f: () => Future[T],
+                                   retries: Int,
+                                   timeout: FiniteDuration,
+                                   delay: FiniteDuration
+                                  )(implicit actorSystem: ActorSystem,
+                                    executionContext: ExecutionContext,
+                                    mc: MarkerContext
+                                  ): () => Future[T] = {
+    def logFailures[T](f: () => Future[T])(
+      implicit executionContext: ExecutionContext, mc: MarkerContext
+    ): () => Future[T] = {
+      () => {
+        f().transform {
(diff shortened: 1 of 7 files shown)


thrall/app/lib/RetryHandler.scala
```diff
@@ -0,0 +1,60 @@
...
+      var count = 0
+
+      def attempt = () => {
+        count = count + 1
+        Logger.info(s"Attempt $count of $retries")
+        f()
+      }
+      retry(attempt, retries, delay)
+    }
+    println("HELLO I AM DOING SOMETHING RETRY RELATED")
```
jonathonherbert:
```suggestion
```

thrall/app/lib/RetryHandler.scala
```diff
@@ -0,0 +1,59 @@
...
+      Future.firstCompletedOf(Seq(timeout, f()))
+    }
+
+    def handleWithRetry[T](f: () => Future[T], retries: Int, delay: FiniteDuration): () => Future[T] = () => {
+      implicit val scheduler = actorSystem.scheduler
+      var count = 0
+
+      def attempt = () => {
+        count = count + 1
+        Logger.info(s"Attempt $count of $retries")
```
akash1810: Would be good to add a bit more context to this log line to include the message subject, or the markers of the `UpdateMessage`. This way we can gain more insight into what gets retried for further investigation.

thrall/test/lib/elasticsearch/ElasticSearchTest.scala
```diff
@@ -29,9 +29,6 @@ class ElasticSearchTest extends ElasticSearchTestBase {
 
           val images: List[Image] = List(imageOne, imageTwo)
 
-          // in a clean index, we should have 0 documents
```
akash1810: If this causes the tests to pass in GHA and the test on line 41 still passes:

```scala
// after bulk inserting, we should have 2 documents
ES.client.execute(ElasticDsl.count(ES.initialImagesIndex)).await.result.count shouldBe images.length
```

It might suggest that force refreshing between tests might be a good thing to do?

thrall/app/lib/RetryHandler.scala
AWare: The markercontext is passed in as an implicit, and these fields are present in kibana.

thrall/test/lib/elasticsearch/ElasticSearchTest.scala
AWare: Yeah, that's a good idea.

PR #2800, opened 2020-02-18
making image projection-atomic

mkuzdowicz:
##  What does this change?
making image projection more atomic
in projection API if any downstream api is in a bad state throw exception

## Tested?
- [x] locally
- [x] on TEST


admin-tools/lib/src/main/scala/com/gu/mediaservice/ImageDataMerger.scala
```diff
@@ -174,8 +187,22 @@ class ImageDataMerger(config: ImageDataMergerConfig, gridClient: GridClient) ext
...
+  private def validateResponse(res: ResponseWrapper, url: URL): Unit = {
+    import res._
+    if (statusCode != 200 && statusCode != 404) {
+      val message = s"breaking the circuit, downstream API: $url is in a bad state, code: $statusCode"
+      val errorJson = Json.obj(
+        "errorStatusCode" -> statusCode,
+        "message" -> "message"
+      )
+      logger.error(errorJson.toString())
+      throw new IllegalArgumentException(message)
```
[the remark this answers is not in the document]
mkuzdowicz: the exception is like a circuit breaker here it is suppose to stop the projection and return 500 or 502 i think this is what API gateway will return

If the image projection will return anything that is not 200 or 404
the image ingestion process will skip that projection, log error and process it at later time

admin-tools/lib/src/main/scala/com/gu/mediaservice/ImageDataMerger.scala
```diff
@@ -14,6 +14,14 @@ import play.api.libs.json._
 
 import scala.concurrent.{ExecutionContext, Future, Promise}
 
+object ImageDataMergerConfig {
+  private val gridClient = GridClient(5)
```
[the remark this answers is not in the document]
mkuzdowicz: done

admin-tools/lib/src/main/scala/com/gu/mediaservice/GridClient.scala
```diff
@@ -0,0 +1,82 @@
...
+          "message" -> s"http-failure: ${response.message()}"
+        )
+        logger.error(errorJson.toString())
+      }
+      val json = if (code == 200) Json.parse(body.string) else Json.obj()
+      response.close()
+      ResponseWrapper(json, code)
+    } catch {
+      case e: Exception =>
+        // propagating exception
```
akash1810: should we log here?

admin-tools/lib/src/main/scala/com/gu/mediaservice/ImageDataMerger.scala
```diff
@@ -151,7 +87,7 @@ object ImageMetadataOverrides extends LazyLogging {
 
 }
```
mkuzdowicz: that is the main change, the rest is refactoring

admin-tools/lib/src/main/scala/com/gu/mediaservice/GridClient.scala
mkuzdowicz: done

PR #2802, opened 2020-02-18
add a cheat to the bulk insert test

akash1810:
## What does this change?
The issue raised in https://github.com/guardian/grid/pull/2785 is still being witnessed. This is a cheating solution.

Ideally, before we bulk insert we would assert document count is 0 and after, it is 2. However there is a race condition in the tests (only seen in GH Actions) which results in state being shared between tests, so we're just checking the document count has increased by the correct amount.

More reliable CI.

n/a

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/thrall/test/lib/elasticsearch/ElasticSearchTest.scala
+++ b/thrall/test/lib/elasticsearch/ElasticSearchTest.scala
@@ -29,13 +29,19 @@ class ElasticSearchTest extends ElasticSearchTestBase {
 
           val images: List[Image] = List(imageOne, imageTwo)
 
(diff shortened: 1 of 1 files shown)


AWare:
Is the count a useful test/did refreshing the index help?

akash1810:
> Is the count a useful test/did refreshing the index help?

Count is essential for this test as we're bulk inserting.

Not tried refreshing yet. Was going to do this as a separate PR.

akash1810:
Merging to unblock @mkuzdowicz . Working on refreshing in `afterEach` step in parallel.

PR #2808, opened 2020-02-20
Add subdomain-health-lambda

stephengeller:
## What does this change?
This adds a lambda that pushes the current healthcheck status code of the subdomains in the Grid to Cloudwatch

Seeing if the metrics show up in Cloudwatch, which they do.

image
This can be used for status dashboards:
subdomain-health

## Tested?
- [X] locally
- [X] on TEST


akash1810:
This feels like it would be useful as a [standalone repo](https://github.com/organizations/guardian/repositories/new) with its cloudformation co-located with it.

PR #2810, opened 2020-02-20
Image batch index record known issues state

mkuzdowicz:
### What
- Image batch index record `known issues state` on images that failed to project because of know issues

### Why
- to avoid to pick up them again on next batch

## Tested?
- [x] locally
- [x] on TEST


admin-tools/lib/src/main/scala/com/gu/mediaservice/indexing/package.scala
```diff
@@ -11,6 +11,7 @@ package object indexing {
     val NotFound = ProduceProgress("not found", 2)
     val Finished = ProduceProgress("finished", 3)
     val Reset = ProduceProgress("reset because of failure", 0)
+    val KnownError = ProduceProgress("reset because of failure", 4)
```
akash1810:
```suggestion
    val KnownError = ProduceProgress("reset because of known failure", 4)
```

mkuzdowicz: done

PR #2816, opened 2020-02-21
log when Content-Length header isn't present

akash1810:
## What does this change?
The alarm `Sum FailedUploads GreaterThanOrEqualToThreshold 50.0` is tripping fairly often. The logs have some suspicious entries around the time they trip:

```
java.lang.NumberFormatException: null
	at java.lang.Integer.parseInt(Integer.java:542)
	at java.lang.Integer.parseInt(Integer.java:615)
	at scala.collection.immutable.StringLike.toInt(StringLike.scala:304)
	at scala.collection.immutable.StringLike.toInt$(StringLike.scala:304)
	at scala.collection.immutable.StringOps.toInt(StringOps.scala:33)
	at lib.Downloader.$anonfun$download$1(Downloader.scala:25)
	at scala.concurrent.Future$.$anonfun$apply$1(Future.scala:658)
	at scala.util.Success.$anonfun$map$1(Try.scala:255)
	at scala.util.Success.map(Try.scala:213)
	at scala.concurrent.Future.$anonfun$map$1(Future.scala:292)
	at scala.concurrent.impl.Promise.liftedTree1$1(Promise.scala:33)
	at scala.concurrent.impl.Promise.$anonfun$transform$1(Promise.scala:33)
	at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
	at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:55)
	at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:91)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
	at scala.concurrent.BlockContext$.withBlockContext(BlockContext.scala:85)
	at akka.dispatch.BatchingExecutor$BlockableBatch.run(BatchingExecutor.scala:91)
	at akka.dispatch.TaskInvocation.run(AbstractDispatcher.scala:40)
	at akka.dispatch.ForkJoinExecutorConfigurator$AkkaForkJoinTask.exec(ForkJoinExecutorConfigurator.scala:44)
	at akka.dispatch.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
	at akka.dispatch.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
	at akka.dispatch.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
[...]

PR #2820, opened 2020-02-24
handle image download failure better

akash1810:
## What does this change?
Log the exception thrown.

Follows https://github.com/guardian/grid/pull/2816

n/a

n/a

## Who should look at this?

@jennygrahamjones 

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/image-loader/app/lib/Downloader.scala
+++ b/image-loader/app/lib/Downloader.scala
@@ -11,7 +11,7 @@ import okhttp3.{OkHttpClient, Request}
 import play.api.Logger
 
 import scala.concurrent.{ExecutionContext, Future}
-import scala.util.Try
+import scala.util.{Failure, Success, Try}
 
 case object TruncatedDownload extends Exception
 case object InvalidDownload extends Exception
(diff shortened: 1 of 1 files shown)


image-loader/app/lib/Downloader.scala
```diff
@@ -25,14 +25,14 @@ class Downloader(implicit ec: ExecutionContext) {
...
     val response = client.newCall(request).execute()
 
-    val maybeExpectedSize = Try{response.header("Content-Length").toInt}.toOption
+    val maybeExpectedSize = Try{response.header("Content-Length").toInt}
 
     maybeExpectedSize match {
-      case None => {
-        Logger.error(s"Missing content-length header from $uri")
+      case Failure(exception) => {
+        Logger.error(s"Missing content-length header from $uri", exception)
```
jennygrahamjones: Do you need something like `.message` or `.getMessage` on the exception?

akash1810: The type signature suggests this is possible. I imagine this allows the log line to include the stack trace.

PR #2823, opened 2020-02-24
Introduce GC logging and file rotation

sihil:
this won't take more than 10Mb of disk space so should be safe everywhere

## What does this change?
We are currently unable to see GC data from grid instances and this turns on logging for that.

## Tested?
- [ ] locally
- [x] on TEST


sihil:
Nope. For now they are not shipped, we're still trying to figure out why image loader dies.

PR #2824, opened 2020-02-24
thin out Image Loader Controller a bit more

akash1810:
## What does this change?
Co-locating file operations for better organisation. I suspect we don't need both a sync and async version of temp file creation, however I couldn't get the compiler to like me, so doing this in the mean time.

Thin controllers are easier to read.

n/a

## Tested?
- [ ] locally
- [ ] on TEST


akash1810:
I think we can take a different approach to this. Closing.

PR #2829, opened 2020-02-25
ability to pass isolated endpoint for image-loader for projection

mkuzdowicz:
## What does this change?
- ability to pass isolated endpoint for image-loader for projection to not to disrupt usual grid usage
- when calling projection endpoint we want to be able to switch to a isolated image-loader endpoint

when calling projection endpoint we call different instance of image-loader  the the that is used by most users

## Tested?
- [x] locally
- [x] on TEST

### DIFF
--- a/admin-tools/lib/src/main/scala/com/gu/mediaservice/ImageDataMerger.scala
+++ b/admin-tools/lib/src/main/scala/com/gu/mediaservice/ImageDataMerger.scala
@@ -16,13 +16,17 @@ import scala.concurrent.{Await, ExecutionContext, Future}
 object ImageDataMergerConfig {
   private val gridClient = GridClient(maxIdleConnections = 5)
 
-  def apply(apiKey: String, domainRoot: String): ImageDataMergerConfig = {
+  def apply(apiKey: String, domainRoot: String, imageLoaderEndpointOpt: Option[String]): ImageDataMergerConfig = {
     val services = new Services(domainRoot, ServiceHosts.guardianPrefixes, Set.empty)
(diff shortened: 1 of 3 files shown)


admin-tools/lib/src/main/scala/com/gu/mediaservice/ImageDataMerger.scala
```diff
@@ -16,13 +16,17 @@ import scala.concurrent.{Await, ExecutionContext, Future}
 object ImageDataMergerConfig {
   private val gridClient = GridClient(maxIdleConnections = 5)
 
-  def apply(apiKey: String, domainRoot: String): ImageDataMergerConfig = {
+  def apply(apiKey: String, domainRoot: String, imageLoaderEndpointOpt: Option[String]): ImageDataMergerConfig = {
```
akash1810: why Option?

mkuzdowicz: if not passed we use the one from Services class look at: https://github.com/guardian/grid/blob/d83b3e370b5c3e5fab53b62130467ec2bc1ecfa4/admin-tools/dev/app/controllers/AdminToolsCtr.scala#L16

i can change the var name `imageLoaderEndpointOpt` -> `alternativeImageLoaderEndpointOpt`

maybe it will make more sense then

mkuzdowicz: i can make it string and treat as restricted to Pass if you prefer ?

PR #2832, opened 2020-02-26
 dont update phantom usages

AWare:
## What does this change?

ElasticSearch queries can now return typed errors for es failures, exposing more detail about what went wrong. Now for updates, we can catch these specific errors and not cause a retry.

Able to take usages for non-existent images without crashing.

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -9,7 +9,8 @@ val commonSettings = Seq(
   version := "0.1",
   scalacOptions ++= Seq("-feature", "-deprecation", "-language:higherKinds", "-Xfatal-warnings"),
   libraryDependencies ++= Seq(
-    "org.scalatest" %% "scalatest" % "3.0.5"
+    "org.scalatest" %% "scalatest" % "3.0.5",
+    "org.mockito" % "mockito-core" % "2.18.0"
   )
 )
 
@@ -83,8 +84,8 @@ lazy val commonLib = project("common-lib").settings(
     // needed to parse conditional statements in `logback.xml`
     // i.e. to only log to disk in DEV
     // see: https://logback.qos.ch/setup.html#janino
-    "org.codehaus.janino" % "janino" % "3.0.6"
-  ),
+    "org.codehaus.janino" % "janino" % "3.0.6",
+),
   dependencyOverrides += "org.apache.thrift" % "libthrift" % "0.9.1"
 )
 
@@ -114,7 +115,6 @@ lazy val mediaApi = playProject("media-api", 9001).settings(
     "org.apache.commons" % "commons-email" % "1.5",
     "org.parboiled" %% "parboiled" % "2.1.5",
     "org.http4s" %% "http4s-core" % "0.18.7",
-    "org.mockito" % "mockito-core" % "2.18.0",
     "com.softwaremill.quicklens" %% "quicklens" % "1.4.11",
     "com.whisk" %% "docker-testkit-scalatest" % "0.9.8" % Test,
     "com.whisk" %% "docker-testkit-impl-spotify" % "0.9.8" % Test
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchExecutions.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchExecutions.scala
@@ -1,7 +1,6 @@
 package com.gu.mediaservice.lib.elasticsearch
 
-import com.sksamuel.elastic4s.{ElasticClient, Executor, Functor, Handler, Response}
-
+import com.sksamuel.elastic4s.{ElasticClient, ElasticError, Executor, Functor, Handler, Response}
 import net.logstash.logback.marker.LogstashMarker
 import net.logstash.logback.marker.Markers.appendEntries
 import play.api.{Logger, MarkerContext}
@@ -26,7 +25,10 @@ trait ElasticSearchExecutions {
       case Success(r) =>
         r.isSuccess match {
           case true => Success(r)
-          case false => Failure(new RuntimeException("query response was not successful: " + r.error.reason))
+          case false => r.status match {
+            case 404 => Failure(ElasticNotFoundException)
+            case _ => Failure(ElasticException(r.error))
+          }
         }
       case Failure(f) => Failure(f)
     }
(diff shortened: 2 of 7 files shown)


thrall/app/lib/kinesis/MessageProcessor.scala
```diff
@@ -51,14 +53,19 @@ class MessageProcessor(es: ElasticSearch,
...
+
     withId(message) { id =>
       withUsageNotice(message) { usageNotice =>
         withLastModified(message) { lastModifed =>
           val usages = usageNotice.usageJson.as[Seq[Usage]]
-          Future.sequence(es.updateImageUsages(id, asJsLookup(usages), dateTimeAsJsLookup(lastModifed)))
+          //DANGER: Here be an entirely useless sequence and map
+          Future.sequence(es.updateImageUsages(id, asJsLookup(usages), dateTimeAsJsLookup(lastModifed)).map(_.recoverWith {
+            case ElasticNotFoundException => Future.successful(ElasticSearchUpdateResponse())
+          }))
```
sihil: This would be better expressed with a `traverse`

build.sbt
```diff
@@ -83,8 +83,10 @@ lazy val commonLib = project("common-lib").settings(
     // needed to parse conditional statements in `logback.xml`
     // i.e. to only log to disk in DEV
     // see: https://logback.qos.ch/setup.html#janino
-    "org.codehaus.janino" % "janino" % "3.0.6"
-  ),
+    "org.codehaus.janino" % "janino" % "3.0.6",
+    "org.mockito" % "mockito-core" % "2.18.0"
```
sihil: Shouldn't this be:
```suggestion
    "org.mockito" % "mockito-core" % "2.18.0" % Test
```

thrall/app/lib/kinesis/MessageProcessor.scala
AWare: thanks :)

build.sbt
AWare: This doesn't work for me, can I get a minute tomorrow to look at this please?

akash1810:
Just noticed E4S has a `NotFound404ResponseHandler` which might be of interest here. https://github.com/sksamuel/elastic4s/blob/master/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/ResponseHandler.scala#L63-L69

image-counter-lambda/artifact.json
```diff
@@ -0,0 +1,11 @@
+{
```
akash1810: Don't think this should be part of this PR?

AWare: Honestly no idea how it came back

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchExecutions.scala
```diff
@@ -40,7 +42,12 @@ trait ElasticSearchExecutions {
     result.failed.foreach { e =>
       val elapsed = System.currentTimeMillis() - start
       val markers = MarkerContext(durationMarker(elapsed))
-      Logger.error(s"$message - query failed after $elapsed ms: ${e.getMessage} cs: ${e.getCause}")(markers)
+      e match {
+        case ElasticNotFoundException => Logger.error(s"$message - query failed: Document not Found")(markers)
+        case ElasticException(error) => Logger.error(s"$message - query failed after: ${error.reason} type: ${error.`type`}")(markers)
```
akash1810:
```suggestion
        case ElasticException(error) => Logger.error(s"$message - query failed because: ${error.reason} type: ${error.`type`}")(markers)
```

thrall/test/lib/elasticsearch/ElasticSearchTestBase.scala
```diff
@@ -1,14 +1,19 @@
 package lib.elasticsearch
 
-import com.gu.mediaservice.lib.elasticsearch.{ElasticSearchConfig, Mappings}
+import com.gu.mediaservice.lib.elasticsearch.ElasticSearchConfig
+import com.sksamuel.elastic4s.ElasticDsl._
+import com.sksamuel.elastic4s.ElasticDsl
+
+
+// import com.sksamuel.elastic4s.ElasticDsl._  This brings in the handler. Don't optimise imports.
```
akash1810:
```suggestion

```

thrall/test/lib/kinesis/MessageProcessorTest.scala
```diff
@@ -0,0 +1,89 @@
...
+            leaseId = None,
+            crops = None,
+            mediaLease = None,
+            leases = None,
+            syndicationRights = None,
+            bulkIndexRequest = None
+          )
+          noException should be thrownBy Await.result(messageProcessor.updateImageUsages(message), fiveSeconds)
+        }
+        "not crash for an image that doesn't exist 👻🖼" in {
```
akash1810: I'm not sure this test is exhaustive. IIUC this PR is designed to fail fast on a 404, that is, do not retry the message more than once. This test is only asserting that an exception isn't thrown. Could we add a test against the retry behaviour in a 404 scenario too please?

AWare: Adding tests to the retry handler should be its own pr.

akash1810: Ok. How can we determine success of this change? The current structure may make adding this type of test obscure; can we add detailed logging which we can observe? Logging the attempt number with the subject, for example? At the moment, it's not clear that this PR solves the issue we've witnessed where a message gets retried and blocks the remaining messages.

AWare: We found one of the root causes being retries on missing image usages. This makes those changes return a success instead of a failure state- and tests for this. 

This PR also makes that information available in the response from executeAndLog and increases of the specificity of logging. 

Extra work is needed in:
- some tests for the retry handler
- better handling of logmarkers to make this more explicit 
- using the api change in this PR to solve the syndication form of this bug

I don't think this PR is enough to fix the issue, but imo everything that is, has been blocked by this PR for the past two weeks.

aug24:
I would very much like to get this into PROD even without tests.
As we didn't have tests on this behaviour for the previous version, I'm not going to beat Alex up too much.
We have live users being affected by this problem and we need more information to diagnose the cause ASAP.

akash1810:
I'm unsure how we can measure success of this; having data in the form of logs or metrics or having more test coverage would solve this. I'm conscious that there has been disruption to the picture desk a few times recently and I'd feel more comfortable if we are able to empirically know that this change is helping. Happy to pair.

thrall/test/lib/kinesis/MessageProcessorTest.scala
akash1810: > This makes those changes return a success instead of a failure state- and tests for this.

Can this test test for that? As in, rather than asserting no exception was thrown, can we assert on the return type? IIRC, we retry iff return isn't `Success` (and not timeout). Asserting for `Success` would provide more confidence that the message won't get retried.

AWare: Tests modified to check for Success.

PR #2835, opened 2020-02-26
exectute and log as either

AWare:
## What does this change?

This makes executeAndLog return an either.

In looking at the usages for non-existent images work, we can avoid explicit checks if we know why the query failed. And it stands to some reason that there are other places that can/will benefit from this knowledge.

Very draft, but just to see if the refactor is a good idea.

Especially interested if we can find a good type signature for the `Left`

## Tested?
- [ ] locally
- [ ] on TEST


AWare:
This is based on https://github.com/guardian/grid/pull/2832

PR #2838, opened 2020-02-27
Mkuzdowicz/ DynamoDB scripts use-full while Re-ingestion

mkuzdowicz:
## What does this change?
adding  DynamoDB scripts use-full while Re-ingestion
- reset state of images to re-ingest in dynamodb
- get stats about re-ingestion `pregress_state`
- reset known errors

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/admin-tools/scripts/src/main/scala/com/gu/mediaservice/ImagesGroupByProgressState.scala
+++ b/admin-tools/scripts/src/main/scala/com/gu/mediaservice/ImagesGroupByProgressState.scala
@@ -1,0 +1,61 @@
+package com.gu.mediaservice
+
+import com.gu.mediaservice.indexing.{IndexInputCreation, ProduceProgress}
+import com.typesafe.scalalogging.LazyLogging
+
+import scala.collection.JavaConverters._
+
+
+object ImagesGroupByProgressState extends App with LazyLogging {
+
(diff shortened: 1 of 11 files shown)


admin-tools/scripts/src/main/scala/com/gu/mediaservice/ImagesGroupByProgressState.scala
```diff
@@ -0,0 +1,58 @@
...
+    def stateNameToCount(progressType: ProduceProgress) = {
+      logger.info(s"calculating stateNameToCount for $progressType")
+      val result = stateIndex.query(getAllMediaIdsWithinStateQuery(progressType.stateId))
+      val iterator = result.iterator()
+      var count: Long = 0
+      while (iterator.hasNext) {
+        count += 1
+        iterator.next()
+      }
+      val ans = progressType.name -> count
```
sihil: This is quite Java-esque

```suggestion
      val ans = progressType.name -> result.iterator.asScala.length
```
seems better? (You'll also need to `import scala.collection.JavaConverters._`)

mkuzdowicz: god point, looks much better now

PR #2840, opened 2020-02-28
re-ingestion adjusting timeouts

mkuzdowicz:
## What does this change?
re-ingestion adjusting timeouts

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/admin-tools/lib/src/main/scala/com/gu/mediaservice/BatchIndexHandler.scala
+++ b/admin-tools/lib/src/main/scala/com/gu/mediaservice/BatchIndexHandler.scala
@@ -34,14 +34,13 @@ class BatchIndexHandler(cfg: BatchIndexHandlerConfig) extends LoggingWithMarkers
 
   import cfg._
 
-  private val ProjectionTimoutInMins = 11
-  private val GetIdsTimoutInMins = 1
-  private val OthersTimoutInMins = 1
-  private val MainProcessingTimoutInMins = ProjectionTimoutInMins + OthersTimoutInMins
+  private val ProjectionTimoutInSec = 740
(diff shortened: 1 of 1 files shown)


admin-tools/lib/src/main/scala/com/gu/mediaservice/BatchIndexHandler.scala
```diff
@@ -34,14 +34,13 @@ class BatchIndexHandler(cfg: BatchIndexHandlerConfig) extends LoggingWithMarkers
...
 
-  private val ProjectionTimoutInMins = 11
-  private val GetIdsTimoutInMins = 1
-  private val OthersTimoutInMins = 1
-  private val MainProcessingTimoutInMins = ProjectionTimoutInMins + OthersTimoutInMins
-
-  private val GetIdsTimeout = new FiniteDuration(GetIdsTimoutInMins, TimeUnit.MINUTES)
-  private val GlobalTimeout = new FiniteDuration(MainProcessingTimoutInMins, TimeUnit.MINUTES)
-  private val ImagesProjectionTimeout = new FiniteDuration(ProjectionTimoutInMins, TimeUnit.MINUTES)
+  private val ProjectionTimoutInSec = 740
```
akash1810: Could we express this as a multiple of 60 to make it easier to determine it's scale? For example, it is easier to determine that `5 * 60` is 5 minutes vs `300`.

akash1810: Just realised 740 isn't easy to express as a factor of 60. Let's get this PR in as it improves what's there and we can refactor more later if necessary.

PR #2842, opened 2020-02-28
change signature of loadImage to use stricter types

akash1810:
## What does this change?
Rather than accepting an `Option[String]` as a param and then converting it to a DateTime in the function body, have a stricter signature.

This is part of a refactor to make the ImageLoaderController easier to follow. The ambition is to make the controller a transforming layer, going from an http request into a strictly typed set of objects that can be pushed into other functions.

Also makes `loadImage` private, because it is!

Simpler code.

n/a

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/image-loader/app/controllers/ImageLoaderController.scala
+++ b/image-loader/app/controllers/ImageLoaderController.scala
@@ -176,7 +176,7 @@ class ImageLoaderController(auth: Authentication, downloader: Downloader, store:
     }
   }
 
-  def loadFile(digestedFile: DigestedFile, user: Principal,
+  private def loadFile(digestedFile: DigestedFile, user: Principal,
                uploadedBy: Option[String], identifiers: Option[String],
                uploadTime: Option[String], filename: Option[String], requestLoggingContext: RequestLoggingContext): Future[Result] = {
     val DigestedFile(tempFile_, id_) = digestedFile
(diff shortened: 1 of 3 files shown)


image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -176,9 +176,11 @@ class ImageLoaderController(auth: Authentication, downloader: Downloader, store:
     }
   }
 
-  def loadFile(digestedFile: DigestedFile, user: Principal,
+  private def dateTimeFromValueOrNow(value: Option[String]): DateTime = Option{new DateTime(value)}.getOrElse(DateTime.now)
```
AWare: This use of option is new to me. Am I right in saying there's some kind of implicit `Option[String] => String` happening and the braces are handling any exceptions in some way?

akash1810: Sorry, this should be a `Try`. `Option{foo}` would return a `Some`; it doesn't have any exception handling. ([source](https://github.com/scala/scala/blob/v2.12.8/src/library/scala/Option.scala#L29)).

akash1810: Oof actually, this PR still isn't correct. I'll write some tests.

PR #2844, opened 2020-03-02
Upversion Elastic4S-http

jonathonherbert:
## What does this change?

Bump the version of elastic4s to a version which supports elasticsearch 7.

If media-api et al can parse elasticsearch responses when a node running 7.5.2 is in the cluster, then this has succeeded.

N/A

## Who should look at this?

Anyone.

## Tested?
- [X] locally
- [X] on TEST


.java-version
```diff
@@ -0,0 +1 @@
+1.8
```
akash1810: Why do we need a specific java version now?

aug24: Because it doesn't work against the latest.  I'll raise a ticket for investigating that.

PR #2847, opened 2020-03-06
Update grid imaging ami to latest ubuntu (not ESR)

AWare:
## What does this change?
This updates Ubuntu to Ubuntu 19.10 (Eoan Ermine) which brings in the GraphicsMagick package 1.4+really1.3.33.

Not the most reassuring package name, but it fixes the bug we've been experiencing with 1.3.28 in Bionic.

Fewer image load errors in the reingest. 

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/riff-raff.yaml
+++ b/riff-raff.yaml
@@ -91,7 +91,7 @@ deployments:
         ImagingAmiId:
           BuiltBy: amigo
           AmigoStage: PROD
-          Recipe: grid-imaging
+          Recipe: grid-imaging-eoan
         ImgOpsAmiId:
           BuiltBy: amigo
           AmigoStage: PROD


AWare:
@paperboyo suggests merging this just before we start reingest

paperboyo:
> suggests merging this just before we start reingest

1. Only listen to him if you agree.
2. But after sorting out other Thrall hiccups.
3. And then, stop and revert as soon as we are seeing errors we’ve never seen before (hopefully won’t happen)

My tests suggests it will fix `Improper call to JPEG library in state 201` which constitutes 99.8% of all errors in most recent reingestion, while may introduce just one, very rare, error we aren’t currently seeing (`Unable to manage color`) from which it would be nice to recover (just strip the dud profile similarly to [what we do here](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala#L43-L55)).

riff-raff.yaml
```diff
@@ -91,7 +91,7 @@ deployments:
         ImagingAmiId:
           BuiltBy: amigo
           AmigoStage: PROD
-          Recipe: grid-imaging
+          Recipe: grid-imaging-eoan
```
akash1810: Can I suggest avoiding adding distro names to the AMIgo recipe as it makes the next upgrade harder:
- Find current recipe
- Clone settings
- Bake
- Raise PR
- Delete old recipe

An alternative would be to update the base image on the `grid-imaging` recipe and bake.

AWare: This will need upgrading at the end of April when the next LTS is released. At that point I'll update grid-imaging to then new LTS and delete this recipe. I want to keep both recipes around until we're confident that the error rate changes in the new graphicsmagick (discussed in planning) doesn't effect us.

PR #2848, opened 2020-03-06
Better log error messages for the batch reingest lambda

jonathonherbert:
## What does this change?

This adds better error message logging to the batch reingest lambda.

Beforehand, we just logged everything we received from the image projection response body. Here, we extract error messages and log them as separate keys, to help us visualise error types and their frequencies.

We should see the `errorMessage` and `downstreamErrorMessage` fields populated in logs.

## Tested?
- [x] locally
- [ ] on TEST


jonathonherbert:
Good points both – I've just amended.

PR #2849, opened 2020-03-06
rename elasticsearch6 dir to elasticsearch

akash1810:
## What does this change?
This matches the path used in [docker-compose.yml](https://github.com/guardian/grid/blob/master/docker-compose.yml#L6) and allows the local ES server to start (previously docker was failing due to file not found).

Local dev works again.

n/a

## Who should look at this?

@paperboyo 

## Tested?
- [ ] locally
- [ ] on TEST


paperboyo:
👍

PR #2850, opened 2020-03-06
Test Kinesis GetRecords.IteratorAgeMilliseconds threshold

aug24:
## What does this change?
This changes adds a test that the configured Kinesis stream for this stage is currently returning all low values for GetRecords.IteratorAgeMilliseconds

If the kinesis stream runs slowly, lambda will log a refusal to process images until the stream runs quickly again.

## Who should look at this?
All of Ed Tools

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/admin-tools/lib/src/main/scala/com/gu/mediaservice/BatchIndexHandler.scala
+++ b/admin-tools/lib/src/main/scala/com/gu/mediaservice/BatchIndexHandler.scala
@@ -25,7 +25,10 @@ case class BatchIndexHandlerConfig(
                                     dynamoTableName: String,
                                     batchSize: Int,
                                     kinesisEndpoint: Option[String] = None,
-                                    maxIdleConnections: Int
+                                    kinesisMaximumMetric: Option[(String, Integer)] = None,
+                                    maxIdleConnections: Int,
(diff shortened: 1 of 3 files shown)


admin-tools/lib/src/main/scala/com/gu/mediaservice/BatchIndexHandler.scala
```diff
@@ -57,6 +60,14 @@ class BatchIndexHandler(cfg: BatchIndexHandlerConfig) extends LoggingWithMarkers
   private implicit val ec: ExecutionContext = scala.concurrent.ExecutionContext.Implicits.global
 
   def processImages(): Unit = {
+
+    if (checkKinesisIsNiceAndFast)
+      processImagesOnlyIfKinesisIsNiceAndFast()
```
jonathonherbert: Couldn't be happier with these names 😁

AWare:
todo:

- [ ] add LATENCY_THRESHOLD	 PROJECTION_ENDPOINT to stack
- [ ] add iam policy allowing us to read CloudWatch

AWare:
Stack changes made. 👌

PR #2851, opened 2020-03-09
Remove Lorna Roach…

paperboyo:
…from Staff list.


paperboyo:
Closing as I think this would invalidate all past content. We don’t have a mechanism to do it sensibly (eg batch edit all past content to something free). Closing.

PR #2854, opened 2020-03-10
Improve Elasticsearch error logging

aug24:
## What does this change?
Better logging of the root cause of ES failures.

Useful information being available in ELK

## Tested?
- [ ] locally
- [ ] on TEST


AWare:
There's a bit of duplication with https://github.com/guardian/grid/pull/2832 here

PR #2856, opened 2020-03-11
Guard against null original usage

aug24:
## What does this change?
This guards against the following scenario: If there is no original usage, the .clone() call will cause a failure.  

We are not certain that this error cause is ever hit, but there are various "cs: null" errors.  This prevents one possible cause.

N/A

## Who should look at this?
Anyone.

## Tested?
- [ ] locally
- [ ] on TEST


sihil:
@aug24 Any chance of adding a test for this and merging?

aug24:
After further investigation, it is impossible to write a test for this because the code does not permit getting into the 'bad' state.  Therefore I think this can not be a fix to the observed problem.

Closing.

PR #2857, opened 2020-03-11
more detail in log markers in Thrall

AWare:
## What does this change?

Every place where previously we'd created a new log marker in thrall, I've created a new `MarkerContext` and passed that down. 

We create new `MarkerContext`s from an existing (explicit) `MarkerContext` and a `Map[String,Any]`- use the map to create a brand new Marker, add any Marker present in the existing context and then create and return a new `MarkerContext`. 

We've agreed that this should be as explicit as possible, so I'm going through and making it so.

I think it's currently functional, but I'm going to go and make it explicit when it creates the initial context from the message. 

Where we have a function which doesn't do anything to a `MarkerContext` I've just left it implicit. 
 
More details in logs!

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchException.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchException.scala
@@ -1,0 +1,40 @@
+package com.gu.mediaservice.lib.elasticsearch
+
+import com.gu.mediaservice.lib.logging.{LogMarker, MarkerMap}
+import com.sksamuel.elastic4s.ElasticError
+import net.logstash.logback.marker.LogstashMarker
+
+trait ElasticSearchError {
+  self: Throwable =>
+  def error: ElasticError
+
+  def markerContents: Map[String, Any]
+}
+
+object ElasticSearchException {
+  def apply(e: ElasticError): Exception with ElasticSearchError = {
+    e match {
+      case ElasticError(t, r, _, _, _, Seq(), _, _, _, _) => // No root causes provided.
+        new Exception(s"query failed because: $r type: $t") with ElasticSearchError {
+          override def error: ElasticError = e
+
+          override def markerContents: Map[String, Any] = Map("reason" -> r, "type" -> t)
+        }
+      case ElasticError(t, r, _, _, _, s, _, _, _, _) =>
+        new Exception(s"query failed because: $r type: $t root cause ${s.mkString(", ")}") with ElasticSearchError {
+          override def error: ElasticError = e
+
+          override def markerContents: Map[String, Any] = Map("reason" -> r, "type" -> t, "rootCause" -> s.mkString(", "))
+        }
+      case _ => new Exception(s"query failed because: unknown error") with ElasticSearchError {
+        override def error: ElasticError = e
+
+        override def markerContents: Map[String, Any] = Map("reason" -> "unknown Elastic Search error")
+      }
+    }
+  }
+
+  def unapply(arg: ElasticSearchError): Option[(ElasticError, LogMarker)] = Some((arg.error, MarkerMap(arg.markerContents)))
+}
+
+case object ElasticNotFoundException extends Exception(s"Elastic Search Document Not Found")
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/logging/MarkerUtils.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/logging/MarkerUtils.scala
@@ -2,15 +2,28 @@ package com.gu.mediaservice.lib.logging
 
 import net.logstash.logback.marker.LogstashMarker
 import net.logstash.logback.marker.Markers.appendEntries
+import play.api.MarkerContext
+import scala.language.implicitConversions
+
 import scala.collection.JavaConverters._
 
-trait LogMarker { def toLogMarker: LogstashMarker }
+trait LogMarker {
+  def toLogMarker: LogstashMarker
+  def markerContents: Map[String, Any]
+}
 
-case class MarkerMap(values: Map[String, Any]) extends LogMarker {
-  override def toLogMarker: LogstashMarker = appendEntries(values.asJava)
+
+
+case class MarkerMap(markerContents: Map[String, Any]) extends LogMarker {
+  override def toLogMarker: LogstashMarker = appendEntries(markerContents.asJava)
+}
+
+object MarkerMap {
+  def apply(entries: (String, Any)*):MarkerMap = MarkerMap(entries.toMap)
 }
 
 trait MarkerUtils {
   val FALLBACK: String = "unknown"
-  def combineMarkers(markers: LogMarker*): LogstashMarker = markers.map(_.toLogMarker).reduce((first, second) => first.and[LogstashMarker](second))
+  def combineMarkers(markers: LogMarker*): LogMarker = MarkerMap(markers.flatMap(_.markerContents.toSeq).toMap)
+  implicit def fromLogMarker(logMarker: LogMarker):MarkerContext = MarkerContext(logMarker.toLogMarker)
 }
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/logging/MarkerUtils.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/logging/MarkerUtils.scala
@@ -8,15 +8,11 @@ import scala.language.implicitConversions
 import scala.collection.JavaConverters._
 
 trait LogMarker {
-  def toLogMarker: LogstashMarker
+  def toLogMarker: LogstashMarker = appendEntries(markerContents.asJava)
   def markerContents: Map[String, Any]
 }
 
-
-
-case class MarkerMap(markerContents: Map[String, Any]) extends LogMarker {
-  override def toLogMarker: LogstashMarker = appendEntries(markerContents.asJava)
-}
+case class MarkerMap(markerContents: Map[String, Any]) extends LogMarker
 
 object MarkerMap {
   def apply(entries: (String, Any)*):MarkerMap = MarkerMap(entries.toMap)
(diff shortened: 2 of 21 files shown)


thrall/app/lib/RetryHandler.scala
```diff
@@ -43,12 +44,13 @@ object RetryHandler {
...
     }
 
-    def handleWithRetry[T](f: () => Future[T], retries: Int, delay: FiniteDuration): () => Future[T] = () => {
+    def handleWithRetry[T](f: () => Future[T], retries: Int, delay: FiniteDuration)(implicit markerContext: MarkerContext): () => Future[T] = () => {
       implicit val scheduler = actorSystem.scheduler
       var count = 0
 
       def attempt = () => {
         count = count + 1
+        MarkerAugmentation.augmentMarkerContext(Map("attempt" -> count))
```
jonathonherbert: Could we be more explicit here?

```suggestion
        MarkerAugmentation.augmentMarkerContext(Map("retryCount" -> count))
```

thrall/app/lib/RetryHandler.scala
```diff
@@ -3,24 +3,27 @@ package lib
...
 import akka.pattern.{after, retry}
+import com.gu.mediaservice.lib.logging.MarkerAugmentation
 import play.api.{Logger, MarkerContext}
 
 import scala.concurrent.{ExecutionContext, Future, TimeoutException}
 import scala.concurrent.duration.FiniteDuration
 import scala.util.{Failure, Success}
 
 object RetryHandler {
+  def withExplicitMarkerContext[T](f: () => Future[T], mc: MarkerContext): Future[T] = {
```
AWare: This doesn't give me good feelings.

common-lib/src/main/scala/com/gu/mediaservice/lib/logging/MarkerAugmentation.scala
```diff
@@ -0,0 +1,25 @@
...
+  /** *
+    *
+    * @param markerContext
+    * @param augmentation : more details!
+    * @return a new markercontext!
+    *         it wouldn't be java if it could be cloned
+    */
+  def augmentMarkerContext(markerContext: MarkerContext, augmentation: Map[String, Any]): MarkerContext = {
+    val newMarker = appendEntries(augmentation.asJava)
+    markerContext.marker.foreach(newMarker.add)
```
akash1810: What's the value of `markerContext.marker` here?

common-lib/src/main/scala/com/gu/mediaservice/lib/logging/MarkerAugmentation.scala
```diff
@@ -0,0 +1,25 @@
...
+    MarkerContext(newMarker)
+  }
+  def augmentMarkerContext(markerContext: MarkerContext, augmentations: (String, Any)*): MarkerContext = {
```
akash1810: What scenario would we use this function over the one above?

common-lib/src/main/scala/com/gu/mediaservice/lib/logging/MarkerAugmentation.scala
AWare: `markerContext.marker` is an `Option[Marker]`

common-lib/src/main/scala/com/gu/mediaservice/lib/logging/MarkerAugmentation.scala
AWare: it's just sugar

common-lib/src/main/scala/com/gu/mediaservice/lib/logging/MarkerUtils.scala
```diff
@@ -2,15 +2,28 @@ package com.gu.mediaservice.lib.logging
...
+
+
+case class MarkerMap(map: Map[String, Any]) extends LogMarker {
+  override def toLogMarker: LogstashMarker = appendEntries(map.asJava)
+}
 
-case class MarkerMap(values: Map[String, Any]) extends LogMarker {
-  override def toLogMarker: LogstashMarker = appendEntries(values.asJava)
+object MarkerMap {
+  def apply(entries: (String, Any)*):MarkerMap = MarkerMap(entries.toMap)
```
akash1810: NICE!

common-lib/src/main/scala/com/gu/mediaservice/lib/logging/MarkerUtils.scala
```diff
@@ -2,15 +2,28 @@ package com.gu.mediaservice.lib.logging
...
 import net.logstash.logback.marker.Markers.appendEntries
+import play.api.MarkerContext
+import scala.language.implicitConversions
+
 import scala.collection.JavaConverters._
 
-trait LogMarker { def toLogMarker: LogstashMarker }
+trait LogMarker {
+  def toLogMarker: LogstashMarker
+  def map: Map[String, Any]
```
akash1810: can we name this to be more intention revealing?

kahuna/package.json
```diff
@@ -50,7 +50,7 @@
     "test": "npm run lint && jest --coverage",
     "lint": "eslint public/js && htmllint public/js/**/*.html",
     "lint-fix": "eslint public/js --fix && htmllint public/js/**/*.html",
-    "dist": "webpack --progress --config webpack.config.prod.js",
+    "dist": "webpack --config webpack.config.prod.js",
```
akash1810: Let's make this it's own PR?

dev-start.sh
```diff
@@ -116,8 +116,8 @@ setupLocalKinesis() {
   # ignore stream already exists error
   set +e
   stream_name='media-service-DEV-ThrallMessageQueue-1N0T2UXYNUIC9'
-  aws --profile media-service --region=eu-west-1 --endpoint-url=http://localhost:4568 kinesis create-stream --shard-count 1 --stream-name "${stream_name}"
-  aws --profile media-service --region=eu-west-1 --endpoint-url=http://localhost:4568 kinesis list-streams
+  aws --profile media-service --region=eu-west-1 --endpoint-url=http://localhost:4568 kinesis create-stream --shard-count 1 --stream-name "${stream_name}" | cat
```
akash1810: Let's make this it's own PR? https://github.com/guardian/grid/pull/2867

project/plugins.sbt
```diff
@@ -1,4 +1,4 @@
-resolvers += "Typesafe repository" at "http://repo.typesafe.com/typesafe/releases/"
+resolvers += "Typesafe repository" at "https://repo.typesafe.com/typesafe/releases/"
```
akash1810: YES!!

scripts/ci/grid.sh
```diff
@@ -13,6 +13,8 @@ setupNvm() {
 }
 
 buildJs() {
+  echo "##teamcity[compilationStarted compiler='webpack']"
```
akash1810: NICE!

thrall/app/lib/RetryHandler.scala
```diff
@@ -3,57 +3,73 @@ package lib
...
                                     executionContext: ExecutionContext,
-                                    mc: MarkerContext
-                                  ): () => Future[T] = {
-    def logFailures[T](f: () => Future[T])(
-      implicit executionContext: ExecutionContext, mc: MarkerContext
-    ): () => Future[T] = {
-      () => {
-        f().transform {
+                                  ): Future[T] = {
+    println("ALEX THREE")
```
akash1810: let's remove the `println`s

common-lib/src/main/scala/com/gu/mediaservice/lib/logging/MarkerUtils.scala
AWare: it's now called `markerContents`

AWare:
Dotting i's and crossing t's

- [ ] image
- [ ] reindex-image
- [ ] delete-image
- [ ] update-image
- [ ] delete-image-exports
- [ ] update-image-exports
- [ ] update-image-user-metadata
- [ ] update-image-usages
- [ ] replace-image-leases
- [ ] add-image-lease
- [ ] remove-image-lease
- [ ] set-image-collections
- [ ] delete-usages
- [ ] upsert-rcs-rights
- [ ] update-image-photoshoot
- [ ] batch-index

common-lib/src/main/scala/com/gu/mediaservice/lib/logging/MarkerUtils.scala
```diff
@@ -2,15 +2,28 @@ package com.gu.mediaservice.lib.logging
...
 import net.logstash.logback.marker.LogstashMarker
 import net.logstash.logback.marker.Markers.appendEntries
+import play.api.MarkerContext
+import scala.language.implicitConversions
+
 import scala.collection.JavaConverters._
 
-trait LogMarker { def toLogMarker: LogstashMarker }
+trait LogMarker {
+  def toLogMarker: LogstashMarker
```
akash1810: The implementation of this is the same (other than in `MarkerMap`); is it worth providing a default concrete implementation in the trait which can be overridden if you need to something different?

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchException.scala
```diff
@@ -0,0 +1,40 @@
...
+import net.logstash.logback.marker.LogstashMarker
+
+trait ElasticSearchError {
+  self: Throwable =>
+  def error: ElasticError
+
+  def markerContents: Map[String, Any]
+}
+
+object ElasticSearchException {
```
akash1810: does this need to extend `LogMarker`?

AWare: No, because it has an unapply that gives us a `LogMarker`

common-lib/src/main/scala/com/gu/mediaservice/lib/logging/MarkerUtils.scala
AWare: Yeah that's a good idea.

AWare:
Put this on TEST, and it's working for me. 

Not sure how to trigger the remaining messages, would like to just double check them once live.

PR #2858, opened 2020-03-12
track how long it took to process a message in thrall

akash1810:
## What does this change?
Thrall has received a message off kinesis, start a stopwatch and log the elapsed time on completion, where completion is one of:
- message subject is unknown
- message processed successfully within the retry constraints
- message processing timed out within the retry constraints
- message processing threw an error

`Stopwatch` is generic enough to be used in other places where we want to understand how long an operation takes.

Creates a `LogMarker` trait to make combining markers easier and a `MarkerMap` to make providing more markers easier too as it requires fewer imports.

Example usage:

```scala
import com.gu.mediaservice.lib.logging._

val stopwatch = Stopwatch.start

val requestMarkers = MarkerMap(Map(
  "route" -> "review-pr"
))

// do work

val extraMarkers = MarkerMap(Map(
  "status" -> "+1"
))

Logger.info("example")(combineMarkers(requestMarkers, extraMarkers, stopwatch.elapsed))
```

We understand how long thrall takes to process each message and can graph message subject by duration. This would be useful when trying to understand where performance bottlenecks may lie.

n/a

## Tested?
- [ ] locally
- [ ] on TEST


akash1810:
Deployed this to TEST and it seems to be preventing the log lines from appearing 😢 . Needs more work.

common-lib/src/main/scala/com/gu/mediaservice/lib/logging/MarkerUtils.scala
```diff
@@ -0,0 +1,16 @@
...
+import scala.collection.JavaConverters._
+
+trait LogMarker { def toLogMarker: LogstashMarker }
+
+case class MarkerMap(values: Map[String, Any]) extends LogMarker {
+  override def toLogMarker: LogstashMarker = appendEntries(values.asJava)
+}
+
+trait MarkerUtils {
+  val FALLBACK: String = "unknown"
```
jonathonherbert: Just out of interest (I can see it's not unique to this PR) – what is this for?

sihil:
Think it would be really great to track the period of time from putting a message onto the thrall queue and completion. Don't know if thrall messages contain the enqueue wall clock time but if they do this could be calculated from the wall clock time at completion.

common-lib/src/main/scala/com/gu/mediaservice/lib/logging/MarkerUtils.scala
akash1810: It's playing the role of a constant - when creating makers from `Option`s we need to `getOrElse`, this is the `orElse` value, keeping things DRY.

jonathonherbert:
@sihil – this would be v. useful, `facia-press` receives a similar value. Presumably beyond the scope of this PR.

akash1810:
> Think it would be really great to track the period of time from putting a message onto the thrall queue and completion. Don't know if thrall messages contain the enqueue wall clock time but if they do this could be calculated from the wall clock time at completion.

This would be valuable, yes! Would this use the created attribute of the message (assuming that's a thing) or would the message producer send this over?

In either case, I think this is a different log message and as such it's own PR.

akash1810:
> Would this use the created attribute of the message

Ah, we have this! https://github.com/guardian/grid/blob/1b67855fa025348aaff0a0cafc09c2c4bd8874ef/thrall/app/lib/kinesis/ThrallEventConsumer.scala#L72

PR #2859, opened 2020-03-12
Add "reason" for error in logs

stephengeller:
## What does this change?
When logging at ERROR level, there isn't an easily identifiable field for what _type_ of error is occurring, making it hard to visualise the frequency of it happening.

This adds a field called `reason` to the logs, which will add the class name of the error in order to allow it to be grouped in visualisation metrics.

The `reason` key is seen as part of error logs in ELK.

## Who should look at this?
@guardian/digital-cms 

## Tested?
- [X] locally
- [X] on TEST

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchExecutions.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchExecutions.scala
@@ -39,7 +39,10 @@ trait ElasticSearchExecutions {
 
     result.failed.foreach { e =>
       val elapsed = System.currentTimeMillis() - start
-      val markers = MarkerContext(durationMarker(elapsed))
+      val markers = MarkerContext(appendEntries(Map(
+        "duration" -> elapsed,
+        "reason" -> e.getClass.getName
+      ).asJava))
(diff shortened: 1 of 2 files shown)


AWare:
Can this wait for https://github.com/guardian/grid/pull/2832 please? I think we'll have this level of logging from that pr

stephengeller:
> Can this wait for #2832 please? I think we'll have this level of logging from that pr

Yes, absolutely! Sorry, @akash1810 mentioned that would be a sensible idea last night and I forgot. Will close this for now.

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchExecutions.scala
```diff
@@ -39,7 +39,10 @@ trait ElasticSearchExecutions {
 
     result.failed.foreach { e =>
       val elapsed = System.currentTimeMillis() - start
-      val markers = MarkerContext(durationMarker(elapsed))
+      val markers = MarkerContext(appendEntries(Map(
+        "duration" -> elapsed,
+        "reason" -> e.getClass.getName
```
AWare: This won't work as the error details are being swallowed by the `RuntimeException` on line 29

AWare: (should refresh before commenting again)

PR #2860, opened 2020-03-13
try and test the retrier

AWare:
draft pr for testing retry handler, can't figure out the implicit actor system

## Tested?
- [ ] locally
- [ ] on TEST


AWare:
I'm not going to get a chance to work on this for a while and I don't think it provides much benefit.

PR #2863, opened 2020-03-16
Improve image-loader controller structure

aug24:
## What does this change?
Improves the structure of image loader, which is currently quite overloaded.
Fix cases where a temp file is left behind.

Code is clearer, behaviour is unchanged except that temp files are cleared down.

N/A

## Who should look at this?
@guardian/digital-cms 

## Tested?
- [X] locally
- [X] on TEST

Uploading and projecting work fine.  I'm concerned in case there's some other functionality I should test, but tests pass, and LGTM.


image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -43,207 +38,130 @@ class ImageLoaderController(auth: Authentication, downloader: Downloader, store:
...
-        case Failure(error) => {
-          Logger.error(s"image projection failed", error)(requestContext.toMarker())
-          respondError(InternalServerError, "image-projection-failed", error.getMessage)
-        }
+    val tempFile = createTempFile(s"projection-$imageId")
+    Logger.info("XXX" )
+    auth.async { _ =>
+      val result = for {
+        result <- imageProjector.projectS3ImageById(imageUploadProjector, imageId, requestContext, tempFile)
+      } yield result
```
jonathonherbert: Correct me if wrong, but I think this desugars to 

```
imageProjector.projectS3ImageById(imageUploadProjector, imageId, requestContext, tempFile).map(identity)
```

that is to say, we can write this more simply as

```
val result = imageProjector.projectS3ImageById(imageUploadProjector, imageId, requestContext, tempFile)
```

akash1810:
Looks good. I think we can further improve image-loader by removing the import from uri path, but in a different PR.

PR #2864, opened 2020-03-17
Handle unknown errors as seperate state and too big metadatas

AWare:
## What does this change?

This adds two new statuses for images that have big projections or failed for an unknown reasons.
This also adds `End of data reached.` as a known error.

Todo: add new params as stack params

Not gumming up thrall.

## Tested?
- [ ] locally
- [ ] on TEST


AWare:
working good in TEST

PR #2867, opened 2020-03-17
fix kinesis local

AWare:
## What does this change?
aws was being clever on my machine and opening its results in less, piping the result into cat fixed this
@sihil 
 
(I think this is a side effect of awscli 2)

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/dev-start.sh
+++ b/dev-start.sh
@@ -116,7 +116,8 @@ setupLocalKinesis() {
   # ignore stream already exists error
   set +e
   stream_name='media-service-DEV-ThrallMessageQueue-1N0T2UXYNUIC9'
-  aws --profile media-service --region=eu-west-1 --endpoint-url=http://localhost:4568 kinesis create-stream --shard-count 1 --stream-name "${stream_name}" | cat
+  export AWS_PAGER=""
+  aws --profile media-service --region=eu-west-1 --endpoint-url=http://localhost:4568 kinesis create-stream --shard-count 1 --stream-name "${stream_name}"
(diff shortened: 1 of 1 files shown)


dev-start.sh
```diff
@@ -116,8 +116,8 @@ setupLocalKinesis() {
   # ignore stream already exists error
   set +e
   stream_name='media-service-DEV-ThrallMessageQueue-1N0T2UXYNUIC9'
-  aws --profile media-service --region=eu-west-1 --endpoint-url=http://localhost:4568 kinesis create-stream --shard-count 1 --stream-name "${stream_name}"
-  aws --profile media-service --region=eu-west-1 --endpoint-url=http://localhost:4568 kinesis list-streams
+  aws --profile media-service --region=eu-west-1 --endpoint-url=http://localhost:4568 kinesis create-stream --shard-count 1 --stream-name "${stream_name}" | cat
```
akash1810:
```suggestion
  export AWS_PAGER=""
  aws --profile media-service --region=eu-west-1 --endpoint-url=http://localhost:4568 kinesis create-stream --shard-count 1 --stream-name "${stream_name}"
```

dev-start.sh
```diff
@@ -116,8 +116,8 @@ setupLocalKinesis() {
...
+  aws --profile media-service --region=eu-west-1 --endpoint-url=http://localhost:4568 kinesis list-streams | cat
```
akash1810:
```suggestion
  aws --profile media-service --region=eu-west-1 --endpoint-url=http://localhost:4568 kinesis list-streams
```

PR #2875, opened 2020-03-20
Rename reindex -> reingest

jonathonherbert:
## What does this change?

Rename reindex -> reingest. Reingestion adds an image that already exists in our S3 bucket, but is not in Elasticsearch.

You should feel a distinct sense of semantic satisfaction.

## Tested?
- [x] locally
- [ ] on TEST


jonathonherbert:
@aug24 – great question. There is, but the code that writes that message to the queue hasn't been added to `grid-cli` yet (this is for a single reingestion).

PR #2876, opened 2020-03-22
Clicking new images ticker scrolls to top

paperboyo:
## What does this change?
It makes clicking ticker scroll to top to see these new arrivals. Took me three days of blindly pasting half the internets in between random parentheses.
Now, we need to have a better way of *always* keeping current view (of specific images): when images disappear because of deletion, on coming back from the viewer to the browser that loaded more images at the top or, as much as we can, even on search changes. But clicking this ticker is a clear declaration of intent: _I want to see them_. Once we have a good way of keeping current view, we could debate if we should scroll to the top like here or, maybe, to where new arrivals meet what one already have seen. For now, this will do. 

Success is for the weak.

## Who should look at this?
Mr. Bot.

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/kahuna/public/js/search/results.js
+++ b/kahuna/public/js/search/results.js
@@ -248,7 +248,7 @@ results.controller('SearchResultsCtrl', [
             // FIXME: should ideally be able to just call $state.reload(),
             // but there seems to be a bug (alluded to in the docs) when
             // notify is false, so forcing to true explicitly instead:
-            window.scrollTo(0,0);
+            $window.scrollTo(0,0);
             $state.transitionTo($state.current, $stateParams, {
                 reload: true, inherit: false, notify: true
             });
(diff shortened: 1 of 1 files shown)


kahuna/public/js/search/results.js
```diff
@@ -248,6 +248,7 @@ results.controller('SearchResultsCtrl', [
             // FIXME: should ideally be able to just call $state.reload(),
             // but there seems to be a bug (alluded to in the docs) when
             // notify is false, so forcing to true explicitly instead:
+            window.scrollTo(0,0);
```
akash1810: I think this should be [`$window`](https://github.com/guardian/grid/pull/2876/files#diff-b8c6d4770b5b04294cc8c88d27ae65d3L78) - [Angular wraps](https://docs.angularjs.org/api/ng/service/$window) native objects for namespacing, which makes testing easier.

```suggestion
            $window.scrollTo(0,0);
```

PR #2877, opened 2020-03-23
Revert "Revert "Improve image-loader controller structure"" and fix the response code.

aug24:
## What does this change?
Improves the structure of image loader, which is currently quite overloaded.
Fix cases where a temp file is left behind.

Code is clearer, behaviour is unchanged except that temp files are cleared down.

N/A

## Who should look at this?
@guardian/digital-cms 

## Tested?
- [X] locally
- [X] on TEST

Uploading and projecting work fine.

### DIFF
--- a/image-loader/app/controllers/ImageLoaderController.scala
+++ b/image-loader/app/controllers/ImageLoaderController.scala
@@ -145,7 +145,7 @@ class ImageLoaderController(auth: Authentication, downloader: Downloader, store:
         .map {
           r => {
             Logger.info("importImage request end")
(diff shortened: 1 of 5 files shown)


akash1810:
Could we add more detail to the description for future generations please?

akash1810:
Would be good to document _why_ a [202 response](https://github.com/guardian/grid/pull/2877/commits/f39f51c10c263a8802ddff38d71dc6d5f2085346) is needed for future generations. Otherwise LGTM.

PR #2879, opened 2020-03-23
[Single image reingest]: Check images aren't already present before reingesting

jonathonherbert:
## What does this change?

Adds a check to ensure images aren't already present in Elasticsearch before they're reingested. If they're already there, this is a no-op.

It should be impossible to reingest images already present in Elasticsearch.

It will be fairly difficult to check this as a user, as the cli tool will run its own checks before attempting reingestion, but the tests should confirm the logic.

## Tested?
- [x] locally
- [ ] on TEST


jonathonherbert:
Mothballing as it's probably better to allow this at the point of processing in favour of a `--force` option on the client.

PR #2885, opened 2020-03-26
Upgrade babel and everything build side js wise TO enable sourcemaps

AWare:
## What does this change?
This upgrades babel etc. so we can use more modern js.


jonathonherbert:
Looks good on TEST, and sourcemaps are there – nice! 👍

PR #2889, opened 2020-03-28
Switch from vanilla KCL processor to Akka streams

sihil:
## What does this change?
This switches Thrall from using the vanilla KCL library to consuming the Kinesis stream via an akka stream.

This runs locally and there is evidence that it correcty processes and checkpoints events.

From here it should be a hop, skip and a jump to add a second incoming kinesis stream with priority merging using a `MergePrioritised` graph shape.

This should work in exactly the same way as the existing code.

## Who should look at this?

Pre-rollout
-----------

 - [x] CFN update must be applied for new permission: https://github.com/guardian/editorial-tools-platform/pull/412

## Tested?
- [X] locally
- [x] on TEST


thrall/app/lib/kinesis/ThrallEventConsumer.scala
```diff
@@ -65,13 +66,14 @@ class ThrallEventConsumer(es: ElasticSearch,
     }
   }
 
-  private def parseRecord(r: Record):Option[UpdateMessage] = {
+  def parseRecord(r: Record):Option[UpdateMessage] = parseRecord(r.getData.array, r.getApproximateArrivalTimestamp.toInstant)
+
+  def parseRecord(r: Array[Byte], timestamp: Instant):Option[UpdateMessage] = {
```
jonathonherbert: Is there a reason this is overloaded in this way? I'm reading this as sugar for two val assignments at the start of the second declaration of `parseRecord`, perhaps to clarify the way the `Record` is used in the function?

thrall/app/controllers/HealthCheck.scala
```diff
@@ -4,18 +4,17 @@ import com.gu.mediaservice.lib.argo.ArgoHelpers
...
 import lib.elasticsearch._
-import lib.kinesis.ThrallMessageConsumer
 import play.api.Logger
 import play.api.mvc._
 
-import scala.concurrent.ExecutionContext
+import scala.concurrent.{ExecutionContext, Future}
 
-class HealthCheck(elasticsearch: ElasticSearch, messageConsumer: ThrallMessageConsumer, config: ThrallConfig, override val controllerComponents: ControllerComponents)(implicit override val ec: ExecutionContext)
+class HealthCheck(elasticsearch: ElasticSearch, streamRunning: Future[_], config: ThrallConfig, override val controllerComponents: ControllerComponents)(implicit override val ec: ExecutionContext)
```
akash1810: This type signature looks really odd 🤔

thrall/app/lib/ThrallStreamProcessor.scala
```diff
@@ -0,0 +1,79 @@
...
+      from = config.from
+    ).withKinesisEndpoint(thrallKinesisEndpoint)
+     .withDynamoDBEndpoint(thrallKinesisDynamoEndpoint)
+  }
+
+  val kinesisSource: Source[KinesisRecord, Future[Done]] = KinesisSource(kinesisConfig)
+
+  implicit val dispatcher = actorSystem.getDispatcher
+
+  def run(): Future[Done] = {
```
akash1810: Not quite sure I understand how we're returning a `Future[Done]` here.

thrall/app/lib/kinesis/ThrallEventConsumer.scala
sihil: This was overloaded as I had originally preserved the code of both ways (direct KCL and KCL via Akka streams) of consuming the Kinesis stream and they produced different types. As such one was a shim for the other without muddying the calling code.

However I removed the dual code towards the end of my work but didn't clean up the methods in this class.

thrall/app/lib/ThrallStreamProcessor.scala
sihil: `Future[Done]` is essentially `Future[Unit]`. It is returned by akka streams - in this case from the `runForEach`.

thrall/app/controllers/HealthCheck.scala
sihil: Is `=> Boolean` better? Leave the implementation to the wiring?

PR #2890, opened 2020-03-28
Allow remembering Metadata Panel state between sessions

paperboyo:
## What does this change?
I always wanted to do this, but this is for discussion. It makes Metadata panel behave like Collections panel: its state (shown/hidden, locked/unlocked) is kept between reloads/sessions in localstorage.

For: consistency of UI; useful for heavy users of this panel
Against: may be annoying to occasional or involuntary users of the panel as it stayes empty until something is selected (maybe it should always show on selection, I bet that would make for a good onboarding?)

Whether it’s more helpful and less annoying. Whether long-term it promotes editing metadata in the browser (including batch edits).

## Who should look at this?
Interested in @akash1810’s opinion, will ask users around (maybe @JonnyWeeks?)

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/kahuna/public/js/search/index.js
+++ b/kahuna/public/js/search/index.js
@@ -89,6 +89,7 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
             ctrl.metadataPanel = panels.metadataPanel;
 
             panelService.setAndSaveState($scope, 'collections', ctrl.collectionsPanel);
+            panelService.setAndSaveState($scope, 'metadata', ctrl.metadataPanel);
 
             keyboardShortcut.bindTo($scope).add({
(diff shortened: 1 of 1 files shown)


akash1810:
should we merge this?

paperboyo:
So: I would like to test my assumptions with heavy users, like Jo and Jonny. And then see if light users aren’t annoyed too. Hard to do remotely, though, so I’m not in a hurry. I would rather not have to revert, cos ppl are annoyed, because then a good idea will be burned.

Had a look at opening panel automagically on any selection, but we can’t do this because if it will open in a locked state, it will shift selected image out of view if you’re scrolled down :-(. Always remembering the place is too hard for me…

PR #2891, opened 2020-03-30
Fix regression on supported MIME types

aug24:
## What does this change?
https://trello.com/c/6vHImcof/1777-uploading-unsupported-filetypes-no-longer-produces-a-meaningful-error

Correct error on uploading unsupported mime type file.

## Who should look at this?
Mateusz

## Tested?
- [ ] locally
- [ ] on TEST


thrall/test/helpers/Fixtures.scala
```diff
@@ -21,7 +21,7 @@ trait Fixtures {
                  ): Image =
     Image(
      id = id,
-     uploadTime = DateTime.now(),
+     uploadTime = DateTime.now(DateTimeZone.UTC),
```
jonathonherbert: 😅

PR #2896, opened 2020-03-31
Thrall dual lanes

sihil:
## What does this change?
Implements high and low priority lanes in thrall. The high priority lane will get processed first and the low priority will get processed eventually.

When there is a record on the high priority stream, it get's processed first, regardless of what's on the low priority stream.
- Let logs with `subject=image` be A
- Let logs with `subject=update-image-usages` be B
- A with have `recordPriority` `high`
- B with have `recordPriority` `low`
- A's `recordArrivalTime` and `start` will be close
- B's `recordArrivalTime` and `start` will vary
- Some Bs will have `recordArrivalTime` close to A's, A will be started first as it is higher priority

We can confirm this with [this script](https://gist.github.com/akash1810/7dcc81004022f3e2349c0750ebf30ca0).

View logs [here](https://logs.gutools.co.uk/goto/d5654cf2be4f65d66bc1ccb8b601aa8c).

Column 4 is `recordArrivalTime`. Column 5 is `start`.
image

## Tested?
- [x] locally
- [x] on TEST

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,12 +40,31 @@ jobs:
         # Wait for elasticsearch to report healthy before continuing.
         # see https://github.com/actions/example-services/blob/master/.github/workflows/postgres-service.yml#L28
         options: -e "discovery.type=single-node" --expose 9200 --health-cmd "curl localhost:9200/_cluster/health" --health-interval 10s --health-timeout 5s --health-retries 10
+      localstack:
+        image: localstack/localstack:0.10.8
+        env:
+          SERVICES: kinesis:4568,dynamodb:4569
+          DEFAULT_REGION: eu-west-1
+          KINESIS_ERROR_PROBABILITY: 0.0
+          PORT_WEB_UI: 5050
+        ports:
+          - 5050:5050
+          - 4568:4568
+          - 4569:4569
+        options: >-
+          --health-cmd "curl localhost:5050/health"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 10
     steps:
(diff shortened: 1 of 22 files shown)


.github/workflows/ci.yml
```diff
@@ -40,12 +40,31 @@ jobs:
         # Wait for elasticsearch to report healthy before continuing.
         # see https://github.com/actions/example-services/blob/master/.github/workflows/postgres-service.yml#L28
         options: -e "discovery.type=single-node" --expose 9200 --health-cmd "curl localhost:9200/_cluster/health" --health-interval 10s --health-timeout 5s --health-retries 10
+      localstack:
+        image: localstack/localstack:0.10.8
+        env:
+          SERVICES: kinesis:4568,dynamodb:4569
```
akash1810, line 46: I'd love for this to be DRYer as it's repeating the `docker-compose.yml` but not sure how.

paperboyo:
💎!

.github/workflows/ci.yml
jennygrahamjones: I think that while both of these are defined with YAML this will repetition will be necessary, since you can't share any configuration across documents.

akash1810: I think there might be a way with [`options`](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idservicesoptions) as they follow the [docker options spec](https://docs.docker.com/engine/reference/commandline/create/#options). Something like `--env-file`? Alas, I [couldn't get it to work](https://github.com/guardian/grid/runs/573750783?check_suite_focus=true#step:2:65).

PR #2900, opened 2020-04-01
Justinrowles/upload layered tiffs

aug24:
## What does this change?
Allows the user to upload layered tiffs.

Successfully upload layered tiffs; continue to be able to upload all other supported image formats.

N/A

## Who should look at this?
Anyone.  Mateusz has already tested in TEST

## Tested?
- [X] locally
- [X] on TEST


common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
```diff
@@ -133,10 +133,33 @@ class ImageOperations(playPath: String) {
...
+
+  private def checkForOutputFileChange(f: File): Future[Unit] = Future {
+    val fileBits = f.getAbsolutePath.split("\\.").toList
+    val mainPart = fileBits.dropRight(1).mkString(".")
+    val extension = fileBits.last
+
+    // f2 is the blah-0 name that gets created from a layered tiff.
+    val f2 = new File(List(s"${mainPart}-0", extension).mkString("."))
+    if (f2.exists()) {
+      // f HAS been renamed to blah-0.  Rename it right back!
```
akash1810, line 148: Do we know why/where/how this happens? It's not clear to me from the detail in this PR.

aug24: I will improve the comment.

image-loader/app/model/ImageUpload.scala
```diff
@@ -375,20 +372,22 @@ object Uploader {
...
     uploadRequest.mimeType match {
       case Some(mime) => mime match {
         case _ if config.transcodedMimeTypes.contains(mime) =>
           for {
-            transformedImage <- imageOps.transformImage(uploadedFile, uploadRequest.mimeType, config.tempDir)
-          } yield transformedImage
+            transformedImage <- imageOps.transformImage(uploadRequest.tempFile, uploadRequest.mimeType, config.tempDir)
+          } yield (uploadRequest
+            // This file has been converted.
+            .copy(mimeType = Some(Jpeg))
```
jonathonherbert, line 384: Do we always convert to a jpeg here? I thought under some situations we converted to `png`, but I might be thinking of @paperboyo's future plans!

common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
```diff
@@ -140,10 +140,39 @@ class ImageOperations(playPath: String) {
...
+    } yield outputFile
+  }
+
+  // When a layered tiff is unpacked, the temp file (blah.something) is moved 
+  // to blah-0.something and contains the composite layer (which is what we want).
+  // Other layers are then saved as blah-1.something etc.
+  // As the file has been renamed, the file object still exists, but has the wrong name
+  // We will need to put it back where it is expected to be found, and clean up the other
+  // files. 
+  private def checkForOutputFileChange(f: File): Future[Unit] = Future {
```
jonathonherbert, line 153: Good lord. I hope our library continues to employ this logic indefinitely 🙈

common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
```diff
@@ -140,10 +140,39 @@ class ImageOperations(playPath: String) {
...
+    val fileBits = f.getAbsolutePath.split("\\.").toList
+    val mainPart = fileBits.dropRight(1).mkString(".")
+    val extension = fileBits.last
+
+    // f2 is the blah-0 name that gets created from a layered tiff.
+    val f2 = new File(List(s"${mainPart}-0", extension).mkString("."))
+    if (f2.exists()) {
```
jennygrahamjones, line 160: Is it possible for `f2` to not exist here?

image-loader/app/model/ImageUpload.scala
paperboyo: We definitely shouldn’t convert *always* to JPEG! Not even under current regime! (otherwise, we wouldn’t be able to support transparency)

common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
aug24: Yes, absolutely,

aug24:
> I wonder if we should store uploads sharded by request id. That is, rather than store everything in `/tmp` we'd create a `/tmp/<requestid>` directory and then the cleanup becomes `rm /tmp/<requestid>`

It's a nice idea - for a different card.  Currently I think we correctly delete temp files under all circumstances (brave assertion, I know).

common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
aug24: If the file is not layered, then it only makes the basic "blah.something" file.  It's only if it's layered that it creates blay-0, blah-1 etc.

PR #2907, opened 2020-04-03
add more detail to stopwatch

akash1810:
## What does this change?
Adding a `start` and `end` marker to stop watch logs for more information.

This is taken from https://github.com/guardian/grid/pull/2896, alas the commits in that PR were not easy to cherry pick out 😢 .

The more information we have, the better.

n/a

## Tested?
- [ ] locally
- [ ] on TEST


common-lib/src/test/scala/com/gu/mediaservice/lib/logging/StopwatchTest.scala
```diff
@@ -5,23 +5,15 @@ import org.scalatest.{FunSpec, Matchers}
...
     doWork
 
-    val marker = stopwatch.elapsed.toLogMarker
-    val map = stopwatch.elapsed.markerContents
+    val markers = stopwatch.elapsed.markerContents
 
-
-    map("duration") should be >= fiveSeconds
-
-    (marker.toString match {
```
akash1810, line 20: This is testing the map to string implementation, which I don't think we need.

PR #2910, opened 2020-04-07
further restrict image deletions

akash1810:
## What does this change?
Stricter conditions for image deletion. Essentially, you can only delete an image when there is nothing in the various dynamodb tables.

In processing a `delete-image` message, it sends a further message to [metadata editor](https://github.com/guardian/grid/blob/186741448e3bb1f7bc9eb01850792a481f7817e7/thrall/app/lib/kinesis/MessageProcessor.scala#L166) for it to [remove entries from dynamodb](https://github.com/guardian/grid/blob/2cc135c8fc8e3e2d92e5c092d90cf1b46beaa554/metadata-editor/app/lib/MetadataSqsMessageConsumer.scala#L18). As we're not doing the same for leases or collections, we can have a scenario where there are phantom records in these dynamodb tables.

By restricting the conditions under which an image is eligible for deletion, we can ensure these tables are empty before the image gets deleted. That is, if you want to delete an image, first remove its leases, collections, usages and crops.

The alternative to this is for lease and collections to get notified by thrall, similar to metadata-editor.

Cleaner dynamodb tables.

n/a

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/Image.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/Image.scala
@@ -35,7 +35,11 @@ case class Image(
 
   def hasUsages = usages.nonEmpty
 
-  def canBeDeleted = !hasExports && !hasUsages
+  def hasLeases = leases.leases.nonEmpty
+
+  def hasCollections = collections.nonEmpty
+
+  def canBeDeleted = !hasExports && !hasUsages && !hasLeases && !hasCollections
 
   def rcsPublishDate: Option[DateTime] = syndicationRights.flatMap(_.published)
 


jonathonherbert:
Stop. those. leaks 🚰 😁

jonathonherbert:
@akash1810 we've added more conditions for deletion – is it clear to users that those conditions must be met before images can be deleted?

akash1810:
This needs some messaging added to the UI. I'll pick this up soon.

paperboyo:
I think this may be very dangerous until digital Usage is 100% correct, no?

akash1810:
> I think this may be very dangerous until digital Usage is 100% correct, no?

How so? This makes you do more to delete an image. Previously you just had to delete an image's crops and usages. Now you have to do that _and_ delete it's leases and collections.

paperboyo:
> How so?

False sense of security (esp. with new messaging): suppose there is an active Usage, but it’s not shown, this new messaging will give an impression it’s safe to delete an image, even though it isn’t.

Regardless, I think that while (100% fixed) Usage should definitely block deletion, active Leases and Collections definitely shouldn’t. One should be free to delete an image without having to delete its Leases or its membership in Collections (just like one shouldn’t have to delete its Labels and metadata overridesetc). Upon deletions, they should be deleted automagically (part of this PR, right?).

[Related](https://trello.com/c/HfyEEjdy/762-deleting-usages-and-deleting-crops-verboten-if-active-digital-usage-in-place).

PR #2917, opened 2020-04-08
Queue for batch trackings

AWare:
## What does this change?

This runs batch tracking through the new queue for apipoll.

They have seperate worker counts. 

Faster large batches/larger batches that don't overwhelm anything.

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/kahuna/public/js/util/queue.js
+++ b/kahuna/public/js/util/queue.js
@@ -5,14 +5,28 @@
  * Enqueued functions are called concurrently up to the worker limit, and retried with an
  * exponential backoff.
  *
- * @param {object} options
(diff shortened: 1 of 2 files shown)


kahuna/public/js/util/queue.js
```diff
@@ -81,5 +89,6 @@ export const createQueue = ({
     }
   };
 
-  return { add };
+  const x = { add };
```
jonathonherbert, line 92: Was this added for debugging purposes? It looks like it can be removed.

AWare: Need to double check the JSDOC inference works without it.

PR #2921, opened 2020-04-14
Output test reports

akash1810:
## What does this change?
By outputting test reports, we can update CI to more easily identify which tests are failing; TC will display a list of executed tests in a separate location from the build log.

See https://github.com/guardian/janus/pull/1673

CI output is easier to understand.

image

## Who should look at this?

@guardian/digital-cms

## Tested?
- [x] locally
- [x] on TEST


akash1810:
This is failing CI because master is failing. Adding this to master makes it easier to determine why master is failing.

img

PR #2923, opened 2020-04-14
use latest version of localstack

akash1810:
## What does this change?
The latest version of localstack has a proxy running on 4566, this means we can talk to dynamo and kinesis on the same port which is a lot simpler.

Keeping up to date.

Doing this in the hope that https://github.com/guardian/grid/pull/2896 can be merged back into master and CI succeeds.

n/a

## Who should look at this?

@guardian/digital-cms

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,12 +18,10 @@ services:
     ports:
       - "9090:9000"
   localstack:
-    image: localstack/localstack:0.10.8
+    image: localstack/localstack:latest
     ports:
-      - "4567-4584:4567-4584"
-      - "8081:8080"
+      - "4566:4566"
     environment:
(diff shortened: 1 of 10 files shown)


docker-compose.yml
```diff
@@ -18,12 +18,10 @@ services:
     ports:
       - "9090:9000"
   localstack:
-    image: localstack/localstack:0.10.8
+    image: localstack/localstack:latest
```
jonathonherbert, line 21: Should we lock this to a specific version?

jonathonherbert: (I'd be worried about it changing underneath us)

akash1810: 👍 I guess it's like any other dependency.

Issue #1799, opened 2016-03-03
Label suggester won't work for Capitalised labels, also suggests badly


paperboyo:
Hello,

Above should summarise the below quite well I hoped.

![suggester stuck](https://cloud.githubusercontent.com/assets/6032869/13500772/e0dcb194-e15b-11e5-8748-8792d07b314c.gif)

![suggester glitch](https://cloud.githubusercontent.com/assets/6032869/13500782/e8a5caa0-e15b-11e5-8c36-19e7bbb4f71e.gif)

Related: https://github.com/guardian/grid/issues/1624

Regards
Mateusz

PR #2925, opened 2020-04-15
Fix label suggester

paperboyo:
## What does this change?
For a long time, our labels suggester was broken (fixes https://github.com/guardian/grid/issues/1799). It was using defunct `labelsSuggest` which performs ppl-who-bough-also-bought kind of suggesting, which may be useful one day somewhere, but never when autocompleting existing labels! This switches it to use `labelSearch` instead.

Sane autocompletion for labels.

image
Hello?!

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/kahuna/public/js/search/structured-query/query-suggestions.js
+++ b/kahuna/public/js/search/structured-query/query-suggestions.js
@@ -139,8 +139,8 @@ querySuggestions.factory('querySuggestions', ['mediaApi', 'editsApi', function(m
     }
 
     function suggestLabels(prefix) {
-        return mediaApi.labelsSuggest({q: prefix}).
-            then(labels => labels.data);
+        return mediaApi.labelSearch({q: prefix}).
+        then(results => results.data.map(res => res.key));
     }
 
     function suggestPhotoshoot(prefix) {


paperboyo:
This doesn’t work at all 😞. Below info from most helpful @itsibitzi, who had a cheeky look:

> Had a cheeky look at this - it looks like the label search is backed by an elasticsearch query but that query is wrong
> log message includes: 
`aggregate search: edits / AggregateSearchParams(label,Some(a a a),List(Match(AnyField,Words(where are my labels))) /`
with "where are my labels" being the search term
`Match` against `AnyField` using a `Words` search is probably not what we want
we need some kind of prefix search on the label field specifically

I will try to get some help. It was working brilliantly at the time of https://github.com/guardian/grid/pull/1247 as this GIF nicely shows:

image

paperboyo:
No idea if it matters, but at the time it was working splendidly, mappings [were storing](https://github.com/guardian/grid/blob/ff876916c4439a9e633d1af6f55d909a5c2d222f/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala#L97) `label`, while now they [are storing](https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala#L187) `labels`.

paperboyo:
This didn’t get anywhere 😢 . Closing.

AWare:
Delete this function and its references in `media-api`

AWare:
Closing, work is in #2945

PR #2926, opened 2020-04-16
produce a thrall message with the correct subject for usage deletions

akash1810:
## What does this change?
Pairing with @paperboyo, this corrects the subject of the message written to the kinesis stream for thrall when usages are deleted.

Thrall is looking for [`delete-usages`](https://github.com/guardian/grid/blob/ea3fe602b307e81c24273683ba03426e78fe054f/thrall/app/lib/kinesis/MessageProcessor.scala#L39) however we were producing ` delete-usages` 😅 and so thrall was dropping the message.

More reason to create the [sum type](https://github.com/guardian/grid/pull/2759)!

Usage deletions work again!

img

## Who should look at this?

@guardian/digital-cms

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/usage/app/controllers/UsageApi.scala
+++ b/usage/app/controllers/UsageApi.scala
@@ -193,7 +193,7 @@ class UsageApi(auth: Authentication, usageTable: UsageTable, usageGroup: UsageGr
         respondError(InternalServerError, "image-usage-delete-failed", error.getMessage)
     }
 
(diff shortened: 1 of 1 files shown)


akash1810:
CI is failing due to the the known dual lanes flaky test. Merging this in now and will take a look at the dual lane test again separately.

PR #2928, opened 2020-04-17
Change the reingest behaviour to produce a message for each image

jonathonherbert:
## What does this change?

This PR alters the behaviour of the reingest lambda. It used to produce a single message per invocation. These messages could be large, resulting in us having to impose restrictions on the number and size of the images to be included.

Now we have a low-priority stream, we can produce a message per image and enqueue them on that, without worrying that the large number of messages will affect the throughput of higher-priority messages on the queue.

Simultaneously, we can remove the code for creating batch messages, and the infra surrounding it (an S3 bucket), too.

This behaviour has been checked on TEST. Images marked as successfully reingested were present in Kahuna.

Trigger the reingestion lamba in TEST. It should produce the same results as before [in the metrics.](https://metrics.gutools.co.uk/d/U7pixN_Zk/media-service-image-reingestion?orgId=1) (although they behave a little erratically for TEST reingestions, as the dashboard is not parameterised for some graphs.)

You should see a [single `reingest-image`](https://logs.gutools.co.uk/s/editorial-tools/goto/89577eba7954451934533fcf5495787c) message produced for [each found image.](https://logs.gutools.co.uk/s/editorial-tools/goto/e79bf53e018ed2d2f8e4cadc75a1ee36)

## Tested?
- [x] locally
- [x] on TEST


admin-tools/lib/src/main/scala/com/gu/mediaservice/BatchIndexHandler.scala
```diff
@@ -115,14 +115,6 @@ class BatchIndexHandler(cfg: BatchIndexHandlerConfig) extends LoggingWithMarkers
   }
 
   def processImages(): Unit = {
-
-    if (AwsHelpers.checkKinesisIsNiceAndFast(stage, threshold))
```
jonathonherbert, line 119: We no longer care whether Kinesis is nice and fast – we have a low-priority stream! 🎉 

... should we keep this and extend the threshold to e.g. hours to avoid pointlessly enqueuing images when the queue is saturated? @aug24

aug24: Discussed and dealt with.

PR #2929, opened 2020-04-21
restructure ThrallStreamProcessor and update stream merging test

akash1810:
## What does this change?
The current stream merging test is flaky in master. This could be for a few reasons, including time it takes to initialise the KCL. This PR updates the test for the high and low priority stream processing.

Injecting `Source`s into `ThrallStreamProcessor` means we can create a simpler test and remove localstack and the KCL as a variable, that is, we test the bare minimum. This has shown that `MergePreferred` seems to take a number of records from both inlets before favouring the `preferred` one. The tests now reflect this with `alternatingSegment` indicating this behaviour.

CI becomes more reliable.

n/a

## Tested?
- [ ] locally
- [ ] on TEST


thrall/test/lib/ThrallStreamProcessorTest.scala
```diff
@@ -0,0 +1,49 @@
+package lib
```
akash1810, line 1: hmm, this is a [`git mv`](https://github.com/guardian/grid/pull/2929/commits/29014f6aa33bbac49ab53ccaaa62728979ac99b5) with edits, which apparently results in a delete and new file on GH?!

build.sbt
```diff
@@ -8,6 +8,12 @@ val commonSettings = Seq(
   organization := "com.gu",
   version := "0.1",
   scalacOptions ++= Seq("-feature", "-deprecation", "-language:higherKinds", "-Xfatal-warnings"),
+
+  // The Java SDK uses CBOR protocol
+  // We use localstack in TEST. Kinesis in localstack uses kinesislite which requires CBOR to be disabled
+  // This is likely go away soon, see https://github.com/localstack/localstack/issues/1930
+  envVars in Test := Map("AWS_CBOR_DISABLE" -> "true"),
```
akash1810, line 15: Setting this env var here makes running the tests in any environment simpler as we don't have to remember to set it ourselves.

thrall/test/lib/ThrallStreamProcessorTest.scala
sihil: It is heuristics based so depends how many edits in the file. If you want to be sure it's usually best to rename in one commit (which should be picked up as a mv) and then edits in a second commit.

akash1810: > If you want to be sure it's usually best to rename in one commit (which should be picked up as a mv) and then edits in a second commit.

I think I did this: the second commit on the branch is the move and the third is the edit 🤷‍♂️ .

PR #2930, opened 2020-04-22
Improve visibility of batch image status changes in logs

jonathonherbert:
## What does this change?

This PR improves the visibility of batch image status changes in logs. Previously, we were struggling to answer questions like 'which images were changed to this status', or 'what changes did this image undergo during the reingestion process'.

Adding structured logs for the image id, progress state and name should let us do just that.

This PR also adds additional configuration to logback to ensure that log levels (previously missing) are added correctly.

When images change status, a log line should be visible that includes structured information about the progress status number, its name, and the image id. That log should be marked as `level: info.`

## Tested?
- [x] locally
- [ ] on TEST


akash1810:
A rebase of master would resolve the issues in CI.

PR #2927, opened 2020-04-16
Revise running Grid from scratch

akash1810:
## What does this change?
AKA Grid in a box, this PR updates Grid to use [localstack](https://github.com/localstack/localstack) in DEV and optionally launches a pan-domain-authentication stack locally allowing you to run Grid with no additional setup.

Why? Currently, it's very hard to run the Grid from outside of the Guardian organisation. This is sad and we can be better! By using localstack we get closer to Grid in a box (at least for DEV).

View documentation how to use this new setup [here](https://github.com/guardian/grid/tree/aa-localstack/docs).

This is quite a big diff, I've split it out into as many small chunks as seems possible.

TODO:
- [x] convert panda/auth into localstack
- [x] check permissions
- [x] script generation of config files
- [x] test scripts on different machines
- [x] documentation
- [x] check if we still want to provide the cloudformation template for DEV
- [x] check usages
- [x] check cropping
- [x] check collections
- [x] check metadata edits
- [x] check uploading

Note: admin-tools fails to run when using local-auth as it [isn't using](https://github.com/guardian/grid/blob/master/admin-tools/dev/app/lib/AdminToolsConfig.scala) the standard configuration and so can't switch to using the local panda bucket. Leaving this for this PR as it's a bigger change to admin-tools in DEV to read from `/etc/gu`.

## How can success be measured?
Running Grid becomes:
- clone
- run `./setup.sh`
- run `./start.sh`
- open browser

## How to test?
Follow [the docs](https://github.com/guardian/grid/tree/aa-localstack/docs) to get Grid running from scratch!

## Screenshots (if applicable)
Running `setup.sh` and `start.sh`, then accessing Grid in a clean browser, uploading and cropping an image. Sequences shortened.

localstack

## Who should look at this?

@guardian/digital-cms

## Tested?
- [x] locally
- [x] on TEST

PR #2931, opened 2020-04-22
localstack doesn't like S3 user metadata in snake_case

akash1810:
## What does this change?
In https://github.com/guardian/grid/pull/2927 we're enabling running Grid locally in [localstack](https://github.com/localstack/localstack). It's quite a big PR which makes it hard to properly reason about. This change can be shipped independently. Small and often over big bang.

Localstack doesn't support snake_case in S3 object metadata, so move to kebab-case:
- Write metadata in kebab-case
- Read metadata as kebab-case and fallback to snake_case for backwards compatibility

More info from localstack [here](https://github.com/localstack/localstack/issues/459).

no-op

n/a

## Who should look at this?

@guardian/digital-cms

## Tested?
- [x] locally
- [x] on TEST

### DIFF
--- a/cropper/app/lib/CropStore.scala
+++ b/cropper/app/lib/CropStore.scala
@@ -8,6 +8,8 @@ import com.gu.mediaservice.lib.S3ImageStorage
 import com.gu.mediaservice.lib.logging.RequestLoggingContext
 import com.gu.mediaservice.model._
 
+import scala.util.Try
+
 class CropStore(config: CropperConfig) extends S3ImageStorage(config) {
   import com.gu.mediaservice.lib.formatting._
 
@@ -17,16 +19,16 @@ class CropStore(config: CropperConfig) extends S3ImageStorage(config) {
   def storeCropSizing(file: File, filename: String, mimeType: MimeType, crop: Crop, dimensions: Dimensions)(implicit requestContext: RequestLoggingContext) : Future[Asset] = {
     val CropSpec(sourceUri, Bounds(x, y, w, h), r, t) = crop.specification
     val metadata = Map("source" -> sourceUri,
-                       "bounds_x" -> x,
-                       "bounds_y" -> y,
-                       "bounds_w" -> w,
-                       "bounds_h" -> h,
+                       "bounds-x" -> x,
+                       "bounds-y" -> y,
+                       "bounds-width" -> w,
+                       "bounds-height" -> h,
                        "type" -> t.name,
                        "author" -> crop.author,
                        "date" -> crop.date.map(printDateTime),
                        "width" -> dimensions.width,
                        "height" -> dimensions.height
-                   ) ++ r.map("aspect_ratio" -> _)
+                   ) ++ r.map("aspect-ratio" -> _)
 
     val filteredMetadata = metadata.collect {
       case (key, Some(value)) => key -> value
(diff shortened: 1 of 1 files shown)


cropper/app/lib/CropStore.scala
```diff
@@ -56,15 +58,19 @@ class CropStore(config: CropperConfig) extends S3ImageStorage(config) {
...
             source <- userMetadata.get("source")
-            x      <- userMetadata.get("bounds_x").map(_.toInt)
-            y      <- userMetadata.get("bounds_y").map(_.toInt)
-            w      <- userMetadata.get("bounds_w").map(_.toInt)
-            h      <- userMetadata.get("bounds_h").map(_.toInt)
+
+            // we've moved to kebab-case as localstack doesn't like `_`
+            // fallback to reading old values for older crops
+            // see https://github.com/localstack/localstack/issues/459
+            x      <- Try(userMetadata.getOrElse("bounds-x", userMetadata("bounds_x"))).toOption.map(_.toInt)
```
akash1810, line 65: Wrap in a `Try` as the `getOrElse` fallback of `userMetadata("bounds_x")` isn't guaranteed to succeed.

AWare: I'd like to see this wrapped in a function.

akash1810: I did this but it made it less readable and was less obvious what was going on imo:

```scala
private def getMetadataValue(map: Map[String, String], preferredKey: String, fallbackKey: String): Option[String] = {
  Try(map.getOrElse(preferredKey, map(fallbackKey))).toOption
}

val updatedCrop = for {
  x <- getMetadataValue(userMetadata, "bounds-x", "bounds_x").map(_.toInt)
  y <- getMetadataValue(userMetadata, "bounds-y", "bounds_y").map(_.toInt)
  ... etc
}
```

akash1810: I think this is down to naming; `getOrElseOrNone` makes it clearer.

cropper/app/lib/CropStore.scala
```diff
@@ -44,6 +46,11 @@ class CropStore(config: CropperConfig) extends S3ImageStorage(config) {
     }
   }
 
+  private def getOrElseOrNone(theMap: Map[String, String], preferredKey: String, fallbackKey: String): Option[String] = Try(
+    // Return the `preferredKey` value in `theMap` or the `fallbackKey` or `None`
+    theMap.getOrElse(preferredKey, theMap(fallbackKey))
+  ).toOption
```
sihil, line 52: Using Try is a weird pattern here. In essence it is throwing and catching and processing an exception which seems to be overkill and confusing. This seems a lot clearer to me.

```suggestion
  private def getOrElseOrNone(theMap: Map[String, String], preferredKey: String, fallbackKey: String): Option[String] = {
    // Return the `preferredKey` value in `theMap` or the `fallbackKey` or `None`
    theMap.get(preferredKey).orElse(theMap.get(fallbackKey))
  }
```

sihil: It seems barely longer than the call to this function so it's tempting to put this inline.

cropper/app/lib/CropStore.scala
```diff
@@ -56,15 +63,19 @@ class CropStore(config: CropperConfig) extends S3ImageStorage(config) {
...
             // Note: if any is missing, the entry won't be registered
             source <- userMetadata.get("source")
-            x      <- userMetadata.get("bounds_x").map(_.toInt)
-            y      <- userMetadata.get("bounds_y").map(_.toInt)
-            w      <- userMetadata.get("bounds_w").map(_.toInt)
-            h      <- userMetadata.get("bounds_h").map(_.toInt)
+
+            // we've moved to kebab-case as localstack doesn't like `_`
+            // fallback to reading old values for older crops
+            // see https://github.com/localstack/localstack/issues/459
```
sihil, line 69: They claim this is 'kludged'. Possibly not merged. Makes me sad that we are working around an issue in a localstack but 🤷

PR #2932, opened 2020-04-22
co-locate files for DEV for clarity

akash1810:
## What does this change?
In #2927 we're enabling running Grid locally in localstack. It's quite a big PR which makes it hard to properly reason about. This change can be shipped independently. Small and often over big bang.

Group DEV stuff in a `dev` directory to provide more organisation.

img

img

## Who should look at this?

You.

## Tested?
- [x] locally
- [ ] on TEST


akash1810:
Snyk is failing against the `package.json` from `generate-dot-properties`. I think it's safe to ignore this for the time being as that project doesn't hit PROD in any way and is just a helper script.

PR #2935, opened 2020-04-23
Enable integration testing by adding IDs to elements 

stephengeller:
## What does this change?

This PR adds IDs to various elements in the Grid UI, so they can be used by integration tests for the Grid. As the integration tests grow, more IDs or other identifiers are likely to be added, but these are the necessary IDs to get the tests currently implemented [here](https://github.com/guardian/editorial-tools-integration-tests).

Elements with IDs added still work as intended and the integration tests pass.

## Tested?
- [X] locally
- [X] on TEST


kahuna/public/js/components/gr-add-label/gr-add-label.html
```diff
@@ -1,4 +1,5 @@
-<button ng-class="{'small': ctrl.grSmall}"
+<button id="it-add-label-button"
```
akash1810, line 1: I think this component is used on each cell in the search page. That is, this will produce invalid HTML as IDs should be unique in a DOM tree.

image

jonathonherbert:
Re: @akash1810's comment about ids and classes, I wonder whether it'd be worth using `data-testid` rather than `id` if we're adding attributes to tags for testing purposes. In this way, developers changing them will know what they they're altering something that's coupled to tests elsewhere. We've used this strategy elsewhere on e.g. [the fronts tool](https://github.com/guardian/facia-tool/blob/8aae8b46a36b5aab2e4180c7480442afab88586f/fronts-client/src/components/OptionsModal.tsx#L56) for our testing; there's an argument in a [Kent C Dodds article on test resilience](https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change) for this approach.

stephengeller:
@jonathonherbert @akash1810 Thanks for this. I wasn't aware of that limitation with repeated IDs, and even more so `data-testid`. After a read through the article it looks like that could be a good solution, and [it is even encourage by Cypress](https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements) so I'll have a look at implementing that instead.

PR #2939, opened 2020-04-24
Add a script to pull the batch ingest table image ids and statuses

jonathonherbert:
## What does this change?

Adds a script to write the current state of the batch ingest table to a file. The file is tab delimited in the format `<id> <progress_state>`.

Run the script with the appropriate arguments – `node scan-dynamo-table-for-pk.js <table-name> <output-file>`. It should behave as expected.

## Tested?
- [x] locally
- [ ] on TEST


jonathonherbert:
@SHession yeah, that'd be handy as a flag! I'll merge this for now but if we find we're repeating ourselves later on this could be a nice addition.

PR #2940, opened 2020-04-24
Move to ubuntu LTS for imaging

AWare:
## What does this change?

~This is blocked on ansible properly supporting ubuntu 20.04. 
https://github.com/ansible/ansible/pull/69161~

We are using the normal ubuntu package and not the ppa.

This changes the imaging ami to `grid-imaging` from `grid-imaging-eoan`
Eoan wasn't LTS, so will run out of support soon. Now there's a new LTS version we can use that. 

Supercedes https://github.com/guardian/grid/pull/2847

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/riff-raff.yaml
+++ b/riff-raff.yaml
@@ -91,7 +91,7 @@ deployments:
         ImagingAmiId:
           BuiltBy: amigo
           AmigoStage: PROD
-          Recipe: grid-imaging-eoan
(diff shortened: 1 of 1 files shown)


AWare:
Roll out: Deploy https://github.com/guardian/editorial-tools-platform/pull/433 and then merge this

PR #2941, opened 2020-04-28
add instructions for creating root collections

akash1810:
## What does this change?
Root collections do not have a UI at the moment and the only way to create them is via the API. Provide a script to simplify this process.

This PR also updates `setup.sh` to create an API key.

We should replace this with [grid-cli](https://github.com/guardian/grid-cli#grid-collectionadd-root-name), however `grid-cli` needs to be installed and configured first...

Collections are friendlier.

n/a

## Who should look at this?

@guardian/digital-cms

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/docs/03-apis/02-collections.md
+++ b/docs/03-apis/02-collections.md
@@ -1,0 +1,23 @@
+# Collections
+
+Collections are a way of organising images in Grid. Collections have a hierarchy similar to:
+
+```
+ROOT COLLECTION 1
+  ⎿ Collection A
+    ⎿ Collection B
+       ⎿ ...
+ROOT COLLECTION 2
+  ⎿ Collection C
+    ⎿ Collection D
+       ⎿ ...
+```
+
+## Root collections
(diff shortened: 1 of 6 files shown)


docs/03-apis/02-collections.md
```diff
@@ -0,0 +1,19 @@
...
+ROOT COLLECTION 2
+  ⎿ Collection C
+    ⎿ Collection D
+       ⎿ ...
+```
+
+## Root collections
+Currently, root collections can only be created via the API. That is, Kahuna doesn't have a UI for it.
+
+We can use [create-root-collection.sh](../../dev/script/create-root-collection.sh) to create a root collection.
```
jennygrahamjones, line 19: Is it just a case of running the script with a single argument: the name of the collection?

akash1810: Yep. Maybe this should be more explicit here too.

akash1810: Done.

PR #2945, opened 2020-05-05
Remove label suggestion js

AWare:
## What does this change?

This is a more thorough expurgation of the label suggestion, but @paperboyo's pr description is the more thorough exposition of it: https://github.com/guardian/grid/pull/2925

## Who should look at this?

@paperboyo, to see whether I have gotten ahold of totally the wrong end of the stick.

## Tested?
- [ ] locally
- [ ] on TEST


sihil:
Closing as this should be fixed properly as "hope is a gift that we should give ourselves everyday" - @jonathonherbert

PR #2947, opened 2020-05-05
Sentry strat 1: Catch no-op HTTP errors before they make it to Sentry

jonathonherbert:
## What does this change?

A way of preventing Sentry from seeing HTTP errors – adding a catch handler to each function that's missing one, with a utility function that filters out HTTP failures but allows other errors to propagate.

I've added handler code to `mediaApi` calls for now – if this approach is good, we can extend it to other hypermedia services.

HTTP errors no longer make it to Sentry.

## Tested?
- [x] locally
- [ ] on TEST


jonathonherbert:
Closed in favour of #2953

PR #2948, opened 2020-05-09
cache sbt dependencies in github actions

akash1810:
## What does this change?
Caching should improve build time.

See: https://github.com/actions/cache/blob/master/examples.md#scala---sbt

Scala build time in GitHub Actions is faster.

n/a

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,6 +62,11 @@ jobs:
       with:
         path: ~/.ivy2/cache
         key: ${{ runner.os }}-sbt-ivy-cache-${{ hashFiles('**/build.sbt') }}
+    - name: Cache SBT coursier cache
+      uses: actions/cache@v1
+      with:
+        path: ~/.coursier/cache
+        key: ${{ runner.os }}-sbt-coursier-cache-${{ hashFiles('**/build.sbt') }}
     - name: Cache SBT
       uses: actions/cache@v1
       with:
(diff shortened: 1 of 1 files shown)


akash1810:
Looks like this isn't quite performing how it should:

image

.github/workflows/ci.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           npm install
           npm run riffraff-artefact
   ScalaBuild:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
```
AWare, line 36: I can see this jumping to an ubuntu version without java 8? (I'm not sure how this interacts with the mozilla sbt image, so this may not be useful)

.github/workflows/ci.yml
```diff
@@ -55,13 +55,33 @@ jobs:
...
+        key: ${{ runner.os }}-sbt-${{ hashFiles('**/build.sbt') }}
     - name: SBT
-      uses: ./.github/actions/sbt
       env:
         USE_DOCKER_FOR_TESTS: false
         ES6_TEST_URL: http://elasticsearch:9200
         LOCALSTACK_ENDPOINT: http://localstack:4566
-      with:
-        args: clean compile test
+      run: |
```
AWare, line 86: Is this pipe thing yaml magic? Would it work if it was on one line?

akash1810:
Whilst the latest build of this appears to have worked, in that the GHA build has passed, it has gotten a lot slower! Closing this as a result.

akash1810:
I think i just mis-configured this. https://github.com/guardian/sbt-riffraff-artifact/pull/52 is an example of an sbt project being built with GHA with caching enabled. The timings are delightful!

PR #2949, opened 2020-05-12
Filmstrip in Image view

itsibitzi:
Towards next-next-next. 

Raising a PR to track progress. Very much still a work in progress.

Pulls out the management of fetching images into an AngularJS service which more than one view can hook into.

Adjacent images are rendered at the bottom of the image view in a style suspiciously similar to Lightroom. A picture speaks a thousand words, but how many does a gif?

nextnextnext

Navigation works with arrow keys, but you can also click on a particular image to jump to it too.

The previous CSS used a _lot_ of `calc` which meant it was fragile to change so I've moved over to using `flex` and `flex-grow` for layout.

- [ ] Finish moving search state out of the results component.
- [ ] Let's chat about UI/UX!!!!!
- [ ] Clean up styling
- [ ] Lazy load images, scale scroll bar (sadly results grid on the search page isn't generic enough)
- [ ] Centre filmstrip when moving out of view
- [ ] Make work in fullscreen
- [ ] Finalise style, fit and finish.

Some bonus features:
* Padding on the download button fixed (been bothering me for years)
* Selecting a crop won't squish the image anymore.
* Lots of overflowing elements in the image details panel, making tiny scroll bars.

### DIFF
--- a/kahuna/public/js/components/gr-downloader/gr-downloader.css
+++ b/kahuna/public/js/components/gr-downloader/gr-downloader.css
@@ -5,4 +5,6 @@ gr-downloader:hover {
 gr-downloader .download {
     display: block;
     height: 100%;
+    /* Lopsided padding because the icon on the right provides 10px of padding in that direction */
+    padding-left: 10px;
 }
--- a/kahuna/public/js/image/controller.js
+++ b/kahuna/public/js/image/controller.js
@@ -99,11 +99,16 @@ image.controller('ImageCtrl', [
 
     ctrl.images = imagesService.getImages();
 
-
     ctrl.tabs = [
       {key: 'metadata', value: 'Metadata'},
       {key: 'usages', value: `Usages`, disabled: true}
     ];
+
+    ctrl.showFilmstrip = true;
+    ctrl.toggleShowFilmstrip = () => {
+      console.log('toggle');
+      ctrl.showFilmstrip = !ctrl.showFilmstrip;
+    }
 
     ctrl.selectedTab = 'metadata';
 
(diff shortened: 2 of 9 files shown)


paperboyo:
💎 💎 💎

philmcmahon:
➡️ ➡️ ➡️ ⏮️ 
keep it up

blishen:
Woah! This is lovely. And will save me a job if I can deprecate my full screen slideshow thing they use currently...

AWare:
@itsibitzi what's needed from that big list to get this merged vs a new pr?

paperboyo:
My, uncharacteristically pragmatic, take on it is as follows.
- I think this is one of the most important and useful changes to Grid UX-wise and will transform and speed up editors’ work immensely
- priority is speed

I would see it done in steps, to make sure it’s in with minimal disruption and more quickly too:
1. Merge changes that mention using “flex and flex-grow” but without any UI changes yet (to realise **current* UI in newer ways and fix plethora of current UI glitches
2. Merge main engine change allowing arrow keys to move through pics in Viewer using current search context (and sort order) from Browser (no UI changes, not even any chevrons!)
3. Communicate to Picture Desk
4. Incorporate preloading/caching for speed; think about speeding up Browser<>Viewer page transitions
5. Then, and only then, think about actually building Filmstrip UI (requires collapsible bottom panel, keyboard shortcuts and many other things that can, and should, wait for later)

I strongly believe, doing 1&2 and releasing it in the wild would already provide huge benefit for people browsing Grid their whole work day!

itsibitzi:
For MVP. The key code change for this is moving the management of loading new image data out of the angular UI and into a separate service. Sadly, right now they're quite intertwined. 

Most the UI stuff is trivial once that's done.

There's an issue with full screen right now see:
https://github.com/guardian/grid/blob/6130e85e83d80470e1729897b94ead8535c620e6/kahuna/public/js/image/controller.js#L108-L133

The current implementation basically lifts the easel element into a full screen view, which is fine when it's static. When angular needs to re-render the image elements the new full screen window closes. I've not given a massive amount of thought to how this could be solved. Perhaps we could full screen a separate element that isn't managed by angular.

PR #2951, opened 2020-05-14
Prevent swimming images

paperboyo:
This tries to prevents further unnecessary scrolling possible:

Moves
 

Possibly related to my [earlier change](https://github.com/guardian/grid/pull/2950), but possibly not.

## Tested?
- [ ] locally
- [ ] on TEST


paperboyo:
This breaks more than it fixes. Closing.

PR #2618, opened 2019-08-20
Supplier list matches

felixcs1:
## What does this change?
In order for this to be merged, both [2612](https://github.com/guardian/grid/pull/2612) and then [2614 ](https://github.com/guardian/grid/pull/2614) must be merged first.

This extracts the matching on the image source and credit into Usage rights config (which is read into from `usage_rights.json` on s3). This allows credit matching to be updated without the need for a redeploy. It is also very useful for us (BBC) as we have slightly different matches. The matching is done using regex so you can include regex patterns in the json.

For each parser, we first match the image source and credit against the lists stored in s3 (with a common function in the the ImageProcessor trait)  and then if that fails matches on the special cases that were already in the code. 

## How can success be measured?
Update the usage_rights.json with the matches given here: 

<details><summary>usage_rights.json</summary>
<p>

#### The full json file to be put in `s3.config.bucket`

```
{
  "supplierCreditMatches": [{
      "name": "AapParser",
      "creditMatches": ["AAPIMAGE", "AAP IMAGE", "AAP"],
      "sourceMatches": []
    },
    {
      "name": "ActionImagesParser",
      "creditMatches": ["Action Images", "Action Images via Reuters"],
      "sourceMatches": []
    },
    {
      "name": "AlamyParser",
      "creditMatches": ["Alamy", "Alamy Stock Photo"],
      "sourceMatches": []
    },
    {
      "name": "AllStarParser",
      "creditMatches": ["Allstar Picture Library"],
      "sourceMatches": []
    },
    {
      "name": "ApParser",
      "creditMatches": ["ap", "associated press"],
      "sourceMatches": []
    },
    {
      "name": "BarcroftParser",
      "creditMatches": ["barcroft media", "barcroft images", "barcroft india", "barcroft usa", "barcroft cars"],
      "sourceMatches": ["barcroft media", "barcroft images", "barcroft india", "barcroft usa", "barcroft cars"]
    },
    {
      "name": "BloombergParser",
      "creditMatches": ["Bloomberg"],
      "sourceMatches": []
    },
    {
      "name": "CorbisParser",
      "creditMatches": [],
      "sourceMatches": ["Corbis"]
    },
    {
      "name": "EpaParser",
      "creditMatches": [".*\\bEPA\\b.*"],
      "sourceMatches": []
    },
    {
      "name": "GettyXmpParser",
      "creditMatches": [],
      "sourceMatches": []
    },
    {
      "name": "GettyCreditParser",
[...]

PR #2953, opened 2020-05-17
Add “ via ” as byline and credit tokeniser

paperboyo:
This adjusts our byline and credit fields cleaners to treat ` via ` exactly as they treat `/`. It provides more consistency in how we display information from different suppliers. It allows more of our cleaners to operate on wider set of bylines and credits. Some suppliers’ logic and its tests were adjusted too.

|Before:|After:|
|--|--|
|byline: `Agencja Gazeta`|byline: `Agencja Gazeta`|
|credit: `via Reuters`|credit: `Reuters`|
| | |
|byline: `Andrew Boyers`|byline: `Andrew Boyers`|
|credit: `Action Images via Reuters`|credit: `Action Images/Reuters`|
| | |
|byline: `Ed Garvey`|byline: `Ed Garvey`|
|credit: `Manchester City FC via EMPICS/PA Photos`|credit: `Manchester City FC/EMPICS/PA Photos`|
| | |
|byline: `Bloomberg`|byline: `Bloomberg`|
|credit: `Bloomberg via Getty Images`|credit: `Getty Images`|

One remaining issue is that this doesn’t deal with cleaners not being able to clean fields where those start with `via ` (`via Reuters` example at the top above is dealt with by supplierProcessor [change](https://github.com/guardian/grid/commit/86f0fb1e500b39c3710c87b9cb9db0432ef14d6e#diff-4b854d8f2e0bc0725c5b4804936b7875R255) separately*), so we can still have 

originalByline: `PIPPA FOWLES/10 DOWNING STREET`
original credit: `via REUTERS`
being cleaned to
byline: `Pippa Fowles`
credit: `10 DOWNING STREET/via REUTERS`

To sort this out, we would need to:
1. Adjust cleaners to remove tokenisers at the beginning of fields
2. Adjust supplierProcessor to work on last token in credit instead of the whole credit.

*two Reuters-related changes in this PR should be their own PR, but aren’t. Changes to business logic [shouldn’t require](https://github.com/guardian/grid/pull/2618) a PR anyway ;-)

@guardian/digital-cms & @itsibitzi (cos cleaners)

## Tested?
- [x] locally
- [x] on TEST

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
@@ -252,7 +252,7 @@ object ReutersParser extends ImageProcessor {
   def apply(image: Image): Image = image.metadata.credit match {
     // Reuters and other misspellings
     // TODO: use case-insensitive matching instead once credit is no longer indexed as case-sensitive
-    case Some("REUTERS") | Some("Reuters") | Some("RETUERS") | Some("REUTERS/") | Some("via REUTERS") | Some("VIA REUTERS") | Some("via Reuters") => image.copy(
+    case Some("REUTERS") | Some("Reuters") | Some("RETUERS") | Some("REUETRS") | Some("REUTERS/") | Some("via REUTERS") | Some("VIA REUTERS") | Some("via Reuters") => image.copy(
       usageRights = Agency("Reuters"),
       metadata = image.metadata.copy(credit = Some("Reuters"))
     )
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
@@ -59,7 +59,7 @@ object AapParser extends ImageProcessor {
 
 object ActionImagesParser extends ImageProcessor {
   def apply(image: Image): Image = image.metadata.credit match {
-    case Some("Action Images") | Some("Action Images via Reuters") => image.copy(
+    case Some("Action Images") | Some("Action Images/Reuters") => image.copy(
       usageRights = Agency("Action Images")
     )
     case _ => image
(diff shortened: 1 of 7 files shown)


paperboyo:
[Add ` via ` as a tokeniser](https://trello.com/c/FsNPHzmN/1878-add-via-as-a-tokeniser)

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -252,7 +252,7 @@ object ReutersParser extends ImageProcessor {
   def apply(image: Image): Image = image.metadata.credit match {
     // Reuters and other misspellings
     // TODO: use case-insensitive matching instead once credit is no longer indexed as case-sensitive
-    case Some("REUTERS") | Some("Reuters") | Some("RETUERS") | Some("REUTERS/") | Some("via REUTERS") | Some("VIA REUTERS") => image.copy(
+    case Some("REUTERS") | Some("Reuters") | Some("RETUERS") | Some("REUETRS") | Some("REUTERS/") | Some("via REUTERS") | Some("VIA REUTERS") | Some("via Reuters") => image.copy(
```
jonathonherbert, line 255: Do we catch all of these permutations, or do some slip through? We could use a fuzzy match to detect spelling errors in this and other fields if it was worth the effort ... (in another PR 😁)

paperboyo: Good point, but this whole business logic will change as soon as we review and merge BBC PRs, so I would just make those able to deal with that.

AWare: Maybe we should do that instead?

paperboyo: > we should do that

Do what? Review and merge BBC PRs? Oh, yes! (but not instead, we can adjust supplierProcessor logic after). We shoud also engineer BBC PRs such, that all manner of crazy stuff currently there and stuff we would like to do, but can’t now, should be possible. Eg for here, as noted, matching a supplier on last token would help: we are loosing some pics from free tier because of that.

jonathonherbert:
Keen to approve, @paperboyo – what's the easiest way to check this code in local/test? Would we need a pre-gridified agency image?

paperboyo:
Sorry for late reply. Easiest would be to fill Author and Credit metadaat using editing app or exiftool. Alternatively, one can run tests manually (?). Or, upload these IDs to TEST/local or PROD/TEST (with this branch deployed) and check out the difference:
56df1d7be09d978403e7d87d59fbadf17c725557, f5fa2fc83521203b33090615f0381f6a2a84366b, f7f17f21d54f07ec0e224343700e0d5720f31fd3, ab31ea9dfe6262d890a306491557399e22540e77.

PR #2957, opened 2020-06-10
fixes to mappings script

AWare:
## What does this change?

This makes the es-mappings script compile.

It doesn't work yet.

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/dev/script/es-ssh-ssm-tunnel.sh
+++ b/dev/script/es-ssh-ssm-tunnel.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+set -e
+
+function HELP {
+>&2 cat << EOF
+
+  Usage: ${0} -t ES-HOST [-s TEST] [-u ubuntu]
+
+  This script sets up an ssh tunnel from localhost port 9200 to the ElasticSearch provided on port 9200
+
+    -v            Verbose ssh output
+
+    -h            Displays this help message. No further functions are
+                  performed.
+
+EOF
+exit 1
+}
+
+BACKGROUND="-f"
+VERBOSE=""
+
+# Process options
+while getopts u:s:t:hv FLAG; do
+  case $FLAG in
+    h)  #show help
+      HELP
+      ;;
+    v)
+      VERBOSE="-v"
+      ;;
+    s)
+      STAGE=$OPTARG
+  esac
+done
+shift $((OPTIND-1))
+
+
+if [ -z "${STAGE}" ]; then
+  STAGE="TEST"
+fi
+
+echo "🛰 fetching connection details from ssm"
+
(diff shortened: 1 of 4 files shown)


dev/script/es-ssh-ssm-tunnel.sh
```diff
@@ -0,0 +1,53 @@
...
+ELB_NAME=""
+
+echo "🛰 fetching connection details from ssm"
+
+SSM_COMMAND=$(ssm ssh --profile media-service -t elasticsearch-data,grid-elasticsearch,$STAGE --newest --ssm-tunnel --raw)
+
+echo "📠 ESTABLISHING CONNECTION"
+
+echo "$SSM_COMMAND"
+eval $SSM_COMMAND $VERBOSE -L 9300:localhost:9200 -N
```
akash1810, line 53:
```suggestion
eval $SSM_COMMAND $VERBOSE -L 9300:localhost:9300 -N
```

dev/script/es-ssh-ssm-tunnel.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
+set -e
+
+function HELP {
+>&2 cat << EOF
+
+  Usage: ${0} -t ES-HOST [-s TEST] [-u ubuntu]
+
+  This script sets up an ssh tunnel from localhost port 9300 to the ElasticSearch provided on port 9300
```
akash1810, line 9: Why 9300 over 9200?

dev/script/es-ssh-ssm-tunnel.sh
```diff
@@ -0,0 +1,53 @@
...
+done
+shift $((OPTIND-1))
+
+
+if [ -z "${STAGE}" ]; then
+  STAGE="TEST"
+fi
+echo "📡 getting ELB name from AWS"
+
+ELB_NAME=""
```
akash1810, line 44: should this be the empty string?

dev/script/es-ssh-ssm-tunnel.sh
AWare: I think it was in the README or something equally not good as an answer. Will fix.

dev/script/es-ssh-ssm-tunnel.sh
AWare: This step isn't needed as I got the direct tunnel working, will remove.

akash1810:
Could we update the CI step to ensure we catch this class of regression earlier in the future?

jonathonherbert:
Great point @akash1810, me and @AWare discussed this on Friday and I'd forgotten it over the weekend. This would be 💯

PR #2669, opened 2019-10-25
Feature/api search by people

jamie-pitts:
This requires both https://github.com/guardian/grid/pull/2664 and https://github.com/guardian/grid/pull/2667 to be merged first.

## What does this change?
Allows for you to search for a 'person' using either the api or via the UI. It also displays the people in an image on the ui in the metadata section if there are any people present. When searching for the name of a person it must be an exact match and case sensitive - probably something we would want to improve on later on.

Screenshot 2019-10-25 at 12 16 10

Screenshot 2019-10-25 at 12 16 19

PR #2958, opened 2020-06-15
 Feature/api search by people #2669 

AWare:
this is #2669
also includes all the commits from https://github.com/guardian/grid/pull/2956

# Original PR
This requires both https://github.com/guardian/grid/pull/2664 and https://github.com/guardian/grid/pull/2667 to be merged first.

## What does this change?
Allows for you to search for a 'person' using either the api or via the UI. It also displays the people in an image on the ui in the metadata section if there are any people present. When searching for the name of a person it must be an exact match and case sensitive - probably something we would want to improve on later on.

Screenshot 2019-10-25 at 12 16 10

Screenshot 2019-10-25 at 12 16 19

### DIFF
--- a/kahuna/public/js/edits/metadataDiff.js
+++ b/kahuna/public/js/edits/metadataDiff.js
@@ -1,30 +1,25 @@
-function isDefined(value){return typeof value !== 'undefined';}
-function isUndefined(value) { return typeof value === 'undefined'; }
-//These two functions have been copied from angular.js and will be deleted.
-
 export const getMetadataDiff = (image, metadata) => {
-  var diff = {};
-
+  const diff = {};
+  const originalMetadata = image.data.originalMetadata;
   // jscs has a maximumLineLength of 100 characters, hence the line break
-  var keys = new Set(Object.keys(metadata).concat(
-    Object.keys(image.data.originalMetadata)));
-
-  // Keywords, peopleInImage and subjects are arrays, the comparison below only works with string comparison.
-  // For simplicity, ignore them as we're not updating these fields at the moment.
-  keys.delete('keywords');
-  keys.delete('subjects');
-  keys.delete('peopleInImage');
+  var keys = new Set(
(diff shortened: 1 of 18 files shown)


AWare:
The original metadata change failed on prod as `getMetadataDiff` could not handle arrays, and special cased the known arrays. This adds a test suite for that function, which should handle all the functionality that's described in it, and then replaces the implementation with something simpler which should handle arrays in any position.

kahuna/public/js/edits/metadataDiff.js
```diff
@@ -0,0 +1,26 @@
...
+    Object.keys(metadata).concat(Object.keys(originalMetadata))
+  );
+
+  keys.forEach((key) => {
+    if (originalMetadata[key] === metadata[key]) {
+      return;
+    }
+
+    //This only works with string fields and does not support arrays
+    if (Array.isArray(metadata[key]) || Array.isArray()) {
```
jonathonherbert, line 15: `Array.isArray()` will always return false – what's the purpose of this second condition?

AWare: That's supposed to be checking the other potential value for an array

kahuna/public/js/edits/metadataDiff.js
```diff
@@ -0,0 +1,26 @@
+export const getMetadataDiff = (image, metadata) => {
+  const diff = {};
+  const originalMetadata = image.data.originalMetadata;
+  // jscs has a maximumLineLength of 100 characters, hence the line break
+  var keys = new Set(
+    Object.keys(metadata).concat(Object.keys(originalMetadata))
+  );
+
+  keys.forEach((key) => {
+    if (originalMetadata[key] === metadata[key]) {
```
jonathonherbert, line 10: This logic is expressed much more cleanly here – as far as I can see it's like for like 🤞

PR #2959, opened 2020-06-15
Add documentation for reingest process

jonathonherbert:
## What does this change?

Adds docs for the reingest process.

There are also a few minor refactors that made sense as we were writing this up.

If you read the docs, and you feel able to attempt the process, that feels like a tick. If we need more detail, let's add it!

## Tested?
- [x] locally 😁 
- [ ] on TEST


AWare:
Just a thought, but I guess the primary use of this in the future is more likely to be reprocessing metadata- so we might want to take another look at that code.

jonathonherbert:
@AWare correct – there are a few obstacles to getting that done that AFAIK are documented on the health trello.

PR #2963, opened 2020-06-18
Add has: aliases for peopleInImage and for leases.leases

paperboyo:
[because of discussion below, the scope of this PR is much reduced, it now only does what the title says it does]

This adds two `has:` aliases:
- `people` for `peopleInImage`
- `leases` for `leases.leases`

To test, try searching `has:people` or `has:leases`.

The below changes have been abandoned after a discussion (keeping here, because there are some good points there about fields that support multiple values and a feature request too).

There is a slight UI mismatch that viewer shows

image

but when one launches a 🍟 , it displays as

image

This is an attempt at using consistent terminology. And, `people` will be as correct as `person` as soon as the 🍟 can take an array of values.

This also introduces a 🍟 alias, so that one can do `+has:people`. And, because it’s bad practice to introduce unrelated changes within one PR, it also does this to `leases.leases`, so that [this madness](https://media.gutools.co.uk/search?query=has:leases%20-has:leases.leases&nonFree=true) doesn’t happen.

## Tested?
- [ ] locally
- [x] on TEST

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/ImageFields.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/ImageFields.scala
@@ -50,6 +50,7 @@ trait ImageFields {
     "croppedBy" -> "exports.author",
     "filename"  -> "uploadInfo.filename",
     "photoshoot"-> photoshootField("title"),
+    "leases" -> "leases.leases",
     "leasedBy" -> "leases.leases.leasedBy",
     "people" -> "peopleInImage"
   )
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/ImageFields.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/ImageFields.scala
@@ -52,7 +52,7 @@ trait ImageFields {
     "photoshoot"-> photoshootField("title"),
     "leases" -> "leases.leases",
     "leasedBy" -> "leases.leases.leasedBy",
-    "people" -> "peopleInImage"
+    "people" -> metadataField("peopleInImage")
   )
 
   def getFieldPath(field: String) = field match {
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/ImageFields.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/ImageFields.scala
@@ -50,7 +50,8 @@ trait ImageFields {
     "croppedBy" -> "exports.author",
     "filename"  -> "uploadInfo.filename",
     "photoshoot"-> photoshootField("title"),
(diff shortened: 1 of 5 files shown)


AWare:
I'm not sure that this is more consistent, we use label and labels for the chip and viewer, and person/people follows this pattern¹. I'm not sure whether plural chips are a good idea or not. 


1. Technically _persons_ would be following this better, but it's a horrible word.

paperboyo:
> we use label and labels for the chip and viewer

Ideal it is not. But better than using two words for the same thing. And we can move to `labels` chip as soon as we support multiple values (this is beyond me).

I think it’s still better, cos it’s the same thing called the same in viewer, in 🍟 and in `has`. Less confusion. No?

paperboyo:
> I'm not sure that this is more consistent, we use label and labels for the chip and viewer, and person/people follows this pattern

More to this. I think this distinction makes for a less predictable experience between what’s in the UI, what the chip shows and what `has:` shows. It applies only to fields that can hold multiple values (hence singular/plural) and person/people is but a most stark example.

I propose all these fields (`people`, `labels`, `keywords`, `subjects`) that have this discrepancy between how they display in UI and in API and in `has:` searches (plural) and how they display in their own chip search (singular) be made consistent by making their own 🍟 accept multiple values (with comma?) and then be switched to plural everywhere.

Two downsides to this: a) ppl’s bookmarks may stop working (unless we do something for them to be translated); b) in XMP many more fields can contain multiple values (eg byline or description in different languages): does it mean every single metadata field should be presented in plural form (bit much).

paperboyo:
> We should be consistent for the `has` search.

We cannot be consistent if it’s called two things. Unless we allow two aliases (dumb).

AWare:
Allowing `has` to support the singular and plural forms of all terms sounds entirely reasonable.

paperboyo:
> Allowing `has` to support the singular and plural forms of all terms sounds entirely reasonable.

No. If it’s alias to the same thing: it’s unnecessary complication and confusion (would work only for fileds that support multi values, not for others). If it’s to be able to search for pics that has one value vs. pics that have more than one – this is even more absurd.

paperboyo:
I have reverted all the renaming to go with what Alex suggested. I have retained a `has:` alias to be `people` to be consistent with `has:labels`, `has:keywords` etc.
One day we may want to support multiple values and get rid of the inconsistency.

jennygrahamjones:
@paperboyo Could you change the description to just explain what the PR now includes? Also how to test if that's relevant 🙏 It will make it easier to understand in the future to understand specifically what this PR introduced.

paperboyo:
Good point, @jennygrahamjones, fixed now. Left the rest for archeologists.

PR #2966, opened 2020-06-19
Revert "Revert "Feature/add people in image metadata""

AWare:
Reverts guardian/grid#2961

https://github.com/guardian/grid/pull/2958
https://github.com/guardian/grid/pull/2966


AWare:
So when this was merged the first time, it triggered a bug in the metadata diff which is fixed here. https://github.com/guardian/grid/pull/2958

PR #2972, opened 2020-06-30
PR and Handout to be Restricted by default

paperboyo:
This is a Picture Desk request. These two categories of Usage Rights seem to often be (mis)used as a _I’m-not-sure-what-this-is-let-me-choose-first-one_ option. This will prevent that.

## Tested?
- [ ] locally
- [x] on TEST


akash1810:
Nice! Could we add a little more context to the PR description please? Why is it needed? What problem does it solve? etc.

PR #2973, opened 2020-07-01
ignore usage updates from CAPI that cannot have originated from Grid

akash1810:
## What does this change?
Grid image IDs don't follow the pattern `gu-image-`. These represent images that pre-date Grid. We can safely ignore these messages when we see them on the CAPI streams.

Less noise in logs.

n/a

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/usage/MediaUsage.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/usage/MediaUsage.scala
@@ -25,9 +25,13 @@ case class MediaUsage(
 ) {
 
   def isGridLikeId: Boolean = {
-    // remove events from CAPI that represent images previous to Grid existing
-    Logger.info(s"MediaId $mediaId doesn't look like a Grid image. Ignoring usage $usageId.")
-    !mediaId.startsWith("gu-image-")
+    if (mediaId.startsWith("gu-image-")) {
+      // remove events from CAPI that represent images previous to Grid existing
+      Logger.info(s"MediaId $mediaId doesn't look like a Grid image. Ignoring usage $usageId.")
+      false
+    } else {
+      true
+    }
   }
 
   def isRemoved: Boolean = (for {
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/usage/MediaUsage.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/usage/MediaUsage.scala
@@ -1,27 +1,35 @@
 package com.gu.mediaservice.model.usage
 
 import org.joda.time.DateTime
+import play.api.Logger
 
 case class UsageId(id: String) {
   override def toString = id
 }
 
 case class MediaUsage(
-                       usageId: UsageId,
-                       grouping: String,
-                       mediaId: String,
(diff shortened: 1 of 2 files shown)


paperboyo:
This is great!

In theory we could match up (some!) by colocation of `gu-image` and `picdarURN` (was doing this manually and Gustaff tries), eg

[This](https://www.theguardian.com/education/cartoon/2014/jun/30/ros-asquith-lines-school-place-shortages-bite) gives
`https://content.guardianapis.com/education/cartoon/2014/jun/30/ros-asquith-lines-school-place-shortages-bite?format=json&api-key=[YOUR KEY]&show-elements=all&show-atoms=all&show-rights=all&show-fields=all&show-tags=all&show-blocks=all&show-references=all`

and contains `"mediaId": "gu-image-440652880"` and `"picdarUrn": "GD*44886439"` that matches up to same [image in much nicer resolution](https://media.gutools.co.uk/search?query=GD*44886439&nonFree=true).

But I guess that’s crazy?

common-lib/src/main/scala/com/gu/mediaservice/model/usage/MediaUsage.scala
```diff
@@ -1,26 +1,34 @@
...
+  frontUsageMetadata: Option[FrontUsageMetadata],
+  lastModified: DateTime,
+  dateAdded: Option[DateTime] = None,
+  dateRemoved: Option[DateTime] = None
+) {
+
+  def isGridLikeId: Boolean = {
+    // remove events from CAPI that represent images previous to Grid existing
+    Logger.info(s"MediaId $mediaId doesn't look like a Grid image. Ignoring usage $usageId.")
+    !mediaId.startsWith("gu-image-")
```
jennygrahamjones, line 29: Will the log message here always be appropriate? I may be missing context, but it looks like this will write regardless of whether the image looks like a Grid image.

akash1810: Excellent point @jennygrahamjones ! Have corrected this now.

jennygrahamjones (approved):

akash1810:
> This is great!
> 
> In theory we could match up (some!) by colocation of `gu-image` and `picdarURN` (was doing this manually and Gustaff tries), eg
> 
> [This](https://www.theguardian.com/education/cartoon/2014/jun/30/ros-asquith-lines-school-place-shortages-bite) gives
> `https://content.guardianapis.com/education/cartoon/2014/jun/30/ros-asquith-lines-school-place-shortages-bite?format=json&api-key=[YOUR KEY]&show-elements=all&show-atoms=all&show-rights=all&show-fields=all&show-tags=all&show-blocks=all&show-references=all`
> 
> and contains `"mediaId": "gu-image-440652880"` and `"picdarUrn": "GD*44886439"` that matches up to same [image in much nicer resolution](https://media.gutools.co.uk/search?query=GD*44886439&nonFree=true).
> 
> But I guess that’s crazy?

I think this is a valid use-case, however I wouldn't classify it as a usage. Maybe a secondary usage or something to denote that it's not directly tied to the image in Grid.

PR #2974, opened 2020-07-02
Guard against null image ids

aug24:
## What does this change?
We have repeated errors in DynamoDB client where it appears the provided image id is null.  This is an attempt to report the problem more clearly to facilitate debugging.

Unclear AWS Exception is replaced with clear custom exception.

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/usage/app/model/UsageTable.scala
+++ b/usage/app/model/UsageTable.scala
@@ -35,6 +35,10 @@ class UsageTable(config: UsageConfig) extends DynamoDB(config, config.usageRecor
   }
 
   def queryByImageId(id: String): Future[Set[MediaUsage]] = Future {
+
+    if (id.isEmpty)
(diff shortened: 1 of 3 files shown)


usage/app/model/UsageTable.scala
```diff
@@ -35,6 +35,10 @@ class UsageTable(config: UsageConfig) extends DynamoDB(config, config.usageRecor
   }
 
   def queryByImageId(id: String): Future[Set[MediaUsage]] = Future {
+
+    if (id.isEmpty)
```
jennygrahamjones, line 39: Is it worth doing a `.trim` here, on the off chance that `"  "` is passed in as the ID?

jonathonherbert: Looking back, this is a heroic spot @jennygrahamjones! 👍

PR #2977, opened 2020-07-11
Download to usages

akash1810:
## What does this change?
This is https://github.com/guardian/grid/pull/2670 rebased with `master` to bring it back into a mergeable state. I can't work out how to push to that particular branch, presumably because ownership resides in the fork.

From the original PR:

> Usages has a new 'downloads' type which records who has downloaded what and when. This is in a similar way to how the other usages are set up, and takes an API call to mark an image as downloaded (along with who downloaded it and a time stamp).
Media API has been updated so when downloads occur it will send a async request over to usages to mark it as download. This is a best effort call and no error handling is in place, as I don't think we should be too bothered if this call fails or not.

Sequence of events is:
1. User clicks download, issuing a `GET` to `/download` in `media-api`
1. Server fetches image from S3 and issues a `POST` to `usages`
1. On receipt of the `POST` request, `usages` persists information in dynamodb (pre-existing behaviour)
1. `usages` places a message on kinesis for `thrall` to consume (pre-existing behaviour)
1. `thrall` consumes message and indexes it into elasticsearch (pre-existing behaviour)

I not sure this is the correct way to achieve this as it breaks the implicit rule in Grid that two services don't talk directly to each other, enqueuing a message to be processed asynchronously is preferred:

1. User clicks download, issuing a `GET` to `/download` in `media-api`
1. Server fetches image from S3 and enqueues message for `usages` to consume
1. `usages` consumes message and persists information in dynamodb (pre-existing behaviour)
1. `usages` places a message on kinesis for `thrall` to consume (pre-existing behaviour)
1. `thrall` consumes message and indexes it into elasticsearch (pre-existing behaviour)

I'd suggest we do not do this in this PR however to keep the diff small and as close to the original PR as possible.

Note: Images with usages are kept indefinitely. Therefore images that are downloaded from Grid will begin to be persisted with this change. This might not be the behaviour we want. I'd recommend we talk with the picture desk about this and address it in a separate PR if necessary. We're currently keeping the vast majority of images already, so it might not be an issue.

Image downloads are recorded as a usage both in the usage-api (dynamodb) and media-api (elasticsearch).

**Kahuna**
image

**Usage API**
image

## Tested?
- [x] locally
- [x] on TEST

## Deploy sequence
This requires an update of the elasticsearch mapping. This is a manual step so care is needed when deploying.

1. Disable continuous deployment
1. Merge
1. Update mappings
1. Deploy
1. Enable continuous deployment

### DIFF
--- a/usage/app/controllers/UsageApi.scala
+++ b/usage/app/controllers/UsageApi.scala
@@ -192,9 +192,9 @@ class UsageApi(auth: Authentication, usageTable: UsageTable, usageGroup: UsageGr
         errorKey = "download-usage-parse-failed",
         errorMessage = JsError.toJson(e).toString
       ),
-      fur => {
-        GridLogger.info("recording download usage", req.user.accessor, fur.mediaId)
-        val group = usageGroup.build(fur)
+      usageRequest => {
+        GridLogger.info("recording download usage", req.user.accessor, usageRequest.mediaId)
+        val group = usageGroup.build(usageRequest)
         usageRecorder.usageSubject.onNext(group)
         Accepted
       }
--- a/usage/app/controllers/UsageApi.scala
+++ b/usage/app/controllers/UsageApi.scala
@@ -184,6 +184,23 @@ class UsageApi(auth: Authentication, usageTable: UsageTable, usageGroup: UsageGr
     )
   }}
 
+  def setDownloadUsages() = auth(parse.json) { req => {
+    val request = (req.body \ "data").validate[DownloadUsageRequest]
+    request.fold(
+      e => respondError(
+        BadRequest,
+        errorKey = "download-usage-parse-failed",
+        errorMessage = JsError.toJson(e).toString
+      ),
+      fur => {
+        GridLogger.info("recording download usage", req.user.accessor, fur.mediaId)
+        val group = usageGroup.build(fur)
+        usageRecorder.usageSubject.onNext(group)
+        Accepted
+      }
+    )
+  }}
+
   def deleteUsages(mediaId: String) = auth.async {
     usageTable.queryByImageId(mediaId).map(usages => {
       usages.foreach(usageTable.deleteRecord)
--- a/usage/app/lib/MediaUsageBuilder.scala
+++ b/usage/app/lib/MediaUsageBuilder.scala
@@ -17,6 +17,7 @@ object MediaUsageBuilder {
     None,
     None,
     None,
+    None,
     printUsage.dateAdded
   )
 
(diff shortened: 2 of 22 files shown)


usage/app/controllers/UsageApi.scala
```diff
@@ -184,6 +184,23 @@ class UsageApi(auth: Authentication, usageTable: UsageTable, usageGroup: UsageGr
...
 
+  def setDownloadUsages() = auth(parse.json) { req => {
+    val request = (req.body \ "data").validate[DownloadUsageRequest]
+    request.fold(
+      e => respondError(
+        BadRequest,
+        errorKey = "download-usage-parse-failed",
+        errorMessage = JsError.toJson(e).toString
+      ),
+      fur => {
```
jonathonherbert, line 195: What's ... what's a fur? 😅

usage/app/lib/MediaUsageBuilder.scala
```diff
@@ -69,8 +72,27 @@ object MediaUsageBuilder {
...
+    val usageId = UsageIdBuilder.build(downloadUsageRequest)
+
+    MediaUsage (
+      usageId,
+      groupId,
+      downloadUsageRequest.mediaId,
+      DownloadUsage,
+      mediaType = "image",
+      downloadUsageRequest.status,
+      printUsageMetadata = None,
```
jonathonherbert, line 90: Just out of interest – would it be possible/desirable to represent these as a sum type in some future refactor, rather than having `n` metadata fields for each usage type? It would reduce the number of lines changes when we added new metadata types.

usage/app/controllers/UsageApi.scala
akash1810: Great question! Might be too late to ask @jamie-pitts the original author of the PR (a cost of us taking so long to get to it, sorry!!). It's likely to be a typo for `dur` as it's of type `DownloadUsageRequest`. A more intention revealing name would be good here, for sure, I'll do this on this PR.

akash1810: Ah, it's copy pasted from the FrontsUsageRequest handling above this, so not necessarily a typo.

paperboyo:
May not be important, given that I think we should kill low-res completely, but will this count and/or distinguish low-res downloads?

usage/app/lib/MediaUsageBuilder.scala
akash1810: Yes, that would be ace! There would be two approaches to this, both with +/-:

- Change case class and database model and reindex database to ensure pre-existing data gets transformed into new shape
- Change case class, keep database model as is and incur the cost of the slight hidden complexity of different representations of data across domains

Either make sense, I think and worth the cost of a sum type as it yields simpler code.

akash1810:
> May not be important, given that I think we should kill low-res completely, but will this count and/or distinguish low-res downloads?

Low res will be recorded as a usage. Usages will not currently distinguish what gets downloaded.

jamie-pitts:
Awesome to see this is getting merged in! Hope all is going well there :)

PR #2984, opened 2020-07-23
Add config value for whether downloads should be recorded as usages

jennygrahamjones:
## What does this change?
Following on from https://github.com/guardian/grid/pull/2977, this PR introduces a config value that can be used to determine whether downloads should be recorded as usages. By default, this value is set to `false` and the Grid will not record downloads as usages. 

Running the Grid locally, upload an image. Click the 'Download' button, and observe that the Usages tab remains as in the image below:

Screenshot 2020-07-23 at 12 43 14

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/dev/script/generate-dot-properties/service-config.js
+++ b/dev/script/generate-dot-properties/service-config.js
@@ -120,6 +120,7 @@ function getMediaApiConfig(config) {
         |quota.store.key=rcs-quota.json
         |security.cors.allowedOrigins=${getCorsAllowedOriginString(config)}
         |metrics.request.enabled=false
+        |image.record.download=false
         |`;
 }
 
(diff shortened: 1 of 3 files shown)


jennygrahamjones:
[Grid's download_usage should not persist images / prevent deletions](https://trello.com/c/4sVEkyy7/1981-grids-downloadusage-should-not-persist-images-prevent-deletions)

dev/script/generate-dot-properties/service-config.js
```diff
@@ -120,6 +120,7 @@ function getMediaApiConfig(config) {
         |quota.store.key=rcs-quota.json
         |security.cors.allowedOrigins=${getCorsAllowedOriginString(config)}
         |metrics.request.enabled=false
+        |image.record.download=false
```
AWare, line 123: Maybe this should be true for dev?

jennygrahamjones: Are you thinking it might be useful for something in particular when developing locally?

PR #2986, opened 2020-07-29
Highlight search terms in the preview

AWare:
## What does this change?

This highlights the search term in the mini image preview.

An image may be a search hit if the search term appears late in the description, if it does, the chance of it being the correct image is diminished. 

This may be better to match the entire term, instead of by word in the query. This is just a proof of concept. (TODO: make it work/capitalisation)

Does it help avoid accidental wrong images and is it liked by end users.

image
image

## Who should look at this?
@guardian/digital-cms 

## Tested?
- [X] locally
- [ ] on TEST

### DIFF
--- a/kahuna/public/js/preview/image.html
+++ b/kahuna/public/js/preview/image.html
@@ -82,7 +82,15 @@
         <p class="preview__description"
            title="{{ctrl.image.data.metadata.description || ctrl.image.data.metadata.title}}">
             <!-- Ensure contents in P to maintain height -->
-            {{ctrl.image.data.metadata.description || ctrl.image.data.metadata.title || '&nbsp;'}}
+            <span ng-repeat="word in ctrl.descriptionHighlighted">
+              <span ng-if="word.highlight">
+                <mark>{{word.term}}</mark>
+              </span>
+              <span ng-if="!word.highlight">
+                {{word.term}}
+              </span>
+
+            </span>
         </p>
     </div>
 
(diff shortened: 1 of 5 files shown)


paperboyo:
Couple of observations:
- yellow highlight feels foreign and would be distractive on a page full of results, so, maybe, standard Grid-blue
- what if the term highlighted is later in the description? do we show that later part then? what about the diminished value of always showing the beginning of description?
- often hits are not in the description. then, there is no way to signal it in Browser

All in all, not sure about that, as indicated IRL. Would need to apply to Viewer (pic page) too: highlighting terms everywhere: desc, keywords, restriction text etc.

BTW, it’s not clear which fields free-text search is actually looking at. I spent days trying to come up with a list (and with different rules/indexers per field) and capitulated…

AWare:
(forgive the ooh reply, I'm on a/l tomorrow so will pick this back up for a bit on Friday and hopefully get it onto TEST)

_> yellow highlight feels foreign and would be distractive on a page full of results, so, maybe, standard Grid-blue_

that's sensible, this is just to show what the effect would be so we can see if it's useful

_> what if the term highlighted is later in the description?_ 

the idea is that the absence of a term from the description is a signal that they should inspect the image more closely if they want to use it, this is just a very quick heuristic; and we'll see if it's fruitful on TEST 

_> Would need to apply to Viewer (pic page) too: highlighting terms everywhere: desc, keywords, restriction text etc._

quite probably, this is just a quick sketch to evaluate the idea.

PR #2988, opened 2020-07-30
Search for usage references

akash1810:
## What does this change?
Adds a new chip `usages@reference` to search for a [UsageReference](https://github.com/guardian/grid/blob/5ca61af6ecaecffc4acc142511bd596dd698b693/common-lib/src/main/scala/com/gu/mediaservice/model/usage/UsageReference.scala).

We'll search against the usage reference `name` and `uri` properties. Not every usage reference has a `name` and not every usage reference has a `uri`.

A digital usage, recorded from the CMS (Composer) creates these references:
```json
"references": [
  {
    "type": "frontend",
    "uri": "https://www.theguardian.com/sport/live/2020/jul/30/england-v-ireland-first-cricket-odi-live",
    "name": "England v Ireland: first ODI – live!"
  },
  {
    "type": "composer",
    "uri": "https://composer.gutools.co.uk/content/5f2274f98f08f42f5d729f79"
  }
]
```

A print usage creates these references:
```json
"references": [
  {
    "type": "indesign",
    "name": "2020-07-30, Guardian, World (News), Page 29"
  }
]
```

Searching both fields means we can keep the UI simple; a search for a paper reference is the same as a search for a URL reference.

**Note**
This moves to using an `ExactMatchValue` instead of a `MatchValue`. Previously we were using a `MatchValue` which would result in the use of the englishSStemmerAnalyzer analyzer ([LOC link](https://github.com/guardian/grid/blob/5ca61af6ecaecffc4acc142511bd596dd698b693/media-api/app/lib/elasticsearch/QueryBuilder.scala#L22-L25)). This analyzer does a number of things, one of which is ignoring forward slashes ("/"). It's a fuzzy match.

When searching for a usage reference (against field `usages.references.uri` or `usages.references.name`), we'd want to be able to search by Composer URL for ease. The `MatchValue` doesn't work here.

`ExactMatchValue` is actually preferred in a nested query; you'd use a nested query when looking for a specific value, thus performing a fuzzy match is less performant than a term match.

See https://discuss.elastic.co/t/query-string-query-forward-slash-how-to-explained/16637.

**Alternative solutions**
We could search against the `composerUrl` in a [DigitalUsageMetadata](https://github.com/guardian/grid/blob/5ca61af6ecaecffc4acc142511bd596dd698b693/common-lib/src/main/scala/com/gu/mediaservice/model/usage/DigitalUsageMetadata.scala#L11), however that's very specific to the Guardian and to Composer. Using usage references keeps things a bit more generic and provides the benefit of searching for print usage references.

We can search for images used in a particular bit of content, using either the Composer URL, article URL or the paper reference.

Other nested queries (all the usage chips) continue to work as expected.

**Grid search and its corresponding Composer piece**
image

image

**Searching for print reference**
image

## Who should look at this?

@guardian/digital-cms @blishen 

## Tested?
- [x] locally
- [x] on TEST

### DIFF
--- a/kahuna/public/js/search/structured-query/query-suggestions.js
+++ b/kahuna/public/js/search/structured-query/query-suggestions.js
@@ -31,6 +31,7 @@ export const filterFields = [
     'usages@>added',
     'usages@platform',
     'usages@status',
+    'usages@reference',
     'has',
     'croppedBy',
     'filename',
--- a/media-api/app/lib/querysyntax/QuerySyntax.scala
+++ b/media-api/app/lib/querysyntax/QuerySyntax.scala
@@ -94,7 +94,7 @@ class QuerySyntax(val input: ParserInput) extends Parser with ImageFields {
 
   def AllowedParentFieldName = rule { "usages" }
   def AllowedNestedFieldName = rule {
-    "status" | "platform" | "section" | "publication" | "orderedBy"
+    "status" | "platform" | "section" | "publication" | "orderedBy" | "reference"
   }
   def AllowedFieldName = rule {
     "illustrator" |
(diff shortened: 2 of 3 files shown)


paperboyo:
💎 Will it work with partial matches eg. `5f2274f98f08f42f5d729f79` instead of `"https://composer.gutools.co.uk/content/5f2274f98f08f42f5d729f79"`?

PR #2990, opened 2020-08-19
Proposal: Add cover card crop for editions.

philmcmahon:
## What does this change?
This adds a new crop size for use in the [editions card builder](https://editions-card-builder.gutools.co.uk/). Cover cards are a fixed size, so it doesn't make sense for the production team to be using a freeform crop.

The obvious downside to this is that this is adding a very guardian specific (and one team of the guardian at that) functionality to a general purpose image search tool. So maybe it could go behind a permission or client side feature flag or something...or maybe it doesn't belong in the grid.

We ask the editions production team if it has saved them some time.

Screenshot 2020-08-19 at 17 36 24

## Who should look at this?

Discussed initially with @AWare , maybe @paperboyo might also have some thoughts? 

## Tested?
- [ ] locally
- [x] on TEST

### DIFF
--- a/kahuna/public/js/util/crop.js
+++ b/kahuna/public/js/util/crop.js
@@ -8,7 +8,7 @@ const portrait = {key: 'portrait', ratio: 4 / 5, ratioString: '4:5'};
 const video = {key: 'video', ratio: 16 / 9, ratioString: '16:9'};
 const square = {key: 'square', ratio: 1, ratioString: '1:1'};
 const freeform = {key: 'freeform', ratio: null};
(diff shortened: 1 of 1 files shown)


akash1810:
> The obvious downside to this is that this is adding a very guardian specific (and one team of the guardian at that) functionality to a general purpose image search tool. So maybe it could go behind a permission or client side feature flag or something...or maybe it doesn't belong in the grid.

Grid currently supports a crop type query string ([introduction PR](https://github.com/guardian/grid/pull/2072)). Maybe it's worth having supporting another param for crop ratio to keep Grid generic?

philmcmahon:
Thanks @akash1810 I'll close this for now and look at doing that!

PR #2991, opened 2020-08-25
Comment out Jonny Weeks from Staff list

AWare:
Jonny's work comes to us externally as well as internally, so it's not safe to mark him as staff for now.


paperboyo (approved):
Let’s see if he prefers it…

Issue #2676, opened 2019-11-14
Merge config change contributions


mbarton:
The following PRs refactoring out some config have been open for quite a while:

- [ ] https://github.com/guardian/grid/pull/2995 (supercedes https://github.com/guardian/grid/pull/2612)
- [ ] https://github.com/guardian/grid/pull/2614
- [ ] https://github.com/guardian/grid/pull/2618 (depends on previous two)
- [ ] https://github.com/guardian/grid/pull/2620 (depends on previous three)
- [ ] https://github.com/guardian/grid/pull/2622 (depends on previous four)

We'd like to get them merged but will need to verify that we are still correctly parsing the affected fields and assigning rights correctly.

Here's the plan for merging and testing each one:

- Extract out JSON config from the current code on `master`
- Run in TEST with config and upload an image
- Check the API response is the same
  - NB: signed URLs may be different but that won't matter

cc @jamie-pitts @felixcs1

jamie-pitts:
That's great to hear @mbarton appreciate there's quite a bit of testing and work to do on your side for these, so no worries for them being open for a while. Let us know if you have any problems!

akash1810:
> Check the API response is the same

We can use [grid-cli](https://github.com/guardian/grid-cli) for this 🎉.  Here's a little script - https://gist.github.com/akash1810/c681d57155f8e48e95214ca39a8f8cdf.

paperboyo:
As the PRs mentioned are all closed or superseded, I will close this even though probably more configuration needs to move out of code. But it won’t be tracked by this issue.

PR #2612, opened 2019-08-05
Extract hardcoded names in MetadataConfig

felixcs1:
## What does this change?
This removes all hard coded photographers names in MetaDataConfig. Instead these are now read from an s3 file into MetaDataStore and parsed as a MetadataConfig class. This allows updates to these lists to be made without the need for a redeploy.

If the json cannot be parsed on startup an error is thrown which stops the app. But after intial load if subsequent parsing fails (if someone makes the json file invalid) only a warning is logged and the last successful load is kept in the store.

The `s3.config.bucket` is where the json should be held, I have sent the guardian's version of this to @akash1810.

## How can success be measured?
Put the json file in the `s3.config.bucket` with the name of `photographers.json`
Add the bucket name to your `/etc/gu` properties for image-loader, api and metadata-editor: `s3.config.bucket=<bucket name>`. 

Updating the json config in `s3.config.bucket` will make the photographers names update (after 10 mins). 

## Tested?
- [x] locally
- [ ] on TEST

PR #2995, opened 2020-09-04
Extract hardcoded names in MetadataConfig (2020 redux to merge conflicts)

mbarton:
**NB: #2612 with merge conflicts fixed.  It also creates an empty config file as part of setup to support local dev**

## What does this change?
This removes all hard coded photographers names in MetaDataConfig. Instead these are now read from an s3 file into MetaDataStore and parsed as a MetadataConfig class. This allows updates to these lists to be made without the need for a redeploy.

If the json cannot be parsed on startup an error is thrown which stops the app. But after intial load if subsequent parsing fails (if someone makes the json file invalid) only a warning is logged and the last successful load is kept in the store.

The `s3.config.bucket` is where the json should be held, I have sent the guardian's version of this to @akash1810.

Put the json file in the `s3.config.bucket` with the name of `photographers.json`
Add the bucket name to your `/etc/gu` properties for image-loader, api and metadata-editor: `s3.config.bucket=<bucket name>`. 

Updating the json config in `s3.config.bucket` will make the photographers names update (after 10 mins). 

## Tested?
- [x] locally
- [x] on TEST
- [ ] test an image following instructions here: https://github.com/guardian/grid/issues/2676


mbarton:
As this change reads JSON, we lose the ability to write comments against the metadata which can be quite handy (for example [Jonny Weeks sometimes being staff and sometimes not](https://github.com/guardian/grid/blob/331dabaaab24cee33bb80d2f41158cf05804aca9/common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala#L51)).

Not a blocker for merging this but would be nice perhaps to discuss at the next catch-up whether we could use YAML or similar to support comments @akash1810?

mbarton:
I've dployed this PR to `TEST` and it still seems to work, agency images still coming in. We should do the double check described in https://github.com/guardian/grid/issues/2676 as well

mbarton:
Yep I think we should get all the BBC PRs that pull out config merged and then consider YAML. From a quick glance over it feels like maybe an embedded scripting engine running Guardian and BBC-specific scripts might be the proper solution. The scripts could modify the fields on the entire as it saw fit. Things like `staffIllustrator` feel Guardian use-case specific to me

mbarton:
I've tried the check in `TEST` with this image: https://media.gutools.co.uk/images/bdba7a463bb47188360e663d521afcc5e971d9a8

Usage rights for the image in `PROD` show:

```
category: "staff-photographer"
photographer: "Mateusz Karpow"
publication: "The Guardian"
```

@paperboyo you're not in the [list of Staff Photographs](https://github.com/guardian/grid/blob/f092deff989cd218fc7aef869d936918a3fac876/common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala#L42) so was that added manually after you uploaded the photo?

I picked that one for testing because it was hackday and downloads now count as usages so I didn't want to confuse anyone with proper PROD data. Am I being over-cautious? Any suggestions for what images to double check are welcome, otherwise I think this PR is ready to ship, pending approval

paperboyo:
> was that added manually after you uploaded the photo?

Exactly: `"originalUsageRights": {},`. Same one on TEST, correctly has no rights, but just try any from list of snappers or agencies or something. 

> Am I being over-cautious?

Definitely! Do whatever you like with TEST!

image-loader/test/scala/model/ProjectorTest.scala
```diff
@@ -33,7 +34,9 @@ class ProjectorTest extends FunSuite with Matchers with ScalaFutures with Mockit
   private val config = ImageUploadOpsCfg(new File("/tmp"), 256, 85d, Nil, "img-bucket", "thumb-bucket")
 
   private val s3 = mock[AmazonS3]
-  private val projector = new Projector(config, s3, imageOperations)
+
+  // TODO MRB: put these behind an interface at the top level so we can mock them here
+  private val projector = new Projector(config, s3, imageOperations, null, null)
```
mbarton, line 39: The actual test here is ignored, it was presumably useful when we built the projection service during the missing images work

dev/script/setup.sh
```diff
@@ -226,6 +226,31 @@ uploadApiKey() {
...
+EOF
+
+  aws s3 cp "$target" \
+    "s3://$configBucket/photographers.json" \
+    --endpoint-url $LOCALSTACK_ENDPOINT
+
+  rm "$target"
+  echo "  uploaded"
+}
+
```
jonathonherbert, line 253: 👏

image-loader/test/scala/model/ProjectorTest.scala
jonathonherbert: hmm 😬

metadata-editor/app/model/UsageRightsProperty.scala
```diff
@@ -19,22 +19,25 @@ case class UsageRightsProperty(
   examples: Option[String] = None
 )
 
-
 object UsageRightsProperty {
   type OptionsMap = Map[String, List[String]]
```
jonathonherbert, line 23: It's not directly related to this PR, but I found this very difficult to read – not Scala's `Option`, but a small-O option for the `UsageRightsProperty`, related to – usage rights, but I'm still not sure what this is exactly! Doesn't help that I don't understand the domain too well, but a clearer name might make things easier. Happy to submit another PR if needed.

metadata-editor/app/model/UsageRightsProperty.scala
```diff
@@ -19,22 +19,25 @@ case class UsageRightsProperty(
...
-  val props: List[(UsageRightsSpec) => List[UsageRightsProperty]] =
+  val props: List[(UsageRightsSpec, MetadataConfig) => List[UsageRightsProperty]] =
     List(categoryUsageRightsProperties, restrictionProperties)
 
-  def getPropertiesForSpec(u: UsageRightsSpec): List[UsageRightsProperty] = props.flatMap(f => f(u))
+  def companyListToMap(companies: List[Company]): OptionsMap = Map(companies
+    .map(c => c.name -> c.photographers): _*)
+
+  def optionsFromCompanyList(companies: List[Company]): Options = sortList(companyListToMap(companies).keys.toList)
+
```
jonathonherbert, line 39: Possible I've misunderstood – why do we need to go through an intermediate map to get a sorted list of company names?

jonathonherbert:
NB: leaving this review for now as per @sihil –

> As it stands we need to convert our rules to JSON. I don't believe that we should do that ...
>
>
> JSON is terrible for configuration (unable to comment stuff) and also we might want to do things that you can do in code but not in JSON without extending the config. Finally you can't write tests for JSON as readily to ensure that it is parsable and does the right thing (I mean I guess you can, but the compiler is there for a reason....)

mbarton:
Discussion about the right way to do this continues in https://github.com/guardian/grid/pull/3004

PR #2614, opened 2019-08-14
Feature/extract config usage rights

felixcs1:
## What does this change?
Similar to [pull 2612](https://github.com/guardian/grid/pull/2612) but for usageRightsConfig

- Extracts `UsageRightsConfig.scala` into a json file on s3 that is read in using a store. Allows suppliers list to be customised without re deploy (and BBC to have different supplier list without a code difference in our fork!)  
- Also makes the UsageRightsSpec in `UsageRights.scala` list customisable from the json file.

As with  [pull 2612](https://github.com/guardian/grid/pull/2612), if the json cannot be parsed on startup an error is thrown which stops the app. But after intial load if subsequent parsing fails (if someone makes the json file invalid) only a warning is logged and the last successful load is kept in the store.

## How can success be measured?
Put the json file in the s3.config.bucket with the name of `usage_rights.json`
ensure the bucket name is added to your /etc/gu properties for api and metadata-editor: `s3.config.bucket=<bucket name>`.

Updating the json config in s3.config.bucket will update supplier list  (after 10 mins).
## Tested?
- [x] locally
- [ ] on TEST

PR #2997, opened 2020-09-07
Extract Usage Rights to config (2020 redux to fix conflicts)

mbarton:
NB: #2614 with merge conflicts fixed. Depends on #2995, don't merge until that PR is merged!

## What does this change?
Similar to [pull 2612](https://github.com/guardian/grid/pull/2612) but for usageRightsConfig

- Extracts `UsageRightsConfig.scala` into a json file on s3 that is read in using a store. Allows suppliers list to be customised without re deploy (and BBC to have different supplier list without a code difference in our fork!)  
- Also makes the UsageRightsSpec in `UsageRights.scala` list customisable from the json file.

As with  [pull 2612](https://github.com/guardian/grid/pull/2612), if the json cannot be parsed on startup an error is thrown which stops the app. But after intial load if subsequent parsing fails (if someone makes the json file invalid) only a warning is logged and the last successful load is kept in the store.

Put the json file in the s3.config.bucket with the name of `usage_rights.json`
ensure the bucket name is added to your /etc/gu properties for api and metadata-editor: `s3.config.bucket=<bucket name>`.

Updating the json config in s3.config.bucket will update supplier list  (after 10 mins).

## Tested?
- [x] locally
- [ ] on TEST
- [ ] test an image following instructions here: https://github.com/guardian/grid/issues/2676


mbarton:
Discussion about the right way to do this continues in https://github.com/guardian/grid/pull/3004

PR #3002, opened 2020-09-14
fix build.sbt bug

fadhilijuma:
## What does this change?
This PR fixes a bug in built.sbt. It removes the try..catch. The idea is to either get the gitCommitId string value or return "Unknown" string.
Screenshot 2020-09-14 at 14 57 27
Screenshot 2020-09-14 at 14 58 15

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -227,10 +227,7 @@ def project(projectName: String, path: Option[String] = None): Project =
 val buildInfo = Seq(
   buildInfoKeys := Seq[BuildInfoKey](
     name,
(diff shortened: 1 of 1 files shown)


sihil:
Hi @fadhilijuma!

Thanks for this contribution. Whilst this works it unecessarily shells out to get a string. Based on our out of band conversations I don't believe there is a problem with `build.sbt` but rather a highlighting error in IntelliJ. A cleaner approach that also fixes the highlighting error might be #3003

sihil:
Closing this in preference for #3003

PR #3004, opened 2020-09-14
Make ImageProcessor pluggable

sihil:
## What does this change?
This is a PR that makes image processors configurable via late binding (i.e. you can include an ImageProcessor in the configuration file even if it wasn't available at the time the grid was compiled).

## Background
As the Grid becomes used by more than one organisation it is becoming apparent that we need to introduce a higher degree of customisation in a few areas, most notably authentication, authorisation and image processing business logic.

There are a couple of different approaches to this. One simple approach is to have different options baked into the Grid and selected via configuration. At the Guardian we use [Pan Domain Authentication](https://github.com/guardian/pan-domain-authentication) for authentication and [Permissions](https://github.com/guardian/permissions) for authorisation. We could do an internal refactor and add options for other users selected via config. 

Similarly we have a lot of code that transforms the metadata of images. A series of PRs extract this logic and make it configurable via JSON data files stored in S3. This approach mimics existing behaviour for API keys and quota configuration but is not ideal for a number of reasons:
 - JSON isn't a syntax designed for use by humans (it doesn't support comments and has a lot of quotes)
 - JSON config limits the flexibility to what is available in the code that applies the config whilst Scala code is Turing complete
 - The validation and testing story of Scala code is somewhat better as there is a compiler and test suites

An alternative to configuration and external data files is to create API hooks to allow users downstream of the main Grid codebase to confirgure the Grid to load their custom code at initialisation using late binding.

## Configuring image processing using late binding
First of all every image-processor-like entity is re-written to conform to the `ImageProcessor` interface (there were a couple of interfaces that essentially implemented `Image => Image` but used a different function name or only operated on a single field of `Image`). A `ComposedImageProcessor` is also created to make it trivial to chain a number of processors together into an `ImageProcessor` in code.

We then introduce a configuration option that allows you to specify which processors should be run. They will be applied in the order that you list them in the `image.processors` property. Each processor can take extra configuration if desired. The property will take a list that can be a mix of simple strings that denote imapge processors to load and richer objects that are able to include configuration.

`ImageProcessor`s are loaded by reflection. We can load companion object, classes that have a no arg constructor and classes that have a constructor with a single parameter of type `play.api.Configuration` which will receive the configuration for that processor.

### Example configurations
The current default config looks like this:
```
image.processors = [
  // no arg class
  "com.gu.mediaservice.lib.cleanup.GuardianMetadataCleaners",
  // companion object (note the trailing $)
  "com.gu.mediaservice.lib.cleanup.SupplierProcessors$"
]
```

You could also override this entirely - in this case mixing existing standard processors with your own processors, one of which takes futher HOCON configuration that is passed into the constructor as an instance of `play.api.Configuration`:
```
image.processors = [
  "com.gu.mediaservice.lib.cleanup.GettyXmpParser$",
  "com.yourdomain.YourImageProcessor",
  // a class with a single arg constructor taking a play Configuration object
  {
    className: "com.yourdomain.YourImageProcessorWithConfig"
    config: {
      configKey1: value1
      configKey2: {
        map: of
        values: "for fun and profit"
      }
    }
  }
]
```

Can the BBC subsequently apply the logic from their series of image processing PRs without having to convert our own config to JSON or similar.

## Remaining work
 - [x] Figure out a better way to allow configuration of custom plugins (converting all config to Typesafe Config and passing in a sub-tree of config looks promising)
 - [ ] Move existing builtin processors into a guardian subproject
 - [ ] Spike applying the BBC PRs into a BBC subproject as an alternate set of processors

## Tested?
- [X] locally
- [ ] on TEST

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/config/ImageProcessorLoader.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/config/ImageProcessorLoader.scala
@@ -1,0 +1,109 @@
+package com.gu.mediaservice.lib.config
+
+import com.gu.mediaservice.lib.cleanup.ImageProcessor
+import com.typesafe.config.ConfigException.BadValue
+import com.typesafe.config.{Config, ConfigObject, ConfigOrigin, ConfigValue, ConfigValueType}
+import com.typesafe.scalalogging.StrictLogging
+import play.api.{ConfigLoader, Configuration}
+
+import scala.collection.JavaConverters._
+import scala.language.existentials
+import scala.util.Try
+import scala.util.control.NonFatal
+
+object ImageProcessorLoader extends StrictLogging {
+  case class ImageProcessorConfigDetails(className: String, config: Option[Configuration], origin: ConfigOrigin, path: String)
+
+  implicit val imageProcessorsConfigLoader: ConfigLoader[Seq[ImageProcessor]] = (config: Config, path: String) => {
+    config
+      .getList(path)
+      .iterator()
+      .asScala.map { configValue =>
+      parseConfigValue(configValue, path)
+    }.map(loadImageProcessor).toList
+  }
+
+  implicit val imageProcessorConfigLoader: ConfigLoader[ImageProcessor] = (config: Config, path: String) => {
+    val configDetails = parseConfigValue(config.getValue(path), path)
+    loadImageProcessor(configDetails)
+  }
+
+  private def parseConfigValue(configValue: ConfigValue, path: String): ImageProcessorConfigDetails = {
+    configValue match {
+      case plainClass if plainClass.valueType == ConfigValueType.STRING =>
+        ImageProcessorConfigDetails(plainClass.unwrapped.asInstanceOf[String], None, plainClass.origin, path)
+      case withConfig:ConfigObject if validConfigObject(withConfig) =>
+        val config = withConfig.toConfig
+        val className = config.getString("className")
+        val processorConfig = config.getConfig("config")
+        ImageProcessorConfigDetails(className, Some(Configuration(processorConfig)), withConfig.origin, path)
+      case _ =>
+        throw new BadValue(configValue.origin, path, s"An image processor can either a class name (string) or object with className (string) and config (object) fields. This ${configValue.valueType} is not valid.")
+    }
+  }
+
+  private def validConfigObject(configObject: ConfigObject): Boolean = {
+    val config = configObject.toConfig
+    config.hasPath("className") && config.hasPath("config")
+  }
+
+  private def loadImageProcessor(details: ImageProcessorConfigDetails): ImageProcessor = {
+    ImageProcessorLoader
+      .loadImageProcessor(details.className, details.config.getOrElse(Configuration.empty))
+      .getOrElse { error =>
+        val configError = s"Unable to instantiate image processor from config: $error"
+        logger.error(configError)
+        throw new BadValue(details.origin, details.path, configError)
+      }
+  }
+
+  def loadImageProcessor(className: String, config: Configuration): Either[String, ImageProcessor] = {
+    for {
+      imageProcessorClass <- loadClass(className)
+      imageProcessorInstance <- instantiate(imageProcessorClass, config)
+    } yield imageProcessorInstance
+  }
+
+  private def loadClass(className: String): Either[String, Class[_]] = catchNonFatal(Class.forName(className)) {
+    case _: ClassNotFoundException => s"Unable to find image processor class $className"
+    case other =>
+      logger.error(s"Error whilst loading $className", other)
+      s"Unknown error whilst loading $className, check logs"
+  }
+
+  private def instantiate(imageProcessorClass: Class[_], config: Configuration): Either[String, ImageProcessor] = {
+    // Fail fast if config is provided but the specified image processor ignores it
+    def assertNoConfiguration[T](ok: Right[String, T]): Either[String, T] = {
+      if (config.keys.nonEmpty) {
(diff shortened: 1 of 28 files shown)


image-loader/app/lib/ImageLoaderConfig.scala
```diff
@@ -26,4 +30,25 @@ class ImageLoaderConfig(override val configuration: Configuration) extends Commo
   val transcodedMimeTypes: List[MimeType] = getStringSetFromProperties("transcoded.mime.types").toList.map(MimeType(_))
   val supportedMimeTypes: List[MimeType] = List(Jpeg, Png) ::: transcodedMimeTypes
 
+  val imageProcessors: List[ImageProcessor] = {
+    val ImageProcessorClass: Regex = "class:(.*)".r
```
mbarton, line 34: Could we use the fully qualified class name in all cases and avoid the regex?

image-loader/app/lib/ImageProcessorClassReflector.scala
```diff
@@ -0,0 +1,47 @@
...
+  def catchNonFatal[T](block: => T): Either[Throwable, T] = {
+    try {
+      Right(block)
+    } catch {
+      case NonFatal(e) => Left(e)
+    }
+  }
+
+  def loadImageProcessor(className: String, config: ImageProcessorConfig): Either[String, ImageProcessor] = {
+    val mirror = universe.runtimeMirror(getClass.getClassLoader)
```
mbarton, line 20: Oh wow, I had forgotten how verbose Scala reflection is. Lets just use the Java API instead: https://docs.oracle.com/javase/tutorial/reflect/member/ctorInstance.html

image-loader/app/lib/ImageProcessorClassReflector.scala
```diff
@@ -0,0 +1,47 @@
...
+        classSymbol <- catchNonFatal(mirror.staticClass(className)).left.map(_ => s"Class $className not found")
+        isImageProcessor = classSymbol.baseClasses.exists(symbol => symbol.fullName == classOf[ImageProcessor].getCanonicalName)
+        _ <- if (isImageProcessor) Right(()) else Left(s"Configured image processor $className does not extend ${classOf[ImageProcessor].getCanonicalName}")
+        classMirror = mirror.reflectClass(classSymbol)
+        constructor = classSymbol.primaryConstructor.asMethod
+        params = constructor.paramLists.head
+        hasConfigParam = {
+          val configParamType = Class.forName(params.head.info.toString, true, getClass.getClassLoader)
+          val assignable = classOf[ImageProcessorConfig].isAssignableFrom(configParamType)
+          System.err.println(s"${params.length} ${classOf[CommonConfig].getCanonicalName} ${configParamType} ${assignable}")
```
mbarton, line 33: Is logging not set up yet?

mbarton (commented):
Great stuff, this is exactly what I was thinking.

No major comments other than using Java reflection to avoid all the Scala runtime universe stuff

image-loader/app/lib/ImageProcessorClassReflector.scala
sihil: Hah - that was accidentally left in from testing in the console where logging wasn't initialised. Will remove.

image-loader/app/lib/ImageProcessorClassReflector.scala
sihil: Converted and basic unit tests added.

image-loader/app/lib/ImageLoaderConfig.scala
sihil: A few reasons I did this:
 - I wanted to differentiate between built in logic and customer supplied logic. 
 - Where it's possible to do a compile time binding we should, even if the config is still used. 
 - This will also make it easier to figure out where the classes here are used.
 - Finally referencing a singleton object is non-obvious, but I guess we can sort that by converting the SupplierProcessors to a class.

mbarton: Makes sense. Maybe a `builtin:` prefix for the ones we have, just to make it clear? Not a big deal though

mbarton (commented):
Looks good, I think this is ready to send around as a proposal

image-loader/app/lib/ImageLoaderConfig.scala
sihil: I changed my mind on this.

common-lib/src/main/scala/com/gu/mediaservice/lib/config/ImageProcessorLoader.scala
```diff
@@ -0,0 +1,109 @@
...
+    // Fail fast if config is provided but the specified image processor ignores it
+    def assertNoConfiguration[T](ok: Right[String, T]): Either[String, T] = {
+      if (config.keys.nonEmpty) {
+        Left(s"Attempt to initialise image processor ${imageProcessorClass.getCanonicalName} failed as configuration is provided but the constructor does not take configuration as an argument.")
+      } else {
+        ok
+      }
+    }
+
+    val maybeCompanionObject = Try(imageProcessorClass.getField("MODULE$").get(imageProcessorClass)).toOption
```
mbarton, line 84: Do we need to support companion objects? I'm worried this would break in future versions of Scala

mbarton (approved):

common-lib/src/main/scala/com/gu/mediaservice/lib/config/ImageProcessorLoader.scala
sihil: I originally didn't but a lot of the existing code is objects and actually that is what they should be if the processor is pure. I wrote a test for all of these cases so if it does break we'll notice swiftly. In that case we can either fix it to use the scala reflection or disable support.

PR #3006, opened 2020-09-15
Handle image file failures as exceptions

aug24:
## What does this change?
Implements better error handling and logging in the event of an image being provided which cannot be loaded.

Clearer exceptions in logs.

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/image-loader/app/controllers/ImageLoaderController.scala
+++ b/image-loader/app/controllers/ImageLoaderController.scala
@@ -86,9 +86,12 @@ class ImageLoaderController(auth: Authentication,
         case e =>
           Logger.error("loadImage request ended with a failure", e)
           (e match {
-            case e: UnsupportedMimeTypeException => FailureResponse.unsupportedMimeType(e, config.supportedMimeTypes)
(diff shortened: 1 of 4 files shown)


image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -86,9 +86,12 @@ class ImageLoaderController(auth: Authentication,
         case e =>
           Logger.error("loadImage request ended with a failure", e)
           (e match {
-            case e: UnsupportedMimeTypeException => FailureResponse.unsupportedMimeType(e, config.supportedMimeTypes)
-            case _: ImageProcessingException => FailureResponse.notAnImage(config.supportedMimeTypes)
-            case e => println(e.getMessage); InternalServerError(Json.obj("error" -> e.getMessage))
+
```
sihil, line 89: Did you mean to remove the handling for the `UnsupportedMimeTypeException`?

sihil (commented):
Missing exception handler and also omission of the IM code change. Not sure if either were intentional.

sihil (approved):

PR #3008, opened 2020-09-18
Rename `Staff image` to `GNM-owned` in tooltip

paperboyo:
@JonnyWeeks thinks this is confusing as it applies also to Commissioned:

image

Since it applies to all these:

image

let’s rename it to `GNM-owned`. Slight problem is: this is Guardian-specific. But we have this problem in the chip too, so can change it in all the places one day (?).

Opinions welcomed.


AWare:
If we moved this into the image model, this could be a good use of @sihil pluggable image cleaners.

jennygrahamjones (approved):
Seems like a reasonable, small change for now.

paperboyo:
> If we moved this into the image model, this could be a good use of @sihil pluggable image cleaners.

This and the `is:` chip names, too: 👏 .

PR #3011, opened 2020-09-23
HOCON config

sihil:
## What does this change?
The grid currently uses a mix of HOCON conf files and Java proerties files for configuration. This PR unifies the configuration to support HOCON everywhere and removes guardian specific locations.

Notably:
 - The new config locations can be in /etc/grid/ or ~/.grid/ so dev configs don't require admin access to the filesystem
 - The order of precedence is (when keys are in multiple files the top item wins):
   - The contents of ~/.grid/<app>.conf
   - The contents of /etc/grid/<app>.conf
   - The legacy properties file at /etc/gu/<app>.properties
   - The play config files
 - All `lazy` vals have been removed so that the config loading is easier to reason about
 - No external files are loaded when running unit tests so config on a developer laptop can never differ from CI when testing
 - We fall back to existing locations for now to provide backwards compatibility (but now read properties via the typesafe config library) - this will be removed in a future PR

We can:
 - [x] Continue to develop locally
 - [x] Develop locally using HOCON config files in developer home directories
 - [ ] Deploy to production continuing to use properties files
 - [ ] Switch production over to HOCON files

## Tested?
- [x] locally
- [x] on TEST

## Todo
 - [X] Edit TEST properties files to include newly-required properties (now lazy vals are no more)
 - [x] Sort out lambda deploys and verify new config for lambdas
 - [x] Edit PROD properties files to include newly-required properties (now lazy vals are no more)
 - [ ] Merge this PR
 - [ ] Convert TEST to HOCON files
 - [ ] Convert PROD to HOCON files
 - [ ] Raise PR to change usage to output to low priority stream (as config cross configures)


media-api/app/lib/MediaApiConfig.scala
```diff
@@ -13,67 +13,65 @@ case class StoreConfig(
...
+  val quotaStoreKey: String = string("quota.store.key")
+  val quotaStoreConfig: StoreConfig = StoreConfig(configBucket, quotaStoreKey)
 
   // quota updates can only be turned off in DEV
-  lazy val quotaUpdateEnabled: Boolean = if (isDev) properties.getOrElse("quota.update.enabled", "false").toBoolean else true
-
-  lazy val recordDownloadAsUsage: Boolean = boolean("image.record.download")
+  val quotaUpdateEnabled: Boolean = if (isDev) boolean("quota.update.enabled") else true
 
-  lazy val imagesAlias: String = properties.getOrElse("es.index.aliases.read", configuration.get[String]("es.index.aliases.read"))
```
akash1810, line 33: Wow, this was unnecessarily complicated before. Much nicer now!

s3watcher/scripts/properties.js
```diff
@@ -1,12 +1,43 @@
...
+    const fieldElements = field.split(".")
+    function getField(obj, fields) {
+      if (fields.length === 0) return obj;
+      getField(obj[fields[0]], fields.slice(1))
+    }
+    return getField(field.data, fieldElements);
+  } else {
+    console.error("Unknown config type")
+    return undefined;
+  }
```
akash1810, line 37: minor: as we're returning early, we can omit the `else` to _slightly_ improve readability.

```suggestion
  if (config.type === "properties") {
    return config.data[field];
  }
  if (config.type === "hocon") {
    const fieldElements = field.split(".")
    function getField(obj, fields) {
      if (fields.length === 0) return obj;
      getField(obj[fields[0]], fields.slice(1))
    }
    return getField(field.data, fieldElements);
  }
  console.error("Unknown config type")
  return undefined;
```

akash1810 (approved):
Really nice change, LGTM.

sihil:
@akash1810 did you successfully run this up locally?

s3watcher/scripts/properties.js
sihil: I'm always in two minds about this. The if ... else structure always seems more explicit in the expression of flow as it's clear that the block will only get executed if the condition doesn't match whilst in the traditional JS style you have to have noticed the return in order to realise that the following code is only executed if the condition doesn't match. I get that this is convention and that my convention comes from Scala but I still prefer it, although not strongly enough if others think I should stick with JS's conventions.

akash1810: Maybe a `switch` would address these concerns? This was a minor observation, feel free to ignore.

akash1810:
> @akash1810 did you successfully run this up locally?

Ha! Yes this is a good idea! I'll do it now.

common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
```diff
@@ -29,30 +20,36 @@ trait CommonConfig extends AwsClientBuilderUtils {
 
   final val sessionId = UUID.randomUUID().toString
 
-  override val awsRegion: String = properties.getOrElse("aws.region", "eu-west-1")
+  // TODO:SAH - remove these and favour explicit config for anything that is derived from here
+  val stage: String = string(GridConfigLoader.STAGE_KEY)
```
akash1810, line 24: Does this mean the HOCON config files can have a `grid.stage` property, which overrides any value read from `/etc/grid/stage`?

akash1810 (commented):
Have tested locally with no issues.

`./dev/script/setup.sh --clean && ./dev/script/start.sh` produced a working Grid.

As did `./dev/script/setup.sh --clean --with-local-auth && ./dev/script/start.sh --with-local-auth `.

common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
sihil: Yes. I'm thinking about deprecating `/etc/grid/stage` and keeping config all in one place. Can't think of a good reason not too.

PR #3015, opened 2020-09-24
Add more better error checking

aug24:
## What does this change?
Exposes a `caused by` section in the Elastic Error in logs _even if_ there is a `root cause`.

Often we have script failures which have a root cause which is entirely unhelpful.  If we have a root cause, we do not currently log the caused by field, whether or not it is populated.  This change adds more logging to clarify if there is a caused by error.

More information in logs.

## Tested?
- [ ] locally
- [ ] on TEST


jonathonherbert (approved):
`rootCause` seems ... optimistically named, but the extra logging looks good 👍

sihil (approved):

PR #3016, opened 2020-09-24
Change pingler to use watch so it doesn't scroll

sihil:
## What does this change?
Pingler's constant scrolling is a distraction when developing. By using watch we can make the status information stay still.

Some notes:
 - Watch's `--color` param doesn't support the 256 color pallette so I've changed the escape codes to still work
 - I've made it denser so it takes up less real estate
 - Watch doesn't show anything until it has the complete output, to make this work better I've limited the curl time to 300ms

Better dev experience everywhere.

Old pingler

new_pingler(1)

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/dev/script/the_pingler.sh
+++ b/dev/script/the_pingler.sh
@@ -1,36 +1,50 @@
 #!/bin/zsh
 
-set -o shwordsplit
+function pingle() {
+  set -o shwordsplit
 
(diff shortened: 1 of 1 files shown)


akash1810 (approved):
Lovely!

jennygrahamjones:
This is great, thank you!

PR #3017, opened 2020-09-29
Add customRatio parameter to allow custom crop ratios.

philmcmahon:
## What does this change?
Following from advice given on https://github.com/guardian/grid/pull/2990 this PR adds a query string parameter `customRatio` to the grid which will add an additional crop ratio to the UI. The format is `customRatio=<label>,<x ratio>,<y ratio>` 

So visiting `https://media.local.dev-gutools.co.uk/search?customRatio=cover%20card1,2` and opening the cropper results in something looking like this:
Screenshot 2020-09-29 at 15 52 29

The 'custom crop' gets stored in session storage, so opening a new grid tab throws it away.

I don't break the grid! And long term the editions cover card builder can use this in it's links to the grid so that the production team can crop an image exactly the same way that the card builder will do it

see above

## Who should look at this?

@guardian/digital-cms 

## Tested?
- [x] locally
- [x] on TEST

### DIFF
--- a/kahuna/public/js/crop/controller.js
+++ b/kahuna/public/js/crop/controller.js
@@ -21,7 +21,6 @@ crop.controller('ImageCropCtrl', [
   'optimisedImageUri',
   'keyboardShortcut',
   'defaultCrop',
-  'cropOptions',
   'cropSettings',
   'square',
   'freeform',
@@ -36,7 +35,6 @@ crop.controller('ImageCropCtrl', [
     optimisedImageUri,
     keyboardShortcut,
     defaultCrop,
-    cropOptions,
     cropSettings,
     square,
     freeform) {
@@ -201,7 +199,7 @@ crop.controller('ImageCropCtrl', [
           callback: () => ctrl.callCrop()
         });
 
-      cropOptions.forEach(option => {
+      cropSettings.getCropOptions().forEach(option => {
         keyboardShortcut.bindTo($scope).add({
           combo: option.key.charAt(0),
           description: `Start ${option.key} crop`,
--- a/kahuna/public/js/crop/controller.js
+++ b/kahuna/public/js/crop/controller.js
@@ -22,7 +22,7 @@ crop.controller('ImageCropCtrl', [
   'keyboardShortcut',
   'defaultCrop',
   'cropOptions',
-  'cropTypeUtil',
+  'cropSettings',
   'square',
   'freeform',
   function(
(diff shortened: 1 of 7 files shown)


paperboyo:
So basically, this won’t ever show for someone not accessing via this special URL? Super-nice! Thanks for going an extra mile, Phil!

philmcmahon:
> So basically, this won’t ever show for someone not accessing via this special URL?

yep!

philmcmahon:
The most recent commit follows from a suggestion by @paperboyo and @aug24 that it should be possible to enforce the extra crop the same way we do with `cropType`. 

e.g.:

Screenshot 2020-09-29 at 17 32 50

akash1810:
Discussed in chat, adding here for prosperity: I think this change results in a very noisy screen. I think we could improve by either:
- Reducing the list of `radio`s into a `select`
- Presenting ratio input for a freeform crop. This PR could then set and lock this value, removing the need to add another radio with a custom label.

That said, the changes in this PR are seen under very specific circumstances so there's an argument to shipping to deliver value and refactor later.

philmcmahon:
Just referencing here - https://github.com/guardian/grid/pull/3018 is an option to help address UX concerns

kahuna/public/js/crop/controller.js
```diff
@@ -37,22 +37,23 @@ crop.controller('ImageCropCtrl', [
     keyboardShortcut,
     defaultCrop,
     cropOptions,
```
akash1810, line 39: It looks like we're getting crop options from `cropSettings.getCropOptions()` now. Can we remove the `cropOptions` import now?

kahuna/public/js/util/crop.js
```diff
@@ -21,30 +26,69 @@ cropUtil.constant('freeform', freeform);
...
+    const customCrop =  storage.getJs(CUSTOM_CROP_STORAGE_KEY, true);
+    return customCrop ? cropOptions.concat(customCrop) : cropOptions;
+  }
+
+  const isValidCropType = cropType => getCropOptions().some(_ => _.key === cropType);
+
+  const isValidRatio = ratio => {
+    const splitRatio = ratio.split(',');
+    return !(splitRatio.length < 3 || isNaN(splitRatio[1]) || isNaN(splitRatio[2]));
+  };
```
akash1810, line 40: We could use [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) to be explicit about what `ratio` should look like. I think using the happy path (`&&`) is clearer too as it removes the global `!`:

```suggestion
  const isValidRatio = ratio => {
    const [label, x, y, ...rest] = ratio.split(',');
    return label && isNaN(x) && isNaN(y);
  };
```

kahuna/public/js/util/crop.js
```diff
@@ -21,30 +26,69 @@ cropUtil.constant('freeform', freeform);
...
 
-  function set({cropType}) {
-    if (!cropType) {
-      return;
+  const parseRatio = ratio => {
+    // example ratio 'longcrop,1,5'
+    if (isValidRatio(ratio)) {
+      const splitRatio = ratio.split(',');
```
akash1810, line 45: Similar to above, destructuring would make this that little bit more readable.

akash1810 (approved):
Couple of comments, looks good overall. I'm uncertain on the UX however let's ship it and get feedback; we can always refactor later.

PR #3018, opened 2020-09-30
Hide alternative crop options when a valid cropType is set.

philmcmahon:
## What does this change?
This modifies the UI of the grid so that if the `cropType` param is set - which locks the grid session to a specific crop size - the other crop size options are hidden. 

This was raisd as a possible UX improvement following discussions triggered by this PR https://github.com/guardian/grid/pull/3017

Happier users?

Before:
Screenshot 2020-09-30 at 15 31 15

After:
Screenshot 2020-09-30 at 15 28 23

## Who should look at this?

@guardian/digital-cms 

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/kahuna/public/js/crop/controller.js
+++ b/kahuna/public/js/crop/controller.js
@@ -52,11 +52,13 @@ crop.controller('ImageCropCtrl', [
         ? `${cropOption.key} (${cropOption.ratioString})`
         : cropOption.key;
 
-      ctrl.cropOptions = cropOptions.map(option => Object.assign(option, {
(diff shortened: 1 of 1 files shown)


jennygrahamjones:
To reiterate what I said on our team chat, I believe that we should only show a disabled button to a user if it's possible for that button to become active during the session, or if the interface is made easier to understand by having the button(s) displayed while disabled. In this case, I think neither of those apply, so this feels like a really sensible change 👍

akash1810 (approved):
LGTM

PR #3019, opened 2020-10-01
S3 watcher modernisation

sihil:
## What does this change?
S3 watcher has barely been touched in four years and was undeployable due to AWS deprecating the node 8 runtime. This is a modernisation effort that:
 - converts codebase to Typescript
   - restructure and introduce types
   - removes rxjs
   - introduces async/await
 - replaces `request` (deprecated) with `node-fetch`
 - ship logs via cloudwatch rather than directly to kinesis
 - add linting
 - write a comprehensive test suite
 - move the build and deploy to main CI / deployable so this doesn't happen again

S3 watcher can be deployed again. In the future it should be easier to maintain, but hard to measure that.

## Deployment
Note that this update has to be done in tandem with [a CFN change](https://github.com/guardian/editorial-tools-platform/pull/467) and must be rolled out together to avoid downtime of ingestion.

## Tested?
- [X] locally
- [X] on TEST


sihil:
The s3Watcher build is deprecated as it has been bundled into the main grid build.

s3watcher/lambda/__tests__/Transfer.test.ts
```diff
@@ -0,0 +1,140 @@
...
+  const importAction = jest.fn().mockReturnValueOnce({ succeeded: true })
+  const result = transfer(
+    logger,
+    s3Client,
+    cloudwatch,
+    importAction,
+    action,
+    ingestConfig
+  )
+  await expect(result).resolves.toBeUndefined()
```
akash1810, line 34: minor: would a more opinionated return type from `transfer` be useful here?

s3watcher/lambda/lib/GridApi.ts
```diff
@@ -0,0 +1,104 @@
...
+  [name: string]: string
+}
+
+interface Params {
+  [name: string]: string
+}
+
+export interface GridImportRequest {
+  key: string
+  url: string
```
akash1810, line 15: Could this be a `URL` type?

s3watcher/lambda/lib/GridApi.ts
```diff
@@ -0,0 +1,104 @@
...
+  const uploadedBy = buildUploadedBy(s3Event.path)
+
+  const params: Params = {
+    filename: s3Event.filename,
+    uploadedBy: uploadedBy,
+    stage: config.stage,
+    uri: imageUri,
+  }
+
+  // @todo – use URLSearchParams?
```
akash1810, line 60: 👍🏽

s3watcher/lambda/lib/Lambda.ts
```diff
@@ -0,0 +1,91 @@
...
+    path,
+    key,
+    filename,
+    size,
+  }
+}
+
+const isIngestConfig = function (config: any): config is IngestConfig {
+  return (
+    typeof config === "object" &&
```
akash1810, line 47: Possibly overkill for this project, however [io-ts](https://github.com/gcanti/io-ts) could make this a bit nicer.

s3watcher/lambda/lib/Lambda.ts
```diff
@@ -0,0 +1,91 @@
...
+  event: ImportAction
+): Promise<string> {
+  const response = await s3Client
+    .getObject({
+      Bucket: event.bucket,
+      Key: "config.json",
+    })
+    .promise()
+  const data = response.Body?.toString()
+  if (data === undefined) {
```
akash1810, line 84: I think the `response.Body?` would mean we never land here?

scripts/ci/grid.sh
```diff
@@ -5,17 +5,25 @@ set -e
...
-
-    nvm install
-    nvm use
+  export NVM_DIR="$HOME/.nvm"
+  if [ -s "$NVM_DIR/nvm.sh" ]; then
+    # This loads nvm if installed directly
+    source "$NVM_DIR/nvm.sh"
+  elif [ -s "$(brew --prefix nvm)/nvm.sh" ]; then
+    # This will load nvm if installed via brew
+    source "$(brew --prefix nvm)/nvm.sh"
```
akash1810, line 14: As this is the CI step, would we ever expect nvm to be installed via brew here?

s3watcher/lambda/lib/EnvironmentConfig.ts
```diff
@@ -0,0 +1,32 @@
+interface EnvironmentConfig {
+  stack: string
+  stage: string
+  app: string
+  isDev: boolean
+  region: string
+  profile: string
+  loggingRoleArn?: string
+  loggingStream?: string
```
akash1810, line 9: As we're moving to log via cloudwatch, are these props still needed?

s3watcher/lambda/__tests__/Lambda.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  ImportAction,
+  IngestConfig,
+  parseIngestConfig,
+} from "../lib/Lambda"
+
+test("createActionFromNotification correctly parses a record", () => {
+  const record: S3EventRecord = {
+    awsRegion: "eu-west-1",
+    eventName: "Put",
+    eventVersion: "",
```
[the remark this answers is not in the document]
sihil: They are not used on any code path - this is just to satify the enormous S3EventRecord. I guess I could cast from a much smaller object.

scripts/ci/grid.sh
sihil: Fair point, but I was testing the CI step locally so this was useful for me and I'd anticipate someone doing the same in the future.

s3watcher/lambda/lib/EnvironmentConfig.ts
sihil: Nope.

s3watcher/lambda/lib/Lambda.ts
sihil: `Body` can be undefined and the optional chaining operator will do the same as mapping over an option as I unserstand it so if `Body` is undefined then `data` will be undefined but otherwise data will be the result of `toString()`.

s3watcher/lambda/lib/Lambda.ts
sihil: Thought about it with Jon but agreed that it would be overkill.

s3watcher/lambda/__tests__/Fixtures.ts
```diff
@@ -0,0 +1,42 @@
...
+import { ImportAction, IngestConfig } from "../lib/Lambda"
+import { createLogger, Logger, LoggingFields } from "../lib/Logging"
+
+export const ingestConfig: IngestConfig = {
+  region: "eu-west-1",
+  baseUrl: "https://grid.example.net",
+  apiKey: "top-secret",
+  failBucket: "grid-failure-bucket",
+  s3UrlExpiry: 60,
+  stage: "TEST",
```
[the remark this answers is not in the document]
sihil: Stage can be anything in theory and I don't believe that code should have to be modified to deploy a new stage so I don't think it should be done.

s3watcher/lambda/__tests__/EnvironmentConfig.test.ts
```diff
@@ -0,0 +1,48 @@
+import { readConfig } from "../lib/EnvironmentConfig"
+
+const OLD_ENV = process.env
+
+beforeEach(() => {
+  jest.resetModules() // most important - it clears the cache
+  process.env = { ...OLD_ENV } // make a copy
```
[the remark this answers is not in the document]
sihil: This is a deep copy because otherwise pushing new values into the object during a test would modify the original copy stored in `OLD_ENV`. I've actually changed this to set the enviroment to be empty to isolate the tests from the user/build environment.

s3watcher/lambda/__mocks__/aws-sdk.ts
```diff
@@ -0,0 +1,59 @@
+export const getObjectPromise = jest.fn().mockReturnValue(Promise.resolve(true))
+export const copyObjectPromise = jest
+  .fn()
+  .mockReturnValue(Promise.resolve(true))
+export const deleteObjectPromise = jest
+  .fn()
+  .mockReturnValue(Promise.resolve(true))
+export const getSignedUrlPromisePromise = jest
+  .fn()
+  .mockReturnValue(Promise.resolve(true))
```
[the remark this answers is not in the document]
sihil: No and during the tests we tell them to return something more sensible. To avoid confusion I'll remove the mocking from here entirely.

s3watcher/lambda/__tests__/Lambda.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    stage: "CONFIG_TEST",
+  })
+})
+
+test("parseIngestConfig should fail to parse an incomplete config", () => {
+  const config = `{
+    "stage": "CONFIG_TEST"
+  }`
+  const action: Promise<IngestConfig> = parseIngestConfig(config)
+  expect(action).rejects.toThrowError("")
```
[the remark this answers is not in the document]
sihil: TIL that if you provide a string to `toThrowError` in jest it uses it as a substring. Pretty sure I put `""` in here to make it tell me what really got thrown and then didn't notice the test passed. Ooops 😬

s3watcher/lambda/__tests__/Metrics.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  ])
+  expect(metricWithDim?.Timestamp).toEqual(timestamp)
+  expect(metricWithDim?.Unit).toEqual("Count")
+  expect(metricWithDim?.Value).toBe(1)
+
+  const metricNoDim = metricData.MetricData.find(
+    (m) => m.Dimensions === undefined
+  )
+  expect(metricWithDim?.MetricName).toEqual("UploadedImages")
+  expect(metricWithDim?.Dimensions).toBeUndefined
```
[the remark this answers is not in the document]
sihil: It should! `toBeUndefined` isn't called either which is why I didn't notice. Good spot 👍

s3watcher/lambda/lib/GridApi.ts
sihil: Wasn't sure about doing this but seems that URL is sane in javascript and fetch already takes it without any call site conversion so 👍

s3watcher/lambda/lib/GridApi.ts
sihil: Oh yes. Done.

akash1810 (approved):

PR #3020, opened 2020-10-01
use new dev-nginx command to ensure nginx is running

akash1810:
## What does this change?
This avoids the slightly annoying behaviour of restarting your machine (i.e. not having nginx running), starting grid, opening the browser and being greeted with the `ERR_CONNECTION_REFUSED` error as you've forgotten to start nginx.

Scripting nginx allows us to be lazy.

Requires an update to `dev-nginx`: `brew upgrade guardian/devtools/dev-nginx`.

See:
- https://github.com/guardian/dev-nginx/pull/22
- https://github.com/guardian/homebrew-devtools/pull/59

Fewer manual steps.

n/a

## Who should look at this?

@guardian/digital-cms

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/dev/script/start.sh
+++ b/dev/script/start.sh
@@ -122,12 +122,17 @@ checkForJavaHome() {
   fi
 }
 
+ensureNginxIsRunning() {
+  dev-nginx start -g
+}
+
 main() {
   checkForJavaHome
   hasCredentials
   checkRequirements
   checkNodeVersion
   startDockerContainers
+  ensureNginxIsRunning
   buildJs
   startPlayApps
 }


sihil:
To throw in a spanner...

dev-nginx depends on nginx from homebrew, which is presumably where most people have got it from. That implies that the right way to start it is with `brew services start nginx`. Running that will `Start the service formula immediately and register it to launch at login (or boot).`. I don't think you even need sudo as it works via the launchctl daemon. So dev-nginx could do that once and then it's not necessary to modify projects like the grid.

akash1810:
Ah brew services is a nice solution here, just remembering that @AWare suggested that ages ago too and I was meant to look into it.

It looks like both dev-nginx and grid do not dictate macOS as your host. If you're on a linux distro you won't have brew services. That is, I think there is value in the `start` command in general.

Maybe a simpler solution is to add a [caveat](https://www.rubydoc.info/github/Homebrew/brew/Formula:caveats) to the dev-nginx brew formula with a note: `Thanks for installing dev-nginx. You'll also want to run 'brew services start nginx' to run nginx at launch.`, and closing this PR?

sihil:
Whilst brew technically works on Linux do we actually support that in any of our projects? I know that even the handful of devs that were still on linux have now switched to macOS. As such I doubt that any projects are actually tested on Linux so I suspect it is wishful thinking that they would work.

I wonder if you can put `services start nginx` into the brewfile?

sihil:
After further thought I wonder whether making dev-nginx restart smarter will solve the problems. Running `nginx` directly is a naive approach regardless of platform. On linux we should use `systemctl` and on macos we should use `LaunchCtl` - likely via `brew services`. We can detect both the OS and whether nginx came from brew pretty easily. Then restart can use the right command (although the semantics are different I think, systemctl restart might only work if already running whilst brew services restart definitely works when stopped so will always be "graceful"). As a side effect of doing so I think will end up restarting on macOs boot. We could do systemctl enable to ensure that the service starts on boot there too if we think it is relevant.

AWare:
`brew services start nginx` will need to be run as root, otherwise nginx will fail to run

sihil:
`brew services` lied and said it was running but you're quite right, it doesn't. I did wonder how it circumvented root to get access to the lower ports.

akash1810:
IIUC we're suggesting to solve the issue of witnessing `ERR_CONNECTION_REFUSED` because we forgot to start nginx within dev-nginx itself as it results in a smaller surface area of change (fixing at source).

In which case, I'll close this PR.

PR #3034, opened 2020-10-14
bump metadata-extractor to stay up to date

akash1810:
## What does this change?
Bumps metadata-extractor to stay up to date.


jennygrahamjones (approved):
Tested by @paperboyo 👍

PR #3035, opened 2020-10-19
Add date tests and new allowable date format

aug24:
## What does this change?
Dates will now accept a new, even more stupid format: `yyyy` and `yyyy-mm`. This applies to a date coming from `photoshop:DateCreated` XMP field, because [our composites](https://github.com/guardian/grid/blob/2c8c1ef40afc7fea3743fe764cac216ed79c0888/image-loader/app/lib/imaging/FileMetadataReader.scala#L92-L108) for multiple IPTC and exif fields (`Date Time Created Composite` and `Date/Time Original Composite`) should never contain partial data.

No-op as far as behaviour, but images with dates of this year-only and year-and-month-only formats will have their dateTaken recorded. For just year, missing data is filled with first month of the year, first day of the month and 00:00 for missing hours.

## Tested?
- [ ] locally
- [x] on TEST


jonathonherbert (approved):
The formats are stupid ... but are they stupid _enough_?

Lovely to see the tests 👏

common-lib/src/test/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverterTest.scala
```diff
@@ -419,5 +420,37 @@ class ImageMetadataConverterTest extends FunSpec with Matchers {
     imageMetadata.peopleInImage should be (Set("person 1","person 2"))
   }
 
+  it("should cope with awful dates") {
```
jennygrahamjones, line 423: Could this have a name that explains a little more about what functionality this is actually testing? 🙏

aug24: I thought that was quite explicit!

paperboyo: I since updated the PR description (without asking consent). Is that enough or should it be in code?

jennygrahamjones: "Awful" _is_ both descriptive and explicit, but it doesn't really say anything about what we consider awful, or what specifically about all of the dates we're asserting against is awful. Is this "should accept incomplete dates" or "should accept dates in a non-standard format" or both or something else? I could probably infer that from looking at all of the assertions, or by looking up the PR, but it would also be useful to have that context right there in the test itself.

PR #3036, opened 2020-10-26
Allow easily searching for suppliersReference and specialInstructions 

paperboyo:
Replaces https://github.com/guardian/grid/pull/2716.

This allow searching for both these as fields (and chips) and also in free-text search. 

~Currently, for some unknown reason, suppliersReference search by-field, doesn’t work (I’ve tried bot `keywordField` and `standardAnalysed` mapping! I would love to be told what’s wrong. Even more so, because this change requires re-mapping (so, a [reindex](https://github.com/guardian/grid/tree/main/scripts#scripts))).~
-- turned out ot be the order of ’em (see comment in code)

Terms from (parts of) specialInstructions show up in free-text search. One can easily search for suppliersReference.

## Who should look at this?
Will shyly ask Akash.

## Tested?
- [x] locally
- [x] on TEST


AWare (approved):

AWare (changes requested):
Having some issues running a remap locally so putting a hold on this pr until resolved. 
image

AWare:
I think we should review this before we reingest.

andrew-nowak (approved):
tested on test, lgtm!

PR #3037, opened 2020-11-02
Config convertor

sihil:
## What does this change?
This adds a script to convert configuration from .properties files to .conf (HOCON) files. This is helpful during the upgrade for #3011.

Note that this imports the typesafe config library so that the output can be customised. Ideally we'd push the changes to the upstream library although they have a large backlog of unmerged PRs, so it might not be worth the effort.

## Tested?
This has been used to convert the Guardian's TEST and PROD configurations into HOCON.


jennygrahamjones:
Do we anticipate that we will need to make this conversion again in the future?

sihil:
No, but the BBC might so I wanted to check in my script for posterity.

jennygrahamjones (approved):

PR #3038, opened 2020-11-03
Bump localstack to fix kinesis physical IDs

sihil:
## What does this change?
With the previous version of localstack the configuration was not being properly populated as the localstack API was not returning the `PhysicalId` for Kinesis streams. This is working in the latest version of localstack.

[Localstack 11.6](https://github.com/localstack/localstack/releases/tag/v0.11.6) drops support for the S3 service ports so we switch to the main port.

Can spin up a cleanly configured stack using `setup.sh --clean`

## Tested?
- [X] locally

### DIFF
--- a/dev/imgops/nginx.conf
+++ b/dev/imgops/nginx.conf
@@ -32,7 +32,7 @@ http {
       image_filter resize $arg_w $arg_h;
 
       # connect to the localstack docker container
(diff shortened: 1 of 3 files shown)


akash1810 (approved):

akash1810 (approved):
Have tested this locally and its working as expected.

PR #3040, opened 2020-11-05
Sanitise logging

sihil:
## What does this change?
This makes a couple of key changes to the way that the grid does logging:
 - Move the log shipping initialisation earlier (at present the log shipping is initialised after the app components are created so any errors or messages from this process are lost)
 - Switch entirely from play.api.Logger to [https://github.com/lightbend/scala-logging](lightbend/scala-logging) which is a performant macro wrapper around SLF4J loggers - this means that loggers are created for each class making it easier to understand where a log entry originates from.
 - Switch from using the `GridLogger` object to a `GridLogging` trait that initialises the above and includes some helpful extra features via class extension.

More consistent logging throughout the Grid.

## Tested?
- [X] locally
- [X] on TEST


collections/app/CollectionsComponents.scala
```diff
@@ -5,9 +5,7 @@ import play.api.ApplicationLoader.Context
 import router.Routes
 import store.CollectionsStore
 
-class CollectionsComponents(context: Context) extends GridComponents(context) {
-  final override lazy val config = new CollectionsConfig(configuration)
-
+class CollectionsComponents(context: Context) extends GridComponents(context, new CollectionsConfig(_)) {
```
jonathonherbert, line 8: Huh, how does the `_` work? Does Scala work it out by type?

common-lib/src/main/scala/com/gu/mediaservice/lib/logging/GridLogging.scala
```diff
@@ -0,0 +1,36 @@
...
+import com.gu.mediaservice.lib.auth.ApiAccessor
+import com.typesafe.scalalogging.{Logger, StrictLogging}
+import net.logstash.logback.marker.Markers
+
+import scala.collection.JavaConverters._
+
+trait GridLogging extends StrictLogging {
+  case class ImageId(id: String)
+
+  implicit class LoggerWithHelpers(logger: Logger) {
```
jonathonherbert, line 11: Just out of interest, and if you've got time – I'm unsure about this implicit class. Is it effectively an override for an implicit provided in `StrictLogging` (where we provide some sugar for common ops)?

jonathonherbert (approved):
Looks good 👍  a few questions about some Scala things I didn't understand.

common-lib/src/main/scala/com/gu/mediaservice/lib/logging/GridLogging.scala
sihil: The implicit class is scala's mechanism for support extension methods. This is essentially adding the methods in the body of the class to the entity in the single parameter constructor. This is useful if you want to add a method onto a class that you can't edit the code of (i.e. in a third party library). It's all resolved at compile time so there isn't a runtime overhead. This is the third overloaded use of `implicit` in scala and doesn't relate to the other uses of implicit.

jonathonherbert: Ah, gotcha – thanks for the explanation. Yes, that's not something I would have guessed, because as you say ...
> This is the third overloaded use of implicit in scala and doesn't relate to the other uses of implicit.

😅

collections/app/CollectionsComponents.scala
sihil: It's currying. The type of the second param of `GridComponents[Config <: CommonConfig]` is `Configuration => Config` so we need to pass in a function that takes a play configuration and returns something that is a subclass of `CommonConfig`. `new CollectionsConfig(_)` is short for `c => new CollectionsConfig(c)` which is turning the constructor into a generic `Function1[Configuration, CollectionsConfig]` shape. The type argument of GridComponents is then inferred from the return value of the function we're passing in.

jonathonherbert: > new CollectionsConfig(_) is short for c => new CollectionsConfig(c) 

Ah, this is what I was missing, thanks!

common-lib/src/main/scala/com/gu/mediaservice/lib/logging/GridLogging.scala
sihil: Think they are changing this in dotty, so hopefully it will become more obvious. 🤞

PR #3042, opened 2020-11-05
Split the S3 clients so we can use V4 signing

sihil:
## What does this change?
The grid has a very noisy error (`Attempting to re-send the request to <bucket>.s3.eu-west-1.amazonaws.com with AWS V4 authentication. To avoid this warning in the future, please use region-specific endpoint to access buckets located in regions that require V4 signing.`). One part of our code needs to use the legacy v2 signing at the moment due to the way that signed URLs are proxied through nginx. 

In order to get around this we create two clients for different purposes. Most of the time we can happily use V4 signing. When we can't we use the legacy client which disables it. This also avoids redirects so should improve performance.

Far fewer (possibly zero) errors like the above in the grid logs. We currently log over 12,000 of these errors per day.

## Who should look at this?
@guardian/digital-cms

## Tested?
- [X] locally
- [X] on TEST

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
@@ -31,6 +31,8 @@ class S3(config: CommonConfig) extends GridLogging {
   import S3Ops.objectUrl
 
   lazy val client: AmazonS3 = S3Ops.buildS3Client(config)
+  // also create a legacy client that uses v2 signatures for URL signing
+  private lazy val legacySigningClient: AmazonS3 = S3Ops.buildS3Client(config, forceV2Sigs = true)
   private val log = LoggerFactory.getLogger(getClass)
 
   private def removeExtension(filename: String): String = {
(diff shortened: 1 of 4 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
```diff
@@ -168,14 +170,16 @@ class S3(config: CommonConfig) extends GridLogging {
...
-    // imgops proxies direct to S3, passing the AWS security signature as query parameters
-    // This does not work with AWS v4 signatures, presumably because the signature includes the host
+  def buildS3Client(config: CommonConfig, forceV2Sigs: Boolean = false, localstackAware: Boolean = true): AmazonS3 = {
+
     val clientConfig = new ClientConfiguration()
-    clientConfig.setSignerOverride("S3SignerType")
+    // Option to disable v4 signatures (https://github.com/aws/aws-sdk-java/issues/372) which is required by imgops
+    // which proxies direct to S3, passing the AWS security signature as query parameters. This does not work with
+    // AWS v4 signatures, presumably because the signature includes the host
+    if (forceV2Sigs) clientConfig.setSignerOverride("S3SignerType")
```
akash1810, line 182: minor: is it worth logging when we enter this path, just to get a sense of scale?

akash1810 (approved):

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
sihil: Essentially it'll happen once per instance for a service that calls `signUrl` (but only once as it's a lazy val so client will be kept around after that). As far as I can see it's only the MediaApi that does this. I'm not sure it will tell us much that can't be inferred from a few minutes reading the source code.

PR #3043, opened 2020-11-05
Fail fast if deprecated configuration is found

sihil:
## What does this change?
After #3011 we no longer need to support properties files. This PR removes the files from the list of configuration consulted and also refuses to start up if any properties files are present. The aim of this is to make sure that a service is not inadvertently mis-configured when we stop reading the files (i.e. a mix of different configuration formats).

We no longer support old configuration files.

## Who should look at this?
@guardian/digital-cms

## Tested?
- [X] locally
- [x] on TEST


AWare (approved):

sihil:
@wainaina @ochiengolanga - I'm interested in merging this fairly soon. Are the BBC still using properties files for configuration? If so I'll close this and we can revisit it in April. Otherwise I'm keen to merge now.

sihil:
The BBC have confirmed that they are still using the legacy properties files at the moment. I'm going to close this and open an issue instead.

PR #3046, opened 2020-11-19
Improve the failure logging to include more context

sihil:
## What does this change?
Some payloads we send to kinesis are too large. We log that this happens but this PR adds more context to make it easier to understand what's happening.

We can track oversized payload errors more easily.

## Tested?
- [x] locally
- [x] on TEST

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/Kinesis.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/Kinesis.scala
@@ -9,9 +9,10 @@ import com.gu.mediaservice.lib.config.CommonConfig
 import com.gu.mediaservice.lib.json.JsonByteArrayUtil
 import com.gu.mediaservice.model.usage.UsageNotice
 import net.logstash.logback.marker.{LogstashMarker, Markers}
-import play.api.libs.json.{JodaWrites, Json}
+import play.api.libs.json.{JodaWrites, Json, Writes}
 import com.amazonaws.auth.AWSCredentialsProvider
 import com.gu.mediaservice.lib.logging.GridLogging
+import org.joda.time.DateTime
 
 case class KinesisSenderConfig(
   override val awsRegion: String,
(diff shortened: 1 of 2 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/aws/Kinesis.scala
```diff
@@ -51,7 +52,7 @@ class Kinesis(config: KinesisSenderConfig) extends GridLogging{
       logger.info(s"Published kinesis message: $result")
     } catch {
       case e: Exception =>
-        logger.error(s"kinesis putRecord exception message: ${e.getMessage}")
+        logger.error(markers, s"kinesis putRecord failed for ID: ${message.id} File: ${message.image.map(_.source.file)}", e)
```
akash1810, line 55: ID get's [pulled out as a marker](https://github.com/guardian/grid/blob/2c8c1ef40afc7fea3743fe764cac216ed79c0888/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/ThrallMessageSender.scala#L57) too so having it in the message may be necessary? Makes searching easier, I guess.

I wonder if `UpdateMessage` should [expose](https://github.com/guardian/grid/blob/2c8c1ef40afc7fea3743fe764cac216ed79c0888/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/ThrallMessageSender.scala#L53-L61) more information as markers?

akash1810 (approved):
LGTM. It might be useful to expose more markers from the content of an `UpdateMessage` too.

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/Kinesis.scala
sihil: I've moved the fileName to the message marker now whilst retaining the exception. Does that look better @akash1810 ?

akash1810 (approved):

PR #3047, opened 2020-11-25
Remove node from Brewfile

mxdvl:
## What does this change?

Removes node from the Brewfile.

Many devs use `fnm` or `nvm` to handle their installation of Node and its multiple versions.

Brew bundle is more robust.

## Who should look at this?

@akash1810 

## Tested?
- [X] locally -- this does not solve the whole issue yet
- [ ] on TEST

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -1,6 +1,5 @@
 brew "awscli"
 brew "sbt"
-brew "node"
 brew "nginx"
 brew "libgd"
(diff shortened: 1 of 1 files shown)


mxdvl:
This was born out of discussion on how to set up the grid locally, but does not seem necessary and too specific to my unique Homebrew set-up.

PR #3048, opened 2020-11-25
That’s not my name

mxdvl:
## What does this change?

Adds the ability to edit the People in an image. Developed with @zeek01 as part of Hack Day 2020.

Some of the “prompted change” alluded to in [Mistakes with identities have dismayed our readers – and prompted change](https://www.theguardian.com/commentisfree/2020/aug/09/mistakes-identities-dismayed-readers-change).

This required to change the underlying Scala type from `Map[String, String]` &rarr; `Map[String, JsValue]`. The `service.js` also handle the `peopleInImage` field differently than other.

Picture editors should be able to correct this metadata if it is wrong.

## Screenshots

joe-to-john-real

## Who should look at this?

@paperboyo, @akash1810, @guardian/digital-cms

## Tested?
- [X] locally
- [x] on TEST

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,3 +8,6 @@ end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
+
+[*.html]
+indent_size = 4
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
@@ -82,7 +82,7 @@ class DynamoDB(config: CommonConfig, tableName: String) {
     if (value) booleanSet(id, key, value)
     else removeKey(id, key)
 
-  def stringSet(id: String, key: String, value: String)
+  def stringSet(id: String, key: String, value: JsValue)
                 (implicit ex: ExecutionContext): Future[JsObject] =
     update(
       id,
@@ -121,7 +121,7 @@ class DynamoDB(config: CommonConfig, tableName: String) {
       get(id, key).map(item => asJsObject(item))
 
   // We cannot update, so make sure you send over the WHOLE document
-  def jsonAdd(id: String, key: String, value: Map[String, String])
+  def jsonAdd(id: String, key: String, value: Map[String, JsValue])
              (implicit ex: ExecutionContext): Future[JsObject] =
     update(
       id,
@@ -238,10 +238,10 @@ class DynamoDB(config: CommonConfig, tableName: String) {
     case value => value
   }
 
-  def valueMapWithNullForEmptyString(value: Map[String, String]) = {
+  def valueMapWithNullForEmptyString(value: Map[String, JsValue]) = {
     val valueMap = new ValueMap()
-    value.map     { case(k, v) => (k, if (v == "") null else v) }
-         .foreach { case(k, v) => valueMap.withString(k, v) }
+    value.map     { case(k, v) => (k, if (v == JsNull) null else v) }
+         .foreach { case(k, v) => valueMap.withJSON(k, Json.stringify(v)) }
 
     valueMap
   }
--- a/kahuna/public/js/edits/metadataDiff.spec.js
+++ b/kahuna/public/js/edits/metadataDiff.spec.js
@@ -20,15 +20,15 @@ describe('metadataDiff', () => {
   it("finds a changed string field in the presence of an array", () => {
     const initial = { field: "value", array1: ["value 2"] };
     const changed = { field: "changed", array2: ["value 2"] };
-    const expected = { field: "changed" };
+    const expected = { field: "changed", array1: "", array2: ["value 2"] };
     const diff = getMetadataDiff({ data: { originalMetadata: initial } }, changed);
     expect(diff).toStrictEqual(expected);
    });
 
-  it("discards an array field", () => {
+  it("handles an array field", () => {
     const initial = { field: "value", keywords: ["value 2"] };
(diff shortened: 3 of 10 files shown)


.editorconfig
```diff
@@ -8,3 +8,6 @@ end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
+
+[*.html]
+indent_size = 4
```
mxdvl, line 13: I added this to make my editor happy, but happy to drop this change.

kahuna/public/js/edits/service.js
```diff
@@ -258,6 +258,13 @@ service.factory('editsService',
         var proposedMetadata = angular.copy(metadata);
         proposedMetadata[field] = value;
 
+        // peopleInImage is a special case. This turns a comma-separated string into an array trimmed of whitespace
```
paperboyo, line 261: I haz no idea, but just thinking about users editing it sloppily (as we are giving them a simple text field for editing multiple entities): should it also be trimmed of commas?

mxdvl: We could make it behave more like keywords. This is a bit of an in-between solution, where any tags are separated with a comma. I need to add more tests before this is ready for review.

paperboyo: Absolutely good for it to go in like that IMHO regarding editing. Just was thinking of someone wanting to delete first entry in
`Tom Jerry, McDonald Trump`
and actually deleting *only* `Tom Jerry` leaving
`, McDonald Trump`.

mxdvl: That’s handled. It would return an array containing only `McDonald Trump`.

kahuna/public/js/edits/metadataDiff.spec.js
```diff
@@ -20,15 +20,15 @@ describe('metadataDiff', () => {
   it("finds a changed string field in the presence of an array", () => {
     const initial = { field: "value", array1: ["value 2"] };
     const changed = { field: "changed", array2: ["value 2"] };
-    const expected = { field: "changed" };
+    const expected = { field: "changed", array1: "", array2: ["value 2"] };
```
akash1810, line 23: Shouldn't `array1` be absent from `expected` here?

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
```diff
@@ -238,10 +238,10 @@ class DynamoDB(config: CommonConfig, tableName: String) {
     case value => value
   }
 
-  def valueMapWithNullForEmptyString(value: Map[String, String]) = {
+  def valueMapWithNullForEmptyString(value: Map[String, JsValue]) = {
     val valueMap = new ValueMap()
-    value.map     { case(k, v) => (k, if (v == "") null else v) }
-         .foreach { case(k, v) => valueMap.withString(k, v) }
+    value.map     { case(k, v) => (k, if (v == JsNull) null else v) }
+         .foreach { case(k, v) => valueMap.withJSON(k, Json.stringify(v)) }
```
akash1810, line 244: Would be good to have some tests against this change.

metadata-editor/app/controllers/EditsController.scala
```diff
@@ -229,8 +229,9 @@ class EditsController(auth: Authentication, store: EditsStore, notifications: No
   }
 
   // FIXME: At the moment we can't accept keywords as it is a list
```
akash1810, line 231: This PR resolves this `FIXME`, we can remove this comment 🎉 .

akash1810 (commented):
Looks great, let's add some tests.

kahuna/public/js/edits/metadataDiff.spec.js
mxdvl: I think we `array1` should return an empty array. In any case, I’ve updated the tests.

mxdvl:
Grid CI / ScalaBuild (pull_request) failed on `FileMetadataReaderTest.scala` for some reason, but re-running the jobs solved it…

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
AWare: @akash1810 I think this is done now?

AWare:
👍 Happy with this, have seen it on TEST and it's writing sensible things to the database. @akash1810 are you happy with the tests?

.editorconfig
akash1810: You must have a wide screen 😅 .

akash1810 (approved):
This change is great! Not least because it addresses a ~5 year issue!

Works as described on TEST; perhaps @paperboyo wants to give it a final try? We should also send some comms out via Central Production.

This is also likely to be of interest to BBC so cc @ochiengolanga and @wainaina. We can also bring it up at our next meeting.

paperboyo:
> @paperboyo wants to give it a final try?

Him did already. Works a treat!

> We should also send some comms out via Central Production.

Defo! cc @marianadiaspereira

> it addresses a ~5 year issue!

Would that be editing arrays?

> also likely to be of interest to BBC

Yeah, they have [this cool thing](https://github.com/bbc/grid/pull/52) (although I can’t tell which fields are possible to add and if `peopleInImage` is one of them). CP comms would be even cooler if we could bring this to our fork.

.editorconfig
mxdvl: Not so much, but that's in line with what's currently used in the Angular html files.

I'm not wedded to this 😉

PR #3055, opened 2020-12-07
Handle upload failures better

aug24:
## What does this change?
Grid image upload currently logs server errors, several times a day, as follows:
```
play.api.UnexpectedException: Unexpected exception[EntityStreamException: Entity stream truncation]
	at play.api.http.HttpErrorHandlerExceptions$.throwableToUsefulException(HttpErrorHandler.scala:247)
	at play.api.http.DefaultHttpErrorHandler.onServerError(HttpErrorHandler.scala:178)
	at play.filters.cors.AbstractCORSPolicy$$anonfun$1.applyOrElse(AbstractCORSPolicy.scala:125)
	at play.filters.cors.AbstractCORSPolicy$$anonfun$1.applyOrElse(AbstractCORSPolicy.scala:123)
	at scala.concurrent.Future.$anonfun$recoverWith$1(Future.scala:417)
	at scala.concurrent.impl.Promise.$anonfun$transformWith$1(Promise.scala:41)
```
These are believed to be upload failures, caused by flaky networks.  Therefore they should not be considered 5XX errors but 4XX errors.

This change traps those errors and returns `422 - Unprocessable Entity` which will not be logged at ERROR.

Fewer errors in logs, no changes to user experience.

## Tested?
- [ ] locally
- [ ] on TEST


jonathonherbert (approved):
Tested in TEST with throttled connection, reproduced w/ info level message 🎉

common-lib/src/main/scala/com/gu/mediaservice/lib/play/EntityStreamExceptionFilter.scala
```diff
@@ -0,0 +1,19 @@
+package com.gu.mediaservice.lib.play
+
+import akka.http.scaladsl.model.EntityStreamException
+import akka.stream.Materializer
+import com.typesafe.scalalogging.StrictLogging
+import play.api.mvc.{Filter, RequestHeader, Result, Results}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class EntityStreamExceptionFilter(override val mat: Materializer)(implicit ec: ExecutionContext)
```
sihil, line 10: This is a really odd looking filter so perhaps we could add a couple of lines of docs?
A more descriptive name (for purpose rather than function) might be better but can't think of one.

build.sbt
```diff
@@ -97,6 +97,8 @@ lazy val commonLib = project("common-lib").settings(
     // i.e. to only log to disk in DEV
     // see: https://logback.qos.ch/setup.html#janino
     "org.codehaus.janino" % "janino" % "3.0.6",
+    akkaHttpServer,
+    ws
```
sihil, line 101: Minor: As play components, I'd prefer these to be with the play import.

common-lib/src/main/scala/com/gu/mediaservice/lib/play/EntityStreamExceptionFilter.scala
aug24: `ConnectionBrokenFilter`?

sihil (approved):

PR #3056, opened 2020-12-07
Two Antonios is one Antonio

paperboyo:
This has been discussed with Picture Desk and with Antonio himself. I will deploy to TEST to test two things:
1. What will happen to pictures already ingested as `Antonio Zazueta` `The Guardian` (ideally, both the Usage Rights and Byline will change and won’t require manually changing 6k pics)
2. Will uploading his pics (which has `Antonio Olmos` in byline suddenly make them being auto-categorised as `The Guardian` (because of the order in this file) which would be sad for two reasons: a) change of behaviour and b) most of his work is Observer

Depending on outcome of pt. 2, I may further change the order in the file.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
@@ -96,7 +96,7 @@ object MetadataConfig {
 
   val contractedPhotographers: Map[String, String] = Map(
     "Alicia Canter"       -> "The Guardian",
-    "Antonio Zazueta"     -> "The Guardian",
+    "Antonio Olmos"       -> "The Guardian",
     "Christopher Thomond" -> "The Guardian",
     "David Levene"        -> "The Guardian",
     "Eamonn McCabe"       -> "The Guardian",
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
@@ -8,38 +8,40 @@ import Scalaz._
 
 import com.gu.mediaservice.model.{StaffPhotographer, ContractPhotographer, Photographer}
 
+case class KnownPhotographer(name: String, publication: String)
+
 object PhotographersList {
-  type Store = Map[String, String]
   type CreditBylineMap = Map[String, List[String]]
 
-  import MetadataConfig.{ staffPhotographers, contractedPhotographers }
+  import MetadataConfig.{ Store, staffPhotographers, contractedPhotographers }
 
   def creditBylineMap(store: Store): CreditBylineMap = store
-      .groupBy{ case (photographer, publication) => publication }
-      .map{ case (publication, photographers) => publication -> photographers.keys.toList.sortWith(_.toLowerCase < _.toLowerCase) }
+      .groupBy{ case KnownPhotographer(_, publication) => publication }
+      .map{ case (publication, photographers) => publication -> photographers.map(_.name).sortWith(_.toLowerCase < _.toLowerCase) }
 
-  def creditBylineMap(stores: List[Store]): CreditBylineMap =
-    stores.map(creditBylineMap).reduceLeft(_ |+| _)
-
-  def list(store: Store) = store.keys.toList.sortWith(_.toLowerCase < _.toLowerCase)
-
-  def getPublication(store: Store, name: String): Option[String] = store.get(name)
+  def list(store: Store) = store.map(_.name).sortWith(_.toLowerCase < _.toLowerCase)
 
   def caseInsensitiveLookup(store: Store, lookup: String) =
-    store.find{case (name, pub) => name.toLowerCase == lookup.toLowerCase}
+    store.reverse.find{case KnownPhotographer(name, _) => name.toLowerCase == lookup.toLowerCase}
 
   def getPhotographer(photographer: String): Option[Photographer] = {
     caseInsensitiveLookup(staffPhotographers, photographer).map {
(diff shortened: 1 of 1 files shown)


paperboyo:
Sad! Tests are inconclusive, as it looks like Grid is incapable to have the same name for both publications. This will need a rethink to be able to satisfy Picture Desk request (and sanity: this is one person).

paperboyo:
After @twrichards’s help, this works: thank you!

Testing it:

1. Existing pictures stay categorised as Contract Photographer, their byline stays as `Antonio Zazueta`. They can be manually adjusted using the dropdown which now correctly works for both publications

2. Uploading a pic with `Antonio Olmos` in the byline matches as `The Observer` which suggests the last value in the list wins. Which happens to be correct from the business POV (he shots more for the Obs).

Great!

- [x] tested on TEST

twrichards:
@sihil, @paperboyo mentioned you were doing some work on the config stuff (related to BBC) and so would really appreciate your thoughts on this...

@paperboyo has tested my changes in https://github.com/guardian/grid/pull/3056#issuecomment-740596810 and they appear to solve the problem, but wanted to be sure the changes to the model seem OK to you (they're fairly well confined to this file - i.e. the shape of the things used outside this file remain the same AFAIK)

common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
```diff
@@ -9,24 +9,18 @@ import Scalaz._
...
+  import MetadataConfig.{ Store, staffPhotographers, contractedPhotographers }
 
   def creditBylineMap(store: Store): CreditBylineMap = store
-      .groupBy{ case (photographer, publication) => publication }
-      .map{ case (publication, photographers) => publication -> photographers.keys.toList.sortWith(_.toLowerCase < _.toLowerCase) }
+      .groupBy{ case (_, publication) => publication }
+      .map{ case (publication, photographers) => publication -> photographers.map(_._1).sortWith(_.toLowerCase < _.toLowerCase) }
 
-  def creditBylineMap(stores: List[Store]): CreditBylineMap =
-    stores.map(creditBylineMap).reduceLeft(_ |+| _)
```
twrichards, line 22: never used (and due to type erasure was ambiguous with its overload now that `Store` is a `List` rather than `Map`)... so removed it

common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
```diff
@@ -9,24 +9,18 @@ import Scalaz._
...
-
-  def list(store: Store) = store.keys.toList.sortWith(_.toLowerCase < _.toLowerCase)
-
-  def getPublication(store: Store, name: String): Option[String] = store.get(name)
```
twrichards, line 26: the new shape of store means this would either need a `.toMap` or to instead use `.find` like the `caseInsensitiveLookup` below, but it's never used so removed it instead

common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
```diff
@@ -8,38 +8,39 @@ import Scalaz._
...
   }
 }
 
 object MetadataConfig {
 
-  val externalStaffPhotographers: Map[String, String] = Map(
+  type Store = List[KnownPhotographer] // not a Map at this point to allow for duplicate keys (as some photographers take pics for multiple publications)
+
+  import scala.language.implicitConversions
+  implicit def toKnownPhotographer(value: (String, String)) = KnownPhotographer(value._1, value._2)
```
jonathonherbert, line 41: Be lovely to match on this so we can name values 1 through 2 👍

jonathonherbert:
I'll defer to @sihil for comments on stakeholder-facing things, but 👏

common-lib/src/main/scala/com/gu/mediaservice/lib/config/MetadataConfig.scala
twrichards: i went to go do that and I wonder if this is niiiice or confusing
```suggestion
  implicit def toKnownPhotographer(value: (String, String)) = KnownPhotographer.apply _
```

twrichards: i just wonder if `match` is too verbose
```suggestion
  implicit def toKnownPhotographer(value: (String, String)) = value match {
    case (name, publication) => KnownPhotographer(name, publication)
  }
```

jonathonherbert: yeah, I guess the fact that this is implicit (yeesh) makes this less troubling – it'd be nice if you could get the semantics from the call sig, not the implementation, but with implicits I guess the idea is that you're not actually calling the thing yourself. Weird.

I'd favour the verbosity over scala-neatness, but it does seem reasonable to go for either.

sihil: This might be an alternative:
```scala
  implicit class Photographer(name: String) {
    def ->(publication: String): KnownPhotographer = KnownPhotographer(name, publication)
  }
```

twrichards: ahh nice, I've gone for the `implicit class` approach, but called `KnownPhotographerOps` rather than `Photographer` (as that is already a thing, `com.gu.mediaservice.model.Photographer` inside `UsageRights.scala`).

sihil (approved):
I think this is well enough isolated for it not to impact other users. At some stage I anticipate that we will introduce a `guardian` package or sub-project and this will be one of the things that moves there as it is obviously specific to us.

PR #3057, opened 2020-12-07
document cerebro

AWare:
## What does this change?

we have cerebro installed, but nothing on how to access it 

that we can now access it without archeology or guesswork

image

## Tested?
- [X] locally
- [ ] on TEST


paperboyo (approved):
💯 (not checked actual URLs listed!)

PR #3058, opened 2020-12-08
Add pluggable authentication in the grid

sihil:
## What does this change?
The grid is currently hardwired into the Guardian's https://github.com/guardian/pan-domain-authentication library. This is specific to the guardian and it makes sense that other adopters would want to blaze their own trail.

This PR:
 - implements the ability to plug in your own authentication provider by specifying user and machine authentication provider classes implementing appropriate traits in the configuration.
 - implements pan domain authentication and api key implementations of the user and machine authentication providers to replicate existing behaviour.
 - implements a generic provider loader mechanism used to dynamically load image processors and authentication providers (and likely more providers in the future). 

Consumers need to implement a `UserAuthenticationProvider` trait. They can also implement a `MachineAuthenticationProvider` trait to override the way machine to machine calls can be done. These are defined as shown below.
```scala
sealed trait AuthenticationProvider {
  def initialise(): Unit = {}
  def shutdown(): Future[Unit] = Future.successful(())

  /**
    * A function that allows downstream API calls to be made using the credentials of the inflight request
    * @param request The request header of the inflight call
    * @return A function that adds appropriate authentication headers to a WSRequest or returns a runtime error
    */
  def onBehalfOf(request: Principal): Either[String, WSRequest => WSRequest]
}

object AuthenticationProvider {
  type RedirectUri = String
}

trait UserAuthenticationProvider extends AuthenticationProvider {
  /**
    * Establish the authentication status of the given request header. This can return an authenticated user or a number
    * of reasons why a user is not authenticated.
    * @param request The request header containing cookies and other request headers that can be used to establish the
    *           authentication status of a request.
    * @return An authentication status expressing whether the
    */
  def authenticateRequest(request: RequestHeader): AuthenticationStatus

  /**
    * If this provider supports sending a user that is not authorised to a federated auth provider then it should
    * provide a function here to redirect the user. The function signature takes the the request and returns a result
    * which is likely a redirect to an external authentication system.
    */
  def sendForAuthentication: Option[RequestHeader => Future[Result]]

  /**
    * If this provider supports sending a user that is not authorised to a federated auth provider then it should
    * provide a function here that deals with the return of a user from a federated provider. This should be
    * used to set a cookie or similar to ensure that a subsequent call to authenticateRequest will succeed. If
    * authentication failed then this should return an appropriate 4xx result.
    * The function should take the Play request header and the redirect URI that the user should be
    * sent to on successful completion of the authentication.
    */
  def sendForAuthenticationCallback: Option[(RequestHeader, Option[RedirectUri]) => Future[Result]]

  /**
    * If this provider is able to clear user tokens (i.e. by clearing cookies) then it should provide a function to
    * do that here which will be used to log users out and also if the token is invalid.
    * This function takes the request header and a result to modify and returns the modified result.
    */
  def flushToken: Option[(RequestHeader, Result) => Result]
}

trait MachineAuthenticationProvider extends AuthenticationProvider {
  /**
    * Establish the authentication status of the given request header. This can return an authenticated user or a number
    * of reasons why a user is not authenticated.
    * @param request The request header containing cookies and other request headers that can be used to establish the
    *           authentication status of a request.
    * @return An authentication status expressing whether the
    */
  def authenticateRequest(request: RequestHeader): ApiAuthenticationStatus
}
```

Much like the pluggable image processor pipeline, these authentication providers are loaded dynamically at runtime. The loader supports companion objects and also classes with constructors that can optionally have arguments (a `play.api.Configuration` instance and/or a `AuthenticationProviderResources` instance - shown below).
```scala
/**
  * Case class containing useful resources for authentication providers to allow concurrent processing and external
  * API calls to be conducted.
  * @param commonConfig the Grid common config object
  * @param context an execution context
  * @param actorSystem an actor system
  * @param wsClient a play WSClient for making API calls
  */
case class AuthenticationProviderResources(commonConfig: CommonConfig,
                                           actorSystem: ActorSystem,
                                           wsClient: WSClient,
                                           controllerComponents: ControllerComponents)
```

## TODO
 - [X] Integrate the new `Authentication` class into the application (and remove the existing version)
 - [X] Write an class loader to load and instantiate providers (or perhaps make the image processor one generic)
 - [x] Review the `AuthenticationProviderResources` that can be passed into a provider
 - [X] Make `PandaUser` more generic (possibly including a `TypedMap` for extensions?)
 - [x] Write tests to confirm and assert the behaviour of the various components
   - [x] Authentication.authenticateRequest
   - [x] Authentication.onBehalfOf
   - [ ] ~PandaAuthenticationProvider~ (fairly hard to do and only used by the Guardian)
   - [x] ApiKeyAuthenticationProvider
 - [x] Add required configuration changes to PROD before merging

# Follow on PRs
 - Write documentation on implementing extension points (including authentication) #3103
 - Implementing a default provider that simply lets users in with a hardcoded user ID for simpler spinning up #3104
 - Implementing bundled provider to support OIDC #3105

Other organisations are able to implement their own idea of authentication without having to fight with the Guardian's pan domain authentication library. 

## Who should look at this?
It would be great to have @mbarton and @akash1810 review this PR.

## Tested?
- [x] locally
- [x] on TEST


common-lib/src/main/scala/com/gu/mediaservice/lib/auth/provider/Authentication.scala
```diff
@@ -0,0 +1,111 @@
...
+                     providers: AuthenticationProviders,
+                     override val parser: BodyParser[AnyContent],
+                     override val executionContext: ExecutionContext)
+  extends ActionBuilder[Authentication.Request, AnyContent] with ArgoHelpers {
+
+  // make the execution context implicit so it will be picked up appropriately
+  implicit val ec: ExecutionContext = executionContext
+
+  val loginLinks = List(
+    Link("login", config.services.loginUriTemplate)
```
mbarton, line 26: Minor point but this should probably come from the provider rather than config now, assuming Kahuna does actually use it

common-lib/src/main/scala/com/gu/mediaservice/lib/auth/provider/AuthenticationProvider.scala
```diff
@@ -0,0 +1,82 @@
...
+    */
+  def sendForAuthentication: Option[(RequestHeader, Option[Principal]) => Future[Result]]
+
+  /**
+    * If this provider supports sending a user that is not authorised to a federated auth provider then it should
+    * provide an Play action here that deals with the return of a user from a federated provider. This should be
+    * used to set a cookie or similar to ensure that a subsequent call to authenticateRequest will succeed. If
+    * authentication failed then this should return an appropriate 4xx result.
+    */
+  def processAuthentication: Option[RequestHeader => Future[Result]]
```
mbarton, line 63: Minor naming point: I'd put `callback` in the name of this method somewhere just to drive home what it will be used for

mbarton (commented):
I think the level of abstraction makes sense. Probably worth us using this to connect to auth0 or another easy to set up provider to shake out any issues though.

common-lib/src/main/scala/com/gu/mediaservice/lib/auth/provider/AuthenticationProvider.scala
sihil: `sendForAuthenticationCallback` to make it clear that it's related to the above?

common-lib/src/main/scala/com/gu/mediaservice/lib/auth/provider/Authentication.scala
sihil: I've made this customisable.

common-lib/src/main/scala/com/gu/mediaservice/lib/auth/provider/AuthenticationProvider.scala
sihil: I've gone for the above.

auth/app/auth/AuthController.scala
```diff
@@ -31,70 +29,85 @@ class AuthController(auth: Authentication, val config: AuthConfig,
...
   }
 
 
   def isOwnDomainAndSecure(uri: URI): Boolean = {
     uri.getHost.endsWith(config.domainRoot) && uri.getScheme == "https"
   }
   def isValidDomain(inputUri: String): Boolean = {
-    Try(URI.create(inputUri)).filter(isOwnDomainAndSecure).isSuccess
+    val success = Try(URI.create(inputUri)).filter(isOwnDomainAndSecure).isSuccess
+    if (!success) logger.warn(s"Provided login redirect URI is invalid: $inputUri")
```
mbarton, line 74: 💯

common-lib/src/main/resources/application.conf
```diff
@@ -23,3 +23,22 @@ image.processors = [
...
 ]
+
+authentication.providers {
+  machine {
+    className = "com.gu.mediaservice.lib.auth.provider.ApiKeyAuthenticationProvider"
+    config {
+      # authKeyStoreBucket = <s3-bucket-with-api-keys>
+    }
+  }
+  # TODO: short term we put panda here for backwards compatibility but the default provider should be something better
```
mbarton, line 34: Minor point: lets say something like "the default provider should be something that involves zero configuration for a simple test setup", just to help any future archaeologists infer what we mean by "better" here

common-lib/src/main/resources/application.conf
```diff
@@ -23,3 +23,22 @@ image.processors = [
...
+  user {
+    className = "com.gu.mediaservice.lib.guardian.auth.PandaAuthenticationProvider"
+    config {
+      # all of the things relating to pan domain auth (these are currently sensibly defaulted in code)
```
mbarton, line 38: Worth saying something like "sensibly defaulted for Guardian usage"?

sihil: I kind of figured that's implicit in that the provider is Guardian specific, but I will make that clearer.

common-lib/src/main/scala/com/gu/mediaservice/lib/auth/Authentication.scala
```diff
@@ -1,122 +1,105 @@
...
+    // Authenticate request. Try with API authenticator first and then with user authenticator
+    providers.apiProvider.authenticateRequest(requestHeader) match {
+      case Authenticated(authedUser) => Right(authedUser)
+      case Invalid(message, throwable) => Left(unauthorised(message, throwable))
+      case NotAuthorised(message) => Left(forbidden(s"Principal not authorised: $message"))
+      case NotAuthenticated =>
+        providers.userProvider.authenticateRequest(requestHeader) match {
+          case NotAuthenticated => Left(unauthorised("Not authenticated"))
+          case Expired(principal) => Left(expired(principal))
+          case GracePeriod(authedUser) => Right(authedUser)
```
mbarton, line 59: Do we need `GracePeriod` in the provider API at all? From a quick glance at the code I think we could make it an internal concern to the Panda provider, but there might be something I missed

media-api/app/MediaApiComponents.scala
```diff
@@ -30,6 +32,7 @@ class MediaApiComponents(context: Context) extends GridComponents(context, new M
   val usageQuota = new UsageQuota(config, actorSystem.scheduler)
   usageQuota.quotaStore.update()
   usageQuota.scheduleUpdates()
+  applicationLifecycle.addStopHook(() => Future{usageQuota.stopUpdates()})
```
mbarton, line 35: This looks good. I'm assuming it's for cleanliness rather than it having any implications on authentication?

mbarton (approved):

akash1810:
> Add required configuration changes to PROD before merging

Worth explicitly detailing what changes are needed in the PR description for ease of adoption?

akash1810 (approved):
Couple of minor comments, LTGM overall.

PR #3061, opened 2020-12-09
some tweaks to the batch api

AWare:
## What does this change?

I _think_ some of the difficulty in batch edits is that the check queue is causing some contention. This adds a 100ms between workers taking things off the queue. 

This does 30 in five seconds, and 136 in four minutes. This is a ten times slowdown. I have no idea. 

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/kahuna/public/js/util/async.js
+++ b/kahuna/public/js/util/async.js
@@ -53,7 +53,7 @@ async.service("queue", ['$timeout', ($timeout) => {
   console.log("queue created!");
   return createQueue({
     timeout: $timeout,
-    maxWorkers: 2
+    maxWorkers: 4
   });
 }]);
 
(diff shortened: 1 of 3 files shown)


AWare:
136 in 4:40 (unchanged)
136 in 5:54 (30,1)
136 in 7:30 (60,1)
136 in 5:32 (30,2)
136 in 5:15 (15,2)
136 in 5:05 (8,8) (gets caught, _feels_ slow)
136 in 4:55 (4,4) (feels even slower)
136 in 4:10 (30,30) (4:40 in ff)

AWare:
Eventually managed to cause the apipoll storm that caused this in the first place. 

Going to split out the poll function from the batch call somehow and come back with a pr that does that.

Issue #1984, opened 2016-08-13
Invalidity indicator not updating state in selection mode


paperboyo:
Hi,

Everything else is updating UI correctly when in selection mode, just this badge is not. Fine after refresh:

![rights not updating](https://cloud.githubusercontent.com/assets/6032869/17643047/9c6bdeac-6155-11e6-990b-e42d84be4562.gif)

Regards
Mateusz

paperboyo:
Fixed by https://github.com/guardian/grid/pull/3064. Thanks!

PR #3064, opened 2020-12-15
Speed up batch jobs

AWare:
## What does this change?

Most of the consumers of `batch-tracking.js` emit and `image-changed`. Careful introspection of that code reveals _very little_ which could be responsible for creating http requests. *Yet here we are* whenever one of these messages is sent, it would appear that every resource requested by the various calls to `apiPoll` is repeated. Or something like that, the only causality I can find is that when that is called, it's really slow, and when it's not, it isn't. 

This totally refactors `trackAll` so that it can split out the various stages of updating an image and control their concurrency a little more carefully. This is also a lot simpler, as it no longer uses `queue.js` which was a very complex attempt to solve this request explosion. https://github.com/guardian/grid/pull/2917 and https://github.com/guardian/grid/pull/2690 and https://github.com/guardian/grid/pull/3061 represent the previous work on this.

This now uses `async`/`await` and [`p-queue`](https://www.npmjs.com/package/p-queue) to handle this. It's not apparently normal to mix angularjs promises with async code, but it works, and is significantly easier to reason about. 

Now that we fire `image-changed` at the end of big batches, we're looking at hundreds of messages dropped into the queue at once. Not entirely sure that I have to mention angular doesn't like this so:

This also changes the `image-changed` handlers to be accompanied by `images-changed` which takes and array of images- and then to trickle feed images into this at some sort of semi-reasonable rate- while showing the user that the job they actually wanted done is done. 

The refactors don't affect other consumers of `trackAll` which I'll work through. 

Update: `results.set` causes all kinds of bad performance. Have added the immutable `map` function and that seems to clear it up. (@paperboyo  this is in test)

Adding 134 labels takes about 45' instead of 4:45 🐢. Half of this time is spent updating the UI. 
330 labels completes in about 4 minutes instead of 32:30
## Tested?
- [ ] locally
- [ ] on TEST

- leases don't currently update the UI.
- crops don't appear in composer


AWare:
This also resolves https://github.com/guardian/grid/issues/1984

AWare:
@paperboyo reports a big slowdown on the last few entries. instrumenting

paperboyo:
This is amazing. The speedup for 330 labels is actually 33 times. One problem (see more info in internal chat) is that data in the metadata panel seems to have lost its refreshed state until page reload:

labels not in sync

kahuna/public/js/components/gr-add-label/gr-add-label.js
```diff
@@ -46,7 +46,8 @@ addLabel.controller('GrAddLabelCtrl', [
 
         }
 
-        function saveFailed() {
+        function saveFailed(e) {
+          console.error(e);
```
jonathonherbert, line 50: Still needed?

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
```diff
@@ -13,153 +13,150 @@ export const module = angular.module('gr.imageMetadata', [
...
+
+    const freeUpdateHandler = (updatedImage) => {
+      console.log('image update recvd in gr-image-metadata', updatedImage);
+      ctrl.image = updatedImage;
+      ctrl.usageRights = imageService(ctrl.image).usageRights;
+      ctrl.metadata = updatedImage.data.metadata;
+      ctrl.setUsageCategory(ctrl.usageCategories, ctrl.usageRights.data.category);
+    };
+
+    const freeUpdateListener = $rootScope.$on('images-updated', (e, updatedImages) => updatedImages.map(updatedImage => freeUpdateHandler(updatedImage)));
```
jonathonherbert, line 105: This is a little odd – are we only ever expecting a single image in `updatedImages`, or do we just care about the last one? (It seems that `freeUpdateHandler` only cares about the image it's currently receiving). 

If we just care about the last one, can we make that explicit by calling it once with the last element in `updatedImages`?

(v. possible I have wrong end of stick!)

jonathonherbert, line 105: What is a 'free update'? Is there a better name? (I know it's not your name – but it might be an opportunity to make things clearer!)

kahuna/public/js/edits/service.js
```diff
@@ -117,12 +117,14 @@ service.factory('editsService',
...
 
-    function registerUpdateRequest(resource, originalImage) {
+    function registerUpdateRequest(resource, originalImage, inBatch = false) {
         const requestPool = createRequestPool();
         const promise = withWatcher(resource, requestPool.promise).
               then(({ edit, image }) => {
-                  $rootScope.$emit('image-updated', image, originalImage);
-                  return edit;
+                if (!inBatch) {
+                  $rootScope.$emit('images-updated', [image]);
```
jonathonherbert, line 125: To confirm my understanding – the idea here is that if this caller knows it's not being called as a part of a batch, it sends an event itself; if it _is_ called as part of a batch, it delegates to the caller, which presumably is sending the event itself?

If that's correct, be good to document in the function description.

kahuna/public/js/edits/service.js
```diff
@@ -143,12 +145,12 @@ service.factory('editsService',
      * @param originalImage {Resource} the image used to check if we've re-indexed yet
      * @returns {Promise.<Resource>} completed when information is synced
```
jonathonherbert, line 146: We should update the function description here to match the new param (as above re: description)

kahuna/public/js/image/controller.js
```diff
@@ -229,9 +229,10 @@ image.controller('ImageCtrl', [
       });
     }
 
-    const freeImageUpdateListener = $rootScope.$on('image-updated', (e, updatedImage) => {
-      if (ctrl.image.data.id === updatedImage.data.id) {
-        ctrl.image = updatedImage;
+    const freeImagesUpdateListener = $rootScope.$on('images-updated', (e, updatedImages) => {
+      const maybeUpdatedImage = updatedImages.some(updatedImage => ctrl.image.data.id === updatedImage.data.id);
+      if (maybeUpdatedImage) {
+        ctrl.image = maybeUpdatedImage;
```
jonathonherbert, line 235: 👏 👏

kahuna/public/js/util/batch-tracking.js
```diff
@@ -1,27 +1,53 @@
...
-        reject,
-        resolve
-      })
-    );
+import PQueue from "p-queue";
+
+const concurrency = 30;
+
+
+export const trackAll = async ($rootScope, key, input, fns, emit) => {
```
jonathonherbert, line 6: This is a golden opportunity for us to document the intent of this function, and what each thing we're passing does, so the next developer who sees it stands a better chance of understanding it without going through the wiring ✨

kahuna/public/js/services/photoshoot.js
```diff
@@ -16,38 +16,38 @@ photoshootService.factory('photoshootService', [
...
-                    $rootScope.$emit('image-updated', newImage, image);
-                    return newImage;
-                });
+        function batchRemove({ images }) {
+          return trackAll(
+            $rootScope,
+            "photoshoot",
+            images,
+            [
+              (image) => imageAccessor.getPhotoshoot(image).delete(),
```
jonathonherbert, line 43: Wherever we're using `trackAll` in this way, I think it makes a lot of sense to be explicit as to what we're waiting for. For example (hopefully I have the semantics right!),

```js
function batchAdd({ data, images }) {
          const putPhotoshoot = (image) => { /* ... etc */ }
          const waitForPhotoshootInApi = (image, newPhotoshoot) => (image, newPhotoshoot) =>  { /* ... etc */ }
          return trackAll(
            $rootScope,
            "photoshoot",
            images,
            [putPhotoshoot, waitForPhotoshootInApi],
            "images-updated"
          );
        }
```

jonathonherbert:
@AWare this is great! Strat of batching everything by default and ignoring individual updates makes sense to me. 

There are a few console.logs etc. that are worth stripping out.

Aware (heh) I've just left a tonne of comments. V. happy to pair to go through, some are purely for my own comprehension!

kahuna/public/js/components/gr-add-label/gr-add-label.js
AWare: Not per se, but the alternative is to swallow the errors which imo is worse. If we're lucky these might end up in sentry even.

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
AWare: I think you're right, but I didn't really want to alter the logic.

AWare: Agreed, this is terribly named.

kahuna/public/js/edits/service.js
AWare: Yeah, where I could I pulled the functions apart properly. But `service.js` is too widely used and the implementation is far too clever to attempt refactoring- so I just bunged in an optional `inBatch` parameter.

kahuna/public/js/services/photoshoot.js
AWare: yeah, this is sensible.

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
AWare: Without better understanding of the function and the context in which it's run, I'm going to assume that `free` is important to the name in some obscure way that will become clear with sufficient introspection- and leave it be.

jonathonherbert: Oh, I think I understand – it's the result of assigning an event handler. Calling it presumably removes the handler.

Eh, this is probably lack of angular domain knowledge rather than a naming problem – although 'remove' might make things clearer. Guessing it's a convention throughout, so best leave it 👍

AWare: That makes sense, I'm going to get rid of the `free` from the handlers, which should imply this (absolutely genius) angular api design

kahuna/public/js/search/results.js
```diff
@@ -408,21 +408,34 @@ results.controller('SearchResultsCtrl', [
...
+          var index = ctrl.images.findIndex(i => i.data.id === updatedImage.data.id);
+          if (index !== -1) {
+            ctrl.images[index] = updatedImage;
+          }
+
+          var indexAll = ctrl.imagesAll.findIndex(i => i && i.data.id === updatedImage.data.id);
+          if (indexAll !== -1) {
+            ctrl.imagesAll[indexAll] = updatedImage;
+
+            // TODO: should not be needed here, the results list
```
tjsilver, line 423: Is this commented section still required?

AWare: yes, but not the `results.set` call and it's a few lines too high now (fixed)

tjsilver: 👍

kahuna/public/js/edits/service.js
```diff
@@ -117,12 +117,17 @@ service.factory('editsService',
     // update requests for a given Resource
     const updateRequestPools = new Map();
 
-    function registerUpdateRequest(resource, originalImage) {
+    // inBatch determines whether the function chain should eventually emit an angular message
+    // as emitting multiple times is very performance heavy
+    // ideally this should be refactored out.
+    function registerUpdateRequest(resource, originalImage, inBatch = false) {
```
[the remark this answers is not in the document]
tjsilver: Makes sense, thanks 👍

kahuna/public/js/services/label.js
```diff
@@ -34,44 +35,43 @@ labelService.factory('labelService',
...
-                return image;
-            });
+      return batchAdd([image], labels);
     }
 
-    function batchAdd (images, labels) {
-        return trackAll($rootScope, "label", images, image => add(image, labels));
+
+
+                        function batchAdd(images, labels) {
```
tjsilver, line 47: Nit: Lots of indentation in this file

AWare: prettier absolutely hates angular, happy to reflow

tjsilver (approved):
Great work! To my 'fresh' eyes it all makes sense. What an amazing speed improvement! 💯

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
jonathonherbert: 👍

jonathonherbert (approved):
Tested locally w/ 10 & 20 images, nice and quick, and the images in the search display seem to update correctly in response to e.g. rights changes. LGTM 👍

PR #3065, opened 2020-12-15
Add a comment about the nature of standard metadata fields

paperboyo:
For the record, they are being read and we reconcile them [here](https://github.com/guardian/grid/blob/85db4bc41d79d5906bb09f53a8279dd1bf40e68f/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala#L13-L97). We (mostly) try to follow [MWG guidelines](https://github.com/guardian/grid/files/4113995/mwg_guidance2.0.pdf).

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/ImageMetadata.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/ImageMetadata.scala
@@ -6,6 +6,8 @@ import play.api.libs.functional.syntax._
 import play.api.libs.json._
 
 case class ImageMetadata(
+/* these are standard metadata fields that exist in multiple schemas,
+most canonical being https://www.iptc.org/std/photometadata/specification/IPTC-PhotoMetadata */
   dateTaken:           Option[DateTime] = None,
   description:         Option[String]   = None,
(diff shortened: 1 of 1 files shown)


common-lib/src/main/scala/com/gu/mediaservice/model/ImageMetadata.scala
```diff
@@ -6,6 +6,8 @@ import play.api.libs.functional.syntax._
 import play.api.libs.json._
 
 case class ImageMetadata(
+/* these are standard metadata fields that exist in multiple schemas,
+most canonical being https://www.iptc.org/std/photometadata/specification/IPTC-PhotoMetadata */
```
sihil, line 10: Traditionally this comment would be before the start of the class if referring to the entire class and above a field if only referring to a field. Thus this looks like it just applies to `dateTaken` which is obviously not what you mean....

paperboyo: Tradition is correct. Here.

sihil (approved):

PR #3066, opened 2020-12-17
Switch app logo and favicon to SVG

paperboyo:
All the hard work done by @mxdvl, thank you!

This will give crisper results on higher resolution screens, while maintaining crispness on low resolution ones too.

The below shows a HiDPI screen (there is no difference on LoDPI ones); SVG versions on the right:

SVG all the thingz

- [x] tested on LOCAL


mxdvl:
Generated with: https://observablehq.com/@mxdvl/grid-logo

kahuna/public/images/grid-logo.svg
```diff
@@ -0,0 +1,3 @@
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 32 32" width="512" height="512">
```
mxdvl, line 1: You can and should use the same file for the favicon, if at all possible!

```suggestion
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 32 32" width="32" height="32">
```

kahuna/public/stylesheets/main.css
```diff
@@ -489,7 +489,8 @@ textarea.ng-invalid {
     text-indent: -9999px;
     width: 50px;
     height: 50px;
-    background: url(/assets/images/grid-logo-32.png) no-repeat center;
+    background: url(/assets/images/grid-logo.svg) no-repeat center;
+    background-size: 32px;
```
mxdvl, line 493: If you accept suggestion on height you can get rid of this!

```suggestion
```

mxdvl (changes requested):

kahuna/app/views/main.scala.html
```diff
@@ -22,7 +22,7 @@
     <link rel="preconnect" href="@mediaApiUri"/>
     <link rel="preconnect" href="@authApiUri"/>
 
-    <link rel="shortcut icon" type="image/png" href="@routes.Assets.versioned("images/grid-favicon-32.png")"/>
+    <link rel="shortcut icon" type="image/svg+xml" href="@routes.Assets.versioned("images/grid-favicon.svg")"/>
```
mxdvl, line 25: What’s the browser support for Grid? Link icons SVG [isn’t supported everywhere](https://caniuse.com/link-icon-svg).

paperboyo: I knew that and thought it’s fine. Apple continues to surprise me every single time. Thx, will amend. (and check if sizing changes haven’t broken “hinting”)

mxdvl:
Also, if using an SVG, I would recommend using a single file for both the `logo` and `favicon`.

And if you need PNG support, there are scripts that allow you to convert SVG to PNG, such as CLI `vips` and JS `sharp`.

paperboyo:
> Also, if using an SVG, I would recommend using a single file for both the `logo` and `favicon`.

They differ, because favicon has a background +colour). I’ve just left the old file as a fallback. We do not officially support Safari, but I agree it’s good practice.

kahuna/app/views/main.scala.html
paperboyo: Fixed.

kahuna/public/images/grid-favicon.svg
```diff
@@ -0,0 +1,4 @@
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 32 32" width="512" height="512">
```
mxdvl, line 1:
```suggestion
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 32 32" width="16" height="16">
```

mxdvl (approved):
LGTM!

[The future of favicon is bright](https://css-tricks.com/svg-favicons-and-all-the-fun-things-we-can-do-with-them/)

PR #3069, opened 2021-01-04
[BBC] BBC Pluggable ImageProcessors and the introduction of ImageProcessorResources

gribeiro:
## What does this change?

1. Makes an actor system available to processors so that updates can be scheduled for stores.
2. Rationalises the way the processor local and application/module config is passed through.
3. This is done by introducing a `ImageProcessorResources` type that includes the processor config, the common config and the actor system for injection into dynamically loaded processors.
4. Making the play actor system available in the config classes so it can be injected for the above.
5. Update with a first version of the BBC's metadata image processors.

We need to ensure that the changes introduced in item 1 do not impact the current application Image Processor and all the Image Processors still works with this change.

## Tested?
- [ x] locally
- [ ] on TEST


sihil:
Thanks for this @gribeiro - I'll take a look shortly.

common-lib/src/main/scala/com/gu/mediaservice/lib/bbc/components/BBCImageProcessorsDependencies.scala
```diff
@@ -0,0 +1,43 @@
...
+import com.gu.mediaservice.lib.bbc.BBCImageProcessorConfig
+import play.api.Configuration
+
+import scala.collection.mutable
+import scala.concurrent.ExecutionContext
+
+object BBCImageProcessorsDependencies {
+  implicit val ec = ExecutionContext.global
+
+  private def memoizeOnce[A, B](f: A => B): A => B = new ((A) => B) {
```
sihil, line 12: For me this memoisation logic could be confusing (were multiple configurations to be used for different processors, this will result in only the first one being initialised actually working). Having `save` be a Map[A, B] and only computing the output once for each different input might resolve this, assuming A correctly implements `Comparable`. 

That said this is logic that won't be used outside the BBC so I don't consider it as a blocker for merging.

common-lib/src/main/scala/com/gu/mediaservice/lib/bbc/components/BBCMetadataStore.scala
```diff
@@ -0,0 +1,55 @@
...
+import com.gu.mediaservice.lib.BaseStore
+import com.gu.mediaservice.lib.config.CommonConfig
+import org.joda.time.DateTime
+import play.api.Logger
+import play.api.libs.json.Json
+
+import scala.concurrent.ExecutionContext
+import scala.util.{Failure, Success, Try}
+
+class MetadataStore(bucket: String, config: CommonConfig)(implicit ec: ExecutionContext)
```
sihil, line 12: Might be best for consistency to name this class `BBCMetadataStore` to match the file.

image-loader/app/lib/ImageLoaderConfig.scala
```diff
@@ -50,12 +50,11 @@ class ImageLoaderConfig(playAppConfiguration: Configuration) extends CommonConfi
...
     * constructor. The constructor can either be no-arg or have a single argument of `play.api.Configuration`.
     *
-    * If configuration is specified but not used (a companion object or class with no arg constructor is specified)
-    * then loading the image processor will fail so as to avoid configuration errors.
+    * If a configuration is needed by is not provided by the config, the module configuration will be used instead.
     */
   val imageProcessor: ComposedImageProcessor = {
     val processors = configuration
-      .get[Seq[ImageProcessor]]("image.processors")(ImageProcessorLoader.imageProcessorsConfigLoader)
+      .get[Seq[ImageProcessor]]("image.processors")(ImageProcessorLoader.imageProcessorsConfigLoader(configuration))
```
sihil, line 57: I'd advocate passing in CommonConfig rather than the plain play config (so passing in `this` here).

image-loader/test/scala/lib/ImageProcessorLoaderTest.scala
```diff
@@ -65,27 +68,20 @@ class ImageProcessorLoaderTest extends FreeSpec with Matchers with EitherValues
...
-
-    "should fail to load a no arg processor that doesn't take configuration with non-empty configuration" in {
-      val instance = ImageProcessorLoader.loadImageProcessor(classOf[NoArgImageProcessor].getCanonicalName, testNonEmptyConfig)
-      instance.left.value shouldBe "Attempt to initialise image processor scala.lib.NoArgImageProcessor failed as configuration is provided but the constructor does not take configuration as an argument."
-    }
-
-    "should fail to load an object processor that doesn't take configuration with non-empty configuration" in {
-      val instance = ImageProcessorLoader.loadImageProcessor(ObjectImageProcessor.getClass.getCanonicalName, testNonEmptyConfig)
-      instance.left.value shouldBe "Attempt to initialise image processor scala.lib.ObjectImageProcessor$ failed as configuration is provided but the constructor does not take configuration as an argument."
-    }
```
sihil, line 77: I think both of these tests are useful to have and can be retained more easily be introducing a resources type.

image-loader/test/scala/lib/ImageProcessorLoaderTest.scala
```diff
@@ -143,21 +139,6 @@ class ImageProcessorLoaderTest extends FreeSpec with Matchers with EitherValues
...
-            "noClassName" -> "scala.lib.ConfigImageProcessor",
-            "config" -> Map("parameter" -> "value")
-          )
-        )
-      ))
-      val thrown = the[BadValue] thrownBy {
-        conf.get[Seq[ImageProcessor]]("some.path")
-      }
-      thrown.getMessage should include ("An image processor can either a class name (string) or object with className (string) and config (object) fields. This OBJECT is not valid.")
-    }
```
sihil, line 160: Likewise - this is a useful test as it helps ensure that an image processor cannot easily be mis-configured.

sihil (changes requested):
Hi @gribeiro - thanks for this PR.

It's great to see the pluggable image processor stuff being used and it's clear that you've hit the limitations that I imagined. I think you've made good steps to remove those limitations but it would be good to expand it and make it generally more powerful.

Rather than overload the single play `Configuration` argument with both the global/module config and also the processor config - I'd suggest introducing an `ImageProcessorResources` type that can carry multiple configurations and resources separately and use that as the single argument to image processors. At this stage we don't need to be backwards compatible so can just change it.

That also means we can include an actor system, which is required to schedule updates to the metadata store (if indeed you want them to dynamically update rather than update at startup only).

Getting an actor system is a bit non-trivial and will need a change to the way the configuration is handled in `GridComponents`.

I don't see the requirement to remove the tests as they are covering cases that still make sense to me. Likewise the assertion they rely on still makes sense to me. Please do make an argument for their removal if you disagree.

I've made a quick PR against your branch to sketch out some improvements. Happy to pair on hangout/zoom if you want to run through anything.

https://github.com/bbc/grid/pull/61

common-lib/src/main/scala/com/gu/mediaservice/lib/bbc/components/BBCImageProcessorsDependencies.scala
gribeiro: I agree, it makes sense if only it is used with the CommonConfiguration. The first version was a Map, but it has the potential to have some memory leak. For instance, if you let hot reload the config many times.
I will change the signature from ` ImageProcessorResources => MetadataStore` to `CommonConfig =>MetadataStore ` in order to be less confusing.

gribeiro:
Hi @sihil. 

I merged all the suggestions and did some minor changes and fixes.

1. Changed the name of the class MetadataStore to BBCMetadataStore.
2. Small fix on loadImageProcessor -- the implementation of actorSystem.
3. Created a BBCDependenciesConfig with the configuration that should be loaded only once.

Thank you :)

sihil (approved):
This is great work. I've left a couple of minor comments that you might want to address so I won't merge this until you give the go ahead - but I will test it in our integration environment so we are ready to merge when you're happy.

sihil:
@gribeiro We've tested this in our test environment and are happy to merge this when you're ready.

common-lib/src/main/scala/com/gu/mediaservice/lib/bbc/BBCSupplierProcessors.scala
```diff
@@ -0,0 +1,72 @@
...
+  ActionImagesParser,
+  AlamyParser,
+  AllStarParser,
+  ApParser,
+  CorbisParser,
+  EpaParser,
+  PaParser,
+  ReutersParser,
+  RexParser,
+  RonaldGrantParser
```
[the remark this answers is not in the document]
gribeiro: Yes. We will introduce more changes, we need this first version to do some testing first.

gribeiro:
Hello @sihil. Thank you for all the great suggestions.

I update a comment and added the descriptions to the ImageProcessors.
I also update the description and the title of the PR to better reflect our work.

Now the checks are failing on `lib.elasticsearch.SyndicationRightsOpsTest`.
I'm not sure what is happening. Do you have any idea what this might be?

PR #3073, opened 2021-01-06
Log whenever we add or remove collections.

AWare:
## What does this change?
Log whenever we add or remove collections.

that we log whenever we add or remove collections.
## Who should look at this?

we should

## Tested?
- [X] locally
- [ ] on TEST

### DIFF
--- a/collections/app/controllers/CollectionsController.scala
+++ b/collections/app/controllers/CollectionsController.scala
@@ -123,6 +123,7 @@ class CollectionsController(authenticated: Authentication, config: CollectionsCo
 
         store.add(collection).map { collection =>
           val node = Node(collection.path.last, Nil, collection.path, collection.path, Some(collection))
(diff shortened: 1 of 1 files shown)


collections/app/controllers/CollectionsController.scala
```diff
@@ -157,6 +158,7 @@ class CollectionsController(authenticated: Authentication, config: CollectionsCo
           s"$collectionPath has children, can't delete!"
         )
       } else {
+        logger.info(s"Deleting collection ${path.mkString("/")}")
```
jonathonherbert, line 161: Do we have any user information we can add here?

AWare: yes, had to have a good look for the right invokation

jonathonherbert (approved):
Tested locally, logs LGTM!

PR #3074, opened 2021-01-06
Filters deleted collections from collection children

SHession:
## What does this change?
This PR aims to fix the issue whereby all sub-collections wouldn't render if there was an orphaned sub-collection.

When an orphaned sub-collection exists, the other collections still render correctly.

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
+++ b/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
@@ -102,6 +102,7 @@ grCollectionsPanel.controller('GrNodeCtrl',
 
     const pathId = ctrl.node.data.data.pathId;
 
+    ctrl.children = ctrl.node.data.children.filter(node => !!node.data.data);
     ctrl.saving = false;
     ctrl.removing = false;
     ctrl.deletable = false;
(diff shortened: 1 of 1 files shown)


paperboyo:
cc @akash1810 for an opinion. My only worry is that this may hide the instances of errors, so we may be not alerted to them as currently only user feedback is responsible for alerting us. Granted, using users as reporting device is a bit 👿 .

SHession:
> cc @akash1810 for an opinion. My only worry is that this may hide the instances of errors, so we may be not alerted to them as currently only user feedback is responsible for alerting us. Granted, using users as reporting device is a bit 👿 .

There was a [separate PR](https://github.com/guardian/grid/pull/3073) to add logging for this issue on the server side. Therefore, we should still be able to tell when it happens.

akash1810 (commented):
Should this be done server side? Having this done client side suggests the server's response is correct, for some use cases - is that right?

AWare (approved):

SHession:
@akash1810, @AWare has attempted to implement this fix on the server side and found it to be much more complicated. This fix should be beneficial for now and we can remove it as/when we implement the server fix.

PR #3076, opened 2021-01-07
Revert "That’s not my name"

AWare:
Reverts guardian/grid#3048


akash1810:
Can we detail why this is needed please? What's wrong with #3048? Does anything need to be done to undo it's impact? etc.

AWare:
If I'd had a good reason to revert, I would have changed the PR. I opened this while investigating and then forgot to close it. Sorry.

PR #3078, opened 2021-01-07
Add check for date format

tjsilver:
## What does this change?

Adds a check that a date is in ISO 8601 format, follows #3077

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/thrall/app/lib/elasticsearch/ElasticSearch.scala
+++ b/thrall/app/lib/elasticsearch/ElasticSearch.scala
@@ -18,6 +18,7 @@ import lib.ThrallMetrics
 import org.joda.time.DateTime
 import play.api.MarkerContext
 import play.api.libs.json._
+import java.time.LocalDateTime
 
 import scala.concurrent.{ExecutionContext, Future}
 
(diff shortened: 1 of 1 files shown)


tonytw1 (approved):

akash1810 (changes requested):
Not certain about this, I think it would be better to fix this at the point the message is produced rather than within thrall. Adding this to thrall would mean any graphs/metrics we have of message production vs. consumption would no longer match up.

I can't quite remember how exceptions are handled within thrall. I _think_ they're handled gracefully somewhere in the process, but we'd need to double check. That is, check that throwing an `Exception` doesn't cause thrall to exit and stop processing other messages on the stream. I don't think an `Exception` is needed here, a log line or returning an `Either` would be better. That is, exceptions are for totally unexpected events, rather than controlling code flow.

Lastly, I think this PR demonstrates that thrall's design could be improved. To come this far in the process (we're in `updateImageUsages ` and about to write into the datastore) and still be uncertain about the validity of the data feels wrong. Thrall should serialise the payload on first receipt and use a strictly typed object from then, rather than `updateImageUsages ` taking a `JsLookupResult`. That is, fail fast. Though more involved, I think this would be a better change to make. Happy to pair.

sihil:
@tjsilver - when you reutrn can you comment on what the next steps are for this? Does any of the work done by @aug24 to improve the types passed to ES queries address the concerns that @akash1810 has?

tjsilver:
> @tjsilver - when you reutrn can you comment on what the next steps are for this? Does any of the work done by @aug24 to improve the types passed to ES queries address the concerns that @akash1810 has?

This was something we tried a few weeks ago when working on the 'Grid rights updates failures' issue, but it wasn't pursued. I think it's been superseded by an alternative solution. I will close this PR for now.

PR #3081, opened 2021-01-08
Adding BBC pan-domain lib version

wainaina:
## What does this change?
The BBC and Guardian use different versions of the pan-domain library. Whilst we wait for #3058 to be implemented and adopted this means that we must diverge our builds. This PR introduces the ability to control build dependencies based on an environment variable (`BUILD_ORG`).

If `BUILD_ORG` is set to `bbc` then the build pulls in the BBC specific pan-domain library: v0.9.2-SNAPSHOT to library dependencies.  Otherwise we use the published Guardian version.

When the environment variable BUILD_ORG=bbc is set, the library used is pan-domain library: v0.9.2-SNAPSHOT and this changes the flow to use BBC Login. 

image

## Who should look at this?
@guardian/digital-cms

## Tested?
- [x] locally
- [x] on TEST

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,6 +2,7 @@ import play.sbt.PlayImport.PlayKeys._
 
 import scala.sys.process._
 import scala.util.control.NonFatal
+import scala.collection.JavaConverters._
 
 val commonSettings = Seq(
   scalaVersion := "2.12.10",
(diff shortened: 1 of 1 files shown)


jonathonherbert (approved):
Code looks fine, I've tested this to verify that the absence of the env var builds as normal, and its presence induces a failure (on my machine) to resolve `0.9.2-SNAPSHOT`, as we'd expect.

As @sihil mentions out of band, you may find you want to introduce a check for the presence of this variable, or some output to indicate that dep X rather than Y is used to make it harder to make mistakes in your environment configuration, but we can do that later if it's found to be needed.

Let me know if there's anything else we'd like to test!

akash1810 (changes requested):
In an effort to keep Grid organisationally agnostic (and ignoring the Guardian heavy image ingestion process for now!), could we make the env var related to the pan domain library?

That is, rather than `BUILD_ORG`, read the env var `PAN_DOMAIN_VERSION`. This would also be more intention revealing:

```scala
val panDomainAuthVersion = sys.env.getOrElse("PAN_DOMAIN_VERSION", "0.8.2")
```

We can then use this `val` in a similar way as we're using `awsSdkVersion`.

sihil:
That's a smart idea @akash1810. I would normally accept it except the BBC have already implemented my less good idea in their CI and pragmatically, given the temporary nature of this, I'd prefer to merge this and ensure it's all removed when the pluggable auth is completed (likely in the next few weeks). 

Likewise, the Guardian heavy image ingestion process, whilst the default, is no longer bound in code - it's possible to replace by config, as the BBC have done. It would be great to implement a default set of rules in the future.

akash1810 (approved):
> the BBC have already implemented my less good idea in their CI and pragmatically, given the temporary nature of this, I'd prefer to merge this and ensure it's all removed when the pluggable auth is completed (likely in the next few weeks).

Hadn't realised this - happy to get this merged now for ease and removed in the near future.

PR #3082, opened 2021-01-10
Image quarantine

Asmau1234:
## What does this change?

This adds an AWS S3 bucket which is used to quarantine images before the are scanned by the chosen scanner..The aim of the quarantine bucket is to make sure no virus infected images are allowed into the ingest bucket.  From the quarantine bucket, the selected installed scanner should scan images then import them into the ingest bucket if they are virus free. 

When images are uploaded onto the grid, ImageLoader pushes the images to quarantine bucket:

- Sling-scanner (selected scanner for bbc grid) downloads the image from the quarantine bucket.
- Sling-scanner scans the image.
- If clean, the image is uploaded by sling-scanner to the ingest bucket and deleted from the quarantine bucket. If the image is not clean (i.e contains a virus), a notification is raised and the image is deleted from the quarantine bucket.
- The ingest lambda receives the scanned image upload event in the ingest bucket via SNS and does a POST request to ImageLoader which downloads and processes the image.

This is an implementation to this proposal opened as an issue [here](https://github.com/guardian/grid/issues/3123). 
## Who should look at this?
@guardian/digital-cms 

## Tested?
- [X] locally by contributor
- [x] locally by Guardian team
- [x] on Guardian TEST environment

### DIFF
--- a/dev/script/generate-config/service-config.js
+++ b/dev/script/generate-config/service-config.js
@@ -56,11 +56,13 @@ function getImageLoaderConfig(config) {
         |aws.region="${config.AWS_DEFAULT_REGION}"
         |s3.image.bucket="${config.coreStackProps.ImageBucket}"
         |s3.thumb.bucket="${config.coreStackProps.ThumbBucket}"
+        |s3.quarantine.bucket="${config.coreStackProps.QuarantineBucket}"
         |s3.config.bucket="${config.coreStackProps.ConfigBucket}"
         |aws.local.endpoint="https://localstack.media.${config.DOMAIN}"
         |security.cors.allowedOrigins="${getCorsAllowedOriginString(config)}"
         |metrics.request.enabled=false
         |transcoded.mime.types="image/tiff"
+        |upload.quarantine.enabled=false
         |`;
 }
 
--- a/image-loader/app/ImageLoaderComponents.scala
+++ b/image-loader/app/ImageLoaderComponents.scala
@@ -3,8 +3,8 @@ import com.gu.mediaservice.lib.logging.GridLogging
 import com.gu.mediaservice.lib.play.GridComponents
 import controllers.ImageLoaderController
 import lib._
-import lib.storage.ImageLoaderStore
-import model.{Projector, Uploader}
+import lib.storage.{ImageLoaderStore, QuarantineStore}
+import model.{Projector, Uploader, QuarantineUploader}
 import play.api.ApplicationLoader.Context
 import router.Routes
 
@@ -18,15 +18,20 @@ class ImageLoaderComponents(context: Context) extends GridComponents(context, ne
 
   val store = new ImageLoaderStore(config)
   val imageOperations = new ImageOperations(context.environment.rootPath.getAbsolutePath)
-
   val notifications = new Notifications(config)
   val downloader = new Downloader()
   val uploader = new Uploader(store, config, imageOperations, notifications)
-
   val projector = Projector(config, imageOperations)
-
+  val quarantineUploader: Option[QuarantineUploader] = (config.uploadToQuarantineEnabled, config.quarantineBucket) match {
+    case (true, Some(bucketName)) =>{
+      val quarantineStore = new QuarantineStore(config)
+      Some(new QuarantineUploader(quarantineStore, config))
+    }
+    case (true, None) => throw new IllegalArgumentException(s"Quarantining is enabled. upload.quarantine.enabled = ${config.uploadToQuarantineEnabled} but no bucket is configured. s3.quarantine.bucket isn't configured.")
+    case (false, _) => None
+  }
   val controller = new ImageLoaderController(
-    auth, downloader, store, notifications, config, uploader, projector, controllerComponents, wsClient)
+    auth, downloader, store, notifications, config, uploader, quarantineUploader, projector, controllerComponents, wsClient)
 
   override lazy val router = new Routes(httpErrorHandler, controller, management)
 }
--- a/image-loader/app/controllers/ImageLoaderController.scala
+++ b/image-loader/app/controllers/ImageLoaderController.scala
@@ -13,11 +13,11 @@ import com.gu.mediaservice.model.UnsupportedMimeTypeException
 import lib._
 import lib.imaging.{NoSuchImageExistsInS3, UserImageLoaderException}
 import lib.storage.ImageLoaderStore
-import model.{Projector, Uploader}
-import play.api.libs.json.Json
+import model.{Projector, Uploader, QuarantineUploader}
+import play.api.libs.json.{JsObject, Json}
 import play.api.libs.ws.WSClient
 import play.api.mvc._
-
+import model.upload.UploadRequest
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.Try
 import scala.util.control.NonFatal
@@ -28,6 +28,7 @@ class ImageLoaderController(auth: Authentication,
                             notifications: Notifications,
                             config: ImageLoaderConfig,
                             uploader: Uploader,
+                            quarantineUploader: Option[QuarantineUploader],
                             projector: Projector,
                             override val controllerComponents: ControllerComponents,
                             wSClient: WSClient)
@@ -45,7 +46,12 @@ class ImageLoaderController(auth: Authentication,
 
   def index: Action[AnyContent] = auth { indexResponse }
 
-  def loadImage(uploadedBy: Option[String], identifiers: Option[String], uploadTime: Option[String], filename: Option[String]): Action[DigestedFile] = {
+  def quarantineOrStoreImage(uploadRequest: UploadRequest)(implicit logMarker: LogMarker) = {
+    quarantineUploader.map(_.quarantineFile(uploadRequest)).getOrElse(uploader.storeFile(uploadRequest))
+  }
+  
+  def loadImage(uploadedBy: Option[String], identifiers: Option[String], uploadTime: Option[String], filename: Option[String]): Action[DigestedFile] =  {
+
     implicit val context: RequestLoggingContext = RequestLoggingContext(
       initialMarkers = Map(
         "requestType" -> "load-image",
@@ -72,9 +78,8 @@ class ImageLoaderController(auth: Authentication,
           DateTimeUtils.fromValueOrNow(uploadTime),
           filename.flatMap(_.trim.nonEmptyOpt),
           context.requestId)
-        result <- uploader.storeFile(uploadRequest)
+        result <- quarantineOrStoreImage(uploadRequest)
       } yield result
-
       result.onComplete( _ => Try { deleteTempFile(tempFile) } )
 
       result map { r =>
@@ -88,8 +93,8 @@ class ImageLoaderController(auth: Authentication,
             case e: UnsupportedMimeTypeException => FailureResponse.unsupportedMimeType(e, config.supportedMimeTypes)
             case e: ImageProcessingException => FailureResponse.notAnImage(e, config.supportedMimeTypes).as(ArgoMediaType)
             case e: java.io.IOException => FailureResponse.badImage(e).as(ArgoMediaType)
-            case e =>
-              logger.error("Failed upload", e)
+            case _ => 
+              logger.error("Failed upload", e) 
               InternalServerError(Json.obj("error" -> e.getMessage)).as(ArgoMediaType)
           }).as(ArgoMediaType)
       }
--- a/image-loader/app/lib/ImageLoaderConfig.scala
+++ b/image-loader/app/lib/ImageLoaderConfig.scala
@@ -10,6 +10,8 @@ class ImageLoaderConfig(resources: GridConfigResources) extends CommonConfig(res
   val imageBucket: String = string("s3.image.bucket")
 
   val thumbnailBucket: String = string("s3.thumb.bucket")
+  val quarantineBucket: Option[String] = stringOpt("s3.quarantine.bucket")
+  val uploadToQuarantineEnabled: Boolean = boolean("upload.quarantine.enabled")
 
   val tempDir: File = new File(stringDefault("upload.tmp.dir", "/tmp"))
 
--- a/image-loader/app/model/QuarantineUploader.scala
+++ b/image-loader/app/model/QuarantineUploader.scala
@@ -1,0 +1,53 @@
+package model
+
+
+import com.gu.mediaservice.lib.argo.ArgoHelpers
+import com.gu.mediaservice.lib.auth.Authentication
+import com.gu.mediaservice.lib.auth.Authentication.Principal
+import com.gu.mediaservice.lib.aws.{S3Object, UpdateMessage}
+import com.gu.mediaservice.lib.imaging.ImageOperations
+import com.gu.mediaservice.lib.logging.{LogMarker, Stopwatch, addLogMarkers}
+import com.gu.mediaservice.lib.logging.MarkerMap
+import com.gu.mediaservice.lib.resource.FutureResources._
+import com.gu.mediaservice.model._
+import lib.ImageLoaderConfig
+import lib.storage.QuarantineStore
+import net.logstash.logback.marker.LogstashMarker
+import org.joda.time.DateTime
+import play.api.Logger
+import play.api.libs.json.{JsObject, Json}
+import com.gu.mediaservice.lib.formatting._
+import model.upload.UploadRequest
+import java.net.URLEncoder
+import java.nio.charset.StandardCharsets
+import scala.concurrent.{ExecutionContext, Future}
+
+class QuarantineUploader(val store: QuarantineStore,
+               val config: ImageLoaderConfig)
+              (implicit val ec: ExecutionContext) extends ArgoHelpers {
+
+  private def storeQuarantineFile(uploadRequest: UploadRequest)
+                         (implicit logMarker: LogMarker) = {
+    val meta = Uploader.toMetaMap(uploadRequest)
+    store.storeQuarantineImage(
+      uploadRequest.imageId,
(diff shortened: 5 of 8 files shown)


akash1810 (dismissed):
Thanks for the contribution @Asmau1234 ! Is there any reason why all the file permissions are changing to include execute permissions?

Asmau1234:
> Thanks for the contribution @Asmau1234 ! Is there any reason why all the file permissions are changing to include execute permissions?

Thank you for the review @akash1810 I have pushed new commits with the right file permissions now.

akash1810:
I vaguely remember seeing a diagram for how this change works - would it be possible to add the diagram to the PR to make this that much easier to review?

Asmau1234:
> I vaguely remember seeing a diagram for how this change works - would it be possible to add the diagram to the PR to make this that much easier to review?

Yup. I'll add it.

thrall/app/lib/ThrallConfig.scala
```diff
@@ -34,6 +34,7 @@ class ThrallConfig(resources: GridConfigResources) extends CommonConfig(resource
   val writeAlias: String = string("es.index.aliases.write")
 
   val thumbnailBucket: String = string("s3.thumb.bucket")
+  val quarantineBucket: String = string("s3.quarantine.bucket")
```
sihil, line 37: This isn't used by thrall so it would be best to not require it in the configuration.

common-lib/src/main/scala/com/gu/mediaservice/lib/ImageIngestOperations.scala
```diff
@@ -38,6 +38,10 @@ class ImageIngestOperations(imageBucket: String, thumbnailBucket: String, config
   private def storeOptimisedImage(storableImage: StorableOptimisedImage)
                        (implicit logMarker: LogMarker): Future[S3Object] =
     storeImage(imageBucket, optimisedPngKeyFromId(storableImage.id), storableImage.file, Some(storableImage.mimeType))
+  
+  def sendToQuarantine(id: String, file: File, mimeType: Option[MimeType], meta: Map[String, String] = Map.empty)
```
sihil, line 42: I wonder whether this would better live in a `ImageQuarantineOperations` class. By separating this we can separate responsibilities and no longer need to configure the bucket in the Thrall service.

dev/config/valid_emails.json
```diff
@@ -0,0 +1,3 @@
+[
+  "grid-demo-account@example.com"
+]
```
sihil, line 3: I think that this file might have leaked in from auth work? I can't see anywhere it is referenced.

dev/script/generate-config/service-config.js
```diff
@@ -142,6 +144,7 @@ function getThrallConfig(config) {
         |aws.region="${config.AWS_DEFAULT_REGION}"
         |s3.image.bucket="${config.coreStackProps.ImageBucket}"
         |s3.thumb.bucket="${config.coreStackProps.ThumbBucket}"
+        |s3.quarantine.bucket="${config.coreStackProps.QuarantineBucket}"
```
sihil, line 147: hopefully this can be removed if `sendToQuarantine` is extracted into a `ImageQuarantineOperations` as suggested above.

image-loader/app/model/Uploader.scala
```diff
@@ -77,6 +84,43 @@ case class ImageUploadOpsDependencies(
...
+      deps: QuarantineImageUploadOpsDependencies)(
+      implicit ec: ExecutionContext,
+      logMarker: LogMarker): Future[JsObject] = {
+
+    import deps._
+
+    uploadAndStoreQuarantineImage(config,
+                                  storeOriginalFile,
+                                  uploadRequest,
+                                  deps)
```
sihil, line 103: Both the deps object and individual fields from it are passed through into this function. It would be cleaner if one approach was chosen.

sihil (commented):
Hi @Asmau1234 thanks for this PR.

I've made a number of suggestions for improvement. Happy to follow up here, on slack or zoom to clarify anything that's not clear.

As @akash1810 has already said, the PR description doesn't make it clear how the big picture works. From looking at the code I'm assuming that:
 - the user uploads to `POST /images` as usual
 - image-loader pushes the file to quarantine with appropriate object metadata
 - a lambda picks up the file and scans it
   - if quarantine passes, the file is then pushed into POST /imports with the object metadata passed in as query parameters and the normal process resumes and a user sees the image appear
   - if quarantine fails, the file is dropped on the floor and the user doesn't get notified (without changes to the frontend in #3085)

~It would be good to describe that process in the PR, possibly with a diagram.~ (Update: I see you've already done this, thanks!) 

I would also even consider having some documentation for this under `docs/07-image-quarantine/`? It would also be great to have the code of the lambda that pushes the images back in available in the main grid repo if it is generic enough - as this would help complete the picture from a code perspective.

sihil:
At a glance this looks like the quarantine code is much more isolated which is great. I'll do a follow up review later today.

Asmau1234:
> At a glance this looks like the quarantine code is much more isolated which is great. I'll do a follow up review later today.

Thanks Simon. I tried to make it a lot more isolated based on your recommendations. I don't mind further modifications though so i'll look out for your feedback later. I'll also be adding some documentation to `docs/07-image-quarantine/` later.

image-loader/app/model/QuarantineUploader.scala
```diff
@@ -0,0 +1,120 @@
...
+                                  storeFile: UploadRequest => Future[S3Object],
+                                  uploadRequest: UploadRequest
+                              )
+                                 (implicit ec: ExecutionContext, logMarker: LogMarker): Future[JsObject] = {
+    Logger.info("storing quarantine file")
+      val sourceStoreFuture = storeFile(uploadRequest)
+       val s3Source = for{
+                        s3Source <- sourceStoreFuture
+                      }yield (s3Source)
+      s3Source.map(result => Json.obj("status" -> "uploaded to quarantine bucket")) recover {case e => Json.obj("error" -> e.getMessage)}
```
sihil, line 59: I don't think this `map` adds anything here and as discussed (at the call site) the `recover` discards the `Failure` state. Furthermore, the `Success` value that is returned is discarded anyway. Perhaps this function best returns a `Future[Unit]`?

image-loader/app/model/QuarantineUploader.scala
```diff
@@ -0,0 +1,120 @@
...
+  private def uploadAndStoreQuarantineImage(
+                                  storeFile: UploadRequest => Future[S3Object],
+                                  uploadRequest: UploadRequest
+                              )
+                                 (implicit ec: ExecutionContext, logMarker: LogMarker): Future[JsObject] = {
+    Logger.info("storing quarantine file")
+      val sourceStoreFuture = storeFile(uploadRequest)
+       val s3Source = for{
+                        s3Source <- sourceStoreFuture
+                      }yield (s3Source)
```
sihil, line 58: Minor: Indentation.

image-loader/app/model/QuarantineUploader.scala
```diff
@@ -0,0 +1,120 @@
...
+    val meta = toMetaMap(uploadRequest)
+    store.storeQuarantineImage(
+      uploadRequest.imageId,
+      uploadRequest.tempFile,
+      uploadRequest.mimeType,
+      meta
+    )
+  }
+
+    def quarantineFile(uploadRequest: UploadRequest)(
```
sihil, line 107: Minor: indentation.

image-loader/app/model/Uploader.scala
```diff
@@ -267,7 +267,7 @@ class Uploader(val store: ImageLoaderStore,
               (implicit val ec: ExecutionContext) extends ArgoHelpers {
 
 
-
+  import Uploader._
```
sihil, line 270: superflous change - this import isn't used.

image-loader/app/model/Uploader.scala
```diff
@@ -57,7 +57,7 @@ case object ImageUpload {
     )
   }
 }
-
+
```
sihil, line 60: Minor: noise.

image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -88,9 +89,7 @@ class ImageLoaderController(auth: Authentication,
             case e: UnsupportedMimeTypeException => FailureResponse.unsupportedMimeType(e, config.supportedMimeTypes)
             case e: ImageProcessingException => FailureResponse.notAnImage(e, config.supportedMimeTypes).as(ArgoMediaType)
             case e: java.io.IOException => FailureResponse.badImage(e).as(ArgoMediaType)
-            case e =>
-              logger.error("Failed upload", e)
-              InternalServerError(Json.obj("error" -> e.getMessage)).as(ArgoMediaType)
```
sihil, line 93: Not sure if I missed this last time, but this log line seems useful to me. Otherwise we do not log the full exception of the problem preventing file upload (the next line only extracts the message from the exception).

image-loader/app/lib/ImageLoaderConfig.scala
```diff
@@ -10,6 +10,11 @@ class ImageLoaderConfig(resources: GridConfigResources) extends CommonConfig(res
   val imageBucket: String = string("s3.image.bucket")
 
   val thumbnailBucket: String = string("s3.thumb.bucket")
+  val quarantineBucket: String = string("s3.quarantine.bucket")
+  val uploadToQuarantineEnabled: Boolean = stringDefault("upload.quarantine.enabled", "false").toLowerCase match {
+    case "true" => true
+    case _ => false
+  }
```
sihil, line 17: I meant to say last time that this would be better expressed using the boolean helper method.
```suggestion
  val uploadToQuarantineEnabled: Boolean = boolean("upload.quarantine.enabled")
```

image-loader/app/model/QuarantineUploader.scala
```diff
@@ -0,0 +1,120 @@
...
+
+  def fromQuarantineUploadRequestShared(uploadRequest: UploadRequest, deps: QuarantineImageUploadOpsDependencies)
+                             (implicit ec: ExecutionContext, logMarker: LogMarker): Future[JsObject] = {
+
+    Logger.info("Starting image ops")
+    
+      uploadAndStoreQuarantineImage(
+        deps.storeFile,
+        uploadRequest
+        )
```
sihil, line 46: Minor: indentation in this method.

sihil (commented):
The relocation of this functionality is much better now and falls out nicely in the `ImageLoaderComponents`. It's great that it no longer touches Thrall.

My main remaing concern is an issue with the way that quarantine upload errors are being handled as I suspect any such failure will be failing silently.

There are also a handful of minor things around style and indentation (we should really put a linter on this project)!

sihil (commented):
Code wise I think this is ready to go. I'll test it locally in our test environment today. The only change to `Uploader.scala` is now whitespace so it would be great if we could eliminate those. Finally, since there are no changes to Kahuna, we don't want to commit ` kahuna/package-lock.json`.

Otherwise I think this is good to go.

sihil:
Thanks for this PR @Asmau1234. I've just tried deploying it to our test environment but unfortunately the deployment failed as the quarantine bucket is not configured there:
```
java.lang.RuntimeException: Required string configuration property missing: s3.quarantine.bucket
        at scala.sys.package$.error(package.scala:30)
        at com.gu.mediaservice.lib.config.CommonConfig.missing(CommonConfig.scala:98)
        at com.gu.mediaservice.lib.config.CommonConfig.$anonfun$string$1(CommonConfig.scala:81)
        at scala.Option.getOrElse(Option.scala:189)
        at com.gu.mediaservice.lib.config.CommonConfig.string(CommonConfig.scala:81)
        at lib.ImageLoaderConfig.<init>(ImageLoaderConfig.scala:13)
        at ImageLoaderComponents$$anonfun$$lessinit$greater$1.apply(ImageLoaderComponents.scala:11)
        at ImageLoaderComponents$$anonfun$$lessinit$greater$1.apply(ImageLoaderComponents.scala:11)
        at com.gu.mediaservice.lib.play.GridComponents.<init>(GridComponents.scala:22)
        at ImageLoaderComponents.<init>(ImageLoaderComponents.scala:11)
        at AppLoader$$anonfun$$lessinit$greater$1.apply(AppLoader.scala:3)
        at AppLoader$$anonfun$$lessinit$greater$1.apply(AppLoader.scala:3)
        at com.gu.mediaservice.lib.play.GridAppLoader.load(GridAppLoader.scala:13)
        at play.core.server.ProdServerStart$.start(ProdServerStart.scala:51)
        at play.core.server.ProdServerStart$.main(ProdServerStart.scala:25)
        at play.core.server.ProdServerStart.main(ProdServerStart.scala)
```

Ideally we would be able to deploy this, with it defaulting to not quarantining, without needing the configuration.

image-loader/app/ImageLoaderComponents.scala
```diff
@@ -18,15 +18,20 @@ class ImageLoaderComponents(context: Context) extends GridComponents(context, ne
...
   val imageOperations = new ImageOperations(context.environment.rootPath.getAbsolutePath)
-
   val notifications = new Notifications(config)
   val downloader = new Downloader()
   val uploader = new Uploader(store, config, imageOperations, notifications)
-
   val projector = Projector(config, imageOperations)
-
+  val checkQuarantineFlag = config.uploadToQuarantineEnabled 
+  val quarantineUploader: Option[QuarantineUploader] = checkQuarantineFlag match {
```
sihil, line 26: I think `checkQuarantineFlag` is superflous here. I'd put it directly inline.

```suggestion
  val quarantineUploader: Option[QuarantineUploader] = config.uploadToQuarantineEnabled match {
```

image-loader/app/lib/ImageLoaderConfig.scala
```diff
@@ -10,6 +10,8 @@ class ImageLoaderConfig(resources: GridConfigResources) extends CommonConfig(res
   val imageBucket: String = string("s3.image.bucket")
 
   val thumbnailBucket: String = string("s3.thumb.bucket")
+  val quarantineBucket: String = stringDefault("s3.quarantine.bucket", "")
```
sihil, line 13: `stringOpt` is a much better way of expressing a configuration option that might not be present. This will give you an Option[String].

image-loader/app/ImageLoaderComponents.scala
```diff
@@ -18,15 +18,20 @@ class ImageLoaderComponents(context: Context) extends GridComponents(context, ne
...
+    case true =>{
+      val quarantineStore = new QuarantineStore(config)
+      Some(new QuarantineUploader(quarantineStore, config))
+    }
+    case false => None
+  }
```
sihil, line 32: If you make quarantineBucket optional then you can do something like this to deal with the Boolean and Option[String]:

```suggestion
  val quarantineUploader: Option[QuarantineUploader] = (config.uploadToQuarantineEnabled, config.quarantineBucket) match {
    case (true, Some(bucketName)) =>{
      val quarantineStore = new QuarantineStore(bucketName, config)
      Some(new QuarantineUploader(quarantineStore, config))
    }
    case (true, None) => throw new IllegalArgumentException("Quarantining is enabled but no bucket is configured")
    case (false, _) => None
  }
```

The advantage of this approach is that a bad config (when quarantine is enabled but no bucket is specified) fails fast.

image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -72,9 +74,8 @@ class ImageLoaderController(auth: Authentication,
           DateTimeUtils.fromValueOrNow(uploadTime),
           filename.flatMap(_.trim.nonEmptyOpt),
           context.requestId)
-        result <- uploader.storeFile(uploadRequest)
+        result <- if (config.uploadToQuarantineEnabled && !config.quarantineBucket.isEmpty) quarantineUploader.get.quarantineFile(uploadRequest) else uploader.storeFile(uploadRequest)
```
sihil, line 77: You've already checked the config when figuring out whether to create the QuarantineUploader. So this can simply be `if (quarantineUploader.nonEmpty) quarantineUploader.get.quarantineFile(uploadRequest) else uploader.storeFile(uploadRequest)`. Or, more idiomatically `quarantineUploader.map(_.quarantineFile(uploadRequest)).getOrElse(uploader.storeFile(uploadRequest)`

Might be nicer to put it into a small function at the top so as to express the intent more clearly.

image-loader/app/ImageLoaderComponents.scala
sihil: Note that this now passes the quaratine bucket directly into QuarantineUploader (which deals with the config being an option).

image-loader/app/ImageLoaderComponents.scala
```diff
@@ -18,15 +18,20 @@ class ImageLoaderComponents(context: Context) extends GridComponents(context, ne
...
   val uploader = new Uploader(store, config, imageOperations, notifications)
-
   val projector = Projector(config, imageOperations)
-
+  val quarantineUploader: Option[QuarantineUploader] = (config.uploadToQuarantineEnabled, config.quarantineBucket) match {
+    case (true, Some(bucketName)) =>{
+      val quarantineStore = new QuarantineStore(config)
+      Some(new QuarantineUploader(quarantineStore, config))
+    }
+    case (true, None) => throw new IllegalArgumentException("Quarantining is enabled but no bucket is configured")
```
sihil, line 30: A nice touch would be to include the configuration names in this message. i.e. `Quarantining is enabled (xxx.yyy=true) but no bucket is configured (zzz.aaa isn't configured)`.

This would save users having to look at the source code.

sihil (approved):
This works as expected on the Guardian's TEST environment.

One minor comment, but approving and happy to merge to main.

sihil (approved):

PR #3086, opened 2021-01-11
Fix credit:Allstar to be correctly recognised

paperboyo:
This is a request from Picture Desk for these to be correctly categorised as Allstar-supplied.

I would love for the credit to be rewritten so that both the `prefix` and `suffix` would appear **before** Allstar string, eg.:

Current credit: `Allstar/UNIVERSAL PICTURES`
Better credit: `Universal Pictures/Allstar` 
[let’s use title case, it will be more correct than not; ugly capitalisation is ugly]

This PR covers the following tasks: 
1. Recognise as 'Allstar' (without subcollection) when the whole credit reads 'Allstar'
2. Move 'Allstar' to the end of the credit
3. Title case the prefix/suffix
4. Remove 'Allstar' or 'Allstar Picture Library' from byline and append to end of credit as 'Allstar', if not already present
5. Update tests to reflect these changes

## Tested?
- [x ] locally
- [ ] on TEST


tjsilver:
[Fix Allstar processor](https://trello.com/c/petCrAZv/2147-fix-allstar-processor)

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -74,33 +75,109 @@ object AlamyParser extends ImageProcessor {
...
 
-object AllStarParser extends ImageProcessor {
-  val SlashAllstar = """(.+)/Allstar""".r
-  val AllstarSlash = """Allstar/(.+)""".r
+object AllStarParser extends CanonicalisingImageProcessor {
+  override def getAgencyName = "Allstar Picture Library"
+  override def getCanonicalName: String = "Allstar"
+
+  case object AllstarRegexResultExtractor extends RegexResultExtractor {
+    private val AllstarInSlashDelimitedString = "((.*)/)?(Allstar( Picture Library)?)(/(.*))?".r
```
jonathonherbert, line 83: This API is fairly expansive, given that AFAICS we only care about the prefix/suffix. (Adding a new RegexResultExtractor will mean doing the capture group dance.) Given the two examples here, feels possible to write something with an API that just consumes the regex to find the agency, and uses that to find prefixes/suffixes. IIUC this would mean a single definition of the search regex, rather than a custom unapply method,  for each extractor.

But this is well explained and tested – more asking for my own curiosity, as I can't think of a case w/ prefix/suffix that wouldn't work with this smaller API – perhaps there is one, though?

jonathonherbert (approved):
Fancier than expected – tests and comments are really helpful ✨  one point about the API that may/may not apply, but otherwise LGTM.

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
aug24: In part, you're correct.  We didn't need the unapply, which was clever but not clear.  We've replaced with an abstract method.  This is needed because the regexes are not necessarily the same between the two sub-classes, and may have different numbers of capture groups.

jonathonherbert: Gotcha. My suggestion above assume the form `prefix/agency/suffix`. Out-of-band discussions have suggested that this may not always be the case (I can't see examples in the tests, but a) I may have missed and b) there may be some in future), and we'd need a more expansive  API to deal with that, so happy to leave as is.

PR #3087, opened 2021-01-11
Get rid of clone calls in Painless

AWare:
## What does this change?

Replaces our calls to `.clone()` in painless code (which is actually bizzaro java) with a proper new hashmap and `putAll`. This is because sometimes it seemed that the maps did not implement `Clonable`. This was an absolute mystery and we've nothing useful to say regarding our investigations. 

https://discuss.elastic.co/t/painless-map-copies-are-painful/106258 

Except that elastic don't mention cloning where it would be obvious and recommend using puts instead. 

Reliable rights updates.

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/thrall/app/lib/elasticsearch/ElasticSearch.scala
+++ b/thrall/app/lib/elasticsearch/ElasticSearch.scala
@@ -518,11 +518,14 @@ class ElasticSearch(config: ElasticSearchConfig, metrics: Option[ThrallMetrics])
   }
 
   private val refreshMetadataScript = """
-      | ctx._source.metadata = ctx._source.originalMetadata.clone();
+      | ctx._source.metadata = new HashMap();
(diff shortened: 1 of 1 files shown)


jonathonherbert (approved):
The rationale makes sense – well done for finding 👏 Approving on the grounds that this is urgent – is it possible to test this? Or do our current tests cover the core functionality (copying across metadata) for all but a few difficult to reproduce edge cases?

AWare:
If you mean, is it possible to make the current tests fail because of the clone call:

No. We've not been able to find a reproduction of this error locally at all.

Otherwise: 
The current test suite validates that this works normally.

sihil:
Have we tried to identify the structure of the data that was cauing the problems? If we could and could recreate it precisely one would hope it would be possible to reproduce?

PR #3088, opened 2021-01-12
Improve signature of ElasticSearch methods.

aug24:
## What does this change?
Currently the methods which write to ElasticSearch take a JSDefined object.

This, of course, has no type safety - any serialised json object will do.  We should not serialise until the last possible moment.  This PR pushes the serialisation into the ElasticSearch code to ensure type safety until json is actually required.

Additionally, there is inconsistency in the use of lastModified.  In some cases it is required (from the message), in others it is generated from the system.  In once case it was passed in from the message and then ignored in favour of the system date.  

As a result of recent changes, it is currently possible to write an older lastModified date (from the low-priority stream) over a more recent lastModified date (from the high-priority stream).

Code is refactored to provide a consistent approach and write back the higher of the two values: current lastModified date and the lastModified date from the message.

The ElasticSearch class methods should not have any "weakly typed" arguments.
The last modified data should always be the date of the latest write.

## Tested?
- [ ] locally
- [ ] on TEST


SHession (approved):
Reviewed code changes thoroughly with @aug24 and tested locally!

PR #3091, opened 2021-01-13
js fixes for batch upload

AWare:
## What does this change?

changes a `.some` which returns `true` to a `.find` which returns what was found 	

## Tested?
- [ ] locally
- [ ] on TEST


jonathonherbert (approved):

jonathonherbert:
Tested locally, fix seems clear: shame we don't have an easy way of writing an integration test to prevent regressions, but this is producing a large number of errors in Sentry, so it makes sense to get a fix out.

PR #3092, opened 2021-01-13
remove id that's not being used and isn't unique

AWare:
## What does this change?

removes `id="access-select"` from select component that's used for each uploaded image and isn't used anywhere

not spamming console on upload page

## Tested?
- [ ] locally
- [ ] on TEST


AWare:
merged in #3093

PR #3093, opened 2021-01-13
Fix batch tracking on upload page (and speed it up)

AWare:
## What does this change?

Reduces the timeout for polling to .5 seconds, reducing min time from 4s. 

Also calls the dispatch loop because batchTracking uses native promises.

(Also includes #3092)

- Updates from upload page work properly
- single updates are fast

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/kahuna/public/js/util/async.js
+++ b/kahuna/public/js/util/async.js
@@ -56,7 +56,7 @@ const queue = new PQueue({ concurrency: 10 });
 async.factory("apiPoll", [
   () => {
     const wait = () => new Promise(resolve => {
-      setTimeout(() => resolve(), 4000);
+      setTimeout(() => resolve(), 500);
     });
     const poll = async (func, n) => {
       const [{ status, value }] = await Promise.allSettled([
--- a/kahuna/public/js/util/batch-tracking.js
+++ b/kahuna/public/js/util/batch-tracking.js
@@ -1,4 +1,5 @@
 import PQueue from "p-queue";
+import { idleTimeout } from "./idleTimeout";
 
 const concurrency = 30;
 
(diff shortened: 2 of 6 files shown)


kahuna/public/js/util/batch-tracking.js
```diff
@@ -58,5 +59,6 @@ export const trackAll = async ($rootScope, key, input, tasks, emit) => {
   $rootScope.$broadcast("events:batch-operations:complete", { key });
 
   $rootScope.$emit(emit, successes);
+  idleTimeout(() => { $rootScope.$apply(); });
```
jonathonherbert, line 62: Oh wow, so the recipients of the events are not triggering dispatch at the end of a batch? Interesting, good to see this. Might be worth a comment to indicate why this is necessary?

kahuna/public/js/util/async.js
```diff
@@ -56,7 +56,7 @@ const queue = new PQueue({ concurrency: 10 });
 async.factory("apiPoll", [
   () => {
     const wait = () => new Promise(resolve => {
-      setTimeout(() => resolve(), 4000);
+      setTimeout(() => resolve(), 500);
```
jonathonherbert, line 59: Just to double check – does this mean (4000/500)/s more polls per image in batch until timeout (my read)? Or does this only happen once? Probably this isn't too problematic: my understanding is that we only poll for a limited number of images at once, and our API is pretty robust, but I want to check that a) my understanding is correct and that b) we've thought about the impact on the service.

AWare: Yes it does, this was the old poll rate, and now we know that polling isn't the weak link, we can go back to it.

kahuna/public/js/util/batch-tracking.js
AWare: The reason is angular.

jonathonherbert (approved):
Tested in CODE, single and batch edits work correctly (addition and removal) and are nice and snappy 👍 

We discussed tests, which are difficult to write at the moment, but it does feel like the sooner we get some kind of CI integration tests for the Grid, the sooner we're able to ensure improvements like this cannot regress.

PR #3094, opened 2021-01-13
Feature/bbc authentication provider

adrielulanovsky:
## What does this change?
This PR implements a `UserAuthenticationProvider` for the BBC, to be able to have a BBC specific authentication. It is based off of https://github.com/guardian/grid/pull/3058.

There's also been a change in the `setup.sh` script so that it checks in an environment variable if the organization is BBC, and then modifies configuration files to reflect that.

BBC grid authenticates successfully with BBC Login, while not affecting Guardian's authentication process.

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -69,7 +69,7 @@ lazy val commonLib = project("common-lib").settings(
     "com.typesafe.play" %% "filters-helpers" % "2.6.20",
     akkaHttpServer,
     ws,
-    "com.gu" %% "pan-domain-auth-play_2-6" % "0.8.2",
+    "com.gu" %% "pan-domain-auth-play_2-6" % "0.9.2-SNAPSHOT",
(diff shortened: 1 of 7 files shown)


sihil:
This is currently dangerous for us to merge as it will break the build and change the default behaviour that we rely on. Marking this as draft so that the conversation can continue.

adrielulanovsky:
> This is currently dangerous for us to merge as it will break the build and change the default behaviour that we rely on. Marking this as draft so that the conversation can continue.

Yeah sorry for that Simon, I wasn't intending to actually merge this, just to show current progress on this task

sihil:
Hi @adrielulanovsky - looks like GitHub helpfully closed this when I merged (and auto-deleted) the branch this PR was set against. I think you'll have to create a new PR by the looks of things as you don't seem to be able to change the base branch in this state. Apologies for the inconvenience.

PR #3095, opened 2021-01-14
[BBC] Introducing search by file type (tiff, png, jpeg)

gribeiro:
## What does this change?

This change introduces a new "advanced search" filter for searching by file type, named "fileType". 
The user can input some file formats like "jpg", "tif" (or "tiff"), "png" and the grid will search for the respective MIME type of the source image, i.e., "image/jpeg".

This change was introduced as part of https://github.com/bbc/grid/pull/55.

Edit: As per @paperboyo suggestion, we have added the offering of suggestions:

fileTypeSuggestions

Upload file of types jpg, tiff, png and try to use the new filter: "fileType".

searchByFileType

## Tested?
- [x] locally
- [x] on TEST


media-api/app/lib/querysyntax/QuerySyntax.scala
```diff
@@ -178,6 +191,20 @@ class QuerySyntax(val input: ParserInput) extends Parser with ImageFields {
     })
   }
 
+  def MatchMimeTypeValue = rule {
+    (QuotedString | String) ~> parseMimeType _
+  }
+
+  def parseMimeType(expr: String): Value = {
```
akash1810, line 198: Could we make use of the [`MimeType` trait](https://github.com/guardian/grid/blob/e7d1d7f9f8263039c831b3c0c9255c6f0d318952/common-lib/src/main/scala/com/gu/mediaservice/model/MimeType.scala#L8) here to increase type safety?

gribeiro: Great idea. I also implemented a rule for just accepting the formats.

paperboyo:
Great! From a user POV, since there is only a handful of fileformats supported, is there a way for it to offer suggestions like some other search chips?

image

gribeiro:
Hello @paperboyo.
Nice suggestion. Just updated the PR with it. 😄

akash1810:
This is looking great! Could we add a test too? [This PR](https://github.com/guardian/grid/pull/2190/files) may provide some inspiration.

gribeiro:
Hello, @akash1810.

I've updated with the tests.
The build has some errors, but I think this is unrelated to my last commit. I had similar problem with my previous PR, Simon just ran it again and it works, maybe it is the same case?
If the build is ok, I'm good on the merging.

Thank you.

AWare (approved):
Works for me on TEST and looks fine to me 🚢

akash1810 (approved):

PR #3097, opened 2021-01-14
Put Photoshoot editing behind edit_metadata permission

paperboyo:
As all other metadata (but Labels and Collections), Photoshoot editing is now behind the usual permission. It looks like an oversight on our part. There may be an argument to be made that more of non-crucial metadata shouldn’t be permissioned, but that’s a different discussion.

We will quickly know if anyone depends on this loophole.

This makes metadata editing rules more consistent and predictable.

## Tested?
- [ ] locally
- [ ] on TEST

### DIFF
--- a/kahuna/public/js/components/gr-photoshoot/gr-photoshoot.html
+++ b/kahuna/public/js/components/gr-photoshoot/gr-photoshoot.html
@@ -1,6 +1,7 @@
 <section ng-if="! ctrl.editInline">
     <button data-cy="it-photoshoot-edit-button"
             class="image-info__edit"
+            ng-if="ctrl.userCanEdit"
             ng-click="photoshootEditForm.$show()"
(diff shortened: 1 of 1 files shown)


paperboyo:
Maybe not. Turns out, this is more broken than just being editable when one has no permission. All other fields but this (and recently introduced People) do not display at all when they are empty **and** user cannot edit them. So both Photoshoot and People need to be fixed to disappear under those circumstances completely (eg. currently People invites you to `click to edit` even if you can’t). But simulating Permissions locally requires Janus access I don’t have, plus I don’t know how to pretend I haven’t uploaded a picture to LOCAL (both need to be true for `!ctrl.userCanEdit` to be true).
But really, what I’m saying is that I feel embarrassed to have to deploy to TEST every time I stick random conditions in random places. What I learned is that there are more of them that I don’t know where to put than I initially thought. :-(

PR #3101, opened 2021-01-15
various angular fixes

AWare:
@jonathonherbert this is my branch from earlier. going to keep on it on monday but mostly just tidy up and rationalise my changes to the lease service.

ps @paperboyo this adds a counter when uploading, is this useful?

## What does this change?

- Wraps `trackAll` and `apiPoll` in $q to make angular happier.
- Makes leases play better with the digest loop. 
- Changes to lease batching on upload
	- Batches the sending of update events in the batch handler to prevent quadratics. 
	- This is done with rx because we don't have lodash, I'm very sorry.
	- Adds an updating state to the lease to make it obvious that a batch update is happening. 
- Lease updates already handled by lease updated event handler refreshing leases¹ so trying to make sure this is the _way_ for leases².
	1. On upload this means every lease for every lease, yikes.
	2. Image updates are mostly handled by the images-updated event, I think this too should be the way, and we should delete the comments disparaging this and embrace as a single way of doing angular updates as it appears to work much more reliably than just randomly mutating scope variables in controllers- which appears to be what we wanted to do.
	- Maybe this event should be specific for which lease is updated.
	- It isn't.
- A progress counter for image uploads, because testing this locally without one was tedious. 
- Running `image.get()` before all lease updates because we only check that there's a more recent lease on the server than we currently have, and this is all kinds of race condition- which causes the UI to display semi-stale¹ update results.
	1. Stale enough to make breadcrumbs from, too stale to toast.
	-  Feeding this value through the batch tracker would have been a _little_ gnarly, so it's in a parent promise instead. 
- Replacing `moment` with the Date constructor in the date comparison, as it's not needed and seemed to be giving `moment(date).isAfter(moment(date)) == true` which was not helping the stale race condition.
- TrackAll can now emit multiple events, maybe this is not useful as we actually want to emit different event data for images and leases.
- Loads of new console.log statements 
- Loads of whitespace changes that vscode just randomly adds whenever I make a change. 

Leases can be reliably updated on the upload screen.

image

## Who should look at this?
@guardian/digital-cms 

## Tested?
- [X] locally
- [X] on TEST (thank you @paperboyo)

### DIFF
--- a/kahuna/public/js/leases/leases.js
+++ b/kahuna/public/js/leases/leases.js
@@ -87,8 +87,10 @@ leases.controller('LeasesCtrl', [
                     : "This image already has a syndication lease. It will be overwritten. " +
                       "Do you wish to proceed?";
                   const shouldApplyLeases = $window.confirm(confirmText);
-                  if (!shouldApplyLeases) {
-                    return;
+                    if (!shouldApplyLeases) {
+                      //We're not saving it if the user cancels, so clear up
+                      ctrl.adding = false;
+                      return;
                   }
                 }
 
@@ -108,10 +110,25 @@ leases.controller('LeasesCtrl', [
         const batchRemoveLeasesEvent = 'events:batch-apply:remove-leases';
 
         if (Boolean(ctrl.withBatch)) {
-            $scope.$on(batchAddLeasesEvent,
-                    (e, leases) => leaseService.replace(ctrl.images[0], leases));
+          $scope.$on(batchAddLeasesEvent,
+              (e, leases) => {
+                  ctrl.adding = true;
+                  leaseService.replace(ctrl.images[0], leases).then((a) => {
+                //These get run before the lease is shown in the UI. Fix.
+                    ctrl.adding = false;
+                });
+            });
             $scope.$on(batchRemoveLeasesEvent,
-                    () => leaseService.clear(ctrl.images[0]));
+                () => {
+                  ctrl.adding = true;
+
+                    leaseService.clear(ctrl.images[0]).then((a) => {
+                        //Shown before leases updated. Fix.
+                        ctrl.adding = false;
+
+
+                    });
+                });
 
             ctrl.batchApplyLeases = () => {
               if (ctrl.leases.leases.length > 0) {
--- a/kahuna/public/js/leases/leases.js
+++ b/kahuna/public/js/leases/leases.js
@@ -196,8 +196,14 @@ leases.controller('LeasesCtrl', [
             return leasedBy;
         };
 
-        ctrl.inactiveLeases = (leases) => leases.leases.filter((l) => !ctrl.isCurrent(l)).length;
-        ctrl.activeLeases = (leases) => leases.leases.filter((l) => ctrl.isCurrent(l)).length;
+        ctrl.inactiveLeases = (leases) => {
+            console.log(leases)
+            leases.leases.filter((l) => !ctrl.isCurrent(l)).length;
+        }
+        ctrl.activeLeases = (leases) => {
+            console.log(leases)
+            leases.leases.filter((l) => ctrl.isCurrent(l)).length;
+        }
 
         ctrl.resetLeaseForm = () => {
             ctrl.newLease = {
--- a/kahuna/public/js/leases/leases.js
+++ b/kahuna/public/js/leases/leases.js
@@ -46,16 +46,23 @@ leases.controller('LeasesCtrl', [
         ctrl.editing = false;
         ctrl.adding = false;
         ctrl.showCalendar = false;
+        ctrl.receivingBatch = false;
+        console.log("HEY");
 
         ctrl.midnightTomorrow = moment().add(1, 'days').startOf('day').toDate();
 
-        ctrl.cancel = () => ctrl.editing = false;
+        ctrl.cancel = () => {
+            ctrl.editing = false;
+            ctrl.updateLeases();
+        };
 
         ctrl.save = () => {
             if (!ctrl.accessDefined()) {
                 $window.alert('Please select an access type (Allow or Deny)');
             } else {
                 ctrl.adding = true;
+                ctrl.editing = false;
+
                 ctrl.newLease.createdAt = new Date();
                 ctrl.newLease.access = ctrl.access;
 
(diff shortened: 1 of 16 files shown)


AWare:
TODO: need to fix leases on the search page, but this should work on upload @paperboyo

kahuna/public/js/leases/leases.js
```diff
@@ -87,31 +94,56 @@ leases.controller('LeasesCtrl', [
                     : "This image already has a syndication lease. It will be overwritten. " +
                       "Do you wish to proceed?";
                   const shouldApplyLeases = $window.confirm(confirmText);
-                  if (!shouldApplyLeases) {
+                    if (!shouldApplyLeases) {
+                      //We're not saving it if the user cancels, so clear up
+                    //   return ctrl.updateLeases();
```
jonathonherbert, line 99: Still needed?

kahuna/public/js/leases/leases.js
```diff
@@ -87,31 +94,56 @@ leases.controller('LeasesCtrl', [
...
-                .finally(() => {
+                  .finally(() => {
+                      console.log("HELO");
                     ctrl.resetLeaseForm();
                 });
             }
         };
 
         // These events allow this control to work as a hybrid on the upload page
+        // Update from future: not very well.
```
jonathonherbert, line 120: Is this any better now?

kahuna/public/js/leases/leases.js
```diff
@@ -87,31 +94,56 @@ leases.controller('LeasesCtrl', [
...
         const batchRemoveLeasesEvent = 'events:batch-apply:remove-leases';
 
         if (Boolean(ctrl.withBatch)) {
-            $scope.$on(batchAddLeasesEvent,
-                    (e, leases) => leaseService.replace(ctrl.images[0], leases));
+          $scope.$on(batchAddLeasesEvent,
+              (e, leases) => {
+                  ctrl.receivingBatch = true;
+                  leaseService.replace(ctrl.images[0], leases).then((a) => {
+                //These get run before the lease is shown in the UI. Fix.
```
jonathonherbert, line 129: Is this fix due in this PR?

kahuna/public/js/leases/leases.js
```diff
@@ -87,31 +94,56 @@ leases.controller('LeasesCtrl', [
...
+                });
+            });
             $scope.$on(batchRemoveLeasesEvent,
-                    () => leaseService.clear(ctrl.images[0]));
+                () => {
+                    ctrl.receivingBatch = true;
+
+
+                    leaseService.clear(ctrl.images[0]).then((a) => {
+                        //Shown before leases updated. Fix.
```
jonathonherbert, line 140: As above.

kahuna/public/js/leases/leases.js
AWare: It's still a massively dangerous pattern.

kahuna/public/js/leases/leases.js
AWare: Ooops, that comment doesn't belong any more. 

When `replace` is run, it triggers the `leasesUpdated` event which refreshes all the leases on the page, which means we can just wait for that. (Will update comment)

kahuna/public/js/leases/leases.js
AWare: (likewise)

kahuna/public/js/services/api/leases.js
```diff
@@ -5,87 +5,100 @@ import '../../services/image-list';
...
-      });
+            if (angular.isDefined(newLease.notes) && newLease.notes.trim().length === 0) {
+                newLease.notes = null;
+            }
+            return image.perform('add-lease', { body: newLease });
+        }
+
+        function batchAdd(lease, images) {
+            //We check whether the leases in the image have a later lastModified date,
+            //so make sure that we don't have anything in the pipeline.
```
jonathonherbert, line 87: Re: `make sure that we don't have anything in the pipeline.`, unsure how this maps to the logic below (what's making sure of this?)

kahuna/public/js/services/api/leases.js
```diff
@@ -110,20 +128,48 @@ leaseService.factory('leaseService', [
...
+          return results.map(({ image, leases }) => {
+              emitLeaseUpdate(leases);
+              console.log(images, leases);
+                emitImageUpdate(image);
+          });
       });
     }
 
+    const imageUpdates$ = new Subject();
+        imageUpdates$.bufferWithTime(1000).subscribe((images) => {
```
jonathonherbert, line 141: Oh cool, so we create a buffer, pump updates into it and issue updates periodically – basically a debounce?

kahuna/public/js/services/api/leases.js
```diff
@@ -110,20 +128,48 @@ leaseService.factory('leaseService', [
...
+            console.log("DUCK", leases, leases.length);
+            $rootScope.$emit('leases-updated');
+    });
+
+    function emitLeaseUpdate(lease) {
+        leaseUpdates$.onNext(lease);
+    }
+
+      //This is a race condition, but I think it's a rare one.
+        //Mitigate by refreshing images with refreshImages before posting.
```
jonathonherbert, line 164: Is this work done?

kahuna/public/js/services/api/leases.js
AWare: This relates more to `refreshImages(images).then` will move.

kahuna/public/js/services/api/leases.js
AWare: Yeah, basically, but we want to keep the id's about when we send the message.

kahuna/public/js/leases/leases.js
jonathonherbert: To rephrase 😁 does a developer in the future need this comment? If so, will they understand it as it stands?

kahuna/public/js/services/api/leases.js
AWare: Yeah, might not be the best wording but I intended that as a warning to anyone using `untilLeasesChange` in the future.

jonathonherbert: Ah, so `mitigated`?

kahuna/public/js/leases/leases.js
AWare: Reworded.

kahuna/public/js/leases/leases.js
AWare: Deleted

kahuna/public/js/leases/leases.js
AWare: Resolved

kahuna/public/js/services/api/leases.js
AWare: Moved.

kahuna/public/js/services/api/leases.js
AWare: Yup.

kahuna/public/js/leases/leases.js
AWare: Nope, removed

jonathonherbert (approved):
Tested in CODE, lease changes now seem reliable in downloads and side panel ✨

PR #3108, opened 2021-01-19
Adds watch to node children, to update if new collections are added

SHession:
## What does this change?
This PR fixes an issue where new sub-collections were not being displayed in the UI until the page was refreshed(see gif). This issue was due to nothing watching the node children for changes, meaning the sub-collections were not being updated. 

This [PR](#3074) introduced a change which would filter out collections with no data associated with them (causing all sibling collections to disappear). The children used to be passed directly to the "gr-nodes". The PR changed this to use a filtered collection, however nothing was watching the actual collection for changes. Therefore, new collections were not reflected in the UI. A new watch on the actual children is now being used to check for changes and update the UI.

unnamed

Adding a new sub-collection appears in the UI immediately. 

## Tested?
- [x] locally
- [ ] on TEST

### DIFF
--- a/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
+++ b/kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
@@ -123,8 +123,12 @@ grCollectionsPanel.controller('GrNodeCtrl',
     ctrl.getCollectionQuery = path => getCollection(path);
 
     $scope.$watch('ctrl.showChildren', onValChange(show => {
-            collectionsTreeState.setState(pathId, show);
+      collectionsTreeState.setState(pathId, show);
     }));
+
+    $scope.$watch('ctrl.node.data.children', children => {
+      ctrl.children = children.filter(node => !!node.data.data);
+    });
 
     ctrl.init = function(grCollectionTreeCtrl) {
         const selectedImages$ = grCollectionTreeCtrl.selectedImages$;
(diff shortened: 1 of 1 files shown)


kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -123,9 +123,13 @@ grCollectionsPanel.controller('GrNodeCtrl',
     ctrl.getCollectionQuery = path => getCollection(path);
 
     $scope.$watch('ctrl.showChildren', onValChange(show => {
-            collectionsTreeState.setState(pathId, show);
+      collectionsTreeState.setState(pathId, show);
     }));
 
+    $scope.$watch('ctrl.node.data.children', children => {
+      ctrl.children = children.filter(node => !!node.data.data);
```
AWare, line 130: I think this filter should be a function.

SHession: Fair point, any ideas where it should live? On the controller or in some util file somewhere?

AWare: Yeah, just in the controller to make it obvious what it's doing.

kahuna/public/js/components/gr-collections-panel/gr-collections-panel.js
```diff
@@ -103,7 +103,12 @@ grCollectionsPanel.controller('GrNodeCtrl',
     const pathId = ctrl.node.data.data.pathId;
 
     //This filter remove child nodes with missing data, preventing display errors from occurring
-    ctrl.children = ctrl.node.data.children.filter(node => !!node.data.data);
+    $scope.filterChildren = children => children.filter(node => !!node.data.data);
```
AWare, line 106:
```suggestion
    ctrl.filterChildren = children => children.filter(node => !!node.data.data);
```

AWare: sorry to nit, but I don't think we want to put the function on scope
(do we?)

AWare (approved):

PR #3113, opened 2021-01-19
Downgrade test library to fix JSON ordering

sihil:
## What does this change?
This fixes the JSON re-ordering side effect introduced in #3058.

The play-json library mangles the order of fields from `2.6.11` until `2.8.0`. See https://github.com/playframework/play-json/issues/236. This solves the issue by downgrading the scalatest library that puled in the later version. At some stage we'll upgrade Play and will need to jump to 2.8 to avoid this behaviour.

There is also a test to detect any future inadvertent upgrade of the play-json library.

## Contention
Some people disagreed with the need for this, but it does make manual viewing of the API significantly easier. There are also now tests upstream so this shouldn't break again.

## Screenshots

Before:
Screenshot 2021-01-19 at 17 34 01

After (ordering maintained in API):
image

## Tested?
- [X] locally by committer
- [X] locally by Guardian reviewer

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -18,7 +18,7 @@ val commonSettings = Seq(
 
   testOptions in Test ++= Seq(Tests.Argument(TestFrameworks.ScalaTest, "-o"), Tests.Argument(TestFrameworks.ScalaTest, "-u", "logs/test-reports")),
   libraryDependencies ++= Seq(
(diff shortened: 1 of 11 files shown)


paperboyo (approved):
If this doesn’t have adverse effects, yes we prefer ordered json API response. Hopefully, this won’t get us stuck on some outdated version of anything.

AWare (approved):

PR #3114, opened 2021-01-20
Apipoll error

tjsilver:
## What does this change?

If apiPoll tries too many times (in this case, 100) it throws an error. In the case of batch uploads, this error appears under the image on the upload page.

Further work to show an error modal to the user when other batch processes fail, listing the failed images, but this is for a future PR.

The user sees an error when apiPoll is taking too long to upload an image.

## Screenshots

image

## Who should look at this?
@guardian/digital-cms

## Tested?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/util/async.js
+++ b/kahuna/public/js/util/async.js
@@ -70,7 +70,7 @@ async.factory("apiPoll", [
       }
       // Something has gone wrong, so we can let the user know
       if (n > 100) {
-        throw new Error('apiPoll failed - Please try again.');
+        throw new Error('gave up after 100 tries (apiPoll failed)');
       }
       await wait();
       return poll(func, n + 1);
--- a/kahuna/public/js/util/async.js
+++ b/kahuna/public/js/util/async.js
@@ -70,7 +70,7 @@ async.factory("apiPoll", [
       }
       // Something has gone wrong, so we can let the user know
       if (n > 100) {
-        throw new Error('failed');
+        throw new Error('apiPoll failed - Please try again.');
       }
       await wait();
       return poll(func, n + 1);
--- a/kahuna/public/js/util/async.js
+++ b/kahuna/public/js/util/async.js
@@ -68,6 +68,10 @@ async.factory("apiPoll", [
       if (status === 'fulfilled') {
         return $q.resolve(value);
       }
+      // Something has gone wrong, so we can let the user know
+      if (n > 100) {
+        throw new Error('failed');
+      }
       await wait();
       return poll(func, n + 1);
     };


tjsilver:
[Add user error message when image updates fail](https://trello.com/c/2v2W3nS3/2138-add-user-error-message-when-image-updates-fail)

jonathonherbert (commented):
Change is nice and simple! Would it be worth making this more understandable to the user – I don't think they'll know what apiPoll means, but they might understand `gave up after 100 tries` or similar (worth chatting to Mariana or @paperboyo?)

tjsilver:
> Change is nice and simple! Would it be worth making this more understandable to the user – I don't think they'll know what apiPoll means, but they might understand `gave up after 100 tries` or similar (worth chatting to Mariana or @paperboyo?)

Good idea, will do.

paperboyo:
My 2p. I know it’s Grid’s way to know something’s finished when it’s finally answered “it’s finished” after asking if it finished constantly. So first one of my reservations may have no resolution, but I still wanted to express it:
1. would that error **ever** show **before** an operation would actually completed successfully (albeit – glacially)? Let’s say, I will make a mistake of batch editing 2k of pics (don’t try at home!), wait an era, see an error, go for 2 weeks holidays and when back I see this operation that errored actually finished OK?
2. If all we know about the nature of the error is that we haven’t heard an answer after an arbitrary number of questions posed, I think it’s OK to have the first part just say `[operation] failed:` and the second could well say `API poll timeout` or just `timeout`. The second part would be a user report quick glance debug worth only.

I think pt. 1 is actually important. Are we not only setting an arbitrary number of questions, but **also** quit asking afterwards (unlike now)? It may not matter practically, but this doesn’t sound like an error message to me. More like an information that stuff takes too long (which I most probably already know: I’m on holidays ;-). But if we also quit…

Makes sense? I’m not sure, TBH.

kahuna/public/js/util/async.js
```diff
@@ -68,6 +68,10 @@ async.factory("apiPoll", [
       if (status === 'fulfilled') {
         return $q.resolve(value);
       }
+      // Something has gone wrong, so we can let the user know
+      if (n > 100) {
+        throw new Error('apiPoll failed - Please try again.');
```
akash1810, line 73: Does this cancel the promise?

AWare: The promise has already been settled at this point.

kahuna/public/js/util/async.js
```diff
@@ -68,6 +68,10 @@ async.factory("apiPoll", [
       if (status === 'fulfilled') {
         return $q.resolve(value);
       }
+      // Something has gone wrong, so we can let the user know
+      if (n > 100) {
+        throw new Error('gave up after 100 tries (apiPoll failed)');
```
tjsilver, line 73: This is a compromise text with the user-friendly text followed by some technical info for debugging.

jonathonherbert (approved):
👍  (@paperboyo, were your comments above addressed? Re: pt.1, yes, I think the op has failed at this point and will never succeed)

paperboyo:
@jonathonherbert If the operation won’t ever succeed, then I’m happy! Messaging all good. As mentioned, will there be logging firing for when this shows up? (if you think that’s useful!)

tjsilver:
> @jonathonherbert If the operation won’t ever succeed, then I’m happy! Messaging all good. As mentioned, will there be logging firing for when this shows up? (if you think that’s useful!)

I'll create a ticket for the logging as a separate piece of work.

PR #3115, opened 2021-01-20
isInferred may not be in document

AWare:
## What does this change?

If isInferred isn't present in the elasticsearch document, which is the case for some old images. 
(See thread for example) then parsing it fails. 

We used to parse it implicitly until #3088 where we inferred false, but #2299 would imply that we should do otherwise. 

## Who should look at this?

Afaict we don't have many invalid json tests, this should go with them.
Also, I'm not sure if this is the correct fix- but it should provide similar behaivour to before #3088 

## Tested?
- [ ] locally by committer
- [X] not really possible (we could really do with something that pulled whole grid documents into local by id)


AWare:
I'd also be tempted to make https://github.com/guardian/grid/blob/9953390b07a4c117da13bf9ce45745a767c77628/kahuna/public/js/search/results.js#L175 log an error into sentry, to give us more visibility over errors like this.

akash1810 (approved):

PR #3116, opened 2021-01-20
Feature/bbc authentication provider

adrielulanovsky:
## What does this change?
This PR implements a UserAuthenticationProvider for the BBC, to be able to have a BBC specific authentication. For now, it is mostly based on PandaAuthenticationProvider, the main difference being a different implementation for validateUser, that either checks a whitelist of valid emails, or checks the user has the correct permissions.

The BBCAuthenticationProvider lives in a new subproject specific to the BBC, which will be compiled only if the BUILD_ORG environment variable is set to "bbc". This way, we avoid compilation errors caused by shared code but different versions of pan-domain-auth.

BBC grid authenticates successfully with BBC Login, while not affecting Guardian's authentication process.

## Who should look at this?
@sihil , @wainaina , @gribeiro 

## Tested?
- [x] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/bbc/components/BBCValidEmailsStore.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/bbc/components/BBCValidEmailsStore.scala
@@ -1,0 +1,56 @@
+package com.gu.mediaservice.lib.bbc.components
+
+import com.gu.mediaservice.lib.BaseStore
+import com.gu.mediaservice.lib.config.CommonConfig
+import org.joda.time.DateTime
+import play.api.Logger
+import play.api.libs.json.Json
+
+import scala.concurrent.ExecutionContext
+import scala.util.{Failure, Success, Try}
+
+class BBCValidEmailsStore(bucket: String, config: CommonConfig)(implicit ec: ExecutionContext)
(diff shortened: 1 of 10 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/bbc/components/BBCValidEmailsStore.scala
```diff
@@ -0,0 +1,56 @@
...
+import com.gu.mediaservice.lib.BaseStore
+import com.gu.mediaservice.lib.config.CommonConfig
+import org.joda.time.DateTime
+import play.api.Logger
+import play.api.libs.json.Json
+
+import scala.concurrent.ExecutionContext
+import scala.util.{Failure, Success, Try}
+
+class BBCValidEmailsStore(bucket: String, config: CommonConfig)(implicit ec: ExecutionContext)
```
sihil, line 12: Can this new class be moved to the new sub project alongside the auth provider?

sihil (commented):
This looks sane to me. I reckon that the Guardian will also create a sub project in much the same way for code that is specific to us. Yhe valid email store should probably be in the sub project too. 

Long term I'd expect all BBC code to be in this sub project, but leaving that to another PR makes sense to me.

PR #3117, opened 2021-01-21
Add deserialise tests for syndication rights

aug24:
## What does this change?
Only tests.

We recently had a failure when syndication rights deserialisation was altered to parse the date, which accidentally removed the defaulting from `isInferred`.  This was corrected in an earlier PR.  This PR represents the 'missing test' to prevent this error.

## Tested?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/test/scala/com/gu/mediaservice/model/SyndicationRightsTest.scala
+++ b/common-lib/src/test/scala/com/gu/mediaservice/model/SyndicationRightsTest.scala
@@ -1,0 +1,54 @@
+package com.gu.mediaservice.model
+
+import org.scalatest.{FunSpec, Matchers}
+import play.api.libs.json.Json
+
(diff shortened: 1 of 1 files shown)


AWare (approved):

common-lib/src/test/scala/com/gu/mediaservice/model/SyndicationRightsTest.scala
```diff
@@ -0,0 +1,54 @@
...
+      """
+        |{
+        |  "published": "2021-1-21T21:21:21.21Z",
+        |  "suppliers": [],
+        |  "rights": [],
+        |  "isInferred": false
+        |}
+        |""".stripMargin
+    val parsedRightsJson = Json.parse(serialisedRights)
+    println(parsedRightsJson)
```
akash1810, line 19: do we need the println?

PR #3118, opened 2021-01-21
Allow no last modified date on json UpdateMessages for Thrall

aug24:
## What does this change?
As the code creation of `UpdateMessage` will ensure a `lastModified` date is present, it was thought that it could be treated as a field which would always be present in the json.

However, it appears that there are sources of `UpdateMessage` which do not send a `lastModified` date (and so presumably do not use the Grid code to construct their json).

As a result, we have errors in logs stating the message could not be parsed as it does not have a lastModified date.

This change allows a message without a `lastModified` date to be deserialised, and provides a `lastModified` date of `Now` in UTC.

No more errors.

## Tested?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/ThrallMessageSender.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/ThrallMessageSender.scala
@@ -7,7 +7,8 @@ import com.gu.mediaservice.model.leases.MediaLease
 import com.gu.mediaservice.model.usage.UsageNotice
 import net.logstash.logback.marker.LogstashMarker
 import org.joda.time.{DateTime, DateTimeZone}
-import play.api.libs.json.{JodaReads, JodaWrites, Json}
+import play.api.libs.functional.syntax.toFunctionalBuilderOps
+import play.api.libs.json.{JodaReads, JodaWrites, Json, __}
 
 // TODO MRB: replace this with the simple Kinesis class once we migrate off SNS
 class ThrallMessageSender(config: KinesisSenderConfig) {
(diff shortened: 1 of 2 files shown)


AWare (approved):

AWare:
Would prefer to not use a hand written JSON parser, but alternatives feel "too clever".

paperboyo:
> This change allows a message without a `lastModified` date to be deserialised, and provides a `lastModified` date of `Now` in UTC.

Does it mean these (external systems, I assume) are now gonna start writing `lastModified`? Asking, coz this will make it further from human-useful `lastModified` (similar to a discussion around Composer’s one). One day, when users will (finally!) be able to eg. sort by it (they can already search), they would rather not care about some systems writing some stuff only about concious humans. Like themselves.

To that (and if that is true and makes sense): do you think we have to have `lastModifiedByHuman` concept? (goes for Composer and other systems too)

aug24:
No, it means the opposite: we don't care if they do, and will assume now() if they don't send one.

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/ThrallMessageSender.scala
```diff
@@ -34,7 +35,23 @@ object UpdateMessage {
...
   implicit val writes = Json.writes[UpdateMessage]
-  implicit val reads = Json.reads[UpdateMessage]
+  implicit val reads =
+    (
+      (__ \ "subject").read[String] ~
+        (__ \ "image").readNullable[Image] ~
+        (__ \ "id").readNullable[String] ~
+        (__ \ "usageNotice").readNullable[UsageNotice] ~
+        (__ \ "edits").readNullable[Edits] ~
+        // We seem to get messages from _somewhere which don't have last modified on them.
```
akash1810, line 45: Can we log the `subject` of the messages missing `lastModified` to narrow this down and fix?

Issue #3100, opened 2021-01-15
Search on the value of arbitrary fields


gribeiro:
## Description

In the light of the issue https://github.com/guardian/grid/issues/3084, it was discussed that we need the search on the value for arbitrary fields.
This search initially would search any field for any content. 
For this we need to start a discussion on the best way to implement it.

Some key points:
1. Some way for the user input two strings, a field name and a field content.
2. Perhaps some new UI element should be done in order to accommodate this new search?
3. We may need limit the scope of this search for some group of fields.
3. We need a consensus name for this search.
4. A discussion regarding on best way on how to do it in Grid must take place, including the changes on the QuerySyntax and the building of the Query.

## Draft
For the purposes of starting a discussion, I have started a (working) draft solution and I have taken the liberty to give some placeholder solutions and naming for some of the key points mentioned above. Please suggest some better naming and/or syntax that would reflect in a better usage of the Grid.

This initial solution revolves around creating a new rule on the QuerySyntax.scala. This rule let us do the aforementioned search with the syntax `+hasWith:some.field=value` or with quotes `+hasWith:some.field="value"` and the QueryBuilder proceeds with adding the Query: `matchQuery(getFieldPath(field), value).operator(Operator.AND)`

This draft solution can be found here: https://github.com/guardian/grid/compare/main...bbc:feature/has-search-value

## Screenshots

Example of a working search:
![arbitrarySearch1](https://user-images.githubusercontent.com/128418/104691310-bf7b6780-56e4-11eb-99cc-f09d0b4b3450.PNG)
The inspect of this object:
![arbitrarySearch2](https://user-images.githubusercontent.com/128418/104691315-c30eee80-56e4-11eb-8bfa-3d3199a07f14.PNG)

Example working on **fileMetadata**:
![fileMetadataSearchOk](https://user-images.githubusercontent.com/128418/104797810-c1eec780-579f-11eb-9925-02f9de2d1cfa.PNG)

sihil:
Hi @gribeiro,

Apologies for the delay on responding to this.

We've had some discussions about this. The historical convention for searching (influenced by lucene search syntax) is `<fieldName>:<fieldValue>` (so no need for `hasWith` or an alternative). We don't use `=` elsewhere and would be keen to retain consistency.

[...]

PR #3119, opened 2021-01-21
Allow even more date formats

aug24:
## What does this change?
We have found yet another form of date which is sent in exif data: `yyyyMMdd`.

This is added to the list of tolerated date formats (in the right place).  At the same time, a few other likely candidates were added, plus tests for them.

Fewer errors on load.

## Tested?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
@@ -126,6 +126,9 @@ object ImageMetadataConverter extends GridLogging {
     DateTimeFormat.forPattern("E MMM dd HH:mm:ss.SSS 'BST' yyyy").withZone(DateTimeZone.forOffsetHours(1)),
     DateTimeFormat.forPattern("E MMM dd HH:mm:ss 'BST' yyyy").withZone(DateTimeZone.forOffsetHours(1)),
 
+    DateTimeFormat.forPattern("yyyyMMdd"),
+    DateTimeFormat.forPattern("yyyyMM"),
+    DateTimeFormat.forPattern("yyyyddMM"),
     DateTimeFormat.forPattern("yyyy"),
     DateTimeFormat.forPattern("yyyy-MM"),
 
--- a/common-lib/src/test/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverterTest.scala
+++ b/common-lib/src/test/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverterTest.scala
@@ -475,8 +475,4 @@ class ImageMetadataConverterTest extends FunSpec with Matchers {
     ImageMetadataConverter.parseRandomDate("2000-02-31") should be(None)
   }
 
-  it("whatever") {
-    DateTimeFormat.forPattern("yyyyMMdd").parseDateTime("20010203").withZone(DateTimeZone.UTC) should be (day(2001, 2, 3))
-  }
-
 }
--- a/common-lib/src/test/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverterTest.scala
+++ b/common-lib/src/test/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverterTest.scala
@@ -420,58 +420,58 @@ class ImageMetadataConverterTest extends FunSpec with Matchers {
       .withSecondOfMinute(s)
       .withMillisOfSecond(ss)
 
-  it("should cope with full date formats and return None otherwise ") {
+  it("should cope with full date formats") {
     ImageMetadataConverter.parseRandomDate("2001-02-03T04:05:06.007Z") should be(Some(day(2001, 2, 3, 4, 5, 6, 7)))
   }
-  it("should cope with offset with space date formats and return None otherwise") {
+  it("should cope with offset with space date formats") {
     ImageMetadataConverter.parseRandomDate("2001-02-03T04:05:06 +00:00") should be (Some(day(2001, 2, 3, 4, 5, 6)))
   }
-  it("should cope with offset without space date formats and return None otherwise") {
+  it("should cope with offset without space date formats") {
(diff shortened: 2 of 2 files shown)


paperboyo (approved):
This will help not only recognise these (illegal by spec!) dates, but will also stop failing the ingestion of images carrying them too. This latter point should be guarded against separately, obviously. And then, the failed images in s3 (which do not exist in ES) reingested. Internal [ticket here](https://trello.com/c/qwohLB8Y/2158-fail-to-index-stores-s3-objects-missing-from-es-beacuse-of-a-daft-date).

common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
```diff
@@ -126,6 +126,9 @@ object ImageMetadataConverter extends GridLogging {
     DateTimeFormat.forPattern("E MMM dd HH:mm:ss.SSS 'BST' yyyy").withZone(DateTimeZone.forOffsetHours(1)),
     DateTimeFormat.forPattern("E MMM dd HH:mm:ss 'BST' yyyy").withZone(DateTimeZone.forOffsetHours(1)),
 
+    DateTimeFormat.forPattern("yyyyMMdd"),
+    DateTimeFormat.forPattern("yyyyMM"),
+    DateTimeFormat.forPattern("yyyyddMM"),
```
AWare, line 131: I'm not sure whether we want to try both of these unless we've seen them, as they'll overlap for some dates- and we'll need to thandle that.

paperboyo: I thought they are tried from the top. So we are (correctly) Europe-centric (and ambigous ones are `yyyMMdd`-first). In case that’s not true, you are right.
Currently, there is no way to tell what is in the wild in those (any) fields. But after glorious [this](https://github.com/guardian/grid/issues/3100) is in (and we make it accept wildcards), we will be able to (finally) tell.
Date parsing logic could be much smarter than currently (comparing, looking for oldest and most detailed), but this is a good step regardless.

aug24: They are indeed tried from the top, fail-down.  So that's OK, we handle that by simple ordering, and there are tests to prove it too.

aug24: NB the yyyyMM has to come before yyyy for precisely this reason.  Any corner cases which are suggested will be added to tests...??

AWare: I'm not super happy about American dates of this format silently changing for the first 12 days of the month and then are parsed correctly. I'd much rather have them silently succeed for 12 days and fail the rest of the month until we're sure they exist; at which point we can figure out how to distinguish them.

paperboyo: Just to note: these dates were observed only as `yyyyMMdd` in both `photoshop:DateCreated` and `iptc:Date Created` (which is [mixed by metadata-extractor](https://github.com/guardian/grid/blob/ddb9effd49fa39bf63bd7e3bcd0fad3af78a19f1/image-loader/app/lib/imaging/FileMetadataReader.scala#L97-L105) into `Date Time Created Composite`. They are **illegal** values for both of those fields. We are doing them a favour by not dropping them on the floor. Instead of trying to figure out if this is silly American or correct European format, which may not be ultimately possible, we could improve other areas of date parsing (starting with not failing to ingest or storing in s3, but not in ES). Or, might I say, starting with virtually anything else ;-)

sihil: Can we set up an auto feedback loop that emails suppliers when they send us images with illegal dates?

paperboyo: > Can we set up an auto feedback loop that emails suppliers when they send us images with illegal dates?

Without much exaggeration, I’m afraid, for most big agencies, it would be an email for every image, if we would to do it. 😜 And the examples of these came via `email` route, so we would need to build tech that preserves the sender in metadata.

Look, this isn’t ideal and I could definitely be less facetious about the American date format. We can remove it from the above subprime position, but then we would need to somehow deal with it when it comes in anyway, right? Sniffing possible formatting sounds near impossible.

And all of it is just more code, for the examples that we haven’t encountered, while the ones that we have are trickling into the Grid as I type this too long comment **and they literally end up in s3 only, without any record in ES** generating upload errors. I would rather spend any effort into looking at why we fail to ingest an image, because some metadata fields has some illegal value. Illegal or poor metadata is, sadly, part of life, as our cleaner well illustrate.

AWare, line 131:
```suggestion
```

paperboyo: ^ as I commented above, I’m not against. But what does this change mean for a “date” that is `20202201`? (pretty sure it errors and errors badly: s3-only, no ES record!)

AWare: That issue would be better fixed by dropping the field instead of the image.

paperboyo: > That issue would be better fixed by dropping the field instead of the image.

Goes without saying! And without me saying that it goes without saying! :-) But! Your change means we will drop them, so the question is: would we rather pretend there is no date or have a date wrong by half a year?

twrichards (approved):
LGTM - based on the code, not tested it myself but based on the lovely tests 👏 ... I reckon it's fine

PR #3120, opened 2021-01-21
Don't treat missing images as errors in some cases (or missing messages) 

AWare:
## What does this change?
We have a real-world possibility showing up as a result of our integration testing:

 * Create image
 * Add something metadata-ish
 * Delete something metadata-ish (actually merely a message goes on to the low-priority queue)
 * Delete image
 * Attempt to do the delete something metadata-ish off the queue - *Oh Noes, no such image.*

This PR takes the following philosophical position:

A "delete" is not a task to be fulfilled, more of a "make sure this does not exist".

If the entire image does not exist, then by definition, the metadata we wish to ensure does not exist also does not exist.

This _is_ unexpected, but it is not an error.  We should therefore log at WARN, not ERROR 

Back to the integration tests... when those tests run, this happens, every time.  And we run those tests a lot.  
In fact, so much that it makes the thrall logs full of errors.
*We do not want logs full of errors.* (Ideally, every server-side error logged should be recorded as a problem and investigated.)

Reduced noise from thrall logs.

## Who should look at this?
 @guardian/digital-cms 

## Tested?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchExecutions.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchExecutions.scala
@@ -9,7 +9,7 @@ trait ElasticSearchExecutions extends GridLogging {
 
   def client: ElasticClient
 
-  def executeAndLog[T, U](request: T, message: String)(implicit
+  def executeAndLog[T, U](request: T, message: String, notFoundSuccessful: Boolean = false)(implicit
                                                        functor: Functor[Future],
                                                        executor: Executor[Future],
                                                        handler: Handler[T, U],
@@ -24,6 +24,7 @@ trait ElasticSearchExecutions extends GridLogging {
         r.isSuccess match {
           case true => Success(r)
           case false => r.status match {
+            case 404 if notFoundSuccessful => Success(r)
             case 404 => Failure(ElasticNotFoundException)
             case _ => Failure(ElasticSearchException(r.error))
           }
@@ -55,6 +56,5 @@ trait ElasticSearchExecutions extends GridLogging {
     }
     result
   }
-
 
 }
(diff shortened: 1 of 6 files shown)


AWare:
After discussion with @aug24 we're not going to discriminate on priority. 

Also, taking lack of engagement to be assent for this change.

jonathonherbert:
_> taking lack of engagement to be assent for this change._

This might be said in jest, but I'll be po-faced 😅 and say – I think this is the wrong assumption to make! Speaking personally, I haven't given an opinion here because I didn't understand the reason for this change from the PR description, it was under discussion elsewhere, and as a result I felt ill-qualified to intervene.

I've discussed w/ @aug24 and now better understand the reason for the change – might be worth adding that description be added to the PR for others to see.

Happy to approve once (IIUC!) the lane-specific changes are gone.

AWare:
By assent I just mean that noone really objects to me doing to work, so I'll finish it up.

AWare:
Not sure if `Boolean` is the right type here, considering a case object to make the call `executeAndLog(...,PermitMissingImage)`

thrall/app/lib/kinesis/ThrallEventConsumer.scala
```diff
@@ -47,7 +47,7 @@ class ThrallEventConsumer(es: ElasticSearch,
         Some(updateMessage)
       }
       case Success(None)=> {
-        logger.error(s"No message present in record at $timestamp")
+        logger.warn(s"No message present in record at $timestamp")
```
aug24, line 50:
```suggestion
        logger.warn(s"No message present in record at $timestamp", new String(r))
```

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchExecutions.scala
```diff
@@ -24,6 +24,7 @@ trait ElasticSearchExecutions extends GridLogging {
         r.isSuccess match {
           case true => Success(r)
           case false => r.status match {
+            case 404 if notFoundSuccessful => Success(r)
```
aug24, line 27:
```suggestion
            case 404 if notFoundSuccessful => {
                   logger.warn(s"No image found for $message.")
                   Success(r)
            }
```

aug24 (approved):

jonathonherbert (approved):

PR #3121, opened 2021-01-22
Log messages with missing last modified date

aug24:
## What does this change?
After a recent change to date handling, we had a live issue with messages which did not have a last modified date.  The change had made this mandatory rather than optional.  A fix was applied to default in a value of `now()` if needed, but it would be good to know more about this unexpected error condition.

This PR only adds logging at `warn` level to enable further analysis.

## Tested?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/ThrallMessageSender.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/ThrallMessageSender.scala
@@ -1,13 +1,13 @@
 package com.gu.mediaservice.lib.aws
 
 import com.gu.mediaservice.lib.config.CommonConfig
-import com.gu.mediaservice.lib.logging.{LogMarker, MarkerMap}
+import com.gu.mediaservice.lib.logging.{GridLogging, LogMarker, MarkerMap}
 import com.gu.mediaservice.model._
 import com.gu.mediaservice.model.leases.MediaLease
 import com.gu.mediaservice.model.usage.UsageNotice
 import net.logstash.logback.marker.LogstashMarker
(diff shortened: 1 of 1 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/aws/ThrallMessageSender.scala
```diff
@@ -43,7 +43,13 @@ object UpdateMessage {
         (__ \ "usageNotice").readNullable[UsageNotice] ~
         (__ \ "edits").readNullable[Edits] ~
         // We seem to get messages from _somewhere which don't have last modified on them.
-        (__ \ "lastModified").readNullable[DateTime].map(_.getOrElse(DateTime.now(DateTimeZone.UTC))) ~
+        (__ \ "lastModified").readNullable[DateTime].map{ d => d match {
+          case Some(date) => date
+          case None => {
+            logger.warn("Message received without a last modified date", __.toJsonString)
```
AWare, line 49: Would be ecstatic if we could pull subject to its own log marker but

AWare (approved):

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/ThrallMessageSender.scala
aug24: Well, we sort of could, but I thought dumping the whole thing was probably useful.

akash1810 (approved):
Logging the entire message might result in a long log line, the `subject` and `id` _might_ be all you need. Maybe we can do this in a later PR, if we find the logs unusable.

LGTM.

PR #3122, opened 2021-01-22
Disallow "date taken" values in the future

aug24:
## What does this change?
In an attempt to be strict on our date parsing, we have worked out one possible way in which we could wrongly load a date:

If we get a pic today (22nd January 2021) with a date string of 20211201, it cannot have been taken in the future!
Therefore we can choose to parse as (eg) US (12th Jan 2021), not EU (1st Dec 2021).

So we can refuse the (apparently successful) EU parse result, which comes first in the list.

Future dates will not be accepted.

## Tested?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
@@ -52,10 +52,11 @@ object ImageMetadataConverter extends GridLogging {
       res
     }
 
+    val currentDateTime = Some(DateTime.now(DateTimeZone.UTC))  // We expect dateTaken to be before this date.
     ImageMetadata(
-      dateTaken           = (fileMetadata.exifSub.get("Date/Time Original Composite") flatMap parseRandomDate) orElse
-                            (fileMetadata.iptc.get("Date Time Created Composite") flatMap parseRandomDate) orElse
-                            (readXmpHeadStringProp("photoshop:DateCreated") flatMap parseRandomDate),
+      dateTaken           = (fileMetadata.exifSub.get("Date/Time Original Composite") flatMap (parseRandomDate(_, currentDateTime))) orElse
+                            (fileMetadata.iptc.get("Date Time Created Composite") flatMap (parseRandomDate(_, currentDateTime))) orElse
+                            (readXmpHeadStringProp("photoshop:DateCreated") flatMap (parseRandomDate(_, currentDateTime))),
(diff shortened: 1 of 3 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
```diff
@@ -52,10 +52,11 @@ object ImageMetadataConverter extends GridLogging {
       res
     }
 
+    val currentDateTime = Some(DateTime.now(DateTimeZone.UTC))  // We expect dateTaken to be before this date.
```
sihil, line 55: By always taking now - this is non-deterministic. If reingested in the future this could change. I think we should use the initial upload date (i.e. the timestamp on the S3 object if it exists). If it is later projected (in 2022 using your example) then the parsing won't change as it will parse it based on when it was first seen.

sihil (approved):
I don't like `parseRandomDate` as a name as it's not particularly descriptive. But you inherited it, so 🤷.

common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
```diff
@@ -134,11 +134,21 @@ object ImageMetadataConverter extends GridLogging {
...
+      case (None, formatter) => safeParsing(
+        formatter.parseDateTime(str) match {
+          // NB We refuse parse results which result in future dates, if a max date is provided.
+          // eg If we get a pic today (22nd January 2021) with a date string of 20211201 we can be pretty sure
+          // that it should be parsed as (eg) US (12th Jan 2021), not EU (1st Dec 2021).
+          // So we refuse the (apparently successful) EU parse result.
+          case d if maxDate.forall(d.isBefore(_)) => d
+          case d => throw new IllegalArgumentException(s"Date $d is after $maxDate")
+        }
+      )
```
sihil, line 149:
```suggestion
      // NB We refuse parse results which result in future dates, if a max date is provided.
      // eg If we get a pic today (22nd January 2021) with a date string of 20211201 we can be pretty sure
      // that it should be parsed as (eg) US (12th Jan 2021), not EU (1st Dec 2021).
      // So we refuse the (apparently successful) EU parse result.
      case (None, formatter) => safeParsing(formatter.parseDateTime(str)).filter(maxDate.isBefore(_))
```

aug24: Won't work, because maxDate is an Option[DateTime].  Have rewritten.

Issue #3123, opened 2021-01-25
Image Quarantine + Virus Scanning Proposal


Asmau1234:
**Short Description**
There is a need to quarantine images uploaded by a user, scan them for viruses before they are uploaded, processed, pushed to post-processing images buckets and made available for use by users on the Grid.

**Current Situation**
When images are uploaded onto the Grid by a user, they currently go through the ImageLoader service which is responsible for processing uploaded images. Once the uploaded file is found to be a valid image, it passes through the metadata extraction, image optimisation (if necessary), a thumbnail image is created and stored in the S3 thumbnail bucket and the original image to the original image bucket.

**Current Challenge**
There is currently no way to prevent virus infected images from getting uploaded, processed and made available for use to users on the Grid.

**Proposed Solution**

- Image gets uploaded onto the Grid by a user via Web App (Kahuna)
- Web App polls /uploadStatus endpoint on the ImageLoader service by image id to get the “upload status” of the image just uploaded in step 1

- _Scenario 1: upload.quarantine.enabled configuration on ImageLoader is set to false_

  - Image is ingested on ImageLoader via POST /images
  - ImageLoader creates and sets image uploadStatus on DynamoDB to pending
  - ImageLoader processes image (extract metadata and upload original image and post-processing images to S3)
  - ImageLoader updates the image uploadStatus on DynamoDB to completed
  - Web App receives uploadStatus completed response and begins to poll /images/:id endpoint on the MediaAPI service

- _Scenario 2: upload.quarantine.enabled configuration on ImageLoader is set to true and Image contains a virus_

  - Image is sent to the configured quarantine S3 bucket by ImageLoader
  - Virus scanner scans the image in quarantine S3 bucket and finds the image to be infected
  - A virus alert lambda is triggered and sends an update uploadStatus request to ImageLoader service /uploadStatus
  - ImageLoader updates the image uploadStatus on DynamoDB to failed
  - Web App receives uploadStatus failed response, stops polling /uploadStatus endpoint on ImageLoader service
  - Web App displays to the logged in user the image upload status with a description

- _Scenario 3: upload.quarantine.enabled configuration on ImageLoader is set to true and Image does not contain a virus_

  - Image is sent to the configured quarantine S3 bucket by ImageLoader
  - Virus scanner scans the image in quarantine S3 bucket and finds the image to be clean
  - The clean image is copied to the usual ingest S3 bucket
  - Ingest lambda is triggered which sends a request POST to /imports endpoint on ImageLoader
[...]

PR #3126, opened 2021-01-25
Implement image upload-status endpoint and changing upload status on load and import

abdelrahmansd:
## What does this change?

This adds a Dynamo DB Table which is used to store images upload status 
```
case class UploadStatusRecord(
                               id: String,
                               fileName: Option[String],
                               uploadedBy: String,
                               uploadTime: String,
                               identifiers: Option[String],
                               status: StatusType,
                               errorMessage: Option[String],
                               expires: Long
                             )
```
and two endpoints 
- `GET       /uploadStatus/:imageId` , Kahuna will be polling it to Get Image upload status
- `POST    /uploadStatus/:imageId` A virus alert lambda will send an update upload Status request to it

also, Update upload status for POST /images 
- in case quarantining isn't enabled

**When doing a curl request** to`POST https://loader.local.dev-gutools.co.uk/uploadStatus/:imageId`
with body
`{"uploadStatus":
{"id":"<imageId>", "uploadedBy":"<uploaded_by>", "fileName":"<file_name>","status":"<status>"}}`

**The expected response should be:**
`{"data":{"id":"<image_id>","uploadedBy":"<uploaded_by>","fileName":"<file_name>","status":"<status>"}}`

**When doing a curl request** to`GET https://loader.local.dev-gutools.co.uk/uploadStatus/:imageId`

**The expected response should be**
`{"data":{"id":"<image_id>","uploadedBy":"<uploaded_by>","fileName":"<file_name>","status":"<status>"}}`

This is an implementation to this proposal opened as an issue [here](https://github.com/guardian/grid/issues/3123).
## Tested?
- [x] locally by committer

## Pre-merge deployment checklist (for Guardian)
 - [x] Deploy CFN update to PROD (see https://github.com/guardian/editorial-tools-platform/pull/480)
 - [x] Add configuration for new table to PROD env configuration

### DIFF
--- a/dev/cloudformation/grid-dev-core.yml
+++ b/dev/cloudformation/grid-dev-core.yml
@@ -181,6 +181,7 @@ Resources:
         ReadCapacityUnits: 1
         WriteCapacityUnits: 1
 
+
   PreviewContentPollTable:
     Type: AWS::DynamoDB::Table
     Properties:
@@ -194,6 +195,23 @@ Resources:
       ProvisionedThroughput:
         ReadCapacityUnits: 1
         WriteCapacityUnits: 1
+
+  UploadStatusTable:
+    Type: AWS::DynamoDB::Table
+    Properties:
+      TableName: UploadStatusTable
+      AttributeDefinitions:
+      - AttributeName: id
+        AttributeType: S
+      KeySchema:
+      - AttributeName: id
+        KeyType: HASH
+      ProvisionedThroughput:
+        ReadCapacityUnits: 1
+        WriteCapacityUnits: 1
+      TimeToLiveSpecification:
+        AttributeName: expires
+        Enabled: true
 
 Outputs:
   ImageBucket:
@@ -234,3 +252,5 @@ Outputs:
     Value: !Ref 'LiveContentPollTable'
   PreviewContentPollTable:
     Value: !Ref 'PreviewContentPollTable'
+  UploadStatusTable:
+    Value: !Ref 'UploadStatusTable'
--- a/image-loader/app/controllers/ImageLoaderController.scala
+++ b/image-loader/app/controllers/ImageLoaderController.scala
@@ -2,29 +2,35 @@ package controllers
 
 import java.io.File
 import java.net.URI
-
 import com.drew.imaging.ImageProcessingException
 import com.gu.mediaservice.lib.argo.ArgoHelpers
 import com.gu.mediaservice.lib.argo.model.Link
 import com.gu.mediaservice.lib.auth._
+import com.gu.mediaservice.lib.formatting.printDateTime
 import com.gu.mediaservice.lib.logging.{FALLBACK, GridLogging, LogMarker, RequestLoggingContext}
 import com.gu.mediaservice.lib.{DateTimeUtils, ImageIngestOperations}
 import com.gu.mediaservice.model.UnsupportedMimeTypeException
-import lib._
+import com.gu.scanamo.error.ConditionNotMet
+import lib.FailureResponse.Response
+import lib.{FailureResponse, _}
 import lib.imaging.{NoSuchImageExistsInS3, UserImageLoaderException}
 import lib.storage.ImageLoaderStore
-import model.{Projector, Uploader, QuarantineUploader}
+import model.{Projector, QuarantineUploader, StatusType, UploadStatus, UploadStatusRecord, Uploader}
 import play.api.libs.json.{JsObject, Json}
 import play.api.libs.ws.WSClient
 import play.api.mvc._
 import model.upload.UploadRequest
+import org.joda.time.DateTime
+
+import java.time.Instant
 import scala.concurrent.{ExecutionContext, Future}
-import scala.util.Try
+import scala.util.{Failure, Success, Try}
 import scala.util.control.NonFatal
 
 class ImageLoaderController(auth: Authentication,
                             downloader: Downloader,
                             store: ImageLoaderStore,
+                            uploadStatusTable: UploadStatusTable,
                             notifications: Notifications,
                             config: ImageLoaderConfig,
                             uploader: Uploader,
@@ -49,7 +55,7 @@ class ImageLoaderController(auth: Authentication,
   def quarantineOrStoreImage(uploadRequest: UploadRequest)(implicit logMarker: LogMarker) = {
     quarantineUploader.map(_.quarantineFile(uploadRequest)).getOrElse(uploader.storeFile(uploadRequest))
   }
-  
+
   def loadImage(uploadedBy: Option[String], identifiers: Option[String], uploadTime: Option[String], filename: Option[String]): Action[DigestedFile] =  {
 
     implicit val context: RequestLoggingContext = RequestLoggingContext(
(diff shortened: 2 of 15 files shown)


sihil:
Hi @abdelrahmansd - thanks for this PR! 

I assume this is the start of the client side part of #3123? Please have a read of https://github.com/guardian/grid/blob/main/.github/CONTRIBUTING.md if you've not already. Specifically, please ensure that the description of the PR follows our guidelines.

If you are looking for specific feedback ahead of marking this ready to review, please let us know here or on slack.

Thanks again!

abdelrahmansd:
Thanks, @sihil, updating it now.

image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -170,6 +179,8 @@ class ImageLoaderController(auth: Authentication,
         .map {
           r => {
             logger.info("importImage request end")
+            uploadStatusTable.setStatus(UploadStatus(uri.split("/").last, uploadedBy.getOrElse(FALLBACK),
+              filename.getOrElse(FALLBACK), UploadStatus.COMPLETED))
```
sihil, line 183: This should be a partial update - we only need to update the status here, not overwrite the other fields.

This only deals with the happy path. If the import fails we do not update the `UploadStatus`.

You might want to do this in an `onComplete` so that both the success and failure cases can be dealt with in a single place. However the `.recover` section converts `Failure` back into `Success` so it would need to be before that.

Remember also that we need to let the user know _what_ went wrong. As such, the UploadStatus needs an optional error message.

image-loader/app/lib/UploadStatusTable.scala
```diff
@@ -0,0 +1,23 @@
+package lib
+
+import com.gu.mediaservice.lib.aws.DynamoDB
+import com.gu.mediaservice.lib.aws.DynamoDB.caseClassToMap
+import model.UploadStatus
+
+import scala.concurrent.ExecutionContext.Implicits.global
+
+
+class UploadStatusTable(config: ImageLoaderConfig) extends DynamoDB(config, config.uploadStatusTable) {
```
sihil, line 10: Newer DynamoDB table should use Scanamo, a higher level library, for reading and writing to DynamoDB. Leases is currently the only place that does this, but we'd prefer it to use this if possible.

image-loader/app/lib/UploadStatusTable.scala
```diff
@@ -0,0 +1,23 @@
...
+
+  def getStatus(imageId: String) = {
+    jsonGet(imageId, key) map { dynamoEntry =>
+      (dynamoEntry \ key ).toOption
+    }
+  }
+
+  def setStatus(uploadStatus: UploadStatus) = {
+    jsonAdd(uploadStatus.id, key, caseClassToMap(uploadStatus))
+  }
```
sihil, line 22: As well as this setStatus, you'll need a partial update mechanism that can set just the status and error.

image-loader/app/model/UploadStatus.scala
```diff
@@ -0,0 +1,20 @@
+package model
+
+import play.api.libs.json._
+
+case class UploadStatus(
+                         id: String,
+                         uploadedBy: String,
+                         fileName: String,
```
sihil, line 8: Both `uploadedBy` and `fileName` would be better modelled as `Option[String]`. There are also some missing fields here for `identifiers` and `uploadTime` - which again can both be `Option[String]`.

image-loader/app/model/UploadStatus.scala
```diff
@@ -0,0 +1,20 @@
...
+                         status: String
+                       )
```
sihil, line 10: We will need to include an error field on the UploadStatus so as to tell users what exactly it was that went wrong.

image-loader/conf/routes
```diff
@@ -3,6 +3,12 @@ POST    /images                     controllers.ImageLoaderController.loadImage(
 POST    /imports                    controllers.ImageLoaderController.importImage(uri: String, uploadedBy: Option[String], identifiers: Option[String], uploadTime: Option[String], filename: Option[String])
 GET     /images/project/:imageId    controllers.ImageLoaderController.projectImageBy(imageId: String)
 
+# Upload Status
+GET     /uploadStatus/:imageId      controllers.UploadStatusController.getUploadStatus(imageId: String)
+POST    /uploadStatus               controllers.UploadStatusController.updateUploadStatus
```
sihil, line 8: I would suggest that the URI for this POST mirrors GET (i.e. `/uploadStatus/:imageId`). If this endpoint then only takes the status and optional error in the JSON body then you'll have the imageId in the path.

dev/cloudformation/grid-dev-core.yml
```diff
@@ -195,6 +196,20 @@ Resources:
...
+      TableName: UploadStatusTable
+      AttributeDefinitions:
+      - AttributeName: id
+        AttributeType: S
+      KeySchema:
+      - AttributeName: id
+        KeyType: HASH
+      ProvisionedThroughput:
+        ReadCapacityUnits: 1
+        WriteCapacityUnits: 1
```
sihil, line 211: This table should probably have a TTL of a few minutes to avoid us having to add code to tidy up rows.

sihil (changes requested):
Nice and tidy PR which follows code standards and is fundamentally sound. Quite a lot of initial feedback I'm afraid. Happy to run through in a Zoom call tomorrow if helpful.

In addition to the comments below:
 * At the start of `ImageLoaderController.importImage` we should be checking for the existance of a row in the `UploadStatus` table (after the file has been digested but before we create the `uploadReqeust`). If it exists then the contents should be used for filename/uploadedBy/uploadTime/identifiers.

image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -85,16 +88,22 @@ class ImageLoaderController(auth: Authentication,
       result map { r =>
         val result = Accepted(r).as(ArgoMediaType)
         logger.info("loadImage request end")
+        if(!config.uploadToQuarantineEnabled){
+        uploadStatusTable.setStatus(UploadStatus(req.body.digest, context.initialMarkers("uploadedBy"),
+          context.initialMarkers("filename"), UploadStatus.COMPLETED))
+        }
```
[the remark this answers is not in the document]
sihil: If the ID doesn't exist then this should return 404.

image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -85,16 +88,22 @@ class ImageLoaderController(auth: Authentication,
...
         result
       } recover {
         case e =>
           logger.error("loadImage request ended with a failure", e)
+          uploadStatusTable.setStatus(UploadStatus(req.body.digest, context.initialMarkers("uploadedBy"),
+            context.initialMarkers("filename"), UploadStatus.FAILED))
```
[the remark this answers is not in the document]
sihil: Also here: if the ID doesn't exist then this should return 404 rather than update any DB entry.

abdelrahmansd:
rebased onto the latest main.

image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -148,38 +155,58 @@ class ImageLoaderController(auth: Authentication,
       )
 
       logger.info("importImage request start")
+      val existingImageStatus = uploadStatusTable.getStatus(uri.split("/").last).map {
```
sihil, line 158: Splitting the uri here is only going to work if the image ID is the last component which is not guaranteed. It would be much better to use the actual ID of the file which is available as the `digest` from the `digestedFile` in the for comprehension.

image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -148,38 +155,58 @@ class ImageLoaderController(auth: Authentication,
...
+      val importResult = for {
+        record <- existingImageStatus
         validUri <- Future { URI.create(uri) }
         digestedFile <- downloader.download(validUri, tempFile)
         uploadRequest <- uploader.loadFile(
           digestedFile,
           request.user,
-          uploadedBy,
-          identifiers,
+          record._1.orElse(uploadedBy),
```
sihil, line 171: It's best to avoid `_N` notation as it's hard to comprehend what the content of each `N` is (more so if the assignment is several lines earlier). I think here it would be better if `record` was an option. Then you can do `maybeRecord.flatMap(_.uploadedBy).orElse(uploadedBy)`.

image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -148,38 +155,58 @@ class ImageLoaderController(auth: Authentication,
...
+          record._2.orElse(identifiers),
           DateTimeUtils.fromValueOrNow(uploadTime),
```
sihil, line 173: uploadTime should also come from the record if it is supplied.

image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -148,38 +155,58 @@ class ImageLoaderController(auth: Authentication,
       )
 
       logger.info("importImage request start")
+      val existingImageStatus = uploadStatusTable.getStatus(uri.split("/").last).map {
+        case Some(Right(record)) => (record.uploadedBy, record.identifiers, record.fileName)
+        case _ => (None, None, None)
+      }.recover{case _ => (None, None, None)}
```
sihil, line 161: If you make the whole record an option you can write these three lines as `_.flatMap(_.toOption)`

sihil (changes requested):
Thanks for adding code to consume the upload status record on import. It will be more robust if we use the ID from the digest rather than require that the user of /imports provides a URI with the ID on the end.

sihil (approved):
I think this is ready to be merged. We need to create the upload status table in our PROD environment but we can do that today if you're happy for this to be merged.

Issue #1523, opened 2015-12-03
No results when searching 'Will' 


tsop14:
Searching

[Will](https://media.gutools.co.uk/search?query=Will) - no images

[Will "Self"](https://media.gutools.co.uk/search?query=Will%20%22Self%22) - no images

 ["Will Self"](https://media.gutools.co.uk/search?query=%22Will%20Self%22) -  images with "self" in description 

["Will" Self](https://media.gutools.co.uk/search?query=%22Will%22%20Self) - images of Will Self. ✔️

I think as 'Will' is ignored, "Will Self" becomes a type of "Self". 
There are also different number of results for searching: 
" Self"
"Self"
Self" 

Expected: Where there's a 'will'... there are images with 'will' in the description 

Perhaps similar to #1422

AWare:
Fixed in #3130.

PR #3130, opened 2021-01-26
We will always have (results for) Paris

AWare:
## What does this change?

The BBC mentioned that searches for `in:Paris` were failing unless it was quoted `in:"Paris"`. These quotes trigger the `Phrase` analyser to be used instead of `Words`. This was deliberately setting a stemmer, which appeared to be stemming Paris to Pari. (pronounced same) 

This removes the explicit setting of the english S stemmer for multi queries. In [elastics docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-match-query.html) this allows fields with heterogeneous analysers to be combined in cross match queries (allowing a match to be located in any one of the fields present). However, [these are all](https://github.com/guardian/grid/blob/b24a767c4150ff25418a5f13eaf7ba25778a7ebd/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala#L96) `standardAnalysed` fields- which would suggest only applying the stemmer.

## ~Questions~ Answers

1. Are we stemming when we write the fields? 
	- Yes, but to the `englishAnalysedCatchAll` field, I don't think this query would read that unless ✨elasticsearch✨ was doing magic.
2. Here https://github.com/guardian/grid/blob/2a2303bcdf112628f8b9ae7a651f8b2871eb4ea9/media-api/app/lib/querysyntax/QuerySyntax.scala#L123-L140 where `in` maps to `location`, why doesn't it actually map to `subLocation`.
	- Cannot find a good reason for this, so I've gone and fixed it.
	- There is no mapping for `location` apart from `subLocation`.
3. Is the test (changed in this PR) deliberate about it's need for the stemmer or does the test just match the existing code?
	- Discussed with @mbarton and we can't see a reason this stemmer would have been applied to the query.
	- Looks like it came across in the es migration.
4. Are we correctly constructing the query here, what is the _right_ stemmer etc to use for our mappings?
	- We could outsource more of this to es defaults, but that's a lot of pain for marginal (if any) gain.

Number of images in:Paris.

This may be related to https://github.com/guardian/grid/issues/1523 

## Screenshots

image
image

## Tested?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/IndexSettings.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/IndexSettings.scala
@@ -22,9 +22,9 @@ object IndexSettings {
     )
 
     val filters: List[TokenFilter] = List(
-// I (Justin) don't think we need to specify these, but can just refer to them by name (below)
-//      LowercaseTokenFilter,
-//      AsciiFoldingTokenFilter,
+      // I (Justin) don't think we need to specify these, but can just refer to them by name (below)
+      // LowercaseTokenFilter,
+      // AsciiFoldingTokenFilter,
       StemmerTokenFilter(name = english_possessive_stemmer, lang = "possessive_english"),
       StopTokenFilter(name = gu_stopwords, stopwords = Seq("_english_")),
       StemmerTokenFilter(name = s_stemmer, lang = "minimal_english")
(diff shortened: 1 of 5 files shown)


AWare:
Insufficiently verbose.

mbarton (approved):

AWare:
resolves #1523

PR #3131, opened 2021-01-26
Pluggable authorisation

sihil:
## What does this change?
This is an initial cut at making the authorisation part of the auth+auth story pluggable in the same way as authentication and image processors.

The two main imagined sources of authorisation data is:
 * aqcuired during authentication
 * queried from an external source

In the first case, data can be attached to the principal in the attributes field during authentication (this will currently need to happen in the AuthenticationProvider).

In the second case, data can be obtained using the provided http client and cached within the provider as necessary.

I'm not completely happy with the provider API yet. I have tried to add the ability to support parameterised permissions (i.e. a permission that has a dimension of some sort - in order to support multi-tenanting in the future for example). I suspect that this will not be completely suitable in reality so might remove it for the time being rather than tying us to a particular implementation.

The BBC and any other Grid users can write their own custom authorisation logic.

## Who should look at this?
@guardian/digital-cms @RichardLe

## Tested?
- [x] locally by committer

## TODO
- [X] apply https://github.com/guardian/editorial-tools-platform/pull/484 in TEST
- [x] apply https://github.com/guardian/editorial-tools-platform/pull/484 in PROD


rest-lib/src/main/scala/com/gu/mediaservice/lib/auth/Authorisation.scala
```diff
@@ -0,0 +1,51 @@
...
+    principal: Principal => {
+      principal match {
+        // a machine principal with internal tier can always see anything
+        case MachinePrincipal(ApiAccessor(_, Internal), _) => true
+        case _ => provider.hasPermissionTo(permission, principal)
+      }
+    }
+  }
+
+  def visibilityFilterFor[T](permission: PermissionWithParameter[T], principal: Principal): VisibilityFilter[T] =
```
gribeiro, line 45: Hello @sihil.
We are wondering if this method would work for filtering images by permission, i.e., showing to user a set of images depending on his permission.

sihil: That's the intention. Obviously parameterised permissions are a little down the road and this probably won't be entirely fit for purpose so I expect at the time it will need to be refined. In particular I imagine that we might sometimes want to shape an ES query from a set of `T` that a user can see, where `T` might be organisation.

rest-lib/src/main/scala/com/gu/mediaservice/lib/guardian/auth/PermissionsAuthorisationProvider.scala
```diff
@@ -0,0 +1,69 @@
...
+
+
+  override def hasPermissionTo[T](permissionContext: PermissionContext[T], principal: Principal): Boolean = {
+    def hasPermission(permission: PermissionDefinition): Boolean = {
+      principal match {
+        case UserPrincipal(_, _, email, _) => permissions.hasPermission(permission, email)
+        case _ => false
+      }
+    }
+    permissionContext match {
```
aug24, line 53: Surely we could dry this a bit?

aug24 (changes requested):

rest-lib/src/main/scala/com/gu/mediaservice/lib/guardian/auth/PermissionsAuthorisationProvider.scala
sihil: Not sure how. Note that the permission on the left is the grid's idea of a permission and the one on the right is the Guardian's permission library idea of a permission so this is mapping from one domain to another.

We could, and I'm regretting starting to type this already, do a getClass.getName.toString and then use reflection. But I think we can all agree that this would be a bad idea.

aug24 (approved):

twrichards (approved):
LGTM 👍

rest-lib/src/main/scala/com/gu/mediaservice/lib/auth/Authorisation.scala
sihil: I've stripped out the more complex permissions model for the time being. We can perhaps arrange for the BBC/Guardian to pair on it once we have a concrete use case. I've deliberately left it as an unsquashed commit so it is there if we want it back.

Issue #3138, opened 2021-01-27
>200 images on screen at once won't load


AWare:
## Short description
there's a max query size of 200, if more images than this will fit on your screen the query will just get rejected
## Steps to reproduce
keep zooming out and reload
## Expected results

this limit should be respected when fetching

PR #3139, opened 2021-01-28
respect max query in ui

AWare:
## What does this change?

Media-Api sets a limit of 200 to the query size, but kahuna doesn't respect it. This causes it to fail to fetch search results if more than 200 images fit in the user viewport. 

This restricts the maximum query size that gu-lazy-table can make, this allows kahuna to repeat the query and get all the results.

Fixes https://github.com/guardian/grid/issues/3138

Load the grid, zoom out until >200 images are displayed.
Reload the page. 

Before: no images will be shown.
After: works as expected. 

## Screenshots

Let's presume a special version of the grid for these screenshots, called the 'gr'. 
For some reason the max query size is 5.

Before:
image
After:
image

## Tested?
- [X] locally by committer

### DIFF
--- a/kahuna/public/js/components/gu-lazy-table/gu-lazy-table.js
+++ b/kahuna/public/js/components/gu-lazy-table/gu-lazy-table.js
@@ -25,6 +25,8 @@ export var lazyTable = angular.module('gu.lazyTable', [
     'util.seq'
 ]);
 
+// Set in ElasticSearchModel.scala
+const maxSize = 200;
 
 
 function asInt(string) {
@@ -97,6 +99,13 @@ lazyTable.controller('GuLazyTableCtrl', ['range', function(range) {
             filter(({$start, $end}) => $start !== -1 && $end !== -1).
             // Ignore if $start after $end (incomplete combine$ state)
             filter(({$start, $end}) => $start <= $end).
+            // Max query size
+            map(({$start, $end})=> {
+              if (($end - $start) < maxSize) {
+                return {$start, $end};
+              }
+              return {$start, $end: $start + maxSize - 1};
+            }).
             distinctUntilChanged(({$start, $end}) => `${$start}-${$end}`);
 
         // Placeholders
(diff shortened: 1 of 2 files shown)


kahuna/public/js/components/gu-lazy-table/gu-lazy-table.js
```diff
@@ -97,6 +97,13 @@ lazyTable.controller('GuLazyTableCtrl', ['range', function(range) {
             filter(({$start, $end}) => $start !== -1 && $end !== -1).
             // Ignore if $start after $end (incomplete combine$ state)
             filter(({$start, $end}) => $start <= $end).
+            // Max query size of 200
+            map(({$start, $end})=> {
+              if ($end - $start > 200) {
```
sihil, line 102: 🤔 Should this be `$end - $start <= 200`?

AWare: oops, it was right the first time. 

the equal is equivalent though, so I'm not particular

kahuna/public/js/components/gu-lazy-table/gu-lazy-table.js
```diff
@@ -25,6 +25,8 @@ export var lazyTable = angular.module('gu.lazyTable', [
     'util.seq'
 ]);
 
+// Set in ElasticSearchModel.scala
```
sihil, line 28: 💯

media-api/app/lib/elasticsearch/ElasticSearchModel.scala
```diff
@@ -189,6 +189,8 @@ object SearchParams {
 
   type SearchParamValidation = Validation[InvalidUriParams, SearchParams]
   type SearchParamValidations = ValidationNel[InvalidUriParams, SearchParams]
+
+  // Also adjust in gu-lazy-table.js
```
sihil, line 193: 💯

sihil (approved):
Looks good to me. Glad to see comments to remind us that this constant is in two places.

PR #3145, opened 2021-01-29
fix lockfile version error

AWare:
accidentally commited an npm lockfile from the wrong npm version, ran npm install on main to fix


paperboyo (approved):
This started happening to me on LOCAL too recently…

PR #3146, opened 2021-01-30
Read country verbal names before falling back to country codes

paperboyo:
https://github.com/guardian/grid/pull/3140/ introduced reading country codes and that’s useful if verbal names are not there. But it turns out, if unrecognised codes are encountered, the verbal names are not read at all now, because our CountryCode cleaner operates on the first values encountered.

And we are seeing some images with unrecognised codes (eg. `POR`, `ROM`, `GZA`), but with correct (or just more useful) verbal names (respectively, `Portugal`, `Romania`, `Gaza`).

This won’t be perfect, but we can be only as good when trying to clean the data provided. It will fail when a verbal name contains gibberish, but a country code is correct. We are taking a leap of faith that this is a less common scenario than what this PR tries to solve.

It should provide better location data.


AWare (approved):
think we're cleaning at the wrong point, let's get this in for now

akash1810:
Be good to have a test for this - https://github.com/guardian/grid/blob/3b03881f9aa9a6331bb34c99967712e842540998/image-loader/test/scala/lib/imaging/FileMetadataReaderTest.scala could be used for inspiration?

AWare:
Refactoring the cleaner now.

PR #3150, opened 2021-02-01
Revert "Revert "update cropperjs""

AWare:
Reverts guardian/grid#3148


paperboyo (approved):
I trust ya :-).

AWare:
_safe_Apply was not actually calling apply. switched to `$timeout` which will call apply outside the digest loop and now it works.

jonathonherbert (approved):
Tested on CODE 👏

PR #3153, opened 2021-02-02
Make country extraction more reliable.

AWare:
## What does this change?

This seperates the logic for finding a fallback country _name_ from finding a country name from an ISO code. 

This also swaps our weird Locale array search for a pair of `Map`s for 2 and 3 letter codes. 

The ordering remains the same, and the same fields are pulled for each check. Previously, this mattered because we would take the first `Some` and populate `metadata.country` from it. The order was tweaked to attempt to get a country name before an unidentifiable country code.  

Now, we check each field where a country may be for a code- and then see whether of those are hits. If one is a hit, then it's used. If nothing hits, then the first of them in the extractor is used. 

We could thus, drop the 'code' extractors from the extractor, and the 'name' ones from the cleaner. I'm not sure whether image data backs this up- so have left all in for completeness. 

Also, now these are maps instead of `Locales` then we can add support for more esoteric country codes- and probably drop the 2-3 letter code distinction by just testing all of the names against the maps. 

Better country metadata. 
Hopefully the existing tests for this cleaner will both just work, validating this change. 🤞🏻

## Who should look at this?

@paperboyo what ordering should we use for this?

## Tested?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/CountryCode.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/CountryCode.scala
@@ -22,7 +22,16 @@ object CountryCode extends ImageProcessor with GridLogging {
       localeCodesAndNames.map(c => c.threeCode -> c.displayName).toMap ++
       Map(
         "UK" -> "United Kingdom",
-        "GB" -> "United Kingdom"
+        "GB" -> "United Kingdom",
+        "GDR" -> "German Democratic Republic",
+        "GRG" -> "Georgia",
+        "KOS" -> "Kosovo",
+        "POR" -> "Portugal",
+        "ROM" -> "Romania",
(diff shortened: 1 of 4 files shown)


paperboyo:
> what ordering should we use for this?

[This ordering](https://github.com/guardian/grid/pull/3153/files#diff-4b80adb60ae76232daa91bbe97b3a33a14f4cbdbd6a4aa536af15622a9ea1256R59-R62) (as in current PROD) makes sense.

> `// newest? the link shows 2008 for location`
Yep, the link leads to new schema we don’t read  :-). But not many suppliers are using it, TBH. Not sure adding this comment is useful.

jonathonherbert (commented):
Code looks good – is there additional behaviour introduced by e.g. this new ordering behaviour? If so, we should add tests.

jonathonherbert (approved):
🎉

PR #3163, opened 2021-02-12
Add fetch from s3 for reindex functionality

aug24:
## What does this change?
This PR provides an additional GET endpoint which triggers a reload of an image file already present in S3.

This can be used to rectify failed loads eg files which previously failed date processing, and we left 'orphaned' in S3.

No changes to current behaviour, orphaned files can be uploaded.

## Tested?
- [X] locally by committer


image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -56,15 +53,54 @@ class ImageLoaderController(auth: Authentication,
...
+      for {
+        meta <- downloader.fetchImage(identifier, tempFile)
+        digestedFile = DigestedFile(tempFile, identifier)
+        // get original upload time etc from s3 meta
+        uploadedBy = meta.get("x-amz-meta-uploaded_by").getOrElse("Unknown")
+        uploadTime = meta.get("x-amz-meta-upload_time")
+        filename = meta.get("x-amz-meta-file_name")
+        // Add to logging markers
+        widerContext = context.copy(initialMarkers = context.initialMarkers + (
+          "uploadedBy" -> uploadedBy,
```
AWare, line 80: this is shadowed from the args on 75, I think the logs might best have the uploadedBy from the function args?

image-loader/conf/routes
```diff
@@ -1,5 +1,7 @@
 GET     /                           controllers.ImageLoaderController.index
-POST    /images                     controllers.ImageLoaderController.loadImage(uploadedBy: Option[String], identifiers: Option[String], uploadTime: Option[String], filename: Option[String])
+POST    /images                     controllers.ImageLoaderController.loadImage(uploadedBy: String, identifiers: Option[String], uploadTime: Option[String], filename: Option[String])
+# Not exposed in UI - force complete reload of file already correctly stored in S3.
+GET     /imageReload                controllers.ImageLoaderController.reloadImage(uploadedBy: Option[String], identifiers: String)
```
AWare, line 4: Should this be a POST instead?

aug24: As I said to Simon, there is a very good reason for this....  I'll discuss before merge.

AWare:
Otherwise happy.

AWare:
Can we test this PR against images which do not match their hashes but are otherwise _servicable_?

PR #3165, opened 2021-02-12
Sanitise creation and reading of S3 user metadata

sihil:
## What does this change?
This normalises and cleans up the way the user metadata on the S3 objects is created and read back in. The main changes are that we:
 - Have a set of vals for the header names and prefixes so they are not duplicated
 - Always URL encode values stored in user metadata (and URL decode on retrival)
 - Lowercase all incoming identifier keys

Having a set of vals rather than duplicating the same string in multiple places is obviously a good thing.

Only US-ASCII data can be stored in user metadata, so we should use an encoding to be sure that the data will be successfully stored and will not be corrupted. We use URL encoding for `file-name` already so we should extend that to other fields. We can now simplify the logic by doing it on everything. 

Identifiers such as picdarURN end up as picdarurn when stored in and retreived via S3 user metadata as all such metadata keys are always lowercased. Unfortunately we don't lowercase them before storing them in ElasticSearch so when we reindex the identifier keys will change. We could special case the existing examples but I suggest that we just put up with it and change any external tooling as necessary.

### What about our historical data?
When deployed, this new code will mean that fields that were not URL encoded will be URL decoded. Taking a sample of 3.5M images I've not found any values that do not decode back to themselves (other than `file-name` which was already encoded so is safe). 

I did this using a couple of little tools to do analysis of existing file metadata which I've checked in for posterity.

## Tested?
- [x] review wider corpus of identifier values in existing data
- [x] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/ImageStorage.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/ImageStorage.scala
@@ -13,9 +13,9 @@ import com.gu.mediaservice.model.MimeType
 object ImageStorageProps {
   val cacheDuration: Duration = 365 days
   val cacheForever: String = s"max-age=${cacheDuration.toSeconds}"
-  val filenameMetadataKey: String = "file_name"
-  val uploadTimeMetadataKey: String = "upload_time"
-  val uploadedByMetadataKey: String = "uploaded_by"
+  val filenameMetadataKey: String = "file-name"
+  val uploadTimeMetadataKey: String = "upload-time"
+  val uploadedByMetadataKey: String = "uploaded-by"
   val identifierMetadataKeyPrefix: String = "identifier!"
 }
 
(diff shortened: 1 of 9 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/ImageStorage.scala
```diff
@@ -13,6 +13,10 @@ import com.gu.mediaservice.model.MimeType
 object ImageStorageProps {
   val cacheDuration: Duration = 365 days
   val cacheForever: String = s"max-age=${cacheDuration.toSeconds}"
+  val filenameMetadataKey: String = "file_name"
```
akash1810, line 16: Does the latest version of localstack handle snake_case metadata properly now? Related https://github.com/guardian/grid/pull/2931.

sihil: OMG 🤯 This is why we're not allowed nice things. Python shouldn't be allowed to be used by anyone older than 12.

Seriously though - also another example of relying on a simulator that isn't close enough to the real thing because it is by definition not the real thing. I guess the counter argument is that if we'd used it from the beginning then it would all be fine.

Thanks for pointing out.

AWare (approved):

PR #3166, opened 2021-02-15
Festina lente

paperboyo:
## What does this change?
We are using a very nice fading of thumbnails and main preview images. Looks very premium. Sadly, it depends on them being loaded in full. We have [imgOps configured](https://github.com/guardian/grid/blob/9588b1c2b5074866ffd11a3dd8176592271cb432/dev/imgops/nginx.conf#L30) to provide progressive JPEGs. Progressive JPEGs start showing up before they are fully loaded. This is a draft PR to test the hypothesis that the app will feel more responsive if images start to show up earlier.

This also changes the thumbnail generation, so that they are also progressive JPEGs.

Things to look out for:
- `background-image` chequerboard from underneath main preview image starts showing, like holes in socks
- progressive rendering looks less cool even though it may feel speedier

App feels faster without feeling uglier.

## Screenshots
Progressive on top, current at the bottom; Slow 3G throttling.

### Main preview
both

### Thumbnails (played at 25% speed for effect)
both_thumb5

## Who should look at this?
Someone who knows how to hide the chequerboards and blue frames (?) until images fully load?

## Tested?
- [x] locally by committer


AWare:
I think if we kept the image onLoad logic, and add a short circuit to it so if it takes longer than (say) 1 second, it shows what it has.

paperboyo:
> so if it takes longer than (say) 1 second

I’m open to any adjustments! Would just be hard to come up with any specific timeout, since they can basically be instantaneous instead of prolonged. Especially for thumbnails.

Do you think it may sense to deploy to TEST and just play around. I don’t think the fact that in this period TEST will create progressively encoded thumbnails matter… Unless I have missed something, which is very possible…

AWare:
I don't want to be mr "i hate angular" about this, but I think this is getting caught in the dispatch loop which is causing it to be slower than otherwise. _I think_ that when you get a pagefull of images fade in at once, it's the promises or timeouts in the directive all getting batched into the same cycle

paperboyo:
Superseded by superior https://github.com/guardian/grid/pull/3169.

PR #3169, opened 2021-02-15
Faster fades on load and progressive display of images

AWare:
## What does this change?

We are using a very nice fading of thumbnails and main preview. Sadly, it depends on them being loaded in full. We have [imgOps configured](https://github.com/guardian/grid/blob/9588b1c2b5074866ffd11a3dd8176592271cb432/dev/imgops/nginx.conf#L30) to provide progressive JPEGs.

This PR:
- changes the fading so that it doesn’t depend on an image being fully loaded
- changes the thumbnail generation, so that they are also progressive JPEGs

Currently, the image fade in on load applies an element style using the
angular digest loop. As this style stuff is done on the element directly
there's no state here for angular to track. I've reorganised things to:
- make the reveal happen before we do any $q promisey stuff, so no more waiting on the digest loop
- set a timeout (which ironically will fire in the digest loop) so that after a while we just show the image

I think this would work better with #3166 from @paperboyo than just dropping the fade. The app should feel more snappy.

## Screenshots
(bottom: current situation)
### Viewer:
both
### Browser  (played at 25% speed for effect):
both_thumb5

## Who should look at this?
@guardian/digital-cms 

## Tested?
- [X] locally by committer

### DIFF
--- a/kahuna/public/js/directives/gr-image-fade-on-load.js
+++ b/kahuna/public/js/directives/gr-image-fade-on-load.js
@@ -7,21 +7,25 @@ imageFade.directive('grImageFadeOnLoad',
                      function ($q, $timeout) {
 
     // TODO: customise duration, transition
-    const animationThreshold = 50; // ms
-    const animationDuration = 200; // ms
+                         const animationDuration = 200; // ms
+                         const revealAfter = 200; // ms
 
     return {
         restrict: 'A',
         link: function (scope, element) {
-            // If not loaded after animationThreshold, hide and wait
+            // If not loaded, hide and wait
             // until loaded to fade in
-            $timeout(() => {
                 if (! isLoaded()) {
                     hide();
-                    whenLoaded().finally(reveal);
+                    whenLoaded();
                 }
-            }, animationThreshold);
 
+            // If not loaded after revealAfter
+            // show the image, it's progressive so we should
+            // have something to show for our time
+            $timeout(() => {
+                reveal();
+            }, revealAfter);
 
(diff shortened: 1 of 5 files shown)


sihil:
Waht are you thinking about this @AWare. I'm keen not to pick up any more work like this, but if this is close to shippable then it makes sense to realise the value by getting it over the line rather than parking it to rot.

paperboyo:
This indeed works better than #3166 (see additional explanation/rationale and comparisons with current PROD there). Here are further comparisons. This PR – top, #3166 – bottom:
## Viewer:
alextopmatboittom_b
## Browser:
alextopmatboittom_browser

kahuna/public/js/directives/gr-image-fade-on-load.js
```diff
@@ -7,41 +7,52 @@ imageFade.directive('grImageFadeOnLoad',
                      function ($q, $timeout) {
 
     // TODO: customise duration, transition
-    const animationThreshold = 50; // ms
-    const animationDuration = 200; // ms
+                         const animationDuration = 200; // ms
+                         const revealAfter = 1500; // ms
```
twrichards, line 11: [nit] think this indentation is accidental

AWare: yeah, the joys of prettier

twrichards (approved):
logic LGTM 👍  (i'm in the middle of a rebase of my grid work so I haven't tested it locally)

PR #3170, opened 2021-02-16
better log message

AWare:
## What does this change?

`BadInputException` now has an override for `getMessage` so the logs are more useful.

Better log messages from usages. 

## Tested?
- [x] locally by committer

I threw this in AppController and it gave a useful message, so hopefully this will make usages easier to log dive.

### DIFF
--- a/usage/app/lib/BadInputException.scala
+++ b/usage/app/lib/BadInputException.scala
@@ -1,3 +1,5 @@
 package lib
 
-class BadInputException(message: String) extends Exception
+class BadInputException(message: String) extends Exception {
+  override def getMessage: String = message
+}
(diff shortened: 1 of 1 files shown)


usage/app/lib/BadInputException.scala
```diff
@@ -1,3 +1,5 @@
 package lib
 
-class BadInputException(message: String) extends Exception
+class BadInputException(message: String) extends Exception {
+  override def getMessage: String = message
+}
```
sihil, line 5: Would passing it straight through to `Exception` work?
```suggestion
class BadInputException(message: String) extends Exception(message)
```

AWare: Almost certainly.

sihil (approved):
LGTM

PR #3172, opened 2021-02-17
add GET upload status API poll

abdelrahmansd:
## What does this change?
This Introduce new API Polling for `GET /image-loader/UploadStatus` to get image upload status 
and show it in the UI

In case of quarantine enabled and quarantine processor failed to process an image 
the quarantine status should show up below the image 

## Screenshots

Screenshot from 2021-03-03 12-53-42

## Tested?
- [x] locally by committer

### DIFF
--- a/image-loader/app/model/QuarantineUploader.scala
+++ b/image-loader/app/model/QuarantineUploader.scala
@@ -36,7 +36,7 @@ class QuarantineUploader(val store: QuarantineStore,
       meta
     )
   }
-  
+
   def quarantineFile(uploadRequest: UploadRequest)(
     implicit ec: ExecutionContext,
     logMarker: LogMarker): Future[JsObject] = {
@@ -45,7 +45,7 @@ class QuarantineUploader(val store: QuarantineStore,
 
     for {
       _ <- storeQuarantineFile(uploadRequest)
-      uri = s"${config.apiUri}/images/${uploadRequest.imageId}"
+      uri = s"${config.rootUri}/uploadStatus/${uploadRequest.imageId}"
     } yield {
       Json.obj("uri" -> uri)
     }
--- a/image-loader/app/model/Uploader.scala
+++ b/image-loader/app/model/Uploader.scala
@@ -343,7 +343,7 @@ class Uploader(val store: ImageLoaderStore,
       updateMessage = UpdateMessage(subject = "image", image = Some(imageUpload.image))
       _ <- Future { notifications.publish(updateMessage) }
       // TODO: centralise where all these URLs are constructed
-      uri = s"${config.apiUri}/images/${uploadRequest.imageId}"
+      uri = s"${config.rootUri}/uploadStatus/${uploadRequest.imageId}"
     } yield {
       Json.obj("uri" -> uri)
     }
--- a/kahuna/public/js/upload/jobs/upload-jobs.js
+++ b/kahuna/public/js/upload/jobs/upload-jobs.js
@@ -18,6 +18,7 @@ export var jobs = angular.module('kahuna.upload.jobs', [
 
 jobs.controller('UploadJobsCtrl', [
     '$rootScope',
+    '$q',
     '$scope',
     '$window',
     'apiPoll',
@@ -27,6 +28,7 @@ jobs.controller('UploadJobsCtrl', [
     'editsService',
 
     function($rootScope,
+            $q,
             $scope,
             $window,
             apiPoll,
@@ -43,6 +45,15 @@ jobs.controller('UploadJobsCtrl', [
     // State machine-esque async transitions
     const eventName = 'Image upload';
 
+    function getUploadStatus(image) {
+        return image.get().then((status) => {
+            if (status.data.status === 'COMPLETED' || status.data.status === 'FAILED') {
+                return status;
+            } else {
+                return $q.reject();
+            }
+        });
+    }
     ctrl.jobs.forEach(jobItem => {
         jobItem.status = 'uploading';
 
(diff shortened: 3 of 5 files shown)


image-loader/conf/routes
```diff
@@ -3,6 +3,8 @@ POST    /images                     controllers.ImageLoaderController.loadImage(
 POST    /imports                    controllers.ImageLoaderController.importImage(uri: String, uploadedBy: Option[String], identifiers: Option[String], uploadTime: Option[String], filename: Option[String])
 GET     /images/project/:imageId    controllers.ImageLoaderController.projectImageBy(imageId: String)
 
+GET     /follow/:imageId            controllers.ImageLoaderController.follow(imageId: String)
```
akash1810, line 6: This additional `follow` endpoint is redundant here. Something like this might be simpler:

1. User POSTS an  image to `/images`
2. image-loader returns a URI to `/uploadStatus/:imageId`
3. Client polls this status endpoint
4. If the status is successful, the payload will tell the client where to start polling media-api
5. If the status is unsuccessful, client shows a message

As Grid's APIs follow the [argo-spec](https://github.com/argo-rest/spec), we should be able to utilise argo's `links` here.

image-loader/app/model/Uploader.scala
```diff
@@ -329,7 +329,7 @@ class Uploader(val store: ImageLoaderStore,
       updateMessage = UpdateMessage(subject = "image", image = Some(imageUpload.image))
       _ <- Future { notifications.publish(updateMessage) }
       // TODO: centralise where all these URLs are constructed
-      uri = s"${config.apiUri}/images/${uploadRequest.imageId}"
+      uri = s"${config.rootUri}/follow/${uploadRequest.imageId}"
```
akash1810, line 332: Following the comment below, this should be the `GET /uploadStatus/:imageId ` endpoint.

kahuna/public/js/errors/global.html
```diff
@@ -46,4 +46,9 @@
         A server error has occurred; the Grid may not be fully functional at this time.
         If the problem persists please <a href="{{ctrl.supportEmailLink}}" class="coloured-link">email the Grid team</a>.
     </div>
+
+    <div class="global-error error" ng-message="quarantineFailed">
+        Image is infected with a virus!
```
akash1810, line 51: Should the messaging be driven by what we [store in the database](https://github.com/guardian/grid/blob/f50d816f1d236f0145c1ee84f92a30c721bcab8b/image-loader/app/model/UploadStatus.scala#L20)?

Not sure this should be a global error as it's localised to a specific operation - can we position it locally to the image upload? [This](https://github.com/guardian/grid/blob/f50d816f1d236f0145c1ee84f92a30c721bcab8b/kahuna/public/js/upload/jobs/upload-jobs.html#L23-L32) should allow for messages to be shown next to the image(s) affected.

image-loader/app/model/Uploader.scala
akash1810: Would we also need to update the [`quarantineFile` endpoint](https://github.com/guardian/grid/blob/9f288a37bf26876b0103692c5ca115ad4f019e99/image-loader/app/model/QuarantineUploader.scala#L40-L51) to make it make it match this. It might be better to push this change up [to here](https://github.com/guardian/grid/blob/552671ef1445f02dbfd2aa657d99ed4cae8678b8/image-loader/app/controllers/ImageLoaderController.scala#L55-L57)?

kahuna/public/js/util/async.js
```diff
@@ -79,6 +79,28 @@ async.factory("apiPoll", [
   }
 ]);
 
+async.factory("apiStatusPoll", [
+  "$q",
```
AWare, line 83: This function isn't really needed as apiPoll takes a function which returns a promise and will continue to run it repeatedly until it the promise resolves.

Here's an example where it's used to do a similar check.
https://github.com/guardian/grid/blob/b1d75cd0bb349374d114f21b8e5aee2958e3e75f/kahuna/public/js/services/label.js#L52

AWare (commented):
Good progress, few things stand out at the moment. 
We look forward to reviewing again. Can you attach a GIF/video to the final PR please? We're happy to have a joint session if you'd like some time. 

@AWare, @sihil & @akash1810

kahuna/public/js/upload/jobs/upload-jobs.js
```diff
@@ -43,93 +45,123 @@ jobs.controller('UploadJobsCtrl', [
     // State machine-esque async transitions
     const eventName = 'Image upload';
 
+    function getUploadStatus(image) {
+      debugger;
```
sihil, line 49: Left over `debugger`

sihil (commented):
Morning @abdelrahmansd,

At a glance this looks good. 

Can you remove the debugger statement and fix the indenting and then we'll get another local build for this kicked off and try it out in our TEST environment.

kahuna/public/js/upload/jobs/upload-jobs.js
```diff
@@ -43,93 +45,124 @@ jobs.controller('UploadJobsCtrl', [
...
-                  null,
-                  'Failure',
-                  null,
-                  { 'Failed on': 'index'}
-                );
-            });
+           jobItem.status = 'indexing';
+           jobItem.resource = resource;
+
+           // TODO: grouped polling for all resources we're interested in?
```
AWare, line 64:
```suggestion
```

I don't think this todo is relevant any more.

kahuna/public/js/upload/jobs/upload-jobs.js
```diff
@@ -43,93 +45,124 @@ jobs.controller('UploadJobsCtrl', [
...
+           const imageUploadStatusResource = apiPoll(() => getUploadStatus(resource));
+           imageUploadStatusResource.then(status => {
+             if (status.data.status === "FAILED"){
+                 const quarantineNotificationDiv = document.getElementById(jobItem.name);
```
AWare, line 68: This should use the controller to insert the warning instead of manually manipulating the DOM.

kahuna/public/js/upload/jobs/upload-jobs.js
```diff
@@ -43,93 +45,124 @@ jobs.controller('UploadJobsCtrl', [
...
+                      jobItem.image = image;
+                      jobItem.thumbnail = image.data.thumbnail;
+
+                      ctrl.remaining -= 1;
+
+                      imageService(image).states.canDelete.then(deletable => {
+                           jobItem.canBeDeleted = deletable;
+                      });
+
+                      // TODO: we shouldn't have to do this ;_;
```
AWare, line 86:
```suggestion
```
This TODO is also probably worth removing. As sad as people were that `$rootScope.on` was used for state updates, it's used everywhere, and seems to be the only way that works.

kahuna/public/js/upload/jobs/upload-jobs.js
```diff
@@ -43,93 +45,124 @@ jobs.controller('UploadJobsCtrl', [
...
+                         eventName,
+                         null,
+                         'Success',
+                         null,
+                         { 'Labels' : presetLabels.length}
+                       );
+                    },
+                     error => {
+                       jobItem.status = 'upload error';
+                       jobItem.error = error.message;
```
AWare, line 120: I think this would work instead of the manual DOM stuff above.

image-loader/conf/routes
```diff
@@ -7,8 +7,6 @@ GET     /images/project/:imageId    controllers.ImageLoaderController.projectIma
 GET     /uploadStatus/:imageId      controllers.UploadStatusController.getUploadStatus(imageId: String)
 POST    /uploadStatus/:imageId      controllers.UploadStatusController.updateUploadStatus(imageId: String)
 
-
-
```
sihil, line 11: Noise, please remove.

kahuna/public/js/upload/jobs/upload-jobs.js
```diff
@@ -43,93 +45,121 @@ jobs.controller('UploadJobsCtrl', [
...
-                  null,
-                  { 'Failed on': 'index'}
-                );
-            });
+           jobItem.status = 'indexing';
+           jobItem.resource = resource;
+
+           const imageUploadStatusResource = apiPoll(() => getUploadStatus(resource));
+           imageUploadStatusResource.then(status => {
+             if (status.data.status === "FAILED"){
```
sihil, line 66: We should consistently use single quotes for JS strings. That said we'll put a linter on this code base soon to make this easier for contributors.

kahuna/public/js/upload/jobs/upload-jobs.js
```diff
@@ -43,93 +45,121 @@ jobs.controller('UploadJobsCtrl', [
...
-                $rootScope.$emit(
-                  'track:event',
-                  eventName,
-                  null,
-                  'Failure',
-                  null,
-                  { 'Failed on': 'index'}
-                );
-            });
+           jobItem.status = 'indexing';
```
sihil, line 61: This whole block still has changed indentation. Please fix it to match what was here before so that diffs don't have unnecessary noise.

sihil (commented):
Hi Ebu, one code comment - but mostly noise to resolve in the form of touched but semantically unchanged files and bad indentation. Thanks!

image-loader/app/model/QuarantineUploader.scala
```diff
@@ -36,7 +36,7 @@ class QuarantineUploader(val store: QuarantineStore,
       meta
     )
   }
-  
+
```
abdelrahmansd, line 39: remove extra space

image-loader/conf/routes
abdelrahmansd: those are extra lines

sihil: My point here is that this file isn't actually modified (semantically) but it has been modified making the PR look untidy (as it is making unnecessary changes).

abdelrahmansd: yup makes sense 👌🏾

sihil (approved):

PR #3173, opened 2021-02-17
Quarantine alert lambda

abdelrahmansd:
## What does this change?
An AWS Lambda to consume SNS (Quarantine processor pushes positive status events to it)  events and update image upload status
through the image-loader service `POST    /uploadStatus/:imageId`.
This is an implementation to this proposal [here](https://docs.google.com/presentation/d/1ccoVJyPL_pErMXwbGU0E7D2zaaYZk2PIVpLYSV7jEoQ/edit#slide=id.p).
## Tested?
- [ ] locally by committer


quarantine-status/download-deps.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+pip install -r src/requirements.txt -t src/libs
```
sihil, line 3: No requirements.txt file is provided, possibly an oversight?

quarantine-status/stacks/lambda.json
```diff
@@ -0,0 +1,170 @@
+{
```
sihil, line 1: One of the JSON or YAML versions of the CFN is sufficient. I'd vote for keeping the YAML version.

quarantine-status/stacks/lambda.yml
```diff
@@ -0,0 +1,106 @@
+Parameters:
+  LambdaHandler:
+    Default: "virus-alert-lambda.lambda_handler"
```
sihil, line 3: could this default match the python file?

quarantine-status/stacks/lambda.yml
```diff
@@ -0,0 +1,106 @@
...
+  Lambda:
+    Type: "AWS::Lambda::Function"
+    Properties:
+      Code:
+        ZipFile:
+          Fn::Join:
+            - "\n"
+            -
+              - "def lambda_handler(event, context):"
+              - " "
```
sihil, line 93: Doesn't this need to be the name of a zip file?

sihil (commented):
We're unlikely to run this code but I'm happy for it to sit in the grid repo. Would be nice for it to be a little more consistent with each other with a outline description on how it should be deployed.

sihil (approved):
It is useful to have this in here as a data point for anyone else implementing quarantining. Thanks @abdelrahmansd!

Issue #2594, opened 2019-06-24
Consistently read and reconcile colourModel and colourModelInformation


paperboyo:
We [read the above](https://github.com/guardian/grid/blob/master/image-loader/app/lib/imaging/FileMetadataReader.scala#L177) using different libraries depending on the file format. We should, maybe, use only one to have consistent output and/or reconcile different outputs into a consistent set of values.

- [ ] consider not [being dependant on EXIF](https://github.com/guardian/grid/blob/master/image-loader/app/lib/imaging/FileMetadataReader.scala#L191), cos it can be nuked/unavailable
- [x] set `hasAlpha` to `false` for all JPEGs (fixed in https://github.com/guardian/grid/pull/3604)
- [x] set bits per sample to 8bit for all JPEGs (we do not care about the dodo 12-bit JPEGs) (fixed in https://github.com/guardian/grid/pull/3604)
- [ ] populate `colourModelInformation` for JPEGs
- [ ] reconcile Bitmap, Greyscale, RGB, CMYK and Lab `colourModel`s (eg. it’s `colourModel:CMYK` for JPEGs, but `colourModelInformation>photometricInterpretation:CMYK` for TIFs)
- [ ] reconcile bits per sample (eg. it’s `bitsPerSample:16 16 16 bits/component/pixel` for TIFFs, but `bitsPerSample:16` for PNGs)

A set of [test images](https://media.test.dev-gutools.co.uk/search?query=%23%22colourModel%20Tests%22) on TEST.

PR #3174, opened 2021-02-18
Show chequerboard only if there is transparency and during cropping too

paperboyo:
## What does this change?
In the Viewer, this makes chequerboard not show up underneath images with no transparency. [We’ve noticed](https://user-images.githubusercontent.com/6032869/107898699-52056580-6f34-11eb-964e-cb3557141810.gif) that sometimes there was a brief flash of it, even though it immediately hid behind the image fully. It keeps the chequerboard for images with transparency.

We decide on transparency based on the existence of `optimisedPng`. This isn’t ideal, but we hid `hasAlpha` under fileMetadata and we do not even [have it there consistently for all images](https://github.com/guardian/grid/issues/2594). `optimisedPng` isn’t a perfect transparency synonym, but it [should get better](https://github.com/guardian/grid/issues/2395), and in any case, it won’t ever not be there if there is transparency (it may be there where there isn’t, but that’s OK as far as chequerboard is concerned).

Lastly, this permanently overrides cropper.js CSS to always show the chequerboard behind the image. We haven’t noticed a flash of it for opaque images, so no transparency check is performed here. There is [this cropper.js option](https://github.com/fengyuanchen/cropperjs#background), but I thought we would want our chequerboard look to be consistent.

These changes will make Viewer look less glitchy (no useless flashes of chequerboard) and will bring us closer to always presenting images with transparency on a chequerboard background (at the cropping stage). Thumbnail generation is the last place where we bake transparent images onto a background.

### Viewer
check nocheck
[current on the left, proposed on the right]

### Cropping stage
image
[current on the left, proposed on the right]

## Tested?
- [x] locally by committer


kahuna/public/js/image/view.html
```diff
@@ -145,7 +145,13 @@
...
              draggable="true"
              ui-drag-data="ctrl.image | asImageDragData">
-          <img class="easel__image easel__image--checkered__background"
+          <img ng-if="!ctrl.image.data.optimisedPng" class="easel__image"
+               alt="preview of original image"
+               ng-src="{{:: ctrl.optimisedImageUri}}"
+               grid:track-image="ctrl.image"
+               grid:track-image-location="original"
+               grid:track-image-loadtime/>
+          <img ng-if="ctrl.image.data.optimisedPng" class="easel__image easel__image--checkered__background"
```
AWare, line 154:
```suggestion
          <img ng-if="ctrl.image.data.optimisedPng" ng-class="[easel__image, {easel__image--checkered__background: ctrl.image.data.optimisedPng }]"
```

AWare (commented):
Think we can better do this with `ng-class`

kahuna/public/js/image/view.html
AWare: Ooops, kept the `ng-if`, this works now.

AWare (approved):

PR #3177, opened 2021-02-18
Make Grid HTTP services common

aug24:
## What does this change?
Currently the Grid offers various microservices which are all presented as HTTPS/REST interfaces.

We have code to simplify cross-service calls, but it is only present in the admin microservice.  We have intentions that Thrall and Projection code will no longer trust message content and ElasticSearch results respectively, but will fetch up-to-date data from the microservice (and thus from the source of truth, Dynamo).

This means that those services need access to the GridClient class, which entails moving it into the common library.

_NB The lambda now needs to have the domain root for the stage as an environment variable_
See https://github.com/guardian/editorial-tools-platform/pull/483

## Tested?
- [ ] locally by committer


AWare:
I'm not 100% sure on using apiKeys for this, and equally I'm not sure that we should send enough to authenticate the user request to thrall.

AWare (approved):
Let's get this in and then get it using `Principal`

PR #3178, opened 2021-02-19
Only process user/provided metadata if image has not previously been uploaded

sihil:
## What does this change?
At the moment the `uploadedBy`, `uploadTime`, `fileName` and `identifiers` are overwritten if a file is uploaded a second time. There are two places this data is stored: as user metadata on the S3 object in the original image bucket and also in the ES index. The user metadata is always completely replaced (all existing data is lost) whilst the filename is lost in ES.

This change should mean that we do not lose the original version of this data and will not propagate incorrect data through to the ES index.

Original metadata is retained both on the S3 object and in the ES index when the same image file is reuploaded allowing reingestions to get back to the same state as ES.

## Who should look at this?
@aug24 and @AWare are likely to have opinons.

## Tested?
- [ ] locally by committer
- [X] on the Guardian's TEST environment

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/ImageStorage.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/ImageStorage.scala
@@ -34,7 +34,8 @@ trait ImageStorage {
   /** Store a copy of the given file and return the URI of that copy.
     * The file can safely be deleted afterwards.
     */
-  def storeImage(bucket: String, id: String, file: File, mimeType: Option[MimeType], meta: Map[String, String] = Map.empty)
+  def storeImage(bucket: String, id: String, file: File, mimeType: Option[MimeType],
+                 meta: Map[String, String] = Map.empty, overwrite: Boolean)
                 (implicit logMarker: LogMarker): Future[S3Object]
 
   def deleteImage(bucket: String, id: String): Future[Unit]
--- a/image-loader/app/model/Uploader.scala
+++ b/image-loader/app/model/Uploader.scala
@@ -146,6 +146,7 @@ object Uploader extends GridLogging {
     val eventualImage = for {
       browserViewableImage <- eventualBrowserViewableImage
       s3Source <- sourceStoreFuture
+      mergedUploadRequest = patchUploadRequestWithS3Metadata(uploadRequest, s3Source)
       optimisedFileMetadata <- FileMetadataReader.fromIPTCHeadersWithColorInfo(browserViewableImage)
       thumbViewableImage <- createThumbFuture(optimisedFileMetadata, colourModelFuture, browserViewableImage, deps)
       s3Thumb <- storeOrProjectThumbFile(thumbViewableImage)
@@ -165,7 +166,7 @@ object Uploader extends GridLogging {
       val thumbAsset = Asset.fromS3Object(s3Thumb, thumbDimensions)
 
       val pngAsset = s3PngOption.map(Asset.fromS3Object(_, sourceDimensions))
-      val baseImage = ImageUpload.createImage(uploadRequest, sourceAsset, thumbAsset, pngAsset, fullFileMetadata, metadata)
+      val baseImage = ImageUpload.createImage(mergedUploadRequest, sourceAsset, thumbAsset, pngAsset, fullFileMetadata, metadata)
 
       val processedImage = processor(baseImage)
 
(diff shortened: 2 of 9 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/ImageStorage.scala
```diff
@@ -34,7 +34,8 @@ trait ImageStorage {
   /** Store a copy of the given file and return the URI of that copy.
     * The file can safely be deleted afterwards.
     */
-  def storeImage(bucket: String, id: String, file: File, mimeType: Option[MimeType], meta: Map[String, String] = Map.empty)
+  def storeImage(bucket: String, id: String, file: File, mimeType: Option[MimeType],
```
AWare, line 37: Not 100% on calling this `storeImage` when it's default argument means it doesn't.

aug24 (approved):

common-lib/src/main/scala/com/gu/mediaservice/lib/ImageStorage.scala
aug24: It's default argument means it *does*.

aug24 (approved):

common-lib/src/main/scala/com/gu/mediaservice/lib/ImageStorage.scala
AWare: huh, I'm sure it said `false`

AWare (changes requested):
I'm concerned how this will affect the use case where someone without edit permissions reuploads an image to change it's metadata. If an image is reuploaded, a user might never know, and if they use the batch edit functionality, they might never know that their metadata change failed.

paperboyo:
Current behaviour overwrites the metadata on the s3 object, but **does not** update `x-amz-meta-upload_time` and `x-amz-meta-uploaded_by` in ES (it does update `x-amz-meta-file_name` in ES, though). (This means we may want to backfill these two for objects where ES contains an earlier date than s3).

Current behaviour keeps a reuploaded pic on the upload page only until a page reload and allows the (re)uploader to edit its metadata **only** there and **only** until the reload. It does not allow subsequent updates to the reuploded picture in the browser. As such, it may be considered a glitch. It is useful to see the image you reuploaded on your initial upload page to be able to find it, but to be consistent with our model and without overhauling it, we may want to disallow editing of the metadata there too (some UI to indicate that would be useful). We can also keep the current behaviour, if that’s easier, even though it does constitute a hole in the permissions model.

paperboyo:
> As this PR stands, whilst the metadata will not be replaced in S3 it will go on to be stored in ElasticSearch

Best behaviour, without overhauling the model (to store more than the original uploader and, maybe, more than the original upload date) is not to replace any of the existing data in ES. Also, as noted above, `x-amz-meta-upload_time` and `x-amz-meta-uploaded_by` aren’t overwritten in ES now.

common-lib/src/main/scala/com/gu/mediaservice/lib/ImageStorage.scala
sihil: defaulting a boolean to true is confusing to be fair. I think this is cleaner now.

AWare:
Good to see.

AWare (approved):

sihil:
In an incidental conversation with @akash1810 he seemed surprised that we'd want this behaviour so wonder whether he should also take a look.

image-loader/app/model/Uploader.scala
```diff
@@ -257,6 +258,16 @@ object Uploader extends GridLogging {
       case None => Future.failed(new Exception("This file is not an image with an identifiable mime type"))
     }
   }
+
+  def patchUploadRequestWithS3Metadata(request: UploadRequest, s3Object: S3Object): UploadRequest = {
```
aug24, line 262: This is nice and clear.

aug24 (approved):

akash1810:
> In an incidental conversation with @akash1810 he seemed surprised that we'd want this behaviour so wonder whether he should also take a look.

I might have misunderstood. We should always keep the first version of this metadata.

In the current version of Grid, if:
- Person A uploads an image on Monday the image appears in their "recent uploads"
- Person B uploads the same image on Tuesday, the image does not appear in their "recent uploads" as they're not attributed as the uploader (in ES at least)

paperboyo:
> We should always keep the first version of this metadata.

In ES we always did (apart from filename), but we were overwriting it in s3. Now, we will be keeping first version everywhere.

PR #3179, opened 2021-02-22
add config options to support loading arbitrary scripts (with permission support)

twrichards:
Co-authored-by: @tjsilver 

https://trello.com/c/TIeyuEff/511-1st-pinboard-integration-grid

## What does this change?
- New config option in `KahunaConfig`(i.e. `kahuna.conf`) entitled `scriptsToLoad` (list of objects) which allows the loading of arbitrary scripts (immediate use case is https://github.com/guardian/editorial-tools-pinboard) which optionally have a permission requirement on them, for example...
image

- Refactors to make the `Content-Security-Policy` more configurable by...
   - adding new config option `security.connectSources` (list of string) to control the `connect-src` portion of the CSP
   - re-using the 'hosts' from the new `scriptsToLoad` config option in the `script-src` portion of the CSP

## Tested?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
@@ -1,12 +1,12 @@
 package com.gu.mediaservice.lib.config
 
 import java.util.UUID
-
 import com.gu.mediaservice.lib.aws.{AwsClientBuilderUtils, KinesisSenderConfig}
-import com.typesafe.config.ConfigException
+import com.typesafe.config.{Config, ConfigException}
 import com.typesafe.scalalogging.StrictLogging
 import play.api.Configuration
 
+import scala.collection.JavaConverters.collectionAsScalaIterableConverter
 import scala.util.Try
 
 
@@ -66,6 +66,10 @@ abstract class CommonConfig(val configuration: Configuration) extends AwsClientB
     case _:ConfigException.WrongType => configuration.get[String](key).split(",").toSeq.map(_.trim)
   }.map(_.toSet)
    .getOrElse(Set.empty)
+
+  def getConfigList(key:String): List[_ <: Config] =
+    if (configuration.has(key)) configuration.underlying.getConfigList(key).asScala.toList
+    else List.empty
 
   final def apply(key: String): String =
     string(key)
--- a/kahuna/app/KahunaComponents.scala
+++ b/kahuna/app/KahunaComponents.scala
@@ -34,29 +34,43 @@ object KahunaSecurityConfig {
       config.services.guardianWitnessBaseUri
     )
 
+    val gaHost = "www.google-analytics.com"
+
     val frameSources = s"frame-src ${config.services.authBaseUri} ${config.services.kahunaBaseUri} https://accounts.google.com"
     val frameAncestors = s"frame-ancestors ${config.frameAncestors.mkString(" ")}"
-    val connectSources = s"connect-src ${(services :+ config.imageOrigin).mkString(" ")} 'self' www.google-analytics.com"
+    val connectSources = s"connect-src 'self' ${(services :+ config.imageOrigin).mkString(" ")} $gaHost ${config.connectSources.mkString(" ")}"
 
-    val imageSources: List[String] = List(
+    val imageSources = s"img-src ${List(
       "data:",
       "blob:",
       URI.ensureSecure(config.services.imgopsBaseUri).toString,
       URI.ensureSecure(config.fullOrigin).toString,
       URI.ensureSecure(config.thumbOrigin).toString,
       URI.ensureSecure(config.cropOrigin).toString,
-      URI.ensureSecure("www.google-analytics.com").toString,
+      URI.ensureSecure(gaHost).toString,
       URI.ensureSecure("app.getsentry.com").toString,
       "'self'"
-    )
+    ).mkString(" ")}"
 
     val fontSources = s"font-src data: 'self'"
+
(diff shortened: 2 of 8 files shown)


sihil:
Ah - totally understand our conversation the other day now. Remind me again, why is the pinboard client not being rolled into [gustaf](https://github.com/guardian/gustaf)?

twrichards:
> Ah - totally understand our conversation the other day now. Remind me again, why is the pinboard client not being rolled into [gustaf](https://github.com/guardian/gustaf)?

Because we didn't want to have to wait for changes to the extension to propagate through to people's browsers, when we release pinboard (currently around 3sec deploy ⚡️) it's instantly available for all users. We could consider moving the loader part to gustaf but it works neatly as is and the integration is pretty light touch and neatly all behind a permission.

Also worth reminding that there is no teleporter for grid (appreciate I haven't checked if it in fact still runs on grid and just doesn't render the Ⓣ icon because there's no unique ID).

twrichards:
@sihil do you think this change is relatively open-source friendly (i.e. we're not coupling pinboard and grid here, just making it possible to add it if desired)?

sihil:
It is loosely coupled and the open source element is well dealt with.

My concern stems from the subtle increase in the number of update sources that could break the grid. It is becomes feasible to make and deploy a change to pinboard that makes grid unusable. As a pattern it concerns me for the same reason that I dislike dependency definitions that are not pinned to a specific version: I'm spent too much time hunting down issues caused by a subtle change to a dependency. In short, there is a cost to making pinboard quick to deploy which is increasing the brittleness of the grid as a system - it's not as self contained.

Gustaf already falls into this category but is arguably a well known exception (presence does too, but in a rebuild I'd change it to be a normal library). For gustaf, it's easy to prove/disprove it being a problem by disabling the extension, although I guess we can block the pinboard script for the same debugging experience.

Whilst the integration is light touch, it's also light touch across all of the systems you want pinboard on. Gustaf (which does have [permission](https://github.com/guardian/gustaf/blob/master/extension/manifest.json#L18) could be even easier, although I don't think you can dynamically load JS into an extension so rollout would indeed be slower.

twrichards:
@sihil 

> It is loosely coupled and the open source element is well dealt with.

👍

> My concern stems from the subtle increase in the number of update sources that could break the grid. It is becomes feasible to make and deploy a change to pinboard that makes grid unusable. As a pattern it concerns me for the same reason that I dislike dependency definitions that are not pinned to a specific version: I'm spent too much time hunting down issues caused by a subtle change to a dependency. In short, there is a cost to making pinboard quick to deploy which is increasing the brittleness of the grid as a system - it's not as self contained.
> 
> Gustaf already falls into this category but is arguably a well known exception (presence does too, but in a rebuild I'd change it to be a normal library). For gustaf, it's easy to prove/disprove it being a problem by disabling the extension, although I guess we can block the pinboard script for the same debugging experience.

I would argue that if a problem is reported its less hassle for user for us to redeploy a safe version of Pinboard (sub 5 seconds) than have the user mess about disabling extensions, and in the process loosing Teleporter functionality, then waiting for a revert release (or fix-forward release) of gustaf to roll out.

> Whilst the integration is light touch, it's also light touch across all of the systems you want pinboard on. Gustaf (which does have [permission](https://github.com/guardian/gustaf/blob/master/extension/manifest.json#L18) could be even easier, although I don't think you can dynamically load JS into an extension so rollout would indeed be slower.

Pinboard itself already has permission integration as the main line of defence, but the permission stuff in this PR (and in https://github.com/guardian/flexible-content/pull/3706) is simply to avoid loading JS that isn't going to be allowed to run (and in the process keeps load on pinboard's `bootstrapping-lambda` to a minimum). I appreciate gustaf has permission built in already but pinboards current approach allows us to iterate quickly.

rest-lib/src/main/scala/com/gu/mediaservice/lib/auth/provider/AuthorisationProvider.scala
```diff
@@ -3,6 +3,7 @@ package com.gu.mediaservice.lib.auth.provider
 import com.gu.mediaservice.lib.auth.Authentication.Principal
 import com.gu.mediaservice.lib.auth.SimplePermission
 import com.gu.mediaservice.lib.config.{CommonConfig, Provider}
+import com.gu.permissions.PermissionDefinition
```
sihil, line 6: The Guardian permissions library has leaked into the abstraction here. This more or less defeats the point so I'm afraid I won't allow this. I suggest that as an alternative we create a temporary permission, perhaps `PinboardAssetsFeature` with an appropriate comment on it and an intention to get rid of it at some point in the future.

I plan to refactor the codebase in the coming months so that this sort of leak isn't possible (Guardian specific code lives in a separate sbt subproject so these common libraries can't see the dependency).

twrichards: my subconscious was telling me this might be problematic. happy to do that, as it reduces the complexity. Do you think the config model introduced in this PR should (for the time being) remove the notion of `app` so the `permission` field associated with a script in `scriptsToLoad` should just be a String (which will be checked against the `name` field of the `SimplePermission`?

sihil: Or even hardcoded to just work with pinboard for now. I do think we should pursue a proper feature switch system to solve all of these problems, although it might be that when we do bring permissions with params back in that might also work for this.

jonathonherbert (approved):
Code looks fine, LGTM and other commenters with more informed opinions re: specific parts have expressed approval out of band – 👍

Issue #3167, opened 2021-02-15
Make additional image fields visible on the UI


ochiengolanga:
## Short description
Currently there's a finite set of visible fields in the UI which are present in the `$.metadata` ([see ImageMetadata.scala](https://github.com/guardian/grid/blob/9588b1c2b5074866ffd11a3dd8176592271cb432/common-lib/src/main/scala/com/gu/mediaservice/model/ImageMetadata.scala#L10)) in the `GET image by id` JSON payload i.e. (_`dateTaken, description, credit, creditUri, byline, bylineTitle, title, copyright, supplier…`_). Extra useful metadata exists in the fileMetadata ([see  FileMetadata.scala](https://github.com/guardian/grid/blob/9588b1c2b5074866ffd11a3dd8176592271cb432/common-lib/src/main/scala/com/gu/mediaservice/model/FileMetadata.scala#L10)) fields stored in Elasticsearch as part of the Image document. However, the FileMetadata is currently not visible on the UI despite an API endpoint existing “/images/{:imageId}/fileMetadata” on the media-api service and the presence of some useful information in fileMetadata. 

This proposal introduces an approach to surface some of the metadata fields that have been extracted from the image and stored in Elasticsearch by making them visible through configuration with some UX considerations.

An admin / developer should be able to configure which fields from the fileMetadata they would like users to see, search by and specify a more human readable name of the metadata field aka an alias e.g. a user would prefer to see `Programme Number` rather than `xmp.bbc:ProgrammeNumber`.

NB: On the search, this proposal intends to improve the UX on the feature introduced by [BBC] Search on the value of arbitrary fields (https://github.com/guardian/grid/pull/3132) PR.

## Approach

1. Introduce a model that represents field aliases configuration with the following configuration parameters "`elasticsearchPath`", “`label`”, “`displaySearchHint`” and “`alias`” that's read from kahuna.conf and media-api.conf
2. Use the field aliases configuration to create a JSON object `aliases` with filemetadata fields that's added to the `GET image by id` JSON payload.
3. Introduce a “show image metadata” UI component that renders filemetadata fields in the `aliases` JSON object.

AWare:
2/3: can we store this in localStorage instead?

paperboyo:
The original issue and this 

> 2/3: can we store this in localStorage instead?

assume this is configurable by a user. Even if all users would be willing and able to do that, I think this is unnecessary and potentially can lead to chaos (lees so with different views of metadata, definitely with different aliases).

[...]

PR #3183, opened 2021-02-23
[BBC] Make additional fields visible on the UI via configuration

ochiengolanga:
## What does this change?
It introduces a way of rendering additional fields on the UI via configuration. This is achieved by adding the following configuration to Kahuna service and Media API service configuration file i.e. 

- `$USER/.grid/kahuna.conf` or `/etc/grid/kahuna.conf` 
- `$USER/.grid/media-api.conf` or `/etc/grid/media-api.conf` 

where applicable (dev stage or prod stage respectively).

```hocon
# Field aliases configurations that are used to define which metadata field is visible, searchable and has an alias on the UI
field.aliases = [
  {
    # Sets the image metadata tag to be used when filtering FileMetadata field
    elasticsearchPath = "fileMetadata.xmp.bbc:Programme-Maker"
    # Sets an alias for use in search, shouldn't have whitespace or special characters
    alias = "BBCProgrammeMaker"
    # Sets a human readable label that will rendered on the UI
    label = "BBC Programme Maker"
    # Sets the field alias to be displayed as a search hint when you type `+`
    displaySearchHint = false
  }
  {
    elasticsearchPath = "fileMetadata.xmp.xmp:CreatorTool"
    alias = "creatorTool"
    label = "Creator Tool"
    displaySearchHint = false
  }
  {
    elasticsearchPath = "fileMetadata.exif.Artist"
    alias = "artist"
    label = "Artist"
    displaySearchHint = true
  }
  {
    elasticsearchPath = "fileMetadata.iptc.Caption Writer/Editor"
    alias = "captionWriter"
    label = "Caption Writer / Editor"
    displaySearchHint = true
  }
  {
    elasticsearchPath = "fileMetadata.xmp.Iptc4xmpExt:DigitalSourceType"
    alias = "digitalSourceType"
    label = "Digital Source Type"
    displaySearchHint = false
  }
]
```

The approach is described in the proposal: https://github.com/guardian/grid/issues/3167

On the Image view page, the configured fields are visible on the “Show additional metadata” section when expanded.

## Screenshots

Screenshot at 2021-03-02 20-44-10

Screenshot at 2021-03-02 20-45-57

## Who should look at this?

@AWare , @sihil , @paperboyo and @akash1810 

## Tested?
- [x] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
@@ -9,7 +9,6 @@ import play.api.Configuration
 import scala.collection.JavaConverters.collectionAsScalaIterableConverter
 import scala.util.Try
 
-
 abstract class CommonConfig(val configuration: Configuration) extends AwsClientBuilderUtils with StrictLogging {
   final val stackName = "media-service"
 
@@ -57,6 +56,8 @@ abstract class CommonConfig(val configuration: Configuration) extends AwsClientB
   val corsAllowedOrigins: Set[String] = getStringSet("security.cors.allowedOrigins")
 
   val services = new Services(domainRoot, serviceHosts, corsAllowedOrigins)
+
+  val fieldAliasConfigs: Seq[FieldAlias] = configuration.get[Seq[FieldAlias]]("field.aliases")
 
   private def getKinesisConfigForStream(streamName: String) = KinesisSenderConfig(awsRegion, awsCredentials, awsLocalEndpoint, isDev, streamName)
 
--- a/kahuna/app/views/main.scala.html
+++ b/kahuna/app/views/main.scala.html
@@ -10,6 +10,7 @@
   usageRightsHelpLink: Option[String],
   invalidSessionHelpLink: Option[String],
   supportEmail: Option[String],
+  fieldAliases: String,
   scriptsToLoad: List[ScriptToLoad]
 )
 <!DOCTYPE html>
@@ -47,7 +48,8 @@
           feedbackFormLink: "@Html(feedbackFormLink.getOrElse(""))",
           usageRightsHelpLink: "@Html(usageRightsHelpLink.getOrElse(""))",
           invalidSessionHelpLink: "@Html(invalidSessionHelpLink.getOrElse(""))",
-          supportEmail: "@Html(supportEmail.getOrElse(""))"
+          supportEmail: "@Html(supportEmail.getOrElse(""))",
+          fieldAliases: @Html(fieldAliases)
         }
     </script>
 
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
@@ -272,7 +272,7 @@
                 ng-click="metadataExpanded = !metadataExpanded">
             <span ng-hide="metadataExpanded">▸ Show</span>
             <span ng-show="metadataExpanded">▾ Hide</span>
-            full metadata
+            additional metadata
         </button>
 
         <div ng-if="metadataExpanded" class="metadata metadata-line image-info__group--dl">
@@ -280,6 +280,12 @@
                 <dt class="metadata-line__key image-info__group--dl__key--full-metadata">{{key | spaceWords}}</dt>
                 <dd class="metadata-line__info image-info__group--dl__value--full-metadata">{{value}}</dd>
             </dl>
+
+            <dl ng-repeat="(key, value) in ctrl.additionalMetadata" class="metadata__body image-info__group--dl">
+                <dt class="metadata-line__key image-info__group--dl__key--full-metadata">{{key}}</dt>
+                <dd class="metadata-line__info image-info__group--dl__value--full-metadata">{{value}}</dd>
+            </dl>
+
             <dl ng-repeat="(key, value) in ctrl.identifiers" class="metadata__body image-info__group--dl">
                 <dt class="metadata-line__key image-info__group--dl__key--full-metadata">{{key | spaceWords}}</dt>
                 <dd class="metadata-line__info image-info__group--dl__value--full-metadata">{{value}}</dd>
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
@@ -33,8 +33,19 @@ module.controller('grImageMetadataCtrl', [
 
     let ctrl = this;
 
+    ctrl.fieldAliases = window._clientConfig.fieldAliases;
     ctrl.showUsageRights = false;
     ctrl.usageRights = imageService(ctrl.image).usageRights;
+    ctrl.additionalMetadata = Object.fromEntries(
+      Object.entries(ctrl.image.data.aliases)
+        .map(([key, val]) => {
+          let match = ctrl.fieldAliases.find(_ => _.alias === key);
+          if (match) {
+            return [match.label, val];
+          } else {
+            return [key, val];
+          }
+        }));
 
     // Alias for convenience in view
     ctrl.metadata = ctrl.image.data.metadata;
--- a/media-api/app/lib/ImageResponse.scala
+++ b/media-api/app/lib/ImageResponse.scala
@@ -1,7 +1,5 @@
 package lib
 
-import java.net.URI
-
 import com.gu.mediaservice.lib.argo.model._
 import com.gu.mediaservice.lib.auth.{Internal, Tier}
 import com.gu.mediaservice.lib.collections.CollectionsManager
@@ -10,12 +8,16 @@ import com.gu.mediaservice.model._
 import com.gu.mediaservice.model.leases.{LeasesByMedia, MediaLease}
 import com.gu.mediaservice.model.usage._
 import com.softwaremill.quicklens._
+import lib.ImageResponse.extractAliasFieldValues
+import lib.elasticsearch.SourceWrapper
 import lib.usagerights.CostCalculator
 import org.joda.time.DateTime
 import play.api.libs.functional.syntax._
 import play.api.libs.json._
 import play.utils.UriEncoding
 
+import java.net.URI
+import scala.annotation.tailrec
 import scala.util.{Failure, Try}
 
 class ImageResponse(config: MediaApiConfig, s3Client: S3Client, usageQuota: UsageQuota)
@@ -48,11 +50,13 @@ class ImageResponse(config: MediaApiConfig, s3Client: S3Client, usageQuota: Usag
 
   def create(
               id: String,
-              image: Image,
+              imageWrapper: SourceWrapper[Image],
               withWritePermission: Boolean,
               withDeleteImagePermission: Boolean,
               withDeleteCropsOrUsagePermission: Boolean,
               included: List[String] = List(), tier: Tier): (JsValue, List[Link], List[Action]) = {
+
+    val image = imageWrapper.instance
 
     val source = Try {
       Json.toJson(image)(imageResponseWrites(image.id, included.contains("fileMetadata")))
@@ -83,6 +87,8 @@ class ImageResponse(config: MediaApiConfig, s3Client: S3Client, usageQuota: Usag
     val persistenceReasons = imagePersistenceReasons(image)
     val isPersisted = persistenceReasons.nonEmpty
 
+    val aliases = extractAliasFieldValues(config, imageWrapper)
+
     val data = source.transform(addSecureSourceUrl(imageUrl))
       .flatMap(_.transform(wrapUserMetadata(id)))
       .flatMap(_.transform(addSecureThumbUrl(thumbUrl)))
@@ -95,7 +101,8 @@ class ImageResponse(config: MediaApiConfig, s3Client: S3Client, usageQuota: Usag
       .flatMap(_.transform(addInvalidReasons(invalidReasons)))
       .flatMap(_.transform(addUsageCost(source)))
       .flatMap(_.transform(addPersistedState(isPersisted, persistenceReasons)))
-      .flatMap(_.transform(addSyndicationStatus(image))).get
+      .flatMap(_.transform(addSyndicationStatus(image)))
+      .flatMap(_.transform(addAliases(aliases))).get
 
     val links: List[Link] = tier match {
       case Internal => imageLinks(id, imageUrl, pngUrl, withWritePermission, valid)
@@ -221,13 +228,17 @@ class ImageResponse(config: MediaApiConfig, s3Client: S3Client, usageQuota: Usag
   def addInvalidReasons(reasons: Map[String, String]): Reads[JsObject] =
     __.json.update(__.read[JsObject]).map(_ ++ Json.obj("invalidReasons" -> Json.toJson(reasons)))
 
+  def addAliases(aliases: Seq[(String, JsValue)]): Reads[JsObject] =
+    __.json.update(__.read[JsObject]).map(_ ++ Json.obj(
+      "aliases" -> JsObject(aliases)
+    ))
+
   def makeImgopsUri(uri: URI): String =
     config.imgopsUri + List(uri.getPath, uri.getRawQuery).mkString("?") + "{&w,h,q}"
 
   def makeOptimisedPngImageopsUri(uri: URI): String = {
     config.imgopsUri + List(uri.getPath, uri.getRawQuery).mkString("?") + "{&w, h, q}"
   }
-
 
   import play.api.libs.json.JodaWrites._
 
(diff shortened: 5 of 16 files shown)


ochiengolanga:
Sample image used:

sample-image

kahuna/app/lib/KahunaConfig.scala
```diff
@@ -21,4 +23,22 @@ class KahunaConfig(resources: GridConfigResources) extends CommonConfig(resource
   val supportEmail: Option[String]= stringOpt("links.supportEmail").filterNot(_.isEmpty)
 
   val frameAncestors: Set[String] = getStringSet("security.frameAncestors")
+  val fileMetadataConfigs: List[FileMetadataConfig] = configObjectOpt("filemetadata.configurations") match {
+    case Some(config) => config.unwrapped()
+      .asInstanceOf[java.util.Map[String, java.util.HashMap[String, java.util.HashMap[String, Object]]]]
```
AWare, line 28: Can we extract them without using `asInstanceOf`?

ochiengolanga: No, from the structure of the HOCON config with dynamic paths. More than happy to get any pointers on how to achieve this without casting.

ochiengolanga: The cast `.asInstanceOf[java.util.Map[String, java.util.HashMap[String, java.util.HashMap[String, Object]]]]` can be cleaner by casting to `ConfigObject` instead i.e. `asInstanceOf[ConfigObject]`

sihil:
We'll do a code review shortly but wanted to put this here as a general comment on the structure of the configuration.

We think that the configuration can be more generic. Specifically:
- Rather than have discreet objects for each metadata type it can be only one
- `elasticsearchPath` replaces `tag` and contains the full path instead of the suffix (means that it can be used for any field in ES)
- the `alias` is used in searching and chips
- `label` is introduced to be the human readable name
- The whole config becomes an array instead of an object as the keys are redundant and it means that ordering in the UI can be derived from the order here

Taking your example:
- `xmp.bbc:Programme-Maker` as Visible and Searchable with an alias “BBCProgrammeMaker”
- `xmp.aux:Lens` as Visible with an alias “AuxLens”
- `xmp.CreatorTool` as visible with an alias “Creator Tool”
- `exif:Artist` as visible and searchable with an alias “Artist”

```hocon
# FileMetadata configurations that are used to define which metadata field is visible, searchable and has an alias on the UI
fieldAliases [
  {
    # Sets the image metadata tag to be used when filtering FileMetadata field
    elasticsearchPath = "fileMetadata.xmp.bbc:Programme-Maker"
    # Sets an alias for use in search, shouldn't have whitespace or special characters
    alias = "BBCProgrammeMaker"
    # Sets a human readable label that will rendered on the UI
    label = "BBC Programme Maker"
    # Sets the field alias to be displayed as a search hint when you type `+`
    displaySearchHint = false
  },
  {
    elasticsearchPath = "fileMetadata.xmp.aux:Lens"
    alias = "auxLens"
    label = "Aux Lens"
    displaySearchHint = false
  },
  {
    elasticsearchPath = "fileMetadata.xmp.CreatorTool"
    alias = "creatorTool"
    label = "Creator Tool"
    displaySearchHint = false
  },
  {
    elasticsearchPath = "fileMetadata.exif.Artist"
    alias = "artist"
    label = "Artist"
    displaySearchHint = true
  }
]
```

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
```diff
@@ -40,6 +40,28 @@ module.controller('grImageMetadataCtrl', [
     ctrl.metadata = ctrl.image.data.metadata;
     ctrl.identifiers = ctrl.image.data.identifiers;
 
+    mediaApi.fileMetadata(ctrl.image).then(resource => {
```
sihil, line 43: We probably don't want to read `fileMetadata` as a matter of course. This can be in excess of 1Mb in many cases and we'll only ever want a tiny fraction of the content. Either the field aliases should be automatically added to the main image endpoint or a new endpoint that only serves those values can be created.

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
```diff
@@ -40,6 +40,28 @@ module.controller('grImageMetadataCtrl', [
...
+      return resource.data;
+    }).then(fileMetadata => {
+      const groupedConfigsByDirectory = window._clientConfig.fileMetadataConfig.reduce((r, a) => {
```
sihil, line 46: Rather than read this here every time it is used it should be parsed once in a controller. e.g. https://github.com/guardian/grid/blob/3b03881f9aa9a6331bb34c99967712e842540998/kahuna/public/js/errors/global.js#L44

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
```diff
@@ -40,6 +40,28 @@ module.controller('grImageMetadataCtrl', [
...
+      let object = {};
+
+      Object.keys(groupedConfigsByDirectory).forEach(directory => {
+        groupedConfigsByDirectory[directory]
+          .filter(config => config.visible)
+          .filter(fileMetadataConfig => fileMetadata[directory][fileMetadataConfig.tag] !== undefined)
+          .map(fileMetadataConfig => {
+            object[fileMetadataConfig.alias] = fileMetadata[directory][fileMetadataConfig.tag];
+          });
+      });
```
sihil, line 60: A cleaner pattern for this (once flattened to remove the directories) uses `Object.fromEntries`. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries#object_transformations

common-lib/src/main/scala/com/gu/mediaservice/lib/config/FileMetadataConfig.scala
```diff
@@ -0,0 +1,21 @@
...
+
+object FileMetadataConfig {
+  import play.api.libs.json._
+  implicit val MetadataConfigurationWrites: Writes[FileMetadataConfig] = (
+    (__ \ "directory").write[String] ~
+      (__ \ "tag").write[String] ~
+      (__ \ "visible").write[Boolean] ~
+      (__ \ "searchable").write[Boolean] ~
+      (__ \ "alias").writeNullable[String]
+    )(unlift(FileMetadataConfig.unapply))
```
sihil, line 20: Think you can probably use a macro here?
```suggestion
  implicit val MetadataConfigurationWrites: Writes[FileMetadataConfig] = Json.writes[FileMetadataConfig]
```

common-lib/src/main/scala/com/gu/mediaservice/lib/config/FileMetadataConfig.scala
```diff
@@ -0,0 +1,21 @@
...
+
+case class FileMetadataConfig(
+                               directory: String,
+                               tag: String,
+                               visible: Boolean = false,
+                               searchable: Boolean = false,
+                               alias: Option[String])
+
+object FileMetadataConfig {
+  import play.api.libs.json._
```
sihil, line 13: Idiomatically this import belongs at the top of the file.

sihil (commented):
A few thoughts for you to look at. We're very happy to run through in a call if you have any questions or want to discuss further.

common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
```diff
@@ -94,6 +92,8 @@ abstract class CommonConfig(val configuration: Configuration) extends AwsClientB
   final def boolean(key: String): Boolean =
     configuration.getOptional[Boolean](key).getOrElse(false)
 
+  final def configObjectOpt(key : String) : Option[ConfigObject] = configuration.getOptional[ConfigObject](key)
```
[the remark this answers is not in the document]
ochiengolanga: Thanks @sihil. Definitely a cleaner implementation now

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
```diff
@@ -272,14 +272,20 @@
                 ng-click="metadataExpanded = !metadataExpanded">
             <span ng-hide="metadataExpanded">▸ Show</span>
             <span ng-show="metadataExpanded">▾ Hide</span>
-            full metadata
+            image metadata
```
sihil, line 275: What's the motivation behind this change? Looking at the UI there is already metadata on display above this so show image metadata doesn't seem to make as much sense. I imagine that the objection of full is that there is more metadata not being displayed?

image

kahuna/app/views/main.scala.html
```diff
@@ -43,7 +44,8 @@
           feedbackFormLink: "@Html(feedbackFormLink.getOrElse(""))",
           usageRightsHelpLink: "@Html(usageRightsHelpLink.getOrElse(""))",
           invalidSessionHelpLink: "@Html(invalidSessionHelpLink.getOrElse(""))",
-          supportEmail: "@Html(supportEmail.getOrElse(""))"
+          supportEmail: "@Html(supportEmail.getOrElse(""))",
+          fileMetadataConfig: @Html(fileMetadataConfigs)
```
sihil, line 48: I'm somewhat apprehensive of having a field here that isn't escaped but it looks like we have a number of other examples and it isn't coming from the user so it's relatively safe.

kahuna/public/js/services/api/media-api.js
```diff
@@ -66,6 +66,10 @@ mediaApi.factory('mediaApi',
         return root.follow('image', {id: id}).get();
     }
 
+    function fileMetadata(image, {include} = {}) {
+      return image.follow('fileMetadata', { include: include }).get();
+    }
```
sihil, line 71: do we need this anymore now we're not fetching `fileMetadata`?

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
ochiengolanga: Removed this bit that's now irrelevant. The visible fileMetadata fields are accessible within `ctrl.image` variable hence no need to fetch `fileMetadata` via the `mediaApi`

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
ochiengolanga: Removed this bit that's now irrelevant. The visible fileMetadata fields are accessible within `ctrl.image`

media-api/app/lib/ImageResponse.scala
```diff
@@ -320,6 +324,28 @@ object ImageResponse {
...
+      }
+    }
+
+    val aliases: Seq[(String, JsValue)] = config.fieldAliasConfigs.flatMap { config =>
+      val parts = config.elasticsearchPath.split('.').toList.filter(_.nonEmpty)
+      val lookupResult = nestedLookup(JsDefined(source.source), parts)
+      lookupResult.toOption.map {
+        config.alias -> _
+      }
+    }
```
sihil, line 343: Fairly minor but it would be pleasant to extract this into another method so that this business logic isn't cluttering this function. Perhaps something like `extractAliasFieldValues(config: MediaApiConfig, source: SourceWrapper[Image]): Seq[(String, JsValue)]` on the `ImageResponse` object.

In fact `addAliases` then becomes `addAliases(aliases: Seq[(String, JsValue)])` and the aliases are actually calculated at the call site around line 89.

sihil (commented):
Morning @ochiengolanga. This PR is in good shape and very close to being ready. Comments are mainly tidying up this time around. With a little refactor of the `ImageResponse` code. Let me know if you have any questions.

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
paperboyo: I agree with @sihil. If `full` is no good, maybe we can go with `more`?

common-lib/src/main/scala/com/gu/mediaservice/model/FileMetadata.scala
```diff
@@ -1,11 +1,9 @@
 package com.gu.mediaservice.model
 
 import com.gu.mediaservice.lib.logging.{LogMarker, MarkerMap}
-import net.logstash.logback.marker.Markers
-import play.api.libs.json._
 import play.api.libs.functional.syntax._
-
-import scala.collection.JavaConverters._
```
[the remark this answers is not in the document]
ochiengolanga: Just optimizing imports here

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
ochiengolanga: This change was inspired by a suggestion on the [proposal](https://github.com/guardian/grid/issues/3167) to change from "Show full metadata" to "Show image metadata".

paperboyo: My suggestion makes no sense: `Hide more metadata` sounds idiotic. Akash, Alex and me think `additional` will work and is better than `full` (not really true).

ochiengolanga: Changing the text to "Show additional metadata" and "Hide additional metadata". Thanks @paperboyo

common-lib/src/main/scala/com/gu/mediaservice/model/FileMetadata.scala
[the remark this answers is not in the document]
sihil: Understood but I'd avoid doing so in a file you have not otherwise touched.

media-api/test/lib/ImageResponseTest.scala
```diff
@@ -1,11 +1,15 @@
...
+import com.gu.mediaservice.model._
 import com.gu.mediaservice.model.usage.{PendingUsageStatus, PrintUsage, Usage}
-import com.gu.mediaservice.model.{Bounds, Crop, CropSpec}
+import lib.elasticsearch.{ElasticSearchTestBase, SourceWrapper}
 import org.joda.time.DateTime.now
+import org.scalatest.Inside.inside
 import org.scalatest.{FunSpec, Matchers}
+import play.api.libs.json._
+
+class ImageResponseTest extends FunSpec with Matchers with ElasticSearchTestBase {
```
sihil, line 11: Unless I've misunderstood this test no longer uses the ElasticSearchTestBase?

ochiengolanga: Refactored `ImageResponseTest` and `ElasticSearchTestBase` to resolve this.

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
```diff
@@ -33,8 +33,20 @@ module.controller('grImageMetadataCtrl', [
...
     ctrl.showUsageRights = false;
     ctrl.usageRights = imageService(ctrl.image).usageRights;
+    ctrl.additionalMetadata = Object.fromEntries(
+      Object.entries(ctrl.image.data.aliases)
+        .map(([key, val]) => {
+          let match = ctrl.fieldAliases.find(_ => _.alias === key);
+          if (match) {
+            return [match.label, val];
+          } else {
+            return undefined;
```
sihil, line 46: Would it be better to fall back to `[key, val]` here? we're not expecting it to happen, but it would IMHO be better to show an alias instead of a label than drop the field completely.

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
```diff
@@ -33,8 +33,20 @@ module.controller('grImageMetadataCtrl', [
...
+          }
+        })
+        .filter(_ => _ !== undefined));
```
sihil, line 49: If you takethe suggestion above then this filter can also go.

common-lib/src/main/scala/com/gu/mediaservice/model/Image.scala
```diff
@@ -127,4 +127,3 @@ object Image {
     )(unlift(Image.unapply))
 
 }
-
```
sihil, line 130: Last bit of noise here.

sihil (approved):
This is a really nice piece of additional functionality. Thanks @ochiengolanga!

This is fundamentally ready to go. I left a couple of minor comments but I'd be OK merging it as it is.

Please do go ahead and squash your commits and then I'll look to merge shortly after that.

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
ochiengolanga: Definitely makes sense to fallback to alias, thanks!

sihil:
Hi @ochiengolanga,
Have confirmed that this squash is good and will merge in the morning. Thanks again for this PR!

sihil (approved):
Tested at Guardian towers. Two minor points around the labels losing their casing and formatting of array values but we'll 🚢 this and raise an issue to resolve those.

image

PR #3185, opened 2021-02-24
Fix REST lib tests

aug24:
## What does this change?
When REST-lib was separated out from common-lib, the new library was not added to the root aggregation, and so the tests for REST services did not run.

This PR fixes that regression.

REST tests run and pass.

## Tested?
- [ ] locally by committer


sihil (approved):
Assuming all the tests now pass 👍

PR #3186, opened 2021-02-24
feature/bbc-pluggable-project: Implement pluggable bbc project mechanism

adrielulanovsky:
## What does this change?
These changes provide BBC a way to have their own specific project, allowing them to use different versions of other libraries or have code that wouldn't otherwise compile on the Guardian version. The bbcLib project is only compiled if the BUILD_ORG environment variable is set to "bbc".

BBC grid works successfully in BBC environments, while not affecting Guardian's Grid behaviour or compilation in any way.

## Who should look at this?
@sihil , @wainaina , @gribeiro

## Tested?
- [x] locally by committer

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -26,8 +26,11 @@ val commonSettings = Seq(
   publishArtifact in (Compile, packageDoc) := false
 )
 
+//Common projects to all organizations
+lazy val commonProjects: Seq[sbt.ProjectReference] = Seq(commonLib, restLib, auth, collections, cropper, imageLoader, leases, thrall, kahuna, metadataEditor, usage, mediaApi, adminToolsLambda, adminToolsScripts, adminToolsDev)
+
 lazy val root = project("grid", path = Some("."))
-  .aggregate(commonLib, restLib, auth, collections, cropper, imageLoader, leases, thrall, kahuna, metadataEditor, usage, mediaApi, adminToolsLambda, adminToolsScripts, adminToolsDev)
+  .aggregate((maybeBBCLib.toList ++ commonProjects):_*)
   .enablePlugins(RiffRaffArtifact)
   .settings(
     riffRaffManifestProjectName := s"media-service::grid::all",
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -67,20 +70,16 @@ val okHttpVersion = "3.12.1"
 
 val bbcBuildProcess: Boolean = System.getenv().asScala.get("BUILD_ORG").contains("bbc")
 
-val bbcCommonLibSettings: SettingsDefinition = if (bbcBuildProcess) {
-  Seq(
-    libraryDependencies ++= Seq("com.gu" %% "pan-domain-auth-play_2-6" % "0.9.2-SNAPSHOT")
```
AWare, line 72: Has using `0.9.2` caused any problems for you?

adrielulanovsky: No, but our new authentication provider doesn't require pan-domain-auth any more, so there's no need to have a bbc specific version of it

AWare: Just earmarking it as a potentially easy target for us to update.

build.sbt
```diff
@@ -289,8 +288,8 @@ val buildInfo = Seq(
   )
 )
 
-def playProject(projectName: String, port: Int, path: Option[String] = None): Project =
-  project(projectName, path)
+def playProject(projectName: String, port: Int, path: Option[String] = None): Project = {
+  lazy val commonProject = project(projectName, path)
```
sihil, line 292: Does this need to be `lazy`?

build.sbt
```diff
@@ -317,3 +316,6 @@ def playProject(projectName: String, port: Int, path: Option[String] = None): Pr
         "-J-XX:GCLogFileSize=2M"
       )
     ))
+  //Add the BBC library dependency if defined
+  maybeBBCLib.fold(commonProject){bbcLib => commonProject.dependsOn(bbcLib) }
```
sihil, line 320:
```suggestion
  maybeBBCLib.fold(commonProject)(commonProject.dependsOn(_))
```

build.sbt
```diff
@@ -67,20 +70,16 @@ val okHttpVersion = "3.12.1"
...
-  )
-} else {
-  Seq(
-    libraryDependencies ++= Seq("com.gu" %% "pan-domain-auth-play_2-6" % "0.8.2")
-  )
-}
+//BBC specific project, it only gets compiled when bbcBuildProcess is true
+lazy val bbcProject = project("bbc").dependsOn(restLib)
+
+lazy val maybeBBCLib: Option[sbt.ProjectReference] = if(bbcBuildProcess) Some(bbcProject) else None
```
sihil, line 76: also probably doesn't really need to be lazy.

build.sbt
```diff
@@ -67,20 +70,16 @@ val okHttpVersion = "3.12.1"
...
-  Seq(
-    libraryDependencies ++= Seq("com.gu" %% "pan-domain-auth-play_2-6" % "0.9.2-SNAPSHOT")
-  )
-} else {
-  Seq(
-    libraryDependencies ++= Seq("com.gu" %% "pan-domain-auth-play_2-6" % "0.8.2")
-  )
-}
+//BBC specific project, it only gets compiled when bbcBuildProcess is true
+lazy val bbcProject = project("bbc").dependsOn(restLib)
```
sihil, line 74: This on the otherhand does need to be lazy, as explained.

sihil (commented):
Happy with this to go in. We'll likely do something similar very soon.

A couple of comments but we should be able to get this merged soon I expect.

build.sbt
adrielulanovsky: Fixed

build.sbt
adrielulanovsky: Fixed

build.sbt
adrielulanovsky: Fixed

sihil (approved):
Thanks for your quick fixes @adrielulanovsky. I've verified that the build works for us and have tested this in our testing environment.

PR #3187, opened 2021-02-24
Jr and aw/convert grid client to play ws

aug24:
## What does this change?
Long term, we intend for Thrall to fetch content rather than blindly trust the content of the thrall message.  Similarly, we have re-projection work which requires some services to talk to other services.

The existing GridClient code is based on OK Http which adds a library (and therefore complexity) when Play's own WS library is _right there_.

This PR converts the client to WS and simplifies the interface as far as possible.

No change to behaviour.  Tests all pass.

## Tested?
- [ ] locally by committer


admin-tools/lib/src/main/scala/com/gu/mediaservice/ImageDataMerger.scala
```diff
@@ -147,42 +157,44 @@ class ImageDataMerger(config: ImageDataMergerConfig) extends LazyLogging {
...
-    val maybeImage: Option[Image] = gridClient.getImageLoaderProjection(mediaId, imageLoaderEndpoint)
-    maybeImage match {
-      case Some(img) =>
-        aggregate(img).map { aggImg =>
-          Some(ImageProjectionOverrides.overrideSelectedFields(aggImg))
-        }
-      case None => Future(None)
-    }
+  private def getMergedImageDataInternal(mediaId: String)(implicit ec: ExecutionContext): Future[Option[Image]] = for {
+    maybeImage <- gridClient.getImageLoaderProjection(mediaId, imageLoaderEndpoint, None)
```
AWare, line 161: as this isn't a "full" image, do we want to call it a maybeStubImage?

admin-tools/lib/src/main/scala/com/gu/mediaservice/ImagesBatchProjection.scala
```diff
@@ -4,49 +4,60 @@ import java.net.URL
...
-import scala.concurrent.duration.Duration
+import scala.concurrent.duration.{DAYS, Duration, HOURS}
 import scala.concurrent.{Await, ExecutionContext, Future}
 
 class ImagesBatchProjection(apiKey: String, timeout: Duration, gridClient: GridClient, maxSize: Int) {
 
   private implicit val ThrottledExecutionContext = ExecutionContext.fromExecutor(java.util.concurrent.Executors.newFixedThreadPool(5))
 
-  def validApiKey(projectionEndpoint: String) = {
+  // TODO I don't like this - checking the API key by making a request to a non-existent endpoint is ... ewww.
```
AWare, line 14: This is done differently in  https://github.com/guardian/grid/pull/3187/files#diff-8714f6045b0cd0fce66f948140ad54deff14c89eab51ed1597d657111207c337R28 

(Personally not sure the client wants a validate method and should instead just log the first http request that goes awry for this)

admin-tools/lib/src/main/scala/com/gu/mediaservice/ImagesBatchProjection.scala
```diff
@@ -4,49 +4,60 @@ import java.net.URL
...
           }
-          case _ if (isAKnownError(response)) => {
+        },
+        Some({ res: ResponseWrapper => {
+          if (isAKnownError(res)) {
             InputIdsStore.setStateToKnownError(id)
-            None
-          }
-          case _ => {
+            new FailedCallException()
```
AWare, line 53: Does this want to take the id for better logging?

common-lib/src/main/scala/com/gu/mediaservice/GridClient.scala
```diff
@@ -132,73 +121,88 @@ class GridClient(apiKey: String, services: Services, maxIdleConnections: Int, de
...
+                                   errorFn: Option[ResponseWrapper => Exception]
+                                 ): Option[T] = {
+    val res = processResponse(response, url)
+    res.statusCode match {
+      case 200 => foundFn(res)
+      case 404 => notFoundFn(res)
+      case failCode => errorFn match {
+        case Some(fn) => throw fn(res)
+        case _ =>
+          val errorMessage = s"breaking the circuit of full image projection, downstream API: $url is in a bad state, code: $failCode"
```
AWare, line 139: This message isn't super relevant any more.

image-loader/test/scala/model/ProjectorTest.scala
```diff
@@ -117,37 +123,62 @@ class ProjectorTest extends FreeSpec with Matchers with ScalaFutures with Mockit
...
+        ),
       originalMetadata = ImageMetadata(
         Some(new DateTime("2015-01-22T00:00:00.000Z").withZone(DateTimeZone.UTC)),
         Some("Austria's Matthias Mayer attends the men's downhill training of the FIS Alpine Skiing World Cup in Kitzbuehel, Austria, on January 22, 2015.       AFP PHOTO / CHRISTOF STACHECHRISTOF STACHE/AFP/Getty Images"),
-        Some("AFP/Getty Images"), None, Some("Christof Stache"), Some("Stringer"),
-        None, Some("CHRISTOF STACHE"), Some("-"),
-        Some("AFP"), None, Nil, None, Some("Kitzbuehel"), None, Some("Austria"), List("sport")),
-      usageRights = Agency("Getty Images", Some("AFP"), None),
-      originalUsageRights = Agency("Getty Images", Some("AFP"), None),
+        Some("AFP/Getty Images"),
```
AWare, line 165: can we do the named fields refactor to this?

admin-tools/dev/app/controllers/AdminToolsCtr.scala
```diff
@@ -6,13 +6,23 @@ import com.gu.mediaservice.model.Image._
...
+class AdminToolsCtr(config: AdminToolsConfig, override val controllerComponents: ControllerComponents, wSClient: WSClient)(implicit val ec: ExecutionContext)
   extends BaseController with ArgoHelpers {
 
+  import akka.actor.ActorSystem
+  import akka.stream.ActorMaterializer
+  import play.api.libs.ws._
+
+  implicit val system = ActorSystem()
+  implicit val materializer = ActorMaterializer()
+  implicit val ws:WSClient  = AhcWSClient()
```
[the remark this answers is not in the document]
aug24: implicitness, which is why I didn't spot that it already existed.  Have amended.

admin-tools/lib/src/main/scala/com/gu/mediaservice/ImageDataMerger.scala
aug24: Reluctantly agreed :)

admin-tools/lib/src/main/scala/com/gu/mediaservice/ImagesBatchProjection.scala
aug24: Not needed, it is caught locally.

common-lib/src/main/scala/com/gu/mediaservice/GridClient.scala
aug24: Agreed.

image-loader/test/scala/model/ProjectorTest.scala
aug24: Done

admin-tools/lib/src/main/scala/com/gu/mediaservice/ImagesBatchProjection.scala
aug24: imageloader calls are implemented differently.  I don't know why.  The  imageloaderendpoint is provided explicitly, for example.

AWare: ah so we can't use the endpoint from there :( I see

AWare (approved):

sihil:
Happy to review this properly early next week if helpful. As discussed today, authFn (currently an option) can simply be `identity` if there is no user. But it should probably be initialised to either be onbehalfof when using internally or to attach the API key when initialised for a CLI or similar.

image-loader/app/lib/Downloader.scala
```diff
@@ -17,42 +17,40 @@ case object TruncatedDownload extends Exception
...
-        logger.error(s"Missing content-length header from $uri", exception)
-        throw InvalidDownload
-      }
-      case Success(expectedSize) => {
-        val input = response.body().byteStream()
-
-        val output = new FileOutputStream(file)
-        val hashedOutput = new HashingOutputStream(digester, output)
+    maybeExpectedSize = Try {
+      response.header("Content-Length").map(_.toInt)
```
sihil, line 27: Constants are awesome: `HeaderNames.CONTENT_LENGTH`

image-loader/app/lib/Downloader.scala
```diff
@@ -17,42 +17,40 @@ case object TruncatedDownload extends Exception
...
+      hashedOutput.close()
 
-        if (actualSize != expectedSize) {
-          throw TruncatedDownload
-        }
+      val actualSize = Files.size(file.toPath)
 
-        DigestedFile(file, hash)
+      if (actualSize != expectedSize) {
+        throw TruncatedDownload
```
sihil, line 51: should this delete the file since it is corrupt?

image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -136,9 +137,8 @@ class ImageLoaderController(auth: Authentication,
           respondError(NotFound, "image-not-found", s"Could not find image: $imageId in s3 at $s3Path")
       } recover {
         case _: NoSuchImageExistsInS3 => NotFound(Json.obj("imageId" -> imageId))
-        case _ => InternalServerError(Json.obj("imageId" -> imageId))
+        case e => InternalServerError(Json.obj("imageId" -> imageId, "exception" -> e.getMessage))
```
sihil, line 140: I think there is a standard format for errors, although it seems to be fairly poorly applied so perhaps there should be one but currently isn't. I note that `getMessage` loses a lot of context such as the exception type (super useful for context) and stack trace (nice to have).

common-lib/src/main/scala/com/gu/mediaservice/GridClient.scala
```diff
@@ -132,73 +121,84 @@ class GridClient(apiKey: String, services: Services, maxIdleConnections: Int, de
...
+    makeGetRequestAsync(
+      url,
+      authFn,
+      {res:ResponseWrapper => Some(res.body.as[Image])},
+      {_:ResponseWrapper => None}
+    )
   }
 
-  def getLeases(mediaId: String)(implicit ec: ExecutionContext): Future[LeasesByMedia] = {
+  def getLeases(mediaId: String, authFn: WSRequest => WSRequest)(implicit ec: ExecutionContext): Future[Option[LeasesByMedia]] = {
```
sihil, line 168: For all of these endpoints I think that it is possible to eliminate the `Option` from the return type as None is never used to signify anything useful. Future can be `Failure` if necessary, and probably should be as if there is an upstream failure. 

I've not yet found a reason this needs to be option. I think the superflous option stems from the functions that are passed into the request function (discussed above) - the suggested refactor will eliminate this but even without that you can remove Option from all of the functions.

sihil (changes requested):
I know that some of the things I've called out were inherited but it would be great to make some of them betterer 😄

image-loader/app/lib/Downloader.scala
aug24: No, it's always deleted in the calling code during onComplete.

admin-tools/dev/app/controllers/AdminToolsCtr.scala
```diff
@@ -6,13 +6,16 @@ import com.gu.mediaservice.model.Image._
...
 import lib.AdminToolsConfig
 import play.api.libs.json.Json
+import play.api.libs.ws.WSClient
+import play.api.libs.ws.ahc.AhcWSClient
 import play.api.mvc.{BaseController, ControllerComponents}
 
 import scala.concurrent.ExecutionContext
 
-class AdminToolsCtr(config: AdminToolsConfig, override val controllerComponents: ControllerComponents)(implicit val ec: ExecutionContext)
+class AdminToolsCtr(config: AdminToolsConfig, override val controllerComponents: ControllerComponents, wSClient: WSClient)(implicit val ec: ExecutionContext)
```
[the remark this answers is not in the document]
aug24: Cleaned up.

common-lib/src/main/scala/com/gu/mediaservice/GridClient.scala
aug24: Done.   Took a little thinking :)

admin-tools/lambda/src/main/scala/com/gu/mediaservice/ImageProjectionLambdaHandler.scala
```diff
@@ -28,12 +41,15 @@ class ImageProjectionLambdaHandler extends ApiKeyAuthentication with LazyLogging
     apiKey match {
       case Some(key) =>
         val cfg: ImageDataMergerConfig = ImageDataMergerConfig(apiKey = key, domainRoot = domainRoot, imageLoaderEndpointOpt = imageLoaderEndpoint)
-        if (!cfg.isValidApiKey()) return getUnauthorisedResponse
+        val merger = new ImageDataMerger(cfg)
+
+        for {
+          ok <- merger.isValidApiKey
+        } yield if (!ok) return getUnauthorisedResponse
```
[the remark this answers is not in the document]
aug24: Removed.

common-lib/src/main/scala/com/gu/mediaservice/GridClient.scala
```diff
@@ -132,73 +121,84 @@ class GridClient(apiKey: String, services: Services, maxIdleConnections: Int, de
...
-        )
-      )
-      logger.error(errorJson.toString())
-      throw new DownstreamApiInBadStateException(errorMessage, downstreamErrorMessage)
+  private def validateResponse[T](
+                                   response: WSResponse,
+                                   url: URL,
+                                   foundFn: ResponseWrapper => Option[T],
+                                   notFoundFn: ResponseWrapper => Option[T],
+                                   errorFn: Option[ResponseWrapper => Exception]
```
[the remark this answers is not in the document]
aug24: Done.  Definitely better.

image-loader/app/model/Projector.scala
```diff
@@ -130,7 +141,23 @@ class Projector(config: ImageUploadOpsCfg,
           identifiers = identifiers_,
           uploadInfo = uploadInfo_
         )
-        imageUploadProjectionOps.projectImageFromUploadRequest(uploadRequest)
+
+        val onBehalfOfFn: OnBehalfOfPrincipal = user
+          .map(u => auth.getOnBehalfOfPrincipal(u))
+          .getOrElse( (r:WSRequest) => r)
```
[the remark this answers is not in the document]
aug24: Yep.  Done (passed in `identity` from ProjectTest as the method now accepts the auth function).

image-loader/app/model/Projector.scala
```diff
@@ -110,7 +117,11 @@ class Projector(config: ImageUploadOpsCfg,
     DigestedFile(tempFile, imageId)
   }
 
-  def projectImage(srcFileDigest: DigestedFile, extractedS3Meta: S3FileExtractedMetadata, requestId: UUID)
+  def projectImage(srcFileDigest: DigestedFile,
+                   extractedS3Meta: S3FileExtractedMetadata,
+                   requestId: UUID,
+                   gridClient: GridClient,
+                   user: Option[Principal])
```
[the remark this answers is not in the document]
aug24: Done.

image-loader/test/scala/model/ProjectorTest.scala
```diff
@@ -3,47 +3,57 @@ package model
...
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.collection.JavaConverters._
+import scala.concurrent.Future
 
 class ProjectorTest extends FreeSpec with Matchers with ScalaFutures with MockitoSugar {
 
-  import ResourceHelpers.fileAt
+  import ResourceHelpers._
+
+  private def noAuthFn = (r:WSRequest) => r
```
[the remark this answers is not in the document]
aug24: Removed.

admin-tools/lib/src/main/scala/com/gu/mediaservice/BatchIndexHandler.scala
```diff
@@ -124,7 +125,7 @@ class BatchIndexHandler(cfg: BatchIndexHandlerConfig) extends LoggingWithMarkers
   }
 
   def processImagesOnlyIfKinesisIsNiceAndFast(): Unit = {
-    if (!validApiKey(projectionEndpoint)) throw new IllegalStateException("invalid api key")
+    validateApiKey(projectionEndpoint)
```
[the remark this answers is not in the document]
aug24: Done.

common-lib/src/main/scala/com/gu/mediaservice/GridClient.scala
```diff
@@ -132,73 +94,91 @@ class GridClient(apiKey: String, services: Services, maxIdleConnections: Int, de
...
-    makeGetRequestAsync(url, apiKey).map { res =>
-      validateResponse(res, url)
-      if (res.statusCode == 200) (res.body \ "data").as[LeasesByMedia] else LeasesByMedia.empty
+    for {
+      response <- makeGetRequestAsync(url, authFn)
+    } yield response match {
+      case Found(json, _) => (json \ "data").as[LeasesByMedia]
+      case NotFound(_, _) => LeasesByMedia.empty
+      case e@Error(_, _, _) => e.logError
     }
```
sihil, line 154: This is identical to the code below if you don't like for comprehensions (and I don't think it adds anything here).

```suggestion
   makeGetRequestAsync(url, authFn).map {
      case Found(json, _) => (json \ "data").as[LeasesByMedia]
      case NotFound(_, _) => LeasesByMedia.empty
      case e@Error(_, _, _) => e.logError
    }
```

common-lib/src/main/scala/com/gu/mediaservice/GridClient.scala
```diff
@@ -1,123 +1,85 @@
...
-      processResponse(response, url)
-    }
+  case class Found(json: JsValue, underlying: WSResponse) extends Response {
+    val status = 200
+  }
+  case class NotFound(body: String, underlying: WSResponse) extends Response {
+    val status = 404
   }
+  case class Error(status: Int, url: URL, underlying: WSResponse) extends Response {
+    def logError: Nothing = {
```
sihil, line 58: Given that this is side effecting this should have empty parens at the end to signify that. Also `Unit` is more idiomatic.

common-lib/src/main/scala/com/gu/mediaservice/GridClient.scala
```diff
@@ -132,73 +94,91 @@ class GridClient(apiKey: String, services: Services, maxIdleConnections: Int, de
...
-  private def validateResponse(res: ResponseWrapper, url: URL): Unit = {
-    import res._
-    if (statusCode != 200 && statusCode != 404) {
-      val errorMessage = s"breaking the circuit of full image projection, downstream API: $url is in a bad state, code: $statusCode"
-      val downstreamErrorMessage = res.bodyAsString
+  private def validateResponse(
+                                   response: WSResponse,
+                                   url: URL
+                                 ): Response = {
+//    val res = processResponse(response, url)
```
sihil, line 112: Nice to see this method disappear. Let's also get rid of the comment.

aug24: Done.

common-lib/src/main/scala/com/gu/mediaservice/GridClient.scala
aug24: As it will always throw an exception it needs to be Nothing.  Unit messes with type inference.
Unless you can suggest something better?  I could return the exception and throw it in the calling code?

sihil: Oh interesting. TIL. Neither of those options are particularly appealing. I try and avoid throwing exceptions with a preference of using an `Either` to communicate back to the caller. But I'm not going to force through another refactor.

As it is, is fine. Empty parens still better and I'd also name the method `logErrorAnd_*cough_sorry*_ThrowException`.

common-lib/src/main/scala/com/gu/mediaservice/GridClient.scala
aug24: Done.

PR #3188, opened 2021-02-26
Removal of the BBC folder from common-lib

gribeiro:
## What does this change?

As we now have https://github.com/guardian/grid/pull/3186, we can move BBC specific code to a separate project.
This PR removes the BBC image processors from common-lib.

This shouldn't affect any current process.

## Who should look at this?
@guardian/digital-cms @wainaina @adrielulanovsky 

## Tested?
- [x] locally by committer


sihil:
Hello @gribeiro - am I right in thinking that you are keeping BBC code in a separate repository and merging it in at build time? This is not quite what I expected (although fair enough). I had imagined that we would keep the code in this repository and my motivation for that was that:
 - if we made breaking changes to any of the provider interfaces we can detect breaks to your code and make atomic commits that make any necessary fixes
 - we'd be able to advise on best practice, although we would not provide the same level of reviewing that we do for code landing in the main codebase

Just a thought - I don't have a strong opinion and long term what you are doing makes sense, although I'd eventually like to see better separation between the projects (your project consuming Grid as a library in some way).

gribeiro:
Hello @sihil.

Thank you for bringing it up these topics, we think that they are valid concerns.

Regarding the first point, the build process has to incorporate the flag on the build. So, It may work for some signature changes, but perhaps it would fail for a more comprehensive test. 

For the second point, I think we can mitigate this by sharing with your team our private repository. 

I think that in the long run, having a separate repository is more beneficial, it brings more previsibility and speed for some of our tasks while still using the most recent changes of this repository. From the project perspective, this model can be replicated for many groups more that would be interested in contributing to the Grid project.

sihil (approved):

PR #3190, opened 2021-02-26
'add to pinboard' button for crop thumbnails

twrichards:
Co-authored-by: @tjsilver 

https://trello.com/c/TIeyuEff/511-1st-pinboard-integration-grid-phase-1

Following https://github.com/guardian/grid/pull/3179 we can now load the pinboard script via config, this PR benefits from the pinboard improvements in https://github.com/guardian/editorial-tools-pinboard/pull/36 which actually does something with the data-attributes on `<pinboard-add-button` HTML tags (i.e. puts them in the `payload` of the `item`s it stores).

## What does this change?
This PR simply adds the `<pinboard-add-button` HTML tag with the following data attributes...
- `data-type="grid-crop"` this is needed to distinguish from other types of item (such as `message-only` or `mam-video`)
- `data-grid-crop-thumbnail="{{:: extremeAssets.smallest | assetFile}}"` the URL pointing directly to the smallest thumbnail for the crop
- `data-grid-image-id="{{ctrl.image.data.id}}"` the main image identifier
- `data-grid-crop-id="{{crop.id}}"` the identifier for the specific crop

... as per the philosophy of pinboard this HTML tag means nothing to the host platform (in this case grid) but is picked up dynamically by pinboard and a button is rendered using React Portals. [TODO add link to pinboard ADR once written].

This is just phase 1 of the grid integration, in future we hope to support 'pinning' the base image and collections as well as drag&drop support (see https://trello.com/c/5WaJkX3Q/587-pinboard-grid-integration-phase-2-base-images-and-collections).

## How to test
With this branch deployed to Grid TEST and `tr-ts-grid-image-support` branch deployed to Pinboard CODE:
- Create a crop of an image
- Click on the `add to 📌` button
- Observe as pinboard opens (if not open already) and a selector popout appears allowing you to choose a pinboard to add the image to
- Select a suitable pinboard e.g. 'test draggable'
- Notice the image preview in the text box
- Observe that you can also discard the image in the selector or text box
- Add optional message and hit send
- Notice the image in the discussion thread above

To test how to use this image, see details in the [Pinboard PR](https://github.com/guardian/editorial-tools-pinboard/pull/36).

Different editorial teams can share grid assets for a piece with accompanying discussion.

## Screenshots
grid_drop

## Tested?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/image/crop.html
+++ b/kahuna/public/js/image/crop.html
@@ -15,4 +15,10 @@
     <span class="flex-spacer"></span>
     <span class="image-crop__creator" title="Cropped by {{:: crop.author}} at {{:: crop.date | date:'medium'}}">{{:: crop.author | getInitials}}</span>
   </div>
+  <pinboard-add-button
+    data-type="grid-crop"
+    data-grid-crop-thumbnail="{{:: extremeAssets.smallest | assetFile}}"
+    data-grid-image-id="{{ctrl.image.data.id}}"
+    data-grid-crop-id="{{crop.id}}"
+  ></pinboard-add-button>
 </div>
--- a/kahuna/public/js/image/crop.html
+++ b/kahuna/public/js/image/crop.html
@@ -19,7 +19,6 @@
     data-source="grid"
     data-source-type="crop"
     data-thumbnail="{{:: extremeAssets.smallest | assetFile}}"
-    data-primary-id="{{ctrl.image.data.id}}"
-    data-secondary-id="{{crop.id}}"
+    data-embeddable-url="{{ctrl.getEmbeddableUrl(crop.id)}}"
   ></asset-handle>
 </div>
--- a/kahuna/public/js/image/crop.html
+++ b/kahuna/public/js/image/crop.html
@@ -15,10 +15,11 @@
     <span class="flex-spacer"></span>
     <span class="image-crop__creator" title="Cropped by {{:: crop.author}} at {{:: crop.date | date:'medium'}}">{{:: crop.author | getInitials}}</span>
   </div>
-  <pinboard-add-button
-    data-type="grid-crop"
-    data-grid-crop-thumbnail="{{:: extremeAssets.smallest | assetFile}}"
-    data-grid-image-id="{{ctrl.image.data.id}}"
-    data-grid-crop-id="{{crop.id}}"
-  ></pinboard-add-button>
+  <asset-handle
+    data-source="grid"
+    data-source-type="crop"
+    data-thumbnail="{{:: extremeAssets.smallest | assetFile}}"
+    data-primary-id="{{ctrl.image.data.id}}"
+    data-secondary-id="{{crop.id}}"
+  ></asset-handle>
 </div>
--- a/kahuna/public/js/image/controller.js
+++ b/kahuna/public/js/image/controller.js
@@ -193,6 +193,9 @@ image.controller('ImageCtrl', [
       $state.go('image', {imageId: ctrl.image.data.id, crop: undefined}, {reload: true});
     };
 
+    ctrl.getEmbeddableUrl = (maybeCropId) =>
+      $state.href($state.current.name, { crop: maybeCropId || null }, { absolute: true });
+
     // TODO: move this to a more sensible place.
     function getCropDimensions() {
       return {


kahuna/public/js/image/crop.html
```diff
@@ -15,4 +15,10 @@
     <span class="flex-spacer"></span>
     <span class="image-crop__creator" title="Cropped by {{:: crop.author}} at {{:: crop.date | date:'medium'}}">{{:: crop.author | getInitials}}</span>
   </div>
+  <pinboard-add-button
+    data-type="grid-crop"
+    data-grid-crop-thumbnail="{{:: extremeAssets.smallest | assetFile}}"
+    data-grid-image-id="{{ctrl.image.data.id}}"
+    data-grid-crop-id="{{crop.id}}"
+  ></pinboard-add-button>
```
sihil, line 23: Out of interest can this be self closing???
```suggestion
  />
```

sihil (commented):
Looks good to me in principal. My only thought is whether we can name this tag semantically rather than referring to an implementation. What is the semantic noun describing what this _is_? That way it could theoretically be useful for other tooling and it would site more nicely in our various tools.

I guess a vaugely similar concept is the label for an input field?

Hopefully what I'm suggesting makes sense.

kahuna/public/js/image/crop.html
twrichards: we did have it self-closing first, but there was problem with it, can't recall exactly what. Perhaps the linter didn't like it. I do know that IntelliJ guards against self-closing tags...
image
... which seems to be explained in this answer https://stackoverflow.com/a/71836/1868921 (not that we have IE8 users - I hope). Its kind of academic, because in the real DOM it will very quickly become not self-closing once Pinboard's React Portal takes over and adds in HTML within.

twrichards:
> Looks good to me in principal. My only thought is whether we can name this tag semantically rather than referring to an implementation. What is the semantic noun describing what this _is_? That way it could theoretically be useful for other tooling and it would site more nicely in our various tools.
> 
> I guess a vaugely similar concept is the label for an input field?
> 
> Hopefully what I'm suggesting makes sense.

I can only kind-of see what you're saying, do have a suggestion for a name that might exemplify your point? I'm a little conflicted about making the name too generic, as people looking for pinboard stuff might find it hard to find?

sihil:
It is representing something that can be pinned or collected. Something like `asset` might work to make this generic. The data fields should probably be generic too (if it's feasible to remove any custom grid logic from pinboard - I guess that the OG stuff might help).
```html
<asset
  data-source="grid"
  data-source-type="crop"
  data-thumbnail="https://example.com/thumbnail.png"
  data-asset-url="https://media.gutools.co.uk/image/5432543254325432/0_0_1000_1000"
></asset>
```

kahuna/public/js/image/crop.html
```diff
@@ -15,4 +15,11 @@
     <span class="flex-spacer"></span>
     <span class="image-crop__creator" title="Cropped by {{:: crop.author}} at {{:: crop.date | date:'medium'}}">{{:: crop.author | getInitials}}</span>
   </div>
+  <asset-handle
+    data-source="grid"
+    data-source-type="crop"
+    data-thumbnail="{{:: extremeAssets.smallest | assetFile}}"
+    data-primary-id="{{ctrl.image.data.id}}"
+    data-secondary-id="{{crop.id}}"
```
sihil, line 23: What's the reason for the split ID here? What do we use this for? In my head we need these to drop the right thing into the Composer editor. The editor understands URIs, so can we use crop URIs here? Otherwise something presumably needs to know how to make the URI anyway?

twrichards: correct, we build the embed URL in pinboard, however in the sprit of your original comment/suggestion we should probably have an `data-embed-url` rather than the primary & secondary data attributes... problem is, I'm struggling to find the code that sets the browser URL when a crop is selected (as its that URL which we're building on the pinboard side, given thats what composer supports)

twrichards: ok I managed to work it out, what do you think of... https://github.com/guardian/grid/pull/3190/commits/2f5c4a8dbc58422be6b696c12409e154072d85a0

sihil (approved):

twrichards:
RE-TESTED in `TEST` (with https://github.com/guardian/editorial-tools-pinboard/pull/36 in `CODE`)

PR #3196, opened 2021-03-03
chore: add Dependabot configuration

akash1810:
## What does this change?

Dependabot is a little nicer than Snyk:
- It has a simpler API, driven by comments on GitHub
- It authors under a consistent username

The downside of this is Dependabot doesn't (yet) support SBT and Snyk does, however I can only see PRs against NPM dependencies so maybe this is OK?

We may want to consider [throttling](https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#open-pull-requests-limit) the number of PRs at the start as it might be overwhelming with so many sub-projects.

See https://github.com/guardian/grid/pulls?q=is%3Apr+%5BSnyk%5D+.

Dependency updates are attributed to a consistent user and all `package.json` files are inspected.

(There are a lot in this repo, which might suggest a need for workspaces which [recently landed in NPM v7](https://docs.npmjs.com/cli/v7/using-npm/workspaces) or [Yarn](https://classic.yarnpkg.com/en/docs/workspaces/))

## Screenshots

n/a

## Who should look at this?

@guardian/digital-cms

## Tested?
n/a

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,40 +9,6 @@ updates:
     commit-message:
       prefix: "chore(deps): "
 
-  # the frontend site
-  - package-ecosystem: 'npm'
-    directory: '/kahuna'
-    schedule:
-      interval: 'weekly'
-    commit-message:
-      prefix: "chore(deps): "
-
-  # things that run in an AWS Lambda environment
-  - package-ecosystem: 'npm'
-    directory: '/reaper'
-    schedule:
-      interval: 'weekly'
-    commit-message:
-      prefix: "chore(deps): "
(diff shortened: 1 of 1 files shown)


sihil:
Thi kind of thing scares me at scale. 

It is the new boilerplate of git based config that will be hard to update across hundreds of repos when dependabot best practice changes. I kind of feel that we need a metabot that has control of `.github`.

sihil (approved):
Hopefully this is better than snyk which I guess we shoudl disable for the repo.

I presume that when we have a new project it will not be detected and we'll need to fix it ourselves?

PR #3197, opened 2021-03-03
Implement permissions for own uploaded images

adrielulanovsky:
## What does this change?
Currently, if you upload an image you can delete it or edit its metadata even if you don't have general permissions to do those actions. With these changes, we add flexibility to the permissions system, by adding 2 new permissions:

- UploaderEditMetadata means the permission to edit metadata for images that were uploaded by the same user trying to edit.
- UploaderDeleteImage means the permission to delete images that were uploaded by the same user trying to delete.

These permissions will be set to true for the Guardian, while they will depend on the permission to delete or edit metadata for BBC.

- A user without permission to edit metadata can still edit metadata from their uploaded images in Guardian Grid.
- A user without permission to delete images can still delete their uploaded images in Guardian Grid.
- A user without permission to edit metadata can no longer edit metadata from their uploaded images in BBC Grid.
- A user without permission to delete images can no longer delete their uploaded images in BBC Grid.

## Who should look at this?
@guardian/digital-cms, @sihil , @wainaina , @gribeiro 

## Tested?
- [x] locally by committer

### DIFF
--- a/media-api/app/controllers/MediaApi.scala
+++ b/media-api/app/controllers/MediaApi.scala
@@ -5,7 +5,7 @@ import com.google.common.net.HttpHeaders
 import com.gu.mediaservice.lib.argo._
 import com.gu.mediaservice.lib.argo.model.{Action, _}
 import com.gu.mediaservice.lib.auth.Authentication._
-import com.gu.mediaservice.lib.auth.Permissions.{DeleteCrops, DeleteImage, EditMetadata}
+import com.gu.mediaservice.lib.auth.Permissions.{DeleteCrops, DeleteImage, UploaderDeleteImage, EditMetadata, UploaderEditMetadata}
 import com.gu.mediaservice.lib.auth._
 import com.gu.mediaservice.lib.aws.{ThrallMessageSender, UpdateMessage}
 import com.gu.mediaservice.lib.formatting.printDateTime
@@ -89,12 +89,13 @@ class MediaApi(
   private def isUploaderOrHasPermission(
     principal: Principal,
     image: Image,
-    permission: SimplePermission
+    permission: SimplePermission,
+    uploaderPermission: SimplePermission
   ) = {
     principal match {
       case user: UserPrincipal =>
         if (user.email.toLowerCase == image.uploadedBy.toLowerCase) {
-          true
+          authorisation.hasPermissionTo(uploaderPermission)(principal)
         } else {
           authorisation.hasPermissionTo(permission)(principal)
         }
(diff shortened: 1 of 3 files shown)


AWare:
I don't think the upload page handles update failures, so I think this would might be necessary here.

adrielulanovsky:
> I don't think the upload page handles update failures, so I think this would might be necessary here.

You're right, I hadn't thought of that use case. Perhaps we could do the same as with /images/:id and hide the actions for deleting or editing metadata if the user doesn't have permission to do so? That way we wouldn't need to handle update failures

media-api/app/controllers/MediaApi.scala
```diff
@@ -89,12 +89,13 @@ class MediaApi(
...
+    uploaderPermission: SimplePermission
   ) = {
     principal match {
       case user: UserPrincipal =>
         if (user.email.toLowerCase == image.uploadedBy.toLowerCase) {
-          true
+          authorisation.hasPermissionTo(uploaderPermission)(principal)
         } else {
           authorisation.hasPermissionTo(permission)(principal)
         }
```
sihil, line 101: The logic here doesn't let someone who has `permission` but not `uploaderPermission` take the action on their own images. In reality anyone with `permission` will probably have `uploaderPermission` but it's a weird corner case.
This is likely better as `(isUploader && hasUploaderPermission) || hasPermission` to deal with that case rather than an if expression.

rest-lib/src/main/scala/com/gu/mediaservice/lib/auth/Permissions.scala
```diff
@@ -14,7 +14,9 @@ object Permissions {
   type PrincipalFilter = Principal => Boolean
 
   case object EditMetadata extends SimplePermission
+  case object UploaderEditMetadata extends SimplePermission
```
sihil, line 17: Let's start putting comments on these so it's easier to understand what their intention is.

rest-lib/src/main/scala/com/gu/mediaservice/lib/guardian/auth/PermissionsAuthorisationProvider.scala
```diff
@@ -52,7 +52,9 @@ class PermissionsAuthorisationProvider(configuration: Configuration, resources:
     }
     permissionContext match {
       case EditMetadata => hasPermission(Permissions.EditMetadata)
+      case UploaderEditMetadata => true
       case DeleteImage => hasPermission(Permissions.DeleteImage)
+      case UploaderDeleteImage => true
```
sihil, line 57: 👍

adrielulanovsky:
Upon further conversation, I'll be closing this since it is not a BBC requirement to avoid users without permission to delete or edit metadata from their uploaded images.

PR #3198, opened 2021-03-04
Convert Crid client to Play WS

aug24:
## What does this change?
Long term, we intend for Thrall to fetch content rather than blindly trust the content of the thrall message.  Similarly, we have re-projection work which requires some services to talk to other services.

The existing GridClient code is based on OK Http which adds a library (and therefore complexity) when Play's own WS library is _right there_.

This PR converts the client to WS and simplifies the interface as far as possible.

No change to behaviour.  Tests all pass.

## Tested?
- [ X] locally by committer
- [ X] on the Guardian's TEST environment


sihil (approved):
LGTM :shipit: 🚢

PR #3199, opened 2021-03-04
Add default local config for pinboard integration

tjsilver:
Co-authored-by: Tom Richards <tom.richards@guardian.co.uk>

## What does this change?
Following #3179, adds default local config for Pinboard integration, ONLY if `BUILD_ORG` env variable is NOT set (to keep things simple and error free for non-Guardian orgs, e.g. BBC).

## Tested?
- [x] locally by committer
- [(N/A)] on the Guardian's TEST environment

### DIFF
--- a/dev/script/generate-config/service-config.js
+++ b/dev/script/generate-config/service-config.js
@@ -72,7 +72,6 @@ function getImageLoaderConfig(config) {
 function getKahunaConfig(config) {
     return stripMargin`${getCommonConfig(config)}
         |aws.region="${config.AWS_DEFAULT_REGION}"
-        |aws.region="${config.AWS_DEFAULT_REGION}"
         |origin.full="images.media.${config.DOMAIN}"
         |origin.thumb="localstack.media.${config.DOMAIN}"
         |origin.images="images.media.${config.DOMAIN}"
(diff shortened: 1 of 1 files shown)


sihil:
Slightly concerned that this is going to start to drop pinboard into all copies of the Grid when run locally. The BBC use these same scripts. Can't immediately think of a good way of telling if someone is a Guardian dev or not though - perhaps we can see if they have the GNM certificate on their machine or something crazy like that.

twrichards:
> Slightly concerned that this is going to start to drop pinboard into all copies of the Grid when run locally. The BBC use these same scripts. Can't immediately think of a good way of telling if someone is a Guardian dev or not though - perhaps we can see if they have the GNM certificate on their machine or something crazy like that.

No need Simon, thats why the scriptToLoad has the permission option... BBC users will not have that permission and thus script will not be loaded.

twrichards:
> Slightly concerned that this is going to start to drop pinboard into all copies of the Grid when run locally. The BBC use these same scripts. Can't immediately think of a good way of telling if someone is a Guardian dev or not though - perhaps we can see if they have the GNM certificate on their machine or something crazy like that.

@sihil following slack discussions with BBC - I've wrapped the addition of pinboard stuff to local kahuna config in a check that `BUILD_ORG` is NOT set - see https://github.com/guardian/grid/pull/3199/commits/fb5ced7be998f93fff0d1d15b4b0ae4dbf5aeff7

`BUILD_ORG` is already being used in https://github.com/guardian/grid/blob/51deec48531cd3615b5183d808eee6881a652b2f/build.sbt#L72
...so seemed safe enough.

sihil (approved):

sihil:
At some stage I intend to set BUILD_ORG to `guardian` to build our providers but we can deal with that when it comes to it.

Issue #3200, opened 2021-03-04
UX Improvement: Search for images by their additional file metadata on the UI


Asmau1234:
### Short Description

This is a stacked proposal on [Make additional image fields visible on the UI](https://github.com/guardian/grid/issues/3167) proposal which introduces extra image field configuration through this [PR #3183](https://github.com/guardian/grid/pull/3183) that configures additional image fields to be rendered on the Grid Image view UI under the “Show additional metadata” section. 
This proposal leverages the configuration and introduces an approach in which additional search hints can be introduced via configuration to the dropdown search hint / suggestion list on the UI when the user types in **+** into the search bar. Users should be able to select a field, input a value and receive a response with all the results that match the parameters. 

**_Sample configuration:_**
 
>  {
>    elasticsearchPath = "fileMetadata.xmp.bbc:elvis_collection"
>    alias = "bbcElvisCollection"
>    label = "BBC Elvis Collection"
>    displaySearchHint = true
>  }

### Approach

1. Read **fieldAliases** configurations injected into the UI.
2. Filter configurations out so only fieldAlias configurations that have the property `displaySearchHint = true` will have their `alias` in the advanced filter suggestions.
3. Use the elasticSearchPath value to query Elasticsearch e.g  `elasticsearchPath= “fileMetadata.xmp.bbc.elvis_collection” `
4. Serve the user matching results

sihil:
Hi @Asmau1234. I think that we would want these aliases to work when querying the API programatically, not just through the Kahuna web app. As such the aliases need to be mapped on the server side. Possibly this can be done in the `QuerySyntax` or `QueryBuilder` files.

paperboyo:
Closed by https://github.com/guardian/grid/pull/3205. Thanks!

PR #3205, opened 2021-03-06
UX Improvement: Ability to search for extra (searchable) file metadata fields on the UI

Asmau1234:
## What does this change?

This references [Proposal 3200](https://github.com/guardian/grid/issues/3200). It introduces a way of adding additional advanced search hints / suggestions via configuration on the Grid UI. This is achieved by reusing the existing `field.aliase`s configuration in `kahuna.conf`.

_Sample configuration:_
_Field aliases configurations that are used to define which metadata field is visible, searchable and has an alias on the UI_

```hocon
field.aliases = [
  {
    # Sets the image metadata tag to be used when filtering FileMetadata field
    elasticsearchPath = "fileMetadata.xmp.bbc:Programme-Maker"
    # Sets an alias for use in search, shouldn't have whitespace or special characters
    alias = "BBCProgrammeMaker"
    # Sets a human readable label that will rendered on the UI
    label = "BBC Programme Maker"
    # Sets the field alias to be displayed as a search hint when you type `+`
    displaySearchHint = false
  },
  {
    elasticsearchPath = "fileMetadata.xmp.bbc:elvis_collection"
    alias = "bbcElvisCollection"
     label = "BBC Elvis Collection"
     displaySearchHint = true
  }
]

```

In the search bar, a user should be able to enter an organization/admin enabled file metadata alias and only images matching that metadata field are displayed: `<property>:<value>`. 
Additionally, a user should be able to select any of the additional field aliases where the `displaySearchHint` is true from the advanced filter by pressing **+**, select a hint, type in search value and be able to view only images matching the search parameters.  

## Screenshots

Screenshot 2021-03-06 at 06 26 34

## Who should look at this?
@guardian/digital-cms
@sihil 
@AWare 

## Tested?
- [X] locally by committer

### DIFF
--- a/kahuna/public/js/search/structured-query/query-suggestions.js
+++ b/kahuna/public/js/search/structured-query/query-suggestions.js
@@ -7,6 +7,10 @@ export const querySuggestions = angular.module('querySuggestions', [
     mediaApi.name
 ]);
 
+const fieldAliases = window._clientConfig.fieldAliases.
+                                  filter(entry => entry.displaySearchHint === true).
+                                  map(entry => entry.alias);
+
 // FIXME: get fields and subjects from API
 export const filterFields = [
     'by',
@@ -38,7 +42,8 @@ export const filterFields = [
     'filename',
     'photoshoot',
     'leasedBy',
-    'is'
+    'is',
+    ... fieldAliases
 ].sort();
 // TODO: add date fields
 
--- a/media-api/app/lib/elasticsearch/QueryBuilder.scala
+++ b/media-api/app/lib/elasticsearch/QueryBuilder.scala
@@ -11,7 +11,16 @@ import com.sksamuel.elastic4s.requests.common.Operator
 import com.sksamuel.elastic4s.requests.searches.queries.Query
 import com.sksamuel.elastic4s.requests.searches.queries.matches.{MultiMatchQuery, MultiMatchQueryBuilderType}
 import lib.querysyntax._
-class QueryBuilder(matchFields: Seq[String], overQuotaAgencies: () => List[Agency]) extends ImageFields with GridLogging {
+import lib.MediaApiConfig
+
+class QueryBuilder(matchFields: Seq[String], overQuotaAgencies: () => List[Agency], config: MediaApiConfig) extends ImageFields with GridLogging {
+
+  def resolveFieldPath(field: String): String = {
+    config.fieldAliasConfigs.find(_.alias == field) match {
+      case Some(x) => x.elasticsearchPath
+      case None => getFieldPath(field)
+    }
+  }
 
   // For some sad reason, there was no helpful alias for this in the ES library
   private def multiMatchPhraseQuery(value: String, fields: Seq[String]): MultiMatchQuery =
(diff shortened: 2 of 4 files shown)


kahuna/public/js/search/structured-query/query-suggestions.js
```diff
@@ -79,6 +79,13 @@ const isSearch = [
 
 querySuggestions.factory('querySuggestions', ['mediaApi', 'editsApi', function(mediaApi, editsApi) {
 
+    window._clientConfig.fieldAliases.
```
AWare, line 82: I think this would be better moved to the definition of filterFields, it's a little uncomfortable that it's exported and then mutated in the factory. Would also probably go for a filter/map combo.

sihil: We had a look at this this morning and think that this is a better way of expressing it:
```javascript
const fieldAliases = window._clientConfig.fieldAliases
  .filter(entry => entry.displaySearchHint)
  .map(entry => entry.alias);

// FIXME: get fields and subjects from API
export const filterFields = [
    'by',
    'category',
    'city',
    'copyright',
    'country',
    'credit',
    'description',
    'fileType',
    'illustrator',
    'in',
    'keyword',
    'label',
    'location',
    'person',
    'source',
    'state',
    'subject',
    'supplier',
    'uploader',
    'usages@<added',
    'usages@>added',
    'usages@platform',
    'usages@status',
    'usages@reference',
    'has',
    'croppedBy',
    'filename',
    'photoshoot',
    'leasedBy',
    'is',
    ... fieldAliases
].sort();
```

Asmau1234: Thanks Simon. I will apply this right away.

media-api/app/lib/elasticsearch/QueryBuilder.scala
```diff
@@ -11,8 +11,21 @@ import com.sksamuel.elastic4s.requests.common.Operator
...
+
+    val fileMetadataPath = config.fieldAliasConfigs.find(_.alias == field) match {
+      case Some(x) => Some(x.elasticsearchPath)
+      case None => None
+    }
+    val simplePath = getFieldPath(field)
+
+    if(fileMetadataPath != None) fileMetadataPath.get
+    else simplePath
+  }
```
sihil, line 28: This function is needlessly complicated. I'd also suggest renaming it to something like `resolveFieldPath`.
```suggestion
  def getAdvancedFieldPath(field: String): String = {
    val fileMetadataPath = config.fieldAliasConfigs.find(_.alias == field) match {
      case Some(x) => x.elasticsearchPath
      case None => getFieldPath(field)
    }
  }
```

media-api/test/lib/elasticsearch/QueryBuilderTest.scala
```diff
@@ -222,6 +230,14 @@ class QueryBuilderTest extends FunSpec with Matchers with ConditionFixtures {
     }
   }
 
+  describe("get advanced file metadata") {
+    it("should return the elasticsearchPath or the field parameter itself if the config doesn't exist"){
```
sihil, line 234: This is testing two things. If you split it into two then you can more precisely describe the actual tests.

sihil (commented):
Hi @Asmau1234. I think this is close to being ready to go. 

Part of me thinks that the field alias expansion should happen in [ImageFields](https://github.com/guardian/grid/blob/main/common-lib/src/main/scala/com/gu/mediaservice/lib/ImageFields.scala#L50-L58) but injecting the configuration into that trait will be very difficult and require a significant refactor of a number of files so I think your solution is the pragmatic one.

Do have a quick look at the simpler implementation of the field path function and it's possible to make the tests a little cleaner.

media-api/app/lib/elasticsearch/QueryBuilder.scala
Asmau1234: I agree with you. I also looked into doing that in [ImageFields](https://github.com/guardian/grid/blob/main/common-lib/src/main/scala/com/gu/mediaservice/lib/ImageFields.scala#L50-L58) but had to revert back to this as it just seems more straight forward and does not require changing multiple files. Will simplify the function now.

sihil (approved):
Thanks for this excellent contribution @Asmau1234.

Issue #539, opened 2015-03-16
Record lastModifiedBy


theefer:
We have added a lastModified timestamp (https://github.com/guardian/media-service/pull/537) but no lastModifiedBy. That could be useful information to have.

Currently the metadata editor and cropper don't record the author of changes. We know we want to save the author of a crop (https://github.com/guardian/media-service/issues/292), and likely the author of metadata edits, so once that is in place, we could index that information in ES.

paperboyo:
Part of _My stuff_:
`uploader:me`
`croppedBy:me`
`lastModified:me`

honorcb:
When you look at the api behind an image,  you can find a links, which the offer a url to edits,  which show you when  “lastModified”: date time,  but not by whom.  Some other changes do record who,  but are hard to find ,as the field before the person’s name vary,  can be author, or describe what they have done, like adding leases is “leasedBy”. 

Plus those listed above

PR #3207, opened 2021-03-09
feature/upload-images-permission

adrielulanovsky:
## What does this change?
This implements a new permission, UploadImages, that controls which users can upload images to the Grid. Users without this permission will not be able to access any loader endpoint, will be shown an error message when trying to enter /upload endpoint in Kahuna, and will have hidden all upload-related UI elements, such as:
- My Uploads checkbox
- My recent uploads button
- Upload button

Users with UploadImages permission are able to access any loader endpoint, are not shown any error message when trying to enter /upload endpoint in Kahuna, and can see and use all upload-related UI elements, such as:
- My Uploads checkbox
- My recent uploads button
- Upload button

Users without UploadImages permission are not able to access any loader endpoint, are shown an error message when trying to enter /upload endpoint in Kahuna, and have hidden all upload-related UI elements, such as:
- My Uploads checkbox
- My recent uploads button
- Upload button

## Screenshots
Top bar for users without upload permission:
image

/upload frontend endpoint with authorisation error:
image

## Who should look at this?
@guardian/digital-cms, @gribeiro , @wainaina 

## Tested?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/search/query.html
+++ b/kahuna/public/js/search/query.html
@@ -51,7 +51,7 @@
             </label>
             </li>
 
-            <li class="search__modifier-item">
+            <li ng-if="searchQuery.canUpload" id="my-uploads-checkbox" class="search__modifier-item">
             <label>
                 <input type="checkbox"
                     ng-model="searchQuery.filter.uploadedByMe" />
--- a/kahuna/public/js/search/view.html
+++ b/kahuna/public/js/search/view.html
@@ -33,4 +33,4 @@
 </gr-panels>
 
 <ui-view name="multiDrag"></ui-view>
-<dnd-uploader></dnd-uploader>
+<dnd-uploader ng-if="ctrl.canUpload"></dnd-uploader>
--- a/kahuna/public/js/search/view.html
+++ b/kahuna/public/js/search/view.html
@@ -5,14 +5,15 @@
     </gr-top-bar-nav>
 
     <gr-top-bar-actions>
-        <a ui-sref="upload"
+        <a ng-if="ctrl.canUpload"
+           ui-sref="upload"
            gr-track-click="Your recent uploads button"
            gr-track-click-data="{'Location': 'Search navigation'}"
            class="top-bar-item clickable side-padded"
            title="my recent uploads">
             <gr-icon-label gr-icon="photo_library">My recent uploads</gr-icon-label>
         </a>
-        <file-uploader class="top-bar-item side-padded"></file-uploader>
+        <file-uploader ng-if="ctrl.canUpload" class="top-bar-item side-padded"></file-uploader>
     </gr-top-bar-actions>
 </gr-top-bar>
 
--- a/kahuna/public/js/upload/view.html
+++ b/kahuna/public/js/upload/view.html
@@ -7,7 +7,7 @@
     <gr-top-bar-actions></gr-top-bar-actions>
 </gr-top-bar>
 
-<div class="page-wrapper">
+<div ng-if="ctrl.canUpload" class="page-wrapper">
     <file-prompt class="section"></file-prompt>
 
     <section class="section"
@@ -27,5 +27,6 @@
         <recent-uploads></recent-uploads>
     </section>
 </div>
+<div ng-if="ctrl.canUpload === false" class="unauthorised-message">You are not authorised to upload images</div>
 
 <dnd-uploader></dnd-uploader>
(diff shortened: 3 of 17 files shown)


kahuna/public/js/search/query.html
```diff
@@ -51,7 +51,7 @@
             </label>
             </li>
 
-            <li class="search__modifier-item">
+            <li ng-if="searchQuery.canUpload" id="my-uploads-checkbox" class="search__modifier-item">
```
AWare, line 54: Do we want to hide the myUploads search for users who can't (at this moment) upload?

adrielulanovsky: Well, I believe this permission will not be something that would be toggled on and off for a user, but a more permanent permission. However, this was not an explicit requirement. I'm happy to remove this if we don't want to hide that search.

paperboyo: I don’t think we have plans of introducing this permission at the Guardian. If indeed this is more likely to be permanent for a give user at the BBC, then it makes sense in my mind to also hide this checkbox (and the dropzone effect and all other upload places/opportunities). As soon as a user is granted the permission, they would all come back anyway. Only the removal would be of (small) concern, but then we are back to the most likely use case.

Any user is always able to run this search manually anyway using `+uploader:` chip. Maybe, one day, even quicker than [by typing the full email address](https://github.com/guardian/grid/issues/539#issuecomment-404420763), too 😃.

adrielulanovsky: @paperboyo actually now that you mentioned it, I wasn't aware of this dropzone effect :sweat_smile: I'll look into deactivating that as well. Do you know of any other upload features that should be blocked as well?

paperboyo: > Do you know of any other upload features that should be blocked as well?

Can’t think of anything. Thanks for taking care of dropzone (on both the search page and upload page)!

sihil:
An initial comment on this. My understanding of the hypermedia spec is that if a user don't have permission to do an action we shouldn't advertise the link in the response. So for example in this case, if the user doesn't have upload permissions then [these two rows](https://github.com/guardian/grid/blob/main/image-loader/app/controllers/ImageLoaderController.scala#L49-L50) returned by `/` on the loader API shouldn't be returned.

Rather than passing a `canUpload` flag separately the UI should infer whether a user can upload or not by checking whether these links were in the loader API hypermedia response.

Can you confirm @akash1810 ?

adrielulanovsky:
> An initial comment on this. My understanding of the hypermedia spec is that if a user don't have permission to do an action we shouldn't advertise the link in the response. So for example in this case, if the user doesn't have upload permissions then [these two rows](https://github.com/guardian/grid/blob/main/image-loader/app/controllers/ImageLoaderController.scala#L49-L50) returned by `/` on the loader API shouldn't be returned.
> 
> Rather than passing a `canUpload` flag separately the UI should infer whether a user can upload or not by checking whether these links were in the loader API hypermedia response.
> 
> Can you confirm @akash1810 ?

@sihil @akash1810 
I've addressed this, what do you think about the current approach? I'm actually basing the decision on the loader link from mediaAPI, because otherwise I think the frontend would need to do a new query to the loader service before knowing if the user has upload permissions or not, and it would add an unnecessary overhead.

kahuna/public/js/search/view.html
```diff
@@ -5,14 +5,15 @@
     </gr-top-bar-nav>
 
     <gr-top-bar-actions>
```
akash1810, line 7: Might be better to have the `ng-if` on this element rather than each individual child element?

kahuna/public/js/upload/view.html
```diff
@@ -27,5 +27,6 @@ <h2 class="flex-spacer">Your past 50 uploads</h2>
         <recent-uploads></recent-uploads>
     </section>
 </div>
+<div ng-if="ctrl.canUpload === false" class="unauthorised-message">You are not authorised to upload images</div>
```
akash1810, line 30: minor: should we provide information on how to resolve this? For example "You are not authorised to upload images. Send an email to admin@example.com if you think this is incorrect."

akash1810 (approved):
This looks good to me.

kahuna/public/js/search/view.html
adrielulanovsky: If used at the gr-top-bar-actions element, we lose the three dots menu with the Quotas, Feedback, Logout options

kahuna/public/js/upload/view.html
adrielulanovsky: Fixed

PR #3208, opened 2021-03-10
Justinrowles/simple json diff

aug24:
## What does this change?
When we load images, various items of information are parsed or created, such as "description", or "byline".

Occasionally, we change the algorithm which parses or creates these fields.  At present, there is no easy way to view what would be created if the image were to be uploaded _now_.

This PR provides an endpoint which gives a json diff between the existing ElasticSearch record and the record which would be written if the image were newly uploaded.

Tests pass, endpoint works, Mateusz is happy.

## Screenshots

image

## Who should look at this?
@guardian/digital-cms

## Tested?
- [X] locally by committer


media-api/conf/routes
```diff
@@ -14,6 +14,7 @@ GET     /images/aggregations/date/:field              controllers.AggregationCon
 # Images
 GET     /images/:id                                   controllers.MediaApi.getImage(id: String)
 GET     /images/:id/_elasticsearch                    controllers.MediaApi.getImageFromElasticSearch(id: String)
+GET     /images/:id/diffProjection                    controllers.MediaApi.diffProjection(id: String)
```
sihil, line 17: Can we use something like `/images/:id/projection/diff` so we can group the different projection endpoints together?

media-api/app/controllers/MediaApi.scala
```diff
@@ -138,6 +148,17 @@ class MediaApi(
     }
   }
 
+  def diffProjection(id: String) = auth.async { request =>
+    val onBehalfOfFn: OnBehalfOfPrincipal = auth.getOnBehalfOfPrincipal(request.user)
+    val apiCheckTimeout = new FiniteDuration(5, TimeUnit.SECONDS)
+    getImageResponseFromES(id, request) map {
+      case Some((source, _, _, _)) =>
+        val projection = Await.result(gridClient.getImageLoaderProjection(id, onBehalfOfFn), apiCheckTimeout)
```
sihil, line 156: What's our concern here about it taking longer than five seconds?

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
```diff
@@ -208,9 +208,6 @@ class S3(config: CommonConfig) extends GridLogging {
     S3Metadata(meta)
   }
 
-  def getUserMetadata(bucket: Bucket, key: Key): Map[Bucket, Bucket] =
-    client.getObjectMetadata(bucket, key).getUserMetadata.asScala.toMap
-
```
sihil, line 213: This seems unrelated?

sihil (commented):
Nice. A couple of minor comments but generally good to go.

media-api/conf/routes
aug24: Will do.

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
aug24: It was unused code.

sihil: I prefer not to do housekeeping in files that are otherwise unchanged as it makes for unclean PRs.

AWare: you can't win :(

media-api/app/controllers/MediaApi.scala
```diff
@@ -138,6 +147,17 @@ class MediaApi(
...
 
+  def diffProjection(id: String) = auth.async { request =>
+    val onBehalfOfFn: OnBehalfOfPrincipal = auth.getOnBehalfOfPrincipal(request.user)
+    getImageResponseFromES(id, request) flatMap {
+      case Some((source, _, _, _)) =>
+        gridClient.getImageLoaderProjection(id, onBehalfOfFn) map (
+        projection => respond(JsonDiff.diff(Json.toJson(source), Json.toJson(projection)))
+      )
+      case _ => Future.successful(ImageNotFound(id))
+    }
```
sihil, line 158: How about this? Longer, but easier in my opinion to comprehend. Plus if works when there is no record in ES/S3. Which feels right.

```suggestion
    for {
      maybeEsImage <- getImageResponseFromES(id, request)
      maybeEsJson = maybeEsImage.map{ case (source, _, _, _) => Json.toJson(source) }
      maybeProjectedImage <- gridClient.getImageLoaderProjection(id, onBehalfOfFn)
      maybeProjectedJson = maybeProjectedImage.map(Json.toJson)
    } yield {
      (maybeEsJson, maybeProjectedJson) match {
        case (None, None) => ImageNotFound(id)
        case (es, projected) => respond(JsonDiff.diff(
          es.getOrElse(JsObject.empty),
          projected.getOrElse(JsObject.empty)
        ))
      }
    }
```

PR #3210, opened 2021-03-10
Do the TODO for Collections

aug24:
## What does this change?
Includes collections when re-projecting an image.

Re-projected images will have collections.

## Tested?
- [ ] locally by committer


jonathonherbert (approved):
That simple? LGTM!

PR #3212, opened 2021-03-12
Enact S3 metadata changes

aug24:
## What does this change?
Once we have found all the metadata which needs updating in S3, we need a tool to enact those changes.

This is that tool.

Given a file (output from Propose S3 Changes tool https://github.com/guardian/grid/pull/3213), we can update S3 user metadata appropriately.

## Tested?
- [ ] locally by committer

### DIFF
--- a/scripts/src/main/scala/com/gu/mediaservice/scripts/EnactS3Changes.scala
+++ b/scripts/src/main/scala/com/gu/mediaservice/scripts/EnactS3Changes.scala
@@ -7,17 +7,17 @@ import org.apache.commons.compress.compressors.bzip2.{BZip2CompressorInputStream
 import play.api.libs.json.Json
 
 import scala.io.Source
-
-import com.amazonaws.regions.Regions
 import com.amazonaws.services.s3.AmazonS3
 import com.amazonaws.services.s3.AmazonS3ClientBuilder
-
+import com.amazonaws.services.s3.model.CopyObjectRequest
+import collection.JavaConverters._
 
 object EnactS3Changes {
 
   private val profile = "media-service"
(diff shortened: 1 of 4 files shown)


sihil:
Before merging this, I'm keen to understand what the performance is going to be like. Is this approach feasible for our anticiapated update of 6m objects?

aug24:
> Before merging this, I'm keen to understand what the performance is going to be like. Is this approach feasible for our anticiapated update of 6m objects?

I'm inclined to make the attempt with a few thousand and see.... but to do that, I think we should merge.

scripts/src/main/scala/com/gu/mediaservice/scripts/EnactS3Changes.scala
```diff
@@ -0,0 +1,90 @@
...
+        .zipWithIndex
+        .foreach {
+          case ((Some(proposed), Some(original)), i) =>
+            if (i % 10000 == 0) System.err.println(s"Processing object metadata line $i")
+            val key = proposed.key
+            either {s3.getObjectMetadata(bucketName, key)} match {
+              case Right(awsObjectMetadata) =>
+                val check = awsObjectMetadata.getUserMetadata.asScala
+                if (check!=original.metadata) {
+                  System.err.println(s"Unable to update - AWS entry for key ${original.key} has changed")
```
sihil, line 62: Rather than just log this to stderr should we actually write this to an audit log which is actionable afterwards. Perhaps we should keep an audit log for updates too.

sihil (approved):
Approving this since we are now running in anger against PROD.

PR #3214, opened 2021-03-17
Create merged metadata on reprojection

aug24:
## What does this change?
Currently, on re-projection (recreation of image data) the metadata is not populated _at all_.

It should be populated by taking the original file metadata, creating image metadata, and then merging down any user metadata changes.  It was missed originally because the re-projection tool was first created for images which were orphaned in S3 and so could not have _any_ user edits.

This missing action has come to light via diff'ing of projections.

Smaller diffs, with correct information.

## Tested?
- [ ] locally by committer


scripts/src/main/scala/com/gu/mediaservice/scripts/EnactS3Changes.scala
```diff
@@ -0,0 +1,90 @@
+package com.gu.mediaservice.scripts
```
AWare, line 1: Is this file supposed to be in the PR?

AWare (approved):
I think the S3 changes stuff might better belong in its own PR.

AWare (approved):

PR #3215, opened 2021-03-19
Debug projection

aug24:
## What does this change?
Several small bugs/improvements in the projection code:
 - user metadata last modified date is closer to current value
 - thumbnails and optimised mime types derive from their parent objects rather than hard coded.

Mostly mime type for optimised files is correct, see diff endpoint.

## Tested?
- [ ] locally by committer

### DIFF
--- a/image-loader/app/model/upload/OptimiseOps.scala
+++ b/image-loader/app/model/upload/OptimiseOps.scala
@@ -33,7 +33,7 @@ object OptimiseWithPngQuant extends OptimiseOps {
 
     val optimisedFile = new File(optimisedFilePath)
     if (optimisedFile.exists()) {
-      (optimisedFile, Png)
+      (optimisedFile, optimiseMimeType)
     } else {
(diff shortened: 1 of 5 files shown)


image-loader/app/model/upload/OptimiseOps.scala
```diff
@@ -49,7 +49,7 @@ object OptimiseWithPngQuant extends OptimiseOps {
           case Some("True Color with Alpha") => true
           case _ => false
         }
-      case Some(Tiff) => true
+      case Some(Tiff) => true // TODO This should be done better, it could be better optimised into a jpeg.
```
paperboyo, line 52:
```suggestion
      case Some(Tiff) => true // TODO This should be done better, it could be better optimised into a jpeg if there is no transparency.
```

jonathonherbert (approved):
Code looks good. Intent much clearer w/ case class and method. Something something tests something. 

... a cracking todo, I always like to see the word 'hope' in a PR.

PR #3220, opened 2021-03-23
Permissions on Metadata service

adrielulanovsky:
## What does this change?
This implements permission checks on the following metadata-editor service endpoints for editing metadata:
```

PUT     /metadata/:id/metadata                          controllers.EditsController.setMetadata(id: String)
POST    /metadata/:id/metadata/set-from-usage-rights    controllers.EditsController.setMetadataFromUsageRights(id: String)

```
With this change, if a request to one of those endpoints is sent by a user who is not the uploader of the image AND does not have EditMetadata permission, the request will be denied.

To get the information about the uploader of the image, a new endpoint was implemented in media-api service:
`
GET     /images/:imageId/uploadedBy                   controllers.MediaApi.uploadedBy(imageId: String)
`
This endpoint returns the uploadedBy field of the image as a string. The metadata-editor service will make use of this endpoint to check if the user trying to edit metadata is the original uploader of the image.

Also, this implements some changes in the UI. In the case of uploading an image already in the system with a diffent uploader, and no EditMetadata permissions, the /upload page reflects this by showing a warning on the Image Metadata section of the conflicting image, and by disallowing edits to the metadata of that image. Batch metadata copying should work without any errors, since the batch process has been updated to avoid copying data into images without permissions.

- As a user with EditMetadata permissions, I'm able to edit the metadata of any image.
- As a user without EditMetadata permissions, I'm only able to edit the metadata of the images I uploaded.
- When an image without EditMetadata permissions is uploaded, the user cannot edit the metadata of it, and an appropriate warning is shown in the UI.

## Screenshots
Batch upload of images, one of which is an image with a different uploader, and with no EditMetadata permissions:
image

## Who should look at this?
@guardian/digital-cms @sihil @ochiengolanga 

## Tested?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/upload/jobs/required-metadata-editor.css
+++ b/kahuna/public/js/upload/jobs/required-metadata-editor.css
@@ -1,0 +1,3 @@
+.job-editor__disabled {
+    opacity: 0.4;
+}
--- a/kahuna/public/js/upload/jobs/required-metadata-editor.html
+++ b/kahuna/public/js/upload/jobs/required-metadata-editor.html
@@ -1,4 +1,4 @@
-<form class="job-editor" name="jobEditor" ng-submit="ctrl.save()" novalidate>
+<form class="job-editor" name="jobEditor" ng-submit="ctrl.save()" novalidate ng-class="{'job-editor__disabled': !ctrl.userCanEdit}">
     <div class="job-editor__inputs">
         <label class="job-info--editor__field flex-center">
             <div class="job-info--editor__label job-info--editor__label flex-center text-small">Description</div>
@@ -13,13 +13,14 @@
                 ng-change="ctrl.save()"
                 ng-model-options="{updateOn: 'default blur', debounce: { default: ctrl.saveOnTime, blur: 0 }}"
                 ng-class="{ 'job-info--editor__input--with-batch': ctrl.withBatch }"
+                ng-disabled="!ctrl.userCanEdit"
             ></textarea>
 
             <button
                 class="job-editor__apply-to-all"
                 title="Apply this description to all your current uploads"
                 type="button"
-                ng-if="ctrl.withBatch"
+                ng-if="ctrl.withBatch && ctrl.userCanEdit"
                 ng-click="ctrl.batchApplyMetadata('description')"
             >⇔</button>
         </label>
@@ -38,13 +39,14 @@
                 ng-change="ctrl.save()"
                 ng-model-options="{updateOn: 'default blur', debounce: { default: ctrl.saveOnTime, blur: 0 }}"
                 ng-class="{ 'job-info--editor__input--with-batch': ctrl.withBatch }"
+                ng-disabled="!ctrl.userCanEdit"
             />
 
             <button
                 class="job-editor__apply-to-all"
                 title="Apply this byline to all your current uploads"
                 type="button"
-                ng-if="ctrl.withBatch"
+                ng-if="ctrl.withBatch && ctrl.userCanEdit"
                 ng-click="ctrl.batchApplyMetadata('byline')"
             >⇔</button>
 
@@ -65,6 +67,7 @@
                     gr-datalist-input
                     ng-model="ctrl.metadata.credit"
                     ng-change="ctrl.save()"
+                    ng-disabled="!ctrl.userCanEdit"
                     ng-model-options="{ updateOn: 'gr-datalist:update blur' }" />
             </gr-datalist>
 
@@ -72,7 +75,7 @@
                 class="job-editor__apply-to-all"
                 title="Apply this credit to all"
                 type="button"
-                ng-if="ctrl.withBatch"
+                ng-if="ctrl.withBatch && ctrl.userCanEdit"
                 ng-click="ctrl.batchApplyMetadata('credit')"
             >⇔</button>
         </div>
@@ -86,13 +89,14 @@
                 ng-change="ctrl.save()"
                 ng-model-options="{updateOn: 'default blur', debounce: { default: ctrl.saveOnTime, blur: 0 }}"
                 ng-class="{ 'job-info--editor__input--with-batch': ctrl.withBatch }"
+                ng-disabled="!ctrl.userCanEdit"
             />
 
             <button
                 class="job-editor__apply-to-all"
                 title="Apply this copyright to all your current uploads"
                 type="button"
-                ng-if="ctrl.withBatch"
+                ng-if="ctrl.withBatch && ctrl.userCanEdit"
                 ng-click="ctrl.batchApplyMetadata('copyright')"
             >⇔</button>
         </label>
@@ -107,13 +111,14 @@
                 ng-change="ctrl.save()"
                 ng-model-options="{updateOn: 'default blur', debounce: { default: ctrl.saveOnTime, blur: 0 }}"
                 ng-class="{ 'job-info--editor__input--with-batch': ctrl.withBatch }"
+                ng-disabled="!ctrl.userCanEdit"
             />
 
             <button
                 class="job-editor__apply-to-all"
                 title="Apply these instructions to all your current uploads"
                 type="button"
-                ng-if="ctrl.withBatch"
+                ng-if="ctrl.withBatch && ctrl.userCanEdit"
                 ng-click="ctrl.batchApplyMetadata('specialInstructions')"
             >⇔</button>
         </label>
--- a/media-api/app/lib/elasticsearch/ElasticSearch.scala
+++ b/media-api/app/lib/elasticsearch/ElasticSearch.scala
@@ -62,6 +62,18 @@ class ElasticSearch(val config: MediaApiConfig, mediaApiMetrics: MediaApiMetrics
     }
   }
 
+  def getImageUploaderById(id: String)(implicit ex: ExecutionContext): Future[Option[String]] = {
+    implicit val logMarker = MarkerMap("id" -> id)
+    val request = get(imagesAlias, id).fetchSourceInclude("uploadedBy")
+    executeAndLog(request, s"get image uploader by id $id").map { r =>
+
+      r.status match {
+        case Status.OK => r.result.sourceFieldOpt("uploadedBy").collect{case s: String => s}
+        case _ => None
+      }
+    }
+  }
+
   def search(params: SearchParams)(implicit ex: ExecutionContext, request: AuthenticatedRequest[AnyContent, Principal]): Future[SearchResults] = {
     implicit val logMarker = MarkerMap()
     def resolveHit(hit: SearchHit) = mapImageFrom(hit.sourceAsString, hit.id)
--- a/metadata-editor/app/controllers/EditsController.scala
+++ b/metadata-editor/app/controllers/EditsController.scala
@@ -3,12 +3,15 @@ package controllers
 
 import java.net.URI
 import java.net.URLDecoder.decode
-
 import com.amazonaws.AmazonServiceException
+import com.gu.mediaservice.GridClient
 import com.gu.mediaservice.lib.argo.ArgoHelpers
 import com.gu.mediaservice.lib.argo.model._
-import com.gu.mediaservice.lib.auth.Authentication
+import com.gu.mediaservice.lib.auth.Authentication.Request
+import com.gu.mediaservice.lib.auth.Permissions.EditMetadata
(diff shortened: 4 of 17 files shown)


media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -62,6 +62,18 @@ class ElasticSearch(val config: MediaApiConfig, mediaApiMetrics: MediaApiMetrics
...
+    implicit val logMarker = MarkerMap("id" -> id)
+    val request = get(imagesAlias, id).fetchSourceInclude("uploadedBy")
+    executeAndLog(request, s"get image uploader by id $id").map { r =>
+
+      r.status match {
+        case Status.OK => r.result.sourceFieldOpt("uploadedBy").collect{case s: String => s}
+        case _ => None
+      }
+    }
+  }
```
sihil, line 75: Lovely 👍

rest-lib/src/main/scala/com/gu/mediaservice/lib/auth/Authorisation.scala
```diff
@@ -22,13 +24,36 @@ class Authorisation(provider: AuthorisationProvider, executionContext: Execution
...
+      case MachinePrincipal(ApiAccessor(_, Internal), _) => true
+      case _ => false
+    }
+  }
+
+  def canUserWriteMetadata(principal: Principal, uploadedBy: String): Boolean =
+    isUploaderOrHasPermission(principal, uploadedBy, EditMetadata)
+
+  def canUserDeleteImage(principal: Principal, uploadedBy: String): Boolean =
+    isUploaderOrHasPermission(principal, uploadedBy, DeleteImage)
```
sihil, line 55: Super happy for `isUploaderOrHasPermission` to be here but I'm quite keen to avoid embedding anything specific to a particular permission in this class. Can these wrapper methods be colocated in the places they are used?

metadata-editor/app/controllers/EditsController.scala
```diff
@@ -39,12 +43,43 @@ import scala.concurrent.{ExecutionContext, Future}
...
+          isAuthorised = authorisation.canUserWriteMetadata(request.user, uploadedBy.getOrElse(""))
+          if isAuthorised
+        } yield {
+          Future.successful(None)
+        }
+        result.flatten.recover{case _ => Some(authorisation.unauthorized(EditMetadata))}
+      }
+    }
+    override protected def executionContext: ExecutionContext = ec
+  }
```
sihil, line 80: This is really nice. 💯 

Without too much work I imagine that this could be made generic and moved into the Authorisation class. Perhaps an `actionFilterForUploaderOr` method? Either the method or `Authorisation` class with need a grid client.

sihil (commented):
Thanks @adrielulanovsky - aside from the specific permissions in Authorisation I'm happy to merge this (after testing). Making the ActionFilter generic would be a bonus but by no means necessary.

metadata-editor/app/controllers/EditsController.scala
adrielulanovsky: I didn't see the use of making this generic, as the only other permission where the uploader of the image matters is DeleteImage, and the deletion of images is done directly in media-api service, so no need to do an external call. However, I may be missing some use case for that, and I'm happy to make this generic if you prefer.

rest-lib/src/main/scala/com/gu/mediaservice/lib/auth/Authorisation.scala
adrielulanovsky: Yes, I moved them here because canUserWriteMetadata is used by both media-api and metadata-editor services, but I can move the wrappers back to media-api and use isUploaderOrHasPermission directly in metadata-editor (or maybe move them to a common class? not sure where...)

metadata-editor/app/controllers/EditsController.scala
adrielulanovsky: I added the refactor as a new commit, showing how it could be used for DeleteImage. However, I didn't use it for that endpoint, since that one already retrieves the whole image to check for syndication rights, so it seemed kind of wasteful to do another query to elasticSearch to get the uploader of the image.

rest-lib/src/main/scala/com/gu/mediaservice/lib/auth/Authorisation.scala
```diff
@@ -22,13 +25,61 @@ class Authorisation(provider: AuthorisationProvider, executionContext: Execution
...
+                                         uploadedBy: String,
+                                         permission: SimplePermission
+                                       ) = {
+    principal match {
+      case user: UserPrincipal =>
+        if (user.email.toLowerCase == uploadedBy) {
+          true
+        } else {
+          hasPermissionTo(permission)(principal)
+        }
```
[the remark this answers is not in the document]
adrielulanovsky: Fixed

rest-lib/src/main/scala/com/gu/mediaservice/lib/auth/Authorisation.scala
adrielulanovsky: Fixed

AWare:
How does this look when it fails in the UI? 
To beat a dead horse, can you get an image in your session upload queue that would fail in this way- and if so, would the user know if an update failed as part of a batch update on that page.

adrielulanovsky:
> How does this look when it fails in the UI?
> To beat a dead horse, can you get an image in your session upload queue that would fail in this way- and if so, would the user know if an update failed as part of a batch update on that page.

Well, the only way I can think of where this could fail in the UI is if an image is loaded when a user has permission to edit metadata (so the page is loaded with the option to edit the metadata) and then the permission is disallowed before the user sends the edit. In all other cases, I believe the UI won't even show the option to edit the metadata of the image. In the former case, there is no error handling in the frontend (should we take that scenario into account?).

I'm not sure what you mean by batch updates. Could you clarify? All the uses I've seen where metadata is edited so far are individual updates on an image, not a batch process... Am I missing something?

AWare:
If you upload a batch of images at the same time, you can use the 
image button to copy that value to all the images you've uploaded in the current session.

adrielulanovsky:
> If you upload a batch of images at the same time, you can use the
> image button to copy that value to all the images you've uploaded in the current session.

Oh I see, I wasn't aware of that functionality... Anyway, I don't think it will be an issue, since if you uploaded the image then you'll be able to edit the metadata no matter what permissions you have.

sihil:
> Oh I see, I wasn't aware of that functionality... Anyway, I don't think it will be an issue, since if you uploaded the image then you'll be able to edit the metadata no matter what permissions you have (we are checking if you have the permission OR if you're the uploader of the image).

The corner case here is that you can 'upload' an image that already exists and can currently edit metadata for that image by virtue of the fact that we do not correctly check your permissions. @paperboyo doesn't think this is used by many people, but it's used by a non-zero number of people so this will break their workflows. A workflow that we absolutely shouldn't support, but one that none the less works.

adrielulanovsky:
> > Oh I see, I wasn't aware of that functionality... Anyway, I don't think it will be an issue, since if you uploaded the image then you'll be able to edit the metadata no matter what permissions you have (we are checking if you have the permission OR if you're the uploader of the image).
> 
> The corner case here is that you can 'upload' an image that already exists and can currently edit metadata for that image by virtue of the fact that we do not correctly check your permissions. @paperboyo doesn't think this is used by many people, but it's used by a non-zero number of people so this will break their workflows. A workflow that we absolutely shouldn't support, but one that none the less works.

So, should we do something about it then? Display an error message when the metadata-editor controller fails? I'm not sure what we could do for the batch process though, if some fail and some don't... Perhaps show indicators on the failed images based on the id?

AWare:
The batching is done by scope messaging in the upload page, so if the image detail page handled it correctly, then it should _just work_ on upload. Sorry this is such a faff.

adrielulanovsky:
After conversation with @AWare , it was decided that the best approach would be to intercept 403 responses in Kahuna and show an appropriate error message. I attach an image for reference:

image

paperboyo:
> would be to intercept 403 responses in Kahuna and show an appropriate error message

The error in the screenshot is a global one. It requires the whole page to be reloaded, so that the batch edits on Upload page stop working completely. 

IIUC, this is to signpost to the user that **at least one** of the batch of images being uploaded
- already exist in Grid and
- was uploaded by somebody else, so that current user will be unable to edit its metadata (incl. as a part of batch edit to the current upload batch)

Because of that, I think the error should not be global (all other images can be edited and batch edited). And it should not cause the user to have to reload the page.

So I think the best approach would be to reuse image-specific error message, instead of a global one. Here is a crappy screenshot to illustrate what I mean. It would be shown only for affected images (all others would appear and work as is):

image
[note that the error is displayed next to the affected images only and that the metadata section is greyed out for affected images only indicating edits are impossible)

paperboyo:
Thanks for taking care of the UI part of this!

adrielulanovsky:
> > would be to intercept 403 responses in Kahuna and show an appropriate error message
> 
> The error in the screenshot is a global one. It requires the whole page to be reloaded, so that the batch edits on Upload page stop working completely.
> 
> IIUC, this is to signpost to the user that **at least one** of the batch of images being uploaded
> 
>     * already exist in Grid and
> 
>     * was uploaded by somebody else, so that current user will be unable to edit its metadata (incl. as a part of batch edit to the current upload batch)
> 
> 
> Because of that, I think the error should not be global (all other images can be edited and batch edited). And it should not cause the user to have to reload the page.
> 
> So I think the best approach would be to reuse image-specific error message, instead of a global one. Here is a crappy screenshot to illustrate what I mean. It would be shown only for affected images (all others would appear and work as is):
> 
> image
> [note that the error is displayed next to the affected images only and that the metadata section is greyed out for affected images only indicating edits are impossible)

I removed the 403 error interceptor, and added:
- Gray-out and disabling of the Image Metadata section when uploading existing images from other uploader and without EditMetadata permissions. It should be disabled for manual editing and for batch editing as well, so no 403s should happen.
- Warning message above that section alerting the user that the image already exists in the grid uploaded by other user and they don't have permissions to edit the metadata.

Screenshot:
image

kahuna/public/js/upload/jobs/required-metadata-editor.html
```diff
@@ -65,14 +67,15 @@
...
+                    ng-disabled="!ctrl.userCanEdit"
                     ng-model-options="{ updateOn: 'gr-datalist:update blur' }" />
             </gr-datalist>
 
             <button
                 class="job-editor__apply-to-all"
                 title="Apply this credit to all"
                 type="button"
-                ng-if="ctrl.withBatch"
+                ng-if="ctrl.withBatch && ctrl.userCanEdit"
```
AWare, line 78: can we also use `ng-disabled` here?

kahuna/public/js/upload/jobs/required-metadata-editor.css
```diff
@@ -0,0 +1,3 @@
+.job-editor__disabled {
+    opacity: 0.4;
```
AWare, line 2: Appreciate it's more faff, but can we do this by altering the colour instead?

AWare (commented):
Might also be worth greying the image. Otherwise good stuff, and I'm happy to +1 with these small changes.

kahuna/public/js/upload/jobs/required-metadata-editor.css
adrielulanovsky: I'm not sure I understand. Do you mean I find each individual color inside the div and make it darker?

kahuna/public/js/upload/jobs/required-metadata-editor.html
adrielulanovsky: I guess we can, but if the ng-if is false I think it won't matter if it's disabled, is that right? or are you saying we replace the ng-if with ng-disabled? in that case, I think the button will still show, but clicking it won't do anything.

AWare: sorry, my mistake

kahuna/public/js/upload/jobs/required-metadata-editor.css
AWare: yeah, it's only that grey though as you won't have invalid metadata on things already in the grid

AWare (approved):
tested on TEST and am happy, all Simons concerns look dealt with 👍

AWare:
(build looks failed because I canceled a duplicate)

Issue #3218, opened 2021-03-23
[Accessibility] Add "new image" alert to tab in browser


ochiengolanga:
## Short description

When new images arrive matching the search terms a user has in the search field, a "X new" icon appears and is updated as new images arrive.

This notification is within and only visible from the window that is open or to the fore when a user is looking.

To improve this, the new image notification can be extended and made visible in the tab so a user can see it what focus is on a different tab. When a user has multiple tabs of Grid open, each tab should display this.

## Actual results

![Screenshot at 2021-03-23 17-30-19](https://user-images.githubusercontent.com/12212239/112163088-bf747800-8bfd-11eb-83cf-a406825b3fd4.png)

![Screenshot at 2021-03-23 17-44-25](https://user-images.githubusercontent.com/12212239/112165375-bedce100-8bff-11eb-8be1-2211be52ff83.png)

## Expected results

![Screenshot at 2021-03-23 17-14-30](https://user-images.githubusercontent.com/12212239/112163109-c56a5900-8bfd-11eb-9ebf-faea12124a44.png)

![Screenshot at 2021-03-23 17-43-51](https://user-images.githubusercontent.com/12212239/112165359-bab0c380-8bff-11eb-9aa5-6e3299c6d6b9.png)

## Approach

1. Use AngularJS `$rootScope.$emit` to emit `new-images` event when new images are available after polling
2. Have the `uiTitle` directive react to the `new-images` event using the `$rootScope.$on` and modifying the text in the `<title ...>search | the grid</title>` element to `<title ...>(2 new) search | the grid</title>`.

paperboyo:
I think this is simple and useful: 👍 .

PR #3222, opened 2021-03-25
[Accessibility] Add "new image" alert to tab in browser

ochiengolanga:
## What does this change?

Adds a "(X new)"  text to the page title when new images arrive after polling.

Proposal - https://github.com/guardian/grid/issues/3218

When the UI (Kahuna) polls for new images:
- and X new images arrive, then the page title should be updated to "(X new) `page_title`"
- using a user defined search term and X new images arrive, then the page title should be updated to "(X new) `page_title`"

## Screenshots

Tab with new image alert

Screenshot at 2021-03-23 17-14-30

Screenshot at 2021-03-23 17-43-51

## Who should look at this?
@guardian/digital-cms @AWare @paperboyo 

## Tested?
- [X] locally by committer

### DIFF
--- a/kahuna/public/js/main.js
+++ b/kahuna/public/js/main.js
@@ -461,7 +461,16 @@ kahuna.directive('uiTitle', ['$rootScope', function($rootScope) {
                   var title = (titleFunc ? titleFunc(toParams) : toState.name) +
                           ' | ' + attrs.uiTitleSuffix;
 
+                  attrs.uiTitle = title;
+
                   element.text(title);
+            });
+
+            $rootScope.$on('events:new-images',
+                function(event, data) {
+                    var title = (data.count ? `(${data.count} new)  ${attrs.uiTitle}` : attrs.uiTitle);
+
+                    element.text(title);
             });
         }
     };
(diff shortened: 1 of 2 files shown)


kahuna/public/js/main.js
```diff
@@ -461,8 +461,17 @@ kahuna.directive('uiTitle', ['$rootScope', function($rootScope) {
...
                           ' | ' + attrs.uiTitleSuffix;
 
+                  attrs.uiTitle = title;
+
                   element.text(title);
             });
+
+            $rootScope.$on('events:new-images', 
+                function(event, data) {
+                    var title = (data.count ? '(' + data.count + ' new) ' + attrs.uiTitle : attrs.uiTitle);
```
akash1810, line 471: We can use a [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) here to improve the readability of the ternary a little:

```suggestion
                    var title = (data.count ? `(${data.count} new)  ${attrs.uiTitle}` : attrs.uiTitle);
```

akash1810 (commented):
Added a comment to improve readability, LTGM otherwise.

Whilst testing this, I've noticed that polling behaves differently based on what you're doing. That is, polling when filtered to a collection or label doesn't work the same as polling on the empty search page. This isn't blocking this PR, but worth noting.

kahuna/public/js/main.js
ochiengolanga: Thanks for the suggestion on improving readability.

akash1810 (approved):

Issue #3219, opened 2021-03-23
[Accesibility] Navigate to quick filter buttons


Asmau1234:
## Short description

The Quick Filter buttons appear to be not included in the tab order (when a user clicks on the tab key, it skips the “Newest first” and “Oldest first” filters) or accessible to the assistive technology. An improvement is being suggested to make the Grid UI more accessible/user friendly.

## Steps to reproduce

When a user clicks on the tab key, the cursor lands on other buttons on the screen except the quick filter buttons. 

## Actual result

When user hits the tab key, the filters are not selected:

![Screenshot at 2021-03-22 16-17-10](https://user-images.githubusercontent.com/6728165/112168854-f2187380-8bf1-11eb-87b0-499504c75969.png)

## Expected result

When user hits the tab key, the filters are selected:

<img width="159" alt="Screenshot 2021-03-24 at 11 26 07" src="https://user-images.githubusercontent.com/6728165/112295110-debde480-8c93-11eb-9ba8-dcd3e65147df.png">

## Approach

When a user navigates the Grid UI, the quick filter buttons will be included in the tab order or be accessible to the assistive technology by adding a visible highlight to the “Newest first” and “Oldest first” filters. When a user uses the tab index navigation, the cursor would be able to land on these labels.

paperboyo:
Thank you, it’s great to see Grid more keyboard-friendly!

paperboyo:
Fixed by https://github.com/guardian/grid/pull/3223? Closing.

PR #3223, opened 2021-03-25
Can use the tab key to navigate to quick filter buttons

Asmau1234:
## What does this change?

This references the proposal [3219](https://github.com/guardian/grid/issues/3219). Ensures the UI is more accessible and responsive to the assistive technology. This adds a tab index navigation on the quick filter buttons (Newest First, Oldest First) so when users use the tab key to navigate the UI, the cursor is able to land on these buttons.

When a user clicks on the tab key, the cursor should be able to land on the quick filter buttons (Newest First, Oldest First) visibly.

## Screenshots

Screenshot at 2021-03-22 16-17-10
Screenshot 2021-03-24 at 11 26 07

## Who should look at this?
@AWare , @sihil , @paperboyo and @akash1810

## Tested?
- [Y] locally by committer

### DIFF
--- a/kahuna/public/js/search/query.html
+++ b/kahuna/public/js/search/query.html
@@ -86,6 +86,7 @@
             <li class="search__modifier-item radio-list">
                 <div class="radio-list__item">
                     <input type="radio"
+                           aria-hidden="false"
                            id="sort-direction__desc"
                            class="radio-list__circle"
                            name="sort-direction"
@@ -100,6 +101,7 @@
                 <div ng-if="!searchQuery.collectionSearch"
                      class="radio-list__item">
                     <input type="radio"
+                           aria-hidden="false"
                            id="sort-direction__asc"
                            class="radio-list__circle"
                            name="sort-direction"
--- a/kahuna/public/stylesheets/main.css
+++ b/kahuna/public/stylesheets/main.css
@@ -2343,7 +2343,17 @@ FIXME: what to do with touch devices
 }
 
 .radio-list__circle {
-    display: none;
+    clip: rect(0 0 0 0);
+    clip-path: inset(100%);
+    height: 1px;
+    overflow: hidden;
+    position: absolute;
+    white-space: nowrap; 
+    width: 1px;
+}
+
+.radio-list__circle:focus + .radio-list__label {
+    border: solid 3px #00adee;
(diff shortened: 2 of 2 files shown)


Asmau1234:
> Can the `tabindex` and `:hover` be added to the checkbox instead of the div?

There is no checkbox rather, a hidden radio button so the `tabindex` cannot navigate to a hidden radio button. I have implemented a better version of this (`tabindex` on the label which is used to trigger the sort order and a key-press function) that does not affect other parts of the application (the hidden radio button is used in other parts of the app and i did not want to change those). I would love your review.

kahuna/public/js/search/query.html
```diff
@@ -91,7 +91,7 @@
                            name="sort-direction"
                            ng-value="undefined"
```
AWare, line 92: I appreciate this is legacy, but I think `"undefined"` should be replaced with "newest" or something throughout.

kahuna/public/js/search/query.html
```diff
@@ -91,7 +91,7 @@
...
                            ng-model="searchQuery.ordering.orderBy">
-                    <label for="sort-direction__desc" class="radio-list__label"
+                    <label tabindex="0" ng-keypress="accesibilityKey($event, 'undefined')" for="sort-direction__desc" class="radio-list__label"
```
AWare, line 94: I think this should be possible without custom js. 

Using https://www.sarasoueidan.com/blog/inclusively-hiding-and-styling-checkboxes-and-radio-buttons/ should allow the checkboxes to be hidden while preserving tab order. (Not using display:none) And then I think https://www.a11ymatters.com/pattern/checkbox/ would help with the focus style, using the `:focus` selector and then using that to get to the label.

Asmau1234: I have applied the requested changes following the suggested blog post approaches. Shout out to the writers.

kahuna/public/js/search/query.html
Asmau1234: I agree with you however, this is out of scope of this PR. I think it will be better if i implemented this in a separate PR.

AWare (approved):
nice, 
sorry for the delay, have tested in TEST and I'm really happy with it

Issue #3221, opened 2021-03-24
Accessibility: Introduce an edit field to Date Picker


abdelrahmansd:
## Short description
The Date Pickers don't appear to have an edit field to allow users to enter the dates. so it is a little bit hard and unusable to some users.
 An improvement is being suggested to introduce an edit field to allow the user to easily enter the dates

## Actual results

![Screenshot from 2021-03-23 18-20-30](https://user-images.githubusercontent.com/33189781/112274988-c5af3680-8c87-11eb-8ff8-ef4696402f65.png)

## Expected results

![Screenshot from 2021-03-23 19-14-52](https://user-images.githubusercontent.com/33189781/112275048-d6f84300-8c87-11eb-9fce-b87c2aa600d0.png)

## Approach

When the user click on the Date Filter and click on the Filter button
The _From Date_ value will be taken from the _From Edit field_ or fall back to the _From Date Picker_
And also The _To Date value_ will be taken from the _To Edit field_ or fall back to the _To Date Picker_

akash1810:
This is wonderful!

A couple of notes to consider:
- The calendar should sync with the text input
- It might make sense for the "clear" button to live inside/next to the text input
- There are currently three date pickers in the repo (sorry!):
  1. [`<gu-date>`](https://github.com/guardian/grid/blob/b76410e9771e22505967e9d16ed43f1de7aa154f/kahuna/public/js/components/gu-date/gu-date.html)
  2. [`<gu-date-range-x>`](https://github.com/guardian/grid/blob/b76410e9771e22505967e9d16ed43f1de7aa154f/kahuna/public/js/components/gu-date/gu-date-range-x.html)
  3. [`<gu-date-range>`](https://github.com/guardian/grid/blob/b76410e9771e22505967e9d16ed43f1de7aa154f/kahuna/public/js/components/gu-date-range/gu-date-range.html)
  As hinted at in [this PR](https://github.com/guardian/grid/pull/2295) `<gu-date-range-x>` was meant to replace `<gu-date-range>` entirely, however that work wasn't completed for some reason! Ideally we'd finish this first, but I don't think we'd reject a PR that copy pastes these accessibility changes into multiple files.

paperboyo:
Fixed by https://github.com/guardian/grid/pull/3225? Closing.

PR #3225, opened 2021-03-25
Introduce an edit field to Date Picker in quick filters and leases

abdelrahmansd:
## What does this change?
this makes the Datepicker's edit field visible so it allows the user to easily enter the dates

User can type the date in the Datepicker's text input and the calendar should sync with the text input

## Screenshots

Screenshot from 2021-03-26 03-50-21

## Tested?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/components/gu-date-range/gu-date-range.html
+++ b/kahuna/public/js/components/gu-date-range/gu-date-range.html
@@ -56,6 +56,7 @@
         <div class="gu-date-range__overlay__pikaday">
             <span class="gu-date-range__overlay__pikaday__wrapper">
                 <h2 class="gu-date-range__overlay__title search__overlay__title">From</h2>
+                <input type="text" class="gu-date-range__input__start--hidden date-edit-field" placeholder="DD-MM-YYYY" ng-model="pikaStartValue">
                 <div class="gu-date-range__overlay__pikaday__container gu-date-range__overlay__pikaday--start"/>
                 <button type="button"
                         class="gu-date-range__overlay__pikaday__clear button-shy"
@@ -66,6 +67,7 @@
 
             <span class="gu-date-range__overlay__pikaday__wrapper">
                 <h2 class="gu-date-range__overlay__title search__overlay__title">To</h2>
+                <input type="text" class="gu-date-range__input__end--hidden date-edit-field" placeholder="DD-MM-YYYY"  ng-model="pikaEndValue">
                 <div class="gu-date-range__overlay__pikaday__container gu-date-range__overlay__pikaday--end"/>
                 <button type="button"
                         class="gu-date-range__overlay__pikaday__clear button-shy"
@@ -90,7 +92,4 @@
             </button>
         </div>
     </div>
-
-    <input hidden type="text" class="gu-date-range__input__start--hidden" ng-model="pikaStartValue">
-    <input hidden type="text" class="gu-date-range__input__end--hidden" ng-model="pikaEndValue">
 </div>
--- a/kahuna/public/js/components/gu-date/gu-date.html
+++ b/kahuna/public/js/components/gu-date/gu-date.html
@@ -17,6 +17,6 @@
 
     <div ng-show="showingOverlay" class="gu-date__container"></div>
 
-    <input hidden type="text" class="gu-date__value--hidden" ng-model="pikaValue">
+    <input type="text" class="gu-date__value--hidden" placeholder="DD-MM-YYYY" ng-model="pikaValue">
 </div>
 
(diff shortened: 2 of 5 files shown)


abdelrahmansd:
Hi @AWare Thanks for the review, I think this will be different than the [proposal](https://github.com/guardian/grid/issues/3221),
@akash1810 what do you think?

akash1810:
@AWare IIUC those links suggest making dates more accessible by making the input more human readable? Sounds like a great idea to me and would suggest that as a different change. I don't think it blocks this PR.

On a related note, Grid already has some [limited support for relative dates](https://github.com/guardian/grid/blob/4813284f701cede49b4465bacc86817877c3a43e/media-api/test/scala/lib/querysyntax/ParserTest.scala#L195-L235), joining this new field to that query string might be possible.

The changes here look good, I'll put them on our pre-PROD environment to test them and will merge once I have.

Thanks for the contribution and apologies my late review!

kahuna/public/js/components/gu-date-range/gu-date-range.html
```diff
@@ -56,6 +56,7 @@ <h2 class="gu-date-range__overlay__title search__overlay__title">Presets</h2>
         <div class="gu-date-range__overlay__pikaday">
             <span class="gu-date-range__overlay__pikaday__wrapper">
                 <h2 class="gu-date-range__overlay__title search__overlay__title">From</h2>
+                <input hidden type="text" class="gu-date-range__input__start--hidden date-edit-field" placeholder="DD-MM-YYYY" ng-model="pikaStartValue">
```
akash1810, line 59: Is this meant to be hidden?

kahuna/public/js/components/gu-date-range/gu-date-range.html
```diff
@@ -66,6 +67,7 @@ <h2 class="gu-date-range__overlay__title search__overlay__title">From</h2>
 
             <span class="gu-date-range__overlay__pikaday__wrapper">
                 <h2 class="gu-date-range__overlay__title search__overlay__title">To</h2>
+                <input hidden type="text" class="gu-date-range__input__end--hidden date-edit-field" placeholder="DD-MM-YYYY"  ng-model="pikaEndValue">
```
akash1810, line 70: Is this meant to be hidden?

kahuna/public/js/components/gu-date/gu-date.html
```diff
@@ -17,6 +17,6 @@
 
     <div ng-show="showingOverlay" class="gu-date__container"></div>
 
-    <input hidden type="text" class="gu-date__value--hidden" ng-model="pikaValue">
+    <input type="text" class="gu-date__value--hidden" ng-model="pikaValue">
```
akash1810, line 20: Should we add a placeholder here?

akash1810 (commented):
Have deployed this to our pre-PROD environment now. The screenshot of the change doesn't match what we're seeing.

We're seeing this on the `gu-date-range` component (no date input):
image

This is down to the `hidden` attribute.

For the `gu-date` we're seeing this:
image

I think it would be preferable for the input field to be shown in the date modal, rather than at the top level of the component.

paperboyo:
@abdelrahmansd This looks like it could support a user request that we had here: for Leases to allow choosing hours (not only days). Similarly, it could be useful for setting ranges for search? Of course, only if that is indeed possible, which I judge *purely* by seeing this user-hostile looking UTC datetime string (we would need to have it formatted in a nice way).

Disregard this comment if this isn’t easy!

kahuna/public/js/components/gu-date-range/gu-date-range.html
abdelrahmansd: fixed

kahuna/public/js/components/gu-date-range/gu-date-range.html
abdelrahmansd: fixed

kahuna/public/js/components/gu-date/gu-date.js
```diff
@@ -46,6 +45,7 @@ guDate.directive('guDate', [function () {
             const input = root.querySelector('input.gu-date__value--hidden');
             const container = root.querySelector('.gu-date__container');
 
+            const pikaFieldFormat = 'DD-MM-YYYY';
```
akash1810, line 48: Can I suggest using `DISPLAY_FORMAT` defined above on L10? The `DD-MM-YYYY` format can be a little ambiguous for our American friends - is 03/04/2021 03 April or 04 March?

`DISPLAY_FORMAT` is `DD MMM YYYY` which will render 03/04/2021 as 03 Mar 2021, which removes the ambiguity.

akash1810 (commented):
Looks great! A nitpick comment on the date format, which applies to each field.

kahuna/public/js/components/gu-date/gu-date.js
abdelrahmansd: yes, make sense

akash1810 (approved):
Have tested these changes on our pre-production environment and they look good to me 👍🏽 .

Will merge in the morning.

Apologies for the delay and thanks for your patience!

PR #3226, opened 2021-03-25
Implement local providers for authentication and authorisation

adrielulanovsky:
## What does this change?
This PR implements a LocalAuthenticationProvider and a LocalAuthorisationProvider which lets everyone through anything with a generic user, for simplicity in local development. It also changes the setup script so that the application.conf file has the correct local providers when running locally.

As a developer, when I'm running the grid locally I don't need to authenticate, and I have permissions for everything.

## Who should look at this?
@guardian/digital-cms @sihil @gribeiro @wainaina 

## Tested?
- [x] locally by committer

### DIFF
--- a/dev/script/setup.sh
+++ b/dev/script/setup.sh
@@ -79,65 +79,6 @@ setupDevNginx() {
   dev-nginx setup-app "$target"
 }
 
-setupPanDomainConfiguration() {
-  if [[ $LOCAL_AUTH != true ]]; then
-    return
-  fi
-
-  echo "setting up pan domain authentication configuration for local auth"
-
-  panDomainBucket=$(getStackResource "$AUTH_STACK_NAME" PanDomainBucket)
-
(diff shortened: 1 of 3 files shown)


AWare:
I think this might be worth making optional.

adrielulanovsky:
> I think this might be worth making optional.

We could use an environment variable in the setup script to use either this one or the previous one. Do you think the current one should be the default?

AWare:
> We could use an environment variable in the setup script to use either this one or the previous one. Do you think the current one should be the default?

This seems reasonable.

adrielulanovsky:
> I think this might be worth making optional.

I added checks for LOCAL_SIMPLE_AUTH_PROVIDER env variable. If that variable is true, the setup script will set the local authentication and authorisation providers. Otherwise, it will behave as it used to.

AWare (approved):

AWare (approved):
tested on TEST and am happy with it

PR #3236, opened 2021-03-29
some node updates

AWare:
## What does this change?

updates a few of the node sub-projects deps

dependabot is less unhappy

## Tested?
- [ ] locally by committer


dev/script/generate-config/.nvmrc
```diff
@@ -1,2 +1,2 @@
-8
+15
```
sihil, line 1: I'd generally suggest sticking with even number releases as they are supported for far longer than the odd releases. Unless there is a good reason to require 15.

AWare, line 1:
```suggestion
14
```

akash1810 (approved):

PR #3237, opened 2021-03-29
Elasticseach mapping test and good to go check

sihil:
## What does this change?
Previously it was fairly easy to make a change to the Grid's model that made it impossible to serialise to the existing ElasticSearch mappings (an additional field for example). This can be fixed by adapting the mapping in the codebase accordingly BUT this is only applied when there is no index so established installations will still fail.

This PR makes evolving the model lower risk by:
 - Adding some unit tests to test serialising and deserialising the Image model into ES using the mappings in the code base
 - Adding a good to go check to Thrall's startup that tries to serialise and deserialise into the configured ES database

These help a developer to catch evolutions when they are made prior to spinning it up or deploying it. They also ensure that operational errors are quickly detected, such as deploying a build with a model that is incompatible with the current index mappings in ES. Previously we would only have a failure when an edit to the altered part of the model was made, now thrall will fail fast.

An upcoming change will add a new `lastModified` field to the Edits model. This change should be picked up immediately by these tests and good to go check to ensure that appropriate mapping changes are made locally and to the deployed databases.

## Tested?
- [X] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/MappingTest.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/MappingTest.scala
@@ -51,7 +51,7 @@ object MappingTest {
     secureUrl = Some(new URL("http://host/filename.jpg"))
   )
 
-  val testUploader = "grid-internal-mapping-test-image"
+  val testUploader = "grid-internal-mapping-test-image" // Do not change this, we use it to clean up old test images
 
   val testImage: Image = Image(
     id = "abcdef1234567890",
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/MappingTest.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/MappingTest.scala
@@ -1,0 +1,227 @@
+package com.gu.mediaservice.lib.elasticsearch
+
+import com.gu.mediaservice.model.leases.{AllowUseLease, LeasesByMedia, MediaLease}
+import com.gu.mediaservice.model.usage._
+import com.gu.mediaservice.model._
+import org.joda.time.{DateTime, Period}
+import play.api.libs.json.{JsArray, JsNumber, JsString}
+
+import java.net.{URI, URL}
+
+/**
+  * This object contains a test image that is fully populated (i.e. all fields of a collection
+  * type including options have a value rather than None, Nil or other empty alternative).
+  * This is used to test that the elasticsearch mapping is still correct when changes are
+  * made to the model.
+  */
+object MappingTest {
+  // various dates for use below - these are deliberately a long time ago in case we accidentally leave anything lying
+  // around
+  private val imageTaken: DateTime = new DateTime(2010, 3, 26, 12, 0)
+  private val imageImported: DateTime = imageTaken.plus(Period.days(1).plusHours(1))
+  private val imageModified: DateTime = imageImported.plus(Period.hours(2))
+
+  private val testImageMetadata: ImageMetadata = ImageMetadata(
+    dateTaken = Some(imageTaken),
+    description = Some("a very lovely photograph"),
+    credit = Some("Bill Withers"),
+    creditUri = Some("https://bill.example.com"),
+    byline = Some("Bill and Friends"),
+    bylineTitle = Some("Bill"),
+    title = Some("Lovely day"),
+    copyright = Some("Bill 1999"),
+    suppliersReference = Some("ASDFFF"),
+    source = Some("bill-withers"),
+    specialInstructions = Some("This is really important."),
+    keywords = List("sunset", "loveliness", "distance"),
+    subLocation = Some("Downtown"),
+    city = Some("LA"),
+    state = Some("California"),
+    country = Some("USA"),
+    subjects = List("Bill", "Friends"),
+    peopleInImage = Set("Bill", "Friends", "Other friends")
+  )
+
+  private val testAsset: Asset = Asset(
+    file = new URI("file://filename.jpg"),
+    size = Some(12345L),
+    mimeType = Some(Jpeg),
+    dimensions = Some(Dimensions(1000, 2000)),
+    secureUrl = Some(new URL("http://host/filename.jpg"))
+  )
+
+  val testImage: Image = Image(
+    id = "abcdef1234567890",
+    uploadTime = imageImported,
+    uploadedBy = "Monkey",
+    lastModified = Some(imageModified),
+    identifiers = Map("id1" -> "value1"),
+    uploadInfo = UploadInfo(
+      filename = Some("filename.jpg")
+    ),
+    source = testAsset,
+    thumbnail = Some(Asset(
+      file = new URI("file://thumb.jpg"),
+      size = Some(1234L),
+      mimeType = Some(Jpeg),
+      dimensions = Some(Dimensions(500, 1000)),
+      secureUrl = Some(new URL("http://host/thumb.jpg"))
+    )),
+    optimisedPng = Some(Asset(
+      file = new URI("file://filename.png"),
+      size = Some(1245L),
+      mimeType = Some(Png),
+      dimensions = Some(Dimensions(1000, 2000)),
+      secureUrl = Some(new URL("http://host/filename.jpg"))
+    )),
+    fileMetadata = FileMetadata(
+      iptc = Map("iptc1" -> "value1"),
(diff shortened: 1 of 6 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/MappingTest.scala
```diff
@@ -0,0 +1,227 @@
...
+        downloadedBy = "me"
+      ))
+    )),
+    leases = LeasesByMedia.build(List(
+      MediaLease(
+        id = Some("leaseA"),
+        leasedBy = Some("you"),
+        startDate = Some(imageTaken),
+        endDate = Some(imageModified),
+        access = AllowUseLease,
```
AWare, line 182: Would making this a deny lease reduce the chance of this image flashing up in the UI?

AWare (approved):

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/MappingTest.scala
paperboyo: Not really, it’s No rights already, so will only show up when one unchecks Free to use. The best way, if possible, and should apply to Integration Tests’s image too!, would be to mock uploadTime to, like, 3 years ago.

sihil: It's no longer NoRights, but all the dates are in 2010 so it shouldn't appear as a recent image. 🤞 

I am very happy to make this a deny lease. I suspect that if this image were to be returned to the client, the client would potentially be quite unhappy.

AWare: Might be worth making it anodyne for the client just in case.

thrall/app/lib/elasticsearch/GoToGoCheck.scala
```diff
@@ -0,0 +1,63 @@
...
+import com.gu.mediaservice.lib.logging.LogMarker
+import com.sksamuel.elastic4s.Response
+import com.sksamuel.elastic4s.requests.delete.DeleteResponse
+import com.typesafe.scalalogging.StrictLogging
+
+import scala.concurrent.{ExecutionContext, Future}
+
+/**
+  * This implements a good to go check that validates when Thrall starts up that it can successfully write and read
+  * and image into the elasticsearch node.
```
aug24, line 14:
```suggestion
  * an image into the elasticsearch node.
```

thrall/app/lib/elasticsearch/GoToGoCheck.scala
```diff
@@ -0,0 +1,63 @@
...
+    // make a test image with an ID that is unique to the node so that multiple nodes can run the check at once
+    val uniqueId = Option(EC2MetadataUtils.getInstanceId).map(_.stripPrefix("i-")).getOrElse("abcdef")
+    val image = MappingTest.testImage.copy(id = s"${MappingTest.testImage.id}$uniqueId")
+    val lastModified = image.lastModified.get
+    implicit val marker: LogMarker = image ++ Map("goToGo" -> true)
+
+    val indexResult: Future[Unit] =
+      for {
+        // perhaps a test image wasn't previously cleaned up, let's err on the side of caution and check and deal with it
+        _ <- Future.successful(logger.info(s"Ensuring test image ${image.id} doesn't exist"))
```
aug24, line 27: Vaguely concerned about races here.  Either they have unique ids and we don't need to do this, or they don't and we have a possibility of two new thralls fighting.  Unless this is more oriented to local testing?  If so, perhaps a comment to that effect?

thrall/test/lib/elasticsearch/ElasticSearchTest.scala
```diff
@@ -1,17 +1,17 @@
 package lib.elasticsearch
 
-import java.util.UUID
-
+import com.gu.mediaservice.lib.elasticsearch.MappingTest
```
aug24, line 3: Are these real changes or refactoring artifacts?  Looks like the whole file could be removed from this PR?

thrall/test/lib/elasticsearch/ImageModelTest.scala
```diff
@@ -0,0 +1,64 @@
...
+       * If this fails then it indicates that a field in the Image model does not contain a value.
+       * This should be fixed by adding appropriate values in MappingTest.scala
+       */
+
+      /* recurse through a typical scala structure */
+      def check(maybeProduct: Any): List[Either[String, Any]] = {
+        maybeProduct match {
+          case None =>
+            List(Left("Empty Option"))
+          case list: TraversableLike[_, _] if list.isEmpty =>
```
aug24, line 25: Couldn't this be `list: Nil`?

thrall/test/lib/elasticsearch/ImageModelTest.scala
```diff
@@ -0,0 +1,64 @@
...
+      val image = MappingTest.testImage
+
+      Await.result(Future.sequence(ES.indexImage(image.id, image, image.lastModified.get)), fiveSeconds)
+      eventually(timeout(fiveSeconds), interval(oneHundredMilliseconds))(reloadedImage(image.id).map(_.id) shouldBe Some(image.id))
+
+      val retrievedImage = reloadedImage(image.id).get
+      /* note that we compare the toString method here as the Joda serialisation/deserialisation ends up with
+       * equivalent by technically different dates due to different ways of internally expressing the chronology
+       * see https://stackoverflow.com/questions/21002385/datetime-does-not-equal-itself-after-unserialization */
+      retrievedImage.toString shouldBe image.toString
```
aug24, line 61: This is nice.  No special cases at all!

aug24 (changes requested):

thrall/app/lib/elasticsearch/GoToGoCheck.scala
sihil: I did indeed hit this when doing local testing and at that point introduced the deletion and shortly after also aded unique IDs so that there were not races between nodes. You're quite right that this won't do a lot - unless thrall is restarting in a loop on the same node??? I will do some more thinking.

thrall/test/lib/elasticsearch/ImageModelTest.scala
sihil: Nil is specifically an empty `List` so wouldn't match the other types of traversables. See https://github.com/scala/scala/blob/2.13.x/src/library/scala/collection/immutable/List.scala#L628-L640

sihil: To be fair, the LHS shouldn't be named `list`. Will change to `collection` or something.

thrall/app/lib/elasticsearch/GoToGoCheck.scala
sihil: I've added a bit of extra context here to describe when this might happen.

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/MappingTest.scala
sihil: Confirmed that the images don't break the client. They don't work properly, but they don't make it upset.

sihil:
I've added a cleanup process as well, to ensure that any old test images are not left lying around. This is run as part of the good to go check. I'm somewhat in two minds about it as it uses a delete by query. Were this to be run without a query then it would obviously delete all of the images in the database. There is an integration test run against ES to prove that the query doesn't delete unrelated images or current test images so this should be fine but other ideas are also welcome.

thrall/test/lib/elasticsearch/ImageModelTest.scala
aug24: I looked into this.... out of interest.  Nil will work for any *Sequence*, but not any *Traversable*.  This is obviously dumb, as an empty traversable (ie no fixed order) is clearly the same as an empty sequence (ie with fixed order).

For your delight:

```
scala> var x = Map()
x: scala.collection.immutable.Map[Nothing,Nothing] = Map()

scala> x.toSeq match { case Nil => println("empty"); case _ => println("other"); }
empty

scala> var x = None
x: None.type = None

scala> x.toSeq match { case Nil => println("empty"); case _ => println("other"); }
empty

scala> var x = Some("banana")
x: Some[String] = Some(banana)

scala> x.toSeq match { case Nil => println("empty"); case _ => println("other"); }
other
```

aug24 (approved):

PR #3260, opened 2021-04-12
Remove conversion to UTC in DateTime parsers

adrielulanovsky:
## What does this change?
Currently there's an automatic conversion to UTC DateTimes when parsing them. This causes GoodToGoCheck to fail on thrall when the server time is in a different timezone than UTC, since the created objects have a local DateTimeZone, but when written into elasticSearch and then read back, the time is converted to UTC, causing the objects to be different.

This fixes it by removing the conversion on parsing.

Thrall service doesn't fail when the server uses a different timezone than UTC.

## Who should look at this?
@sihil 

## Tested?
- [x] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/formatting/package.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/formatting/package.scala
@@ -15,11 +15,10 @@ package object formatting {
     ).map(_.getParser)
     new DateTimeFormatterBuilder().
       append(null, parsers).
-      toFormatter.
-      withZoneUTC
+      toFormatter
   }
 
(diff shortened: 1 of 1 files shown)


aug24:
The change which introduced the UTC zone was deliberately introduced to simplify date handling, by ensuring that all dates were consistently persisted in UTC.

All writes to ES should also be forcing to UTC.  I would suggest that a better answer is to locate the relevant write for this issue and change it to push a UTC time zone into ES.

Any time zoning for display purposes should be done explicitly at display time.

sihil:
@adrielulanovsky can we close this now that the issue at hand has been resolved with https://github.com/guardian/grid/pull/3263 ?

adrielulanovsky:
Closing this as the issue with Thrall has already been sorted.

PR #3261, opened 2021-04-12
time honoured tradition of "does it work now?"

AWare:
## Tested?
- [ ] locally by committer


AWare:
this was due to the bintray brownout
there will be more work to move off it properly, but not a rushed job like this

PR #3265, opened 2021-04-13
Permissions on Metadata service

adrielulanovsky:
## What does this change?
This implements permission checks on the following metadata-editor service endpoints for editing metadata:
```

PUT     /metadata/:id/metadata                          controllers.EditsController.setMetadata(id: String)
POST    /metadata/:id/metadata/set-from-usage-rights    controllers.EditsController.setMetadataFromUsageRights(id: String)

```
With this change, if a request to one of those endpoints is sent by a user who is not the uploader of the image AND does not have EditMetadata permission, the request will be denied.

To get the information about the uploader of the image, a new endpoint was implemented in media-api service:
`
GET     /images/:imageId/uploadedBy                   controllers.MediaApi.uploadedBy(imageId: String)
`
This endpoint returns the uploadedBy field of the image as a string. The metadata-editor service will make use of this endpoint to check if the user trying to edit metadata is the original uploader of the image.

Also, this implements some changes in the UI. In the case of uploading an image already in the system with a diffent uploader, and no EditMetadata permissions, the /upload page reflects this by showing a warning on the Image Metadata section of the conflicting image, and by disallowing edits to the metadata of that image. Batch metadata copying should work without any errors, since the batch process has been updated to avoid copying data into images without permissions.

- As a user with EditMetadata permissions, I'm able to edit the metadata of any image.
- As a user without EditMetadata permissions, I'm only able to edit the metadata of the images I uploaded.
- When an image without EditMetadata permissions is uploaded, the user cannot edit the metadata of it, and an appropriate warning is shown in the UI.

## Screenshots
Batch upload of images, one of which is an image with a different uploader, and with no EditMetadata permissions:
image

## Who should look at this?
@guardian/digital-cms @sihil @ochiengolanga 

## Tested?
- [x] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/GridClient.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/GridClient.scala
@@ -172,6 +172,16 @@ class GridClient(services: Services)(implicit wsClient: WSClient) extends LazyLo
     }
   }
 
+  def getUploadedBy(mediaId: String, authFn: WSRequest => WSRequest)(implicit ec: ExecutionContext): Future[Option[String]] = {
+    logger.info("attempt to get uploadedBy")
+    val url = new URL(s"${services.apiBaseUri}/images/$mediaId/uploadedBy")
+    makeGetRequestAsync(url, authFn) map {
+      case Found(json, _) => Some((json \ "data").as[String])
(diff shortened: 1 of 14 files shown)


AWare (approved):
thanks for fixing so quickly 🚀

PR #3271, opened 2021-04-19
Change to triple input streams

aug24:
Co-authored-by: @twrichards 
Co-authored-by: @tjsilver 

## What does this change?
Adds a new input stream for reingestion and changes the original two input stream names to better reflect what they do, resulting in the following in priority order (highest to lowest):

1. `uiSource` (formerly known as 'high priority') - end user actions
2. `automationSource` (formerly known as 'low priority') - e.g. usages, RCS, etc
3. `reingestionSource` (new) 

Everything should continue to process as normal, this is a no-op. There is currently no mechanism to produce messages for the new reingestion source - this will be added in future PRs.

We take messages from the reingestion queue if the other queues are already (near) empty.

## Who should look at this?
@guardian/digital-cms

## Tested?
- [x] locally by committer

### DIFF
--- a/thrall/app/ThrallComponents.scala
+++ b/thrall/app/ThrallComponents.scala
@@ -54,7 +54,7 @@ class ThrallComponents(context: Context) extends GridComponents(context, new Thr
 
   val highPrioritySource: Source[KinesisRecord, Future[Done]] = KinesisSource(highPriorityKinesisConfig)
   val lowPrioritySource: Source[KinesisRecord, Future[Done]] = KinesisSource(lowPriorityKinesisConfig)
-  val reingestionSource: Source[ReingestionRecord, Future[Done]] = ReingestionSource()
+  val reingestionSource: Source[ReingestionRecord, Future[Done]] = ReingestionSource(es)
 
   val thrallEventConsumer = new ThrallEventConsumer(es, thrallMetrics, store, metadataEditorNotifications, new SyndicationRightsOps(es), actorSystem)
   val thrallStreamProcessor = new ThrallStreamProcessor(highPrioritySource, lowPrioritySource, reingestionSource, thrallEventConsumer, actorSystem, materializer)
--- a/thrall/app/ThrallComponents.scala
+++ b/thrall/app/ThrallComponents.scala
@@ -54,12 +54,26 @@ class ThrallComponents(context: Context) extends GridComponents(context, new Thr
   val highPriorityKinesisConfig: KinesisClientLibConfiguration = KinesisConfig.kinesisConfig(config.kinesisConfig)
   val lowPriorityKinesisConfig: KinesisClientLibConfiguration = KinesisConfig.kinesisConfig(config.kinesisLowPriorityConfig)
 
-  val highPrioritySource: Source[KinesisRecord, Future[Done]] = KinesisSource(highPriorityKinesisConfig)
-  val lowPrioritySource: Source[KinesisRecord, Future[Done]] = KinesisSource(lowPriorityKinesisConfig)
+  val uiSource: Source[KinesisRecord, Future[Done]] = KinesisSource(highPriorityKinesisConfig)
+  val automationSource: Source[KinesisRecord, Future[Done]] = KinesisSource(lowPriorityKinesisConfig)
   val reingestionSource: Source[(UpdateMessage, Instant), Future[Done]] = ReingestionSource(???)
 
-  val thrallEventConsumer = new ThrallEventConsumer(es, thrallMetrics, store, metadataEditorNotifications, actorSystem)
-  val thrallStreamProcessor = new ThrallStreamProcessor(highPrioritySource, lowPrioritySource, reingestionSource, thrallEventConsumer, actorSystem, materializer)
+  val thrallEventConsumer = new ThrallEventConsumer(
+    es,
+    thrallMetrics,
+    store,
+    metadataEditorNotifications,
+    actorSystem
+  )
+
+  val thrallStreamProcessor = new ThrallStreamProcessor(
+    uiSource,
+    automationSource,
+    reingestionSource,
+    thrallEventConsumer,
+    actorSystem,
+    materializer
+  )
 
   val streamRunning: Future[Done] = thrallStreamProcessor.run()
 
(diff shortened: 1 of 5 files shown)


AWare:
Had a go at this with @sihil to continue this. Think we should get it ready and merged, but maybe not the reingest source parts.

thrall/app/ThrallComponents.scala
```diff
@@ -53,9 +55,10 @@ class ThrallComponents(context: Context) extends GridComponents(context, new Thr
 
   val highPrioritySource: Source[KinesisRecord, Future[Done]] = KinesisSource(highPriorityKinesisConfig)
   val lowPrioritySource: Source[KinesisRecord, Future[Done]] = KinesisSource(lowPriorityKinesisConfig)
+  val reingestionSource: Source[(UpdateMessage, Instant), Future[Done]] = ReingestionSource(???)
```
twrichards, line 58: think this should become `migrationSource` based on https://github.com/guardian/grid/pull/3324

thrall/app/ThrallComponents.scala
```diff
@@ -53,9 +55,10 @@ class ThrallComponents(context: Context) extends GridComponents(context, new Thr
 
   val highPrioritySource: Source[KinesisRecord, Future[Done]] = KinesisSource(highPriorityKinesisConfig)
   val lowPrioritySource: Source[KinesisRecord, Future[Done]] = KinesisSource(lowPriorityKinesisConfig)
```
twrichards, line 57: we should rename these to reflect their purpose, rather than their priority, e.g...
```suggestion
  val userActionsSource: Source[KinesisRecord, Future[Done]] = KinesisSource(highPriorityKinesisConfig)
  val backgroundTasksSource: Source[KinesisRecord, Future[Done]] = KinesisSource(lowPriorityKinesisConfig)
```

thrall/app/ThrallComponents.scala
twrichards: decided to do this as part of https://github.com/guardian/grid/pull/3324 instead, as a follow-up

thrall/app/lib/ThrallStreamProcessor.scala
```diff
@@ -1,62 +1,101 @@
...
+case object UiPriority extends Priority {
   override def toString = "high"
 }
-case class TaggedRecord(record: KinesisRecord, priority: Priority) extends LogMarker {
-  override def markerContents = Map(
+
+case class TaggedRecord[T](payload: T,
+                           arrivalTimestamp: Instant,
+                           priority: Priority,
+                           markProcessed: () => Unit) extends LogMarker {
```
jonathonherbert, line 32: I like this. Just to check my understanding, I'm guessing we're asking for these properties explicitly, rather than wrapping a `Record`, b/c [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) or similar? I'm torn as to whether a worth a comment to describe what a TaggedRecord is, or why it exists, might be nice.

thrall/app/lib/ThrallStreamProcessor.scala
```diff
@@ -1,62 +1,101 @@
...
+    automationRecordSource  ~> uiAndAutomationRecordsMerge.in(0)
+
+    // parse the kinesis records into thrall update messages (dropping those that fail)
+    val uiAndAutomationMessagesSource: PortOps[TaggedRecord[UpdateMessage]] =
+      uiAndAutomationRecordsMerge.out
+        .map { taggedRecord =>
+          ThrallEventConsumer
+            .parseRecord(taggedRecord.payload, taggedRecord.arrivalTimestamp)
+            .map(
+              message => taggedRecord.copy(payload = message)
```
jonathonherbert, line 81: Looks like `payload` can represent both a raw message and its parsed version. Would it make sense to model the raw and parsed versions separately, to ensure they're never confused by consuming code?

thrall/test/lib/ThrallStreamProcessorTest.scala
```diff
@@ -1,49 +1,91 @@
...
+      // and then the `preferred` inlet until `preferred` is mostly empty.
+      //  Nonetheless the result should be that a high proportion of the highest priority turn up in the first section.
+      val PERCENTAGE: Double = 95.0 / 100.0
+      val MINIMUM_RECORDS: Int = (PERCENTAGE * COUNT_EACH).toInt
 
-      // it looks like MergedPreferred doesn't strictly process the `preferred` inlet and then the remaining inlets
-      // but rather, it appears to take some a number from all defined inlets and then the `preferred` inlet
-      // `alternatingSegment` represents this behaviour
-      val alternatingSegment = Seq(LowPriority, HighPriority, LowPriority)
+      firstBatch.count(p => p == UiPriority) should be > MINIMUM_RECORDS
```
jonathonherbert, line 82: 👍 👍

jonathonherbert (commented):
This looks great, nice to see comments in the test explaining the methodology. A few comments.

thrall/app/lib/ThrallStreamProcessor.scala
tjsilver: It's not just because of the 'principal of least knowledge', but also because `TaggedRecord` now represents not only messages coming from Kinesis but also our new messages that originate within thrall. @

twrichards: code comment added in https://github.com/guardian/grid/pull/3271/commits/b44df43a3f8aa4cdf335e86c2c7b8ceff330f431

thrall/app/lib/ThrallStreamProcessor.scala
twrichards: I see where you're coming from, however as the comment in https://github.com/guardian/grid/pull/3271/commits/b44df43a3f8aa4cdf335e86c2c7b8ceff330f431 explains, `TaggedRecord` is used to represent both Kinesis records and messages originating in thrall which wouldn't have a raw payload (so we'd have to model those as Option which would be messy). The `P` type param of `TaggedRecord` is sufficient for consumers to know whether it's been parsed or not (`Array[Byte]` when raw, and `UpdateMessage` when parsed, compiler will tell the user when it's the wrong shape.

jonathonherbert: Ah yep, good point re: the type param 👍

jonathonherbert (approved):

PR #3273, opened 2021-04-21
Exclude Tom Nicholson from Getty sniffing

paperboyo:
Ronseal [pretend you are reading a detailed description of this change].


AWare (approved):
I'm not quite sure what you meant by "consectetur adipiscing elit", but this looks fine to me.

PR #3277, opened 2021-04-26
Filter for unprocessable (bad) entities

aug24:
## What does this change?
We log at ERROR (and thus set off alarms) when we receive MediaUsage notifications which have an empty mediaId.

This is not helpful, as we can't do anything about them, and the exception propagation means the entire set of updates is abandoned rather than just the bad messages.

```
lib.BadInputException: Empty string received for image id at model.UsageTable.$anonfun$queryByImageId$1(UsageTable.scala:40) 
```

This PR filters out those messages and logs a warning instead.

No more errors, warnings in logs.

## Tested?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/usage/MediaUsage.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/usage/MediaUsage.scala
@@ -28,6 +28,9 @@ case class MediaUsage(
       // remove events from CAPI that represent images previous to Grid existing
       logger.info(s"MediaId $mediaId doesn't look like a Grid image. Ignoring usage $usageId.")
       false
+    } else if (mediaId.trim.isEmpty) {
+      logger.warn("Unprocessable MediaUsage", this)
+      false
     } else {
       true
     }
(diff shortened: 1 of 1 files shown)


jonathonherbert (approved):

common-lib/src/main/scala/com/gu/mediaservice/model/usage/MediaUsage.scala
```diff
@@ -28,6 +28,9 @@ case class MediaUsage(
       // remove events from CAPI that represent images previous to Grid existing
       logger.info(s"MediaId $mediaId doesn't look like a Grid image. Ignoring usage $usageId.")
       false
+    } else if (mediaId.trim.isEmpty) {
+      logger.warn("Unprocessable MediaUsage", this)
```
sihil, line 32: should we say why? 
```suggestion
      logger.warn("Unprocessable MediaUsage, mediaId is empty", this)
```

sihil (approved):
Nice catch. Would be nice to be clearer why we can't do anything, but better not to have errors.

PR #3278, opened 2021-04-26
Switch to new syndication

aug24:
## What does this change?
The old syndication ops in Thrall relied on the elastic search database as a primary source of syndication and photoshoot information.

The correct architecture is for RCS Poller Lambda to send (RESTful) updates to the (new) syndication service within metadata, which is the primary source, and for metadata to send trivial messages to Thrall, which it will simply persist in ES.

RCS Messages from RCS Poller Lambda are ignored, but messages from Syndication are acted upon.

## Tested?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
@@ -157,7 +157,7 @@ class DynamoDB[T](config: CommonConfig, tableName: String, lastModifiedKey: Opti
               val maybeT = (json \ attributeKey).asOpt[T]
               logger.info(s"Obtained a T of $maybeT from json $json")
               maybeT.map(
-                (json \ IdKey).as[String] -> _
+                attributes.get(IdKey).toString -> _
               )
             })
           logger.info(s"Got $results for request")
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
@@ -147,7 +147,9 @@ class DynamoDB[T](config: CommonConfig, tableName: String, lastModifiedKey: Opti
         else {
           logger.info(s"Fetching records for $request")
           val response = client.batchGetItem(request)
-          val results = response.getResponses.get(tableName).asScala.toList
+          val responses = response.getResponses
+          logger.info(s"Got responses of $responses")
+          val results = responses.get(tableName).asScala.toList
             .flatMap(att => {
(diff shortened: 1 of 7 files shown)


AWare (approved):

metadata-editor/app/lib/Syndication.scala
```diff
@@ -1,67 +1,82 @@
...
-                                (implicit ec: ExecutionContext): Future[Unit.type] = {
+  private def timedFuture[T](label: String, f : => Future[T])(implicit ec: ExecutionContext): Future[T] = {
+    val timeBefore = DateTime.now().getMillis
+    val result = f
+      result.onComplete { _ =>
+        val timeAfter = DateTime.now().getMillis
+        logger.info(s"$label took ${timeAfter - timeBefore} milliseconds")
+      }
+    result
+  }
```
akash1810, line 26: This looks like a [stopwatch](https://github.com/guardian/grid/blob/main/common-lib/src/main/scala/com/gu/mediaservice/lib/logging/Stopwatch.scala), can we reuse it?

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
```diff
@@ -125,7 +128,42 @@ class DynamoDB[T](config: CommonConfig, tableName: String, lastModifiedKey: Opti
 
   def jsonGet(id: String, key: String)
              (implicit ex: ExecutionContext): Future[JsValue] =
-      get(id, key).map(item => asJsObject(item))
+    get(id, key).map(item => asJsObject(item))
+
+  def batchGet(ids: List[String])
```
akash1810, line 133: Should we move to [Scanamo](https://www.scanamo.org/batch-operations.html)?

metadata-editor/app/lib/Syndication.scala
```diff
@@ -1,67 +1,82 @@
...
+
+  private[lib] def publishChangedSyndicationRightsMessages[T](id: String, unchangedPhotoshoot: Boolean = false, photoshoot: Option[Photoshoot] = None)(f: () => Future[T])
```
akash1810, line 28: minor: Could the following renames make this easier to parse?:
  - `f` as `updatePhotoshoot` (??)
  - `before` as `current`
  - `after` as `updated`
 
That is, it would become more of a narrative:
  1. Get current state
  2. Perform update
  3. Get updated state

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
AWare: I'm not convinced that's a great use of Justin's time.

PR #3283, opened 2021-04-30
kcl-akka-stream bintray exfiltration

AWare:
# what

bintray has been shutdown, and to upgrade this library to a version would take more time than is remaining. 

we should upgrade to v5.0.0 as soon as possible- but this also requires upgrading the kcl to v2

## What does this change?

who i trust
enjoyment of public holidays

## Tested?
no
ok, it's in TEST and appears okay


tjsilver (approved):
This seems like the best option in the short term 👍

PR #3287, opened 2021-05-05
Update CollectionsManager.scala so label colours are accessible

pradasa:
## What does this change?
Updated the collectionColours to match Source and make Australia and Travel accessible for all users.

## Screenshots
Screenshot 2021-05-04 at 12 48 44

## Tested?
- [x] checked on figma, not tested


rhystmills (approved):
Looks good to me!

jonathonherbert (approved):
Fancy new colours, LGTM!

PR #3300, opened 2021-05-12
free now means under quota

AWare:
## What does this change?
spike to let us explore making free mean under quota. 
also, any UI should probably look at combining the CostFilter with quotas lest we make too much extra complexity here. 

also might be worth adding payType to the UI for chips
this took 3 minutes to implement

## Tested?
- [ ] locally by committer

### DIFF
--- a/media-api/app/lib/elasticsearch/ElasticSearch.scala
+++ b/media-api/app/lib/elasticsearch/ElasticSearch.scala
@@ -95,7 +95,7 @@ class ElasticSearch(val config: MediaApiConfig, mediaApiMetrics: MediaApiMetrics
     val hasIdentifier = params.hasIdentifier.map(idName => filters.exists(NonEmptyList(identifierField(idName))))
     val missingIdentifier = params.missingIdentifier.map(idName => filters.missing(NonEmptyList(identifierField(idName))))
     val uploadedByFilter = params.uploadedBy.map(uploadedBy => filters.terms("uploadedBy", NonEmptyList(uploadedBy)))
-    val simpleCostFilter = params.free.flatMap(free => if (free) searchFilters.freeFilter else searchFilters.nonFreeFilter)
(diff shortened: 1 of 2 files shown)


paperboyo:
As discussed, my main concern is this: even if users will be beyond happy not seeing all the over-quota imagery that they have to request a Lease to be able to use, in the instance of high over-quota flood (two big agencies at the same time), this will mean this (default?) view will reduce the number of images visible to most users (who never touch any checkboxes and just use free-text search) dramatically hence reducing the quality and diversity of imagery on the site (while lowering the number of requests for Leases from Picture Desk, so it will be a silent regression: everyone will be more happy and site will look worse.

> look at combining the CostFilter with quotas

If that means hiding Restricted-but-not-leased while showing Restricted-but-leased under Free to use: I agree, this would be sensible. Downside to consider: you upload a Handout or PR (they are Restricted by default), you don’t know you should add a Lease, it’s visible to you because *you as an uploader* can crop it, but it won’t be visible to anyone whom you’ve uploaded it for. _Where is this image you said you uploaded? It’s there. I can’t see it. I can. What do you mean? Come over to my house and see for youself. You come over to mine!_

> worth adding payType to the UI for chips

Was wondering about it… Currently, checking My uploads filter *does not* add an `+uploader:my.self@domain.com`, though. And its connection with Angular router, or whatever, is a tad broken, too (sometimes, clicking Home doesn’t clear the filter etc).

In general, the quickest/cheapest way (in my naive view!) to show Free to use for more users with narrower screens/in an iframe is to fix the searchbar width at first-from-widest breakpoint.

Once the above is done, maybe we can squeeze another filter (Hide over-quota) in there if we reword, tweak some CSS *slightly*.

I would be interested in how hard would it be to be able to hide this new Hide over-quota filter when nothing is over quota. Otherwise, we are adding something that is useless and does nothing most of the time.

Gee, sorry for the length.

PR #3301, opened 2021-05-13
someBODY once told me to fix the all star cleaner

AWare:
## What does this change?

It has been noticed that Allstar supplierProcessor is deduplicating byline whne it equals to credit for all images. This change applies the deduping only to images with Allstar [CanonicalName](https://github.com/guardian/grid/blob/6f0eb2228253b546f6c365da185b47f74511acc6/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala#L80).

A supplier processor is deleting byline when it is equal to credit.
May-13-2021 17-23-24

When the tests pass!

## Tested?
- [ ] locally by committer

(ps its the allstar processor)

### DIFF
--- a/common-lib/src/test/scala/com/gu/mediaservice/lib/cleanup/AttributeCreditFromBylineTest.scala
+++ b/common-lib/src/test/scala/com/gu/mediaservice/lib/cleanup/AttributeCreditFromBylineTest.scala
@@ -12,7 +12,7 @@ class AttributeCreditFromBylineTest extends FunSpec with Matchers with MetadataH
     testCleaner.clean(metadata).credit should be (Some("Some Credit"))
   }
 
-  it("should not set the credit if the byline doesn't matche the configured list") {
+  it("should not set the credit if the byline doesn't match the configured list") {
     val metadata = createImageMetadata("byline" -> "Someone else")
     testCleaner.clean(metadata).credit should be (None)
   }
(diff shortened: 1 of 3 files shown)


common-lib/src/test/scala/com/gu/mediaservice/lib/cleanup/AttributeCreditFromBylineTest.scala
```diff
@@ -12,9 +12,14 @@ class AttributeCreditFromBylineTest extends FunSpec with Matchers with MetadataH
     testCleaner.clean(metadata).credit should be (Some("Some Credit"))
   }
 
-  it("should not set the credit if the byline doesn't matches the configured list") {
+  it("should not set the credit if the byline doesn't matche the configured list") {
```
jonathonherbert, line 15: ye olde typographical mistaek

jonathonherbert (approved):
Makes sense, tests look good. My assumption is that if there's not a mention of the canonical name then nothing in that pipeline makes sense to apply – if that's correct, then 👍

common-lib/src/test/scala/com/gu/mediaservice/lib/cleanup/AttributeCreditFromBylineTest.scala
AWare:
```suggestion
  it("should not set the credit if the byline doesn't match the configured list") {
```

AWare:
Yeah, that felt like the most appropriate change to make, but I'm still not 100% on exactly what those cleaners do.

PR #3303, opened 2021-05-14
Unknown text for user without Edit Metadata permission

adrielulanovsky:
## What does this change?
Currently, when a user doesn't have Edit Metadata permission and they open an image without peopleInImage field, they can see the text "Unknown (click ✎ to add)", even when they don't have the ✎ button because of their permissions. This hides the People field for those users (only when the field is empty).

A user without Edit Metadata permission cannot see the text "Unknown (click ✎ to add)" in the Image Metadata section, and instead doesn't see the field.

## Screenshots
Current behaviour:
image

Behaviour after the fix:
image

## Who should look at this?
@guardian/digital-cms

## Tested?
- [ ] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
@@ -235,10 +235,10 @@
                 </span>
             </dd>
 
-            <dt class="image-info__group--dl__key metadata-line__key image-info__wrap">
+            <dt ng-if="ctrl.metadata.peopleInImage || ctrl.userCanEdit" class="image-info__group--dl__key metadata-line__key image-info__wrap">
                 People
             </dt>
-            <dd class="image-info__wrap image-info__group--dl__value metadata-line__info">
+            <dd ng-if="ctrl.metadata.peopleInImage || ctrl.userCanEdit" class="image-info__wrap image-info__group--dl__value metadata-line__info">
                 <button data-cy="it-edit-people-button"
                         class="image-info__edit"
                         ng-if="ctrl.userCanEdit"


paperboyo:
This is useful as it removes an obvious idiocy (they **cannot** click to add!). We may decide that this is a small step forward and should be merged as is (will still need a code review from the team).

But this reminded me of a little 🐰-hole I found last time I looked in this direction. See [point 6](https://github.com/guardian/grid/pull/3083) in this PR, which leads to [this comment](https://github.com/guardian/grid/pull/3097#issuecomment-760556259) describing UI inconsistency that was overlooked when People field was introduced (but, true, was pre-existing too). #3083 is stalled, because I definitely hit a limit of what I can do without being able to code and nobody took had time to baby-sit me ;-)

paperboyo:
In short, @adrielulanovsky, because I tend to write useless unactionable essays, I think it would be preferable to just hide the whole field completely for users who have no power to add data, just as it already happens for eg. a Byline (and other) field:

Permissioned user:
image

Unpermissioned user (no Byline field present):

image

adrielulanovsky:
> In short, @adrielulanovsky, because I tend to write useless unactionable essays, I think it would be preferable to just hide the whole field completely, if it’s empty, for users who have no power to add data, just as it already happens for eg. a Byline (and other) field:
> 
> Permissioned user:
> image
> 
> Unpermissioned user (no Byline field present in the UI):
> 
> image

Sure, I'll do that instead. Thanks for the context, I wasn't aware of it, I just saw the "click ✎ to add" text when not having permissions and thought it was an easy bug to fix :sweat_smile:

adrielulanovsky:
@paperboyo Updated PR with new approach

AWare (approved):

PR #3308, opened 2021-05-17
Fix BBC test dependencies

adrielulanovsky:
## What does this change?
Currently there is a compilation failure on BBC's tests caused by an inability to access common-lib's traits in the test folder. This change allows BBC (and other external projects depending on rest-lib) to access common-lib's and rest-lib's test folders.

See https://stackoverflow.com/questions/8193904/sbt-test-dependencies-in-multiprojects-make-the-test-code-available-to-dependen and https://www.scala-sbt.org/release/docs/Multi-Project.html#Per-configuration+classpath+dependencies for more info on the problem and solution

sbt test:compile runs successfully on BBC and Guardian

## Who should look at this?
@guardian/digital-cms

## Tested?
- [ ] locally by committer

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -72,7 +72,7 @@ val okHttpVersion = "3.12.1"
 val bbcBuildProcess: Boolean = System.getenv().asScala.get("BUILD_ORG").contains("bbc")
 
 //BBC specific project, it only gets compiled when bbcBuildProcess is true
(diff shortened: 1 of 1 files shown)


AWare (approved):

AWare:
pushed this to teamcity, will merge when it builds

PR #3316, opened 2021-05-20
Moved Xenial AMI to Bionic as it is no longer supported

rhystmills:
## What does this change?

Moves AMI from Xenial to Bionic because Xenial is no longer supported.

**Update:** `grid-imgops` and `grid-elasticsearch` recipes have now also been replaced with newly baked recipes.

Related card [here](https://trello.com/c/NLpikw2C/2330-update-xenial-amis-to-use-bionic).

## Has it been successfully deployed?

Yes, [here](**url**). All seems to be working well.
## How to test

Deploy to [TEST](https://media.test.dev-gutools.co.uk/search) via RiffRaff (it's under `media-service::grid::all`), and ensure it deploys and runs correctly.

### DIFF
--- a/riff-raff.yaml
+++ b/riff-raff.yaml
@@ -95,7 +95,7 @@ deployments:
         ImgOpsAmiId:
           BuiltBy: amigo
           AmigoStage: PROD
-          Recipe: grid-imgops
+          Recipe: grid-imgops-bionic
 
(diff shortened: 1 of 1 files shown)


AWare (approved):

tjsilver (approved):
👍

PR #3317, opened 2021-05-20
Feature: Hide/Soft-Delete Agency images

abdelrahmansd:
## What does this change?
This implements a new feature: Hide/Soft Delete images
add`deletedAt` field on Edits model
add`isDeleted` search filter
introduce`viewDeletedImages` permission that control which users can view or hide/soft-delete images on the Grid
introduce endpoint to hide/soft-delete images

- Given images in the Images Catalogue with a status of hidden
When the user has set the search filter `isDeleted = true` or through the `hidden` quick filter button in the UI
Then the user should be able to see all the hidden images in the Images Catalogue

- Given images in the Images Catalogue with a status of hidden
And a user who isn't in the right group 
Then user will not be able to set the search filter `isDeleted = true` and the`hidden`quick filter button in the UI will not be rendered

- Given `PUT     /metadata/:id/hide` endpoint
Then `deletedAt` field should be added in `userMatadata`
`                    "userMetadata": {
                        "archived": false,
                        "deletedAt": "2021-06-01T16:33:08.063+02:00",
                        "metadata": {
                            "description": "JVWF-superbeetle-5615-1200x800"
                        },
                        "lastModified": "2021-06-01T16:33:08.077+02:00",
                        "usageRights": {
                            "category": "public-domain"
                        },
                        "labels": []
                    }`

- Given `PUT     /metadata/:id/hide` endpoint
And a user who isn't in the right group 
Then it should respond with `permission-denied`

#### user with`viewDeletedImages`permission 
Screenshot from 2021-06-01 15-46-56
Screenshot from 2021-06-01 15-47-10

#### user without`viewDeletedImages`permission 
Screenshot from 2021-06-01 15-47-49

## Tested?
- [x] locally by committer


AWare:
I don't think we should add the checkbox there, we've managed to go a few weeks without a combined BBC/gu meeting. @pradasa is looking at this checkbox area for us to do something similar with

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
```diff
@@ -61,6 +61,14 @@ class DynamoDB[T](config: CommonConfig, tableName: String, lastModifiedKey: Opti
     }
   }
 
+  def keySet(id: String, key: String, value: String)
```
AWare, line 64: how does this differ from stringSet, and should this field be stored as a string?

abdelrahmansd: right `stringSet` does the same, updating the PR now

> I don't think we should add the checkbox there, we've managed to go a few weeks without a combined BBC/gu meeting. @pradasa is looking at this checkbox area for us to do something similar with

okay I will remove it for now

kahuna/public/js/search/results.js
```diff
@@ -316,11 +316,11 @@ results.controller('SearchResultsCtrl', [
             if (angular.isUndefined(orderBy)) {
                 orderBy = $stateParams.orderBy;
             }
-
             return mediaApi.search($stateParams.query, angular.extend({
                 ids:        $stateParams.ids,
                 archived:   $stateParams.archived,
                 free:       $stateParams.nonFree === 'true' ? undefined : true,
+                isDeleted:  $stateParams.isDeleted === 'true' ? true : false,
```
akash1810, line 323: I don't think we need the ternary here? `$stateParams.isDeleted === 'true'` should be enough.

paperboyo:
> could we avoid using the word `hidden` and favour `deleted` or similar
&
> Paid images are "hidden" by default too, so being explicit in the naming here stops us overloading the term "hidden".

OTOH, `delete` has already a concrete meaning, no? And not sure paid images are called `hidden` explicitly anywhere (incl. in code)? I would then favour `hidden`. Or a third one altogether?

media-api/app/controllers/MediaApi.scala
```diff
@@ -70,9 +70,12 @@ class MediaApi(
 
     val userCanUpload: Boolean = authorisation.hasPermissionTo(UploadImages)(user)
     val userCanArchive: Boolean = authorisation.hasPermissionTo(ArchiveImages)(user)
+    val userCanViewDeletedImages: Boolean = authorisation.hasPermissionTo(ViewDeletedImages)(user)
+
 
     val maybeLoaderLink: Option[Link] = Some(Link("loader", config.loaderUri)).filter(_ => userCanUpload)
     val maybeArchiveLink: Option[Link] = Some(Link("archive", s"${config.metadataUri}/metadata/{id}/archived")).filter(_ => userCanArchive)
+    val maybeViewDeletedLink: Option[Link] = Some(Link("archive", s"${config.authUri}/{id}/hide")).filter(_ => userCanViewDeletedImages)
```
akash1810, line 78: Should this be `"archive"` here? I think it'd want to match the new route added below?

akash1810:
> OTOH, `delete` has already a concrete meaning, no?

My understanding is "soft delete" would be the only form of deletion made available in Grid? That is our current delete is getting replaced with "soft delete"?

paperboyo:
> My understanding is "soft delete" would be the only form of deletion made available in Grid? That is our current delete is getting replaced with "soft delete"?

Oh! If that’s the case, ie. Delete button will actually soft-delete, then I take all I said back! If that’s the case, we could even relax some of our Permissions/rules around Deletion: eg. currently, it is **impossible** to delete an image which has crops. But if it would only be a soft-delete, they I can’t see a reason why not?

abdelrahmansd:
> > OTOH, `delete` has already a concrete meaning, no?
> 
> My understanding is "soft delete" would be the only form of deletion made available in Grid? That is our current delete is getting replaced with "soft delete"?

twrichards:
Closing this PR as it has been broken down into multiple PRs which follow a somewhat different approach, so far there's...
- #3353
- #3356
- #3360
- (_please edit this comment to add more PRs when they've been created)_

PR #3318, opened 2021-05-20
Let's think about reingestion / multi indexing

AWare:
## 🖋 Whither index/alias

So, to reingest an image, we'll need some kind of multi-index support. Just enough to be able to manually write to a second index. 

This really needs a sensible name. 

## 👻 Ghost of reindex past

In this work, we've been a little lax about whether to call it _reindex_ or _reinest_, and we were similarly lax when we did the past (different) reindexing work. We'll need to either rip out that logic, or keep both sets but be strict.

## 🧵 Thrall

I don't think that the current way that thrall passes messages is great for this work. Discussing with @sihil, we thought about moving from a giant optional fieldy json with a string tag to some kind of `trait`/`case class` combo- where each message type has its own object. 

Right now, I'm just using the `image` field on the multi-purpose object to start the reingest for now.

## 🤷‍♂️ What I'm looking for?

I'd like to get some opinions on this all so I can start carding out parts. I'm thinking:

1) Typesafe messages in thrall
	a) Delete past reindexer from thrall
1) Spare alias configured and available for multi-index running.
1) Reindex an image into the spare.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchConfig.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchConfig.scala
@@ -1,3 +1,3 @@
 package com.gu.mediaservice.lib.elasticsearch
 
-case class ElasticSearchConfig(alias: String, url: String, cluster: String, shards: Int, replicas: Int)
+case class ElasticSearchConfig(alias: String, accretorAlias: String, url: String, cluster: String, shards: Int, replicas: Int)
--- a/thrall/app/lib/elasticsearch/ElasticSearch.scala
+++ b/thrall/app/lib/elasticsearch/ElasticSearch.scala
@@ -26,6 +26,7 @@ class ElasticSearch(config: ElasticSearchConfig, metrics: Option[ThrallMetrics])
   with ImageFields with ElasticSearchExecutions {
 
   lazy val imagesAlias: String = config.alias
+  lazy val accretorAlias: String = config.accretorAlias
   lazy val url: String = config.url
   lazy val cluster: String = config.cluster
   lazy val shards: Int = config.shards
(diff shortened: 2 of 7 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchConfig.scala
```diff
@@ -1,3 +1,3 @@
 package com.gu.mediaservice.lib.elasticsearch
 
-case class ElasticSearchConfig(alias: String, url: String, cluster: String, shards: Int, replicas: Int)
+case class ElasticSearchConfig(alias: String, accretorAlias: String, url: String, cluster: String, shards: Int, replicas: Int)
```
twrichards, line 3: had to look up what https://en.wiktionary.org/wiki/accretor meant - and even then not sure exactly what it means in this context - perhaps we could pick a more mainstream word to describe this alias string?

thrall/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -53,8 +54,11 @@ class ElasticSearch(config: ElasticSearchConfig, metrics: Option[ThrallMetrics])
     List(response.map(_ => ElasticSearchBulkUpdateResponse()))
   }
 
-  def indexImage(id: String, image: Image, lastModified: DateTime)
+  def indexImage(id: String, image: Image, lastModified: DateTime, reindex: Boolean = false)
                 (implicit ex: ExecutionContext, logMarker: LogMarker): List[Future[ElasticSearchUpdateResponse]] = {
+
+    val index = if(reindex){accretorAlias}else{imagesAlias}
```
twrichards, line 60: appreciate the word reindex is becoming overloaded (as per PR description) but would calling it `reIndexAlias` be better?

twrichards: or `reIngestAlias` (given other reindex -> reingest name changes in this PR)

twrichards (commented):
Some comments on naming - but otherwise changes LGTM so far 👍

thrall/app/lib/elasticsearch/ElasticSearch.scala
AWare: We've gone for `migration` as the word.

This PR is mostly for design discussion now as the change I was attempting had too many preconditions.
https://github.com/guardian/grid/pull/3324

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchConfig.scala
AWare: I nearly also used the word [roche](https://en.wikipedia.org/wiki/Interacting_binary_star) as well, which may have given away that I didn't intend to actually go through with this confusing terminology.

PR #3319, opened 2021-05-20
Split thrall messages out into types internally 

AWare:
## What does this change?

So, following from https://github.com/guardian/grid/pull/3318 I think it's not unfair to say the `UpdateMessage` class isn't great- and does not extend to provide reingestion through thrall particularly easily. 

So, this pulls it out into a sealed trait with a whole bunch of nice case classes.

Right now, where we're at, is that I've implemented these for all the messages (except those formerly known as reindex) and I've determined their fields based on their call sites. 

These are then passed into the message processors.

As the new reingest should be entirely within thrall, this should be sufficient to unblock it. 

## Next PR
Afaict: Justin's syndication work has freed us from non-grid thrall producers. But for purposes of rolling out, and rewinding we can't just use these.

The next three PRS would be:
1) Decode these types from JSON
2) Send these types
3) Remove `UpdateMessage` processor

Should not merge decoder until we're happy with the sent messages.
Sender can go a week after both us/bbc have merged the decoder. 
Removal of `UpdateMessage` should not happen until thralls kinesis history is only the new type, which should take a further week.

## Tested?
- [ ] locally by committer


common-lib/src/main/scala/com/gu/mediaservice/model/ThrallMessage.scala
```diff
@@ -0,0 +1,89 @@
+package com.gu.mediaservice.model
+
+ import com.gu.mediaservice.lib.logging.{GridLogging, LogMarker}
+ import com.gu.mediaservice.model.leases.MediaLease
+ import com.gu.mediaservice.model.usage.UsageNotice
+ import org.joda.time.{DateTime, DateTimeZone}
```
tjsilver, line 6: Do we still need to use joda time even though it is deprecated?

AWare: I'm not sure, I just didn't want to alter the types from the existing message. (Aside from removing `Option`)

twrichards: probably best to move to java.time in a sep. PR

common-lib/src/main/scala/com/gu/mediaservice/model/ThrallMessage.scala
```diff
@@ -0,0 +1,88 @@
...
+ import play.api.libs.json.{JodaReads, JodaWrites, Json, __}
+
+sealed trait ThrallMessage extends GridLogging with LogMarker {
+  implicit val yourJodaDateReads = JodaReads.DefaultJodaDateTimeReads.map(d => d.withZone(DateTimeZone.UTC))
+  implicit val yourJodaDateWrites = JodaWrites.JodaDateTimeWrites
+  def subject: () => String = () => this.getClass.getSimpleName
```
twrichards, line 12: [minor] could this be a `val` or will that return `ThrallMessage` rather than the name of case class implementation

common-lib/src/main/scala/com/gu/mediaservice/model/ThrallMessage.scala
```diff
@@ -0,0 +1,88 @@
...
+  val id: String
+  val lastModified: DateTime
+  def additionalMarkers: () => Map[String, Any] = () => Map()
```
twrichards, line 15: could this be a `lazy val`

common-lib/src/main/scala/com/gu/mediaservice/model/ThrallMessage.scala
```diff
@@ -0,0 +1,88 @@
...
+      "subject" -> subject,
+      "id" -> id,
+      "size" -> message.getBytes.length,
+      "length" -> message.length
+    )
+  }
+}
+object ThrallMessage{
+  implicit val yourJodaDateReads = JodaReads.DefaultJodaDateTimeReads.map(d => d.withZone(DateTimeZone.UTC))
+  implicit val yourJodaDateWrites = JodaWrites.JodaDateTimeWrites
```
twrichards, line 29: are these needed in that they're in the sealed trait, might be worth seeing if you can delete either these lines or the two lines in the trait

common-lib/src/main/scala/com/gu/mediaservice/model/ThrallMessage.scala
```diff
@@ -0,0 +1,88 @@
...
+case class RemoveImageLeaseMessage(id: String, lastModified: DateTime, leaseId: String) extends ThrallMessage
+
+case class SetImageCollectionsMessage(id: String, lastModified: DateTime, collections: Seq[Collection]) extends ThrallMessage
+
+case class DeleteUsagesMessage(id: String, lastModified: DateTime) extends ThrallMessage
+
+object DeleteUsagesMessage {
+  implicit val yourJodaDateReads = JodaReads.DefaultJodaDateTimeReads.map(d => d.withZone(DateTimeZone.UTC))
+  implicit val yourJodaDateWrites = JodaWrites.JodaDateTimeWrites
+  implicit val what = Json.format[DeleteUsagesMessage]
```
twrichards, line 80: why does this need its own Format and even if it does how come it needs the joda read/writes?

twrichards (commented):
LGTM so far 👍 
(left a few minor comments)

common-lib/src/main/scala/com/gu/mediaservice/model/ThrallMessage.scala
```diff
@@ -0,0 +1,88 @@
...
+}
```
sihil, line 81: There is another formats impl in the `ThrallMessage` object. Are both needed?

common-lib/src/main/scala/com/gu/mediaservice/model/ThrallMessage.scala
```diff
@@ -0,0 +1,88 @@
...
+
+ import com.gu.mediaservice.lib.logging.{GridLogging, LogMarker}
+ import com.gu.mediaservice.model.leases.MediaLease
+ import com.gu.mediaservice.model.usage.UsageNotice
+ import org.joda.time.{DateTime, DateTimeZone}
+ import play.api.libs.json.{JodaReads, JodaWrites, Json, __}
+
+sealed trait ThrallMessage extends GridLogging with LogMarker {
+  implicit val yourJodaDateReads = JodaReads.DefaultJodaDateTimeReads.map(d => d.withZone(DateTimeZone.UTC))
+  implicit val yourJodaDateWrites = JodaWrites.JodaDateTimeWrites
```
sihil, line 11: Don't think that these are used? Compiles without them.

common-lib/src/main/scala/com/gu/mediaservice/model/ThrallMessage.scala
sihil, line 12: I don't think this needs to return a function?

common-lib/src/main/scala/com/gu/mediaservice/model/ThrallMessage.scala
```diff
@@ -0,0 +1,88 @@
...
+  def subject: () => String = () => this.getClass.getSimpleName
+  val id: String
+  val lastModified: DateTime
+  def additionalMarkers: () => Map[String, Any] = () => Map()
+
+  override def markerContents = {
+    val message = Json.stringify(Json.toJson(this)(ThrallMessage.writes))
+    Map (
+      "subject" -> subject,
```
sihil, line 20: you're passing in a function here, I suspect this should be `subject()`? (or just make this a string as mentioned above)

thrall/app/lib/kinesis/MessageTranslator.scala
```diff
@@ -0,0 +1,71 @@
...
+      case UpdateImageSyndicationMetadata => (updateMessage.id, updateMessage.syndicationRights) match {
+        case (Some(id), maybeSyndicationMetadata) => Right(UpdateImageSyndicationMetadataMessage(id, updateMessage.lastModified, maybeSyndicationMetadata))
+        case _ => Left(MissingFieldsException(updateMessage.subject))
+      }
+      case UpdateImagePhotoshootMetadata => (updateMessage.id, updateMessage.edits) match {
+        case (Some(id), Some(edits)) => Right(UpdateImagePhotoshootMetadataMessage(id, updateMessage.lastModified, edits))
+        case _ => Left(MissingFieldsException(updateMessage.subject))
+      }
+      case _ => Left(ProcessorNotFoundException(updateMessage.subject))
+    }
```
sihil, line 65: Personally I have a pathological dislike of nested pattern matching and prefer something like this (granted this then has a lot of underscores, so I understand the motivation):
```suggestion
    updateMessage match {
      case UpdateMessage(Image | ReingestImage | UpdateImage, Some(image), Some(id), _, _, lastModified, _, _, _, _, _, _, _) =>
        Right(ImageMessage(id, lastModified, image))
      case UpdateMessage(DeleteImage, _, Some(id), _, _, lastModified, _, _, _, _, _, _, _) =>
        Right(DeleteImageMessage(id, lastModified))
      case UpdateMessage(DeleteImageExports, _, Some(id), _, _, lastModified, _, _, _, _, _, _, _) =>
        Right(DeleteImageExportsMessage(id, lastModified))
      case UpdateMessage(UpdateImageExports, _, Some(id), _, _, lastModified, _, _, Some(crops), _, _, _, _) =>
        Right(UpdateImageExportsMessage(id, lastModified, crops))
      case UpdateMessage(UpdateImageUserMetadata, _, Some(id), _, Some(edits), lastModified, _, _, _, _, _, _, _) =>
        Right(UpdateImageUserMetadataMessage(id, lastModified, edits))
      case UpdateMessage(UpdateImageUsages, _, Some(id), Some(usageNotice), _, lastModified, _, _, _, _, _, _, _) =>
        Right(UpdateImageUsagesMessage(id, lastModified, usageNotice))
      case UpdateMessage(UpdateImageUsages, _, Some(id), _, _, lastModified, _, _, _, _, Some(leases), _, _) =>
        Right(ReplaceImageLeasesMessage(id, lastModified, leases))
      case UpdateMessage(AddImageLease, _, Some(id), _, _, lastModified, _, _, _, Some(mediaLease), _, _, _) =>
        Right(AddImageLeaseMessage(id, lastModified, mediaLease))
      case UpdateMessage(RemoveImageLease, _, Some(id), _, _, lastModified, _, Some(leaseId), _, _, _, _, _) =>
        Right(RemoveImageLeaseMessage(id, lastModified, leaseId))
      case UpdateMessage(SetImageCollections, _, Some(id), _, _, lastModified, Some(collections), _, _, _, _, _, _) =>
        Right(SetImageCollectionsMessage(id, lastModified, collections))
      case UpdateMessage(DeleteUsages, _, Some(id), _, _, lastModified, _, _, _, _, _, _, _) =>
        Right(DeleteUsagesMessage(id, lastModified))
      case UpdateMessage(UpdateImageSyndicationMetadata, _, Some(id), _, _, lastModified, _, _, _, _, _, maybeSyndicationMetadata, _) =>
        Right(UpdateImageSyndicationMetadataMessage(id, lastModified, maybeSyndicationMetadata))
      case UpdateMessage(UpdateImagePhotoshootMetadata, _, Some(id), _, Some(edits), lastModified, _, _, _, _, _, _, _) =>
        Right(UpdateImagePhotoshootMetadataMessage(id, updateMessage.lastModified, edits))
      case UpdateMessage(subject, _, _, _, _, _, _, _, _, _, _, _, _) if validSubjects.contains(subject) =>
        Left(MissingFieldsException(updateMessage.subject))
      case UpdateMessage(subject, _, _, _, _, _, _, _, _, _, _, _, _) =>
        Left(ProcessorNotFoundException(updateMessage.subject))
    }
```

sihil: You'll need this in the MessageSubjects trait if you take the above (shame we don't have enumeratum in this project yet...):
```
  val validSubjects = Set(
    Image,
    ReingestImage,
    DeleteImage,
    UpdateImage,
    DeleteImageExports,
    UpdateImageExports,
    UpdateImageUserMetadata,
    UpdateImageUsages,
    ReplaceImageLeases,
    AddImageLease,
    RemoveImageLease,
    SetImageCollections,
    DeleteUsages,
    UpdateImageSyndicationMetadata,
    UpdateImagePhotoshootMetadata,
  )
```

sihil (commented):
This is good. Sensible to make it internal to Thrall to start with.

Given the significant change and minimal test coverage on this part of Thrall I'm somewhat concerned that if we have missed a detail somewhere such that we have modified the way we process a particular message type we may not notice for quite some time. I'm not sure what we can do about that, but it's a risk IMHO.

common-lib/src/main/scala/com/gu/mediaservice/model/ThrallMessage.scala
sihil: that can safely be a val of `String` and will do what we want.

thrall/app/lib/kinesis/MessageTranslator.scala
AWare: Your pathological dislike of nesting pattern matchers meets mine of _'s. I tried to pull this logic out into custom unapplies, but couldn't pull it off.

common-lib/src/main/scala/com/gu/mediaservice/model/ThrallMessage.scala
AWare: yeah, hadn't realised trait values or those in the body of a case class won't get serialised

common-lib/src/main/scala/com/gu/mediaservice/model/ThrallMessage.scala
AWare: it's a bit of a mess, trying to cleanly source these implicits and this wasn't needed. have added a method to the trait that json's it, which allows the convertors to live in the traits companion object rather than spamming the file

AWare:
Comments addressed and it's on TEST.

twrichards (approved):
LGTM 👍  (although haven't tested as PR is lacking testing instructions)
I've left a couple of fairly minor suggestions, bu

PR #3320, opened 2021-05-21
Add classes to dropdown menu options

adrielulanovsky:
## What does this change?
Add classes to dropdown menu options. The BBC would like to hide the Quotas button for the moment, and it will do so by targeting the specific class for that button and hiding it with specific BBC css.

The Guardian can see the Quotas button, while BBC cannot.

## Screenshots
image

## Who should look at this?
@guardian/digital-cms

## Tested?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/common/user-actions.html
+++ b/kahuna/public/js/common/user-actions.html
@@ -5,8 +5,8 @@
         <gr-icon>more_vert</gr-icon>
     </button>
     <ul class="drop-menu__items drop-menu__items--right" ng-if="showUserActions">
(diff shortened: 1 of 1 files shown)


AWare:
Do we want this as config, or do we want some kind of per-org css/js?

adrielulanovsky:
> Do we want this as config, or do we want some kind of per-org css/js?

@AWare I'm not confident to say which way would be better, but I can try to take the css approach if you prefer.

AWare (approved):

PR #3324, opened 2021-05-24
rename reindex/reingest to 'migration' and introduce new message type for starting a migration

AWare:
## What does this change?

This sets the name of the reingestion to "migration" for want of a unique name.

This takes #3319 and adds an Internal/External distinction, so that when we send messages over kinesis in this form, we don't send Migration ones.

Internal messages refer to those originating from within thrall (which is a new thing). External refers to those messages coming into thrall from outside via the kinesis streams (which is all the existing types of messages handled by thrall).

## Context 

We intend to replace the current aliases `read` and `write` (see #3362) with:
- `Images_Current` _which will be the primary index_
- `Images_Migration` _which will be the index into which the migration is written_

Then, when this message is processed by thrall it will:
1. Check for the `Images_Migration` alias, and halt if present. 
2. Create a new index called `images_YYYY-MM-DDTHH:mm:ss_GITHASH`
3. Point the `Images_Migration` alias at this index.

We then, on all writes, will write the name of this index to a marker field on the elasticsearch record. 

Then on all reads, we'll [augment the response](https://github.com/guardian/grid/blob/6dbe1b6124511b4196d70b44893a8ac4850814a4/media-api/app/lib/ImageResponse.scala#L94-L107)
 with 
```ts
{
	...
	esInfo: {
		migratedTo: ['images'...],
		sourceAlias: 'Images_Migration',
		aliasMappings: {
			Images_Current: 'images....',
			Images_Migration: 'images2...'
		}
	}
}
```
Which will make a nice enough way to understand the migration status of an image. This will be especially usefully initially when we're testing multi-read and multi-write.

- inspection
- a series of unit tests which will shortly be written

## Screenshots
Here is the jamboard between @sihil and I, detailing this.
image

## Tested?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/InternalThrallMessage.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/InternalThrallMessage.scala
@@ -1,0 +1,17 @@
+package com.gu.mediaservice.model
+
+import org.joda.time.DateTime
+
+sealed trait InternalThrallMessage extends ThrallMessage {
+
+}
+
+/**
+  * Message to start a new 'migration' (for re-index, re-ingestion etc.)
+  * @param migrationStart timestamp representing when a migration commenced
+  * @param gitHash the git commit hash (of the grid repo) at the point the migration commenced
+  */
+case class CreateMigrationIndexMessage(
+  migrationStart: DateTime,
+  gitHash: String
+) extends InternalThrallMessage
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/ThrallMessage.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/ThrallMessage.scala
@@ -1,85 +1,16 @@
 package com.gu.mediaservice.model
 
- import com.gu.mediaservice.lib.logging.{GridLogging, LogMarker}
- import com.gu.mediaservice.model.leases.MediaLease
- import com.gu.mediaservice.model.usage.UsageNotice
- import org.joda.time.{DateTime, DateTimeZone}
- import play.api.libs.json
- import play.api.libs.json.{JodaReads, JodaWrites, Json, OFormat, OWrites, Reads, __}
+import com.gu.mediaservice.lib.logging.{GridLogging, LogMarker}
 
-sealed trait ThrallMessage extends GridLogging with LogMarker {
-  val subject: String = this.getClass.getSimpleName()
-  val id: String
-  val lastModified: DateTime
-  val additionalMarkers: Map[String, Any] = Map()
-  def toJson = Json.toJson(this)
+trait ThrallMessage extends GridLogging with LogMarker {
+  val subject: String = this.getClass.getSimpleName
+  def additionalMarkers: () => Map[String, Any] = () => Map()
 
-  override def markerContents = {
-    val message = Json.stringify(Json.toJson(this)(ThrallMessage.formats))
+  override def markerContents: Map[String, Any] = {
     Map (
-      "subject" -> subject,
-      "id" -> id,
-      "size" -> message.getBytes.length,
-      "length" -> message.length
+      "subject" -> subject
     )
   }
 }
-object ThrallMessage{
-  implicit val yourJodaDateReads: Reads[DateTime] = JodaReads.DefaultJodaDateTimeReads.map(d => d.withZone(DateTimeZone.UTC))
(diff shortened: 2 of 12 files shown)


common-lib/src/main/scala/com/gu/mediaservice/model/ThrallMessage.scala
```diff
@@ -1,88 +1,21 @@
...
-  implicit val what = Json.format[DeleteUsagesMessage]
-}
-
-case class UpdateImageSyndicationMetadataMessage(id: String, lastModified: DateTime, maybeSyndicationRights: Option[SyndicationRights]) extends ThrallMessage
-
-case class UpdateImagePhotoshootMetadataMessage(id: String, lastModified: DateTime, edits: Edits) extends ThrallMessage
-
-
-
+case class CreateMigrationIndexMessage(migrationStart: DateTime ,gitHash: String)
```
twrichards, line 21: as per PR description the new index will be called `images_YYYY-MM-DDTHH:mm:ss_GITHASH` (this case class looks clearly related).
I like that we're putting a traceable unique reference in the new index name, but since these migrations could last many days (even weeks) does this mean that we can't release other grid things while a migration is underway, since this would then create a further new index and essentially start again - apologies if I've misunderstood what this git hash refers to

AWare: the hash/date is just for convenience, so we can see when it was started.

twrichards: ahh I see, because it's checking for the consistently named `Images_Migration` alias - gotcha

common-lib/src/main/scala/com/gu/mediaservice/model/InternalThrallMessage.scala
```diff
@@ -0,0 +1,9 @@
+package com.gu.mediaservice.model
+
+import org.joda.time.DateTime
+
+sealed trait InternalThrallMessage extends ThrallMessage {
+
+}
+
+case class CreateMigrationIndexMessage(migrationStart: DateTime, gitHash: String) extends InternalThrallMessage
```
jonathonherbert, line 9: What is `gitHash`? Might be nice to have a comment to explain?

common-lib/src/main/scala/com/gu/mediaservice/model/ThrallMessage.scala
```diff
@@ -1,85 +1,16 @@
...
+import com.gu.mediaservice.lib.logging.{GridLogging, LogMarker}
 
-sealed trait ThrallMessage extends GridLogging with LogMarker {
-  val subject: String = this.getClass.getSimpleName()
-  val id: String
-  val lastModified: DateTime
-  val additionalMarkers: Map[String, Any] = Map()
-  def toJson = Json.toJson(this)
+trait ThrallMessage extends GridLogging with LogMarker {
+  def subject: () => String = () => this.getClass.getSimpleName
```
jonathonherbert, line 6: Probably more a test of my Scala knowledge – why the change from `val` to `def` here?

jonathonherbert (approved):
Code looks fine, one minor comment. Re: `Internal/External` – just to double check my understanding of the naming, internal or external to what? Can we be more specific?

common-lib/src/main/scala/com/gu/mediaservice/model/InternalThrallMessage.scala
twrichards: @tjsilver and I moved this code when we refactored/added to Alex's PR, which didn't bring over the GH comment thread about this see https://github.com/guardian/grid/pull/3324#discussion_r638597056

Nonetheless I'll add a code comment

common-lib/src/main/scala/com/gu/mediaservice/model/ThrallMessage.scala
twrichards: good catch, this was an artifact from Alex's changes, I've changed it back to val (and it doesn't need to be a function neither AFAIK) - will test it now

twrichards:
> Re: `Internal/External` – just to double check my understanding of the naming, internal or external to what? Is it possible to be more specific?

@jonathonherbert internal/external to thrall, internal messages refer to those originating from within thrall (which is a new thing), external refers to those messages coming into thrall from outside via the kinesis streams (which is all the existing types of messages handled by thrall)

PR #3325, opened 2021-05-25
Add configurable homeLinkHtml

adrielulanovsky:
## What does this change?
This allows to add a different html for the home link via configuration. 

This is needed for the BBC because it is required to have 2 home link buttons instead of one. 

If the configuration is empty, the default value in the js file is used.

The Guardian can see no difference in the UI with this change.
BBC sees 2 home buttons instead of one.

## Screenshots
BBC view:
BBC Images - Top Nav LIVE

## Who should look at this?
@guardian/digital-cms

## Tested?
- [ ] locally by committer

### DIFF
--- a/kahuna/app/controllers/KahunaController.scala
+++ b/kahuna/app/controllers/KahunaController.scala
@@ -39,7 +39,8 @@ class KahunaController(
       config.supportEmail,
       fieldAliases,
       config.scriptsToLoad.filter(_.permission.map(authorisation.hasPermissionTo).fold(true)(maybeUser.exists)),
-      config.staffPhotographerOrganisation
+      config.staffPhotographerOrganisation,
+      config.homeLinkHtml
     ))
   }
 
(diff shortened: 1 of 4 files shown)


paperboyo:
Not that I should care, but out of curiosity, what are these two separate links lead to? Also, how do they play with other pages:

image
&
image

Top toolbar in browser is quite busy (esp. if chips feel full width of the searchbox) and doesn’t exactly collapse gracefully, so I guess it’s something worth thinking about…

adrielulanovsky:
> Not that I should care, but out of curiosity, what are these two separate links lead to? Also, how do they play with other pages:
> 
> image
> &
> image
> 
> Top toolbar in browser is quite busy (esp. if chips feel full width of the searchbox) and doesn’t exactly collapse gracefully, so I guess it’s something worth thinking about…

image
I don't know how to make that Cancel button appear, but it should look the same I think.

Both links lead to the home page of the Grid. Doesn't make much sense to me, but this was specifically requested to be like that :man_shrugging:

paperboyo:
> Both links lead to the home page of the Grid

Yeah. Odd. Maybe first step in getting it a bit more right would be to get rid of the line between them same links, then ;-) Thx!

adrielulanovsky:
> > Both links lead to the home page of the Grid
> 
> Yeah. Odd. Maybe first step in getting it a bit more right would be to get rid of the line between them same links, then ;-) Thx!

The line is in the provided mocks, so I'm afraid I'll have to leave it there. However this is all done via custom css and html, so Guardian's UI should remain the same.

AWare (approved):

PR #3328, opened 2021-05-26
Manually parse fixture id into suppliers ref for reuters, getty, epa and ap images only

AWare:
## What does this change?

This adds iptc fixture id as suppliers reference to the Reuters cleaner. 

Would also suggest that we put the suppliers reference extraction in the individual supplier cleaners.

- When we finish reindex, we should be able to make the Suppliers Ref better searchable. 
- That @paperboyo is happy

## Tested?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
@@ -363,20 +363,32 @@ object PaParser extends ImageProcessor {
 }
 
 object ReutersParser extends ImageProcessor {
+  def extractFixtureID(image:Image) = image.fileMetadata.iptc.get("Fixture Identifier")
+
   def apply(image: Image): Image = image.metadata.credit match {
+
     // Reuters and other misspellings
     // TODO: use case-insensitive matching instead once credit is no longer indexed as case-sensitive
     case Some("REUTERS") | Some("Reuters") | Some("RETUERS") | Some("REUETRS") | Some("REUTERS/") | Some("via REUTERS") | Some("VIA REUTERS") | Some("via Reuters") => image.copy(
       usageRights = Agency("Reuters"),
-      metadata = image.metadata.copy(credit = Some("Reuters"))
+      metadata = image.metadata.copy(
+        credit = Some("Reuters"),
+        suppliersReference = extractFixtureID(image) orElse image.metadata.suppliersReference
+      )
     )
     // Others via Reuters
     case Some("USA TODAY Sports") => image.copy(
-      metadata = image.metadata.copy(credit = Some("USA Today Sports")),
+      metadata = image.metadata.copy(
+        credit = Some("USA Today Sports"),
+        suppliersReference = extractFixtureID(image) orElse image.metadata.suppliersReference
+      ),
       usageRights = Agency("Reuters")
     )
     case Some("USA Today Sports") | Some("TT NEWS AGENCY") => image.copy(
-      usageRights = Agency("Reuters")
+      usageRights = Agency("Reuters"),
+      metadata = image.metadata.copy(
+        suppliersReference = extractFixtureID(image) orElse image.metadata.suppliersReference
+      )
     )
     case _ => image
   }
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
@@ -312,7 +312,7 @@ object GettyXmpParser extends ImageProcessor with GettyProcessor {
 
     val isExcludedByCredit = image.metadata.credit.exists(isExcluded(_, excludedCredit))
     val isExcludedBySource = image.metadata.source.exists(isExcluded(_, excludedSource))
-    val hasGettyMetadata = image.fileMetadata.getty.nonEmpty && getSuppliersReference(image).nonEmpty
+    val hasGettyMetadata = image.fileMetadata.getty.nonEmpty
 
 
     if(!hasGettyMetadata){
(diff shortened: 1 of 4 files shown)


paperboyo:
👎  Grid already is very confused and conflates two standard schema [fields into one](https://github.com/guardian/grid/blob/33411dbc614c17cbb250eb540e3dd6e4eb4bf55a/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala#L74). This is wrong (what if both are populated?).

This PR adds yet another another standard schema field to that confusion.

Instead of doing this, we need to untangle what we are already doing wrong, instead of making it worse.

AWare:
I'd be inclined to agree with you if we were reading this field from something called `Suppliers Reference` in the metadata.

As it isn't, and `Fixture ID` is the reference used by the supplier, this change is sound and correct. 

It also provides a way out of this problem in the long term: 
Rather than adding extra logic to `ImageMetadataConvertor` for suppliers slipping their internal references into random metadata locations, we copy this model and add the logic to that suppliers custom cleaner.

paperboyo:
> I'd be inclined to agree with you if we were reading this field from something called `Suppliers Reference` in the metadata.

I think I was wrong and you are right, Alex. Apologies! I was wrong that `supplierReference` is a standard-schema field. It is not. It is our invention, and thus, we indeed have control over which standard or custom schema field to read and stick there. And indeed it may differ for different suppliers (:sadpanda:). And indeed `iptc.Fixture Identifier` is better [than both](https://github.com/guardian/grid/blob/33411dbc614c17cbb250eb540e3dd6e4eb4bf55a/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala#L75-L78) `iptc.Original Transmission Reference` and `iptc.Object Name` specifically for Reuters.

I humbly withdraw my 👎.

> Would also suggest that we put the suppliers reference extraction in the individual supplier cleaners.

This makes sense too. It would eg. be `xmp.GettyImagesGIFT:AssetID` for Getty (modulo [this craziness](https://github.com/guardian/grid/blob/33411dbc614c17cbb250eb540e3dd6e4eb4bf55a/image-loader/app/lib/imaging/FileMetadataReader.scala#L146)) and something else for others, it wouldn’t be consistent and would require them not to change their mind, but it would be better than current situation. And if those fields would be empty, we would fill them according to current, crazy, logic.

This will make the field more vulnerable to suppliers’ whims, though. And our Rights dept. may be using this field as is. So let’s discuss these two points.

paperboyo:
Survey of the remaining ones. We need to fix only three in **bold**.

AAP: `iptc.Original Transmission Reference` (already OK)
Action Images: hard to say, maybe `iptc.Unique Document Identifier` before what we already read?
Alamy: `xmp.dc:title` (`iptc.Original Transmission Reference`) (already OK)
Allstar: `xmp.photoshop:TransmissionReference` (`iptc.Original Transmission Reference`) (already OK)
**AP**: `xmp.plus:ImageSupplierImageID` (and [structured](https://www.iptc.org/std/photometadata/specification/IPTC-PhotoMetadata#registry-entry-structure) `xmp.Iptc4xmpExt:RegistryId` key/values)
**EPA**: `iptc.Unique Document Identifier`
**Getty**: `xmp.GettyImagesGIFT:AssetID` (old ones [spelled differently](https://github.com/guardian/grid/blob/33411dbc614c17cbb250eb540e3dd6e4eb4bf55a/image-loader/app/lib/imaging/FileMetadataReader.scala#L146))
PA: `iptc.Original Transmission Reference` (already OK)
Rex: `iptc.Original Transmission Reference` (already OK)
Ronald Grant: `xmp.dc:title` (already OK)

AWare:
Added custom cleaners for the ones in bold.

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -275,6 +287,11 @@ trait GettyProcessor {
 }
 
 object GettyXmpParser extends ImageProcessor with GettyProcessor {
+  def getSuppliersReference(image: Image)={
+    // String defined in ImageMetadataConvertor
+    image.fileMetadata.getty.get("Asset ID").orElse(image.metadata.suppliersReference)
```
paperboyo, line 292: Let’s not read that from our copy of Getty data (even though it conveniently deals with inconsistent naming), because we should (and will) get rid of this stupid copy anyway. It was introduced when we weren’t yet reading the whole of XMP. Getty processor needs to move to identifying Getty images by existence of Asset ID, not, like now, by the existence of Getty namespace only (we will be able to remove all exclusions then, yay!).

Instead, let’s read it from main XMP block. If we are feeling generous, we could include naming reconcilliation of old, rare spelling (I think the old one is `AssetId` and correct, more common, one is `AssetID`).

AWare: spoil sport

AWare: Looking at it, we still use that code in the Getty Cleaner- and it's a bit of a faff to read it  from JsValue.

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -295,15 +312,25 @@ object GettyXmpParser extends ImageProcessor with GettyProcessor {
 
     val isExcludedByCredit = image.metadata.credit.exists(isExcluded(_, excludedCredit))
     val isExcludedBySource = image.metadata.source.exists(isExcluded(_, excludedSource))
-    val hasGettyMetadata = image.fileMetadata.getty.nonEmpty
+    val hasGettyMetadata = image.fileMetadata.getty.nonEmpty && getSuppliersReference(image).nonEmpty
```
AWare, line 315:
```suggestion
    val hasGettyMetadata = image.fileMetadata.getty.nonEmpty
```

paperboyo: Yep! Thanks. Let’s apply it separately when we will be making whole Getty business more sane.

paperboyo:
Functionality is good here, thanks! Ready for review, @guardian/digital-cms.

@ochiengolanga @adrielulanovsky FYI, this makes Guardian’s supplierProcessor errr… supply better supplierReference per-Supplier. You may want to bring those changes to your versions?

twrichards (approved):
Logic looks good 👍  (I see @paperboyo has tested)

PR #3279, opened 2021-04-27
Add pluggable usage rights configuration

ochiengolanga:
## What does this change?

Introduce the ability to plug in your own usage rights configuration provider. Allowing other adopters to specify their own sources e.g. external database, configuration file, values in code e.t.c by creating their own providers.

This is achieved by introducing a new `UsageRightsConfigProvider` - a class that provides usage configuration. Two implementations are provided. One is the `GuardianUsageRightsConfig` - which has hardcoded configuration in code that mimics what was done before. The second is the `RuntimeUsageRightsConfig` that fetches the data from the application configuration.

- The new configuration will look similar as shown below.

To use the hardcoded configuration:
```hocon
usageRightsConfigProvider = "com.gu.mediaservice.lib.guardian.GuardianUsageRightsConfig$"
```

- To use configuration from the application configuration files:

```hocon
usageRightsConfigProvider {
  className: "com.gu.mediaservice.lib.config.RuntimeUsageRightsConfig"
  config {
    externalStaffPhotographers = [ # list of external staff photographers grouped by publication
      {
        name = "Publication 1",
        photographers = ["John Doe"]
      }
      {
        name = "Publication 2",
        photographers = ["Larry Wolf"]
      }
      {
        name = "Publication 3",
        photographers = ["Hafeez Jackson"]
      }
    ]
    internalStaffPhotographers = [ # list of internal staff photographers grouped by publication
      {
        name = "Publication 1",
        photographers = ["Jane Doe"]
      }
    ]
    contractedPhotographers = [ # list of contracted photographers grouped by publication
      {
        name = "Contract Photographers 1",
        photographers = ["Peter Larry"]
      }
    ]
    contractIllustrators = [ # list of contracted illustrators grouped by publication
      {
        name = "Contract Illustrators 1",
        photographers = ["Tom Hardy"]
      }
    ]
    staffIllustrators = ["John Doe", "Jane Doe", "Larry Wolf"] # list of staff illustrators
    creativeCommonsLicense = [ # list of allowed creative commons licenses
      "CC BY-4.0",
      "CC BY-SA-4.0",
      "CC BY-ND-4.0"
    ]
  }
}
```

Introduces the ability for an adopter to specify a list of usage rights that are applicable. This is achieved via configuration as shown below.

```hocon
usageRights.applicable = [
  "com.gu.mediaservice.model.NoRights",
  "com.gu.mediaservice.model.Handout",
  "com.gu.mediaservice.model.PrImage",
[...]

PR #3329, opened 2021-05-27
fix missing options for the 'publication' field on CommissionedPhotographer and CommissionedIllustrator

twrichards:
https://github.com/guardian/grid/pull/3279

Bug introduced in https://github.com/guardian/grid/pull/3313 which was revert of a revert of https://github.com/guardian/grid/pull/3279 (FYI @ochiengolanga)

## What does this change?
Previously the options for the `publicationField` defaulted to the publications associated with staff photographers (i.e. `The Guardian` & `The Observer`, however in https://github.com/guardian/grid/pull/3279 where these were made configurable CommissionedPhotographer and CommissionedIllustrator were not updated to receive the same which resulted in...
image

This PR fixes that by passing the same info but from the new config provider approach.

## Tested?
- [x] locally by committer

### DIFF
--- a/metadata-editor/app/model/UsageRightsProperty.scala
+++ b/metadata-editor/app/model/UsageRightsProperty.scala
@@ -53,8 +53,6 @@ object UsageRightsProperty {
   ) = UsageRightsProperty(name, label, "string", required = true, options,
                           optionsMap, optionsMapKey, examples)
 
-  private def publicationField(required: Boolean): UsageRightsProperty = publicationField(required, List.empty)
-
   private def publicationField(required: Boolean, options: Options)  =
     UsageRightsProperty("publication", "Publication", "string", required,
       Some(sortList(options)))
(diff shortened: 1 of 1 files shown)


twrichards:
Hi @ochiengolanga (and @adrielulanovsky if @ochiengolanga is not available) - please can you confirm that removing the options for publication on CommissionedPhotographer and CommissionedIllustrator wasn't intentional in #3279 - I don't want this PR to break things for you?

SHession (approved):
Code changes and description make sense. I haven't tested locally due to issues with my local setup but happy to +1 for a quick turnaround.

ochiengolanga:
> Hi @ochiengolanga (and @adrielulanovsky if @ochiengolanga is not available) - please can you confirm that removing the options for publication on CommissionedPhotographer and CommissionedIllustrator wasn't intentional in #3279 - I don't want this PR to break things for you?

Hi @twrichards , just seeing this. Apologies for the late response and introducing a bug through the PR. It was not intentional and doesn't break anything. Thanks for the fix

PR #3334, opened 2021-06-01
update panda

AWare:
## What does this change?
this updates panda

that panda is updated, 
some login problems, don't want to debug them against an old version 

## Tested?
- [ ] locally by committer


twrichards (approved):
assuming you've tested on ~CODE~ `TEST` - synk isn't happy but worth a check to see if this helps ongoing login problems

AWare:
I'm not sure whether this will have any useful effect. Will continue debugging but will keep this open.

AWare:
think we should not do this and upgrade to latest panda with play 2.8

PR #3335, opened 2021-06-01
Set a speed on pngquant to make it faster

AWare:
## What does this change?
Set a speed on pngquant to make it faster

pngquant is faster and less likely to fail 

## Tested?
- [ ] locally by committer


paperboyo:
Thanks for not making it `10` :-). This is good. One day, we should stop optimising non-transparent images to PNG, which is stupid (too big files, 256 colours). They should just be [optimised to JPG](https://github.com/guardian/grid/issues/2395).

AWare:
@paperboyo it would have been more work to do so

twrichards (approved):
approving given simplicity of code change, but haven't tested myself (but I assume you and or @paperboyo have)

PR #3341, opened 2021-06-03
endpoint delete cookie for testing

AWare:
## What does this change?

This expires a users cookie when they hit the cookie-monster endpoint on the auth microservice. 

This is to enable some debugging to see why users are seeing a higher than usual number of reauth prompts.

the cookie monster eating a cookie

## Tested?
- [ ] locally by committer


mbarton (commented):
I ❤️ the cookie monster vibe but lets add a full description to the PR and a comment on the route to indicate why we're testing this

mbarton (approved):

PR #3342, opened 2021-06-03
Add inner-service authentication (aka cross-service)

tjsilver:
Co-authored-by: @twrichards 

## Context/Problem
https://trello.com/c/UhDV91uo/2333-grid-cross-service-authentication-process

Projection accesses all canonical sources by talking to all miscroservices in order to provide a diff (`/projection/diff`) between a current shape of the data and the shape the data would take would the image be ingested now.

We need to be able to talk freely to all the other microservices in the Grid. Thrall needs to be able to initiate the calls to these microservices itself, and not rely on api-keys or upstream UI actions triggering the calls.

## What does this change?

Introduces a new `InnerServicePrincipal` to be used when **originating** requests from a service in the Grid to another service in the Grid (for example, `thrall` calling `media-api`). This differs from 'MachinePrincipal` which is used when third-party (or services external to the Grid) originate a call to a Grid service and authenticate using an API key. 

This new inner service mechanism applies the following headers to the request: 
- `X-Inner-Service-Identity`: contains the name of the originating service, e.g. `thrall` and  possibly names of other services the request has gone via (see `onBehalfOf` section below)
- `X-Inner-Service-UUID`: contains a unique identifier for the request
- `X-Inner-Service-Timestamp`
- `X-Inner-Service-Signature`: contains a checksum of the above 3 headers, encrypted using the Play secret (which is already shared between the services).

**There is now an `innerServiceCall` on `Authentication` instances (which is already available in lots of the controllers, etc) which takes a `WSRequest`, signs the request using the mechanism above, and returns a new `WSRequest`.**

#### `onBehalfOf`

All `Principal`s require an `onBehalfOf` implementation so that services can call other services reusing the credentials of the service that called them. In the case of the new `InnerServicePrincipal` we append a "via" to the identity, then create a fresh UUID and timestamp and thus a new signature - so there is traceability of the hops taken from service to service. 

#### `innerServiceStatusCheck`

To demonstrate/check this new mechanism, we created `InnerServiceStatusCheckController` and added a `/management/InnerServiceStatusCheck` endpoint to `thrall` **(which is now regularly exercised via [Pingler](https://github.com/guardian/grid/blob/main/docs/02-running/01-running-locally.md#the_pinglersh))**. It calls all other internal services using `auth.innerServiceCall`, which then go on to call all other internal services again using the `auth.getOnBehalfOf` in order to demonstrate the "via" mentioned above.

This unblocks us from progressing with the reingestion project by allowing us to originate calls from `thrall` in a secure way, without having to generate, store/retrieve and rotate API keys.
 

## Who should look at this?
@guardian/digital-cms 

## Tested?
- [x] locally by committer

## Images
example logging demonstrating https://github.com/guardian/grid/pull/3342/commits/8026cd6d2ae37fb8c3df188d0b58692800e95c01
image

### DIFF
--- a/rest-lib/src/main/scala/com/gu/mediaservice/lib/management/InnerServiceStatusCheckController.scala
+++ b/rest-lib/src/main/scala/com/gu/mediaservice/lib/management/InnerServiceStatusCheckController.scala
@@ -24,9 +24,10 @@ class InnerServiceStatusCheckController(
 
   private def safeJsonParse(maybeJsonStr: String) = Try(Json.parse(maybeJsonStr)).getOrElse(JsString(maybeJsonStr))
 
-  private def callAllInternalServices(authenticator: WSRequest => WSRequest) = {
+  private def callAllInternalServices(depth: Int, authenticator: WSRequest => WSRequest) = {
+    val nextDepth = depth - 1
     val whoAmIFutures = services.allInternalUris.map { baseUri =>
-        authenticator(ws.url(s"$baseUri/management/whoAmI")).get
+        authenticator(ws.url(s"$baseUri/management/whoAmI").addQueryStringParameters("depth" -> nextDepth.toString)).get
           .map(resp => WhoAmIResponse(baseUri, resp.status, safeJsonParse(resp.body)))
           .recover{
             case throwable: Throwable => WhoAmIResponse(baseUri, SERVICE_UNAVAILABLE, Json.obj(
@@ -41,10 +42,14 @@ class InnerServiceStatusCheckController(
     }
   }
 
-  def whoAmI = auth.async { request =>
-    request.user match {
-      case principal: InnerServicePrincipal if !principal.identity.contains(" via ") =>
-        callAllInternalServices(auth.getOnBehalfOfPrincipal(principal))
+  def whoAmI(depth: Int) = auth.async { request =>
+    if (depth < 0 || depth > 2) { Future.successful(BadRequest("'depth' query param must be at least 0 and no more than 2"))}
+    else request.user match {
+      case principal: InnerServicePrincipal if depth > 0 =>
+        callAllInternalServices(
+          depth,
+          authenticator = auth.getOnBehalfOfPrincipal(principal)
+        )
       case _ =>
         Future.successful(
           Ok(Json.toJson(request.user.toString))
@@ -52,7 +57,11 @@ class InnerServiceStatusCheckController(
     }
   }
 
-  def statusCheck = Action.async {
-    callAllInternalServices(auth.innerServiceCall)
+  def statusCheck(depth: Int) = Action.async {
+    if (depth < 1 || depth > 3) { Future.successful(BadRequest("'depth' query param must be at least 1 and no more than 3"))}
+    else callAllInternalServices(
+      depth,
+      authenticator = auth.innerServiceCall
+    )
   }
 }
--- a/rest-lib/src/main/scala/com/gu/mediaservice/lib/management/InnerServiceStatusCheckController.scala
+++ b/rest-lib/src/main/scala/com/gu/mediaservice/lib/management/InnerServiceStatusCheckController.scala
@@ -1,6 +1,8 @@
 package com.gu.mediaservice.lib.management
 
 import com.gu.mediaservice.lib.argo.ArgoHelpers
+import com.gu.mediaservice.lib.auth.Authentication
+import com.gu.mediaservice.lib.auth.provider.InnerServiceAuthenticationProvider
 import com.gu.mediaservice.lib.config.Services
 import play.api.libs.json.{Json, Writes}
 import play.api.libs.ws.WSClient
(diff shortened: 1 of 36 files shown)


AWare:
innerServiceStatusCheck sounds like it'll amplify traffic a bit too neatly.

twrichards:
> innerServiceStatusCheck sounds like it'll amplify traffic a bit too neatly.

Only called by pingler currently. No plans to make it a regular thing in production. that's why we didn't make it part of the main healthcheck

AWare:
Yeah, if part of the main healthcheck, it could lead to some nasty race condition like behaviour on deploys which could cause cycling. I think it would be as useful if it didn't fan out again though, and just checked that each other service is available at all.

twrichards:
> Yeah, if part of the main healthcheck, it could lead to some nasty race condition like behaviour on deploys which could cause cycling. I think it would be as useful if it didn't fan out again though, and just checked that each other service is available at all.

How about by default it doesn't fan out, but it does take a query param to fan out if desired (and that pingler uses said param)

twrichards:
> > Yeah, if part of the main healthcheck, it could lead to some nasty race condition like behaviour on deploys which could cause cycling. I think it would be as useful if it didn't fan out again though, and just checked that each other service is available at all.
> 
> How about by default it doesn't fan out, but it does take a query param to fan out if desired (and that pingler uses said param)

Actually this is a silly compromise I've suggested because in practice the traffic is the same since pingler is really the only thing using this new endpoint and I was suggesting it still forces it to fan out. 

@AWare how strongly do you feel about the fact it shouldn't fan out? Personally, I think it's worth the 81 (9²) very lightweight calls every few seconds (rather than just 9 calls) to guard against anyone breaking the `onBehalfOf` functionality without realising. Are you worried that this extra traffic will affect other services when running locally?

twrichards:
so interestingly, the innerServiceStatusCheck fails in TEST, because... 
image
... ironically I think `admin-tools` is the home of one of the projection endpoints, but as its README points out it runs differently locally compared to TEST & PROD where it runs as a Lambda and looks like it has a different URL structure. Need to work out if this is needed...
- if so, adjust the URL building logic to build correct URL for this service or perhaps request the DNS changes to put the lambda on the expected/consistent domain
- if not, we can remove it from the list of 'inner services'

rest-lib/src/main/scala/com/gu/mediaservice/lib/management/InnerServiceStatusCheckController.scala
```diff
@@ -0,0 +1,58 @@
...
+
+    Future.sequence(whoAmIFutures).map { whoAmIResponses =>
+      val overallStatus = whoAmIResponses.map(_.status).max
+      new Status(overallStatus)(Json.toJson(whoAmIResponses.map(resp => resp.baseUri -> resp).toMap))
+    }
+  }
+
+  def whoAmI = auth.async { request =>
+    request.user match {
+      case principal: InnerServicePrincipal if !principal.identity.contains(" via ") =>
```
jonathonherbert, line 46: Worth parameterising? The only reason I mention – IIUC if this doesn't match (e.g. a dev makes a change to 'via' as it's added) we'll get stuck in an exponentiating loop.

jonathonherbert (approved):
This looks good. Tested locally, and I can see 200s with the appropriate information. I love that we're keeping tracer-bullet UUIDs for these requests.

Do we know what our logging looks like with this in place? I can't see anything added in the code here, and I can't see any markers in my local logs that give e.g. request UUIDs. (I _can_ see `X-Gu-Original-Service`.) I wonder whether it might be useful to surface these new headers, perhaps [here](https://github.com/guardian/grid/blob/e669dfc4b01aa4fdfa82893d535e4a86558319b3/rest-lib/src/main/scala/com/gu/mediaservice/lib/play/RequestLoggingFilter.scala)?

As I understand it, if we introduced a change that broke this, healthcheck would not fail. Using pingler to test that things work locally ad hoc is very useful – but I don't think that relying on Pingler to surface problems before they hit PROD is sensible. Having said that, there are plenty of other business-level behaviours that our healthcheck doesn't verify, so perhaps this is an acceptable risk in an area of the codebase that shouldn't be changed too often.

Other than that – one v. minor comment, otherwise, great work all! 👍

jonathonherbert:
To follow up re: admin-tools (this was posted out-of-band, but including here for completeness)

a) The admin-tools service would be better as patched onto an existing service, rather than requiring a whole other set of precarious (and inconsistent) infra

b) my understanding (based on an incomplete understanding of the current work and also my terrible memory, please correct) is that it's responsible for precisely the sort of operation that Thrall is about to become responsible for, e.g. marshalling data from different services and providing a canonical version. (idk how we 'dry run' this sort of thing w/ Thrall, mind).

I'd say remove from this list until we know we need it. If we do find that we need it, I suspect the proper solution will be to move its responsibilities somewhere else.

twrichards:
Need to update https://github.com/guardian/grid/blob/main/docs/07-extending/03-authentication.md and ahttps://github.com/guardian/grid/blob/main/docs/03-apis/01-authentication.md before merging 
(thanks @paperboyo and @akash1810 for pointing me to these)

rest-lib/src/main/scala/com/gu/mediaservice/lib/management/InnerServiceStatusCheckController.scala
twrichards: good point, funnily enough thats how i first implemented it, then chose the above for succinctness (but on reflection not worth the risk of it going haywire). reimplemented `depth` param in https://github.com/guardian/grid/pull/3342/commits/f1ff6ea5ba88962bb8763bd538e224d091dc6197 (capped at max 3)

PR #3343, opened 2021-06-03
Upgrade Sentry

AWare:
our version of sentry is quite old, this upgrades it to the latest one. 
this allows us better handling of logging, and should let us see into the console logs and get a good idea about the auth timeouts

it will need `https://app.getsentry.com/` in connect-src in kahuna

### DIFF
--- a/kahuna/package-lock.json
+++ b/kahuna/package-lock.json
@@ -4977,6 +4977,74 @@
         }
       }
     },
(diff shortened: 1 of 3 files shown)


mbarton (approved):
LGTM and happy to add sentry to the `connect-src` config in S3

twrichards (approved):

PR #3344, opened 2021-06-03
what's one more terrible hack between friends (hit the root api every fifteen minutes so google auth doesn't time out)

AWare:
pr description left blank from shame

### DIFF
--- a/kahuna/public/js/main.js
+++ b/kahuna/public/js/main.js
@@ -304,12 +304,19 @@ kahuna.run(['$rootScope', '$window', '$q', 'getEntity',
 
 
 kahuna.controller('SessionCtrl',
-                  ['$scope', '$state', '$stateParams', 'mediaApi',
+                  ['$scope', '$state', '$stateParams', '$interval', 'mediaApi',
                    function($scope, $state, $stateParams, mediaApi) {
 
     mediaApi.getSession().then(session => {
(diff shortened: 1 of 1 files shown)


twrichards:
amazing if this works, but would be not better to do something server side on all API endpoints (which presumably fail when auth is expired) which do whatever the root endpoint is doing?

AWare:
the root endpoint isn't doing anything special here, they all fail in the same way (with a 419), which causes pandular to force a reauth (because it wraps $http) 

if the auth session has timed out _a little_, then we can silently recover, but if it's too long, then we need to do an interactive reauth. the aim here is to not let the grid go too long without hitting an auth'd endpoint and thus timing out for the silent reauth

twrichards:
> the root endpoint isn't doing anything special here, they all fail in the same way (with a 419), which causes pandular to force a reauth (because it wraps $http)
> 
> if the auth session has timed out _a little_, then we can silently recover, but if it's too long, then we need to do an interactive reauth. the aim here is to not let the grid go too long without hitting an auth'd endpoint and thus timing out for the silent reauth

makes sense, thanks for the explanation, with that said is their a lighterweight endpoint we could hit?

AWare:
I think it would be unlikely/premature to do so.

twrichards (approved):

PR #3345, opened 2021-06-04
Extract Guardian specific messages into config

adrielulanovsky:
## What does this change?
Currently, there are many Guardian-specific messages in the UI. This is an effort to make these messages configurable.

The following are some of the things that BBC wants to customise:
- References to "the Grid" will be changed to "BBC Images".
- References to "The Guardian" will be changed to "BBC".
- References to "The Observer" will be changed to "BBC Studio".

Regarding the systemName config, I've noticed there's not a uniform capitalisation of it; some places it's written as "the grid", some as "the Grid", others as "The Grid". My approach was to default systemName to "the Grid" to avoid having to add 3 different configs for this value. However, there's one place where I had to capitalise the first letter via code, because it was at the start of a sentence.

In one place "Grid" was used without the article ("Grid-specific usage rights") so I decided to make that whole text configurable.

The other message I extracted completely was "Drop files or GuardianWitness URLs here to upload them instantly to the Grid". BBC wants this phrase to read as "Drop files or URLs here to upload them instantly to BBC Images".

Same thing for the usage of "Observer". One instance had lowercase while the other had uppercase first letter. I decided to use "Observer" as default value.

## Who should look at this?
@guardian/digital-cms

## Tested?
- [x] locally by committer

### DIFF
--- a/kahuna/app/controllers/KahunaController.scala
+++ b/kahuna/app/controllers/KahunaController.scala
@@ -41,12 +41,7 @@ class KahunaController(
       config.scriptsToLoad.filter(_.permission.map(authorisation.hasPermissionTo).fold(true)(maybeUser.exists)),
       config.staffPhotographerOrganisation,
       config.homeLinkHtml,
-      config.systemName,
-      config.dropzoneExplanation,
-      config.advancedSearchExampleExplanation,
-      config.exampleEmail,
-      config.exampleLabel,
-      config.institution
+      config.systemName
     ))
   }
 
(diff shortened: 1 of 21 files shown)


AWare:
It might be easier to do these strings with some webpack magic instead of injecting into the html.

adrielulanovsky:
> It might be easier to do these strings with some webpack magic instead of injecting into the html.

Sorry, I'm not too familiar with webpack. Do you mean extracting configuration into a file, like this? https://webpack.js.org/guides/asset-management/#loading-data. If that's the case, I think there's one instance (systemName) where this is used in the Scala template, so I'm not sure it can be replaced via webpack (unless there's more advanced magic that I'm not aware of). 

I can try to take that approach for the other variables though

AWare:
I'm thinking we have a `strings.js` file and then set the replacements in that.

adrielulanovsky:
> I'm thinking we have a `strings.js` file and then set the replacements in that.

I extracted most of the configs (except systemName, for the reason already mentioned) into a strings.json file

AWare (approved):

paperboyo:
Tested, thanks, @adrielulanovsky!

PR #3350, opened 2021-06-12
Extract crop options

adrielulanovsky:
## What does this change?
BBC would like to display different cropping options than the ones used by The Guardian. To be able to achieve that, this PR extracts crop options to a file that can be overriden during the build process to modify the options and/or order of the list.

Cropping options are not modified in The Guardian's Grid.
BBC can customize their cropping options.

## Screenshots
How the cropping options would show when customized:
image

## Who should look at this?
@guardian/digital-cms

## Tested?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/util/constants/cropOptions.js
+++ b/kahuna/public/js/util/constants/cropOptions.js
@@ -1,0 +1,8 @@
+// `ratioString` is sent to the server, being `undefined` for `freeform` is expected 🙈
+export const landscape = {key: 'landscape', ratio: 5 / 3, ratioString: '5:3'};
+export const portrait = {key: 'portrait', ratio: 4 / 5, ratioString: '4:5'};
+export const video = {key: 'video', ratio: 16 / 9, ratioString: '16:9'};
+export const square = {key: 'square', ratio: 1, ratioString: '1:1'};
+export const freeform = {key: 'freeform', ratio: null};
+
+export const cropOptions = [landscape, portrait, video, square, freeform];
--- a/kahuna/public/js/util/crop.js
+++ b/kahuna/public/js/util/crop.js
@@ -1,20 +1,13 @@
 import angular from 'angular';
 
+import { landscape, portrait, video, square, freeform, cropOptions } from './constants/cropOptions';
+
 const CROP_TYPE_STORAGE_KEY = 'cropType';
 const CUSTOM_CROP_STORAGE_KEY = 'customCrop';
-
-// `ratioString` is sent to the server, being `undefined` for `freeform` is expected 🙈
-const landscape = {key: 'landscape', ratio: 5 / 3, ratioString: '5:3'};
-const portrait = {key: 'portrait', ratio: 4 / 5, ratioString: '4:5'};
-const video = {key: 'video', ratio: 16 / 9, ratioString: '16:9'};
-const square = {key: 'square', ratio: 1, ratioString: '1:1'};
-const freeform = {key: 'freeform', ratio: null};
 
 const customCrop = (label, xRatio, yRatio) => {
   return { key:label, ratio: xRatio / yRatio, ratioString: `${xRatio}:${yRatio}`};
 };
-
-const cropOptions = [landscape, portrait, video, square, freeform];
 
 export const cropUtil = angular.module('util.crop', ['util.storage']);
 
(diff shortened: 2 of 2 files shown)


paperboyo:
If this plays well with https://github.com/guardian/grid/pull/3017, I can’t see why not! cc @akash1810

kahuna/public/js/util/constants/cropOptions.js
```diff
@@ -0,0 +1,8 @@
+// `ratioString` is sent to the server, being `undefined` for `freeform` is expected 🙈
+export const landscape = {key: 'landscape', ratio: 5 / 3, ratioString: '5:3'};
+export const portrait = {key: 'portrait', ratio: 4 / 5, ratioString: '4:5'};
+export const video = {key: 'video', ratio: 16 / 9, ratioString: '16:9'};
+export const square = {key: 'square', ratio: 1, ratioString: '1:1'};
+export const freeform = {key: 'freeform', ratio: null};
+
+export const cropList = [landscape, portrait, video, square, freeform];
```
twrichards, line 8: if we renamed this, then we could remove line 12 of `kahuna/public/js/util/crop.js`
```suggestion
export const cropOptions = [landscape, portrait, video, square, freeform];
```

kahuna/public/js/util/crop.js
```diff
@@ -1,20 +1,15 @@
...
-const video = {key: 'video', ratio: 16 / 9, ratioString: '16:9'};
-const square = {key: 'square', ratio: 1, ratioString: '1:1'};
-const freeform = {key: 'freeform', ratio: null};
-
 const customCrop = (label, xRatio, yRatio) => {
   return { key:label, ratio: xRatio / yRatio, ratioString: `${xRatio}:${yRatio}`};
 };
 
-const cropOptions = [landscape, portrait, video, square, freeform];
+const cropOptions = cropList;
```
twrichards, line 12: this line could be removed if the thing being imported was called cropOptions

kahuna/public/js/util/crop.js
```diff
@@ -24,7 +19,7 @@ cropUtil.constant('video', video);
 cropUtil.constant('square', square);
 cropUtil.constant('freeform', freeform);
 cropUtil.constant('cropOptions', cropOptions);
-cropUtil.constant('defaultCrop', landscape);
+cropUtil.constant('defaultCrop', cropOptions[0]);
```
twrichards, line 22: I think this should stay as is, as the fact that the default is 'landscape' is more explicit/readable
```suggestion
cropUtil.constant('defaultCrop', landscape);
```

twrichards (approved):
few minor comments but LGTM 👍

kahuna/public/js/util/crop.js
twrichards: but if you don't want landscape as default, then ignore this suggestiong and stick with `cropOptions[0]`

adrielulanovsky: The issue here is that BBC doesn't want the landscape option, and the default will be a BBC specific one. That's why I set it to the first option in cropOptions (which in BBC will be a different option, not landscape). 

Perhaps we could add a flag inside the options that says "default: true" and then check on that to make it more explicit. What do you think?

twrichards: thats a nice idea, but don't worry, I think first item in the list is sufficient, probs not worth the extra complexity (in ensuring that its set on one [and only one] entry)

kahuna/public/js/util/constants/cropOptions.js
adrielulanovsky: Fixed

kahuna/public/js/util/crop.js
adrielulanovsky: Fixed

twrichards (approved):

paperboyo:
FYI, just deployed it to TEST and Landscape and Portrait crop options disappeared:

TEST:
image

PROD:
image

adrielulanovsky:
> FYI, just deployed it to TEST and Landscape and Portrait crop options disappeared:
> 
> TEST:
> image
> 
> PROD:
> image

Oops, sorry for that, I accidentally committed BBC's file instead of the Guardian one. Could you try again now?

paperboyo:
> Could you try again now?

Works perfectly, thanks! Merging tomorrow.

twrichards (approved):

PR #3353, opened 2021-06-14
introduce soft delete image endpoint

abdelrahmansd:
## What does this change?
This implements a new endpoint: Soft Delete images

Given `DELETE     media-api/soft-delete/images/:id` endpoint
Then `softDeletedMetadata` fields should be added in the root of the document 
`                      "softDeletedMetadata": {
                        "deleteTime": "2021-06-17T03:52:37.084+02:00",
                        "deletedBy": "johndoe@example.com"
                    }`

## Tested?
- [x] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
@@ -101,6 +101,17 @@ class DynamoDB[T](config: CommonConfig, tableName: String, lastModifiedKey: Opti
       valueMapWithNullForEmptyString(Map(":value" -> value))
     )
 
+  def stringListSet(id: String, keyValues: (String, JsValue)*)
+                   (implicit ex: ExecutionContext): Future[JsObject] = {
+    val keyValueMap = keyValues.toMap
+    val expressionParts = keyValueMap.keys.map(key => s"$key = :$key")
+    val valueMap = keyValueMap.map {case (key, value) => (s":$key", value)}
+    update(
+      id,
+      expression = s"SET ${expressionParts.mkString(",")}",
+      valueMapWithNullForEmptyString(valueMap)
+    )
+  }
 
   def setGet(id: String, key: String)
             (implicit ex: ExecutionContext): Future[Set[String]] =
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
@@ -52,6 +52,7 @@ object Mappings {
         exportsMapping("exports"),
         dateField("uploadTime"),
         keywordField("uploadedBy"),
+        softDeletedMetadataMapping("softDeletedMetadata"),
         dateField("lastModified"),
         dynamicObj("identifiers"),
         uploadInfoMapping("uploadInfo"),
@@ -281,6 +282,11 @@ object Mappings {
   def leasesMapping(name: String): ObjectField = nonDynamicObjectField(name).fields(
     leaseMapping("leases"),
     dateField("lastModified")
+  )
+
+  def softDeletedMetadataMapping(name: String) = nonDynamicObjectField(name).fields(
+    dateField("deleteTime"),
+    keywordField("deletedBy")
   )
 
   private def nonDynamicObjectField(name: String) = ObjectField(name).dynamic("strict")
--- a/media-api/app/controllers/MediaApi.scala
+++ b/media-api/app/controllers/MediaApi.scala
@@ -5,7 +5,7 @@ import com.google.common.net.HttpHeaders
 import com.gu.mediaservice.lib.argo._
 import com.gu.mediaservice.lib.argo.model.{Action, _}
 import com.gu.mediaservice.lib.auth.Authentication._
-import com.gu.mediaservice.lib.auth.Permissions.{ArchiveImages, DeleteCrops, DeleteImage => DeleteImagePermission, EditMetadata, UploadImages}
+import com.gu.mediaservice.lib.auth.Permissions.{ArchiveImages, DeleteCrops, EditMetadata, UploadImages, DeleteImage => DeleteImagePermission}
 import com.gu.mediaservice.lib.auth._
 import com.gu.mediaservice.lib.aws.{S3Metadata, ThrallMessageSender, UpdateMessage}
 import com.gu.mediaservice.lib.formatting.printDateTime
@@ -14,15 +14,15 @@ import lib._
 import lib.elasticsearch._
 import org.apache.http.entity.ContentType
 import org.http4s.UriTemplate
-import org.joda.time.DateTime
+import org.joda.time.{DateTime, DateTimeZone}
 import play.api.http.HttpEntity
 import play.api.libs.json._
 import play.api.libs.ws.WSClient
 import play.api.mvc.Security.AuthenticatedRequest
 import play.api.mvc._
-
 import java.net.URI
 import java.util.concurrent.TimeUnit
+
 import com.gu.mediaservice.GridClient
 import com.gu.mediaservice.JsonDiff
 import com.gu.mediaservice.lib.config.{ServiceHosts, Services}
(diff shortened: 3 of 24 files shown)


twrichards:
LGTM so far - will wait for `deletedBy` before doing a final review

metadata-editor/app/controllers/EditsController.scala
```diff
@@ -212,6 +213,19 @@ class EditsController(
     (Json.toJson(metadata).as[JsObject]).as[Map[String, JsValue]]
   }
 
+  def softDelete(id: String) = auth.async { implicit req =>
+    val result = for {
+      _ <- editsStore.stringSet(id, Edits.DeletedAt, JsString(DateTime.now.toString))
```
twrichards, line 218: it looks like we're discarding the output of this command and mapping on the result of setting the deletedBy only, is there a way we could combine them into a single command to set both fields?

twrichards (commented):
it all looks nice and clean, I will test in our TEST environment tomorrow morning. One important comment on the code in the meantime though.

metadata-editor/app/controllers/EditsController.scala
abdelrahmansd: yes we can use `jsonAdd` to add both
will update the PR now

metadata-editor/app/controllers/EditsController.scala
```diff
@@ -212,6 +213,16 @@ class EditsController(
     (Json.toJson(metadata).as[JsObject]).as[Map[String, JsValue]]
   }
 
+  def softDelete(id: String) = auth.async { implicit req =>
+    editsStore.stringListSet(id, Map(Edits.DeletedAt -> JsString(DateTime.now.toString),
+      Edits.DeletedBy -> JsString(req.user.accessor.identity)))
+      .map(publish(id, UpdateImageUserMetadata))
```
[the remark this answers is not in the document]
twrichards: if however you don't like the _varargs_ part of my other suggestion (I don't have a strong preference), then please can we neaten up the call signature to make clear the two inputs, something like...
```suggestion
    editsStore.stringListSet(
       id, 
       Map(
          Edits.DeletedAt -> JsString(DateTime.now.toString),
          Edits.DeletedBy -> JsString(req.user.accessor.identity)
        )
      )
      .map(publish(id, UpdateImageUserMetadata))
```

twrichards (commented):
thanks for updating to make the two field updates in a single operation

couple of style suggestions, but otherwise looks good. I'm going to test in our TEST environment now.

twrichards:
tested in `TEST`, see the metadata after hitting the endpoint 
image

metadata-editor/app/controllers/EditsController.scala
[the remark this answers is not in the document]
twrichards: done in https://github.com/guardian/grid/pull/3353/commits/b4c9fc9cd92e0267dfcee64540a636036b01b151

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/DynamoDB.scala
```diff
@@ -101,6 +101,16 @@ class DynamoDB[T](config: CommonConfig, tableName: String, lastModifiedKey: Opti
...
+  def stringListSet(id: String, keyValueMap: Map[String, JsValue])
+               (implicit ex: ExecutionContext): Future[JsObject] = {
+    val query = keyValueMap.foldLeft("SET ")((acc, kv) => acc + kv._1 + s" = :${kv._1},").dropRight(1)
+    val valueMap = keyValueMap.map {case (key, value) => (s":$key", value)}
+    update(
+      id,
+      query,
+      valueMapWithNullForEmptyString(valueMap)
+    )
+  }
```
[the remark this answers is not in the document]
twrichards: done in https://github.com/guardian/grid/pull/3353/commits/b4c9fc9cd92e0267dfcee64540a636036b01b151

twrichards (approved):
Now I've tested this, I'm happy to merge tomorrow. Nice work 🙌 

We really ought to add this endpoint to the 'hypermedia' responses on the GET endpoints (i.e. when reading this information it provides the url for all the actions, in this case `soft-delete` so the client doesn't need to build the URL itself). Will leave it up to you whether you want to make that part of this PR or a follow-up? (You'll need it when you move onto the client-side portion)

twrichards:
we're observing that `thrall` isn't booting up on this branch, so I'm going to hold-off from merging for now. The problem relates to adding these fields to ES, you got any thoughts @akash1810 ?
image

paperboyo:
Inability to change the existing mappings is one of the reasons we are doing the reingestion work. But I thought adding a new one is not a problem. No idea if [this script](https://github.com/guardian/grid/tree/main/scripts#updatemapping) is actually how you do it, though…

abdelrahmansd:
@twrichards it will require mapping update because `userMetadataMapping` set to strict

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -188,7 +188,9 @@ object Mappings {
     metadataMapping("metadata"),
     usageRightsMapping("usageRights"),
     photoshootMapping("photoshoot"),
-    dateField("lastModified")
+    dateField("lastModified"),
+    dateField("deletedAt"),
+    keywordField("deletedBy")
```
akash1810, line 193: I don't think this is the correct location for these fields as it suggests this is the time the user metadata was deleted, rather than the image itself, which is what we want to represent. A better location would be at the [root of the document as siblings to `uploadTime` and `uploadedBy`](https://github.com/guardian/grid/blob/c7ccb0ab6603209aaa816188e2547fb7394d90af/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala#L53-L54).

We may also want to call it `deletedTime` for consistency?

twrichards:
> Inability to change the existing mappings is one of the reasons we are doing the reingestion work. But I thought adding a new one is not a problem. No idea if [this script](https://github.com/guardian/grid/tree/main/scripts#updatemapping) is actually how you do it, though…

Running this script did the trick, thanks @paperboyo (and @akash1810 for mobbing on running it). I think the point that @akash1810  raises about placement of these fields is important though, so will hold off merging until for now https://github.com/guardian/grid/pull/3353#discussion_r652569008

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
abdelrahmansd: @akash1810 yes it will be making more sense 
will change the name and move it to the `Image` model

abdelrahmansd: @akash1810 I couldn't add `deleteTime` and `deletedBy` at the root of the document  because we hit  function parameters maximum number `implementation restricts functions to 22 parameters`
so I added them into a nested object `softDeletedMetadata`

akash1810: Ah, Scala, such a wonderful language 😉 .

media-api/app/controllers/MediaApi.scala
```diff
@@ -226,6 +226,21 @@ class MediaApi(
...
+  def softDeleteImage(id: String) = auth.async { request =>
+    implicit val r = request
+
+    elasticSearch.getImageById(id) map {
+      case Some(_) =>
+        val updateMessage = UpdateMessage(subject = SoftDeleteImage, id = Some(id),
+          softDeletedMetadata = Some(SoftDeletedMetadata(
+            DateTime.now(DateTimeZone.UTC),
+            request.user.accessor.identity)))
+        messageSender.publish(updateMessage)
```
twrichards, line 238: no need for val as case class name describes what it is, named params on the SoftDeletedMetadata args and changed line breaks & indentation to make things clearer  
```suggestion
        messageSender.publish(
          UpdateMessage(
            subject = SoftDeleteImage, 
            id = Some(id),
            softDeletedMetadata = Some(SoftDeletedMetadata(
              deleteTime = DateTime.now(DateTimeZone.UTC),
              deletedBy = request.user.accessor.identity
            )
          )
        )
```

twrichards (commented):
Just testing now, one minor comment in the meantime - otherwise looks good 👍 

Also if you don't mind, please could you squash the commits (I would happily do it at merge time but don't want to mess up BBC when you pull the changes back through to the bbc repo)

media-api/app/controllers/MediaApi.scala
abdelrahmansd: @twrichards will update and squash the PR now

twrichards (approved):
re-tested 🎉 
image

Amazing work and thanks for your patience though the various suggestions/iterations.

Will merge once squashed :)

PR #3356, opened 2021-06-21
reusing DeleteImages Permission on the soft-delete endpoint

abdelrahmansd:
## What does this change?
reusing DeleteImages Permission on the soft-delete endpoint to control which users can view or soft-delete images on the Grid

- Given `DELETE     media-api/soft-delete/images/:id` endpoint
Then `softDeletedMetadata` fields should be added in the root of the document 
`                      "softDeletedMetadata": {
                        "deleteTime": "2021-06-17T03:52:37.084+02:00",
                        "deletedBy": "johndoe@example.com"
                    }`
                    

- Given `DELETE     media-api/soft-delete/images/:id` endpoint
And a user who isn't in the right group 
Then it should respond with `permission-denied`

## Tested?
- [x] locally by committer

### DIFF
--- a/media-api/app/controllers/MediaApi.scala
+++ b/media-api/app/controllers/MediaApi.scala
@@ -73,7 +73,6 @@ class MediaApi(
 
     val maybeLoaderLink: Option[Link] = Some(Link("loader", config.loaderUri)).filter(_ => userCanUpload)
     val maybeArchiveLink: Option[Link] = Some(Link("archive", s"${config.metadataUri}/metadata/{id}/archived")).filter(_ => userCanArchive)
-
     val indexLinks = List(
       searchLink,
       Link("image",           s"${config.rootUri}/images/{id}"),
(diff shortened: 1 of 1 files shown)


twrichards (approved):
Thanks for taking onboard the suggestion to re-use the existing `DeleteImagePermission` much simpler/cleaner now. Code looks good 👍 . I've tested and can confirm 
- `405` when attempting to soft-delete an image with usages (expected AFAIK)
- `202` when soft-deleting a random image when I have `DeleteImagePermission` (and `softDeletedMetadata` added as before)
- `403` when soft-deleting a random image without the permission
image
- `202` when deleting an image I uploaded even though I don't have `DeleteImagePermission`

🎉 

Before I merge tomorrow please can you update the PR description to match the final state of the code, might be worth adjusting the PR title too.

Great work @abdelrahmansd 🙌  and appreciate your patience awaiting this review.

PR #3357, opened 2021-06-21
Download crops

adrielulanovsky:
## What does this change?
Currently, when selecting a crop and clicking the Download button, only the original image is downloaded and not the selected crop. Also, when clicking on a link to a crop in the Show Crops section of the metadata, you are redirected to the image instead of downloading it. Another issue is that this doesn't log a usage for the image. 

In this PR I'll try to address these issues by creating an endpoint in the media-api service to download a crop:
`GET     /images/:imageId/export/:exportId/asset/:width/download`

This endpoint will record the download as a usage (if having the 'image.record.download' config), fetch the crop, and return the cropped image the same way the /download endpoint does for original images.

The UI will display an option called "Download selected crop" below "Download low-res" option, and clicking on it will automatically start the download of the crop.

This feature will be hidden behind a config flag: `canDownloadCrop`, since it has been pointed to me that The Guardian doesn't want this functionality.

- The Guardian's Grid maintains its current behaviour. 
- On BBC Grid, when selecting a crop, you can now see a "Download selected crop" option and clicking on it will automatically start the download of the crop. Also, clicking on a link to a crop in the Show Crops section of the metadata will trigger the download and add a usage as well.

## Screenshots
image

## Who should look at this?
@guardian/digital-cms

## Tested?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/image/controller.js
+++ b/kahuna/public/js/image/controller.js
@@ -208,7 +208,8 @@ image.controller('ImageCtrl', [
       return ctrl.image.data.source.dimensions;
     }
 
-    mediaCropper.getCropsFor(image).then(crops => {
+    ctrl.cropsForImage = mediaCropper.getCropsFor(image)
+    ctrl.cropsForImage.then(crops => {
       ctrl.crop = crops.find(crop => crop.id === cropKey);
       ctrl.fullCrop = crops.find(crop => crop.specification.type === 'full');
       ctrl.crops = crops.filter(crop => crop.specification.type === 'crop');
--- a/kahuna/public/js/image/controller.js
+++ b/kahuna/public/js/image/controller.js
@@ -208,8 +208,7 @@ image.controller('ImageCtrl', [
       return ctrl.image.data.source.dimensions;
     }
 
-    ctrl.cropsForImage = mediaCropper.getCropsFor(image)
-    ctrl.cropsForImage.then(crops => {
+    mediaCropper.getCropsFor(image).then(crops => {
       ctrl.crop = crops.find(crop => crop.id === cropKey);
       ctrl.fullCrop = crops.find(crop => crop.specification.type === 'full');
       ctrl.crops = crops.filter(crop => crop.specification.type === 'crop');
--- a/kahuna/public/js/image/controller.js
+++ b/kahuna/public/js/image/controller.js
@@ -208,7 +208,14 @@ image.controller('ImageCtrl', [
       return ctrl.image.data.source.dimensions;
     }
 
-    mediaCropper.getCropsFor(image).then(crops => {
+    mediaCropper.getCropsFor(image).then(cropsResource => {
+      let crops = cropsResource.data;
+      debugger;
+      if ($window._clientConfig.canDownloadCrop) {
+        crops.forEach((crop) =>
+          crop.downloadLink = cropsResource.links.find(link => link.rel.includes(`crop-${crop.id}-download`))?.href
+        );
+      }
       ctrl.crop = crops.find(crop => crop.id === cropKey);
       ctrl.fullCrop = crops.find(crop => crop.specification.type === 'full');
       ctrl.crops = crops.filter(crop => crop.specification.type === 'crop');
--- a/kahuna/public/js/image/controller.js
+++ b/kahuna/public/js/image/controller.js
@@ -210,11 +210,15 @@ image.controller('ImageCtrl', [
 
     mediaCropper.getCropsFor(image).then(cropsResource => {
       let crops = cropsResource.data;
-      debugger;
       if ($window._clientConfig.canDownloadCrop) {
-        crops.forEach((crop) =>
-          crop.downloadLink = cropsResource.links.find(link => link.rel.includes(`crop-${crop.id}-download`))?.href
-        );
+        crops.forEach((crop) => {
+          crop.assets.forEach((asset) =>
+            asset.downloadLink = cropsResource.links.find(link => link.rel.includes(`crop-download-${crop.id}-${asset.dimensions.width}`))?.href
+          );
+          //set the download link of the crop to be the largest asset
+          let largestAsset = crop.assets.find(asset => asset.dimensions.width == crop.master.dimensions.width);
+          crop.downloadLink = largestAsset.downloadLink;
+        });
       }
       ctrl.crop = crops.find(crop => crop.id === cropKey);
       ctrl.fullCrop = crops.find(crop => crop.specification.type === 'full');
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
@@ -76,43 +76,56 @@ class S3(config: CommonConfig) extends GridLogging {
     regex.replaceAllIn(filename, "")
   }
 
-  private def getContentDispositionFilename(image: Image, imageType: ImageType, charset: Charset, maybeCrop: Option[Crop] = None): String = {
+  private def getExtension(image: Image, asset: Asset): String = asset.mimeType match {
+    case Some(mimeType) => mimeType.fileExtension
+    case _ =>
+      logger.warn(image.toLogMarker, "Unrecognised mime type")
+      ""
+  }
 
-    lazy val imageAsset = imageType match {
-      case Source => image.source
-      case Thumbnail => image.thumbnail.getOrElse(image.source)
-      case OptimisedPng => image.optimisedPng.getOrElse(image.source)
-    }
-
-    val asset = maybeCrop.flatMap(_.master).getOrElse(imageAsset)
-
-    val extension: String = asset.mimeType match {
-      case Some(mimeType) => mimeType.fileExtension
-      case _ =>
-        logger.warn(image.toLogMarker, "Unrecognised mime type")
-        ""
-    }
-
-    val maybeCropId = maybeCrop.flatMap(_.id).map(id => s"($id)").getOrElse("")
-    val baseFilename: String = image.uploadInfo.filename match {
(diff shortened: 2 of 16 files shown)


paperboyo:
> Also, when clicking on a link to a crop in the Show Crops section of the metadata, you are redirected to the image instead of downloading it.

After this is merged in:
1. will this link look any different for the Guardian (with this switched off):

image

2. Will these links look different with this switched on? How will they look?

I am asking, because we are contemplating how useful would that functionality be for us. Very small minority of our users still use Show crops palette to basically just quickly grab URLs (Copy Link in the context menu) to specific crops to use in tooling that has no proper integration with Grid.

adrielulanovsky:
> > Also, when clicking on a link to a crop in the Show Crops section of the metadata, you are redirected to the image instead of downloading it.
> 
> After this is merged in:
> 
>     1. will this link look any different for the Guardian (with this switched off)?
> 
> 
> image
> 
>     1. Will these links look different with this switched on? How will they look?
> 
> 
> I am asking, because we are contemplating how useful would that functionality be for us. Very small minority of our users still use Show crops palette to basically just quickly grab URLs (Copy Link in the context menu) to specific crops to use in tooling that has no proper integration with Grid.

1. It will look the same with this switched off.
2. Since I assumed The Guardian didn't have a need for this, for simplicity's sake I decided to only show the full crop and not the low-res ones. So it will look as only one option for each crop. I did this because the endpoint I created only downloads the full crop, and BBC didn't have a need for the other ones. However, I could try to change this so that the endpoint can download lower-res crops as well if this has any value for you guys.

paperboyo:
Thanks, @adrielulanovsky. Will the link be the same or will it contain anything extra, so that it doesn’t launch a pic in a new tab as now, but triggers the download. In other words, when users will right-click on the link, when this is ON, will they get exactly the same URL, eg. `https://media.guim.co.uk/892f053aa9d2b342e96047562878dfa8cb62ac2b/0_0_8400_5649/8400.jpg`?

>     2. So it will look as only one option for each crop.

I know our users are almost always look for smaller than full-size optimised crops (2000px seems to be most popular). If we were to ever have this functionality ON, we would need to see those other sizes. 

cc @blishen @akash1810 and all (because we talked about actually having downloads (usage) reporting for off-platform used crops)

adrielulanovsky:
> Thanks, @adrielulanovsky. Will the link be the same or will it contain anything extra, so that it doesn’t launch a pic in a new tab as now, but triggers the download? In other words, when users will right-click on the link, when this is ON, will they get exactly the same URL, eg. `https://media.guim.co.uk/892f053aa9d2b342e96047562878dfa8cb62ac2b/0_0_8400_5649/8400.jpg`?
> 
> > ```
> > 2. So it will look as only one option for each crop.
> > ```
> 
> I know our users almost always look for smaller than full-size optimised crops (2000px seems to be most popular). If we were to ever have this functionality ON, we would need to see those other sizes.
> 
> cc @blishen @akash1810 and all (because we talked about actually having downloads (usage) reporting for off-platform used crops)

It will behave the same way as the Downoad button. That is, the link it will point to will be the endpoint in media-api for downloading crops ( https://api.images.bbc.co.uk/images/:imageId/export/:exportId/download for the moment, although I'll probably change that to also have the dimensions or an identifier for downloading the other sizes).

adrielulanovsky:
Changing this to a Draft PR to give the endpoint the capability of downloading smaller sized crops as well.

paperboyo:
> the link it will point to will be the endpoint in media-api for downloading crops

This will make it unusable for those (few!) of our users who need URLs (and of varying size) instead of downloads. So it looks like we will have that switched off. We only contemplated switching it on, in case we would want to count downloads as usages (the other thing we have switched off). This would allow us to track pictures that are used in this way (link copied and pic used in a bespoke interactive or design piece that escapes the normal usage reporting from our CMS).

But if it’s not possible to keep the download-crop-as-usage **and** allow for copying links like now, then we may want to skip this feature too. Which should make it easier to review, too ;-)

adrielulanovsky:
> Changing this to a Draft PR to give the endpoint the capability of downloading smaller sized crops as well.

I did this anyway, as it might be a useful thing for BBC too.

twrichards (commented):
Logic looks sound, however I get 
image
when trying to actually download the crops. 

Also more minor, user has to refresh to see the usage? (perhaps that's expected, @paperboyo may have an opinion on whether this needs addressing).

Lastly, is there any documentation that could be created/updated to detail the new config option `canDownloadCrop` and perhaps the existing but pertinent `image.record.download`?

paperboyo:
> Also more minor, user has to refresh to see the usage?

This may be how it works currently. Most obviously it would be preferable for UI to **always** show any updates live. No idea how hard that would be, though…

adrielulanovsky:
> Logic looks sound, however I get
> image
> when trying to actually download the crops.
> 
> Also more minor, user has to refresh to see the usage? (perhaps that's expected, @paperboyo may have an opinion on whether this needs addressing).
> 
> Lastly, is there any documentation that could be created/updated to detail the new config option `canDownloadCrop` and perhaps the existing but pertinent `image.record.download`?

Hm, that's weird... are you certain that the imageId, exportId and width chosen are from an existing crop?

I think the refresh is currently needed to see the usage.

I can create documentation for those configs. Are you aware of an existing document for configs?

paperboyo:
> Are you aware of an existing document for configs?

I think we’ve neglected that part, definitely since we started to add more BBC-specific configuration options. In a way it became harder for a prospective new joiner to set up the system, because it’s not clear what and how the functionality might differ… I think we have a job to fix that. Some relevant bits are (or – should be) here:
https://github.com/guardian/grid/blob/main/docs/01-setup/03-configuring-setup.md
https://github.com/guardian/grid/tree/main/docs/07-extending

twrichards:
Ok so we've re-tested with the `publishing.image.bucket` conf option added to `media-api.conf` and that works nicely now. 

Interestingly the...
image
... always seems to download a 500px wide image, which seems odd. The links down in the bottom right let you download the crop at other resolutions though.

adrielulanovsky:
> Ok so we've re-tested with the `publishing.image.bucket` conf option added to `media-api.conf` and that works nicely now.
> 
> Interestingly the...
> image
> ... always seems to download the second lowest resolution image (e.g. 500px wide), which seems odd. The links down in the bottom right let you download the crop at other resolutions though.

I'll try to fix this, I think I am selecting the first option in the assets array. When I tested I saw that the first option was the highest resolution, but maybe the order is not consistent.

paperboyo:
> When I tested I saw that the first option was the highest resolution, but maybe the order is not consistent.

I don’t know what the order is and if it’s consistent. But the logic would be to take the one that is the same size as `master` one, but **not** the actual `master` (`master`s are horrendously large, they are to be send to image resizer and cannot be used-as is: weigh megabytes… while other assets are – poorly – optimised for immediate use on web)

twrichards:
> > Ok so we've re-tested with the `publishing.image.bucket` conf option added to `media-api.conf` and that works nicely now.
> > Interestingly the...
> > image
> > ... always seems to download the second lowest resolution image (e.g. 500px wide), which seems odd. The links down in the bottom right let you download the crop at other resolutions though.
> 
> I'll try to fix this, I think I am selecting the first option in the assets array. When I tested I saw that the first option was the highest resolution, but maybe the order is not consistent.

I know there's some client-side logic for detecting the smallest and largest assets btw, not sure if the same code exists somewhere serverside or whether you want to keep the serverside generic and pass the size you want as query param (which I think you already did).

adrielulanovsky:
> Ok so we've re-tested with the `publishing.image.bucket` conf option added to `media-api.conf` and that works nicely now.
> 
> Interestingly the...
> image
> ... always seems to download the second lowest resolution image (e.g. 500px wide), which seems odd. The links down in the bottom right let you download the crop at other resolutions though.

@twrichards This should be fixed now.

kahuna/public/js/image/controller.js
```diff
@@ -208,7 +208,18 @@ image.controller('ImageCtrl', [
...
-    mediaCropper.getCropsFor(image).then(crops => {
+    mediaCropper.getCropsFor(image).then(cropsResource => {
+      let crops = cropsResource.data;
+      if ($window._clientConfig.canDownloadCrop) {
+        crops.forEach((crop) => {
+          crop.assets.forEach((asset) =>
+            asset.downloadLink = cropsResource.links.find(link => link.rel.includes(`crop-download-${crop.id}-${asset.dimensions.width}`))?.href
+          );
+          //set the download link of the crop to be the largest asset
+          let largestAsset = crop.assets.find(asset => asset.dimensions.width == crop.master.dimensions.width);
```
twrichards, line 219: this is the logic for finding the 'extreme' assets (i.e. smallest and largest) that I'd hinted at in my previous comment... https://github.com/guardian/grid/blob/ca7fa005e870fe9ddf0a18aabc80e511bedd891a/kahuna/public/js/main.js#L316-L328
...then example usage (of smallest)...
https://github.com/guardian/grid/blob/e669dfc4b01aa4fdfa82893d535e4a86558319b3/kahuna/public/js/image/crop.html#L4 

Totally understand if they're not suitable here.

twrichards (approved):
I've just tested it since your most recent commit and all working nicely - logic looks good too (although I've commented about some existing client side logic for the largest asset - not a blocker though).
Great work 🙌  and appreciate the patience awaiting this review.
Will merge tomorrow.

kahuna/public/js/image/controller.js
adrielulanovsky: Yes, I saw this logic, however this function takes images as input and I needed to take crops instead, so I thought I would just rewrite it instead of trying to generalize this function only for this use case. Also getExtremeAssets is slightly more inefficient since it needs to sort out the assets array, whereas I only need to find the asset with the original crop width.

PR #3360, opened 2021-06-23
Feature/soft delete UI filter

abdelrahmansd:
## What does this change?
introduces a new "advanced search" filter for searching by soft-deleted, named "is:soft-deleted".

Given images in the Images Catalogue
When the user has set the search query filter `is:soft-deleted`
Then the user should be able to see all the soft-deleted images in the Images Catalogue

## Screenshots

Screenshot from 2021-06-24 10-20-38

## Tested?
- [x] locally by committer


paperboyo:
Tiny: since we’ve decided soft-deletion replaced deletion, at least as far as UI/users are concerned, should this simply be `+is:deleted`?

twrichards:
replaced by... 
- https://github.com/guardian/grid/pull/3376

PR #3362, opened 2021-06-28
Refactor ElasticSearch config to support current and migration aliases

tjsilver:
Co-authored-by: @twrichards 

## What does this change?

Removed the idea of 'read' and 'write' aliases  and introduced the idea of 'current' and 'migration' aliases (NB: 'current' is not equivalent to 'read', nor is 'migration' equivalent to 'write' - this is a new concept altogether). This work was done in preparation for the ongoing reingestion project.

Previously, the ElasticSearch config was repeated in the `thrall` and `media-api` config files, which could have led to inconsistencies. We have now hoisted the `es` and `es6` config blocks up to `common.conf` and the properties are now read in through a new class `CommonConfigWithElastic` which extends `CommonConfig`, and now `ThrallConfig` and `MediaApiConfig` extend this new class. 

### Updating config

To provide breathing room while teams/individuals update their config, we have added fallbacks for the `current` alias to the `read`/`write` alias config properties. This will be removed once everyone has updated their config. 

For local/dev machines, simply:
1. Run `/dev/script/setup.sh` (being mindful this may overwrite some of your local config).
2. Run `/dev/script/start.sh` (and best to run `/dev/script/the_pingler.sh` as well to exercise all the services)
3. Check in Cerebro (http://localhost:9090/#!/aliases?host=http:%2F%2Felasticsearch:9200) for the presence of the new "Images_Current" alias, optionally you can now remove "readAlias" and "writeAlias".

The below instructions should be repeated for each real enviroment.

1. Via Cerebro (or whatever ES management tool you use), add an `Images_Current` alias to the main/only index (where the `readAlias` is pointing)
2. Move the `es` and `es6` config blocks from both `thrall.conf` and `media-api.conf` up to `common.conf` (removing any duplication)
3. Remove `es.index.aliases.read` and `es.index.aliases.write`
4. Add `es.index.aliases.current` with a value of "Images_Current"
5. Add `es.index.aliases.migration` with a value of "Images_Migration"
6. Deploy Grid and test 
7. (Optional) Remove old `readAlias` and `writeAlias` from ES, then redeploy and retest.

Once this is done for all environments, please contact @guardian/digital-cms so we can remove fallback code.

Further unblocking the reingestion project by having meaningfully-named aliases.

## Who should look at this?
@guardian/digital-cms 

## Tested? Documented?
- [ ] locally by committer


dev/script/generate-config/README.md
```diff
@@ -3,7 +3,7 @@
 Generate the configuration files for the various Grid services.
 
 ## Requirements
-- Node 8
+- Node 14
```
jonathonherbert, line 6: 😁

jonathonherbert (approved):
Looks sensible to me, and steps for migration seem clear. Has this been tested in e.g. TEST? Just 'cos I haven't exercised them – lmk if you'd like me to try them locally or in a cloud environment.

PR #3368, opened 2021-06-30
filter scriptsToLoad based on new property `shouldLoadWhenIFramed`

twrichards:
## What does this change?
I noticed in the GIF of https://github.com/guardian/prosemirror-elements/pull/45 that pinboard is loading when the Grid is iframed in Composer, this will be confusing for users.

Since the pinboard script is loaded based on configuration, I've added a new optional property `shouldLoadWhenIFramed` on items in `scriptsToLoad` config array. 

Whether UI is loaded in an iframe is determined by [`Sec-Fetch-Dest`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-Fetch-Dest) header being present and set to `iframe`, as can be seen from network tab when Grid is used in composer...
image

No double vision, i.e. no confused users seeing pinboard in grid, when they already see it in composer thats iframing grid.

## Who should look at this?
@guardian/digital-cms 

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/app/controllers/KahunaController.scala
+++ b/kahuna/app/controllers/KahunaController.scala
@@ -21,6 +21,12 @@ class KahunaController(
 
     val maybeUser: Option[Authentication.Principal] = authentication.authenticationStatus(req).toOption
 
+    val isIFramed = req.headers.get("Sec-Fetch-Dest").contains("iframe")
+
+    val scriptsToLoad = config.scriptsToLoad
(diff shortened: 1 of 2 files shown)


twrichards:
now tested in `TEST` ✅

SHession (approved):
Code changes make sense, checked on the aforementioned PR. LGTM!

PR #3376, opened 2021-07-07
introduce is:deleted search filter

abdelrahmansd:
## What does this change?
introduces a new "advanced search" filter for searching by soft-deleted, named "is:deleted".

Given images in the Images Catalogue
When the user has set the search query filter is:deleted
Then the user should be able to see all the soft-deleted images in the Images Catalogue

## Screenshots

Screenshot from 2021-07-07 03-57-13
Screenshot from 2021-07-07 03-57-40

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/media-api/app/lib/elasticsearch/IsQueryFilter.scala
+++ b/media-api/app/lib/elasticsearch/IsQueryFilter.scala
@@ -13,6 +13,7 @@ sealed trait IsQueryFilter extends Query with ImageFields {
     case IsOwnedPhotograph(staffPhotographerOrg) => s"$staffPhotographerOrg-owned-photo"
     case IsOwnedIllustration(staffPhotographerOrg) => s"$staffPhotographerOrg-owned-illustration"
     case IsOwnedImage(staffPhotographerOrg) => s"$staffPhotographerOrg-owned"
+    case _: IsDeleted => "deleted"
     case _: IsUnderQuota => "under-quota"
   }
 }
@@ -26,6 +27,7 @@ object IsQueryFilter {
       case s if s == s"$organisation-owned-illustration" => Some(IsOwnedIllustration(organisation))
       case s if s == s"$organisation-owned" => Some(IsOwnedImage(organisation))
       case "under-quota" => Some(IsUnderQuota(overQuotaAgencies()))
+      case "deleted" => Some(IsDeleted(true))
       case _ => None
     }
   }
@@ -53,4 +55,10 @@ case class IsUnderQuota(overQuotaAgencies: List[Agency]) extends IsQueryFilter {
   override def query: Query = overQuotaAgencies.toNel
     .map(agency => filters.mustNot(filters.terms(usageRightsField("supplier"), agency.map(_.supplier))))
     .getOrElse(matchAllQuery)
+}
+
+case class IsDeleted(isDeleted: Boolean) extends IsQueryFilter {
+  override def query: Query = filters.or(
+    (filters.existsOrMissing("softDeletedMetadata", _))(isDeleted)
+  )
 }
--- a/media-api/app/lib/querysyntax/Parser.scala
+++ b/media-api/app/lib/querysyntax/Parser.scala
@@ -2,8 +2,14 @@ package lib.querysyntax
 
 object Parser {
 
-  def run(input: String): List[Condition] =
-    normalise(parse(input))
+  def run(input: String): List[Condition] = {
+    normalise(
+      parse(
+        if(input.contains("is:deleted")) input
+        else input.concat(" -is:deleted").trim
+      )
+    )
+  }
 
   def parse(input: String): List[Condition] =
     new QuerySyntax(input.trim).Query.run().map(_.toList) getOrElse List()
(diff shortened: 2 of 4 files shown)


twrichards:
great work on this @abdelrahmansd

I've looked it in our TEST environment, and whilst the `is:deleted` filter works just as expected, and although the 'soft deleted' items no longer show when you're on the homepage (i.e. no search criteria) I see that when I search for anything else I see the soft deleted images are visible... looking at the code, I see the `is:not-deleted` is being passed as the default query, which explains them not showing on the homepage (as desired) but I think as soon as any search terms are added they're then visible again. I think we need to figure out a deeper way of filtering out deleted images on ALL queries, EXCEPT those which include `is:deleted` - I'm not sure how complicated this will be (I will try to find something useful)

abdelrahmansd:
> but I think as soon as any search terms are added they're then visible again.

Thanks @twrichards for review , 
yes `is:not-deleted` filter should be added in all cases except if `is:deleted` filter is explicitly added
I will handle it in another commit now

paperboyo:
> is:not-deleted

I wouldn’t know if this is any easier on the backed, but just wanted to comment that any filter can be negated, so no need for an explicit negation: `-is:deleted`, again, unless that way it’s easier to express it in code (which I wouldn’t know).

abdelrahmansd:
yes , it will be done from the backend

media-api/app/lib/querysyntax/Parser.scala
```diff
@@ -2,8 +2,10 @@ package lib.querysyntax
 
 object Parser {
 
-  def run(input: String): List[Condition] =
-    normalise(parse(input))
+  def run(input: String): List[Condition] = {
+    val inputString = if(!input.contains("is:deleted")) input.concat(" is:not-deleted").trim else input
+    normalise(parse(inputString))
```
twrichards, line 7: I think could be expressed more clearly as...
```suggestion
    normalise(
      parse(
        if(input.contains("is:deleted")) input
        else input.concat(" is:not-deleted").trim 
      )
    )
```

twrichards: Also, did you try @paperboyo 's suggestion of `-is:deleted` (where the `-` makes it a negation of `is:deleted`) (see https://github.com/guardian/grid/pull/3376#issuecomment-875768691)? 

_Happy to stick with `not-deleted` btw, as its slightly more explicit/readable unless there is no precedent amongst other filters?_

twrichards (commented):
I'll give this a spin in our test environment shortly, but in the meantime,

media-api/app/lib/querysyntax/Parser.scala
abdelrahmansd: @twrichards yes i tried `-is: deleted`, i will update the PR with it now, it makes sense, rather than introducing a negation filter

AWare: If the default is to hide deleted images, then this js change shouldn't be needed and we should just have an "is not deleted" filter running by default that `is:deleted` reverses. 

I think it may also make sense to have a filter which shows both deleted and not images.

paperboyo: > I think it may also make sense to have a filter which shows both deleted and not images.

Was thinking about that too. But the problem is: Grid has no `OR` searches **at all** (which is bad, very bad).

AWare: If the default state is to search for images which are not deleted, then this should not be a seperate filter. We should use the presence of the `deleted` flag to change the `deleted` filter from showing not-deleted images to showing deleted images. I'd also suggest a filter which shows both deleted and not-deleted images. I believe one of the free to use filters already works like this.

abdelrahmansd: I think it is a similar approach here
https://github.com/guardian/grid/blob/cae0b648ef0804cbfc985cc52d9fd5d86d4adeeb/media-api/app/lib/querysyntax/Parser.scala#L8

media-api/app/lib/elasticsearch/IsQueryFilter.scala
```diff
@@ -26,6 +27,8 @@ object IsQueryFilter {
       case s if s == s"$organisation-owned-illustration" => Some(IsOwnedIllustration(organisation))
       case s if s == s"$organisation-owned" => Some(IsOwnedImage(organisation))
       case "under-quota" => Some(IsUnderQuota(overQuotaAgencies()))
+      case "deleted" => Some(IsDeleted(true))
+      case "not-deleted" => Some(IsDeleted(false))
```
twrichards, line 31: can this be removed following https://github.com/guardian/grid/pull/3376/commits/cae0b648ef0804cbfc985cc52d9fd5d86d4adeeb

twrichards (approved):
I've just done a last test of this in our `TEST` environment... all working as expected.

Nice work 🙌  and thanks again for your patience.

I will merge on Monday (if you could just take a look at the comment and ideally squash your commits a bit to avoid commits like 'fix tests')

media-api/app/lib/elasticsearch/IsQueryFilter.scala
abdelrahmansd: @twrichards `not-deleted` removed and commits are been squashed

PR #3377, opened 2021-07-07
Recognise source:Shutterstock as Rex

paperboyo:
## What does this change?
Categorises images with `source:Shutterstock` as Rex.

Less manual categorisation.

## Tested? Documented?
- [ ] locally by committer


twrichards (approved):
assuming you've tested it @paperboyo

code looks good (and to answer your point in email thread, yeah it looks like you've understood the `_` correctly here*)

* this is a pattern match, and the `_ ` means you don't care about that part of the match, in this case the `image.metadata.source`, instead you're matching against `image.metadata.credit` (note also this will only work if it hasn't matched any of the cases above).
Worth saying the `_ ` means a totally different thing inside say a `.map` where it is essentially a short hand for the thing in question.

paperboyo:
> you don't care about that part of the match, in this case the `image.metadata.source`, instead you're matching against `image.metadata.credit`

Thanks for the info and for spelling this out, because, inexplicably, I was matching on exactly the right thing! (should be `source`, fixed now). Tested, ready for merging tomorra.

PR #3384, opened 2021-07-09
Force vertical scroll bar

adrielulanovsky:
## What does this change?
This change forces the browser to show the vertical scroll bar, to avoid jumps in the page caused by the scroll bar appearing/disappearing when hovering over elements in the image viewer.

See https://css-tricks.com/eliminate-jumps-in-horizontal-centering-by-forcing-a-scroll-bar/ for more info

The vertical scroll bar is always visible.

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/kahuna/public/stylesheets/main.css
+++ b/kahuna/public/stylesheets/main.css
@@ -68,6 +68,7 @@ z-index
 }
 
 html {
+    overflow-y: scroll;
     font: 62.5%/1.5 "Guardian Agate Sans Web", Georgia;
 }
 


paperboyo:
Hi. I don’t think this is a good solution to, slightly broken, indeed, UI. It’s showing the scrollbar at all times, even when there should be none:

image
[just one search result]

image
[multiple scrollbars bonanza]



>  jumps in the page caused by the scroll bar appearing/disappearing when hovering over elements in the image viewer

Could you paste a GIF of the behaviour? I only know that scrollbars are broken on images with the Usage Rights messaging that seems to not be included in the vertical page estimation:

image


This will likely need a more involved solution. Maybe some pointers are available in [this PR](https://github.com/guardian/grid/pull/2949) (only from memory, may not be true)?

honorcb:
This was a ticket I raised on our Jira, because to the user,  it stops you being able to pick the pencil to edit the text , when this issue happens.  To the user it appear that the image has the shivers, as it flickers when you move to edit the text.  For our archivist team users this could be a major problem.  I know that this does not happen if I am working in Firefox,  but Chromes it the BBC default browser.

paperboyo:
I will try to reproduce (I think I’ve seen it), but am unable now. @honorcb, could you possibly record the screen or ask the user – maybe it happens with specific image/screen resolution/browser size?

Having a recording/reproduction steps will help fixing it.

paperboyo:
Maybe, it can be worked around on a Mac with a System Preference setting:

image

Still, it should be possible to fix. I hope.

AWare:
Is the jitter happening in the info panel?

honorcb:
The Jitter appears across the entire screen, but it when you try to access a pencil in the info panel,   So I guess this might be an issue that only bother people who can edit any metadata not just their own images.   it seem quite random when it occurs,  and today I can only find in on BBC uploaded images today but it does occur with agency images,  I was trying to find one of those as you should have that image.

honorcb:
I am very bad at recording screens, 

https://user-images.githubusercontent.com/29702265/125273141-85da9080-e304-11eb-9f20-c3d5084682bd.mp4

paperboyo:
Thanks for the recording! Can you try changing System Preferences>General>Show scrollbars to see if it helps?

I’m sure I have seen it, but cannot repro despite messing around System Preferences and Chrome window size. I can only see the usual bugs:
1. unnecessary horizontal scrollbar:
    horiz
2. Usage Rights banners causing the whole page to move up and down (second vertical scrollbar)
    double

adrielulanovsky:
I've had trouble reproducing this myself. I think it happens either when the browser is very close to adding a scrollbar, and the hovering adds some margins somewhere that make the scrollbar appear, or viceversa (I've seen it in images that have a scrollbar and the hovering makes the scrollbar disappear). I'm referring to the second scrollbar, not the scrollbar in the Metadata section

adrielulanovsky:
https://user-images.githubusercontent.com/20479781/125282049-553b2c80-e2ed-11eb-8e6f-87c31ca6c6a0.mp4

adrielulanovsky:
Would it be ok to force the vertical scrollbar on only on the /images page?

paperboyo:
> Would it be ok to force the vertical scrollbar ON only on the /images page?

Better than everywhere, but still a bit crude, I guess (some images do not and should not have it). I would prefer a surgical solution that will fix the underlying problem properly (strangely, last screen recording suggests it’s the **horizontal** scrollbar that is at fault?!). But I wouldn’t know how much hate engineers have for diving into AngularJS ;-)

twrichards:
> > Would it be ok to force the vertical scrollbar ON only on the /images page?
> 
> Better than everywhere, but still a bit crude, I guess (some images do not and should not have it). I would prefer a surgical solution that will fix the underlying problem properly (strangely, last screen recording suggests it’s the **horizontal** scrollbar that is at fault?!). But I wouldn’t know how much hate engineers have for diving into AngularJS ;-)

Has any one tried permanently disabling the horizontal scrollbar as an alternative? I can't see anything actually being obscured.

AWare:
Maybe we could just tweak the sizes until it stops doing this on hover?

adrielulanovsky:
> > > Would it be ok to force the vertical scrollbar ON only on the /images page?
> > 
> > 
> > Better than everywhere, but still a bit crude, I guess (some images do not and should not have it). I would prefer a surgical solution that will fix the underlying problem properly (strangely, last screen recording suggests it’s the **horizontal** scrollbar that is at fault?!). But I wouldn’t know how much hate engineers have for diving into AngularJS ;-)
> 
> Has any one tried permanently disabling the horizontal scrollbar as an alternative? I can't see anything actually being obscured.

Disabling the horizontal scrollbar apparently fixed it for the image where we can reproduce this, although I'm not sure if it's just that it changed the height of what was showing and now there's no need to show a vertical scrollbar in either case... The problem for me is still in the vertical scrollbar appearing/disappearing, but it has been really difficult to reproduce this consistently (in fact, the image that I sent the video of is not doing that behaviour now, with no apparent changes)

AWare:
I'm a little confused as the div where this is happening has a class `image-details--scroll` which seems to already be enforcing the vertical scroll. So it would seem that this is actually a horizontal scroll thing.

adrielulanovsky:
> I'm a little confused as the div where this is happening has a class `image-details--scroll` which seems to already be enforcing the vertical scroll. So it would seem that this is actually a horizontal scroll thing.

That's not the div where the vertical scrollbar is appearing/disappearing. If you take a close look at the video, the image-details panel always has the vertical scrollbar, the one that's appearing/disappearing is the outer vertical scrollbar

AWare:
I see, I had the no rights warning- which was causing a scrollbar on its own.

honorcb:
that is the issue,  sometimes in Chrome  the same image , and same metadata generate a horizontal scroll bar and associated issue and sometime it does not,  and it does not appear to matter whether you are using full screen,  or what percentage Zoom,  which make it hard, to generate test cases

paperboyo:
> image-details panel always has the vertical scrollbar

May depend on this System Preferences Mac setting, mind you.

> the no rights warning- which was causing a scrollbar on its own

Yep, that was always broken.

AWare:
I think it might well be the add leases button.

adrielulanovsky:
> I think it might well be the add leases button.
> 
> It's the tooltip on the add leases button which is causing an overflow and giving us the horizontal scroll. I'm not sure if fixing it would solve the whole issue, but I think it's worth testing with the tooltip disabled. I've got some meetings now but will give it another go after.

I tried removing the entire Leases section, it's still doing the behaviour:

https://user-images.githubusercontent.com/20479781/125297421-99362d80-e2fd-11eb-9095-21190e4ae934.mp4

honorcb:
Might it just be something to do with a max number of characters to rendering in the metadata box?  could  never seem get an issue with images with short description and very few tags and no title.    I know that adding leases sometimes gave me issue but not always,

adrielulanovsky:
Since The Guardian hasn't been able to reproduce this, it might be a BBC-only bug. I'm closing this PR and will fix it via custom css (forcing the browser to hide the horizontal scrollbar seems like a better approach and apparently fixes the issue)

AWare:
I think hiding the horizontal scrollbar is the better approach and don't object to it being global. Thanks for your patience while we failed to reproduce it.

PR #3385, opened 2021-07-09
Tag S3 object on add to library

adrielulanovsky:
## What does this change?
This sets an S3 tag on the image object stored in the main image bucket when the image is added to library.

See https://github.com/guardian/grid/discussions/3382 for discussion on this feature

When a user clicks on Add to Library
Then an S3 tag is added on that image's object stored in the main image bucket

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [ ] locally by committer


AWare:
Why is this wanted on the S3 object and not in edits or somewhere similar?

adrielulanovsky:
Closing this PR, since BBC will implement this feature via lambdas instead.

PR #3392, opened 2021-07-13
Deprecation warnings for the index healthcheck code

AWare:
Add red ink to the base of ElasticSearchClient to prevent use of hardcoded index/alias values until the index handling logic is superceded by the migration code.

## What does this change?

This adds warnings to all the static references of the images index so we don't use them- and then can delete it all when we get to the index handling in the migration work. 

That we don't use this. 

## Tested? Documented?
- [x] noop
- [ ] locally by committer


twrichards:
This seem such as shame to close given...
image

PR #3393, opened 2021-07-13
Remove guardian-specific prefixes

adrielulanovsky:
## What does this change?
Many services are using Guardian-specific prefixes for uris instead of relying in config, which causes issues for non-Guardian organisations. This replaces most of those usages to rely on config instead.

URI prefixes depend on config and are not hardcoded.

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/image-loader/app/ImageLoaderComponents.scala
+++ b/image-loader/app/ImageLoaderComponents.scala
@@ -38,7 +38,7 @@ class ImageLoaderComponents(context: Context) extends GridComponents(context, ne
     case (false, _) => None
   }
 
-  val services = new Services(config.domainRoot, ServiceHosts.guardianPrefixes, Set.empty)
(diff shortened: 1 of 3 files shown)


AWare:
There's no change to the config loader, so I presume this is a drop in change?

adrielulanovsky:
> There's no change to the config loader, so I presume this is a drop in change?

The config serviceHosts already exists in all services, so there's no need to change the config loader. However Guardian might need to add some configs for it to work ("hosts.<service>Prefix" configs), although those configs already have defaults that match the hardcoded guardianPrefixes, so maybe no changes from your side are needed

AWare (approved):

AWare:
@paperboyo took this for a spin on TEST and is happy, and by that comment and the code it looks pretty safe to go

PR #3398, opened 2021-07-15
update sentry because snyk is sad 🐶

AWare:
supercedes https://github.com/guardian/grid/pull/3391 and https://github.com/guardian/grid/pull/3390


twrichards (approved):
deployed to `TEST`... nothing seems broken, but not been able to do anything which causes an error to be posted off to sentry so hard to be 100% sure its all good - but approving nonetheless

PR #3394, opened 2021-07-14
Parse Kinesis records as ExternalThrallMessages if possible

andrew-nowak:
Co-authored-by: Alex Ware <alex.ware@guardian.co.uk>

## What does this change?

Incoming messages are currently parsed as `UpdateMessage`s - this PR changes to try to parse them as `ExternalThrallMessage`s if possible, and falling back to parsing as `UpdateMessage`s if that fails, then extracting the `ExternalThrallMessage` from that `UpdateMessage`.

## How can success be measured?

All incoming messages are successfully parsed and consumed as `ExternalThrallMessage`s, which is a more flexible type and makes it easier for us to add new message types in the future.

## Tested? Documented?
- [x] locally by committer
- [ ] locally by Guardian reviewer
- [ ] on the Guardian's TEST environment
- [ ] relevant documentation added or amended (if needed)

PR #3400, opened 2021-07-16
Support sending `ExternalThrallMessage` over Kinesis

twrichards:
## What does this change?
Follows on from https://github.com/guardian/grid/pull/3394 (which added parsing of `ExternalThrallMessage` directly from Kinesis - i.e. avoiding translation from `UpdateMessage`) - this PR refactors the sending to also support sending `ExternalThrallMessage` directly over Kinesis.

Simplicity.

## Who should look at this?
@guardian/digital-cms 

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/Kinesis.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/Kinesis.scala
@@ -2,16 +2,14 @@ package com.gu.mediaservice.lib.aws
 
 import java.nio.ByteBuffer
 import java.util.UUID
-
 import com.amazonaws.services.kinesis.model.PutRecordRequest
 import com.amazonaws.services.kinesis.{AmazonKinesis, AmazonKinesisClientBuilder}
-import com.gu.mediaservice.lib.config.CommonConfig
 import com.gu.mediaservice.lib.json.JsonByteArrayUtil
 import com.gu.mediaservice.model.usage.UsageNotice
 import net.logstash.logback.marker.{LogstashMarker, Markers}
 import play.api.libs.json.{JodaWrites, Json, Writes}
 import com.amazonaws.auth.AWSCredentialsProvider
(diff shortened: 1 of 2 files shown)


common-lib/src/main/scala/com/gu/mediaservice/model/ExternalThrallMessage.scala
```diff
@@ -47,7 +47,7 @@ object ExternalThrallMessage{
 
   implicit val createMigrationIndexMessage = Json.format[CreateMigrationIndexMessage]
 
-  implicit val writes = Json.writes[ExternalThrallMessage]
+  val writes = Json.writes[ExternalThrallMessage] // intentionally not implicit, so individual formats will be used
```
AWare, line 50: what's the intention behind this?

twrichards: there's an ambiguous implicit resolution when trying to publish say `CreateMigrationIndexMessage` where the compiler can't decide between `writes` and `createMigrationIndexMessage` further up in the `object ExternalThrallMessage`  (the compiler error doesn't actually affect this PR tbf but best to get it right up front).

AWare: is this right though? this gives me a bit of a feeling it's covering up a different implicit mistake (that i made)

twrichards: hmm I think it makes sense, and the overall writes is used explicitly further up the file in the `toJson` which is in-turn used in the logMarkers. let me take another look though, might be something else I can do to improve

twrichards: yep found an alternative, which leaves it as an implicit

AWare (approved):

PR #3401, opened 2021-07-16
handler for `CreateMigrationIndexMessage` to create new index & alias

AWare:
## What does this change?

This adds the logic to start a migration by fixing some of the unimplemented things in the ES base class and then wires that up to the message handler.

It also splits off the test:compile and update phases of sbt so teamcitys output is legible. 

## How can anything be measured?

objective truth is impossible

## Who should look at this?

you.

## Tested? Documented?
- [ ] locally by committer
- [ ] locally by Guardian reviewer
- [ ] on the Guardian's TEST environment
- [ ] relevant documentation added or amended (if needed)

PR #3403, opened 2021-07-19
wire up thrall 'start migration button' to send `CreateMigrationIndexMessage`

twrichards:
## What does this change?
Adds a new `POST /startMigration` endpoint to `thrall` which sends the `CreateMigrationIndexMessage` (via the UI priority [a.k.a. high priority] kinesis stream) and then polls for 10 seconds (server-side) to wait for the new migration index to be created, redirecting back to the index page if it does and ... 
image
... if it doesn't.

To invoke this new endpoint the button on the control panel has been changed to a form...
image

In order for this POST endpoint to be called from the browser (via the form POST) it required changes to `application.conf` to configure the `trustedProxies` to allow traffic from nginx locally and from the ELB when in AWS - as per https://www.playframework.com/documentation/2.8.x/HTTPServer#Configuring-trusted-proxies . Also worth noting that in order to work properly locally it required a fix to `dev-nginx` - see https://github.com/guardian/grid/pull/3403#issuecomment-887455841 and https://github.com/guardian/dev-nginx/pull/25

Combined with https://github.com/guardian/grid/pull/3401 we can start migrations all from the UI, making it easier for us to progress with the migration project (f.k.a. reingestion project).

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
Worth mentioning I've created [`start-migration-PLUS-aw-migration`](https://github.com/guardian/grid/compare/start-migration-PLUS-aw-migration) branch to combine this PR with https://github.com/guardian/grid/pull/3401 so we can see this work end-to-end.

- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/resources/application.conf
+++ b/common-lib/src/main/resources/application.conf
@@ -4,9 +4,12 @@ play {
   application.loader = "AppLoader"
   application.langs = "en"
 
-  http.session {
-    httpOnly = false
-    secure = true
+  http{
+    session {
+      httpOnly = false
+      secure = true
+    }
+    forwarded.trustedProxies=["0.0.0.0/0", "::/0"]
   }
 
   # Quick hack
--- a/kahuna/conf/routes
+++ b/kahuna/conf/routes
@@ -17,7 +17,6 @@ GET     /assets/*file                                 controllers.Assets.version
 GET     /management/healthcheck                       com.gu.mediaservice.lib.management.Management.healthCheck
 GET     /management/manifest                          com.gu.mediaservice.lib.management.Management.manifest
 GET     /management/whoAmI                            com.gu.mediaservice.lib.management.InnerServiceStatusCheckController.whoAmI(depth: Int)
-GET     /management/innerServiceStatusCheck           com.gu.mediaservice.lib.management.InnerServiceStatusCheckController.statusCheck(depth: Int)
 
 # Shoo robots away
 GET     /robots.txt                                   com.gu.mediaservice.lib.management.Management.disallowRobots
--- a/thrall/app/controllers/ThrallController.scala
+++ b/thrall/app/controllers/ThrallController.scala
@@ -1,18 +1,28 @@
 package controllers
 
+import akka.actor.ActorSystem
+import akka.stream.ActorMaterializer
+import akka.stream.scaladsl.{Sink, Source}
 import com.gu.mediaservice.lib.auth.{Authentication, BaseControllerWithLoginRedirects}
+import com.gu.mediaservice.lib.aws.ThrallMessageSender
 import com.gu.mediaservice.lib.config.Services
+import com.gu.mediaservice.lib.logging.GridLogging
+import com.gu.mediaservice.model.CreateMigrationIndexMessage
 import lib.elasticsearch.ElasticSearch
+import org.joda.time.{DateTime, DateTimeZone}
 import play.api.mvc.ControllerComponents
 
-import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.DurationInt
+import scala.concurrent.{Await, ExecutionContext}
 
 class ThrallController(
   es: ElasticSearch,
+  messageSender: ThrallMessageSender,
+  actorSystem: ActorSystem,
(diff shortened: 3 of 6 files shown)


twrichards:
This is working as expected in `TEST`, 10 seconds after hitting the `Start Migration` button we get...
https://github.com/guardian/grid/pull/3394
... and we see the placeholder show up in the logs... 
image

kahuna/conf/routes
```diff
@@ -17,7 +17,6 @@ GET     /assets/*file                                 controllers.Assets.version
 GET     /management/healthcheck                       com.gu.mediaservice.lib.management.Management.healthCheck
 GET     /management/manifest                          com.gu.mediaservice.lib.management.Management.manifest
 GET     /management/whoAmI                            com.gu.mediaservice.lib.management.InnerServiceStatusCheckController.whoAmI(depth: Int)
-GET     /management/innerServiceStatusCheck           com.gu.mediaservice.lib.management.InnerServiceStatusCheckController.statusCheck(depth: Int)
```
AWare, line 20: I presume this healthcheck runs directly on thrall now and is superfluous?

thrall/app/controllers/ThrallController.scala
```diff
@@ -28,4 +38,43 @@ class ThrallController(
...
+      InternalServerError(msgFailedToFetchIndex)
+    } map {
+      case Some(index) =>
+        BadRequest(s"There is already an index '${index}' for alias '${es.imagesMigrationAlias}', and thus a migration underway.")
+      case None =>
+        messageSender.publish(CreateMigrationIndexMessage(
+          migrationStart = DateTime.now(DateTimeZone.UTC),
+          gitHash = utils.buildinfo.BuildInfo.gitCommitId
+        ))
+        // poll until images migration alias is created, giving up after 10 seconds
```
AWare, line 56: I think this should 200 on sending the message and leave checking status for the user.

kahuna/conf/routes
twrichards: yes exactly (see commit message https://github.com/guardian/grid/pull/3403/commits/82950b6f04a2352a221f6ee8937c89ac812fb290)
(and just to clarify it doesn't run on a schedule or anything - to avoid unnecessary load - its just part of pingler for local dev and can now be hit directly since thrall is visible)

twrichards:
The issues with running this locally are resolved by https://github.com/guardian/dev-nginx/pull/25 (once released, people will need to run setup script again)

@AWare @tjsilver @andrew-nowak if any of you need to do browser form POSTs before the dev-ngnix PR is released (and upgraded on your machine) then simply add...
```proxy_set_header X-Forwarded-Proto $scheme;```
... to the thrall section of `/usr/local/etc/nginx/servers/media-service.conf` (ideally under the `proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;` line)

File: thrall/app/controllers/ThrallController.scala (hunk 2 above)
Comment (twrichards) [reply thrall/app/controllers/ThrallController.scala:56]: That would require the user to refresh the page, and potentially be tempted to click the Start Migration button again (even though server-side code guards against that) OR it would require client side polling - I've tried to avoid having any client-side code for this control panel - for simplicity (to avoid the inevitable package.json, react, nvmrc etc. etc. which all needs to be kept up to date)

**Review by andrew-nowak — APPROVED** [review]

File: common-lib/src/main/resources/application.conf [hunk 3]
```diff
@@ -4,9 +4,12 @@ play {
...
 
-  http.session {
-    httpOnly = false
-    secure = true
+  http{
+    session {
+      httpOnly = false
+      secure = true
+    }
+    forwarded.trustedProxies=["0.0.0.0/0", "::/0"]
```
andrew-nowak, line 12: On our branch, the new dev-nginx was enough to get POSTs working through the dashboard; we didn't need this extra config. It might not be needed?

twrichards: ahh good to know, this is still needed for it to work in AWS (behind the load balancer, see https://www.playframework.com/documentation/2.8.x/HTTPServer#Configuring-trusted-proxies)

PR #3409, opened 2021-07-21
Count multiple download usages per user and image

adrielulanovsky:
## What does this change?
Currently, the grouping id for a download usage is generated by hashing downloadedBy and mediaId, which causes that 2 usages of the same image by the same user are counted as 1, since the usage is overwritten. With this change, 2 downloads of the same image by the same user will count as 2 different usages.

When a user downloads the same image n times, there are n usages listed in the Usages tab by that user.

## Screenshots
image

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [x] locally by committer


AWare (approved):
just going to run it on test before I merge

adrielulanovsky:
> just going to run it on test before I merge

Take into account that in order for downloads to be recorded as usages, you need to set image.record.download=true in media-api (don't know if The Guardian has this config set or not)

AWare:
teamcity build failed because I cancelled a dupe.

PR #3416, opened 2021-07-28
Refine DeleteCrops permission

adrielulanovsky:
## What does this change?
Currently DeleteCrops permission is used to determine if the user can delete crops OR usages. This renames DeleteCrops permission to DeleteCropsOrUsages to make its reach clearer. It also implements permissions on usage DELETE endpoint.

A user without DeleteCropsOrUsages permission cannot delete crops or usages.
A user with DeleteCropsOrUsages permission can delete crops and usages.

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/cropper/app/controllers/CropperController.scala
+++ b/cropper/app/controllers/CropperController.scala
@@ -6,7 +6,7 @@ import _root_.play.api.mvc.{BaseController, ControllerComponents}
 import com.gu.mediaservice.lib.argo.ArgoHelpers
 import com.gu.mediaservice.lib.argo.model.Link
 import com.gu.mediaservice.lib.auth.Authentication.Principal
-import com.gu.mediaservice.lib.auth.Permissions.{DeleteCrops, PrincipalFilter}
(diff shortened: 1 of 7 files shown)


cropper/app/controllers/CropperController.scala
```diff
@@ -79,7 +81,7 @@ class CropperController(auth: Authentication, crops: Crops, store: CropStore, no
     }
   }
 
-  private val canDeleteCrops: PrincipalFilter = authorisation.hasPermissionTo(DeleteCrops)
+  private val canDeleteCrops: PrincipalFilter = authorisation.hasPermissionTo(DeleteCropsOrUsages)
```
AWare, line 84: does this also want to say `OrUsages`?

AWare (approved):
looks good, will take it for a spin before I merge

cropper/app/controllers/CropperController.scala
adrielulanovsky: Didn't see the need here since it's a value in Cropper service and only used for deleting crops. But could rename as well if you think it's appropriate.

PR #3418, opened 2021-07-30
increase api-poll interval

abdelrahmansd:
## What does this change?

increase API poll interval to decrease timeout errors on poor internet connection
## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/util/async.js
+++ b/kahuna/public/js/util/async.js
@@ -57,7 +57,7 @@ async.factory("apiPoll", [
   "$q",
   ($q) => {
     const wait = () => new Promise(resolve => {
-      setTimeout(() => resolve(), 500);
+      setTimeout(() => resolve(), 1500);
     });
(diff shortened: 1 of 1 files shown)


kahuna/public/js/util/async.js
```diff
@@ -57,7 +57,7 @@ async.factory("apiPoll", [
   "$q",
   ($q) => {
     const wait = () => new Promise(resolve => {
-      setTimeout(() => resolve(), 500);
+      setTimeout(() => resolve(), 1500);
```
twrichards, line 60: this is quite a big increase (x3), how about...
```suggestion
      setTimeout(() => resolve(), 1000);
```
combined with a few more retries (maybe increase 100 to 150 on lines 72 & 73 below - that way it will wait for just as long total as your proposed 100x1500 millis)?

twrichards (approved):
nice! I like this back-off approach 👏

PR #3422, opened 2021-08-03
Add a count of documents to the thrall dashboard

andrew-nowak:
## What does this change?
Adds a count of documents to the thrall dashboard

Open thrall dashboard, see how many images are in each index.

## Screenshots

image

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchClient.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchClient.scala
@@ -65,12 +65,11 @@ trait ElasticSearchClient extends ElasticSearchExecutions with GridLogging {
     executeAndLog(request, "Healthcheck").map { _ => true}.recover { case _ => false}
   }
 
-
-  def countImages(): Future[ElasticSearchImageCounts] = {
+  def countImages(indexName: String = "images"): Future[ElasticSearchImageCounts] = {
     implicit val logMarker = MarkerMap()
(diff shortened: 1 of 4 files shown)


thrall/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -37,6 +37,10 @@ class ElasticSearch(config: ElasticSearchConfig, metrics: Option[ThrallMetrics])
     executeAndLog(getAliases(Nil, Seq(alias)), s"Looking up index for alias '$alias'").map(_.result.mappings.keys.headOption)
   }
 
+  def countDocsInIndex(index: Index)(implicit ex: ExecutionContext, logMarker: LogMarker = MarkerMap()): Future[Long] = {
+    executeAndLog(count(index.toIndexes), s"Counting docs in index ${index.name}").map(_.result.count)
+  }
+
```
twrichards, line 43: do the existing stat counts not provide the right info?

twrichards (approved):
brilliant stuff 👏

thrall/app/lib/elasticsearch/ElasticSearch.scala
andrew-nowak: Thanks for pointing me in the right direction - I hadn't realised that a count function already existed! I've pushed a new commit that makes use of that instead. It looks like it'll have a slight performance impact (it makes three queries for each index for different types of counts, of which we only use one here) but I think that's acceptable for powering this dashboard and avoiding duplication of logic

twrichards (approved):

PR #3425, opened 2021-08-05
handle re-upload soft-deleted images

abdelrahmansd:
## What does this change?

- update the Delete button to call the _soft-delete_ endpoint
- update the upload page to check if an image exists and is soft-deleted and give the ability to undelete it for permissioned users
- implement /undelete endpoint

when re-uploading an image that is soft-deleted, an error message should appear says the Image already exists and has been soft deleted
and  permissioned users are able to undelete that image

## Screenshots

Unpermissioned user:
Screenshot from 2021-09-10 15-02-10

Permissioned user:
Screenshot from 2021-09-10 14-57-45

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/image-loader/app/lib/FailureResponse.scala
+++ b/image-loader/app/lib/FailureResponse.scala
@@ -54,6 +54,16 @@ object FailureResponse extends ArgoHelpers with Results {
     )
   }
 
+  def imageAlreadyExists(exception: Exception): Response = {
+    logger.info(s"Rejected request to load file: image file already exists and is soft deleted", exception)
+
+    Response(
+      BadRequest,
+      "image-already-exists",
+      s"Image already exists: image file already exists and is soft deleted."
+    )
+  }
+
   def internalError(exception: Throwable): Response = {
     logger.info(s"Unhandled exception", exception)
 
--- a/image-loader/app/lib/FailureResponse.scala
+++ b/image-loader/app/lib/FailureResponse.scala
@@ -54,16 +54,6 @@ object FailureResponse extends ArgoHelpers with Results {
     )
   }
 
-  def imageAlreadyExists(exception: Exception): Response = {
-    logger.info(s"Rejected request to load file: image file already exists and is soft deleted", exception)
-
-    Response(
-      BadRequest,
-      "image-already-exists",
-      s"Image already exists: image file already exists and is soft deleted."
-    )
-  }
-
   def internalError(exception: Throwable): Response = {
     logger.info(s"Unhandled exception", exception)
 
--- a/kahuna/public/js/components/gr-archiver-status/gr-archiver-status.js
+++ b/kahuna/public/js/components/gr-archiver-status/gr-archiver-status.js
@@ -23,7 +23,15 @@ archiver.controller('ArchiverCtrl',
     ctrl.archive = archive;
     ctrl.unarchive = unarchive;
     ctrl.archiving = false;
-    ctrl.canArchive = false;
+    ctrl.canUndelete = false;
+    ctrl.isDeleted = false;
+
+    ctrl.undelete = undelete;
+
+    mediaApi.getSession().then(session => {
+        if (ctrl.image.data.softDeletedMetadata !== undefined && session.user.permissions.canDelete) { ctrl.canUndelete = true; }
+        if (ctrl.image.data.softDeletedMetadata !== undefined) { ctrl.isDeleted = true; }
+    });
 
     mediaApi.canUserArchive().then(canArchive => {
         ctrl.canArchive = canArchive;
@@ -49,6 +57,13 @@ archiver.controller('ArchiverCtrl',
         return promise.
             catch(()  => $window.alert('Failed to save the changes, please try again.')).
             finally(() => ctrl.archiving = false);
+    }
+
+    function undelete() {
+        const imageId = ctrl.image.data.id;
+        mediaApi.undelete(imageId).then(
+            ctrl.canUndelete = ctrl.isDeleted = false
+        );
     }
 }]);
 
--- a/kahuna/public/js/components/gr-archiver-status/gr-archiver-status.js
+++ b/kahuna/public/js/components/gr-archiver-status/gr-archiver-status.js
@@ -29,7 +29,7 @@ archiver.controller('ArchiverCtrl',
     ctrl.undelete = undelete;
 
     mediaApi.getSession().then(session => {
-        if (ctrl.image.data.softDeletedMetadata !== undefined && session.user.permissions.canDelete || session.user.email === ctrl.image.data.uploadedBy) { ctrl.canUndelete = true; }
+        if (ctrl.image.data.softDeletedMetadata !== undefined && (session.user.permissions.canDelete || session.user.email === ctrl.image.data.uploadedBy)) { ctrl.canUndelete = true; }
         if (ctrl.image.data.softDeletedMetadata !== undefined) { ctrl.isDeleted = true; }
     });
 
--- a/kahuna/public/js/components/gr-archiver-status/gr-archiver-status.js
+++ b/kahuna/public/js/components/gr-archiver-status/gr-archiver-status.js
@@ -29,7 +29,7 @@ archiver.controller('ArchiverCtrl',
     ctrl.undelete = undelete;
 
     mediaApi.getSession().then(session => {
-        if (ctrl.image.data.softDeletedMetadata !== undefined && session.user.permissions.canDelete) { ctrl.canUndelete = true; }
+        if (ctrl.image.data.softDeletedMetadata !== undefined && session.user.permissions.canDelete || session.user.email === ctrl.image.data.uploadedBy) { ctrl.canUndelete = true; }
         if (ctrl.image.data.softDeletedMetadata !== undefined) { ctrl.isDeleted = true; }
     });
 
--- a/kahuna/public/js/components/gr-archiver/gr-archiver.js
+++ b/kahuna/public/js/components/gr-archiver/gr-archiver.js
@@ -40,6 +40,7 @@ module.controller('grArchiverCtrl', [
 
         ctrl.canUndelete = false;
         ctrl.isDeleted = false;
+        ctrl.undeleting = false;
 
         ctrl.undelete = undelete;
 
@@ -112,10 +113,16 @@ module.controller('grArchiverCtrl', [
                 toArray();
         }
         function undelete() {
+            ctrl.undeleting = true
             const imageId = ctrl.image.data.id;
-            mediaApi.undelete(imageId).then(
-                ctrl.canUndelete = ctrl.isDeleted = false
-            );
+            mediaApi.undelete(imageId)
+                .then(
+                  ctrl.canUndelete = ctrl.isDeleted = false
+                ).catch(() => {
+                     $window.alert('Failed to undelete image, please try again.');
+                }).finally(() => {
+                     ctrl.undeleting = false;
+                });
         }
     }
 ]);
--- a/kahuna/public/js/components/gr-archiver/gr-archiver.js
+++ b/kahuna/public/js/components/gr-archiver/gr-archiver.js
@@ -45,7 +45,7 @@ module.controller('grArchiverCtrl', [
         ctrl.undelete = undelete;
 
         mediaApi.getSession().then(session => {
-            if (ctrl.image && ctrl.image.data.softDeletedMetadata !== undefined && session.user.permissions.canDelete) { ctrl.canUndelete = true; }
+            if (ctrl.image && ctrl.image.data.softDeletedMetadata !== undefined && session.user.permissions.canDelete || session.user.email === ctrl.image.data.uploadedBy) { ctrl.canUndelete = true; }
             if (ctrl.image && ctrl.image.data.softDeletedMetadata !== undefined) { ctrl.isDeleted = true; }
         });
 
--- a/kahuna/public/js/components/gr-archiver/gr-archiver.js
+++ b/kahuna/public/js/components/gr-archiver/gr-archiver.js
@@ -45,7 +45,7 @@ module.controller('grArchiverCtrl', [
         ctrl.undelete = undelete;
 
         mediaApi.getSession().then(session => {
-            if (ctrl.image && ctrl.image.data.softDeletedMetadata !== undefined && session.user.permissions.canDelete || session.user.email === ctrl.image.data.uploadedBy) { ctrl.canUndelete = true; }
+            if (ctrl.image && ctrl.image.data.softDeletedMetadata !== undefined && (session.user.permissions.canDelete || session.user.email === ctrl.image.data.uploadedBy)) { ctrl.canUndelete = true; }
             if (ctrl.image && ctrl.image.data.softDeletedMetadata !== undefined) { ctrl.isDeleted = true; }
         });
 
--- a/kahuna/public/js/components/gr-archiver/gr-archiver.js
+++ b/kahuna/public/js/components/gr-archiver/gr-archiver.js
@@ -45,8 +45,8 @@ module.controller('grArchiverCtrl', [
         ctrl.undelete = undelete;
 
         mediaApi.getSession().then(session => {
-            if (ctrl.image.data.softDeletedMetadata !== undefined && session.user.permissions.canDelete) { ctrl.canUndelete = true; }
-            if (ctrl.image.data.softDeletedMetadata !== undefined) { ctrl.isDeleted = true; }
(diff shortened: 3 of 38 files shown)


paperboyo:
Just regarding user-facing warning, I think it would be great:
- for permissioned users to be able to click through to see this image (see how upload looks for images that already exist, thumbnail on an upload page contains a View image link)
- for the wording to not use _soft delete_ phrasing as I think we’ve agreed that it’s to just replace _delete_ all over Grid. So maybe something like “This image already exist in the deleted state” or something (English speaker, help me here!)

paperboyo:
>     * for permissioned users to be able to click through to see this image

Actually, dependent on UX for unpermissioned users, the link may well appear for all? Eg. if an umpermissioned user hits the deleted URL directly, they maybe should see a page saying something like: _This image has been deleted. You have no permissions to see it. Please contact the relevant team if you think you need to use it._ Or something to that effect?

image-loader/app/lib/FailureResponse.scala
```diff
@@ -54,6 +54,16 @@ object FailureResponse extends ArgoHelpers with Results {
     )
   }
 
+  def imageAlreadyExists(exception: Exception): Response = {
+    logger.info(s"Rejected request to load file: image file is already exists and soft deleted", exception)
```
adrielulanovsky, line 58: Should this read "... image file already exists and is soft deleted" instead?

image-loader/app/lib/FailureResponse.scala
```diff
@@ -54,6 +54,16 @@ object FailureResponse extends ArgoHelpers with Results {
...
+
+    Response(
+      BadRequest,
+      "image-already-exists",
+      s"Image already exists: image file already exists and soft deleted."
```
adrielulanovsky, line 63: Same here

image-loader/app/lib/FailureResponse.scala
paperboyo: `image file  already exists in a deleted state` ?

image-loader/app/lib/FailureResponse.scala
paperboyo: Same as above? `state is good for logs and API responses, but can be dropped in user-facing warnings, I guess?

image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -79,39 +79,51 @@ class ImageLoaderController(auth: Authentication,
     val parsedBody = DigestBodyParser.create(tempFile)
 
     AuthenticatedAndAuthorised.async(parsedBody) { req =>
-      val uploadTimeToRecord = DateTimeUtils.fromValueOrNow(uploadTime)
-      val uploadedByToRecord = uploadedBy.getOrElse(Authentication.getIdentity(req.user))
+      val onBehalfOfFn: OnBehalfOfPrincipal = auth.getOnBehalfOfPrincipal(req.user)
+      val isSoftDeleted = gridClient.getImageSoftDeletedStatus(req.body.digest, onBehalfOfFn)
+      isSoftDeleted.flatMap{
```
adrielulanovsky, line 84: An alternative way here would be to do something like this:
`val result = for {
    softDeleted <- isSoftDeleted
    _ <- if(softDeleted) Future.failed(ImageAlreadyExistsException) else Future.successful(true)
    ....`
    and then add a case in the recover section to handle that exception.

AWare:
I'm not sure why this is the desired behavior for attempting to reupload a deleted image.

AWare:
I think this should use the `result-editor` like for a successful upload, with a change there to allow the user to undelete if they have permission. This should link to the image page as normal, and mirror whatever the individual image UX for deleted images. Whatever UX for undeleting images (for users with permissions) in the result editor is, it should strongly consider whether it can apply to the entire upload.

kahuna/public/js/edits/image-editor.html
```diff
@@ -80,6 +83,15 @@
                                 image="ctrl.image">
             </gr-archiver-status>
         </div>
+
+        <div ng-if="ctrl.isDeleted" class="job-status status deleted_image_warning"
+             ng-class="{
+                         'status--invalid': !ctrl.canUndelete
+                     }">
+            <span ng-if="ctrl.canUndelete">Image deleted (this image already exist and has been deleted. Click Add to Library to undelete it)</span>
```
paperboyo, line 91:
```suggestion
            <span ng-if="ctrl.canUndelete">Image deleted (this image already exists and has been deleted. Click Add to Library to undelete it)</span>
```

kahuna/public/js/edits/image-editor.html
```diff
@@ -80,6 +83,15 @@
...
+            <span ng-if="!ctrl.canUndelete">Image deleted (this image already exist and has been deleted. Click View image to see details)</span>
```
paperboyo, line 92:
```suggestion
            <span ng-if="!ctrl.canUndelete">Image deleted (this image already exists and has been deleted. Click View image to see details)</span>
```

Sorry, my typos in mockups.

paperboyo (commented):
Sorry, just my typos in mockups. Native speakers, please adjust further if needed!

image-loader/app/lib/FailureResponse.scala
```diff
@@ -54,6 +54,16 @@ object FailureResponse extends ArgoHelpers with Results {
     )
   }
 
+  def imageAlreadyExists(exception: Exception): Response = {
+    logger.info(s"Rejected request to load file: image file is already exists and is soft deleted", exception)
```
twrichards, line 58: [minor]
```suggestion
    logger.info(s"Rejected request to load file: image file already exists and is soft deleted", exception)
```

kahuna/public/js/edits/image-editor.js
```diff
@@ -29,16 +29,26 @@ imageEditor.controller('ImageEditorCtrl', [
...
              editsService,
              editsApi,
              imageService,
-             collections) {
+             collections,
+             mediaApi) {
 
     var ctrl = this;
+    ctrl.canUndelete = false;
+    ctrl.showUndeleteMsg = false;
```
twrichards, line 45: can this be removed, I can't see it being used (nor set to anything other than false)

twrichards (commented):
the code looks good (minus a few comments)

@tjsilver and I are going to try it out in our TEST environment now...

paperboyo:
Thanks! Playing around with this in our test env, I have some feedback below. Worth noting that this is only UI review feedback (no code). Whether fixes belong in this PR or subsequent ones depends *only* on whether any UI changes are introduced in this PR or not. In other words: we can continue to work on this as long as the UI doesn’t do soft-deletes (last commits), but as soon as it does, we cannot merge until the UI works spotlessly…
1. Double Add to Library buttons in Viewer for a deleted image:
   image
   clicking one on the left doesn’t remove soft-deleted metadata and causes this:
   image
   clicking the one on the right then deletes metadata (good) but the button indicated image is added to Library (`archived`:`true`), while it should just be undeleted (and only the subsequent click on Add to Library should set `archive` bit).
2. Same double Add to Library buttons on the Upload page:
   image
   (worth noting that the Add to Library button bottom right behaves correctly: deletes metadata and doesn’t set `archive` bit)
3. Reuploading a soft-deleted image for a user *without* `delete_image` permission looks like for a user with the permission. It should look like the first screenshot [in here](https://github.com/guardian/grid/discussions/3364#discussioncomment-1148842) instead.

kahuna/public/js/components/gr-archiver/gr-archiver.js
```diff
@@ -101,6 +112,18 @@ module.controller('grArchiverCtrl', [
                 reduce((all, items) => all.union(items)).
                 toArray();
         }
+        function undelete() {
+            ctrl.undeleting = true
```
andrew-nowak, line 116: The linter is complaining about this 

```suggestion
            ctrl.undeleting = true;
```

paperboyo:
Looking much better, especially for unpermissioned users, thank you! Some minor things:

1. Upon reuploading a deleted image, when I click Add to Library button bottom right, the messaging (correctly) disappears, but the hover state for the small Add to Library button next to View image still says `Undelete` (should say `Add to Library` as the image isn’t deleted any more by that time)
2. Upon reuploading a deleted image, when I click small Add to Library button next to View image, the messaging doesn’t disappear. It should behave exactly as in pt. 1
3. I like how Viewer and hover state for the small Add to Library button next to View image says `Undelete` and not `Add to Library`. Maybe larger Add to Library button bottom right on the upload page should also say `Undelete`? Then, the messaging should be revised to say `Image deleted (this image already exists but has been deleted. Click Undelete to bring it back)` (maybe `bring it back` could be `resurrect it`?) (sorry, I know my mockups said `Add to Library` instead of `Undelete`, but I think the latter is clearer and is already so in the Viewer)
4. I think we should hide Delete button for deleted images in Viewer:
   image

For discussion:
A. Currently, it’s possible to have an image in **both** deleted and archived states. Should they be exclusionary states? UI is OK: in Viewer, the button says Undelete and after clicking, it says Remove from Library. So maybe it‘s fine?
B. Is Undeleteing behind archivist permission (which we don’t use at the Guardian)? Or is it under delete_image permission? Which one should it be behind? Latter, I think?

cc @sihil Especially regarding pt. A.

paperboyo:
@abdelrahmansd thank you! Some more testing the UI:

6. For a permissioned user, small Add to Library/Undelete image button next to View image looks broken:
    image
7. Add to Library button in the actions bar for a selection of soft-deleted images doesn’t show (it should say Undelete, just like in Viewer):
    image
8. For an unpermissioned user, small Add to Library/Undelete image button next to View image doesn’t do anything, but highlights on hover. Tootlip shown is correct, but maybe let’s get rid of the highlighting on hover? 
    image
9. Hovering over bigger undelete button on the Upload page weirdly triggers hover state for an image (see badges appearing):
    hover

Because of pt. 6, I cannot test pts. 1–2. Pts. 3–4 fixed 🎉 . Pt. 5 half-fixed (Delete is hooked up, but not Undelete, see pt 7)

abdelrahmansd:
@paperboyo thank you for your feedback
pt.6 : 




>     6. For a permissioned user, small Add to Library/Undelete image button next to View image looks broken:

small Undelete image button has been removed for permissioned users for now due to difficulty to share the state between the two components, and the undelete operation can be done from the large Undelete button on the upload page, however it can be handled in a separate PR

paperboyo:
> however it can be handled in a separate PR

Cool. Good enough (even looking slightly weird) if will be fixed later!

abdelrahmansd:
pt.7 :
>     7\. Add to Library button in the actions bar for a selection of soft-deleted images doesn’t show (it should say Undelete, just like in Viewer):

the same as pt.6 , and undelete operation can be done from the image view page, also it can be handled in a separate PR as well

abdelrahmansd:
pt.8
>     8. For an unpermissioned user, small Add to Library/Undelete image button next to View image doesn’t do anything, but highlights on hover. Tootlip shown is correct, but maybe let’s get rid of the highlighting on hover?

it can be removed as pt.6 for unpermissioned users also

paperboyo:
> it can be removed as pt.6 for unpermissioned users also

Nah. If future PRs can fix 6–8 (7 being the most important), this is alright as-is: these actions are all edge-casey, anyway, so users are unlikely to be confused.

blishen:
Discussed at Guardian BBC meet 13/9. We're going to disable the undelete buttons, and re-introduce in a separate PR, so we're inviting a Guardian code review now.

paperboyo:
> We're going to disable the undelete buttons, and re-introduce in a separate PR

This only applies to already disabled Undelete buttons in selection mode (and small ones on the upload page). Let’s review and see if we can merge it as-is (soft-delete becomes a thing!).

I will move remaining fixes (pts. 6–9) and remaining discussion points (A–D) out of comments here and into existing [Discussion topic](https://github.com/guardian/grid/discussions/3364) after code-review.

paperboyo:
Hello, @abdelrahmansd! We have discussed it extensively internally. We are happy with a code above, thank you! Because merging it constitutes a fundamental change to how Grid works since its inception and to user expectations, we would like to see the below three fixes before we will be able to merge it to PROD. I have no opinion of whether those fixes should be in the same or separate PRs, but they will need to go in before this can go in. (the rest of less important changes are kept [in here](https://github.com/guardian/grid/discussions/3364#discussioncomment-1326553)).

1. Soft-deleted images should be made Unusable. We already have that state (for when there is no credit and/or description). It prevents cropping. It should look something like this:
    image
2. Users without delete_image permissions should not be seeing any results for `+is:deleted` search
3. Current implementation produces one regression to current user behaviour: unpermissioned user uploads an image, deletes it, but cannot undelete it (reuploading doesn’t help)

Ideally, pts. 2 & 3 would be fixed such that:
- users without delete_image permission would only be seeing **their own** deleted images in `+is:deleted` search
- users without delete_image permission would be able to undelete **only their own** uploads

If this is not possible, then, maybe, we can live with the worse implementation:
- users without delete_image permissions would not be seeing any deleted images in `+is:deleted` search (how hard would it be to hide the chip in UI too?)
- users without delete_image permissions would be able to undelete **only their own** uploads by reuploading them (instead of being able to use the UI; normal UI for reuploading deleted images for unpermissioned users would just not show)

Sorry to be introducing these points so late, Ebu! It’s too often the case that the full ramifications of a complex feature become clear only in the process of implementing it (instead of in all the previous discussions). We hope this is not too much to ask, but we feel we need those changes to be sure users won’t be unnecessarily surprised (we will also be using hard-delete CURL commands for real emergencies we decided).

kahuna/public/js/components/gr-archiver-status/gr-archiver-status.js
```diff
@@ -29,7 +29,7 @@ archiver.controller('ArchiverCtrl',
     ctrl.undelete = undelete;
 
     mediaApi.getSession().then(session => {
-        if (ctrl.image.data.softDeletedMetadata !== undefined && session.user.permissions.canDelete) { ctrl.canUndelete = true; }
+        if (ctrl.image.data.softDeletedMetadata !== undefined && session.user.permissions.canDelete || session.user.email === ctrl.image.data.uploadedBy) { ctrl.canUndelete = true; }
```
andrew-nowak, line 32: I think this (and the other conditions like this in this latest commit) needs to have brackets around the or, like this:

`if (ctrl.image.data.softDeletedMetadata !== undefined && (session.user.permissions.canDelete || session.user.email === ctrl.image.data.uploadedBy)) { ctrl.canUndelete = true; }`

without them, the condition evaluates to true even when the image isn't soft deleted - and I can see an 'Undelete' button when uploading an image for the first time!

image

media-api/app/controllers/MediaApi.scala
```diff
@@ -88,7 +88,8 @@ class MediaApi(
       Link("collections",     config.collectionsUri),
       Link("permissions",     s"${config.rootUri}/permissions"),
       Link("leases",          config.leasesUri),
-      Link("admin-tools",     config.adminToolsUri)
+      Link("admin-tools",     config.adminToolsUri),
+      Link("undelete",        s"${config.rootUri}/images/{id}/undelete ")
```
andrew-nowak, line 92:
```suggestion
      Link("undelete",        s"${config.rootUri}/images/{id}/undelete")
```

kahuna/public/js/upload/jobs/upload-jobs.js
```diff
@@ -79,6 +79,7 @@ jobs.controller('UploadJobsCtrl', [
 
                         imageService(image).states.canDelete.then(deletable => {
                             jobItem.canBeDeleted = deletable;
+                             if (image.data.softDeletedMetadata !== undefined) { ctrl.isDeleted = true; }
```
andrew-nowak, line 82: I think this should set `jobItem.isDeleted` instead (and update the condition in upload-jobs.html to match) - it's possible for a user to upload both deleted and not-deleted images in the same batch

andrew-nowak:
Thanks so much for adding the soft-deleted metadata table! It looks really good, but we need to make use of it by including it in image projections. The critical lines of code are here: https://github.com/guardian/grid/blob/main/common-lib/src/main/scala/com/gu/mediaservice/ImageDataMerger.scala#L56 

Only thing to be aware is that the source of the metadata going into the above `aggregate` function must be dynamo, not elasticsearch; I'd suggest implementing via a new endpoint in media-api which retrieves the metadata from dynamo, and a corresponding method in the GridClient, which should then fit into the pattern used for the others.

Sorry to add yet another thing, but this is important because while the Guardian are running migrations, if the image has been soft-deleted in the old index we need to be sure that the metadata will persist into the new index so that images don't appear to become undeleted.

PR #3436, opened 2021-08-13
Periodically fetch ongoing migration index name and store locally

andrew-nowak:
## What does this change?

We anticipate needing to know the name of the index for the ongoing migration. We can fetch this from the cluster when needed, but this incurs the time penalty of yet another round trip. This PR adds a periodically refreshed field to the Elasticsearch clients to allow accessing the index name without needing to call out to the cluster.

Times we know we'll need the migration index:

* After migrating an image, we need to write back the index name that the image was migrated to
* When querying for more images to migrate
* We need to know if the migration is under way when searching
* We can avoid unnecessary reads and writes against the migration alias if there is no migration index

You can get the index name from the Elasticsearch client.

## Tested? Documented?
- [x] locally by committer


common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchClient.scala
```diff
@@ -27,6 +30,20 @@ trait ElasticSearchClient extends ElasticSearchExecutions with GridLogging {
...
+  def scheduler: Scheduler
+
+  val maybeMigrationIndexName = new AtomicReference[Option[String]](None)
+
+  private val migrationIndexNameRefresher = scheduler.schedule(
+    initialDelay = 0.seconds,
+    interval = 1.minute
+  )(
+    () => {
+      // TODO what happens if this times out/fails?
```
twrichards, line 42: did you get to the bottom of this, or are we going to wait and see?

twrichards (approved):
we worked on this together, so perhaps we want other eyes, but here's an approval so you're not blocked

AWare:
Where do we need to know the index name and not the alias?

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchClient.scala
andrew-nowak: the scheduler's scaladoc says

> If the function throws an exception the repeated scheduling is aborted, i.e. the function will not be invoked any more.

so I've updated the scheduled function to log but swallow any errors it encounters, which allows for the error to self-heal between invocations. How does this sound?

andrew-nowak:
@AWare When migrating images, we are writing the success/failure back to the 'current' index with the real index name, as that allows us to distinguish between migrations, so any code which wants to interact with that logic needs to access the name. Also, this PR could allow us to improve `migrationClient.getStatus()` to not need to make the round trip to the cluster, which will be useful when doing the parallel reads/writes.

AWare:
I think some of this may better be served (I fear) in painless scripts.

twrichards:
> @AWare When migrating images, we are writing the success/failure back to the 'current' index with the real index name, as that allows us to distinguish between migrations, so any code which wants to interact with that logic needs to access the name. Also, this PR could allow us to improve `migrationClient.getStatus()` to not need to make the round trip to the cluster, which will be useful when doing the parallel reads/writes.

Just to elaborate on the above, we have lots of scenarios where we need to know whether there is a migration underway and specifically what the resolved index name for the `Images_Migration` alias is, for example...

- when we're searching for images we need to use the multi-target notation (only if there is migration underway) whilst also filtering out results which have a `esInfo.migration.migratedTo` which matches the current migration index
- when doing other read operations, we can avoid a failing query on the migration index for the majority of the time when we're not running a migration
- when writing @andrew-nowak has found its difficult to distinguish between a failure due to the 
- when querying for what needs migrating next we will need this migration index name to filter out previous failures to migrate to that alias (keys of the `esInfo.migration.failures`) and to filter out images that have already been migrated to that specific migration index name (`esInfo.migration.migratedTo`)
- and probably a few more efficiency savings (or rather, avoiding attempting things that will fail)

> I think some of this may better be served (I fear) in painless scripts.

@AWare that might be possible but painless scripts are less accessible to devs (than a simple `Option[String]` with a timer to update it) so if we can minimise the extra lines of painless required for this migration I think that's worth it.

AWare:
Regarding painless, when we attempt to write an image: 
It doesn't seem possible to construct a painless query which can find out the current migration status. If this could be extracted in painless, we could naively perform both updates and let elasticsearch figure out the correct migration status fields.

andrew-nowak:
From a conversation on the above discussion, the two options open to us seem to be:
- Do the extra round trip to find the migration status/index name on each request which needs it.
- Cache the status/name of the migration index (ie. this PR).

It's not yet clear what the performance impact of doing the extra round trip would be yet. It might be possible to measure once the rest of the migration work is done (ie. by measuring throughput of migration with/without extra round trip).

We considered once downside of caching the migration index name: what happens during the time between a migration starting and the new name entering the cache. There was an intention to account for this by updating the cached index name when a new migration is started, but this approach was flawed as the `MigrationClient` and `ElasticSearch` were two separate instances of the `ElasticSearchClient` trait (which managed this cached value) which meant that the `MigrationClient` could only update its own instance of the cached value.

The latest commit on this PR changes `MigrationClient` to contain the cached index name (IMO an improvement in code complexity, moves more migration logic into one file), converts it to a trait which is mixed into the `ElasticSearch` class (a first attempt kept each as independent classes, but this runs into circular dependency issues, because each of `MigrationClient` and `ElasticSearch` wants to use values/methods from the other, and neither can be cleanly created without the other), and moves `MigrationClient` from thrall to common-lib, so that it can also be consumed by media-api.

twrichards (approved):
I'm gonna deploy this to `TEST` now, but code looks great now!

PR #3453, opened 2021-08-26
Migration aware updates

andrew-nowak:
## What does this change?

Updates Thrall's ElasticSearch class to perform updates and deletes against both the current and (if exists) migration indexes, using the migration status refresher introduced in #3436. Requests against the migration index are only sent if there is an in progress migration.

Run an update or deletion against a previously-migrated image, and observe the update/deletion be applied in both indexes.

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/thrall/app/lib/elasticsearch/ElasticSearch.scala
+++ b/thrall/app/lib/elasticsearch/ElasticSearch.scala
@@ -339,17 +339,12 @@ class ElasticSearch(
       nestedQuery("usages").query(existsQuery("usages"))
     )
 
-    // Uses the cached migration status to determine if a migration is in progress: if there is not, this will be None
-    // and get flattened out
-    val migrationAliasIfActive = migrationStatus match {
-      case InProgress(_) => Some(imagesMigrationAlias)
-      case _ => None
+    (migrationStatus match {
+      case InProgress(migrationIndexName) => List(imagesCurrentAlias, migrationIndexName)
+      case _ => List(imagesCurrentAlias)
+    }).map { index =>
+      deleteFromIndex(id, index, deletableImage).map { _ => ElasticSearchDeleteResponse() }
     }
-    List(Some(imagesCurrentAlias), migrationAliasIfActive)
-      .flatten
-      .map { indexAlias => deleteFromIndex(id, indexAlias, deletableImage).map { _ =>
-        ElasticSearchDeleteResponse()
-      }}
   }
 
   def deleteAllImageUsages(id: String,
--- a/thrall/app/lib/elasticsearch/ElasticSearch.scala
+++ b/thrall/app/lib/elasticsearch/ElasticSearch.scala
@@ -2,7 +2,7 @@ package lib.elasticsearch
 
 import akka.actor.Scheduler
 import com.gu.mediaservice.lib.ImageFields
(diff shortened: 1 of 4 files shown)


thrall/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -37,6 +37,32 @@ class ElasticSearch(
...
+    // coalesce all other errors.
+    val runForCurrentIndex: Future[Option[Response[UpdateResponse]]] = executeAndLog(requestFromIndexName(imagesCurrentAlias), logMessageFromIndexName(imagesCurrentAlias), notFoundSuccessful).map(Some(_))
+    // Update requests to the alias throw if the alias does not exist, but the exception is very generic and not cause is not obvious
+    // ("index names must be all upper case")
+    val runForMigrationIndex: Future[Option[Response[UpdateResponse]]] = migrationStatus match {
+      case InProgress(_) => executeAndLog(requestFromIndexName(imagesMigrationAlias), logMessageFromIndexName(imagesMigrationAlias), notFoundSuccessful = true).map(Some(_))
+      case _ => Future.successful(None)
+    }
+    // remove the optionality of the completed futures. runForCurrentIndex will always be Some, so there will always be a head.
+    Future.sequence(List(runForCurrentIndex, runForMigrationIndex)).map(_.flatten.head)
```
twrichards, line 63: excellent comments

thrall/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -295,19 +339,17 @@ class ElasticSearch(
...
     }
-
-    List(eventualDeleteResponse.map { _ =>
-      ElasticSearchDeleteResponse()
-    })
+    List(Some(imagesCurrentAlias), migrationAliasIfActive)
+      .flatten
+      .map { indexAlias => deleteFromIndex(id, indexAlias, deletableImage).map { _ =>
+        ElasticSearchDeleteResponse()
+      }}
```
twrichards, line 352: why do we need to use the migration alias here, when we have the actual index name? does `deleteFromIndex` not like real index names??

twrichards (approved):
Great work! one question (but not a blocker)

✅  tested on `TEST` (verified with REST calls in `cerebro`)

thrall/app/lib/elasticsearch/ElasticSearch.scala
andrew-nowak: I don't think there was any specific reason to use the alias over the index name, they should be equivalent AFAIK. I can switch it to the name if there's a strong preference for that

twrichards: not a strong preference, just we're ignoring the value with an `_` when we could perhaps simplify the whole block and eliminate the variable, something like... 
```scala
migrationStatus match {
  case InProgress(migrationIndexName) => List(imagesCurrentAlias, migrationIndexName)
  case _ => List(imagesCurrentAlias)
}.map { index => 
  deleteFromIndex(id, index, deletableImage).map { _ => ElasticSearchDeleteResponse() }
}
```
(this avoids the implicit conversion too)

twrichards (approved):

PR #3458, opened 2021-09-01
Cast unknown type

andrew-nowak:
## What does this change?

A new major version (v3.0.0) of node-fetch was released yesterday, which was automatically pulled into builds as image-counter-lambda dependency is declared as `latest`.

The only breaking change observed so far is that the newly bundled type definitions declare the response to be `unknown` rather than `any`, which requires us to explicitly cast to the expected type, implemented in this PR.

Alternate solutions/proposals:

- Use `npm ci` in the Github Actions workflow, to ensure the same version is used in subsequent builds.
- Restrict the version range of node-fetch.

Status checks go green.

## Tested? Documented?
- [x] locally by committer


twrichards (approved):
nice work - although we should definitely not have `latest` for any dependency... it should be an explicit version (I suggest we correct that in this PR).

twrichards (approved):

PR #3462, opened 2021-09-06
Ongoing query to migrate images

andrew-nowak:
## Tested? Documented?
- [ ] locally by committer


thrall/app/lib/ThrallStreamProcessor.scala
```diff
@@ -92,9 +97,10 @@ class ThrallStreamProcessor(
         }
 
     // merge in the re-ingestion source (preferring ui/automation)
-    val mergePreferred = graphBuilder.add(MergePreferred[TaggedRecord[ThrallMessage]](1))
+    val mergePreferred = graphBuilder.add(MergePreferred[TaggedRecord[ThrallMessage]](2))
     uiAndAutomationMessagesSource ~> mergePreferred.preferred
-    migrationMessagesSource ~> mergePreferred.in(0)
+    migrationManualMessagesSource ~> mergePreferred.in(0)
+    migrationOngoingMessagesSource ~> mergePreferred.in(1)
```
sihil, line 103: Don't do this, it doesn't do what you think (it will interleave `0` and `1` rather than preferring `0` over `1`). Instead chain two separate MergePreferred pieces.

twrichards: yeah we did wonder about that, thanks for letting us know. 
its not crucial that manual get done first as as the `migrationOngoingMessagesSource` is pulling in batches of 10 anyway, so manual one will respond promptly enough... although @andrew-nowak it's maybe worth the extra lines of code to combine with their own MergePreferred first

andrew-nowak: this makes sense to me, will do this too

Issue #1903, opened 2016-05-19
Add to Collection from image view small issues


paperboyo:
Hi,
1. The close button dismissing Add to Collection button is far away from the modal Collections selection
2. The modal should be dismissible with ESC
3. The modal does not darken everything (red triangle on screenshot).
4. The tooltip for Add to Collection is more verbose than all other Add to... tooltips, for brevity and consistency should read "Add to collection"
5. I should not be allowed to play with UI

![add-to-collection](https://cloud.githubusercontent.com/assets/6032869/15390186/6f988308-1db2-11e6-9eaa-98e03878323b.png)

Regards
Mateusz

paperboyo:
6. Closing as it’s been more than 4 years and it will be easier to just rewrite the whole UI, I suppose.

PR #3465, opened 2021-09-07
Unify Image Viewer's and Image Browser's info panels

adrielulanovsky:
## What does this change?
Following the discussion on https://github.com/guardian/grid/discussions/3448, this attempts to unify the code and UI for the metadata section of the info panels on image viewer and image browser.

Behaviour should stay the same, while using only one template and controller for both panels.

## Tested? Documented?
- [ ] locally by committer


paperboyo:
This is great! Most bugs with those panels are not specific to this work: multiple redundant scrollbars, whole image moving around, editing hints (_Multiple bylines…_) sometimes in italic, sometimes not or [these gremlins](https://github.com/guardian/grid/issues/1903).

It’s great that you’ve described future steps around tokens! I think most can be modelled like Labels, with just some colour/CSS tweaks.

This is what I could find specific to this PR:

1. We do not show leases for users without edit_metadata in Viewer (this is questionable, because leases are not shown, nor their effect, but nevertheless this is how it is). In Browser, we show some useless (_current + inactive_) text which lies about the counts (unlike Usage rights count). If anything, it’s better to copy current Viewer, than current Browser:
    Current PROD Viewer:
    image
    This PR (and current Browser):
    image
2. Special instructions is editable, so should behave like other fields (should show _Multiple intructions (click ✎ to edit all)_) and not disappear:
  Special

Will add more if I find anything. This looks very good. Just a note that there is [this abandoned PR](https://github.com/guardian/grid/pull/3083) with slightly related work…

paperboyo:
> I fixed issue number 2

Yay! Thanks!

> do you want me to hide the Leases panel when there are multiple images selected?

Yeah, I think so. _current leases + inactive leases_ text is useless without the actual counts! Yes, the problem precedes this PR and yes, it would be great to have actual dynamic counts that do not break the CSS, haha.

> image viewer has the following condition for showing leases: permission to edit metadata OR number of leases > 0. Do we want to maintain that condition?

I think it was @andrew-nowak who spotted that we have a bug with the second part of the condition, so that none of the leases ever show for non-permissioned users. We have users complained that if unpermissioned users are asking permissioned users to grant a lease they have no way of knowing this has actually happened apart from the fact that the _Cannot crop_ changes to _Crop_. 

This can be sorted in two ways:
1. either don’t display the leases, but create a special state for unpermissioned users clearly indicating a leased image is fine to use
2. fix the condition and display active leases to unpermissioned users (but obviously read-only). bonus for hiding syndication leases as these are useless for them

I think option 2 is much easier to implement (just fixing a bug), but I’m famous for thinking something I won’t actually be fixing myself is easy to fix :-).

paperboyo:
[ignore, not directly relevant to this PR work]

> they have no way of knowing this has actually happened

Just for the sake of completeness, and for my own record, because I’m testing this as I go, this isn’t strictly true. The other way to know something has been leased for an unpermissioned user is the lack of these additional banners
- No rights imagery without a lease:
    image
- Chargeable category without a lease:
    image

So if we ever think of solution 1, those banner better look different.

It’s news to me (!) that not all Restricted images are equal (given same lease status and other variables):
- Some **can be cropped** by unpermissioned users (so it’s just a warning, really), they only have `"cost": "conditional"` in the API:
    image
- Others, **cannot be cropped** and they have both "cost": "conditional" and `"invalidReasons":"conditional_paid": "This image is restricted use"`.
    image

Weirdly, both of those types are displayed under _Free to use_ search! This smells a bit broken, but I bet nobody ever noticed… I have no idea what controls those two types of Restricted, but it isn’t `UsageRights.scala`.

paperboyo:
Testing, (Date) Taken isn’t shown:
PROD:
image
TEST:
image

adrielulanovsky:
> Testing, (Date) Taken isn’t shown: PROD: image TEST: image
> 
> And if super-easy, I think I like the more subdued missing values from PROD Viewer better (Copyright: unknown) than the same intensity ones from TEST, what do you think?

@paperboyo Fixed

adrielulanovsky:
> Almost there!
> 
>     1. Description text got greyed out:
>        image
> 
>     2. Could we maintain filename wrapping behaviour from current PROD Viewer? Applies to **at least** Description and Filename. Filename shown here:
>        PROD (long filename without spaces):
>        image
>        PROD (long filename with spaces):
>        image
>        TEST (long filename without spaces):
>        image
>        TEST (long filename without spaces):
>        image
> 
>     3. [only if needed after sorting pt. 2] Could we keep the behaviour of [this little guy](https://github.com/guardian/grid/pull/2637) which allows selecting **the whole** of truncated filename with one click? (otherwise it’s impossible to copy the whole of truncated filename) It should only apply to filename field (otherwise it would be impossible to select part of eg. Description etc):
>        sdfsdf
> 
>     4. Could we keep the tooltip showing the whole of any truncated field?
>        image
> 
>     5. In Browser, any non-array, non-editable, always-present fields (eg. Uploaded, Uploader, Filename) should be shown as long as the values are the same for all selected images. As soon as the values differ, the whole field should disappear. It is of no use to say eg. `Filename: Multiple filenames` (since it’s impossible to edit it either). Currently, Filename stays, while eg. Uploader disappears even if it’s the same for all selected photos.

Issues 1 to 5 should be fixed now. Thanks for the feedback

adrielulanovsky:
> And the tiniest one,
> 
>     1. Little plus icons for Leases, Collections and Labels should be `#cccccc` like on Collections, instead of `#999999` like on Leases and Labels:
>        image

5 and 6 should be fixed now, please have a look. Thanks!

PR #3466, opened 2021-09-08
re-use the configurable `scriptsToLoad` config to populate the `frame-src` directive of the Content-Security-Policy

twrichards:
In https://github.com/guardian/editorial-tools-pinboard/pull/58 we introduce an iFrame in pinboard client back to the pinboard domain, Grid's previous `frame-src` blocked that. This PR re-uses the configurable `scriptsToLoad` config to populate the `frame-src` directive of the Content-Security-Policy.


tjsilver (approved):
👍

PR #3467, opened 2021-09-09
Configurable fuzzy search

ochiengolanga:
## What does this change?

Typos when searching for images often happen and returning search results based on exact match without handling typos can reduce the user's experience. 

This PR introduces fuzzy search to allow inexact fuzzy matching to return documents in Elasticsearch that contain terms similar to the search term, as measured by a [Levenshtein edit distance](https://en.wikipedia.org/wiki/Levenshtein_distance).

For instance, a fuzzy search for "lightrom" (which is a typo) would match the word "lightroom" and return documents that have the term lightroom.

Fuzzy search is only applied to Word queries e.g. Lightroom Develop and not Phrase queries e.g. "Lightroom Develop" (wrapped in quotes) as multi-match queries are treated as exact term match.

The configurations needed in `media-api.conf` to enable fuzzy search are:

```hocon
search.fuzziness = {
  enabled = true # optional and defaults to false
  editDistance = 2 # optional and defaults to AUTO - Levenshtein edit distance
}
```

Enable fuzzy search by adding `search.fuzziness.enabled = true` configuration to media-api.conf.

### Sample image with metadata

Sample image

### Search results with "lightroom" search term

Search lightroom

Should return 1 result.

### Search results with a typo in the search term "lightrom" and fuzzy search disabled

Search lightrom - fuzzysearch disabled

Should return 0 results.

### Search results with a typo in the search term "lightrom" and fuzzy search enabled

Search with typo - fuzzysearch enabled

Should return 1 result.

## Who should look at this?

@guardian/digital-cms

## Tested? Documented?
- [X] locally by committer
- [X] relevant documentation added or amended (if needed)

### DIFF
--- a/media-api/app/lib/elasticsearch/QueryBuilder.scala
+++ b/media-api/app/lib/elasticsearch/QueryBuilder.scala
@@ -26,10 +26,18 @@ class QueryBuilder(matchFields: Seq[String], overQuotaAgencies: () => List[Agenc
   private def multiMatchPhraseQuery(value: String, fields: Seq[String]): MultiMatchQuery =
     ElasticDsl.multiMatchQuery(value).fields(fields).matchType(MultiMatchQueryBuilderType.PHRASE)
 
+  private def multiMatchWordQuery(value: String, fields: Seq[String]): MultiMatchQuery = {
+    val multiMatchQuery = ElasticDsl.multiMatchQuery(value).fields(fields).operator(Operator.AND)
+
+    if (config.fuzzySearchEnabled) {
+      multiMatchQuery.matchType(MultiMatchQueryBuilderType.BEST_FIELDS).fuzziness(config.fuzzySearchEditDistance)
+    } else {
+      multiMatchQuery.matchType(MultiMatchQueryBuilderType.CROSS_FIELDS)
+    }
+  }
+
   private def makeMultiQuery(value: Value, fields: Seq[String]): MultiMatchQuery = value match {
-    case Words(value) => ElasticDsl.multiMatchQuery(value).fields(fields).
-      operator(Operator.AND).
-      matchType(MultiMatchQueryBuilderType.CROSS_FIELDS)
+    case Words(value) => multiMatchWordQuery(value, fields)
     case Phrase(string) => multiMatchPhraseQuery(string, fields)
     // That's OK, we only do date queries on a single field at a time
     case e => throw InvalidQuery(s"Cannot do multiQuery on $e")
--- a/media-api/app/lib/MediaApiConfig.scala
+++ b/media-api/app/lib/MediaApiConfig.scala
@@ -37,6 +37,12 @@ class MediaApiConfig(resources: GridConfigResources) extends CommonConfigWithEla
   val cloudFrontDomainThumbBucket: Option[String] = stringOpt("cloudfront.domain.thumbbucket")
   val cloudFrontKeyPairId: Option[String]         = stringOpt("cloudfront.keypair.id")
 
+  val fuzzySearchEnabled: Boolean = boolean("search.fuzziness.enabled")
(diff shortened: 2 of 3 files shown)


paperboyo:
Cool! Does it apply to any “chips” queries? Eg. `+byline:Mateusz` etc? (I don’t think it should?)

ochiengolanga:
@paperboyo with these changes it doesn't. Only applicable to search without search hints and exact search using quotes. I'll be more than happy to include search hints if necessary.

paperboyo:
Thanks! I think it should **only** apply to free text search (not with hints and not when using quotes).

So this is correct in PR description:
> Fuzzy search is only applied to Word queries e.g. Lightroom Develop and not Phrase queries e.g. "Lightroom Develop" (wrapped in quotes)

media-api/app/lib/elasticsearch/QueryBuilder.scala
```diff
@@ -26,10 +26,18 @@ class QueryBuilder(matchFields: Seq[String], overQuotaAgencies: () => List[Agenc
   private def multiMatchPhraseQuery(value: String, fields: Seq[String]): MultiMatchQuery =
     ElasticDsl.multiMatchQuery(value).fields(fields).matchType(MultiMatchQueryBuilderType.PHRASE)
 
+  private def multiMatchWordQuery(value: String, fields: Seq[String]): MultiMatchQuery = {
+    val multiMatchQuery = ElasticDsl.multiMatchQuery(value).fields(fields).operator(Operator.AND)
+
+    if (config.fuzzySearchEnabled) {
+      multiMatchQuery.matchType(MultiMatchQueryBuilderType.BEST_FIELDS).fuzziness(config.fuzzySearchEditDistance)
+    } else {
+      multiMatchQuery.matchType(MultiMatchQueryBuilderType.CROSS_FIELDS)
```
andrew-nowak, line 35: Why do you switch to best_fields when fuzzy searching is turned on, rather than sticking with cross_fields?

ochiengolanga: @andrew-nowak Elasticsearch (upto version 7.14) does not support fuzzy search on cross_fields multi-match queries.

andrew-nowak: @ochiengolanga
thanks! Just to be sure - have your stakeholders seen and approved the trade-off (that users cannot search across fields once fuzziness is enabled)? We'd be unlikely to turn this on on our side while `cross_fields` can't be used with fuzzy searching, as we don't want to remove functionality that users rely on. 
If you do want this behaviour anyway, can you make sure to clearly document how enabling fuzzy searching changes the search behaviour, including the downsides.

paperboyo:
😱 As per [Andrew’s comment](https://github.com/guardian/grid/pull/3467#discussion_r708264148), switching this ON makes free-text search **unable** to search within [multiple fields](https://github.com/guardian/grid/blob/08cb88f23b7b6394cbf776d980e8239aee2ad98f/media-api/app/lib/elasticsearch/MatchFields.scala#L11-L16) at once. This greatly reduces the usefulness of free-text search and makes it less predictable too. I would say it reduces it so much that it renders any gains from fuzziness non-existent, sadly.
In my opinion, if fuzzy search cannot be made to work with searching multiple fields at a time, this should not be merged and will need a rethink.

ochiengolanga:
@paperboyo We are in concensus that enabling this functionality greatly reduces user's search experience and the gains are less valuable compared to the tradeoff. It would be really nice if we could have both (mutlifield sesrch and fuzzy search), until then we'll be parking the improvement and closing the PR.

ochiengolanga:
Closing this PR as the implementation introduces a trade-off (disabling multi field multimatch query) that greatly reduces the user's search experience.

PR #3472, opened 2021-09-16
Fix empty Staff Illustrator Usage Rights

adrielulanovsky:
## What does this change?
There is a bug where if the configuration has an empty option list for staffIllustrators, the backend will still send an options value to the frontend (with an empty list) causing the frontend to not render the text field for the creator input. This makes it impossible to add this usage right, since the creator field is required.

This change makes the backend check if the list is empty, and in that case not send an options value at all, which will allow the frontend to render the text input.

When having an empty option list for staffIllustrators in the usageRightsConfigProvider, users can still select the Staff Illustrator option and input the illustrator's name manually.

## Screenshots
Before change:
image

After change:
image

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/metadata-editor/app/model/UsageRightsProperty.scala
+++ b/metadata-editor/app/model/UsageRightsProperty.scala
@@ -103,8 +103,10 @@ object UsageRightsProperty {
       illustratorField(sortPublicationList(p.contractIllustrators), "publication")
     )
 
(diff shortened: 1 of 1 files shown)


tjsilver (approved):
LGTM 👍

PR #3483, opened 2021-09-23
Improve logging in image-loader projection endpoint

andrew-nowak:
Co-authored-by: Jonathon Herbert <jonathon.herbert@guardian.co.uk>

## What does this change?

Adds a helpful log of error when projection endpoint is not successful.

Developers find it easier to debug errors in the projection endpoint.

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/image-loader/app/controllers/ImageLoaderController.scala
+++ b/image-loader/app/controllers/ImageLoaderController.scala
@@ -71,11 +71,11 @@ class ImageLoaderController(auth: Authentication,
         "filename" -> filename.getOrElse(FALLBACK)
       )
     )
-    logger.info("loadImage request start")
+    logger.info(context, "loadImage request start")
 
     // synchronous write to file
     val tempFile = createTempFile("requestBody")
-    logger.info("body parsed")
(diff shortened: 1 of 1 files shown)


jonathonherbert (approved):

jonathonherbert:
@andrew-nowak ooh, one point re: @paperboyo's suggestion – could we change this to add a log marker containing the image id (e.g. `imageId` -> id)? This will let us index on that marker in our logs, opening up the possibility of a dashboard w/ a table of image failures – which as Mat says, might be quite handy.

paperboyo:
Not sure if this comment applies, just wanted to say that for images that failed ingestion proper, there isn’t a mediaID (filename is the only thing that will identify them, I think).

jonathonherbert:
You are correct, @paperboyo – in some places, we won't get an image id. We do have other context which wasn't coming through – but now is present: tested in code, w/ before/after

Screenshot 2021-09-23 at 16 48 21

jonathonherbert (approved):

PR #3488, opened 2021-09-28
dont optimise not transparent tiffs

AWare:
## Tested? Documented?
- [ ] locally by committer


AWare:
to actually do this we need an implementation without pngquant to use instead so closing

PR #3491, opened 2021-10-01
Better colourModel identification incl. some imaging ops

AWare:
Co-authored-by: Andrew Nowak @andrew-nowak 
Co-authored-by: paperboyo @paperboyo 

## What does this change?
This makes `colourModel` identification more robust. It was missing/wrong for non-JPEG files. Because [it’s used to decide on the colour of the crops](https://github.com/guardian/grid/blob/0f136041c2e73ae14fd190785ab3f108bd5bd4ee/common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala#L65), this caused wildly incorrect colour for eg. CMYK TIFFs.
It partially fixes https://github.com/guardian/grid/issues/2594. It’s not perfect, eg. Bitmap (1-bit) images are identified as `GRAYSCALE`, but it’s much better.

It also improves thumbnail generation: as they are JPEGs, it bakes them into the #33333 background. This is cheating, but it’s better than current situation of having unpredictable and wrongly baked in background colour staring at us from transparent areas. Maybe one day, thumbnails with transparency, should be transparent and show CSS checkerboard pattern, like preview and crop images…

It also reduces the bitdepth of `browserViewable` intermediary PNG image created from a source TIFF. For some reasons, feeding pngquant 8bpc data (when creating `optimisedPng` from this `browserViewable`) results in less dithering (and at least in the example below, half the size). 

More correct metadata. Correct colour for some crops. Better thumbnails.

## Screenshots
image
[correct colour on the left]

Thumbnails BEFORE:

image

Thumbnails AFTER:

image

`optimisedPng` BEFORE (863 bytes):

`optimisedPng` AFTER (417 bytes):

## Tested? Documented?
- [ ] locally by committer


common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/im4jwrapper/ImageMagick.scala
```diff
@@ -31,13 +31,15 @@ object ImageMagick extends GridLogging {
   def interlace(op: IMOperation)(interlacedHow: String): IMOperation = op <| (_.interlace(interlacedHow))
 
   def runConvertCmd(op: IMOperation, useImageMagick: Boolean): Future[Unit] = {
```
AWare, line 33: refactor to not use a boolean?

andrew-nowak:
TODO: 
1. either don't run correctColour when we've created a transformed/browserviewable image (ie. input was tiff) OR make correctColour read iccColourSpace only from original
2. make correctColour never concern itself with colour models which we don't have defaults (ie. is colourModel `LAB` or `Mateusz` or something else? do nothing)

cropper/app/lib/Crops.scala
```diff
@@ -31,15 +31,23 @@ class Crops(config: CropperConfig, store: CropStore, imageOperations: ImageOpera
...
 
     val source   = crop.specification
     val metadata = apiImage.metadata
     val iccColourSpace = FileMetadataHelper.normalisedIccColourSpace(apiImage.fileMetadata)
 
     for {
-      strip <- imageOperations.cropImage(sourceFile, apiImage.source.mimeType, source.bounds, masterCropQuality, config.tempDir, iccColourSpace, colourModel, mediaType)
+      strip <- imageOperations.cropImage(
+        sourceFile, apiImage.source.mimeType, source.bounds, masterCropQuality, config.tempDir,
+        iccColourSpace, colourModel, mediaType, isTransformedFromSource = false
```
jonathonherbert, line 49: Why is `isTransformedFromSource` `false` here?

common-lib/src/main/scala/com/gu/mediaservice/lib/ImageIngestOperations.scala
```diff
@@ -88,7 +88,7 @@ case class StorableOptimisedImage(id: String, file: File, mimeType: MimeType, me
 
 
 
-case class BrowserViewableImage(id: String, file: File, mimeType: MimeType, meta: Map[String, String] = Map.empty, mustUpload: Boolean = false) extends ImageWrapper {
+case class BrowserViewableImage(id: String, file: File, mimeType: MimeType, meta: Map[String, String] = Map.empty, isTransformedFromSource: Boolean = false) extends ImageWrapper {
```
jonathonherbert, line 91: Could we have a comment to expand on what `isTransformedFromSource` means in the broader context?

common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
```diff
@@ -137,19 +165,23 @@ class ImageOperations(playPath: String) extends GridLogging {
                       qual: Double = 100d,
                       outputFile: File,
                       iccColourSpace: Option[String],
-                      colourModel: Option[String])(implicit logMarker: LogMarker): Future[(File, MimeType)] = {
+                      colourModel: Option[String],
+                      isTransformedFromSource: Boolean
```
jonathonherbert, line 169: Let's add a comment to document the new param.

cropper/app/lib/Crops.scala
jonathonherbert: Answer from @andrew-nowak – we're dealing with the original file (see all the references to `source`), so it won't have been through the image pipeline.

common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
jonathonherbert: Refactored w/ @andrew-nowak and @paperboyo to pass a `BrowserViewable` image straight through, negating the need for a comment 🎉

jonathonherbert (approved):
Discussed over VC with @paperboyo and @andrew-nowak, a few comments addressed in situ. LGTM pending any further testing from @paperboyo 👍

PR #3494, opened 2021-10-05
Speed up projection by downloading computed images where possible

andrew-nowak:
## What does this change?

The projection endpoint is slow, and a bottleneck for running migrations. Most of the time in a projection is spent converting images - into thumbnails (always) and the intermediary and optimised PNGs (sometimes). The intermediary PNG is never saved, so we can't avoid that work, but the thumbnails and optimised PNGs were created on the first import to grid, and the new versions created during projection aren't saved, so why bother generating them at all?

This PR changes projection logic to first attempt to download the output image from S3 if it can, and only if that fails fall back to generating from the original.

Projections are fasted (I observed migration throughput go from ~400/min before this change to ~750)

## Tested? Documented?
- [x] locally by committer


paperboyo:
This is gold, thanks!

> new versions created during projection aren't saved

This is something to fix, right? (can be after migration is over or in 2077, but in principle, I mean) If only so that

> if that fails fall back to generating from the original

will actually save them.

Saving newly generated images will be useful in at least two instances: some broken thumbnails from 2015 and if we would to change the image generation code to eg. make thumbnails transparent etc.

andrew-nowak:
> > new versions created during projection aren't saved
> 
> This is something to fix, right? (can be after migration is over or in 2077, but in principle, I mean) If only so that
> 
> > if that fails fall back to generating from the original
> 
> will actually save them.

Yep! With the caveat being we'll lose this speed bonus again :( unless we do some smart optimisation elsewhere

PR #3495, opened 2021-10-05
Close some open image-loader streams

andrew-nowak:
## What does this change?

Cleans up some open file streams the projection endpoint code. These were found to be causing image-loader to keep too many open files during migrations.

This is not an exhaustive pass of unclosed streams - there may well be others lingering in the code base, but this change was sufficient to allow migration to complete without issue.

The number of open file handles (ballpark measurements possible with `sudo ls /proc/$(pgrep java)/fds | wc -l` on an image-loader instance) does not increase over time during heavy usage (eg. during a migration).

## Tested? Documented?
- [ ] locally by committer


jonathonherbert (approved):
Well spotted – wonder how many there are elsewhere 😬

Issue #2066, opened 2017-07-05
Colour wrong for images only tagged _via_ DCF EXIF


paperboyo:
Hello,

_This affects all frontend display of images sent straight from cameras that are set to record AdobeRGB, e.g. images from Tom Jenkins_

![grid-dcf-exif-bug](https://user-images.githubusercontent.com/6032869/27866175-43aca9a2-618d-11e7-934a-f0bd52e53ca7.jpg)

Camera makers often do not actually embed ICC profiles in camera JPEG images (I’ve tested two from Canon and Nikon). Instead they use [DCF EXIF tags to specify sRGB or AdobeRGB (pt. A2)](http://ninedegreesbelow.com/photography/embedded-color-space-information.html#embedded-info). Grid does not recognise this and treats these images as untagged. That is no problem for sRGB images as they are assumed to be sRGB anyway, but colour saturation/tonality suffers for images tagged as AdobeRGB. This affects both Grid’s thumbnails and all exports (including a master asset!).

For all images of an RGB colour model that do not explicitly have embedded ICC profiles, we should read both `0xa001 ColorSpace` and `0x0001 InteropIndex` [EXIF tags](http://www.sno.phy.queensu.ca/%7Ephil/exiftool/TagNames/EXIF.html) and explicitly embed AdobeRGB before performing any image operations when
`0x0001 InteropIndex` = `'R03' = R03 - DCF option file (Adobe RGB)`
and
`0xa001 ColorSpace` = `0x2 = Adobe RGB`
(this pertains only to untagged images; when there is a ICC profile present, it should always take precedence even if the above tags are set)

GraphicsMagick does not identify `0x0001 InteropIndex`, so unless we are [willing to contribute](https://sourceforge.net/p/graphicsmagick/feature-requests/14/#bf4a/da0e) and update it, we can look into using exiftool to do it. Or maybe metadata-extractor itself (we would most probably [need to import](https://github.com/guardian/grid/blob/master/image-loader/app/lib/imaging/FileMetadataReader.scala#L14) additional [EXIF directory](https://github.com/drewnoakes/metadata-extractor/blob/master/Source/com/drew/metadata/exif/ExifInteropDirectory.java) (?)).

Regards
Mateusz

paperboyo:
Yeah, so take affected 45c71141a66d7dcc61fd0e09eb90d8c1c1d091b2.

Read that there is no embedded ICC profile:
`fileMetadata:xmp:icc{}`

Then read
[...]

Issue #894, opened 2015-07-01
Apply correct colourspace conversion when generating on-the-fly preview


theefer:
Not sure how to do this as it's a different conversion path (via nginx module), but let's see if we can think of a way.

See also https://github.com/guardian/grid/issues/784

paperboyo:
Unfortunately, depends on https://github.com/libgd/libgd/issues/136 (https://bitbucket.org/libgd/gd-libgd/issues/136) :sadpanda:

paperboyo:
Hmm, I wouldn’t know, but is [this](https://aws.amazon.com/blogs/compute/resize-images-on-the-fly-with-amazon-s3-aws-lambda-and-amazon-api-gateway/) a possible replacement? ngx_http_image_filter_module not only messes up colour, but also chokes on big images. [AWS alternative](https://github.com/awslabs/serverless-image-resizing) uses [sharp](https://github.com/lovell/sharp) which in turn uses [libvips](https://github.com/jcupitt/libvips) which we [wanted to have a look at](https://github.com/guardian/grid/issues/155) separately as a replacement for GM. From what I can tell, it supports colour management. And promises to be super-fast.

paperboyo:
@akash1810 Because Firefox now supports WebP and because https://github.com/libgd/libgd/issues/136#issuecomment-252424409, maybe we could ask for WebPs? This might have an additional advantage of not having to use `optimisedPng` intermediary asset for cutouts.

paperboyo:
It looks like we would need another nginx module: [pagespeed](https://www.modpagespeed.com/doc/reference-image-optimize).

Although they say that [any conversion/resize removes colour profile](https://www.modpagespeed.com/doc/reference-image-optimize#strip_image_color_profile). Let’s hope they have no idea what are they talking about (which a non-mention of profile conversion prior to profile removal luckily suggests).

paperboyo:
Not sure, but maybe https://github.com/imazen/imageflow can help here? Or in https://github.com/guardian/grid/issues/155. Or – both?

paperboyo:
Another one based on VIPS: https://docs.imgproxy.net/#/configuration?id=avifwebp-support-detection

paperboyo:
Also VIPS? https://github.com/h2non/imaginary

paperboyo:
VIPS too: https://github.com/aws-solutions/serverless-image-handler

PR #3497, opened 2021-10-10
Read ExifInteropDirectory

paperboyo:
## What does this change?
This reads an additional Exif directory using metadata-extractor: ExifInteropDirectory. It contains only two tags. We only care about one of them: `InteropIndex`. Together with already read `Color Space`, it will help us detect images straight from cameras that are in AdobeRGB colour space, but have no embedded ICC profile. It will help us fix [this issue](https://github.com/guardian/grid/issues/2066).

We are ready to gather information needed to fix an issue, ~before we perform the migration.~ (voices of reason rightly recommended a code-frieze until we know how the migration goes…).

## Screenshots
image

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/FileMetadata.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/FileMetadata.scala
@@ -54,7 +54,7 @@ object FileMetadata {
     (__ \ "iptc").read[Map[String,String]] ~
     (__ \ "exif").read[Map[String,String]] ~
     (__ \ "exifSub").read[Map[String,String]] ~
-    (__ \ "exifInterop").read[Map[String,String]] ~
+    (__ \ "exifInterop").readNullable[Map[String,String]].map(_ getOrElse Map()) ~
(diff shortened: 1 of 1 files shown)


paperboyo:
I’m not convinced this is the best way: reading this otherwise useless directory. Maybe instead, we should perform all the needed sniffing and record the ooutcome in `colourModelInformation.DCF-EXIF-impliedAdobeRGB` (`true`, `false`) or whatever sounds better?

Pseudo-code for `FileMetadataReader.scala`:
- if `mimeType` = `Jpeg` AND
- if `ExifDirectoryBase.TAG_COLOR_SPACE` = `Undefined` AND
- if `ExifInteropDirectory."Interoperability Index"` = `Uknown (R03)`
  - set `colourModelInformation.DCF-EXIF-impliedAdobeRGB` to `true`
- else set `colourModelInformation.DCF-EXIF-impliedAdobeRGB` to `false` (or drop whole key? seems worse option)

And then one day, use this new value to apply AdobeRGB colour profile while creating crops (because we [currently can’t](https://github.com/guardian/grid/issues/894) for live preview).

PR #3514, opened 2021-10-29
Close unwanted scrolls

andrew-nowak:
## What does this change?

We don't currently have any code which closes scrolls when we are finished with them. During normal operation this isn't an issue - we're just constantly working through the open scroll with no issues.

When we get to the end of a migration, the current behaviour will see no results from a scroll request, so discard the current scroll ID and open a new one. This new scroll will also return no results, so the scroll ID will be discarded and a new opened, etc. The only problem with this behaviour is that the scrolls aren't automatically closed just because there are 0 results, so they can accumulate quite quickly. ES does enforce a limit on open scrolls so we don't see the quantity run away, but we should still close the scrolls when we discard the ID to keep ourselves clear of that limit and the associated errors.

When a migration reaches completion in the TEST environment, we no longer see many error messages from thrall complaining about reaching the limit of open scrolls in ES.

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/thrall/app/lib/MigrationSourceWithSender.scala
+++ b/thrall/app/lib/MigrationSourceWithSender.scala
@@ -8,7 +8,7 @@ import com.gu.mediaservice.lib.elasticsearch.InProgress
 import com.gu.mediaservice.lib.logging.GridLogging
 import com.gu.mediaservice.model.{MigrateImageMessage, MigrationMessage}
 import com.sksamuel.elastic4s.requests.searches.SearchHit
-import lib.elasticsearch.ElasticSearch
+import lib.elasticsearch.{ElasticSearch, ScrolledSearchResults}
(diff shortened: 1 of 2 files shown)


jonathonherbert (commented):
Code looks sensible! Cannot see evidence this has happened in TEST, although may be [looking in wrong place](https://logs.gutools.co.uk/s/editorial-tools/goto/b7d451cdba4578e788844d9b3c6f07c7), v. happy to pair to confirm all works as expected, or to take a look at the aftermath in logs if you kick off a migration.

twrichards (approved):
approving based on the code, but lets see if we prove it out in TEST before merging. feel free to merge now though if you need to get it into prod before the migration reaches the end

andrew-nowak:
We've run it in TEST now, and seen the number of scrolls stay low as expected

PR #3520, opened 2021-11-03
Add Snyk Github Action (GHA) to monitor all projects

tjsilver:
Co-authored-by: @SHession 

## What does this change?
This adds a Snyk Github Action to check all the manifests on merge to main. The `monitor` command is run and results reported to the Snyk.io dashboard. 

This integration is meant to improve on the original way projects were integrated into Snyk.io. There is only one `build.sbt` file in the project and testing with GHA reports 241 dependencies that are being checked, while the old integration reports 207. The version of SBT is 1.3.8 and the `sbt-dependency-graph` is not installed, so we're not sure how deeply into the dependency tree this is going, but it seems to be further than before.

Note: The `package-lock.json` files for `s3Watcher` and `oidc-provider` are also committed as part of this PR, to enable checking by Snyk.

All projects are checked for vulnerabilities and reported in the Snyk.io dashboard.

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -7,8 +7,8 @@ on:
   schedule:
     - cron: "0 6 * * *"
   push:
-    branches:
-      - main
+#    branches:
+#      - main
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/snyk.yml
```diff
@@ -0,0 +1,29 @@
+# This action runs every day at 6 AM and on every push
+# If the branch it's running on is main then it will run snyk monitor (reports vulnerabilities to snyk.io)
+# Otherwise it will run snyk test
```
andrew-nowak, line 3: is this comment still accurate? as far as I can tell the workflow is only run on pushes to main, and always runs snyk monitor

tjsilver: Good spot! Will update...

andrew-nowak (approved):

PR #3538, opened 2021-11-11
'redact' any `xmp` and `icc` fields where the length is greater than 5,000 characters

twrichards:
…to avoid `OutOfMemory` errors in downstream services (e.g. `thrall`) when parsing JSON payloads (note: the field is not dropped so one can still search for its presence just not its content).

This PR also refactors removal of long `icc` fields (introduced in https://github.com/guardian/grid/pull/2322) to follow this new 'redacting' approach (for consistency / simplicity).

We can ingest (and indeed migrate) images with giant values in their XMP fields, such as https://media.gutools.co.uk/images/1eeed60029abdc8e465298ff75d6c096d19e3359 (which took down `thrall` when attempting to migrate (due to parsing a huge JSON payload which contained the huge fileMetadata)

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [x] locally by committer


image-loader/app/lib/imaging/FileMetadataReader.scala
```diff
@@ -119,6 +119,16 @@ object FileMetadataReader {
       case (key, Some(value)) => key -> value
     }
   }
+
+  private val redactionThreshold = 5000
+  val redactionReplacementValue = s"REDACTED (value longer than $redactionThreshold characters, please refer to the metadata stored in the file itself)"
+  private def redactLongFieldValues(imageId:String)(props: Map[String, String]) = props.map {
+    case (fieldName, value) if value.length > redactionThreshold =>
+      logger.warn(s"Redacting '$fieldName' XMP field for image $imageId, as it's problematically long (longer than $redactionThreshold characters")
```
paperboyo, line 127: XMP or ICC? Or just drop the prefix from msg?

twrichards: ahh great spot, thanks @paperboyo - will fix now

jonathonherbert (approved):
Code looks good – ~happy to help test~ I can see this image on code @ `images/_doc/1eeed60029abdc8e465298ff75d6c096d19e3359?_source=true` (it was already present, and I could see this branch was already deployed, so I reuploaded), but can't see the `redacted` fields – wonder if I'm missing something?

paperboyo:
> can't see the `redacted` fields

For posterit, they are there, but under `fileMetadata`.

jonathonherbert:
Ah, yep, I can see them in the Grid response – I'd be interested in understanding why ES (via Cerebro) didn't give me that data from the above query! Ace 🎉

twrichards:
> Ah, yep, I can see them in the Grid response – I'd be interested in understanding why ES (via Cerebro) didn't give me that data from the above query! Ace 🎉
> images/_doc/1eeed60029abdc8e465298ff75d6c096d19e3359?_source=true

@jonathonherbert When I run this through TEST cerebro i get plenty of instances of the REDACTED message in both `icc` and `xmp` fields, within `fileMetadata` (note that the xmp ones are deep down, its not the whole xmp field that gets redacted just the deepest points, where they're > 5000 chars).

jonathonherbert:
@twrichards performing the same query this morning, I can see exactly what you're seeing – unsure if anything's changed! Anyhow, good stuff 👍

PR #3537, opened 2021-11-11
include syndication rights in projection

andrew-nowak:
## What does this change?

Fetch syndication rights data and include in aggregated data for image projections

## How can success be measured?

Syndication rights data is included with projected images

## Tested? Documented?
- [x] locally by committer
- [ ] locally by Guardian reviewer
- [x] on the Guardian's TEST environment
- [ ] relevant documentation added or amended (if needed)

PR #3544, opened 2021-11-17
FIX duplicate images (from new and old index) following re-queueing of images with syndicationRights

twrichards:
Originally syndicationRights were missing from 'migrated images' (fixed in https://github.com/guardian/grid/pull/3537). Rather than doing a whole new migration, we removed the `migratedTo` from the affected images to effectively re-queue them for migration. This had the side-effect of them appearing twice in the search API responses and the UI behaves funkily here. 

Fortunately when doing this re-queueing we added a field `esInfo.migration.failures.syndicationRightsMissing` on the affected documents in the old index, so this PR filters those out of the search results to ensure for this odd batch of images (approx 1.25 million) we don't show duplicates. 

This PR can be reverted once all those images have been re-migrated (in a few days).

### DIFF
--- a/media-api/app/lib/elasticsearch/ElasticSearch.scala
+++ b/media-api/app/lib/elasticsearch/ElasticSearch.scala
@@ -312,7 +312,10 @@ class ElasticSearch(
       case _ => List(imagesCurrentAlias)
     }
     val migrationAwareQuery = migrationStatus match {
-      case running: Running => filters.and(query, filters.mustNot(filters.term("esInfo.migration.migratedTo", running.migrationIndexName)))
+      case running: Running => filters.and(query, filters.and(
+        filters.mustNot(filters.term("esInfo.migration.migratedTo", running.migrationIndexName)),
+        filters.mustNot(filters.exists(NonEmptyList("esInfo.migration.failures.syndicationRightsMissing"))
+      )))
       case _ => query
     }
     val searchRequest = ElasticDsl.search(indexes) query migrationAwareQuery


media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -312,7 +312,10 @@ class ElasticSearch(
       case _ => List(imagesCurrentAlias)
     }
     val migrationAwareQuery = migrationStatus match {
-      case running: Running => filters.and(query, filters.mustNot(filters.term("esInfo.migration.migratedTo", running.migrationIndexName)))
+      case running: Running => filters.and(query, filters.and(
```
twrichards, line 315: should it be 
```suggestion
      case running: Running => filters.and(query, filters.or(
```

andrew-nowak: I'm pretty sure `and` is right (but I'm having trouble working through booleans this morning) - I think `or` would mean we'd get the duplicates on all the non-syndication related images. written out in a pseudocode 

```
query && (!migratedTo && !syndicationRightsMissing)
vs
query && (!migratedTo || !syndicationRightsMissing)
```

andrew-nowak (approved):

media-api/app/lib/elasticsearch/ElasticSearch.scala
twrichards: yeah checked on TEST - it works how we want with and, was just second guessing myself. im just waiting for `main` to redeploy to TEST to confirm that my replicating of the scenario was fine and that we see the duplicates.

twrichards: yep, i had replicated it correctly and this PR does indeed solve it, merging now...

PR #3546, opened 2021-11-17
FIX `AwaitingReviewForSyndication` queue bug

twrichards:
## RELIES ON: https://github.com/guardian/grid/pull/3550

Co-Authored-By: @andrew-nowak 

## What does this change?

https://trello.com/c/haSpms31/584-bug-grid-syndication-review-queue-shows-already-reviewed-images

The `leases` field is interpreted as a flat array, which has an unfortunate side effect of, images which have an active DenySyndicationLease AND any **expired** non-syndication lease, showing up in the syndication review queue (https://media.gutools.co.uk/search?syndicationStatus=review). 

A possible proper fix is to interpret this field as 'nested' (although this has some performance concerns) but this would require another migration (and could have said performance problems). 

**So instead this PR** introduces an alternate approach, using a new feature called [Runtime Fields](https://www.elastic.co/guide/en/elasticsearch/reference/master/runtime-search-request.html) (unlocked by https://github.com/guardian/grid/pull/3550) which allows us to run a _painless_ script on each of the initial query results to correctly assert that there is no **Active** Deny Syndication lease and then further filter on that.

Despite bumping `elastic4s` to latest there is no support for [Runtime Fields](https://www.elastic.co/guide/en/elasticsearch/reference/master/runtime-search-request.html) (even though they've been around for quite a few ES versions now) so instead we created `SearchHandlerWithRuntimeFields` (which uses the existing `SearchHandler` builder) and retroactively adds the top-level `runtime_mappings` fields to the outgoing request body - in the longer term we hope/plan to make an open source contribution to add support officially (unless `sksamuel` gets there first).

Happier syndication reviewers.

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchExecutions.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchExecutions.scala
@@ -9,13 +9,24 @@ trait ElasticSearchExecutions extends GridLogging {
 
   def client: ElasticClient
 
+  def executeAndLog[T, U](handler: Handler[T, U])(request: T, message: String)(implicit
+                                                                               functor: Functor[Future],
+                                                                               executor: Executor[Future],
+                                                                               manifest: Manifest[U],
+                                                                               executionContext: ExecutionContext,
+                                                                               logMarkers: LogMarker
+  ): Future[Response[U]] = {
+    implicit val implicitHandler = handler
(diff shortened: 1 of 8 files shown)


twrichards:
alas, the `totalHits` is used for more than just the count at the top, its also used to reserve squares on the screen (i.e. placeholders...
image
... so i'm not sure this tactical solution is going to work.

andrew-nowak (approved):

PR #3551, opened 2021-11-25
Add "download" as a "usages@platform" search filter suggestion

ochiengolanga:
## What does this change?
This PR adds _"download"_ among the list of available hints on the _"usages@platform"_ search filter when `image.record.download` config is set to `true`.

Since both media-api and kahuna use the `image.record.download` config, though optional it is now defined in the common config file and not media-api config file.

Given the `image.record.download` config is set to `true`
Then the _"download"_ should be included as a suggested _"usages@platform"_ filter and should filter images that have a recorded "download" usage.

demo

## Screenshots

Search filter hint

Search filter description

## Who should look at this?
@guardian/digital-cms 

## Tested? Documented?
- [X] locally by committer

### DIFF
--- a/kahuna/public/js/search/syntax/syntax.html
+++ b/kahuna/public/js/search/syntax/syntax.html
@@ -123,7 +123,7 @@
                 <gr-chip-example gr-filter-field="usages@platform" gr-example-search="print"></gr-chip-example>
             </dt>
             <dd class="advanced-search-example__explanation">
-                Filter by print, digital or download usage.
+                Filter by print, digital or other usage.
             </dd>
         </dl>
          <dl class="advanced-search-example">
--- a/kahuna/public/js/search/syntax/syntax.html
+++ b/kahuna/public/js/search/syntax/syntax.html
@@ -123,7 +123,7 @@
                 <gr-chip-example gr-filter-field="usages@platform" gr-example-search="print"></gr-chip-example>
             </dt>
             <dd class="advanced-search-example__explanation">
-                Filter by print or digital usage.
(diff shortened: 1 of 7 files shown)


paperboyo:
From usability perspective, this looks very sensible. We do not use downloads-as-usages at the Guardian (it’s behind the config we have switched off). Will this hide the suggested UI filter under these circumstances? (would be great)

ochiengolanga:
@paperboyo No it will not. I will reuse the `image.record.download` configuration to either hide or show the "download" suggestion to keep usages consistent.

paperboyo:
This works very well. Just one minor comment about help text.

kahuna/public/js/search/syntax/syntax.html
```diff
@@ -123,7 +123,7 @@ <h3 class="advanced-search-help__section__sub-heading">How to filter by Usage hi
                 <gr-chip-example gr-filter-field="usages@platform" gr-example-search="print"></gr-chip-example>
             </dt>
             <dd class="advanced-search-example__explanation">
-                Filter by print or digital usage.
+                Filter by print, digital or download usage.
```
paperboyo, line 126: Since this is dependent on config (that can be OFF and `download` platform won’t be available), maybe better to say `Filter by print, digital or other usage.`? Not a biggie, obvs.

ochiengolanga: Thanks @paperboyo for the suggestion. This is now resolved.

andrew-nowak (approved):

PR #3553, opened 2021-11-25
facilitate easier grouping of migration failures

twrichards:
…by no longer including imageId in the migration failure messages (since the failures are attached to image document in ES anyway, so we have that relationship)

Co-Authored-By: @andrew-nowak


twrichards:
Tested in `TEST`

andrew-nowak (approved):

PR #3561, opened 2021-12-02
Increase HTTP client timeout for projections to 300 seconds

andrew-nowak:
## What does this change?

Increase the HTTP client's timeout for calls to the projection endpoint to match server side increases in recent platform PRs.

Also requires setting config `play.ws.timeout.idle = 300 seconds` in Thrall config. Idle timeout can't be configured per request, so this keeps the request level timeout config to the default 120 seconds, and require the specific request to specify higher timeouts.

Docs: <https://www.playframework.com/documentation/2.6.x/ScalaWS#Configuring-Timeouts>

The final large TIFF that is still timing out can be migrated.

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/GridClient.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/GridClient.scala
@@ -10,6 +10,7 @@ import com.typesafe.scalalogging.LazyLogging
 import play.api.http.HeaderNames
 import play.api.libs.json.{JsArray, JsObject, JsValue, Json, Reads}
 
+import scala.concurrent.duration.{Duration, DurationInt}
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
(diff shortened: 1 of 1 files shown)


twrichards (approved):
I see you've tested on TEST so 👍 

will this not affect more than just projection though? since it's the generic async client in common-lib and so used all over the grid... could there be any downsides to that??

andrew-nowak:
@twrichards the requestTimeout in this PR is opt-in, and only the `getImageLoaderProjection` method is opting in. It does mean that any other usages of the grid client for projection will also have this timeout set (media-api's projection diff endpoint, and the admin-tools lambda), but without the `idleTimeout` config updated they'll still get timed out by that. Together this means that only projection requests from thrall will have this updated timeout.

Having thought about this, I think it's worth adding a comment line to `makeGetRequestAsync` that the requestTimeout parameter won't do anything unless the `idleTimeout` config is also increased, as I can imagine that being a future cause for confusion - I'll do that now

twrichards (approved):

PR #3568, opened 2021-12-07
Delete thumbnails after upload

andrew-nowak:
## What does this change?

Most generated assets are placed into a temp dir that is deleted after the request is finished, with the exception of the thumbnails. This means that thumbnails will stick around on disk after request
is finished. Normally this isn't a problem, as thumbnails are small but during a migration they will add up and can even fill up the disk, causing more problems.

Move the generated thumbnails inside this temp dir that is scoped to the lifetime of the request, so that they will also be cleared on completion

Free disk space does not decrease during operation of an image-loader instance (more noticeable on the projection deployment instances during a migration)

## Tested? Documented?
- [x] locally by committer


twrichards (approved):
NIce!!! 
image

PR #3572, opened 2021-12-08
workaround for the fact that sometimes fields like xmp"photoshop:DateCreated" have invalid/partial dates with extra chars e.g `19250200-01-01T00:00:00.000Z`

twrichards:
THIS IS GROSS

## Tested? Documented?
- [ ] locally by committer


twrichards:
I'd ideally like to close this PR in favour of fixing whatever is causing these extra chars, even though `exiftool` outputs without the extra chars

twrichards:
traced this back to the place the extra chars are introduced and worked around it there (see https://github.com/guardian/grid/pull/3575) ... so closing ...

PR #3575, opened 2021-12-10
workaround for the fact that sometimes fields like xmp"photoshop:DateCreated" have invalid/partial dates

twrichards:
Sometimes fields (such as xmp"photoshop:DateCreated") have 'partial' dates, which are padded with zeros, such as `19250200` or `19250000`, these don't parse properly as the date time formatters interpret them as a **really big** year. 

We noticed this during migration, because the existing logic would try to build an ISO date from the really big year, which when projected produced dates like `19250200-01-01T00:00:00.000Z` (which is technically valid, but way way in the future), which would then fail to write to ES (because it rejected this ludicrous date). We don't see this at ingestion, because ingestion filters out dates in the future because they indicate a mistake, for whatever reason, during projection (which executes the same code as ingestion) the value of `s3Source.metadata.objectMetadata.lastModified` is `None` and so no filtering takes place and the mega date isn't dropped like it is with ingestion.

## What does this change?
This PR contains a fairly explicit detection of this scenario where it's already attempting to cleanup metadata dates, with a test of course!

We should be able to successfully retry the 97 images which are failing to migrate because of the above.

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
@@ -156,11 +156,19 @@ object ImageMetadataConverter extends GridLogging {
   private def safeParsing[A](parse: => A): Option[A] = Try(parse).toOption
 
   private def cleanDateFormat = ISODateTimeFormat.dateTime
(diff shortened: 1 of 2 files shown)


twrichards:
✅ tested in `TEST`
image

andrew-nowak (approved):

PR #3580, opened 2021-12-13
when we encounter malformed `fileMetadata` fields, retry the ES insert with the problematic fields dropped

twrichards:
During migration we experienced...
image
... this happens because `fileMetadata` fields are mapped on the fly (using a template in Mappings file) based on the shape of the first value it encounters for a field, and rejects subsequent inserts which don't conform to that shape. In the above case, the existing entries in ES which have `fileMetadata.xmp.avm:Distance` have a `Notes` property within...
image, whereas the 3 files failing just have a string array value directly on `avm:Distance` and thus ES rejects them as it can't index. Unfortunately `https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-malformed.html` can't be used on this type of field (keyword). 

**In this PR we attempt to parse the above error and drop the offending fields and retry the insert.**

### DIFF
--- a/thrall/app/lib/elasticsearch/ElasticSearch.scala
+++ b/thrall/app/lib/elasticsearch/ElasticSearch.scala
@@ -2,7 +2,7 @@ package lib.elasticsearch
 
 import akka.actor.Scheduler
 import com.gu.mediaservice.lib.ImageFields
-import com.gu.mediaservice.lib.elasticsearch.{ElasticSearchClient, ElasticSearchConfig, ElasticSearchExecutions, Running}
+import com.gu.mediaservice.lib.elasticsearch.{ElasticSearchClient, ElasticSearchConfig, ElasticSearchError, ElasticSearchException, ElasticSearchExecutions, Running}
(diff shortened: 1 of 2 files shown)


twrichards:
Closing this as it was getting way too sketchy trying to parse from the error message string the offending field names and excluding them in a reliable way. Unfortunately `ElasticError` doesn't provide a nice list of fields, it just comes as part of the error message string with various different prefixes and suffixes (which are subject to change without warning).

As such we're going to solve the problem another way...

- In order to fix these 3 images and subsequently 'complete' the current migration, we'll release a temporary change to explicitly drop `fileMetadata.xmp.avm:Distance`, reattempt migration on the 3 images, then revert the PR and 'complete' the migration.
- Before the next migration, we will add a permanent change which will replace any `.` in xmp field names (e.g. `avm:Distance.Notes`) with underscores or some other kind of dot. ES will then create the mapping on the fly and this will treat them as entirely different fields (e.g. `avm:Distance` and `avm:Distance_Notes`) to prevent these errors from ever occuring. This can't be done now though due to the existing mappings in ES.

PR #3583, opened 2021-12-16
add original-source usage-rights

abdelrahmansd:
## What does this change?
Add a new usage Rights called "Original Source" 

## Tested? Documented?
- [ ] locally by committer


paperboyo:
Hi, @abdelrahmansd. I have spoken with our editors and they are using two other Usage Rights for this purpose (Handout and Guardian Witness), so could we have this behind a BBC-only config, please? (there are too many to choose from already and users are getting confused) Thanks!

paperboyo:
Oh, @andrew-nowak mentions [this existing config](https://github.com/guardian/grid/blob/861bb12210e9f08ad09a70d39acd58d5d1567527/common-lib/src/main/resources/application.conf#L61), which may already apply, so we may be good. We will put on TEST.

andrew-nowak (approved):

PR #3584, opened 2021-12-16
Upgrade Play to 2.8.11, and SBT to 1.5.7.

jonathonherbert:
_Co-authored-by: @andrew-nowak_ 

## What does this change?

Upgrades Play to 2.8.11, and SBT to 1.5.7.

We have an application that works as before.

## This is long! Where do I start?

We've tried to separate discrete edits into different commits, to tell a story about what changed and when. These should be largely self-explanatory, but we're happy to provide more detail if needed.

## Things worth testing

Tested in CODE:

- ✅ Parts of the application that rely on the scheduler (resumed a migration, and paused again)
- ✅ Stream processing (e.g. Thrall)
- ✅ Logging, including markers, where that's been changed

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -18,7 +18,7 @@ val commonSettings = Seq(
 
   Test / testOptions ++= Seq(Tests.Argument(TestFrameworks.ScalaTest, "-o"), Tests.Argument(TestFrameworks.ScalaTest, "-u", "logs/test-reports")),
   libraryDependencies ++= Seq(
-    "org.scalatestplus.play" %% "scalatestplus-play" % "3.1.2" % Test,
+    "org.scalatestplus.play" %% "scalatestplus-play" % "5.1.0" % Test,
     "org.mockito" % "mockito-core" % "2.18.0" % Test,
     "org.scalamock" %% "scalamock" % "5.1.0" % Test
   ),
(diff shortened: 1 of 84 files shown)


build.sbt
```diff
@@ -103,26 +104,27 @@ lazy val commonLib = project("common-lib").settings(
...
     // i.e. to only log to disk in DEV
     // see: https://logback.qos.ch/setup.html#janino
     "org.codehaus.janino" % "janino" % "3.0.6",
-    "com.typesafe.play" %% "play-json-joda" % "2.6.9",
+    "com.typesafe.play" %% "play-json-joda" % "2.9.2",
     "com.gu" %% "scanamo" % "1.0.0-M8",
     "com.fasterxml.jackson.core" % "jackson-databind" % "2.9.10.7",
-    ws
+    // Necessary to have a mix of play library versions due to scala-java8-compat incompatibility
+    "com.typesafe.play" %% "play-ahc-ws" % "2.8.9"
```
SHession, line 119: I think this may be possible to avoid with a Scala version upgrade, however there may be barriers to that.

SHession (approved):
Great stuff! I have reviewed the code but not done any manual testing, as I believe this has been well covered. LGTM.

PR #3586, opened 2021-12-20
Revert "Upgrade Play to 2.8.11, and SBT to 1.5.7."

jonathonherbert:
Reverts guardian/grid#3584, which seems to be causing issues for users in PROD.


SHession (approved):

twrichards:
Closing this since we've concluded it isn't the cause of the problem. Will edit this comment once PR is ready which does solve the (long standing) problem.

PR #3587, opened 2021-12-20
Ensure `images` index is not aliased with `Images_Current` at start-up

twrichards:
Although we did test https://github.com/guardian/grid/pull/3560 and https://github.com/guardian/grid/pull/3563 we did not attempt a redeploy once the `images` index was no longer aliased by `Images_Current` and therefore missed the fact that this happens during startup, due to some outdated code. This meant `Images_Current` pointed to two places and so writes to ES would fail and API would return double IDs (i.e. breaking the grid quite significantly).

## What does this change?
Fix the check which happened in the `ensureAliasAssigned` function so instead now it checks to ensure the `Images_Current` alias points somewhere - previously it checked explicitly for the presence of the `images` index, which is no longer correct since we've completed our first migration (and the `Images_Current` alias points to a different index now).

Also renamed poorly named functions `ensureIndexExists` to `createIndexIfMissing` and `ensureAliasAssigned` to `ensureIndexExistsAndAliasAssigned`

Thrall should start-up properly and no add an additional `Images_Current` alias and the grid should run smoothly throughout/after deployments.

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchClient.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchClient.scala
@@ -44,11 +44,11 @@ trait ElasticSearchClient extends ElasticSearchExecutions with GridLogging {
   }
 
   //TODO: this function should fail and cause healthcheck fails
-  def ensureAliasAssigned() {
+  def ensureIndexExistsAndAliasAssigned() {
     logger.info(s"Checking alias $imagesCurrentAlias is assigned to index…")
     val indexForCurrentAlias = Await.result(getIndexForAlias(imagesCurrentAlias), tenSeconds)
     if (indexForCurrentAlias.isEmpty) {
-      ensureIndexExists(initialImagesIndex)
+      createIndexIfMissing(initialImagesIndex)
       assignAliasTo(initialImagesIndex, imagesCurrentAlias)
       waitUntilHealthy()
     }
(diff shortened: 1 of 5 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/ElasticSearchClient.scala
```diff
@@ -46,9 +46,10 @@ trait ElasticSearchClient extends ElasticSearchExecutions with GridLogging {
   //TODO: this function should fail and cause healthcheck fails
   def ensureAliasAssigned() {
     logger.info(s"Checking alias $imagesCurrentAlias is assigned to index…")
-    if (getCurrentAlias.isEmpty) {
+    val indexForCurrentAlias = Await.result(getIndexForAlias(imagesCurrentAlias), tenSeconds)
+    if (indexForCurrentAlias.isEmpty) {
       ensureIndexExists(initialImagesIndex)
```
SHession, line 51: This might read slightly better if we renamed `ensureIndexExists` to `createIndexIfMissing`

twrichards: yeah good idea, will do that now

twrichards:
✅ tested in TEST

SHession (approved):
Changes LGTM and this has been deployed to a test environemnt.

PR #3599, opened 2022-01-13
Do not log to stdout when running locally

andrew-nowak:
## What does this change?

There is a high volume of output when running the grid locally (due to
running multiple microservices in the same session), which are difficult
to grok. In addition (anecdotally) running with stdout logging on
reduces the ability to interact with the sbt shell. I found it very
difficult to get sbt to respond to my ctrl-C inputs, which can be a
symptom of the terminal emulator struggling to keep up with the volume
of output.

This PR disables stdout logging when running locally (ie. when stage ==
DEV), which I have found (anecdotally) to significantly improve
responsiveness from sbt.

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/common-lib/src/main/resources/logback.xml
+++ b/common-lib/src/main/resources/logback.xml
@@ -37,18 +37,13 @@
     <logger name="application" level="INFO" />
     <logger name="request" level="INFO" />
 
-    <if condition='property("STAGE").contains("DEV")'>
-      <then>
-        <root level="INFO">
-          <appender-ref ref="LOGFILE"/>
-        </root>
-      </then>
-      <else>
-        <root level="INFO">
-          <appender-ref ref="LOGFILE"/>
+    <root level="INFO">
+      <appender-ref ref="LOGFILE"/>
+      <if condition='!property("STAGE").contains("DEV")'>
+        <then>
           <appender-ref ref="ASYNCSTDOUT"/>
-        </root>
-      </else>
-    </if>
+        </then>
+      </if>
+    </root>
 
 </configuration>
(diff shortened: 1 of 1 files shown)


common-lib/src/main/resources/logback.xml
```diff
@@ -37,9 +37,18 @@
...
+          <appender-ref ref="LOGFILE"/>
+        </root>
+      </then>
+      <else>
+        <root level="INFO">
+          <appender-ref ref="LOGFILE"/>
+          <appender-ref ref="ASYNCSTDOUT"/>
+        </root>
+      </else>
+    </if>
```
twrichards, line 52: could this be moved outside the condition, to avoid repetition and so its clearer that we still always have a log file? something like (written in GH UI, so might not be valid)...
```suggestion
<root level="INFO">
    <appender-ref ref="LOGFILE"/>
    <if condition='!property("STAGE").contains("DEV")'>
      <then>
          <appender-ref ref="ASYNCSTDOUT"/>
      </then>
    </if>
 </root>
```

twrichards (approved):
👏  nice!
approved, but suggested an alternative definition to achieve the same

common-lib/src/main/resources/logback.xml
andrew-nowak: Cheers, tried it out locally and it's working great so added as a commit

twrichards: probs need to check it on TEST just to make sure the negation actually works and it still logs to stdout

andrew-nowak: deployed to TEST, logs are still coming through as expected there

twrichards: nice

twrichards (approved):
🚢

PR #3604, opened 2022-01-18
Better colourModelInformation for JPEGs

jonathonherbert:
## What does this change?

`colourModelInformation` metadata (`hasAlpha`, `bitsPerSample`) wasn’t being returned for JPEGs. But we know them JPEGs! They cannot have transparency and they are always only 8 bit per channel (someone, somewhere saw a 10-bit JPEG once, but nobody ever saw that person…). So we fill them in hard.

It renames `GRAYSCALE` to more British human-readable `Greyscale` in anticipation of a chip. It cleans values for `bitsPerSample`, so that they are more consistent.

Addresses pts. 2–6 in https://github.com/guardian/grid/issues/2594. So, in truth… Fixes https://github.com/guardian/grid/issues/2594.

Slightly more correct colour information for JPEGs.
Future vision from @paperboyo: One day we will be able to search for `+is:Cutout` in chips. For now, we can add this as an alias search…

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/image-loader/app/model/Uploader.scala
+++ b/image-loader/app/model/Uploader.scala
@@ -232,7 +232,7 @@ object Uploader extends GridLogging {
 
   private def toFileMetadata(f: File, imageId: String, mimeType: Option[MimeType]): Future[FileMetadata] = {
     mimeType match {
-      case Some(Png | Tiff) => FileMetadataReader.fromIPTCHeadersWithColorInfo(f, imageId, mimeType.get)
+      case Some(Png | Tiff | Jpeg) => FileMetadataReader.fromIPTCHeadersWithColorInfo(f, imageId, mimeType.get)
       case _ => FileMetadataReader.fromIPTCHeaders(f, imageId)
     }
   }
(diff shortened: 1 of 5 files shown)


image-loader/app/model/Uploader.scala
```diff
@@ -232,7 +232,7 @@ object Uploader extends GridLogging {
 
   private def toFileMetadata(f: File, imageId: String, mimeType: Option[MimeType]): Future[FileMetadata] = {
     mimeType match {
-      case Some(Png | Tiff) => FileMetadataReader.fromIPTCHeadersWithColorInfo(f, imageId, mimeType.get)
+      case Some(Png | Tiff | Jpeg) => FileMetadataReader.fromIPTCHeadersWithColorInfo(f, imageId, mimeType.get)
       case _ => FileMetadataReader.fromIPTCHeaders(f, imageId)
```
andrew-nowak, line 236: if we're happy with this PR, a note to remember to clean up as the `case _` branch here is dead

twrichards: not dead, if `mimeType` is a `None`

twrichards (approved):
once @paperboyo has finished his testing, this LGTM based on the code

image-loader/app/model/Uploader.scala
paperboyo: `fromIPTCHeaders` might as well be dead? But let’s pretend we don’t know, because I don’t :-).

PR #3605, opened 2022-01-19
Version clientside assets

andrew-nowak:
## What does this change?

Adds the sbt-digest plugin, which creates an md5 of each asset at build time. This allows Play to **automatically** switch to serving versioned assets from its controllers! (We'd already halfway opted into the behaviour by using the 'versioned' assets controllers and reverse routes, but without the digest Play falls back to treating them as unversioned). Now that each version has a unique URL, Play will set the Cache-control header to `Cache-Control: public, max-age=31536000, immutable`.

Also adds the sbt-gzip plugin, which will gzip each asset. This is a smaller optimisation as I think cloudfront is already doing compressing of assets for us, do we need/want compression for the hop between grid -> cloudfront?

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -152,7 +152,9 @@ lazy val imageLoader = playProject("image-loader", 9003).settings {
   )
 }
 
-lazy val kahuna = playProject("kahuna", 9005)
+lazy val kahuna = playProject("kahuna", 9005).settings(
+  pipelineStages := Seq(digest, gzip)
+)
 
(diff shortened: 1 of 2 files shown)


davidfurey (commented):
Looks good to me.  Why is it a draft?

andrew-nowak:
thanks! I'd overlooked that it was still in that stage, I've made it ready for review now

davidfurey (approved):

PR #3607, opened 2022-01-21
Mark all records as processed even if parsing failed

andrew-nowak:
## What does this change?

Mark unparseable kinesis messages as processed before they are dropped. We need to do this otherwise the kinesis client can't proceed checkpoints beyond such an unparseable message, and all messages from that point will be replayed upon every reboot/redeploy of Thrall.  Parseable kinesis messages are marked as processed after Thrall has attempted to process them _even if the processing fails_ - which is why we haven't seen this more widely.

Badly formatted messages are still dropped, but kinesis checkpointing continues past the bad message.

## Tested? Documented?
- [ ] locally by committer


twrichards (approved):
Amazing detective work 🕵️ (as per usual @andrew-nowak )

twrichards:
A wider question, do we have alarms on unparseable/dropped messages?

andrew-nowak:
@twrichards I don't think we do no :( I also thought I remembered some sort of a DLQ for thrall which I can't find now, only one for s3watcher

PR #3616, opened 2022-01-24
rename OriginalSource UsageRights name

abdelrahmansd:
## What does this change?
Rename OriginalSource UsageRights name from _"OriginalSource"_ to _"Original Source"_

## Tested? Documented?
- [x] locally by committer


paperboyo (approved):
I think even I can understand what’s going on here, so will approve. Even more so, because  we are not using it here ;-).

PR #3617, opened 2022-01-24
Document migration

andrew-nowak:
## Tested? Documented?
- [ ] locally by committer


docs/05-migration/01-about.md
```diff
@@ -0,0 +1,82 @@
+# Migration
+
+Sometimes also described as reindexing or reingestion.
+
+Image migration is an optional process in the Grid that will create a new index
+in the Elasticsearch cluster containing an up-to-date representation of all
+known images. In simple cases, this is akin to the [Elasticsearch Reindex
+API][es-reindex], i.e. moving all documents to a new index using the latest
+version of the mapping, but migration also contains improvements to ensure a
+seamless transition and improved metadata.
```
twrichards, line 10: I would be hesitate to liken it to reindex API, I think the key thing to convey is that it's rebuilding the ES index from the original image metadata (in S3) AND the user metadata (in DynamoDB) and NOT from the existing ES representation.
```suggestion
known images. Unlike the [Elasticsearch Reindex
API][es-reindex] (i.e. copying all documents to a new index using the latest
version of the mapping) the migration process rebuilds the document  from the original image metadata (in S3) AND the user metadata (in DynamoDB) and NOT from the existing ES representation. 
Migration also contains improvements to ensure a seamless transition and improved metadata.
```

docs/05-migration/01-about.md
```diff
@@ -0,0 +1,82 @@
...
+
+## How it works
+
+### High level
+
+While a migration is [in progress](#migration-status-flag), Thrall will
+repeatedly query Elasticsearch for images which have [not yet been
+migrated](#image-migration-record). Each image ID which is found will be
+[projected](#projection), and that projection will be entered into a
+low-priority queue for processing by Thrall when no other updates need to be
```
twrichards, line 36:
```suggestion
low-priority queue (UI interactions are highest priority, then 'automation' like usages, RCS etc, then lowest priority is migration) for processing by Thrall when no other updates need to be
```

twrichards (approved):
Awesome work @andrew-nowak  (a few suggestions, but not blocking and happy for you disagree/ignore)

Would be lovely to have a review from someone else in @guardian/digital-cms , who hasn't worked on this project.

docs/05-migration/01-about.md
paperboyo: I concur: great work, thank you!

Regarding similarities/differences to reindexing, I think this needs to die (or be adjusted):
https://github.com/guardian/grid/blob/main/docs/05-reingestion/01-how-to.md

Similarly, not sure how much of this still applies or how it should change because of migration?
https://github.com/guardian/grid/tree/main/scripts#reindex
&
https://github.com/guardian/grid/tree/main/scripts#updatemapping

paperboyo: Thanks (sorry for not spotting binning!)

SHession (approved):
This is brilliant documentation, @andrew-nowak, really thorough and it will no doubt prove indispensable in the future. 👏

I can only think of one addition we may want to include, which is a point about performance issues when a migration tasks takes a long time. I.e. a low priority queue job takes a long time to finish, causing users changes to be backed up. I'm not sure if there is a way to mitigate but might be worth a mention.

Issue #155, opened 2014-10-26
Evaluate libvips as alternative for image manipulation/cropping w better performance


theefer:
http://www.vips.ecs.soton.ac.uk/index.php?title=Libvips

May be able to take in image stream?

andrew-nowak:
investigating in #4068

PR #3621, opened 2022-01-25
Switch all imaging operations to ImageMagick

paperboyo:
## What does this change?
[A Sense of History](https://www.imdb.com/title/tt0105355/): Grid started by only using GraphicsMagick, probably because it was thought to be both more performant and more secure. In time, as we added more file formats, its limitations started to show, so we’ve introduced ImageMagick and decided to use them [for some operations](https://github.com/guardian/grid/pull/2579). ~Today we have found a perfectly fine JPEG, a format still serviced by GraphicsMagick, that wouldn’t upload. ImageMagick has no problems with this file.~ (EDIT: this was our fault and fixed now in https://github.com/guardian/grid/pull/3622)

This PR switches all imaging and metadata operations to run off ImageMagick. The reasons are:
- IM is more often more reliable (Lab TIFFs, some ICC profiles etc)
- GM has fallen a bit behind because an undue stress is put on a sole maintainer
- since we have IM anyway, security is not a concern. Or at least – not any more
- performance gains, if any, are not so well publicised for us to be able to refer to them here
- we [should be using libvips](https://github.com/guardian/grid/issues/155) anyway!

~A [particular file](https://trello.com/c/IxKWJhU4/2578-gm-fails-on-some-icc-profiles) uploads correctly. There are no regressions in any imaging and metadata operations relying on GM and IM.~
[EDIT: this only presents a case for removing GM and keeping only IM]

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/imaging/ImageOperations.scala
@@ -249,24 +249,20 @@ object ImageOperations {
   val optimisedMimeType = Png
   def identifyColourModel(sourceFile: File, mimeType: MimeType)(implicit ec: ExecutionContext): Future[Option[String]] = {
     // TODO: use mimeType to lookup other properties once we support other formats
+    val op = new IMOperation()
+    val formatter = format(op)("%[colorspace]")
+    val withSource = addDestImage(formatter)(sourceFile)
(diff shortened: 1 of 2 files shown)


paperboyo:
This is not urgent any more. But is a valuable lesson in how to get rid of GM and move to IM completely. But, maybe, more sensible would be to go to libvips…

PR #3632, opened 2022-02-08
Identify raw images

davidfurey:
## What does this change?

This improves the mimetype detection so that DNG and CR2 files are correctly identified.  Previously these were classified as Tiffs, and then later failed to import, with users seeing an unfriendly error message and developers getting a PagerDuty alert.

### Before:
image (7)

### After
image (6)

Fewer PagerDuty alerts

## Tested? Documented?
- [x] locally by committer


paperboyo:
Tested and it works great! Probably should be revisited (to remove magic identification) after this is resolved upstream: https://github.com/drewnoakes/metadata-extractor/issues/75.

It correctly prevents uploading of `.CR2` and DNG images. If we would want to get magick for more RAW files, probably [Sony](https://github.com/drewnoakes/metadata-extractor-images/blob/master/arw/) and [Nikon](https://github.com/drewnoakes/metadata-extractor-images/tree/master/nef) should be next (Sony bombs, Nikon uploads a 165px thumbnail).

Regarding future, better, support for DNGs, we **could** extract the preview JPEG and upload that **after checking it’s the same dimensions as the RAW file**. This is better tan trying to develop RAW files ourselves, because they often come with proprietary developmnet instructions, so using JPEG preview should get us closer to artists’ intent. But even that wouldn’t be 100% solution as previews may be out of date and even native resolution preview generated by Adobe apps is a highly compressed JPEG :-(.

image-loader/conf/org/apache/tika/mime/custom-mimetypes.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<mime-info>
+  <mime-type type="image/x-raw-canon">
+    <_comment>Canon raw image</_comment>
+    <magic priority="50">
```
andrew-nowak, line 5: nit: This has the same priority as tiff files which I think should also match this magic. Should we increase this priority to make sure that whatever tika uses to tiebreak doesn't accidentally switch back to identifying these as tiffs?

image-loader/app/lib/imaging/MimeTypeDetection.scala
```diff
@@ -27,6 +30,14 @@ object MimeTypeDetection extends GridLogging {
     case Failure(_: Throwable) => Left(new UnsupportedMimeTypeException(FALLBACK))
   }
 
+  // DNG files look like Tiff files, but include a DNGVersion tag (0xC612)
+  // See https://www.adobe.com/content/dam/acom/en/products/photoshop/pdfs/dng_spec_1.4.0.0.pdf
+  private def isDng(file: File) = {
+    val metadata = TiffMetadataReader.readMetadata(file)
+    val directory = metadata.getFirstDirectoryOfType(classOf[ExifIFD0Directory])
+    directory.containsTag(0xC612)
```
andrew-nowak, line 38: nit: I think technically this could NPE (if somehow there wasn't a 0th ifd on the tiff), other places in the code we've  checked that it's not null

andrew-nowak (approved):
working on test 👍

image-loader/conf/org/apache/tika/mime/custom-mimetypes.xml
davidfurey: Sounds reasonable.  I think custom-mimetypes have higher priority already, but no harm in increasing this.

andrew-nowak (approved):

paperboyo:
Thanks!

> Happy to look at this when I get a chance, just send me some example images.
Just download any from the folders I linked to. eg. [this](https://github.com/drewnoakes/metadata-extractor-images/blob/master/arw/Sony%20ILCE-7M3%20(A7M3).arw) and [that](https://github.com/drewnoakes/metadata-extractor-images/blob/master/nef/NikonD200.nef).

> I'm reluctant to attempt to extract JPEG info from a DNG since I don't think this is a workflow we want to encourage. I'd rather we displayed a sensible error message.

Yeah. Thing is, the current state of industry prevents us from always doing what’s best: proprietary Adobe, Sony or any other metadata development commands cannot be interpreted by open source software, because the renderers are proprietary too (not even commercial software can read competitors’ data).

The error message could read something like: `This image is a raw file that cannot be reliably interpreted by Grid. Please convert it to supported file format preserving all metadata.`. But this would be useful only for users who will understand it and can act upon it (either themselves or asking upstream). 

One alternative would be to sniff for acceptable previews (that have developments baked-in). Another would be whatever dcraw does: probably just renders raws using vanilla settings, although it may take some metadata into account, I don’t know (I would expect at least white balance to be interpreted correctly). Latter alternative would render raws/dngs different to how they were saved from raw developing software if they indeed were (although we could exclude those with proprietary-specific metadata?). In any case, this seems involved. Preview ingest less so than trying to be clever about raw development. Users should upload developed files saved as highest quality JPEGs…

davidfurey:
For future reference, I think the Canon raw magic came from http://lclevy.free.fr/cr2/#key_info / https://github.com/lclevy/libcraw2/blob/master/docs/cr2_poster.pdf

But I might have got it from somewhere else and then cross-referenced with this resource

PR #3633, opened 2022-02-08
Es/S3 sync checker

andrew-nowak:
## What does this change?

Adds new functionality to Thrall to run through elasticsearch and s3 and pick up images which are in one but not the other and output them as warning logs.

There is one part not covered by this: because images in s3 are stored behind prefixes, it is possible that an image is stored behind a nonstandard prefix (eg. an image at `x/y/z/xyzimage` should not have been created by the grid). Traversing the entire bucket to search for all possible nonstandard prefixes is difficult to do with the s3 listobjects api, so I've left out of this functionality.

We can find all the discrepancies between ES and S3.

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/thrall/app/lib/SyncChecker.scala
+++ b/thrall/app/lib/SyncChecker.scala
@@ -84,14 +84,10 @@ class SyncChecker(
       s3Ids <- s3IdsFetch
     } yield {
       val extrasInS3 = s3Ids diff esIds
-      if (extrasInS3.nonEmpty) {
-        logger.warn(logMarker, s"SyncChecker found ${extrasInS3.size} extra images in S3 under prefix $s3Prefix: ${extrasInS3.mkString(",")}")
-      }
+      batchedLogs(s"extra images in S3 under prefix $s3Prefix", extrasInS3)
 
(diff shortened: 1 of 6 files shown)


thrall/app/lib/SyncChecker.scala
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  private def createStream() = Source.cycle(() => prefixes.toIterator)
+    .throttle(1, per = 5.seconds)
+    .filter(_ => {
+      es.migrationStatus match {
+        case InProgress(_) => false
+        case _ => true
+      }
+    })
```
twrichards, line 122: if there's a `filterNot`, how about... 
```suggestion
    .filterNot(_ => es.migrationStatus.isInstanceOf[InProgress])
```

otherwise, how about... 
```suggestion
    .filter(_ => !es.migrationStatus.isInstanceOf[InProgress])
```

twrichards (approved):
Amazing work 👏 

One minor comment (which might not work, but looks more readable to me).

PR #3638, opened 2022-02-09
fix usage flakiness

twrichards:
Co-authored-by: @andrew-nowak 

Following on from #3637 - now that we have composer usages coming through we need to address the 'flakiness' that usages have long experienced.

## What does this change?
Here we set the `ConsistentRead` flag to `true` for the three different queries done against the Usage table (to ensure we can safely read immediately after writing). This 'fix' is inspired by an old investigation by @akash1810 @jonathonherbert and @marialivia16 as detailed in https://trello.com/c/adNuLqH0/553-grid-usages-not-reliable#comment-5c542db00f90282a959055f2 
image

Hopefully more reliable/accurate usages 🤞 

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/usage/app/model/UsageTable.scala
+++ b/usage/app/model/UsageTable.scala
@@ -1,6 +1,6 @@
 package model
 
-import com.amazonaws.services.dynamodbv2.document.spec.{DeleteItemSpec, UpdateItemSpec}
+import com.amazonaws.services.dynamodbv2.document.spec.{DeleteItemSpec, QuerySpec, UpdateItemSpec}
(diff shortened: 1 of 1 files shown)


twrichards:
So it turns out that the query where we really need the `ConsistentRead` (the other two we just did for good measure) uses a Global Secondary Index, for which `ConsistentRead` is not available (as per https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.ReadConsistency.html). 
image

So closing this PR and we'll look to address the eventual consistency problem a different way.

PR #3639, opened 2022-02-09
ensure accurate usages stored to ES

twrichards:
Co-authored-by: @andrew-nowak 

Following on from our failed attempt (https://github.com/guardian/grid/pull/3638) to work around DynamoDB eventual consistency (given we query immediately after writing) this is an alternative approach...

## What does this change?
We now check the results returned from the query of all usages for a mediaId (done when building the `UsageNotice` list that goes off to thrall/ElasticSearch) and if the just written MediaUsage is present OR missing we log accordingly. If it is missing however, we now supplement the query results with the just written MediaUsage.

Hopefully more reliable/accurate usages 🤞

## Screenshots
image

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/usage/app/lib/UsageNotifier.scala
+++ b/usage/app/lib/UsageNotifier.scala
@@ -15,10 +15,19 @@ import scala.concurrent.ExecutionContext.Implicits.global
 class UsageNotifier(config: UsageConfig, usageTable: UsageTable)
   extends ThrallMessageSender(config.thrallKinesisLowPriorityStreamConfig) with GridLogging with MessageSubjects {
 
-  def build(mediaId: String) = Observable.from(
-    usageTable.queryByImageId(mediaId).map((usages: Set[MediaUsage]) => {
-      val usageJson = Json.toJson(usages.map(UsageBuilder.build)).as[JsArray]
-      UsageNotice(mediaId, usageJson)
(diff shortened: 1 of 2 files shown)


jonathonherbert (approved):
Logging a great way of making this really clear 💯   

Just to document other possible ways usages might become inaccurate – could two concurrent usages result in inaccurate reporting? E.g. if usage A is already written, and usage B and C are written concurrently, some variation of A & (B | C) will be given to ES.

(Concurrently usages are, I think, really really unlikely?)

twrichards:
> Logging a great way of making this really clear 💯
> 
> Just to document other possible ways usages might become inaccurate – could two concurrent usages result in inaccurate reporting? E.g. if usage A is already written, and usage B and C are written concurrently, some variation of A & (B | C) will be given to ES.
> 
> (Concurrently usages are, I think, really really unlikely?)

We did consider the possibility of concurrent writes, but considered it too unlikely and extremely hard to guard against, doing what we've done in this PR doesn't worsen the concurrent writes scenario you describe, so we've gone ahead and merged.

jonathonherbert:
Absolutely not a blocker to this work, hence the approval – just thinking about documenting in this PR in case somebody's digging about. That's now done 😁

PR #3643, opened 2022-02-10
[usage] improve the accurate/inaccurate read-after-write detection logic

twrichards:
## What does this change?
In #3639 we introduced some logging to detect the ratio of accurate vs inaccurate read-after-writes before sending to thrall/ElasticSearch. Given some of our concerns around the fact only the `.equals` method is overridden in `MediaUsage.scala` (and not `.hashCode` as it should always be when overriding `.equals`), in this PR we add an additional log line to tell us whether the read-after-write does in fact contain the a usage with the same `usageId` but the `.contains` (which we think uses `.hashCode`) considers it not present in the read result (probably due to something else being different).

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/usage/app/lib/UsageNotifier.scala
+++ b/usage/app/lib/UsageNotifier.scala
@@ -1,7 +1,7 @@
 package lib
 
 import com.gu.mediaservice.lib.aws.{ThrallMessageSender, UpdateMessage}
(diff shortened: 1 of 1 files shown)


twrichards:
this read-after-write consistency was largely a red herring, so closing this out of date PR

Issue #1900, opened 2016-05-16
Unicode in search terms disappears from search box


paperboyo:
I think asciifolding filter somehow affects URL<>searchbar communication. Or something.

Try pasting e.g. `czesław miłosz` into searchbox or clicking this link: https://media.gutools.co.uk/search?query=czes%C5%82aw%20mi%C5%82osz and then reloading browser.

Modifying the query in searchbox makes it all broken...

![milosz in grid](https://cloud.githubusercontent.com/assets/6032869/15303991/b76f1cfc-1bb2-11e6-94c5-7e2a202aa090.gif)

> You see how I try
To reach with words
What matters most
And how I fail.

—Czesław Miłosz

paperboyo:
In newer Firefox, Unicode works well, but quote marks still do not:

![image](https://user-images.githubusercontent.com/6032869/41735453-a4e6fe44-7581-11e8-9840-70ed386456e3.png)

paperboyo:
See [comment](https://github.com/guardian/grid/blob/master/kahuna/public/js/search/query.js#L87).

paperboyo:
Fixed by https://github.com/guardian/grid/pull/4398.

PR #3645, opened 2022-02-11
keep Free to use and My Upload checkbox in sync with search filters

abdelrahmansd:
## What does this change?

Description:

When a "Free to Use Only" tick has been selected  to view, 

and then click on an image to view

Return to Search

The tick at the showing at the top of the screen disappears however The filter is still being applied

## Screenshots

      
Images Tick 1
Images Tick 2
Images Tick 3

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/search/query.js
+++ b/kahuna/public/js/search/query.js
@@ -6,6 +6,7 @@ import angular from 'angular';
 // import 'angular-animate';
 import moment from 'moment';
 import {eq} from '../util/eq';
+import '../util/storage';
 import {guDateRange} from '../components/gu-date-range/gu-date-range';
 import template from './query.html';
 import {syntax} from './syntax/syntax';
@@ -17,7 +18,8 @@ export var query = angular.module('kahuna.search.query', [
     eq.name,
     guDateRange.name,
     syntax.name,
-    grStructuredQuery.name
+    grStructuredQuery.name,
+    'util.storage'
 ]);
 
 query.controller('SearchQueryCtrl', [
@@ -26,8 +28,9 @@ query.controller('SearchQueryCtrl', [
   '$state',
   '$stateParams',
   'onValChange',
+  'storage',
   'mediaApi',
-  function($rootScope, $scope, $state, $stateParams, onValChange, mediaApi) {
+  function($rootScope, $scope, $state, $stateParams, onValChange, storage, mediaApi) {
 
     const ctrl = this;
 
@@ -144,8 +147,24 @@ query.controller('SearchQueryCtrl', [
 
 
     $scope.$watchCollection(() => ctrl.filter, onValChange(filter => {
+        storage.setJs("isNonFree", ctrl.filter.nonFree ? ctrl.filter.nonFree : false, true);
+
         filter.uploadedBy = filter.uploadedByMe ? ctrl.user.email : undefined;
+        storage.setJs("isUploadedByMe", ctrl.filter.uploadedByMe, true);
+
         ctrl.collectionSearch = ctrl.filter.query ? ctrl.filter.query.indexOf('~') === 0 : false;
+
+
+        const defaultNonFreeFilter = storage.getJs("defaultNonFreeFilter", true);
+        if (defaultNonFreeFilter && defaultNonFreeFilter.isDefault === true){
+          let newNonFree = defaultNonFreeFilter.isNonFree ? "true" : undefined;
+          if (newNonFree !== filter.nonFree){
+                storage.setJs("isNonFree", newNonFree ? newNonFree : false, true);
+                storage.setJs("isUploadedByMe", false, true);
+                storage.setJs("defaultNonFreeFilter", {isDefault: false, isNonFree: false}, true);
+          }
+          Object.assign(filter, {nonFree: newNonFree, uploadedByMe: false, uploadedBy: undefined});
+        }
 
         $state.go('search.results', filter);
     }));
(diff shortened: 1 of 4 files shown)


paperboyo:
I don’t think this is an entirely correct fix (after testing in our TEST env). The initial state of Free to use checkbox is controlled by a permission. Most our users have this permission (`show_paid`) set to off. In this state, your scenario behaves correctly whether the checkbox is ON (default) or OFF (manual user action).

But with this permission ON, indeed there is a problem: scenario behaves correctly whether the checkbox is OFF (default), but it behaves as you describe when the checkbox is ON (manual user action).

Your PR “fixes” manually overridden checkbox with permission ON, but breaks:
1. completely state with permission OFF – instead of checkbox be ON by default (as, again, it is for majority of our users), the checkbox is OFF, while the search (correctly is `/search`), so they are out of sync
2. clicking on a Grid logo (reset search) with permission ON (instead of `/search?nonFree=true`, clicking the logo gets you to `/search`; checkbox OFF, so out of sync with state)

I guess fixing your scenario properly will involve more than what this PR does. I know My uploads checkbox also behaves erratically. I hope the fix is not as involved as [this issue](https://github.com/guardian/grid/issues/1900)!

paperboyo:
Testing this revealed that current PROD is often wrong: the state of the checkbox doesn’t reflect the state of the running search. This PR doesn’t make it any worse and fixes some of the oddities.

Ideally, we would be consistently keeping the state of the checkbox (incl. when overriden manually) for all operation in the session and never have the checkbox NOT reflect reality. This PR is the step in this direction.

kahuna/public/js/search/query.js
```diff
@@ -144,7 +144,9 @@ query.controller('SearchQueryCtrl', [
 
 
     $scope.$watchCollection(() => ctrl.filter, onValChange(filter => {
+        $rootScope.nonFree = ctrl.filter.nonFree ? ctrl.filter.nonFree : false;
```
jonathonherbert, line 147: I can't see these properties being used on `$rootScope` elsewhere. `$rootScope` is effectively a global namespace – is there a reason we wouldn't scope these properties as variables defined within the controller? E.g.

```js
let isNonFree = false;

    isNonFree = ctrl.filter.nonFree ? ctrl.filter.nonFree : false;
    // ... etc.
```

abdelrahmansd: Thanks, @jonathonherbert for the review,
using the $rootScope to keep the nonFree value across pages,
using the controller variable will not keep the nonFree filter value when moving to another page and the Free-to-use checkbox will not reflect the current state

E.g:
When a "Free to Use Only" tick has been selected to view,
and then click on an image to view
Then Return to Search
The tick  disappears however The filter is still being applied

jonathonherbert: Gotcha, the requirement makes sense.

Can I recommend writing a little singleton service to keep track of this state, rather than adding to `$rootScope`? It'll more clearly encapsulate the values and what they're for, and will avoid adding things to a global variable. (`$rootScope` will become crowded and confusing quite quickly if we continue to take this approach elsewhere.)

I'm thinking really small – an example in this codebase might be `services/api/edits-api.js`, which uses some local state. Happy to find another example or explain further if that's not clear.

I'd be super happy to pair.

kahuna/public/js/search/query.js
```diff
@@ -170,12 +175,27 @@ query.controller('SearchQueryCtrl', [
...
+          ctrl.filter.uploadedByMe =  isUploadedByMe;
+        }
 
-        if (ctrl.filter.nonFree === undefined) {
+        if (isNonFree === undefined) {
           ctrl.filter.nonFree = session.user.permissions.showPaid ?
             session.user.permissions.showPaid : undefined;
+            storage.setJs("isNonFree", session.user.permissions.showPaid ? session.user.permissions.showPaid : false);
+        }
+        else if (isNonFree === true || isNonFree === "true") {
```
jonathonherbert, line 195: non-blocking: IIUC this could be more concisely expressed as `Boolean(isNonFree)`.

kahuna/public/js/search/query.js
```diff
@@ -170,12 +175,27 @@ query.controller('SearchQueryCtrl', [
...
+        if (isUploadedByMe === undefined) {
+              ctrl.filter.uploadedByMe = ctrl.uploadedBy === ctrl.user.email;
+              storage.setJs("isUploadedByMe",ctrl.filter.uploadedByMe);
+        }
+        else {
+          ctrl.filter.uploadedByMe =  isUploadedByMe;
+        }
 
-        if (ctrl.filter.nonFree === undefined) {
+        if (isNonFree === undefined) {
```
jonathonherbert, line 190: Am I right in thinking that this will break when users' permissions are revoked and `isNonFree` is true? If that's the case, perhaps we could return early if the permission is not present, setting `isNonFree` accordingly.

jonathonherbert:
The localStorage solution looks good, @abdelrahmansd – nice work! I've added a few comments.

_> There is plenty to improve, but not related to this PR as it’s broken in PROD too, eg.:_
_>_
_> show_paid is granted_
_> user clicks Free to use to check it_
_> user reloads the page_

Am I right in thinking that localStorage solves this problem, too? (cc. @paperboyo) – I've tested locally and it looks good to me.

kahuna/public/js/search/query.js
twrichards: @jonathonherbert it appears @abdelrahmansd's way is safer, `Boolean` constructor will coerce any truthy value to true

image
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean/Boolean

jonathonherbert: Good spot, thanks for the correction!

kahuna/public/js/search/query.js
jonathonherbert: Resolved in favour of using `localStorage` in https://github.com/guardian/grid/pull/3645/commits/d5bc5006460f3271a84f840328c1af5c0299ac7f.

paperboyo:
Great! All issues are fixed, only preexisting issue B is there. For another PR, this is ready for another formal review. Thanks so much for persevering, @abdelrahmansd!

kahuna/public/js/search/query.js
```diff
@@ -144,10 +147,28 @@ query.controller('SearchQueryCtrl', [
...
+        const defaultNonFreeFilter = storage.getJs("defaultNonFreeFilter", true);
+        if (defaultNonFreeFilter && defaultNonFreeFilter.isDefault === true){
+          let newNonFree = defaultNonFreeFilter.isNonFree ? "true" : undefined;
+          if (newNonFree !== filter.nonFree){
+                storage.setJs("isNonFree", newNonFree ? newNonFree : false, true);
+                storage.setJs("isUploadedByMe", false, true);
+                storage.setJs("defaultNonFreeFilter", {isDefault: false, isNonFree: false}, true);
+          }
+          newFilter = {};
+          newFilter = Object.assign(filter, {nonFree: newNonFree, uploadedByMe: false, uploadedBy: undefined});
```
jonathonherbert, line 168: A few questions:

- What is the purpose of the reassignment on L167? It looks to me like `newFilter` is immediately reassigned to the result of `Object.assign` on the next line.
- To double check – IIUC Object.assign will mutate `filter` and return the target object ([MDN]((https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign))), which will mean that at as a result of this expression, `newFilter` and `filter` will be identical.  Is the mutation of `filter` intended? If so, I think `newFilter` is redundant – passing `filter` to the `$state.go` call should be equivalent, as the variables reference the same data.

jonathonherbert (commented):
Thanks for the revision, @abdelrahmansd – a few more comments, but this is looking good.

kahuna/public/js/search/query.js
abdelrahmansd: @jonathonherbert yup,` Object.assign` will mutate the object not return a new object
I fixed it in the last commit.

jonathonherbert (approved):
Thanks for the revisions, @abdelrahmansd – LGTM!

PR #3649, opened 2022-02-14
Support removing all fields from listish metadata

andrew-nowak:
## What does this change?

Values from image metadata can come from two sources: metadata on the original uploaded image, or user edits via Grid UI. These two sources are stored as separate fields in our Elasticsearch index, and combined by Grid application code. 
When they are merged, user edits are preferred and should be displayed over original metadata.

In our Scala model, we model most of these fields as Options which makes sense: either the field has not been set, or it has (including user edits explicitly setting a field to "", which Grid understands as "explicitly unset"). However, the fields which can contain multiple values are modeled as Lists (and 1 Set). This can model a field that has not been set, or has been set with values, but _cannot model a field that has been explicitly set to no values_. For example, this means that a user cannot remove all people from an image where the people have been identified from the original metadata --- attempting to do so will be treated as if the user has set no override and the original people will be shown again.

This PR instead models these list fields as Option[List] (and Option[Set]), which now allows us to model the difference, as None represents unset, and Some(Nil) represents explicitly set to empty.

We can choose to delete all entries from the 'people' field in grid UI

## Tested? Documented?
- [x] locally by committer


common-lib/src/main/scala/com/gu/mediaservice/model/ImageMetadata.scala
```diff
@@ -40,13 +40,13 @@ case class ImageMetadata(
     suppliersReference = (that.suppliersReference ++ this.suppliersReference).headOption,
     source = (that.source ++ this.source).headOption,
     specialInstructions = (that.specialInstructions ++ this.specialInstructions).headOption,
-    keywords = that.keywords ++ this.keywords,
+    keywords = (that.keywords ++ this.keywords).headOption,
```
twrichards, line 43: I'm not sure `headOption` is right here, its quite different from before, and tbh not sure how it conforms to the type `Option[List[String]]` as i would've thought `headOption` here returns `Option[String]` - I'm sure I'm misunderstanding though

twrichards (commented):
totally onboard with the spirit of this PR (which is well articulated in the description) - I've left one comment about something that confuses me, hopefully you can shed some light? after which i'll approve, given you've tested it.

great work btw 👏

common-lib/src/main/scala/com/gu/mediaservice/model/ImageMetadata.scala
andrew-nowak: The `++` here first does the implicit conversion of `Option[List[String]]` to `Iterable[List[String]]` and _then_ concatenates, so the output will have 0-2 items (`that.keywords ++ this.keywords` ends up being roughly the same as `List(that.keywords, this.keywords).filter(_.isDefined).map(_.get)`), so headOption will give an `Option[List[String]]`

It is quite unlike the old behaviour, but that's because I think the old behaviour was wrong - this func is only used in projection (where `that` is the original metadata and `this` is the user metadata and we're trying to do the merge), and from a quick test: if an image originally had 2 people, one of them was removed by a user, and then image was projected, that other person appears again in the metadata 😬  If subjects or keywords were user editable, I think that projecting could start duplicating items, since they're Lists and not Sets.

This way we should take whatever the user metadata is, and only pick up from the original metadata if no user metadata was set at all. I think this is the correct behaviour, as it makes these fields behave in the same way as all of the other fields and not as exceptions.

twrichards: hmm I wonder if there's a way to make it more explicit, rather than relying on that implicit behaviour (even though it did before)?

andrew-nowak: I've replaced these concats and headOptions with orElse, hopefully that's a bit more straightforward to follow at a glance!

twrichards (approved):
once this has been re-tested on `TEST`, please go ahead and 🚢

andrew-nowak:
Tested again on TEST

PR #3653, opened 2022-02-15
Look up environment vars from /etc/environment, rather than unused default /etc/default/<appname>

andrew-nowak:
## What does this change?

SBT Native Packager will by default set up systemd to look up environment variables from a specific file `/etc/default/<appname>`. This file is packaged into the .deb package we use to deploy. Some of our user data scripts expect to be able to load environment variables from `/etc/environment`, but these currently aren't used. It's difficult to move these variables into `/etc/default/<appname>` as we can't inject the variables until the package has been installed (as that is when the file is created), but if we inject the variables after installation the app has already started and will not read the file again. Instead, instruct systemd to load environment variables from `/etc/environment`, which we can write in our user data script before installation of the package.

Environment variables set in user data scripts will be respected.

## Tested? Documented?
- [x] locally by committer


andrew-nowak:
[Allow imagemagick to use more RAM](https://trello.com/c/LciVf5qH/467-allow-imagemagick-to-use-more-ram)

PR #3654, opened 2022-02-15
Exclude `gu-fc-` from Grid-like mediaIds

paperboyo:
## What does this change?
CAPI `mediaId`s starting with `gu-fc-` do not come from Grid ([example](https://www.theguardian.com/discover-culture/2014/aug/07/stephen-hough-in-recital-sydney)). This excludes those from Usage checks.

I learn if `||` found on the Internet does what I need it to.

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/usage/MediaUsage.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/usage/MediaUsage.scala
@@ -24,7 +24,7 @@ case class MediaUsage(
 ) extends GridLogging {
 
   def isGridLikeId: Boolean = {
-    if (mediaId.startsWith("gu-image-")) || (mediaId.startsWith("gu-fc-")) {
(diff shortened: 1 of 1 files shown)


common-lib/src/main/scala/com/gu/mediaservice/model/usage/MediaUsage.scala
```diff
@@ -24,7 +24,7 @@ case class MediaUsage(
 ) extends GridLogging {
 
   def isGridLikeId: Boolean = {
-    if (mediaId.startsWith("gu-image-")) {
+    if (mediaId.startsWith("gu-image-")) || (mediaId.startsWith("gu-fc-")) {
```
andrew-nowak, line 27:
```suggestion
    if (mediaId.startsWith("gu-image-") || mediaId.startsWith("gu-fc-")) {
```

andrew-nowak (approved):

PR #3656, opened 2022-02-18
Buffer and report Cloudwatch metrics using an Akka actor

andrew-nowak:
## What does this change?

The grid's existing implementation for cloudwatch metrics uses a scalaz stream to buffer metrics to allow reporting as a group (more of a queue than a stream). https://github.com/scalaz/scalaz-stream has never been released for Scala 2.13, so is one of several dependencies blocking us from upgrading Grid. This PR replaces the scalaz stream with an Akka actor to perform the same functionality - buffer metrics and report in a batch on an interval.

This PR does not drop the dependency on scalaz due to other usages. These usages do not use scalaz-stream, but on other scalaz syntaxes that have been removed in later versions, which will be addressed in a follow up PR(s).

Tested on TEST - deployed and Cloudwatch metrics are still reported (`TEST/MediaApi results` is reporting frequently and is best to observe that metrics are still coming through on Cloudwatch, alternatively you can search for "metric data points" in the logs to regular sending of metrics).

No regressions; one step closer to upgrading to 2.13.

## Tested? Documented?
- [x] locally by committer


twrichards (approved):
you've tested in TEST and the code LGTM, however I notice you've consolidated the `recordOne` and `runRecordOne` (and similar) functions and removed various instances of `.run` after an `.increment()`, I imagine those were there intentionally to ensure the metric is delivered right away where needed - has the behaviour changed? or is immediate now the default? (the answer might be worth clarification in the PR description)

andrew-nowak:
> you've tested in TEST and the code LGTM, however I notice you've consolidated the `recordOne` and `runRecordOne` (and similar) functions and removed various instances of `.run` after an `.increment()`, I imagine those were there intentionally to ensure the metric is delivered right away where needed - has the behaviour changed? or is immediate now the default? (the answer might be worth clarification in the PR description)

Actually if you read the implementations of those functions in the previous diff, `runRecordOne` is only a shorthand for `recordOne(...).run` - ie. `run` needs to be called either way. I couldn't find any difference, so opted to consolidate for future ease, especially as the sending to the "queue" is now async.

PR #3664, opened 2022-02-28
Metadata templates

ochiengolanga:
## What does this change?

Introduces new functionality that allows application of metadata field values and usage rights defined in a metadata template to be applied to an image.

Adding the following configuration to kahuna configuration should

- Render "Template select" component on the "My uploads" page and Image view page as shown in the illustrations below.
- Allow a user to select a metadata template and apply the template values to the image metadata and usage rights (if configured)

```hocon
metadata.templates = [
  {
    templateName: "Social Media"
    usageRights: {
      category: "social-media" # usage right category
      restrictions: "SOCIAL MEDIA - Usage requires approval by a Senior Editor" # Optional
    }
    metadataFields: [
      {
        name: "byline" # ImageMetadata field name
        value: "My Byline"
        resolveStrategy: "replace" # Optional, possible values "replace", "prepend", "append" and "ignore". Default value is "replace"
      }
      {
        name: "credit"
        value: "My Credit"
        resolveStrategy: "replace"
      }
      {
        name: "copyright"
        value: "Social media source Copyright"
        resolveStrategy: "replace"
      }
      {
        name: "specialInstructions"
        value: "Approval needed from a senior editor"
        resolveStrategy: "replace"
      }
    ]
  }
  {
    templateName: "User uploads"
    usageRights: {
      category: "Bylines"
      restrictions: "Bylines - Usage requires approval by a Senior Editor"
    }
    metadataFields: [
      {
        name: "copyright"
        value: "My Copyright"
        resolveStrategy: "replace"
      }
    ]
  }
]
```

### My uploads view

Upload page

### Search view

Images page

### Image view

Image page

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [X] locally by committer

### DIFF
--- a/common-lib/src/main/resources/application.conf
+++ b/common-lib/src/main/resources/application.conf
@@ -102,3 +102,5 @@ usageRightsConfigProvider = {
 }
 
 domainMetadata.specifications = []
+
+metadata.templates = []
--- a/kahuna/app/controllers/KahunaController.scala
+++ b/kahuna/app/controllers/KahunaController.scala
@@ -46,6 +46,7 @@ class KahunaController(
     // is on the same domain and can be read by the JS
     val domainMetadataSpecs: String = Json.toJson(config.domainMetadataSpecs).toString()
     val fieldAliases: String = Json.toJson(config.fieldAliasConfigs).toString()
+    val metadataTemplates: String = Json.toJson(config.metadataTemplates).toString()
     val returnUri = config.rootUri + okPath
     Ok(views.html.main(
       config.mediaApiUri,
@@ -65,7 +66,8 @@ class KahunaController(
       config.systemName,
       config.canDownloadCrop,
       domainMetadataSpecs,
-      config.recordDownloadAsUsage
+      config.recordDownloadAsUsage,
+      metadataTemplates
     ))
   }
 
--- a/kahuna/app/lib/KahunaConfig.scala
+++ b/kahuna/app/lib/KahunaConfig.scala
@@ -46,5 +46,7 @@ class KahunaConfig(resources: GridConfigResources) extends CommonConfig(resource
     shouldLoadWhenIFramed = if (entry.hasPath("shouldLoadWhenIFramed")) Some(entry.getBoolean("shouldLoadWhenIFramed")) else None,
   ))
 
+  val metadataTemplates: Seq[MetadataTemplate] = configuration.get[Seq[MetadataTemplate]]("metadata.templates")
+
 }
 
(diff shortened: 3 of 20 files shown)


paperboyo:
Hi Daniel, thanks for this! Some discussion points below. I‘m not so sure on the UX in general, but even in this incarnation,

1. clicking Apply should go back to initial state of the dropdown (_Please select…_; no button active), instead of this:
    image
2. `prepend` doesn’t work (haven’t checked `append`). 
3. What’s the reason for `ignore`? (isn’t the lack of value effectively that?)
4. When selecting multiple images in browser, applying a template doesn’t work at all.
5. When selecting multiple images in browser, the preview of the values (before clicking Apply) doesn’t appear like for single edits.
6. It’s possible to omit Usage Rights in a template, great! But when a template does that, Usage Rights enter editing mode after selecting this template.
7. I like the preview of the affected values a lot (before one clicks Apply). Maybe it should be somehow distinct (to be able to answer quite an important question: what does a specific template do).

It may be that some of the functionality is missing on purpose? We can discuss in Slack if you like, esp. if you’ve done some user research on what would be most useful.

ochiengolanga:
@paperboyo Thanks for the feedback. The recent commits resolve points 1,2,6 and 7 which are ready for testing and review.

paperboyo:
Thanks, @ochiengolanga. I like how the fields now clearly indicate what’s gonna get changed! Let us know if you want to address pts. 4–5 in this PR or in subsequent ones. Otherwise, I think it’s good, maybe two minor things (both illustrated by the GIF):
8. When no template is selected yet, Cancel and Apply buttons should either not be there or have no hover
9. Can you think of any way of indicating that template will affect Usage Rights? I know it cannot be a border, but maybe something else yellow?
10. 
not style

paperboyo:
Thanks, @ochiengolanga, pt. 8 is foxed now, yay! But regarding pt. 9:
– when the template contains Usage Rights – they are expanded now – good (yellow border around them would be nice, but not absolutely necessary)
– when the template **does not** contain Usage Rights – they are also expanded – and this is wrong, because it suggests they will change, while they won’t

ochiengolanga:
@paperboyo Thanks for the timely feedback. I agree yellow borders on the usage rights form will be a nice and consistent visual queue. I will attempt to reproduce the second bug and commit a fix

andrew-nowak:
One more bug here unfortunately - if you start to apply a template, then edit a field that's not affected by the template, the template values will seem to disappear and revert to their current values - until you apply the template, which will apply the template values

Kapture 2022-04-05 at 16 48 20

andrew-nowak (approved):

PR #3671, opened 2022-03-04
Remove scalaz syntaxes

andrew-nowak:
## What does this change?

Removes scalaz from the grid.

Following #3656 there are only a couple of usages of scalaz remaining.

- `<|` operator (runs right hand side function against left hand side object, then return the left hand side. Kinda adds method chaining onto methods which do not support method chaining)
- Nonempty Lists
- Validation types

We _could_ upgrade scalaz to continue using nonempty lists, but the `<|` operator has been removed. I explored how much utility we were getting from Nonempty lists and Validations: it seems to mostly be the `toNel` function, which `None` for an empty list, and `Some` for a list with items, which is straightforward to replicate. Validation types also are `Either`s under the hood, and can easily be replicated with those.

## Tested? Documented?
- [ ] locally by committer


andrew-nowak:
see #4345

PR #3673, opened 2022-03-04
Remove quicklens

andrew-nowak:
## What does this change?

We're importing the quicklens library into grid for a very narrow usage, which is also accomplished with a `copy`. Replace the usage and remove quicklens.

One less dependency, one step closer to 2.13

## Tested? Documented?
- [ ] locally by committer


bryophyta (approved):
🥳 🥳

PR #3678, opened 2022-03-08
Better tokeniser/dedup and catching more exotic Reuters metadata

paperboyo:
_Actually authored by @andrew-nowak_

## What does this change?
This:
- replaces more variants of `via` with a forward slash
- tokenises byline field also by ` - ` (space+hyphen+space) and `#` (research shows that it’s useful in a huge majority of cases) 
- deduplicates tokens in byline and credit case-insensitively
- matches more Reuters imagery by looking also at credits which **end** with all variants of `Reuters` spelling

Cleaner and more consistent metadata. More Reuters’ imagery recognised as such.

## Screenshots

Before | After
-------|------
image | image
image | image
image | image
image | image
image | image

Note: all the Reuters images above weren’t and now are correctly recognised as Reuters.

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
@@ -438,7 +438,7 @@ object ReutersParser extends ImageProcessor {
     )
     // Reuters and other misspellings
     // TODO: use case-insensitive matching instead once credit is no longer indexed as case-sensitive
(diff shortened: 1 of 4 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -418,6 +418,9 @@ object PaParser extends ImageProcessor {
 }
 
 object ReutersParser extends ImageProcessor {
+
+  private val reutersSpellings = """(?i) ?(/ ?)?(via )?(reuters|retuers|reuetrs)/?$""".r
```
jonathonherbert, line 422: Cannot wait for the follow-up PR(s?) 😁

jonathonherbert (approved):
Code looks good. Tests make everything really clear. 💯

PR #3697, opened 2022-03-29
use a `mozilla/sbt` docker image built for `linux/amd64` for GHA ScalaBuild

twrichards:
Scala GHA builds started failing (e.g. https://github.com/guardian/grid/runs/5726787902?check_suite_focus=true) due to a discrepancy in the `platform` of the runner vs the default in docker...
image

It looks like [`mozilla/sbt` docker image](https://hub.docker.com/r/mozilla/sbt/tags) has stopped being built for `linux/amd64` (which is what the GHA runners are)...
image

## What does this change?

... for the time being, use a hard coded image version (which was the last one built for `linux/amd64`) - since SBT will get the correct version of itself anyway. Longer term this isn't ideal.

No failing builds.

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/.github/actions/sbt/Dockerfile
+++ b/.github/actions/sbt/Dockerfile
@@ -1,4 +1,4 @@
-FROM mozilla/sbt
+FROM mozilla/sbt:8u292_1.5.7
 
 RUN apt-get -y update
 


paperboyo (approved):
I approve of this effort.

twrichards:
not sure why the GitHub actions haven't run yet???

jonathonherbert (approved):
Can see the [workflow running](https://github.com/guardian/grid/runs/5736228661?check_suite_focus=true), if that's all good then LGTM!

akash1810 (commented):
I think I was very optimistic with this `Dockerfile` with the addition of imaging libraries. The intent was to perform image operations during CI. IIUC we're not (yet) doing this, so we could move to the GH provided [Java action](https://github.com/actions/setup-java) that provides `sbt`, instead of our own.

twrichards:
@ochiengolanga and @abdelrahmansd you'll need to rebase your active branches onto `main` if you want the `ScalaBuild` to pass

PR #3672, opened 2022-03-04
Remove box dependency

andrew-nowak:
## What does this change?

Removes the dependency upon <https://github.com/guardian/box>, which has not received updates in several years. The dependency is a single file which provides an alternate API to AtomicReference. This PR removes the dependency by implementing the direct API for AtomicReference.

## How can success be measured?

No regressions, one step closer to upgrading to 2.13

## Tested? Documented?
- [ ] locally by committer
- [ ] locally by Guardian reviewer
- [ ] on the Guardian's TEST environment
- [ ] relevant documentation added or amended (if needed)

PR #3698, opened 2022-04-01
Inline Box library

DavidLawes:
## What does this change?
The `frontend` project is updating to scala 2.13. As part of the update we found that [Box](https://github.com/guardian/box) is not compatible with 2.13 and that the lib also contained a bug. The advice we got was to deprecate the Box lib and inline the functionality in projects that are using it. The related `frontend` PR is here: https://github.com/guardian/frontend/pull/24768

In anticipation that `grid` might also update to scala 2.13 (at some point in the future), this commit removes Box as a dependency and inlines the functionality that `grid` requires.

`com.gu.box` is no longer a dependency of this project, inlined Box class provides all necessary functionality.

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -106,7 +106,6 @@ lazy val commonLib = project("common-lib").settings(
     "com.sksamuel.elastic4s" %% "elastic4s-core" % elastic4sVersion,
     "com.sksamuel.elastic4s" %% "elastic4s-client-esjava" % elastic4sVersion,
     "com.sksamuel.elastic4s" %% "elastic4s-domain" % elastic4sVersion,
-    "com.gu" %% "box" % "0.2.0",
     "com.gu" %% "thrift-serializer" % "4.0.0",
     "org.scalaz.stream" %% "scalaz-stream" % "0.8.6",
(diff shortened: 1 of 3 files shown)


andrew-nowak:
Thanks for the PR! I'd come across Box too a few weeks ago; when I looked at it in #3672 I couldn't see that Box was a useful abstraction over AtomicRefs, so I instead replaced the usages with AtomicReference instances directly. Unfortunately, that PR has languished in draft since discovering many other libraries that would need work/replacement before upgrading Grid to 2.13; but since you've also been looking at Box recently I'd be interested if you have an opinion on the approach.

jonathonherbert:
I'd agree with removing `Box` entirely, both – thanks so much for raising this @DavidLawes!

DavidLawes:
Sounds like the better option for `grid` is to remove `box`, so I'll close this in the understanding that https://github.com/guardian/grid/pull/3672 is reopened and merged (at some point) 🙂

PR #3703, opened 2022-04-05
make the location field editable

abdelrahmansd:
## What does this change?

Make the location field editable in the metadata section in the right info panel

Location value is a concatenation of City, State, Country, Sub Location.
Edit Location Form appears when clicking on the edit location button
to add remove and update any of the Location fields

## Screenshots

Screenshot from 2022-04-05 18-05-29
Screenshot from 2022-04-05 18-06-10
Screenshot from 2022-04-05 18-16-57

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
@@ -317,7 +317,7 @@
 
                 <span ng-hide="locationEditForm.$visible" ng-repeat="prop in ['subLocation', 'city', 'state', 'country']"
                       >
-                        <span class="metadata-line__info">
+                        <span>
                             <a
                                ui-sref="{{ ctrl.hasMultipleValues(ctrl.rawMetadata[prop]) ? '' : 'search.results({query: (ctrl.metadata[prop] | queryFilter:ctrl.locationFieldMap[prop])})'}}"
                                ng-class="{'image-info--multiple': ctrl.hasMultipleValues(ctrl.rawMetadata[prop]),
@@ -331,7 +331,7 @@
                         </span>
 
                 </span>
-                <span ng-hide="locationEditForm.$visible" class="metadata-line__info"
+                <span ng-hide="locationEditForm.$visible" class="editable-empty"
                       ng-if="!ctrl.hasLocationInformation() &&
                             ctrl.userCanEdit &&
                             (!ctrl.hasMultipleValues(ctrl.rawMetadata.subLocation) &&
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
@@ -316,7 +316,12 @@
                         ng-click="locationEditForm.$show()"
                         ng-hide="locationEditForm.$visible">✎</button>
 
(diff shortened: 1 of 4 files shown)


paperboyo:
This is a very useful feature. Without going too far (upload page, adding when non-existing and, especially, some data validation), I think it would be great to at least 

1. deal with multiple values when they don’t match or are present only on some pics from a selection. Similarly to how we do it for other metadata fields (eg. _Multiple bylines (click ✎ to edit *all*)_:

    Location

    Since the edit button was already pressed, I’m not sure if in this situation:
    - a. there should be another pencil button next to non-matching field with the text eg. _Multiple cities (click ✎ to edit *all*)_ or
    - b. the field should just be filled with a italic text _Multiple cities_ which would become an empty field on click and accept a new value for all (with Cancel being an escape), but when left untouched would just left these fields untouched in respective images

    Probably b?

2. Clearing all fields that exist in embedded metadata doesn’t work correctly (similar to how it didn’t work in https://github.com/guardian/grid/pull/3531, which was fixed by https://github.com/guardian/grid/pull/3649), Andrew would know better but IIRC, it has something to do with userMetadata not keeping empty values as overrides (?)
3. Also, a tiny bug: after clicking pencil to edit, ESC key should cancel editing (as it does for other fields)

andrew-nowak:
I think this runs into a problem with the recently merged metadata templates. If the field is edited while applying a template, the template UI will break, so it's best to prevent field edits while a template is being edited - see https://github.com/guardian/grid/commit/307dcc76bd28178146b6db4aac411563e61886a3#diff-7661cb91c09677bedfa8e6a6a0b1c081a8502a387a7554e7d9ac28edc1159f36L57-R57

abdelrahmansd:
> I think this runs into a problem with the recently merged metadata templates. If the field is edited while applying a template, the template UI will break, so it's best to prevent field edits while a template is being edited

Hi, @andrew-nowak Thanks for the review,  PR Updated

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
```diff
@@ -308,22 +307,123 @@
...
+                        ng-click="locationEditForm.$show()"
+                        ng-hide="locationEditForm.$visible">✎</button>
+
+                <span ng-hide="locationEditForm.$visible" ng-repeat="prop in ['subLocation', 'city', 'state', 'country']"
+                      >
+                        <span>
+                            <a
+                               ui-sref="{{ ctrl.hasMultipleValues(ctrl.rawMetadata[prop]) ? '' : 'search.results({query: (ctrl.metadata[prop] | queryFilter:ctrl.locationFieldMap[prop])})'}}"
+                               ng-class="{'image-info--multiple': ctrl.hasMultipleValues(ctrl.rawMetadata[prop]),
+                                          'inactiveLink': ctrl.hasMultipleValues(ctrl.rawMetadata[prop])}"
```
andrew-nowak, line 324: This isn't a good way to disable a link - it's still partially accessible (it can be tabbed to and activated with the keyboard) and you need to forcibly override some of the link styling with !important. I'd suggest instead rendering an entirely different element (maybe a span?) if `hasMultipleValues` is true, which I think should also result in simpler elements

abdelrahmansd: Hi, @andrew-nowak  Thanks for the review, PR Updated

andrew-nowak (approved):

PR #3709, opened 2022-04-11
make the dateTaken field editable

abdelrahmansd:
## What does this change?

Make the date taken field editable in the metadata section in the right info panel

## Screenshots

Screenshot from 2022-04-14 02-52-58
Screenshot from 2022-04-14 02-53-05
Screenshot from 2022-04-14 02-53-38
Screenshot from 2022-04-14 02-53-50
Screenshot from 2022-04-14 02-54-06
Screenshot from 2022-04-14 02-54-19

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
@@ -210,8 +210,29 @@
 
     <div class="image-info__group" role="region" aria-label="Image metadata">
         <dl class="image-info__group--dl metadata-line">
-            <dt class="metadata-line image-info__wrap metadata-line__key image-info__group--dl__key--panel" ng-if="ctrl.metadata.dateTaken">Taken on</dt>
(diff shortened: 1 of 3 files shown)


paperboyo:
Hi. Thanks, very useful!

1. When the datetime embedded contains seconds and/or subseconds, only `0` works in these fields, otherwise this happens:
    image

Ideally, it would be possible to enter seconds (don’t think it’s useful to be able to edit subseconds, but if easier – that would be fine too as they may be coming in in pics)

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
```diff
@@ -210,8 +210,40 @@
...
+                      e:ng-class="{'image-info__editor--error': $error,
+                                   'image-info__editor--saving': takenEditForm.$waiting,
+                                   'text-input': true}">
+
+                    <span class="metadata-line__info" ng-if="ctrl.hasMultipleValues(ctrl.rawMetadata.dateTaken) && ctrl.userCanEdit">
+                        Multiple date (click ✎ to edit <strong>all</strong>)
+                    </span>
+
+                    <span class="metadata-line__info" ng-if="ctrl.hasMultipleValues(ctrl.rawMetadata.dateTaken) && !ctrl.userCanEdit">
+                        Multiple date
```
paperboyo, line 235:
```suggestion
                        Multiple dates
```

PR #3714, opened 2022-04-12
Add API to get single collection

andrew-nowak:
## What does this change?

Adds an API to get a single collection.
The collections API contains hrefs to APIs which look as though they would fetch data for a single collection, but no such API existed. Fortunately most of the required logic exists in other APIs which I was able to collate quite quickly.

Caveats: This doesn't fetch the child collection(s) - I think we'd need to scan the whole db to get them which would make this not much better than getting all the collections

We can get data for a single collection without fetching the entire collections tree out of Dynamo (quite slow).

## Tested? Documented?
- [x] locally by committer

Tested with guardian/pinboard#127

### DIFF
--- a/collections/app/controllers/CollectionsController.scala
+++ b/collections/app/controllers/CollectionsController.scala
@@ -101,6 +101,16 @@ def uri(u: String) = URI.create(u)
       (collection) => collection.description)
   }
 
+  def getCollection(collectionPathId: String) = authenticated.async {
+    store.get(uriToPath(collectionPathId)).map {
+      case Some(collection) =>
+        val node = Node(collection.path.last, Nil, collection.path, collection.path, Some(collection))
+        respond(node, actions = getActions(node))
+      case None =>
+        respondNotFound("Collection not found")
+    }
+  }
+
   def getCollections = authenticated.async { req =>
     allCollections.map { tree =>
       respond(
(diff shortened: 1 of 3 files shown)


jonathonherbert (commented):
Tested locally, getting a collection via `collections/:id` works great. One thing to mention – asking for a collection that doesn't exist throws, rather than exercising the `respondNotFound("Collection not found")` path I expected. E.g. for `https://media-collections.local.dev-gutools.co.uk/collections/does-not-exist`:

Screenshot 2022-04-25 at 17 34 25

One other thing – this might be a stretch, but AFAICS this endpoint isn't included in the hypermedia response for collections. I'm not sure whether it should be, or if that is the case whether this is necessary (I don't know if we've a case for discovering this endpoint programmatically yet!) but thought I'd mention.

```
"actions": [
    {
        "name": "add-child",
        "href": "https://media-collections.local.dev-gutools.co.uk/collections/Cypress%20Integration%20Testing",
        "method": "POST"
    },
    {
        "name": "remove",
        "href": "https://media-collections.local.dev-gutools.co.uk/collections/Cypress%20Integration%20Testing",
        "method": "DELETE"
    }
]
```

paperboyo:
@jonathonherbert ❤️ for thinking about hypermedia…

jonathonherbert (approved):

PR #3719, opened 2022-04-25
Create request ID in the RequestLoggingFilter

andrew-nowak:
## What does this change?

Currently request IDs are (sometimes) created in the controllers and (sometimes) passed down the execution flow as log markers from there. But this means that the request ID is not included on the access logs, so it becomes _very_ hard to correlate the request in the access logs with other log messages.

This PR changes the Play Filter responsible for the access logs to also create the request ID, and pass it through to the rest of the application via a custom internal-only header `x-grid-request-id`. This new header is only consumed in image-loader so far as a proof of concept, it should be fairly easy to roll out further if this works well!

This PR also removes the UUID from the UploadRequest case class as it wasn't being used

## Tested? Documented?
- [ ] locally by committer


jonathonherbert (approved):
I can see [logs with a consistent requestId](https://logs.gutools.co.uk/s/editorial-tools/goto/608dab40-c577-11ec-bf09-35dccc95800f) for `image-loader`, which is great!

Taking a look at `RequestLoggingFilter`, it looks like the only thing we'll need to do to use this elsewhere is to use `getRequestId` to extract the header value from requests? (Shame that we need to do this explicitly each time we create markers, but I can't think of another way...)

PR #3732, opened 2022-05-05
Simplify getty processor to rely on getty asset id

andrew-nowak:
## What does this change?

Previously we detected Getty supplied images as those with a Getty metadata block. But other non-Getty submitters might add a Getty block for their own reasons. We think a better detection would be if the image has an Asset ID in the Getty metadata block.

This means that we no longer need to check against these excluded credits and sources.

We also chose to remove the Getty Credit parser - the only way that images will be given Getty usage rights is if there's an Asset ID in the Getty metadata. 

Broadly, I think this pseudo-code describes the change:

before:
```
(GettyXmpParser)
if image has fileMetadata.getty AND credit is not excluded AND source is not excluded {
	usageRights = Agency(getty);
	usageRights.suppliersCollection =metadata.source;
	metadata.credit = "Getty Images" if not already defined
	metadata.suppliersReference = fileMetadata.getty."Asset ID"
} else {
	do nothing
}

then
(GettyCreditParser)
if image.metadata.credit contains "Getty" or "via Getty" or "/Getty" {
	usageRights = Agency(getty);
	usageRights.suppliersCollection = metadata.source;
} else if image.metadata.credit is one of ["AFP", "FilmMagic", "WireImage", "Hulton"] {
	usageRights = Agency(getty);
	usageRgihts.suppliersCollection = metadata.credit;
} else {
	do nothing
}
```

After

```
if image has getty.Asset ID {
	usageRights = Agency(getty); 
	if metadata.credit is one of ["AFP", "FilmMagic", "WireImage", "Hulton"] {
		usageRights.suppliersCollection = metadata.credit;
	} else {
		usageRights.suppliersCollection = metadata.source;
	}
	metadata.suppliersReference = fileMetadata.getty."Asset ID"
	metadata.credit = "Getty Images" if not already defined

} else {
	do nothing
}
```

Better Getty detection

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
@@ -10,7 +10,6 @@ import com.gu.mediaservice.model._
 class SupplierProcessors(resources: ImageProcessorResources)
   extends ComposeImageProcessors(
     GettyXmpParser,
-    GettyCreditParser,
     AapParser,
     ActionImagesParser,
     AlamyParser,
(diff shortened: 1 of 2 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -315,42 +315,17 @@ trait GettyProcessor {
...
-      "London News Pictures Ltd", "Showtime", "Propaganda", "Equinox Features", "Athena Picture Agency Ltd",
-      "www.edinburghelitemedia.co.uk", "WALES NEWS SERVICE", "Sports Inc", "UK Sports Pics Ltd", "Blitz Pictures",
-      "Consolidated News Photos", "MI News & Sport Ltd", "Parsons Media", "Tom Nicholson", "SONY BMG MUSIC ENTERTAINMENT"
-    )
-
-    val isExcludedByCredit = image.metadata.credit.exists(isExcluded(_, excludedCredit))
-    val isExcludedBySource = image.metadata.source.exists(isExcluded(_, excludedSource))
-    val hasGettyMetadata = image.fileMetadata.getty.nonEmpty
 
+    val hasGettyMetadata = image.fileMetadata.getty.contains("Asset ID")
```
twrichards, line 325: @paperboyo didn't you say there's two spellings of `Asset ID`??

twrichards (approved):
logic LGTM 😄 

perhaps @paperboyo can test on TEST?

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
paperboyo: Yes, but this (stupid!) copy of `xmp.GettyImagesGIFT` into our own `getty` contains [reconciliation](https://github.com/guardian/grid/blob/d2673a49e4d756ebefc619504cef9ef568bd27e4/image-loader/app/lib/imaging/FileMetadataReader.scala#L156). This is the *only* reason I’m not advocating removing this stupid copy altogether.

Aracho1 (approved):
I find the new changes much easier to follow ✨

PR #3736, opened 2022-05-09
Add restriction for PA images based in restriction message

andrew-nowak:
## What does this change?

Recognise and apply PA (Press Association) restriction comment as a usage restriction, which PA supplies in the description of the image.

Applying to our back catalogue can be completed with a migration.

We attempted automatically adding a 1 month lease on processing, but that doesn't work well. Adding to ES but not Dynamo is against our data storage model, and makes the lease uneditable. The supplier processors can't call out to the leases API to add the lease as the image hasn't yet been uploaded to ES, meaning that thrall would attempt to process an AddLease message _before_ the IndexImage message, which won't end well. To do this, image-loader _could_ make the api call after sending off the IndexImage message (even if Thrall picks up the message asynchronously, the messages will be processed in order), but this loses the encapsulation of the SupplierProcessors. More thinking/work is needed for that functionality.

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
@@ -3,6 +3,10 @@ package com.gu.mediaservice.lib.cleanup
 import com.gu.mediaservice.lib.config.{RuntimeUsageRightsConfig, UsageRightsConfigProvider}
 import com.gu.mediaservice.lib.metadata.UsageRightsMetadataMapper
 import com.gu.mediaservice.model._
+import com.gu.mediaservice.model.leases.{AllowUseLease, LeasesByMedia, MediaLease}
+import org.joda.time.DateTime
+
+import java.util.UUID
 
 /**
   * This is largely generic or close to generic processing aside from the Guardian Photographer parser.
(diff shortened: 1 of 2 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -418,11 +422,24 @@ object PaParser extends ImageProcessor {
...
-      image.copy(usageRights = Agency("PA", restrictions = restrictions))
+    if (isPa && image.metadata.description.exists(_.contains(restrictionMessage))) {
+        val firstLease = LeasesByMedia.build(List(MediaLease(
+          id = Some(UUID.randomUUID().toString),
+          startDate = Some(DateTime.now()),
+          endDate = Some(DateTime.now().plusMonths(1)),
+          access = AllowUseLease,
+          mediaId = image.id,
+          createdAt = DateTime.now(),
+          leasedBy = None,
```
paperboyo, line 433: Would this ever go in (and work), maybe that field could have some generic “user” helping identify all on-ingest Lease applications? Something like `ingestion.robot` or something. I’m still checking with Editorial. But if it works, it could be hugely valuable to eg. add a 24hrs Deny leases to a future Alamy Live News processor (them pics should be free after 24hrs).

Thanks for checking that!

PR #3749, opened 2022-05-18
Set thrall projection parallelism from config

andrew-nowak:
## What does this change?

The current attempt to migrate is suffering errors. It's not immediately clear where the problem lies, but my suspicion is related to the quantity of simultaneously attempted projections overwhelming the projection instances. Rather than creating a PR every time we tweak this value, allow setting it from config.

Better control over the amount of projections produced should let us run a smoother migration

## Tested? Documented?
- [ ] locally by committer


andrew-nowak:
@jonathonherbert thanks! currently the only consumers are ourselves (bbc do not exercise this functionality), and 50 now seems to be far too high. I think 1 is a good choice because it means that
- consumers who do not want to run migrations do not need to set any configuration
- when running without setting the option, thrall will still do _some_ projection
- consumers who want to migrate at pace will be forced to experiment and choose a value that works for them - this isn't something that we can really do for them, because it largely depends on infra choices, such as number of instances, size of instances, etc.

PR #3750, opened 2022-05-18
Remove admin-tools

andrew-nowak:
## What does this change?

I believe that the admin-tools app and related code is now defunct, replaced by [migration](https://github.com/guardian/grid/blob/main/docs/05-migration/01-about.md). Remove to help simplify repo and reduce confusion (primarily mine, every time I try to look at ImageDataMerger and end up in the admin-tools part rather than the image-loader projection part).

 

## Tested? Documented?
- [ ] locally by committer


jonathonherbert (approved):
If it ticks, I clicks.

✅ 🖱️ 

Guess we should remove the separate build in TC while we're at it.

PR #3760, opened 2022-05-26
make cost filter checkbox label configurable

abdelrahmansd:
## What does this change?
Add configuration For **_cost filter_** checkbox (Free to use only checkbox)
eg:
```
costFilter {
 label: "chargeable",
 chargeable: true
}
```
`costFilter.label` responsible for checkbox label, it will be **"_Free to use only_"** by default if the config doesn't present
`costFilter.chargeable` responsible for checkbox behavior it will be **_false_**  by default if the config doesn't present

if the `costFilter.label` value equals any synonyms of _**chargeable**_ eg ('chargeable', 'paid', ...etc) the `costFilter.chargeable` should set to **_true_** to make checkbox getting paid images when been ticked

if the `costFilter.label` value equals any synonyms of **_Free_** eg ('Free to use', 'non paid', 'gratis' ...etc) the `costFilter.chargeable` should set to **_false_** to make checkbox getting Free images when been ticked

 

without `costFilter` configuration or with `costFilter.chargeable = false` 
it should behave in the same behavior

with `costFilter.chargeable = true`
- users with `showPaid` permission 
   **_cost filter_** checkbox is ticked by default
- users without `showPaid` permission 
   **_cost filter_** checkbox is unticked by default

 if the  `costFilter` configuration doesn't present or `costFilter.chargeable` set to false
 **_cost filter_** checkbox (Free to use only checkbox) should be the same experience
 if the  `costFilter` configuration present and `costFilter.chargeable` set to true
 - users with `showPaid` permission should see **_cost filter_** checkbox is ticked by default
- users without `showPaid` permission should see  **_cost filter_** checkbox is unticked by default
 
## Screenshots
without `costFilter` configuration or with `costFilter.chargeable = false` and users with `showPaid` permission
Screenshot from 2022-05-26 03-43-21
without `costFilter` configuration or with `costFilter.chargeable = false` and users without `showPaid` permission
Screenshot from 2022-05-26 03-43-35

with `costFilter.chargeable = true` and users with `showPaid` permission
Screenshot from 2022-05-26 03-42-14
with `costFilter.chargeable = true` and users without `showPaid` permission
Screenshot from 2022-05-26 03-42-32

## Who should look at this?

@guardian/digital-cms

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/kahuna/public/js/search/query.js
+++ b/kahuna/public/js/search/query.js
@@ -33,6 +33,10 @@ query.controller('SearchQueryCtrl', [
   function($rootScope, $scope, $state, $stateParams, onValChange, storage, mediaApi) {
 
     const ctrl = this;
+    ctrl.costFilterLabel = window._clientConfig.costFilterLabel;
+    ctrl.costFilterChargeable = window._clientConfig.costFilterChargeable;
+    ctrl.costFilterFalseValue =  ctrl.costFilterChargeable ? undefined : "'true'";
+    ctrl.costFilterTrueValue =  ctrl.costFilterChargeable ? "'true'" : undefined;
 
     ctrl.canUpload = false;
 
@@ -214,10 +218,10 @@ query.controller('SearchQueryCtrl', [
         if (isNonFree === null) {
           ctrl.filter.nonFree = session.user.permissions.showPaid ?
             session.user.permissions.showPaid : undefined;
-            storage.setJs("isNonFree", session.user.permissions.showPaid ? session.user.permissions.showPaid : false);
+            storage.setJs("isNonFree", session.user.permissions.showPaid ? session.user.permissions.showPaid : false, true);
         }
         else if (isNonFree === true || isNonFree === "true") {
-            ctrl.filter.nonFree = true;
+            ctrl.filter.nonFree = "true";
         } else {
           ctrl.filter.nonFree = undefined;
         }
--- a/common-lib/src/main/resources/application.conf
+++ b/common-lib/src/main/resources/application.conf
@@ -104,3 +104,8 @@ usageRightsConfigProvider = {
 domainMetadata.specifications = []
 
 metadata.templates = []
+
+# costFilter {
+#  label: "chargeable",
+#  chargeable: true
+# }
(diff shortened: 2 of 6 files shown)


paperboyo:
Just looking at the screenshot, wouldn’t you want it capitalised? (like everything else in the UI)

abdelrahmansd:
> Just looking at the screenshot, wouldn’t you want it capitalised? (like everything else in the UI)
Thanks @paperboyo for giving it a look,yup it can be done through the configuration eg `costFilter.label = "Chargeable"`

kahuna/public/js/search/query.js
```diff
@@ -214,10 +218,10 @@ query.controller('SearchQueryCtrl', [
         if (isNonFree === null) {
           ctrl.filter.nonFree = session.user.permissions.showPaid ?
             session.user.permissions.showPaid : undefined;
-            storage.setJs("isNonFree", session.user.permissions.showPaid ? session.user.permissions.showPaid : false);
+            storage.setJs("isNonFree", session.user.permissions.showPaid ? session.user.permissions.showPaid : false, true);
```
jonathonherbert, line 221: IIUC this will store this value in session storage, not local storage – is this desired? cc. @paperboyo

jonathonherbert: (Ee think this might simply addressing an inconsistency with storage elsewhere? Good to confirm if so.)

jonathonherbert:
Tested w/ @andrew-nowak,  behaves as expected –  thanks! One question that'd be good to clarify before we approve.

kahuna/public/js/search/query.js
paperboyo: I wouldn’t know where we store current things we store? (panel locks etc). If there is no reason for divergence, I would store all things in the same place? But otherwise I have no opinion.

jonathonherbert: Ace, @andrew-nowak suspect we're good to approve in this case?

abdelrahmansd: Thanks, @jonathonherbert and @paperboyo for the review,
 in the past, it was stored in local storage (browser scoped) and it led to unwanted behavior so it been stored in session storage (tab scoped) and for this specific change in line 221, it was missed in the previous [PR](https://github.com/guardian/grid/pull/3645)
as you can see the same in line [154](https://github.com/guardian/grid/blob/d9ab813eb1dd50513e320a87a173867efc44eb46/kahuna/public/js/search/query.js#L154) and line [176](https://github.com/guardian/grid/blob/d9ab813eb1dd50513e320a87a173867efc44eb46/kahuna/public/js/search/query.js#L176)

PR #3761, opened 2022-05-27
Simpler OR queries for single text words

andrew-nowak:
## What does this change?

Users should be able to search for a variety of terms, and view all images which match at least one of those terms.

Tried it before in #3516 to allow searches along the lines of `cat OR dog OR a flock of birds OR +by:squirrel`. That didn't really work well because of the current implementation of the search bar and the way chips interact, plus it didn't really mesh well with the query language.

This sidesteps the problem by restricting the terms that can be or-ed together by adding a new chip which accepts a list of words to OR together. `+or:"cat dog birds squirrel"`

This approach has limits - only single words can be ORed together and chips cannot be used. But it is cheap and provides some functionality until a bigger change is possible.

Deploy to TEST. Can you query for different words and see all of the results for the different querywords in the same page?

image

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/kahuna/public/js/search/query.js
+++ b/kahuna/public/js/search/query.js
@@ -212,9 +212,9 @@ query.controller('SearchQueryCtrl', [
         }
 
         if (isNonFree === null) {
-          ctrl.filter.nonFree = session.user.permissions.showPaid ?
-            session.user.permissions.showPaid : undefined;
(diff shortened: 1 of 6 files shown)


paperboyo:
Added it to the info flyout, feel free to change the wording/everything:
image

PR #3762, opened 2022-05-30
Upsert images into both indices

andrew-nowak:
## What does this change?

Reuse the upsert logic when inserting into the migration index. This helps ensure that any changes made to an image are not hidden.

If a migration was running, and an image had been migrated, any changes made to the image would "disappear" on reuploading because the result of the upload pipeline would be directly inserted into the migration index. We should remember that the image may be a reupload, and safely merge the metadata into the index.

Reupload an image, check that edits (and usage rights, crops, usages, etc.) stay visible in the UI.

Migration is safer

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/thrall/app/lib/elasticsearch/ElasticSearch.scala
+++ b/thrall/app/lib/elasticsearch/ElasticSearch.scala
@@ -87,7 +87,7 @@ class ElasticSearch(
 
 
   def migrationAwareIndexImage(id: String, image: Image, lastModified: DateTime)
-                              (implicit ex: ExecutionContext, logMarker: LogMarker): List[Future[ElasticSearchUpdateResponse]] = {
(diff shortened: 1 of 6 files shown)


twrichards (approved):
might want others to review

twrichards:
just had a thought, we could eliminate this nasty painless logic (which i think is only for re-upload) by explicitly checking if the image already exists in current index and if so projecting it and using the output of that as the `Image` for the `directInsert` into both currrent index and migration index (if applicable).

andrew-nowak:
@twrichards I took a look at this - I think it's a plausible idea, but I'm not comfortable enough with the script and all the places that it's touching to be confident in removing it. Having already been burnt once by thinking we could avoid it, I'm reluctant to do so again in short succession 😂  A good one to consider for the future though

PR #3753, opened 2022-05-20
do not write to logfiles

andrew-nowak:
stdout logs go to journald which are added to syslog; theres no need to store them in 2 files which can only overwhelm the disk

## Tested? Documented?
- [ ] locally by committer
- [ ] locally by Guardian reviewer
- [x] on the Guardian's TEST environment
- [ ] relevant documentation added or amended (if needed)

PR #3768, opened 2022-05-30
Limit the amount of file logs kept 

andrew-nowak:
See also <https://logback.qos.ch/manual/appenders.html#SizeAndTimeBasedRollingPolicy>

## What does this change?

This change is designed for Thrall, as a replacement for #3753 but will affect all services. In practice no service should be nearly as chatty as Thrall during a migration.

Limit the amount of space that can be consumed by Thrall application logs.

Since the migration to fluentbit, we store logs in 3 places - syslog, journald and the classic log files. We are using our infra to manage the rotation of syslog and journal files, we should also configure logback to do the same. We were already rotating logfiles once a day, storing compressed history for 7 days, but even that much can fill out an instance's disk space. This PR adds size based rolling, to rotate logfiles every day or 100MB, whichever comes first, storing 7 files' worth of history up to 500MB.

Deploy to TEST, start a migration and wait. This may take a long time, so you may want to deploy a branch with smaller values for the purpose of testing the configuration.

Logging continues even on a high-throughput process like a migration.

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/resources/logback.xml
+++ b/common-lib/src/main/resources/logback.xml
@@ -14,9 +14,11 @@
     <appender name="LOGFILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
         <file>${LOGS_LOCATION}/application.log</file>
 
-        <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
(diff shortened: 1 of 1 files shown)


jonathonherbert (approved):
These changes look reasonable, happy to approve. If you'd like me to test them by exercising in CODE, lmk and perhaps we can do that in concert with https://github.com/guardian/editorial-tools-platform/pull/610.

Issue #3780, opened 2022-06-08
Shared URLs do not affect app state correctly


paperboyo:
We are receiving user reports that possibly recent work (probably https://github.com/guardian/grid/pull/3645 and https://github.com/guardian/grid/pull/3724?) results in incorrect/confusing behaviour.

### Scenario 1

1. set `show_paid` permission to `true`
2. clear search and check Free to use checkbox
3. navigate to the URL with `&nonFree=true` that includes paid imagery (eg. https://media.gutools.co.uk/search?query=category:chargeable&nonFree=true)

Actual result:
`&nonFree=true` disappears from the URL and the sent search contents is not reflected

Expected result:
The sent URL should override app state.

### Scenario 2

1. set `show_paid` permission to `false`
2. clear search and navigate to the URL with `&nonFree=true` that includes paid imagery (eg. https://media.gutools.co.uk/search?query=category:chargeable&nonFree=true)

Actual result:
`&nonFree=true` disappears from the URL and the sent search contents is not reflected

Expected result:
The sent URL should override app state. Users are often sharing URLs and the receiver is expected to see the exact same results as the sender not dependent on their permissions or app state.

### Scenario 3

1. irrespective `show_paid` permission state (sic!)
2. (if `show_paid` is `false`) uncheck Free to use checkbox
3. navigate to a paid for image
4. click any of the outgoing search links eg. underline byline

Actual result:
Free to use checkbox appears unchecked, **but** (incorrectly), URL looses `&nonFree=true`, so the very image from which the search started is not included in results. 

Expected result:
The search URL should include the current app state. 

5. reload the page

Actual result:
It “fixes itself” after reloading the page.

Expected result:
There should be no mismatch between app state and URL that is cleared on page reload.

andrew-nowak:
this is an example of how we could fix scenario 3 (there may be better methods, it's worth exploring further) https://github.com/guardian/grid/commit/78982025c2de41419e8cca8df04994ac21a5e03f

PR #3782, opened 2022-06-08
try to fix bootstrapping of nonFree query param

andrew-nowak:
The only place in kahuna which should pay attention to the showPaid
permission should be the state for the root (ie. '/'), which after
finding the permission and setting the nonFree state param will
insert that to session storage ang then transition to /search.
If we are on the /search state and do not have nonFree in session
storage, then the user has followed a link directly to a search - in
that case we should respect the current state of the URL (ie.
stateParams.nonFree)

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/kahuna/public/js/search/index.js
+++ b/kahuna/public/js/search/index.js
@@ -104,6 +104,12 @@ search.config(['$stateProvider', '$urlMatcherFactoryProvider',
               });
             };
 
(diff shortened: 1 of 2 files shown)


paperboyo:
Fixes https://github.com/guardian/grid/issues/3780.

PR #3783, opened 2022-06-08
Don't render missing location fields

andrew-nowak:
## What does this change?

Some location fields may be null - previously we still rendered them with a null anchor, which threw errors as seen in #3781 

Without this PR, open almost any image with less than 4 location fields, there should be stack traces in the browser console. Then open the same image with this PR - there should not be stack traces.

Without this PR, there will be anchors with no content for the null location fields, eg. 
image
With this PR, these should all be gone.

Big reduction in errors in Sentry.

## Tested? Documented?
- [x] locally by committer


jonathonherbert (approved):
So quiet 🤫
Screenshot 2022-06-09 at 09 45 00

PR #3785, opened 2022-06-09
add pikaday-time dependency and update leases datepicker

abdelrahmansd:
## What does this change?

add pikaday-time dependency and update **_gu-date_** datepicker Configuration to support time

 

when adding a lease the datepicker in the form of the leases should support the time

when adding a lease and changing the time from the datepicker in the form of the leases 
the lease time should be updated instead of being (start/end of the day) 

## Screenshots
Screenshot from 2022-06-09 11-10-06
Screenshot from 2022-06-09 11-11-12

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/components/gu-date/gu-date.js
+++ b/kahuna/public/js/components/gu-date/gu-date.js
@@ -1,13 +1,13 @@
 import angular from 'angular';
 import moment from 'moment';
-import Pikaday from 'pikaday';
+import Pikaday from 'pikaday-time';
 import 'pikaday/css/pikaday.css';
 
 import template from './gu-date.html';
 import rangeTemplate from './gu-date-range-x.html';
 import './gu-date.css';
 
-const DISPLAY_FORMAT = 'DD MMM YYYY';
+const DISPLAY_FORMAT = 'DD MMM YYYY HH:mm';
 const TEN_YEARS_MILLIS = (10 * 365 * 24 * 60 * 60 * 1000);
 const START_OF_WEEK = 1; // Monday
 
@@ -47,6 +47,14 @@ guDate.directive('guDate', [function () {
 
             const pika = new Pikaday({
                 field: input,
+                showTime: true,
+                showMinutes: true,
+                use24hour: false,
+                incrementHourBy: 1,
+                incrementMinuteBy: 1,
+                incrementSecondBy: 1,
+                autoClose: true,
+                timeLabel: null, // optional string added to left of time select
                 container: container,
                 bound: false,
                 minDate: $scope.minDate && new Date($scope.minDate),
@@ -65,7 +73,7 @@ guDate.directive('guDate', [function () {
             $scope.closeOverlay = () => $scope.showingOverlay = false;
 
             if (angular.isDefined($scope.minDate)) {
-                $scope.dateRounder = (date) => moment(date).endOf('day').toDate();
+                $scope.dateRounder = (date) => moment(date).toDate();
 
                 $scope.$watch('minDate', value => {
                     const dateValue = value ? new Date(value) : new Date();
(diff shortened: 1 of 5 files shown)


paperboyo:
Just having a look at functionality. Pretty sure the weirdness about swapping day and month when entering manually (which only works in Chrome?) is there in PROD, but could we:
1. change the dropdown to provide 24hrs view, like the rest of the app?
2. maybe the dropdown could display whatever hour is already entered?

pika chrome

abdelrahmansd:
>     1. change the dropdown to provide 24hrs view, like the rest of the app?
> 
>     2. maybe the dropdown could display whatever hour is already entered?
> 
Thanks, @paperboyo for the review, I've updated the date picker to use 24 hour and this fixes point 2 also

kahuna/public/js/components/gu-date/gu-date.js
```diff
@@ -65,7 +73,7 @@ guDate.directive('guDate', [function () {
             $scope.closeOverlay = () => $scope.showingOverlay = false;
 
             if (angular.isDefined($scope.minDate)) {
-                $scope.dateRounder = (date) => moment(date).endOf('day').toDate();
+                $scope.dateRounder = (date) => moment(date).toDate();
```
andrew-nowak, line 76: if we no longer need to round dates, you should be able to get rid of these entirely. all these do now is convert a Date into moment then back into Date

kahuna/public/js/leases/leases.html
```diff
@@ -88,6 +88,10 @@
         </div>
     </div>
     <div ng-switch-default>
+<!--        <span ng-if="ctrl.access">-->
+<!--            <input type="datetime-local" id="startDate" ng-model="ctrl.newLease.startDate">-->
+<!--            <input type="datetime-local" id="endDate" ng-model="ctrl.newLease.endDate">-->
+<!--        </span>-->
```
andrew-nowak, line 94: should be deleted?

kahuna/public/js/components/gu-date/gu-date.js
```diff
@@ -47,6 +47,14 @@ guDate.directive('guDate', [function () {
 
             const pika = new Pikaday({
                 field: input,
+                showTime: true,
+                showMinutes: true,
+                use24hour: true,
+                incrementHourBy: 1,
+                incrementMinuteBy: 1,
+                incrementSecondBy: 1,
+                autoClose: true,
```
andrew-nowak, line 56: we find it jarring that the picker closes as soon as you pick a time - if I want to set a day, an hour and a minute I need to reopen the picker 3 times. Does autoClose change the behaviour so that I only need to open it once, pick all 3 things, then I can close it?

abdelrahmansd: unfortunately, `autoClose` will not change the behavior 
as mentioned [here](https://github.com/owenmead/Pikaday/issues/45)
but I tweaked the picker config, and it now keeps open after selection

paperboyo: Great to see it opened! But now the little chevron that used to open and close the picker does nothing?

paperboyo:
Small comment about the chevron.

I guess original pt. 2 is not easy to fix? (when one enters the text manually and then **subsequently** opens the picker, the picker does not reflect the text entry; in other words: if you choose something with the picker, close it and reopen, the picker doesn’† reflect the values)

abdelrahmansd:
> Small comment about the chevron.
> 
> I guess original pt. 2 is not easy to fix? (when one enters the text manually and then **subsequently** opens the picker, the picker does not reflect the text entry; in other words: if you choose something with the picker, close it and reopen, the picker doesn’t reflect the values)

Thanks, @paperboyo for the review I've updated the PR to show/hide the picker manually from the chevron icon and update the picker date on the input text change

kahuna/public/js/components/gu-date/gu-date.js
```diff
@@ -65,7 +67,6 @@ guDate.directive('guDate', [function () {
             $scope.closeOverlay = () => $scope.showingOverlay = false;
 
             if (angular.isDefined($scope.minDate)) {
-                $scope.dateRounder = (date) => moment(date).endOf('day').toDate();
```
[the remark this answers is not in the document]
andrew-nowak: awesome, thanks!

PR #3787, opened 2022-06-13
Revert "Revert "Process duplicate usages""

andrew-nowak:
Reverts guardian/grid#3786 to bring back #3778


Fweddi (approved):
We are confident https://github.com/guardian/grid/pull/3778 was not the cause of the issues (as the ES clusters resolved themselves at a different time to the merging of https://github.com/guardian/grid/pull/3786).

PR #3791, opened 2022-06-17
handle NonFree Filter and keep it in Sync with NonFree stateParam

abdelrahmansd:
## What does this change?

The only place in kahuna which should pay attention to the showPaid
permission should be the state for the root (ie. '/'), which after
finding the permission and setting the nonFree state param will
insert that to session storage and then transition to /search.
If we are on the /search state and do not have nonFree in session
storage, then the user has followed a link directly to a search - in
that case, we should respect the current state of the URL (ie.
stateParams.nonFree)

and update outgoing search links eg. underline byline to include NonFree value from the storage

This Fixes https://github.com/guardian/grid/issues/3780. Closes #3782.

credit goes to @andrew-nowak  https://github.com/guardian/grid/pull/3782/commits/02bdf52846f338b27063a0af4e8276dd57eaa678 , https://github.com/guardian/grid/commit/78982025c2de41419e8cca8df04994ac21a5e03f 

 

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
@@ -280,7 +280,7 @@
 
                         <span class="metadata-line__info" ng-if="!ctrl.hasMultipleValues(ctrl.rawMetadata.byline)">
                             <span ng-if="ctrl.metadata.byline">
-                                <a ui-sref="search.results({query: (ctrl.metadata.byline | queryFilter:'by')})"
+                                <a ui-sref="search.results({query: (ctrl.metadata.byline | queryFilter:'by'), nonFree: ctrl.srefNonfree()})"
                                    aria-label="Search images by {{ctrl.metadata.byline}} byline">{{ctrl.metadata.byline}}</a>
                             </span>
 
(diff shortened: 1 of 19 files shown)


kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
```diff
@@ -280,7 +280,7 @@
 
                         <span class="metadata-line__info" ng-if="!ctrl.hasMultipleValues(ctrl.rawMetadata.byline)">
                             <span ng-if="ctrl.metadata.byline">
-                                <a ui-sref="search.results({query: (ctrl.metadata.byline | queryFilter:'by')})"
+                                <a ui-sref="search.results({query: (ctrl.metadata.byline | queryFilter:'by'), nonFree: ctrl.srefNonfree()})"
```
andrew-nowak, line 283: thanks! there's still a few state transitions which don't have this because they're in other template files - I've seen this in at least labels and keywords. I think the important ones are gr-photoshoot.html, list-editor-compact.html and list-editor-info-panel.html, but you can find a full list by searching the repo for html files containing "search.results"

PR #3799, opened 2022-06-24
Guard against null mediaIDs, and include markers in logs

andrew-nowak:
## What does this change?

We have a couple of rows in the usage table that have null mediaIDs. They're usually not a problem, but the usages stream may throw NPEs when trying to process them (because they're null).

While I'm here, I've added logmarkers to the relevant log lines, and also added a check for "_no_ids" which I've come across in the table. It looks like CAPI can inject this where an element is missing an ID (https://github.com/guardian/content-api/blob/541a5b1c2fb18f5dad2110d5a00fe3ce7a53f04f/concierge/src/main/scala/com.gu.contentapi.concierge/processors/elementsgenerator/ElementsGenerator.scala#L129), though I don't know the specific trigger, and how it is different to the case that emits a null.

With difficulty - it may be possible to contrive an example in TEST dynamodb and send updates from a matching composer piece if required

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/usage/MediaUsage.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/usage/MediaUsage.scala
@@ -1,6 +1,6 @@
 package com.gu.mediaservice.model.usage
 
-import com.gu.mediaservice.lib.logging.GridLogging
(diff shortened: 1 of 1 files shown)


jonathonherbert (approved):
Seems sensible.

At the risk of stating the obvious, adding a unit test would be one way of feeling confident it's behaving as expected 😀

PR #3805, opened 2022-07-01
allow a few xmp fields to avoid redaction

andrew-nowak:
## What does this change?

In #3538 we removed content from any ICC or XMP metadata fields that were over 5000 chars long - some images had several huge fields that caused problems for storage and processing. This has been successful, but there are a few fields that can legitimately be large - these are the description, title and specialInstructions. We have a few thousand images with large descriptions, a handful with large titles, and none yet with specialInstructions that are getting redacted. (We also have one that is getting its suppliersReference redacted, but no real suppliersReference should be that large).

We could raise the content limit, but there's no obvious upper limit to what could be allowable.

Download one of the images that currently has its description redacted (<https://media.gutools.co.uk/search?query=description:%22%20%20%20%20%20REDACTED%20%28value%20longer%20than%205000%20characters,%20please%20refer%20to%20the%20metadata%20stored%20in%20the%20file%20itself%29%20%22&nonFree=true>), and reupload with this PR running. Does the description come through?

No more ugly/confusing redacting of image descriptions.

## Tested? Documented?
- [x] locally by committer


jonathonherbert (approved):
Code looks sensible, tested in TEST and behaves as expected 👍

PR #3806, opened 2022-07-04
Mass retry migration failures

andrew-nowak:
## What does this change?

image

Until now, retrying failed image migrations is an exercise in repeatedly clicking buttons, or making mass manual updates in ES - both undesirable. This PR adds a button that will retry failures a page at a time.

Caveat: manually enqueuing an image to be migrated now returns before a projection is attempted.
Pros: faster feedback for clicks, waiting for 250 images to be queued is much much faster than waiting for 250 images to be projected.
Cons: less feedback for images which fail to project due to problems internal to the image, a bigger and more confusing diff in this PR.

## How should a reviewer review this change?

Review commit-by-commit - I've done my best to split up the change from the prerequisite refactors, and each commit has a message describing the rationale.

Deploy to TEST. There's currently a few thousand failed images there; feel free to try out the button. After clicking, return to the dashboard and do some refreshing. Does the number of failures go down? Go back to the main dashboard and resume the migration (or pause if running). Go back and do a batch retry again. Does it also work this time?

Reduced chance of team members developing RSI

## Who should look at this?
 @guardian/digital-cms

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/thrall/app/lib/Paging.scala
+++ b/thrall/app/lib/Paging.scala
@@ -1,0 +1,22 @@
+package lib
+
+import play.api.mvc.{Result, Results}
+
+import scala.concurrent.Future
+
+case class Paging(page: Int, from: Int, pageSize: Int)
+
+object Paging extends Results {
+  def withPaging(maybePage: Option[Int])(f: Paging => Future[Result]): Future[Result] = {
+    val pageSize = 250
+    // pages are indexed from 1
+    val page = maybePage.getOrElse(1)
+    val from = (page - 1) * pageSize
+
+    if (page < 1) {
+      Future.successful(BadRequest(s"Value for page parameter should be >= 1"))
+    } else {
+      f(Paging(page, from, pageSize))
+    }
+  }
+}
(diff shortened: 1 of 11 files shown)


thrall/app/lib/Paging.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import play.api.mvc.{Result, Results}
+
+import scala.concurrent.Future
+
+case class Paging(page: Int, from: Int, pageSize: Int)
+
+object Paging extends Results {
+  def withPaging(maybePage: Option[Int])(f: Paging => Future[Result]): Future[Result] = {
+    val pageSize = 250
```
jonathonherbert, line 11: Wonder if it's worth noting any reason that this number is chosen in particular.

jonathonherbert (approved):
Tested in TEST, clicked a button and I can see the numbers going down.

IIUC it seems sensible to have the same id-based strategy for both queues, code seem neater as a result and it's easy to reason about one way of doing things.

The commit messages are very clear, and were useful when reviewing.

One minor comment.

thrall/app/lib/Paging.scala
andrew-nowak: Would be a good idea - but there's no particular reason for this number, and no barrier to changing in the future!

PR #3807, opened 2022-07-04
[IMAGEDAM-927] visibility of Programmes/Archives metadata submenu

ornel-lloyd-edano:
## What does this change?

This changes the visibility behavior of Hide/Show Programmes metadata and Hide/Show Archives metadata submenus. Given a user who cannot edit metadata and when selected image has no Programmes/Archives metadata to show anyway, then do not show the UI submenus "Hide/Show Programmes/Archives metadata" for this user

1. Add the configuration below in common-lib/src/main/resources/application.conf
```

domainMetadata.specifications = [
{
    name = "programmes"
    label = "Programmes"
    description = "Model used for Programmes"
    fields = [
      {
        name = "originalTxDate"
        label = "Original TX Date"
        type = "datetime"
      }
      {
        name = "channels"
        label = "Channels"
        type = "string"
      }
      {
        name = "seriesNumber"
        label = "Series Number"
        type = "string"
      }
    ]
  },
  {
    name = "archives"
    label = "Archives"
    description = "Model used for Archives"
    fields = [
      {
        name = "hard_copy_ref_number"
        label = "Hard Copy Ref Number"
        type = "string"
      }
      {
        name = "original_filename"
        label = "Original Filename"
        type = "string"
      }
    ]
  }
]

```

The above configuration will display for the selected image, Programmes metadata submenu with three fields under it (Original TX Date, Channels, Series Number) as well as Archives metadata submenus also with three fields under it (Hard Copy Ref Number, Original Filename)
2. Log in as a user who cannot add or edit metadata
3. Search an image with no values for all the fields under Programmes metadata and/or Archives metadata (**might be easier to upload a new image with empty metadata rather than search an existing one**)
4. Select that image then in the Metadata tab to the right, the Hide/Show Programmes metadata and Hide/Show Archives metadata submenu should not be displayed
5. Log in as a user who can edit metadata
6. Search any image with metadata for Programmes metadata and Archives metadata
7. Select the image then in the Metadata tab to the right, the Hide/Show Programmes metadata and/or Hide/Show Archives metadata submenu should still be visible regardless if some fields have value, all fields have value or no fields have value

Given the domainMetadata.specifications config shown above, a user with no edit rights can view and interact with the Hide/Show Programmes metadata and Hide/Show Archives metadata only if there are metadata present in these submenu. Otherwise, these UI components will be invisible to the user.

## Screenshots

User can edit. Programmes and Archives metadata are not empty (both panels visible)
Screen Shot 2022-07-05 at 1 15 10 AM

User cannot edit. Programmes metadata is not empty (panel visible but empty items hidden). Archives metadata is empty (panel hidden)
Screen Shot 2022-07-05 at 1 40 20 AM

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
@@ -450,7 +450,7 @@ module.controller('grImageMetadataCtrl', [
     };
 
     ctrl.isAdditionalMetadataEmpty = () => {
-      var totalAdditionalMetadataCount = Object.keys(ctrl.metadata).filter(key=> { ctrl.isUsefulMetadata(key); }).length +
+      const totalAdditionalMetadataCount = Object.keys(ctrl.metadata).filter(key => ctrl.isUsefulMetadata(key)).length +
       Object.keys(ctrl.additionalMetadata).length +
       Object.keys(ctrl.identifiers).length;
 
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
@@ -448,6 +448,14 @@ module.controller('grImageMetadataCtrl', [
     ctrl.isDomainMetadataEmpty = (key) => {
       return ctrl.domainMetadata.find(obj => obj.name === key ).fields.every(field => field.value === undefined );
     };
+
+    ctrl.isAdditionalMetadataEmpty = () => {
+      var totalAdditionalMetadataCount = Object.keys(ctrl.metadata).filter(key=> { ctrl.isUsefulMetadata(key); }).length +
+      Object.keys(ctrl.additionalMetadata).length +
+      Object.keys(ctrl.identifiers).length;
+
+      return totalAdditionalMetadataCount == 0;
+    };
 }
 ]);
 
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
@@ -444,6 +444,10 @@ module.controller('grImageMetadataCtrl', [
       ctrl.metadata = metadata;
       ctrl.usageRights.first().data = usageRights;
     };
+
+    ctrl.isDomainMetadataEmpty = (key) => {
+      return ctrl.domainMetadata.find(obj => obj.name === key ).fields.every(field => field.value === undefined );
+    };
 }
 ]);
 
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
@@ -585,7 +585,7 @@
     </dl>
     </div>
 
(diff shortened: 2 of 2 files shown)


paperboyo:
@ornel-lloyd-edano Thanks for this. We are not using domain metadata, but I just tested hiding `Show additional metadata` and found a problem: it hides the whole section if the image(s) lack hardcoded fields (source, suppliersReference and bylineTitle) **even** if additional fields have been configured to appear there (and they exist) via functionality available since https://github.com/guardian/grid/pull/3183.

So for example here, where the `common.conf` contains 
```
# field aliases
field.aliases = [
  {
    elasticsearchPath = "fileMetadata.iptc.Urgency"
    alias = "iptcUrgency"
    label = "IPTC Urgency"
    # searching won't work until we can reingest
    displaySearchHint = false
  },
  {
    elasticsearchPath = "fileMetadata.xmp.dc:subject"
    alias = "xmpKeywords"
    label = "XMP Keywords"
    # searching won't work until we can reingest
    displaySearchHint = false
  },
  {
    elasticsearchPath = "fileMetadata.xmp.xmpRights:WebStatement"
    alias = "xmpRightsUrl"
    label = "Rights URL"
    # searching won't work until we can reingest
    displaySearchHint = false
  },
  {
    elasticsearchPath = "fileMetadata.icc.Profile Description"
    alias = "iccProfile"
    label = "Colour Profile"
    # searching won't work until we can reingest
    displaySearchHint = true
  },
  {
    elasticsearchPath = "fileMetadata.colourModel"
    alias = "colourModel"
    label = "Colour Model"
    displaySearchHint = true
    searchHintOptions = [ "RGB", "CMYK", "GRAYSCALE", "LAB" ]
  },
  {
    elasticsearchPath = "fileMetadata.colourModelInformation.hasAlpha"
    alias = "hasAlpha"
    label = "Cutout"
    displaySearchHint = true
    searchHintOptions = [ "False", "True" ]
  },
  {
    elasticsearchPath = "fileMetadata.colourModelInformation.bitsPerSample"
    alias = "bitsPerSample"
    label = "Bits Per Sample"
    displaySearchHint = true
    searchHintOptions = [ "8", "16", "32" ]
  }
]
```

Show additional metadata section is hidden even though it shouldn’t.

paperboyo:
@ornel-lloyd-edano Thanks for the fix!

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
```diff
@@ -444,6 +444,18 @@ module.controller('grImageMetadataCtrl', [
       ctrl.metadata = metadata;
       ctrl.usageRights.first().data = usageRights;
     };
+
+    ctrl.isDomainMetadataEmpty = (key) => {
+      return ctrl.domainMetadata.find(obj => obj.name === key ).fields.every(field => field.value === undefined );
+    };
+
+    ctrl.isAdditionalMetadataEmpty = () => {
+      var totalAdditionalMetadataCount = Object.keys(ctrl.metadata).filter(key=> { ctrl.isUsefulMetadata(key); }).length +
```
andrew-nowak, line 453:
```suggestion
      const totalAdditionalMetadataCount = Object.keys(ctrl.metadata).filter(key => ctrl.isUsefulMetadata(key)).length +
```

I think you'll need to take the braces off the anonymous function here in order to correctly filter on the response

aracho1 (approved):
👍LGTM

PR #3817, opened 2022-07-18
Restrict image download permissions

aracho1:
## What does this change?
[Trello ticket](https://trello.com/c/XY6xdfgt/590-grid-prevents-downloads-for-unpermissioned-users)

In order to lower the risk of picture misuse, this PR restricts making downloading images from Grid behave in the same way as cropping of Paid, Restricted or No rights imagery is available only to permissioned users. Grid will indicate inability to download similarly to inability to crop.

Before
image

After
image

We have added a config so this feature can be turned on or off. While adding this config, we had to refactor the existing config list to get around the 22 parameters limit in Scala ([context](https://underscore.io/blog/posts/2016/10/11/twenty-two.html#:~:text=Back%20in%202014%2C%20when%20Scala,on%20in%20functions%20and%20tuples.)).

 

1. Deploy this PR to Grid TEST (unfortunately testing this locally is tricky as you would have to upload an image, which gives you the overriding permission to be able to download the image)
2. Quickest way to test this would be to select an existing image and delete its description (make sure you have the Grid `edit_metadata` rights in Permissions CODE) and refresh the page.
3. The top right corner should say both `Cannot download` and `Cannot crop`

Downloads of non Free to use pictures will be restricted for users without the right permission or if the image has key missing metadata such as credits or description (for reference, a full list of validity checks can be found [here]( https://github.com/guardian/grid/blob/13c1aff4c2362f6a595200883a9015dd0f95b00e/media-api/app/lib/ImageExtras.scala#L15-L23))

## Tested? Documented?
- [x] locally by committer


kahuna/app/controllers/KahunaController.scala
```diff
@@ -50,29 +50,18 @@ class KahunaController(
...
-      config.canDownloadCrop,
       domainMetadataSpecs,
-      config.recordDownloadAsUsage,
       metadataTemplates,
       additionalNavigationLinks,
       costFilterLabel,
-      costFilterChargeable
+      costFilterChargeable,
+      restrictDownload,
+      config
```
aracho1, line 64: Refactoring config to reduce the number of parameters in this function due to Scala limitation ([context](https://underscore.io/blog/posts/2016/10/11/twenty-two.html#:~:text=Back%20in%202014%2C%20when%20Scala,on%20in%20functions%20and%20tuples.))

kahuna/public/js/components/gr-downloader/gr-downloader.html
```diff
@@ -3,19 +3,29 @@
...
-            ng-if="ctrl.imageCount() > 1"
+            ng-if="ctrl.imageCount() > 1 && ctrl.canDownload"
             ng-click="ctrl.download('downloadUri')">
         <gr-icon-label gr-icon="file_download">
             Download<span ng-show="ctrl.downloading">ing</span>
         </gr-icon-label>
     </button>
 
-    <a ng-if="ctrl.imageCount() == 1"
+    <span ng-if="ctrl.imageCount() > 1 && !ctrl.canDownload" aria-label="Cannot download" gr-tooltip="Cannot download the selected images.">
```
aracho1, line 13: @paperboyo Any suggestions on what to put for `gr-tooltip` messaging? I left it quite generic for now but happy to follow your suggestion

kahuna/public/js/components/gr-downloader/gr-downloader.html
```diff
@@ -3,19 +3,29 @@
...
+        </gr-icon-label>
+    </span>
+
+    <a ng-if="ctrl.imageCount() == 1 && ctrl.canDownload"
        href="{{ ctrl.firstImageUris.uris.downloadUri }}" download rel="noopener" target="_blank"
        aria-label="Download image">
         <gr-icon-label gr-icon="file_download">Download</gr-icon-label>
     </a>
 
+    <span ng-if="ctrl.imageCount() == 1 && !ctrl.canDownload" aria-label="Cannot download" gr-tooltip="Cannot download this image.">
```
aracho1, line 25: @paperboyo Same here (for a single image)

kahuna/public/js/components/gr-downloader/gr-downloader.html
paperboyo: 1. Is this for multiple selected images? If yes – does one non-downloadable one block the whole download (even if others are fine)? I think it shouldn’t. The download zip should either just silently nopt have non-downloadable images (bad) or ditto, but with some kind of messaging (also bad, no idea how to message without a blocking popup).
2. No idea what would be the best tooltip. Maybe just “Downloading requires a lease” for any Cannot download button (where either one or all images are non-downloadable)?
    image

kahuna/public/js/components/gr-downloader/gr-downloader.html
paperboyo: See above.

kahuna/public/js/components/gr-downloader/gr-downloader.html
aracho1: Yes for now a single non-downloadable image would block the whole download. As you said, not sure where messaging should be positioned in that case, and silently ignoring it could maybe annoy the users? Not sure. 

Thanks for the suggestion on the tooltip, will update that.

paperboyo: Definitely let’s not block the whole batch download. Maybe we can repurpose some banner code to non-blockingly say: x images won’t download because they need lease(s) after clicking Download?

aracho1: Is there an example banner that's currently in use that's similar to what you have in mind? No worries if not, I can have a look in the code as well

paperboyo: There is one on crop page when crop is too small (non-blocking). There is also a cookie one (blocking I think?).

PR #3829, opened 2022-08-01
[IMAGEDAM-960] workaround S3 and Elasticsearch image state inconsistency when image cropping failed

ornel-lloyd-edano:
## What does this change?
Changes how GET /crops/:id gets a list of Crop that is displayed in the left side of the UI
Previous implementation only fetches the data from S3. New implementation gets data from Elasticsearch by calling GET /image/:id and then gets data from S3 GET /crops/:id. The assets from S3 are then matched from the assets from ES and only assets with synchronized state between S3 and ES are returned.

Select any image and perform 2 or more crops on that image. However, simulating a crop image failure that may result to state inconsistency is difficult. You can manually update the state in S3 or Elasticsearch for that image to mock a state inconsistency by deleting items in the exports json array (from GET /image/:id) or in items in data json array (from GET /crops/:id)

Only assets with consistent state between S3 and ES will be displayed on the left side panel and can be downloaded successfully.

## Screenshots

Consistent state between S3 and ES means ids and number of assets (normally 5 assets) should match

Screen Shot 2022-08-01 at 12 47 22 PM

Screen Shot 2022-08-01 at 12 45 32 PM

All crop thumbnails displayed on the left side of UI should be downloadable

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/Export.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/Export.scala
@@ -26,6 +26,7 @@ object Export {
       crop.master,
       crop.assets
   )
+
   implicit val exportWrites: Writes[Export] = (
     (__ \ "id").writeNullable[String] ~
     (__ \ "author").writeNullable[String] ~
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/Export.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/Export.scala
@@ -26,7 +26,6 @@ object Export {
       crop.master,
       crop.assets
   )
-
   implicit val exportWrites: Writes[Export] = (
     (__ \ "id").writeNullable[String] ~
     (__ \ "author").writeNullable[String] ~
--- a/cropper/app/controllers/CropperController.scala
+++ b/cropper/app/controllers/CropperController.scala
@@ -18,7 +18,6 @@ import org.joda.time.DateTime
 import play.api.libs.ws.WSClient
 
 import java.net.URI
-
 import com.gu.mediaservice.syntax.MessageSubjects
 
 import scala.concurrent.{ExecutionContext, Future}
@@ -90,7 +89,6 @@ class CropperController(auth: Authentication, crops: Crops, store: CropStore, no
     store.listCrops(id) map (_.toList) map { crops =>
       val deleteCropsAction =
         ArgoAction("delete-crops", URI.create(s"${config.rootUri}/crops/$id"), "DELETE")
-
       lazy val cropDownloadLinks = for {
         crop <- crops
         asset <- crop.assets
--- a/cropper/app/controllers/CropperController.scala
+++ b/cropper/app/controllers/CropperController.scala
@@ -18,6 +18,7 @@ import org.joda.time.DateTime
 import play.api.libs.ws.WSClient
 
 import java.net.URI
+
 import com.gu.mediaservice.syntax.MessageSubjects
 
 import scala.concurrent.{ExecutionContext, Future}
@@ -89,6 +90,7 @@ class CropperController(auth: Authentication, crops: Crops, store: CropStore, no
     store.listCrops(id) map (_.toList) map { crops =>
       val deleteCropsAction =
         ArgoAction("delete-crops", URI.create(s"${config.rootUri}/crops/$id"), "DELETE")
+
       lazy val cropDownloadLinks = for {
         crop <- crops
         asset <- crop.assets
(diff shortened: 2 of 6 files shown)


andrew-nowak (commented):
thanks for the PR! I can see how this would work around the inconsistency by pulling the ES data from media-api during the cropper request. However, I don't think this is the right approach.

At the time that the clientside is requesting the crops data from cropper (https://github.com/guardian/grid/blob/a451103f4987063885e5078bd0c3ab5a73664ab8/kahuna/public/js/image/controller.js#L221), it has already got a copy of the ES data from media-api in memory. (crops would be available at `ctrl.image.data.exports`) We should use that data and avoid the network request to cropper entirely (which is something that we've considered doing before but not had a reason to put in the time -- resolving this bug is a great time to put in the effort).

There is one piece of data that the cropper response includes which is not available on the media-api response, which is the URL to download the crop, which will need to be added onto the media-api response.

As an aside, the cropper `GET /crops/:id` needs to have no reliance on ES as it is one of the sources of truth used in [projection](https://github.com/guardian/grid/blob/main/docs/05-migration/01-about.md#projection)

ornel-lloyd-edano:
> ), it has already got a copy of the ES data from media-api in memory. (crops would be available at `ctrl.image.data.exports`) We should use that data and avoid the network request to cropper entirely (which is something that we've considered doing before but not had a reason to put in the time -- resolving this bug is a great time to put in the effort).

Hi @andrew-nowak I tried calling image.get before transitioning to `$state.go('image'` https://github.com/guardian/grid/blob/main/kahuna/public/js/crop/controller.js#L156

ornel-lloyd-edano:
> ), it has already got a copy of the ES data from media-api in memory. (crops would be available at `ctrl.image.data.exports`) We should use that data and avoid the network request to cropper entirely (which is something that we've considered doing before but not had a reason to put in the time -- resolving this bug is a great time to put in the effort).

Hi @andrew-nowak I tried calling image.get before transitioning to `$state.go('image'` https://github.com/guardian/grid/blob/main/kahuna/public/js/crop/controller.js#L156
However there is still a race condition and the ctrl.image.data.exports is not refreshed when screen returns (adding a delay before `$state.go('image'` solved the issue but obviously not a good solution)

Can you point me where to fix this synchronization? I suspect the data sent from $rootScope.$emit('events:crop-created' arrives slower than the transition kicked off from `$state.go('image'`

andrew-nowak:
`events:crop-created` isn't relevant here - I believe that it's only used for posting messages to a web app that is embedding the grid (this is how the grid talks to our CMS).

I still think the correct solution is to poll the API, waiting for the crop to be available, and transitioning when it is found. There is an example here: https://github.com/guardian/grid/blob/main/kahuna/public/js/services/api/collections-api.js#L90-L110 of how we do that for collections (obviously, replacing line 93 with a new function `untilCropAppears` and line 96 with the `$state.go`).

andrew-nowak:
thanks! This is making a lot better now. Please can the polling make use of the `apiPoll` function? This makes the polling a lot nicer, as it adds a delay between requests (prevents overloading the server) and will eventually fail if the crop doesn't appear in the response

ornel-lloyd-edano:
> thanks! This is making a lot better now. Please can the polling make use of the `apiPoll` function? This makes the polling a lot nicer, as it adds a delay between requests (prevents overloading the server) and will eventually fail if the crop doesn't appear in the response

Thanks @andrew-nowak for the suggestion. Updated PR to add apiPoll on recursive call of pollUntilImageUpdated

andrew-nowak:
thanks. this just needs a couple more tweaks - to save time describing what needs doing I've made this commit https://github.com/guardian/grid/commit/a6d03590251b73e3208730e8b9e86b975b217660?w=1 which I think is the best way to implement this - preferring promises to callbacks, and using a single definition of how to poll until the crop exists. take a look and test it out, and if it looks good we can cherry-pick it onto this branch and then I think this will be good to ship

ornel-lloyd-edano:
Hi @andrew-nowak I tried applying this commit https://github.com/guardian/grid/commit/a6d03590251b73e3208730e8b9e86b975b217660?w=1 However after cropping the screen does not automatically go back to previous. But when I click 'cancel' to go back, I can see the cropped image among the thumbnails to the left and I no longer need to reload so it seems the poll worked but the state change did not.

andrew-nowak:
whoops, thanks for spotting! that sounds like the right fix. the different UX sounds interesting, but I don't think users often want to create multiple crops in a single session - @paperboyo may have better formed opinions though

paperboyo:
I would keep the UX. Not often peeps crop multiple frames at a time, and I would worry it may upset our integration which is: picture(s) is(are) dropped into CMS/other tools as soon as the crop has finished.

andrew-nowak (approved):

PR #3837, opened 2022-08-11
Improve restrictions around sensible dateTaken values

andrew-nowak:
## What does this change?

a couple of our images have very malformed datetakens, which place the value so far
in the future that we get integer overflow, and attempt to tell
elasticsearch that the images were taken over a hundred million years
ago. after some thinking we reckon it makes a reasonable amount of sense
to restrict dates to those in current calendar era (ie. in AD, ie. 1 AD -> uploadTime)

This PR also restricts datetakens during projection to be before upload time (there's little sense in an image where the dateTaken is after the uploadTime, at least until time travel is invented)

Find some images with malformed (ie. future or pre-AD) datetakens / malform your own image's datetaken values, upload, project, etc. The malformed timestamps should be ignored.

eg. for image 4f304b65dde78b88823154d2607190b14d789f43, which has xmp field `"exif:DateTimeOriginal": "1008051601:00:00 00:00:"`:

Before:
```
"metadata": {
  "dateTaken": "-161056498-05-29T19:08:16.768Z",
  "description": "...",
  "credit": "...",
  "keywords": [
  ],
  "subjects": [
  ],
  "peopleInImage": [
  ]
},
```

After:
```
"metadata": {
  "description": "...",
  "credit": "...",
  "keywords": [
  ],
  "subjects": [
  ],
  "peopleInImage": [
  ]
},
```

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
@@ -10,6 +10,9 @@ import play.api.libs.json.{JsArray, JsString, JsValue}
 
 object ImageMetadataConverter extends GridLogging {
 
+  // keep positive years (ie AD) with a couple of days spare room to avoid accidentally flipping over into BC
+  val earliestSensibleDate: DateTime = DateTime.parse("0001-01-03T00:00:00.000Z")
+
   private def extractSubjects(fileMetadata: FileMetadata): Option[List[String]] = {
     val supplementalCategories = fileMetadata.iptc
       .get("Supplemental Category(s)")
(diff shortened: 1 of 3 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
```diff
@@ -10,6 +10,9 @@ import play.api.libs.json.{JsArray, JsString, JsValue}
 
 object ImageMetadataConverter extends GridLogging {
 
+  // elasticsearch does not like dates with years other than 4 digits long
+  val earliestSensibleDate: DateTime = DateTime.parse("1000-01-01T00:00:00.000Z")
```
davidfurey, line 14: ISO 8601 dates should always have 4 digits, the year before 1000 should be 0999

davidfurey: But if you really want to minimise the chance of ES ever seeing a date with a year of 999 or lower, you might want to go with 2nd January 1000 to avoid being bitten by timezones in future.

andrew-nowak: thanks! that makes a lot more sense than what I had been thinking. With that, I think it's probably sensible to restrict to years 1AD and later. does that also make sens to you?

davidfurey: Sounds reasonable to me

davidfurey (approved):
Seems reasonable until time travel is invented.

PR #3841, opened 2022-08-17
Fix/IMAGEDAM-999/keep metadata after cropping

ornel-lloyd-edano:
## What does this change?

Image metadata is removed during cropping. The Guardian writes back the a small portion of the original metadata but only to master crop. BBC wants to write back as much metadata as possible (configurable) but not only to the master crop but to all crop sizes

```
metadata.writeBack {
  default = [
      "dateTaken",
      "description",
      "credit",
      "creditUri",
      "byline",
      "bylineTitle",
      "title",
      "copyright",
      "suppliersReference",
      "source",
      "specialInstructions",
      "keywords",
      "subLocation",
      "city",
      "state",
      "country",
      "subjects",
      "peopleInImage"
  ],
  domain = []
}
```

This PR adds configurable behavior to choose which metadata to write back after cropping. The presence of this configuration will override the default behavior (Guardian behavior) which only writes back to master crop copyright, credit and supplier reference metadata. Instead it will write back all metadata specified in the config to master crop and all crop sizes.

This change does not affect the UI

Without the config, crop as usual. The metadata should only be present on the master crop. With the config, metadata should be present on all crop sizes.

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [ x ] locally by committer


paperboyo:
> The Guardian writes back the a small portion of the original metadata but only to master crop

I think we write this subset to all crops. I didn’t know it’s original metadata. Should include user overrides TBH!

Thanks for making it configurable! In general, couple of points (mostly, because I can’t see any actual writing code here (exiftool commands)):
1. what happens to **existing** crops if any of the metadata configured to be saved is changed (is it written to them again?)
2. Some of the fields won’t have obvious counterparts in known metadata standards (`creditURI`)
3. Some, will need careful [reconciliation back](https://github.com/guardian/grid/blob/ca2628a362da5b9de8ffcb0dcd38f3c52f9b3820/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala#L58-L101) into all standard fields (see useful [MWG composites](https://exiftool.org/TagNames/MWG.html))
4. This may be a bit confusing, but I think I’m correct. Writing `source` and `credit` back will possibly need some thought (in our syndication code, we write Grid’s `credit` to [`source`](https://www.iptc.org/std/photometadata/specification/IPTC-PhotoMetadata#source) and write Grid’s `byline`+`credit` into [`credit line`](https://www.iptc.org/std/photometadata/specification/IPTC-PhotoMetadata#credit-line)).

honorcb:
Our legal team and Data protection team,  wanted to keep the original credit , source and copyright,  but also wanted us not to pass on details  personal details, such as phone number that might have been put in those fields.

paperboyo:
> wanted to keep the original credit , source and copyright

Original? These fields are editable in Grid, what is the point of writing original values if they have been edited in Grid? Even more so if they wrongly contained personal info which has been edited out in Grid? Maybe we should discuss it at the next meeting? Seems to me writing Grid-held values (even when they overwrote original ones) should always be preferable? (and if we are now writing original ones instead of Grid-held ones, we are doing it wrong)

honorcb:
The issue from the legal team is we have the same image from various "suppliers"  image A from supplier Z ,  has licence to use by the BBC,  but image A from supplier Q does not have licence to use by the BBC,   and often the supplier has  edit the images original metadata,   It is variation from the supplier of this information legal want to see,  when answering questions  about which version of image A has been cropped and used.

paperboyo:
> It is variation from the supplier of this information legal want to see, when answering questions about which version of image A has been cropped and used

I think what I propose achieves that aim.
5. we already write `suppliersReference` which tries its best to gather individual suppliers’ identifiers (via individual [supplier processors](https://github.com/guardian/grid/blob/main/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala#L299), explained [here](https://github.com/guardian/grid/pull/3328#issuecomment-849080100), and then falling back to [standard fields](https://github.com/guardian/grid/blob/main/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala#L78-L79)). We are only writing it to old IPTC 2:103 Original Transmission Reference, we should probably also write it to newer [Job Id](https://www.iptc.org/std/photometadata/specification/IPTC-PhotoMetadata#job-id). 
6. Grid reads and stores all the mentioned fields agencies sent (it should never loose any data):
   a) Credit is largely cleaned up by supplier processors cleanups, it’s also editable
   b) Source is not editable and would stay unchanged, but I think we should actually do what I wrote in pt. 4 (rewrite it)
   c) Copyright comes as-is, but is editable

In short – I think the request is sensible, but the cleaned data (and possible user overrides) of it achieves the goal better and also doesn’t go against Grid being the source of truth (what’s the point of editing wrong data in Grid if the images in the wild will have old, wrong, data from suppliers?).

But if it’s absolutely required to keep data from suppliers intact in Grid crops, I propose to make the config for it also accept input (as well as output) fields, so BBC could have

`originalMetadata.credit` written into `xmp.photoshop:Credit` and `iptc.2:110 Credit`

while the Guardian would have

`metadata.credit` written into `xmp.photoshop:Credit` and `iptc.2:110 Credit`

Guardian would be more correct 😜.

paperboyo:
Oops! I forgot the cleanup is applied **before** data ends up under `originalMetadata`! So to catch exactly what agencies sent, BBC would need to do:

`fileMetadata.xmp.photoshop:Credit` or, if absent, `fileMetadata.iptc.credit` written into `xmp.photoshop:Credit` and `iptc.2:110 Credit`

paperboyo:
If the aim is to be able to identify the image in question, maybe we should just write Grid ID into crops? We would append Grid org (or configurable orgs) and registry ID from [this structure](https://www.iptc.org/std/photometadata/specification/IPTC-PhotoMetadata#registry-entry-structure).

paperboyo:
Oh, this is also related to what we are discussing with our image CDN. We would like Grid to write (and CDN to passthrough) these XMP fields and their IIM counterparts (although that’s just to play nice with the standards) for populating Google Images pages as [described here](https://iptc.org/standards/photo-metadata/quick-guide-to-iptc-photo-metadata-and-google-images/):

- [dc:creator](https://iptc.org/std/photometadata/specification/IPTC-PhotoMetadata#creator) (IPTC IIM 2:80 By-line)
- [photoshop:Credit](https://iptc.org/std/photometadata/specification/IPTC-PhotoMetadata#credit-line) (IPTC IIM 2:110 Credit)
- [dc:rights](https://iptc.org/std/photometadata/specification/IPTC-PhotoMetadata#copyright-notice) (IPTC IIM 2:116 Copyright Notice)
- [xmpRights:WebStatement](https://iptc.org/std/photometadata/specification/IPTC-PhotoMetadata#web-statement-of-rights)
- [plus:LicensorURL](http://ns.useplus.org/LDF/ldf-XMPSpecification#LicensorURL)

Fo the last two, we don’t have infra set up yet, or thought out TBH, nor do we have Grid UI etc to fill them in, but we think it would be great to have Google Images being able to display all this data. Including a “shopping link” to some images we sell (cartoons etc), which is last two fields.

PR #3845, opened 2022-08-23
[IMAGEDAM-1000] 

ornel-lloyd-edano:
either remove save image option from the context menu or disable context menu entirely (in case of crop view)

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/kahuna/public/js/crop/view.html
+++ b/kahuna/public/js/crop/view.html
@@ -103,8 +103,9 @@
 
 <div class="easel" role="main" aria-label="Image cropper">
     <div class="easel__canvas">
-        <div class="easel__image-container">
+        <div class="easel__image-container" oncontextmenu="return false;">
             <img class="easel__image easel__image--cropper"
+                 oncontextmenu="return false;"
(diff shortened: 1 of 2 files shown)


paperboyo:
I was only worried (potentially) about missing Open Link in New Tab on a search page, but seeing that it’s still there, I think this is eminently sensible! Thank you.

paperboyo:
Previously: https://github.com/guardian/grid/pull/2640.

jonathonherbert (approved):

paperboyo:
Hi, @ornel-lloyd-edano, just a note that we had to roll that back in production as it broke ability to drag multiple images into collections… (I missed that in testing :-( ).

ornel-lloyd-edano:
> Hi, @ornel-lloyd-edano, just a note that we had to roll that back in production as it broke ability to drag multiple images into collections… (I missed that in testing :-( ).

Hi @paperboyo I see. How about preventing the context menu to appear on right click? Would this break any UX?

PR #3848, opened 2022-08-30
restore image draggability

andrew-nowak:
A recent change to set images to `pointer-events:none` meant that the
images on the search page cannot be dragged, which breaks our support
for mass-dragging images into collections.

Instead of relying on the implicit draggability of images with css
pointer-events:auto, explicitly make the entire card draggable.
This may count as an improvement over the old behaviour as you can now
drag from anywhere inside the card, but you also cannot now select the
text in the card.

(Does this break anyone's workflow? You only get a few words here,
surely you would want to enter the full image before highlighting the
text?)

## Tested? Documented?
- [x] locally by committer
- [x] by @paperboyo 

### DIFF
--- a/kahuna/public/js/search/results.html
+++ b/kahuna/public/js/search/results.html
@@ -145,8 +145,9 @@
                     </div>
 
                     <ui-preview-image image="image"
+                                      draggable="true"
                                       selection-mode="ctrl.inSelectionMode"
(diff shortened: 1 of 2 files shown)


twrichards (approved):
approving the code, I know @paperboyo has tested in `TEST`

PR #3854, opened 2022-09-07
Add some teal to represent successfully leased images

andrew-nowak:
## What does this change?

Users have suggested that it would be useful to be able to see when an image has been leased more clearly. It is essentially visible on the image detail page (lease is large and visible, and the red danger banner goes yellow/warning once lease is added). But the flags on the search page and the status info boxes when multiple selecting could switch.

We've discussed colour and we are opposed to green (green should be reserved for free images only - paid/overquota/restricted images don't become free once leased) so we've picked *teal* `#008080` (positive but not *too* positive).

We've also removed the blue stripe on leases - it showed if a lease was active but no one knew what it meant. The lease now fades out when inactive.

Some pics:

image
image
image

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-info-panel/gr-info-panel.html
+++ b/kahuna/public/js/components/gr-info-panel/gr-info-panel.html
@@ -11,27 +11,35 @@
             <li ng-repeat="cost in ctrl.selectedCosts | orderBy:'data'">
                 <div ng-switch="cost.data">
                     <div ng-switch-when="free"
-                         class="image-notice image-info__group status cost cost--free">
+                         class="image-notice image-info__group status cost cost--free"
+                    >
                         {{cost.count}} free
                     </div>
 
                     <div ng-switch-when="no_rights"
-                         class="image-notice image-info__group status cost cost--pay">
+                         class="image-notice image-info__group status cost cost--pay"
+                    >
                         {{cost.count}} no rights
                     </div>
 
                     <div ng-switch-when="pay"
-                         class="image-notice image-info__group status cost cost--pay">
+                         class="image-notice image-info__group status cost cost--pay"
+                         style="{{ctrl.stylePercentageLeased(cost)}}"
+                    >
                         {{cost.count}} paid
                     </div>
 
                     <div ng-switch-when="overquota"
-                         class="image-notice image-info__group status cost cost--pay">
+                         class="image-notice image-info__group status cost cost--pay"
+                         style="{{ctrl.stylePercentageLeased(cost)}}"
+                    >
                         {{cost.count}} over quota
                     </div>
 
                     <div ng-switch-when="conditional"
-                         class="image-notice image-info__group cost cost--conditional">
+                         class="image-notice image-info__group cost cost--conditional"
+                         style="{{ctrl.stylePercentageLeased(cost)}}"
+                    >
                         {{cost.count}} restricted
                     </div>
                 </div>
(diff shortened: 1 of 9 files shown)


twrichards (approved):
code looks great! I haven't tried it on `TEST` but I will add @paperboyo as a reviewer (from a usability perspective).

Also, just wanted bring to the attention of @ochiengolanga and @ornel-lloyd-edano as this will have an impact on BBC.

paperboyo:
I like the idea of changing the copy on Inactive leases! Not sure how easy would it be to adjust it as in your example when the lease has not yet started (see above `Starts in a day, Expires in 2 days`). Maybe if that would be too hard, just sticking `Inactive: ` in front would be enough? @andrew-nowak?

image

Regarding the messaging, I agree that users do not understand all the complicated colours and badges. But this PR makes it clearer as currently the leased state is **not indicated at all** in these three examples where this applies (chargeable, over-quota and restricted). So we are not making anything worse, just universally better.

Also, this applies outside of selection state, so is more readily visible. And I don’t think it conflicts with your selection states work. I agree selection states overlays should indicate what this PR does in the more obvious way. We had reports of users asking for an already leased image to be leased and this should prevent that. A bit.

I think after the copy changes this should go in and we can see how it pans out in use and adjust if necessary. What do you think, @louisegoodspeed?

louisegoodspeed:
yes the language is confusing when it comes to the lease not having started yet. In response to 'inactive' = would simply 'not started' work? Will have a ponder. 
Agree, both the flag and selection state warnings would work well together. In terms of colours we probably need to agree either the teal or blue... as I think we should use the same colour unless BBC can use our own?

paperboyo:
> Can we change the green to red for an expired lease too?

Not really as two colours represent two states: allow and deny, so that would confuse things.

> I think we should use the same colour unless BBC can use our own?

Whenever we could, we should try not to bloat configuration, so I think we should and can agree.

Teal suggests more allowing than blue, but not as much as green as Andrew mentioned. Blue is also used as UI accent colour, so is neutral when it comes to messaging.

We can trade mockups in Slack, will be faster than here, if you prefer.

NickyPhelanBBC:
Hello, just chipping in re: colours.

Agree it would be good to reach an agreement if possible and the teal does look good with the red and yellow but unfortunately it's not going to work for us. Firstly we should really be using our enterprise colour palette (see attached) and secondly in the absence of a green elsewhere on the page the user is likely to think that the teal _represents_ a green, as part of a traffic light system as there are already red and yellow elements on the page.

We'll have a think our end about other possible options (purple?) and perhaps we can discuss at Grid Hour next week.

[Enterprise-colour-guide.pdf](https://github.com/guardian/grid/files/9528477/Enterprise-colour-guide.pdf)

paperboyo:
> unfortunately it's not going to work for us. Firstly we should really be using our enterprise colour palette

So, maybe, we need to lift **all** colours into config? That seems… extreme? And in any case seems a bigger job than this PR. The doc does contain

image

I’m super-happy using the teal values from your doc! Deal? ;-)

> the user is likely to think that the teal _represents_ a green

Kinda does represent it (an active Allow lease makes an image usable). But only kinda, because it doesn’t make it free. It definitely represents (or – should) a move on the traffic light spectrum: a move towards “a green end”. Note that there is already this green here we are replacing (and the blue nobody understood):

image

Worth also noting, the icons retain their symbols (a pound, a graph and a flag), so the dual nature is not (completely) lost. Random colour from outside a traffic light spectrum would make less sense, I think. What we are trying to say is less: _hey, there is a lease_ as _hey, this is allowed_.

Super-open to discussing on Slack/meeting, yeah!

PR #3855, opened 2022-09-09
tweaks to the imgops image to run on m1

andrew-nowak:
## What does this change?

- switch from debian to ubuntu (slightly more modern nginx, docker
  images are built natively for arm64)
- set DEBIAN_FRONTEND=noninteractive (prevents images getting stuck
  waiting for input during an apt-get)
- explicitly load the nginx image filter
- allow docker to run the container on the same cpu architecture as the
  host

Run grid locally. Does the imgops docker container spin up and work correctly?

Local development is fully supported on both amd64 and arm64 hosts.

### DIFF
--- a/dev/imgops/Dockerfile
+++ b/dev/imgops/Dockerfile
@@ -1,6 +1,7 @@
-FROM debian:jessie
+FROM ubuntu:focal
 MAINTAINER The Guardian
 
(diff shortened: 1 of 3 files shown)


twrichards (approved):
👏

andrew-nowak:
tested and container also launches and works on amd64 processor

PR #3864, opened 2022-09-23
Add logging contexts to cropper

andrew-nowak:
## What does this change?

Use the RequestLoggingFilter introduced in #3719 to get stable request IDs on cropper endpoints, and passes logMarkers through into functions defined in and used by cropper. It should be easier now to relate error logs in cropper to other logs created by the same request.

try to create a full-frame crop of https://media.test.dev-gutools.co.uk/images/a36e8475249d99bf45fce2f2288aef8ca8b91a18 , which should (currently) fail. Find an error log caused by this, then create a new search on the `requestId` marker - you should now see all related logs from that request.

## Tested? Documented?
- [ ] locally by committer


cropper/app/controllers/CropperController.scala
```diff
@@ -52,23 +50,44 @@ class CropperController(auth: Authentication, crops: Crops, store: CropStore, no
...
+          respondError(BadRequest, "no-source-image", MissingSecureSourceUrl.getMessage)
+        case InvalidCropRequest =>
+          logger.error(logMarker, InvalidCropRequest.getMessage)
+          respondError(BadRequest, "invalid-crop", InvalidCropRequest.getMessage)
+        case ApiRequestFailed =>
+          logger.error(logMarker, ApiRequestFailed.getMessage)
+          respondError(BadGateway, "api-failed", ApiRequestFailed.getMessage)
+        case NonFatal(e) =>
+          logger.error(logMarker, s"export failed with exception", e)
+          respondError(InternalServerError, "unknown-error", e.getMessage)
```
twrichards, line 90: is there a reason we've not moved the logger line into `respondError`

andrew-nowak: `respondError` is shared between all of the services (common-lib, part of the Argo helpers), and I wasn't prepared to audit that logging from it worked for every service in this PR (eg. do all have an implicit logMarkers in scope? are we duplicating other logging? etc.)

One day we'll get all of the services behaving in a common way...

PR #3863, opened 2022-09-23
introduce `--no-auth` flag for `setup.sh` and `start.sh` scripts for easier local development

twrichards:
https://trello.com/c/9qsSL3Am

https://github.com/guardian/grid/pull/3226 was quite confusing and undocumented, here we introduce a new flag `--no-auth` for the `setup.sh` and `start.sh` scripts (which replaces $LOCAL_SIMPLE_AUTH_PROVIDER and better describes the behaviour) and it now changes the auth providers in the generated config rather than re-writing the `application.conf` file (since that could easily get committed).

_NOTE: The `--with-local-auth` is still potentially broken (I couldn't get it working, and other users outside Guardian have reported problems) - I've created https://trello.com/c/3cF1guts/2825-fix-improve-with-local-auth-option-in-grid-setup-start to revisit this. **EDIT:** this looks to have been picked up in https://github.com/guardian/grid/pull/3865_

PR #3865, opened 2022-09-23
get local auth running again

andrew-nowak:
## What does this change?

A couple of places were looking for config on the wrong object / under the wrong key, and oidc-provider seems to now need PKCE disabling to work with grid locally. 

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/dev/oidc-provider/app.js
+++ b/dev/oidc-provider/app.js
@@ -55,6 +55,9 @@ const oidc = new Provider(issuer, {
     openid: [ "sub", "email", "email_verified", "given_name", "family_name", "name" ]
   },
   proxy: true,
+  pkce: {
+    required: () => false
+  },
   findAccount
 });
 
--- a/dev/script/generate-config/service-config.js
+++ b/dev/script/generate-config/service-config.js
@@ -198,7 +198,7 @@ function getGridProdConfig(config) {
   return stripMargin`
     |auth.useLocal=true
     |panda.userDomain="${config.EMAIL_DOMAIN}"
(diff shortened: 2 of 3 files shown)


twrichards:
Amazing, looking forward to reviewing when you're ready. I had made https://trello.com/c/3cF1guts/2825-fix-improve-with-local-auth-option-in-grid-setup-start to cover this, I'll link that card to this ticket. Please see https://github.com/guardian/grid/pull/3863#issuecomment-1256133353 though

twrichards (commented):
code looks good but trying it out locally I get 
image
after signing in with `grid-demo-account@guardian.co.uk` -

NOTE that I forgot the `@guardian.co.uk` first time which signed me in but with an invalid user, I then had to hit http://localhost:9014/.well-known/openid-configuration to find the endpoint for signing out which is http://localhost:9014/session/end. What made me realise I was missing `@guardian.co.uk` was the logs on the `oidc-provider` docker container which ALSO warns about cookie issues which might be related to the above...
image

twrichards (approved):
After pairing with @andrew-nowak we cleared cookies on both `local.dev-gutools.co.uk` and `localhost` and signed-in again... all good. We think stale stuff left behind from using regular/Guardian panda before was the cause (Incognito was fine also obviously, didn't think of that first time).
After a little discussion we're content with this, given Guardian people don't generally use local-auth (as our local Guardian panda login is superior) and devs outside the Guardian will  not have any stale Guardian panda cookies. 🎉

PR #3869, opened 2022-09-26
UI changes on the grid view/image preview tile

ornel-lloyd-edano:
Change size and layout of icons inside an image preview tile. Add label "Uploaded:" before datetime

## What does this change?

No new or changed functionality, just minor UI related changes on the grid view image tile as per discussions during Grid Hour
- date time below an image preview was added with label "Uploaded:"
- warning and restriction icons are moved on the next row after the upload time label
- warning and restriction icons are slightly bigger (20px) than normal icons (16px was the normal icon size)

New Look
Screen Shot 2022-09-26 at 1 10 28 PM

Edit: Fixed layout issues (color changes are as per BBC custom config)
Screen Shot 2022-10-14 at 7 13 26 PM

Edit: Reduced the overall height of a preview tile by 30px from the last PR. Now it is only 10px taller than the original
Screen Shot 2022-10-28 at 1 28 25 PM

More can be seen from the third row, just like original
Screen Shot 2022-10-28 at 1 36 39 PM

Nothing to configure, just run and observe the label and new look of the icons

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-syndication-icon/gr-syndication-icon.html
+++ b/kahuna/public/js/components/gr-syndication-icon/gr-syndication-icon.html
@@ -1,5 +1,5 @@
 <div title="{{ ctrl.states.syndicationReason }}">
-    <gr-icon class="syndication-status--{{ ctrl.states.syndicationStatus }}">
+    <gr-icon style="font-size: 20px;" class="syndication-status--{{ ctrl.states.syndicationStatus }}">
         monetization_on
     </gr-icon>
 </div>
--- a/kahuna/public/js/components/gr-syndication-icon/gr-syndication-icon.html
+++ b/kahuna/public/js/components/gr-syndication-icon/gr-syndication-icon.html
@@ -1,5 +1,5 @@
 <div title="{{ ctrl.states.syndicationReason }}">
-    <gr-icon style="font-size: 20px;" class="syndication-status--{{ ctrl.states.syndicationStatus }}">
+    <gr-icon class="gr-icon--large" class="syndication-status--{{ ctrl.states.syndicationStatus }}">
         monetization_on
     </gr-icon>
 </div>
--- a/kahuna/public/js/components/gr-icon/gr-icon.css
+++ b/kahuna/public/js/components/gr-icon/gr-icon.css
@@ -37,6 +37,10 @@ gr-icon .gr-icon--small {
     font-size: 1.2rem;
 }
 
+.gr-icon--large {
+    font-size: 20px;
+}
+
 gr-icon-label .icon-label {
     vertical-align: middle;
 }
--- a/kahuna/public/js/preview/image-large.html
+++ b/kahuna/public/js/preview/image-large.html
@@ -99,7 +99,7 @@
         <div class="preview__bottom-bar bottom-bar preview__bottom-bar--large">
             <div class="bottom-bar__meta">
                 <span class="preview__upload-time">
-                    {{ctrl.image.data.uploadTime | date:'dd/MM/yy'}}
+                    Uploaded: {{ctrl.image.data.uploadTime | date:'dd/MM/yy'}}
                     {{ctrl.image.data.uploadTime | date:'HH:mm'}}
                 </span>
 
--- a/kahuna/public/js/preview/image.html
+++ b/kahuna/public/js/preview/image.html
@@ -99,75 +99,69 @@ Credit: {{ctrl.image.data.metadata.credit  || '[none]'}}">
             </span>
         </div>
     </div>
+    <div class="preview__bottom-icons">
+        <span class="bottom-bar__meta-item preview__has-crops"
+              title="this image has crops"
+              ng-if="ctrl.states.hasCrops">
+                <gr-icon style="font-size: 22px;">crop</gr-icon>
+        </span>
 
-    <div class="preview__info-flex">
-        <div class="preview__info-inline">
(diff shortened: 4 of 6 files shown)


paperboyo:
Hello, @ornel-lloyd-edano. Testing in our TEST env, I can see badges cut off in both Firefox 105 and Chrome 105 independent on zoom level:

image

paperboyo:
Thanks for all the fixes! Looks much better. I like the consistency in icon sizes.

But I do have one worry. We have made cells much taller (286×286 vs. 288×323). It results in almost the whole row of images missing on a common screen res/zoom level:

image

I must admit I’m slightly terrified someone will ask me: _oi! I scroll images all day long, for years, and you have added one bit of info I already knew while adding literally kilometres to my yearly scroll. Why would you ever do such a thing?!_.

To be able to answer this with (slightly more) confidence, I observe the below screenshot and note:

image

- we have increased vertical spacing even before Description line (?), it’s now lower
- we have increased the vertical spacing between Description and the Date, even though before it was the icons that were closer to Description and now it’s a Date line
- we have increased bottom margin of the card (below icons)
- we have increased vertical spacing between the cards (see first screenshot) yes, it’s more consistent with horizontal spacing, but adds to this problem

I think (hope!) we can remove all these increases to recover some of the vertical space. I understand maybe some of them were added to give some breathing room, but it’s worth noting, most of the time cells are quite empty (see first screenshot), it’s only rarely that they contain all the badges, labels etc. The After in the first screenshot does look a bit like it’s wasting space especially below the dateline (I know this is where the icons would be, but this is how it looks most of the time for majority of users)…

Adding @louisegoodspeed, because we’ve discussed it a bit on Slack. Sadly, I don’t have Andrew’s github handle (will ping him there)…

louisegoodspeed:
Screenshot 2022-10-19 at 09 45 26
@andrew-nowak you mocked up a version with a different height shown in this screengrab, would you mind giving us the pixel height please so @drewcochrane can work to that?

paperboyo:
Hello, @louisegoodspeed, my fault for sharing @andrew-nowak’s mockup without him knowing which was only to suggest we can increase card height if there is no other way to incorporate extra info.

Basically, we need to balance two things:
1. Users being confused as to the nature of the date shown (serious issue, I agree, but once learned, will require them to be seeing it indefinitely, which will make much more sense once they will be able to configure it to show Taken too)
2. Decreased number of pics per page shown and extra scrolling for users who scroll their whole days (not all, but mainly news)

That’s why I think as little increase in height as possible would be best. Now, here is yet another mockup that **should not be treated as final**. It keeps the current height. It definitely looks cramped (with the caveat that most of the time, cells only have Description and date, nothing else!).

What do you think? If it’s all too tight, we can introduce extra breathing space where you think, but I would still think it makes sense to be somewhere in between this below and what this PR has. Again, as close to current height as possible to kill both birds without tripping up on the slingshot:

With the rules showing alignment to current (on the left):

image


100% non-retina PNG:

a196007491-6fedbdaa-dbe7-403a-9af7-77ee55e2bb8f

louisegoodspeed:
@paperboyo I think I'm just going to let @drewcochrane do his thing, I'm sure he'll be able to sort it.

drewcochrane:
Hey @ornel-lloyd-edano can you confirm that the typical image tile is 324px x 272px with a 5px border? This is based on inspecting via a browser inspection tool so that I can explore how much closer I can get to the original size as possible via Sketch.

ornel-lloyd-edano:
> Hey @ornel-lloyd-edano can you confirm that the typical image tile is 324px x 272px with a 5px border?

@drewcochrane the original was 272px by 285.5px with 5px margin. After the suggested changes it becomes 272px by 320px with 5px margin. I can try to reduce the space between the description and uploaded, make the icons and main image just a bit smaller and lets see how close to the original height we can get without looking weird

ornel-lloyd-edano:
Hi @paperboyo my recent commit to this PR, reduces 30px to the increased height. It should be now 10px taller than the original

paperboyo:
Thank you, @ornel-lloyd-edano! This tested in our TEST env looks good now, I think:

longer cards

kahuna/public/js/components/gr-syndication-icon/gr-syndication-icon.html
```diff
@@ -1,5 +1,5 @@
 <div title="{{ ctrl.states.syndicationReason }}">
-    <gr-icon class="syndication-status--{{ ctrl.states.syndicationStatus }}">
+    <gr-icon style="font-size: 20px;" class="syndication-status--{{ ctrl.states.syndicationStatus }}">
```
andrew-nowak, line 2: We try to avoid using style where possible. There is already an option in gr-icon to set a small font-size (`<gr-icon gr-small>...`), perhaps we could add a `gr-large` attribute which will set a font-size of 20px

ornel-lloyd-edano: Hi @andrew-nowak I am having difficulties adding a gr-large attribute to the component, somehow it affects all the gr-icons. However I removed the use of style attribute with class attribute. Is this acceptable?

andrew-nowak: thanks!

andrew-nowak (approved):

Issue #3849, opened 2022-09-01
Auto-link metadata keys from additional metadata section


paperboyo:
## Short description
Just as for all metadata, keys from under Show additional metadata should be links to searches:

<img width="280" alt="image" src="https://user-images.githubusercontent.com/6032869/187917417-5c15aadf-9104-4866-af4a-2c8ab8b4c67e.png">

Please note, that as of https://github.com/guardian/grid/pull/3588, this section may contain keys supplied by configuration.

PR #3871, opened 2022-09-28
Searchlinks in Additional metadata and more consistent Info panel UI

paperboyo:
## What does this change?
_Co-authored by @jonathonherbert_
_Co-authored by @andrew-nowak_

Fixes https://github.com/guardian/grid/issues/3849:
Links values for metadata fields visible under Show additional metadata section to searches bringing them in line with the rest of the Info panel. Please note that identifiers have no links (because they are either unique or defunct). Arrays and other structured fields do not work, but their display is already broken anyway. Both should one day be using our pills (like eg. people) for arrays and forms (like eg. Location) for structured fields. Also, there is no reason why Additional metadata UI could not be displayed and work for multiple selections…

image

For consistency , it also allows searching for multiple images with the same Title by linking from the Title field. Linking from Description makes either no or very little sense. I decided not to link from hour and/or day either. Not sure about Special Instructions, but decided on no.

| Before  | After |
| ------------- | ------------- |
| image | image |

Finally, it brings some minor styling improvements which make the panel more consistent:
1. _Multiple_ values are always italicised (fixed Description and Date taken)
2. makes Photoshoot value look like any other
3. hides non-required fields when they are empty and user can’t edit (Location, Filename and People) (previously some were hidden, but these were not)
4. adds Unknown placeholder for empty required Credit field and for empty (but important) Taken field
5. It uses Title Case for three additional metadata fields (`bylineTitle`, `suppliersReference` and `source`) and for identifiers

## Screenshots

| No. | Before  | After |
| ------------- | ------------- | ------------- |
| 1  | image  | image  |
| 2  | image  | image  |
| 3&4 | image | image |
| 5 | image | image |

Try to search by clicking on Title and values in Additional metadata. Look around for styling fixes shown above.

Editors can discover image groupings by easily searching on Title.
Editors can discover more ways of searching.
Less chaos in the info panel, no headings without values, more consistent styling.

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
@@ -101,10 +101,9 @@
                         ng-hide="descriptionEditForm.$visible">✎</button>
                 <dt class="metadata-line metadata-line__key">Description</dt>
                 <form editable-form name="descriptionEditForm" onaftersave="ctrl.updateDescriptionField('description', $data)">
-                  <div ng-hide="descriptionEditForm.$visible">
+                  <div ng-hide="descriptionEditForm.$visible" ng-class="{'image-info--multiple': ctrl.hasMultipleValues(ctrl.rawMetadata.description)}">
                     <div ng-if="ctrl.userCanEdit">
                         <dd class="image-info__description"
-                            ng-class="{'image-info--multiple': ctrl.hasMultipleValues(ctrl.rawMetadata.description)}"
                             ng-if="ctrl.hasMultipleValues(ctrl.rawMetadata.description)"
                             >Multiple descriptions (click ✎ to edit <strong>all</strong>)
                         </dd>
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
@@ -567,8 +567,8 @@
                 </span>
             </dd>
 
-            <dt ng-if="ctrl.metadata.peopleInImage.length > 0 || ctrl.userCanEdit" class="image-info__wrap metadata-line metadata-line__key image-info__group--dl__key--panel">People</dt>
-            <dd ng-if="ctrl.metadata.peopleInImage.length > 0 || ctrl.userCanEdit" class="image-info__people image-info__wrap metadata-line metadata-line__info image-info__group--dl__value--panel">
+            <dt ng-if="ctrl.rawMetadata.peopleInImage.length > 0 || ctrl.userCanEdit" class="image-info__wrap metadata-line metadata-line__key image-info__group--dl__key--panel">People</dt>
+            <dd ng-if="ctrl.rawMetadata.peopleInImage.length > 0 || ctrl.userCanEdit" class="image-info__people image-info__wrap metadata-line metadata-line__info image-info__group--dl__value--panel">
                 <button data-cy="it-edit-people-button"
                         class="metadata-people__button"
                         ng-class="{'small': ctrl.grSmall}"
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
@@ -598,12 +598,18 @@
         <div ng-if="!ctrl.isMetadataSectionHidden('additionalMetadata')" class="metadata metadata-line image-info__group--dl">
             <dl ng-repeat="(key, value) in ctrl.metadata" ng-if="ctrl.isUsefulMetadata(key)" class="metadata__body image-info__group--dl">
                 <dt class="metadata-line__key image-info__group--dl__key--full-metadata">{{key | spaceWords}}</dt>
-                <dd class="metadata-line__info image-info__group--dl__value--full-metadata">{{value}}</dd>
+                <dd class="metadata-line__info image-info__group--dl__value--full-metadata">
+                <a ui-sref="search.results({query: (value | queryFilter:'{{key}}'), nonFree: ctrl.srefNonfree()})"
+                aria-label="Search images with {{key}} '{{value}}'">{{value}}</a>
+                </dd>
             </dl>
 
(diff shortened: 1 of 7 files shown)


kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
```diff
@@ -219,10 +219,10 @@ module.controller('grImageMetadataCtrl', [
 
       ctrl.additionalMetadata = Object.fromEntries(
         Object.entries(ctrl.singleImage.data.aliases)
-          .map(([key, val]) => {
+          .map(([key, value]) => {
             let fieldAlias = ctrl.fieldAliases.find(_ => _.alias === key);
             if (fieldAlias && fieldAlias.displayInAdditionalMetadata === true) {
-              return [fieldAlias.label, val];
+              return [fieldAlias.label, {value,elasticsearchPath: fieldAlias.elasticsearchPath}];
```
andrew-nowak, line 225: you could try this (complete guess, no idea what angular will try to do with it)
```suggestion
              return [fieldAlias.label, { value, elasticsearchPath: '"' + fieldAlias.elasticsearchPath + '"' }];
```

paperboyo: Thanks! It makes it much better, but still, under certain scenarios, a good search flashes and as if aftyer a reload ui-router (?) messes things up…

aracho1:
I was running this on TEST and noticed the following error in the console:
image

which is pointing to this [line](https://github.com/guardian/grid/pull/3871/files#diff-7661cb91c09677bedfa8e6a6a0b1c081a8502a387a7554e7d9ac28edc1159f36R623). I noticed that the values in the XMP Keywords field appear as an array of strings but they are not actually clickable, is this the expected behaviour? 
image

If not, I wonder if the query is failing in this field potentially, as the other additional metadata fields seem to be working fine.

paperboyo:
> is this the expected behaviour?

Very good spot, @aracho1. This was hinted at in the PR description

> Arrays and other structured fields do not work, but their display is already broken anyway

Arrays and other structured fields have currently no support to show properly. This PR adds “links” to them which do not work. All of those fields would need to be consciously added via `fieldAliases` config (like `XMP keywords` is on our TEST).

It is a pity, but to add support to display them properly (and then fix the linking for them) is a job too clever for this contributor (I have tried). I think it is acceptable as is, since someone would need to configure them to be shown knowing there is no support. But if you think not, I may have a look at adding code that disables linking for arrays and structured fields. Again, much better would be to make them work, but here I would need help from someone…

paperboyo:
Last commit that deals with multiple non-matching Location fields for unpermissioned users reveals we have an ongoing problem with People for unpermissioned users (in current PROD):

image
[_Apply person to all_ and _Remove_ [should say _person from all_] should not be shown for unpermissioned users as clicking them doesn’t do anything; football pic has no People]

This PR makes it look like this:

image
[both images have exactly the same two People]

So we need to fix it here, I guess? I think we need to make current Keywords pills [non-editable ones] work for unpermissioned users in People. Alternatively, I should revert changes to People that will make them as broken as shown for PROD above.

Sadpanda.

paperboyo:
In the end, I fixed the People for unpermissioned users so that it correctly shows up if any of the selected images have any People.

The fact that those umpermissioned users are seeing _Apply/remove person to/from all_ is a preexisting issue.

kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
```diff
@@ -333,14 +347,21 @@
...
-            <dd class="image-info__wrap metadata-line metadata-line__info image-info__group--dl__value--panel">
+            <dt ng-if="(ctrl.hasLocationInformation() || ctrl.hasMultipleValues(ctrl.rawMetadata.subLocation) ||
+                            ctrl.hasMultipleValues(ctrl.rawMetadata.city) ||
+                            ctrl.hasMultipleValues(ctrl.rawMetadata.state) ||
+                            ctrl.hasMultipleValues(ctrl.rawMetadata.country)) || ctrl.userCanEdit" ng-hide="locationEditForm.$visible" class="image-info__wrap metadata-line metadata-line__key image-info__group--dl__key--panel">Location</dt>
+            <dd ng-if="(ctrl.hasLocationInformation() || ctrl.hasMultipleValues(ctrl.rawMetadata.subLocation) ||
+                            ctrl.hasMultipleValues(ctrl.rawMetadata.city) ||
+                            ctrl.hasMultipleValues(ctrl.rawMetadata.state) ||
+                            ctrl.hasMultipleValues(ctrl.rawMetadata.country))  || ctrl.userCanEdit" class="image-info__wrap metadata-line metadata-line__info image-info__group--dl__value--panel">
```
aracho1, line 364: I'm probably missing something - what is the reason for checking `ctrl.hasMultipleValues` on the `city`, `state`, and `country` separately to `hasLocationInformation()`? Is the location information returned on `rawMetadata` different from `metadata`?

paperboyo: Sorry, but the honest answer is: I don’t know. This was kindly suggested by @andrew-nowak after it wasn’t working correctly before. All fault mine, obviously…

aracho1: No worries, I thought I'd check since I understood `ctrl.hasMultipleValues` was being used elsewhere to display `Multiple something` in the field but didn't seem like that's what we want for the location fields

andrew-nowak: thanks! the difference (_I think_) is that previously the location field was always shown, whereas we now want to not show the location field if there's no data and the user cannot edit. In an ideal world the `hasLocationInformation` would reliably tell us if there's location info or not, but currently it only works on single image selection. We could update it to understand multiple image selection, but it is also used elsewhere and we don't want/have time to untangle right now.

(The highlighted code is similar to the existing https://github.com/guardian/grid/blob/ca2628a362da5b9de8ffcb0dcd38f3c52f9b3820/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html#L372-L377)

aracho1 (approved):

PR #3880, opened 2022-10-05
customizing warnings and restrictions messages

ornel-lloyd-edano:
add configuration to customize warning and restriction text for labels and messages

## What does this change?

BBC wants to change the default warning messages for its users. This PR adds configuration to change the warning text and messages displayed on the metadata panel when an image is selected. **FYI The change of colors shown in the screenshot is a BBC only configuration and is not included in this PR**

Screen Shot 2022-10-06 at 4 11 25 PM

Screen Shot 2022-10-06 at 4 10 11 PM

Screen Shot 2022-10-06 at 5 53 19 PM

In application.conf, add the following configuration

> warningText {
>     prefix = "This image has warnings attached."
>     prefixNoRights = "Do not use this image."
>     unusablePrefix = "Do not use this image."
>     validityDescription {
>         no_rights = "No rights to use. Contact your appropriate rights management team to check rights status."
>         missing_credit = "This image requires credit information."
>         missing_description = "This image requires a description."
>         paid_image = "This image is a payable image. There are no rights to use unless an allow lease has been set. See lease information below. If there is no lease or the lease has expired, contact your appropriate rights management team to negotiate a licence to use."
>         over_quota = "The image quota downloaded for this organisation has been exceeded."
>         conditional_paid = "This image has a restriction applied to it, read more in the restricted use section."
>         current_deny_lease = "This image has been denied a lease and cannot be used. See further details in the lease section below."
>     }
> }

Select any image with warnings (ex. no rights, missing description). The custom messages from the above configuration should appear instead of the default.

Without this configuration, the default warning messages will stay the same.

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/app/lib/KahunaConfig.scala
+++ b/kahuna/app/lib/KahunaConfig.scala
@@ -51,5 +51,9 @@ class KahunaConfig(resources: GridConfigResources) extends CommonConfig(resource
 
   val metadataTemplates: Seq[MetadataTemplate] = configuration.get[Seq[MetadataTemplate]]("metadata.templates")
 
+  //BBC custom warning text
+  val warningTextPrefix: String = configuration.getOptional[String]("warningText.prefix").getOrElse("")
+  val warningTextPrefixNoRights: String = configuration.getOptional[String]("warningText.prefixNoRights").getOrElse("")
+  val unusableTextPrefix: String = configuration.getOptional[String]("warningText.unusablePrefix").getOrElse("")
 }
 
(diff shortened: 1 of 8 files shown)


kahuna/app/lib/KahunaConfig.scala
```diff
@@ -52,5 +52,9 @@ class KahunaConfig(resources: GridConfigResources) extends CommonConfig(resource
 
   val metadataTemplates: Seq[MetadataTemplate] = configuration.get[Seq[MetadataTemplate]]("metadata.templates")
 
+  //BBC custom warning text
+  val warningTextPrefix: String = configuration.getOptional[String]("warningText.prefix").getOrElse("")
```
andrew-nowak, line 56: I think 'header' describes these better than 'prefix'

andrew-nowak: plus, can we put the defaults in the getOrElse here? then the UI doesn't have to worry about it; should make the js code simpler

ornel-lloyd-edano: makes sense

andrew-nowak (approved):

PR #3892, opened 2022-10-11
Update the API key docs

andrew-nowak:
## What does this change?

Ronseal.
Also includes update to get-stack-resource script to ensure running, and a new script mkapikey to replace the command in the docs, which generates longer, securer api keys


phillipbarron (approved):
LGTM 👍

PR #3895, opened 2022-10-14
replace git deps in package.json with npm/tarball dependencies

andrew-nowak:
## What does this change?

Speeds up the teamcity build :D 12-13 mins reduced back to the good old 5-6 mins
npm apparently doesn't like installing dependencies with the git protocol, so stop doing that.

angular-ui-router is available from npm registry, can't see a clear reason why we were installing from github 
our fork of titip is not published to npm registry, but you _can_ pass a link to the tarball that github generates for a tagged commit, which works really nicely!

buildtimes:

| before | after |
|--------|-------|
| image | image |

Tested on TEST, routing and tooltips packages seem to be imported and working correctly


mchv:
Great work @andrew-nowak

twrichards (approved):

twrichards (approved):
brilliant!

PR #3896, opened 2022-10-17
add a `--use-TEST` option to `start.sh` to run only `auth`, `kahuna` and `media-api` and to use `TEST` for everything else (including ElasticSearch)

twrichards:
Co-authored-by: @andrew-nowak 
Co-authored-by: @paperboyo 

Most of our development on the grid centres around `kahuna` (the UI) and `media-api` (the API) so it's a pain to...

1. have to spin-up docker with its 5 containers
2. upload images every time (including having to delete the containers between machine/docker restarts, since persistence in localstack is a paid-for feature)

... and for when you need to more than a handful of images (e.g. implementing search changes) it's super painful. SO, why don't we just connect to `TEST` for ElasticSearch and all the other micro-services (except auth, which we want to run locally too). That's what this PR does!

PLUS refactor loading of CloudFront private key to load from S3 once (using new config properties `cloudfront.private-key.bucket` and `cloudfront.private-key.key`)

### DIFF
--- a/dev/script/start.sh
+++ b/dev/script/start.sh
@@ -31,6 +31,10 @@ for arg in "$@"; do
     LOCAL_AUTH=true
     shift
   fi
+  if [ "$arg" == "--use-TEST" ]; then
+    USE_TEST=true
+    shift
+  fi
 done
 
 isInstalled() {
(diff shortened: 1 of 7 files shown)


dev/script/start.sh
```diff
@@ -84,7 +88,21 @@ checkRequirements() {
...
 startDockerContainers() {
-  docker-compose up -d
+  if [[ $USE_TEST == true ]]; then
+    docker-compose down
+    EXISTING_TUNNELS=$(ps -ef | grep ssh | grep 9200 | grep -v grep || true)
+    if [[ -n $EXISTING_TUNNELS ]]; then
+      echo "RE-USING EXISTING TUNNEL TO TEST ELASTICSEARCH (on port 9200)"
+    else
+      TUNNEL_OPTS="-o ExitOnForwardFailure=yes -o ServerAliveInterval=10 -o ServerAliveCountMax=2"
+      SSH_COMMAND=$(ssm ssh --profile media-service -t elasticsearch-data,grid-elasticsearch,TEST --newest --raw)
```
andrew-nowak, line 98: not blocking, but it would be great to figure out a way to clean up the tunnel when stopping, which would stop accidentally connecting to the cluster (since the tunnel is on the default ES port)

andrew-nowak (approved):

dev/script/start.sh
twrichards: ahh in case we switch back to running in normal local mode?

andrew-nowak: yep, or anything else which connects to 9200

twrichards: done!

andrew-nowak (approved):

PR #3902, opened 2022-10-24
Do not show invalid images as no_rights in costState

andrew-nowak:
## What does this change?

I did this in #3854 to try and override all other costs when an image is
missing credit or description (one definition of "valid"). Instead it
overrides all other costs whenever an image is uncroppable (restricted,
overquota, etc., the other definition of "valid").

costState is used to determine both which flags to show on the search
results page, and also the warnings/error messages on the image page,
which means that this change hid the restriction text from ALL
unpermissioned users! So revert this logic now, and commit to finding a
better way of applying the invalid flag in the future.

Look at an image with restrictions while without the edit_metadata permission. Can you see the restriction text? If you view the image in the search results page, you will unfortunately see only a yellow "restricted" flag, rather than a more descriptive red "!" flag.

### DIFF
--- a/kahuna/public/js/image/service.js
+++ b/kahuna/public/js/image/service.js
@@ -28,7 +28,7 @@ imageService.factory('imageService', ['imageLogic', function(imageLogic) {
             cost,
             hasCrops: image.data.exports && image.data.exports.length > 0,
             hasRights,
-            costState: hasRights && image.data.valid ? cost : "no_rights",
+            costState: hasRights ? cost : "no_rights",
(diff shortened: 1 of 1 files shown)


paperboyo (approved):
Code-blind approval (of functionality): 👍.

A card to make Allow lease teal not override invalid imagery (missing description and/or credit) [here](https://trello.com/c/xEVZ6Mgn/888-teal-from-allow-cropping-ease-not-yielding-to-missing-credit-and-description).

PR #3907, opened 2022-10-28
Fix the edit field buttons for title, description + special instructions

andrew-nowak:
## What does this change?

it turns out that the edit buttons work only when siblings of the form they control

Deploy to TEST and successfully edit all metadata fields

## Tested? Documented?
- [x] locally by committer


twrichards (approved):
LGTM

PR #3908, opened 2022-10-28
make keywords editable in UI

twrichards:
Previously keywords were readonly (despite the fact metadata service technically supports them being edited).

In the PR we re-use the pattern & components used for the `People` field, for keywords - to make them editable.

| Before | After |
| --- | --- |
| image | image |

This involved a little refactoring to make the add/remove logic generic to avoid repetition (UI parts could probably be wrapped in their own directive).

PLUS made sure that the remove button beside editable pills is not visible if the user cannot edit (i.e. they're not the uploader or they don't have `EditMetadata` permission) - to avoid it just breaking when clicked by users who cannot edit.

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/MappingTest.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/MappingTest.scala
@@ -35,7 +35,7 @@ object MappingTest {
     suppliersReference = Some("ASDFFF"),
     source = Some("bill-withers"),
     specialInstructions = Some("This is really important."),
-    keywords = Some(List("sunset", "loveliness", "distance")),
+    keywords = Some(Set("sunset", "loveliness", "distance")),
     subLocation = Some("Downtown"),
     city = Some("LA"),
     state = Some("California"),
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/ImageMetadataConverter.scala
@@ -42,13 +42,13 @@ object ImageMetadataConverter extends GridLogging {
     Some((xmpIptcPeople ++ xmpGettyPeople).toSet)
   }
 
-  private def extractKeywords(fileMetadata: FileMetadata): Option[List[String]] = {
-    val fromXMP =  extractXMPArrayStrings("dc:subject", fileMetadata).toList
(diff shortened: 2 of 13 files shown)


twrichards:
✅  TESTED on `TEST`

paperboyo:
💎 Thank you! (also for fixes to generalised pill UI!)

Looks great for a permissioned user. For unpermissioned one, I found two issues:
1. Keywords completely disappear for multiple pic select (they should work like People: show what’s common, what’s not common, just without any editing UI bits)
2. People (and probs Keywords too after pt. 1 is fixed), still show Apply to all editing UI bit even though user cannot perform this action for uncommon values (values common for a selection correctly have no editing UI). Pills should look like they do (indicating common/uncommon), but contain no X and no -/+ UI bits when users cannot edit.

twrichards:
@paperboyo all three points done (and deployed to `TEST`)

As for point three, it was only the right hand side which was missing the radius, which makes sense as the radius was on the remove button (now hidden when not editable).

paperboyo:
Much betterer thank you! I think CSS for partial coverage regressed a bit (so it’s harder to distinguish it from common coverage for permissioned users and impossible to do this for unpermissioned ones). Compare:
4. Partial coverage on PROD:
    image
    Partial coverage on TEST (also hover state on a `+` button is wrong: not shown here):
    image

Pt. 3 still occurs: pills have hard edges on the right (where they lost an `X` button).

twrichards:
@paperboyo can't get into said regression right now, but I didn't test my change w.r.t. to point 3, had applied the rounding CSS to the wrong element (hence it was still unsolved) I've done it right this morning - please see on `TEST`

paperboyo:
🎩,  thanks:

image

Only pt. 4 remains, I think.

paperboyo:
Pt. 4 works correctly now. Also, additionally (thank you!), for multiple instances os a value, tooltip says _Remove keyword from all_ 👏

andrew-nowak:
it looks like keywords don't get deduplicated, meaning you can add a keyword multiple times (either by using a add-to-all button, or by explicitly adding a keyword that already exists to an image) 
image
not quite a problem by itself, except when then selecting that image and an image without the keyword you don't get the option to add to all 
eg. when selecting the above image and an image with 0 keywords:
image

I think the people field gets around this bug by treating people as a Set, I wonder if there's any reason we can't do the same for keywords

twrichards:
great spot @andrew-nowak - changed to use `Set` rather than `List` in https://github.com/guardian/grid/pull/3908/commits/c1734897d8bac7193147ce2ae382aa5061a16953

andrew-nowak (approved):

PR #3915, opened 2022-11-01
support the persisted query param

andrew-nowak:
## What does this change?

Allow to do search for reapable images by adding `persisted=false` (or `true`) in address bar
(I did briefly look at adding to chips, but is definitely harder than query-param :( )

Deploy to TEST, add the param, see the results be filtered accordingly

## Tested? Documented?
- [ ] locally by committer


rhystmills (approved):
Tested on TEST and looks to be working as intended 👍

PR #3923, opened 2022-11-02
FIX `Labels` should be editable by anyone in UI

twrichards:
tiny fix to #3908 to re-instate normal removable behaviour for labels in the UI


andrew-nowak (approved):
untested, but makes sense to me 👍

twrichards:
✅  TESTED on `TEST` (without `EditMetadata` permission)
image

PR #3926, opened 2022-11-04
Add a new filter showing reapable images

rhystmills:
## What does this change?

This adds a search filter for 'reapable' images, Add this in the search bar with `+is:reapable`. Reapable images are older than 20 days and match none of the persistence criteria.

A follow-up PR can add this filter by default on the API end, so that users only see images that are not reapable.

1. Deploy to TEST. 
2. Apply the search filter `+is:reapable`
3. Do you see the amount of images you would expect? Are any 'reapable' images less than 20 days old? (they shouldn't be)

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/media-api/app/lib/elasticsearch/IsQueryFilter.scala
+++ b/media-api/app/lib/elasticsearch/IsQueryFilter.scala
@@ -4,6 +4,8 @@ import com.gu.mediaservice.lib.ImageFields
 import com.gu.mediaservice.model._
 import com.sksamuel.elastic4s.ElasticDsl.matchAllQuery
 import com.sksamuel.elastic4s.requests.searches.queries.Query
+import lib.MediaApiConfig
+import org.joda.time.DateTime
 import scalaz.syntax.std.list._
 
 sealed trait IsQueryFilter extends Query with ImageFields {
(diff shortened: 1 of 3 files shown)


media-api/app/lib/elasticsearch/IsQueryFilter.scala
```diff
@@ -62,3 +66,26 @@ case class IsDeleted(isDeleted: Boolean) extends IsQueryFilter {
     (filters.existsOrMissing("softDeletedMetadata", _))(isDeleted)
   )
 }
+
+case class IsReapable(persistedRootCollections: List[String], persistenceIdentifier: String) extends IsQueryFilter {
+  val moreThanTwentyDaysOld = filters.date("uploadTime", Some(new DateTime(0)), Some(DateTime.now().minusDays(20))).getOrElse(matchAllQuery())
```
andrew-nowak, line 71: you can make the from be `None` to get an open-ended search 
```suggestion
  val moreThanTwentyDaysOld = filters.date("uploadTime", None, Some(DateTime.now().minusDays(20))).getOrElse(matchAllQuery())
```

PR #3927, opened 2022-11-04
Apply -is:reapable filter by default on API end

rhystmills:
## What does this change?

Apply the `-is:reapable` filter by default on the API end.

## Tested? Documented?
- [ ] locally by committer


rhystmills:
Closed because it slowed the loading of searches on Grid.

PR #3928, opened 2022-11-04
Upgrade CAPI scala client

andrew-nowak:
## What does this change?

Upgrades the CAPI scala client to the latest version, which includes the latest content models and a client that implements automatic retries with backoff. I've plucked an initial retry policy out of thin air - we should probably come up with better timings

Deploy to TEST and try the usage reindex endpoint (`/usages/digital/content/<content path>/reindex`), and see if the usages for the images gets backfilled

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -185,7 +185,7 @@ lazy val thrall = playProject("thrall", 9002).settings(
 
 lazy val usage = playProject("usage", 9009).settings(
   libraryDependencies ++= Seq(
-    "com.gu" %% "content-api-client-default" % "18.0.1",
+    "com.gu" %% "content-api-client-default" % "19.0.4",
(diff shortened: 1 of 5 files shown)


rhystmills (approved):
Code looks good, and reindexing working as expected.

Is the FIXME something worth addressing soon?

andrew-nowak:
> Is the FIXME something worth addressing soon?

Yep it's a good point - but my understanding is that access to preview CAPI is much more tightly controlled (the endpoint is not generally accessible, even with any tier of API key) which I suspect is why the code is the way it is. It's probably worth us talking to CAPI to see how they would expect us to retrieve retrievableUpdates received on the preview stream, but since preview usages are of much lower priority than published usages I think that's a problem for the future. Let me type up a trello card...

PR #3929, opened 2022-11-04
add warning and restrictions overlay

ornel-lloyd-edano:
## What does this change?

BBC only UI/UX enhancement which adds the following through config:

Adds a text message superimposed over the selected image preview tile's description
- "This image has warnings attached. It is important you check all details before use." for alert
- "This image has warnings attached. It has a restriction that requires attention." for warning
- "This image has a lease attached, read more before use" for lease attached

Adds a 100% opacity background color to the text message. Also adds a 30% opacity overlay to the selected image preview tile
- red overlay for alert
- brown/amber overlay for warning
- blue overlay for lease attached

 

Screen Shot 2022-11-04 at 7 29 14 PM

Add the following configuration in application.conf
> warningText {
>     imagePreviewFlag {
>         alertCopy = "This image has warnings attached. It is important you check all details before use."
>         warningCopy = "This image has warnings attached. It has a restriction that requires attention."
>         leaseAttachedCopy = "This image has a lease attached, read more before use."
>     }
> }

Without the above config, the UI will behave just like before. No color overlay or warning messages on the selected preview images.

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [x] locally by committer


kahuna/public/js/preview/image.js
```diff
@@ -106,13 +106,31 @@ image.controller('uiPreviewImageCtrl', [
     ctrl.srefNonfree = () => storage.getJs("isNonFree", true) ? true : undefined;
 
     ctrl.hasActiveAllowLease = ctrl.image.data.leases.data.leases.find(lease => lease.active && lease.access === 'allow-use');
+
+    ctrl.showOverlay = (image) => $window._clientConfig.enableWarningFlags && ctrl.isSelected && (ctrl.showAlertOverlay(image) || ctrl.showWarningOverlay(image));
```
jonathonherbert, line 110: Is there a reason we are passing the image here, rather than using ctrl.image? I can't see these functions being called with anything other than the image that is currently in scope.

kahuna/public/stylesheets/main.css
```diff
@@ -2855,3 +2855,36 @@ FIXME: what to do with touch devices
     }
 
 }
+
+.result__select__overlay.alert {
+  background-color: #C433314D;
```
jonathonherbert, line 2860: I can see the element this styling refers to in the DOM, but it doesn't look like it has a height – so I don't see the tinted overlay:

Screenshot 2022-11-10 at 10 56 47

Having said that, I think the messaging is clear without it, and to me the overlay is problematic – it reduces the contrast of text that is already hard to read in some cases. If it's still desirable, would it be better to alter the background of the preview component? Alternatively, using [`background-blend-mode`](https://developer.mozilla.org/en-US/docs/Web/CSS/background-blend-mode) might be one way of applying a hue whilst preserving contract.

jonathonherbert (commented):
Thanks for this, @ornel-lloyd-edano – a few comments.

paperboyo:
Heya, thanks for this! Initial functionality review:

1. “yellow” ones and also others after resizing the browser are cut off:
   image
2. text is different colour sometimes (should always be white?):
   image
3. I thought we decided not to have it for leases? If we do want it, it will require changes (incl. to some other parts of the app):
  - a) it uses a BBC blue for active Allow lease. It should use the Grid [teal](https://github.com/guardian/grid/pull/3854), configurable to be another “teal” in one place
    - a1) teal does not apply to No rights with an active Allow cropping lease (rationale: users should fix No rights first!): consequently, “blue” also should not apply (basically, we need to have it consistent)    
  - b) active Allow leases making badges teal (https://github.com/guardian/grid/pull/3854) didn’t cover making images “red” for active Deny lease. Given functionality here, we should do it – images with active Deny cropping lease should:
    - b1) have a warning triangle added in the browser (even when no selection going on) with a corresponding tooltip giving the Denied reason
    - b2) should have their Info panel counts adjusted to be “red” 
4. it uses a BBC red for active Deny lease, No rights, Paid and Invalid. It should use the Grid red, configurable to be another “red” in one place
5. it uses a BBC yellow for active Restricted, No rights and Paid. It should use the Grid yellow, configurable to be another “yellow” in one place

As these are related to messaging in the browser and Info panel, this highlights some deficiencies worth addressing (possibly, separately, though!) to keep consistency:

6. this marks Invalid imagery (no desc and/or credit) “red”. They currently look like Free to use in the browser and Info panel:
   image
  a) they should get a warning triangle and a corresponding tootlitp (probably `Unusable: missing [description] [and] [credit]`
  b) they should also get a corresponding Info panel badge turn red (`x unusable`?)   
7. more generally about warning texts, I wonder: should some of them be dynamic (or have parts dynamic)?
   a) Invalid could just say what they are missing (desc and/or credit)
   b) Restricted could spell out Restriction text
   c) Leased ones could spell out Lease note
   Eg. `Image has a lease: [Lease note]`, `Image is restricted use: [restriction text]`. It’s true we would need to truncate some longer ones, but maybe it would be more useful than a lot of repeatable text that just basically says `Please read stuff in the other part of the app`? I worry repeatable nature may numb users and they will stop reading the text?

cc @louisegoodspeed @pradasa and others (sorry, can’t easily tag teams) Only pt. 1 is a bug. Making colours consistent (configurable) seems important. Some other points do not relate to the PR directly, but should be tackled for consistency. Pt. 7 is for discussion (and some thought given). cc @andrew-nowak because he worked on related functionality recently

kahuna/public/js/preview/image.js
ornel-lloyd-edano: I see you have valid point @jonathonherbert. I have removed the image param since they are fixed anyway

kahuna/public/stylesheets/main.css
ornel-lloyd-edano: Just noticed I have some uncommitted css. This should now be fixed

ornel-lloyd-edano:
Hi @paperboyo I may have mixed up the BBC/Grid colors here (for example, the flag is teal but overlay is blue)
Fixing it to use all Guardian colors by default. BBC have external custom css to override the colors anyway.

paperboyo:
Testing new commits, thanks!

Pts. 1 & 2 look fixed 🎉.
Pts. 3a, 4 & 5 fixed 🎉.

- Pt. 3a1 is still inconsistent with how “teal” is not applied to a badge for No rights:

  image

  I guess we can:
  - ignore the inconsistency
  - make it behave like “teal” on badges: not colour the cards “teal” for No rights only, leave it red (rationale above)
  - decide rationale is weak and make “teal” on badges colour No rights badge “teal” too (counter-rationale: _yeah maybe you should fix No rights first, but we are showing you **effective** rights_)

- Pt. 3b is still inconsistent with how active Deny leases are not in any way signalled without an ongoing selection (images look like free):
  Active Deny lease on a Free image (Info panel also gives no indication in counts):
  image
   Same image selected:
   image

  I guess we can:
  - ignore the inconsistency
  - fix things described above in pt. 3b (not necessarily in this PR)

Pt. 6 still applies.

I think only pt. 3a1 is worth fixing in this PR. It is an active inconsistency (selection state and non-selection state give conflicting signals).

Pts. 3b and 6, while inconsistent, are weaker: the state is not conflicting, it’s just non-selection state is not as helpful in messaging as a selection one. I think it’s worth registering as an issue to fix later, if there is no apettite to fix it here.

What do you think, @louisegoodspeed /others?

paperboyo:
8. This PR breaks tooltips in selection state when the warnings are on:

    Current PROD:
    sdfsdfsdfsdf

    This PR:
    sdfsdfsdfsdf

@ornel-lloyd-edano is it possible to keep the tooltips? Or will this be a Family Guy CSS GIF-like situation (impossible)?

And kinda related to pt. 7… if it’s possible to keep the tooltips, since hovering over Description isn’t possible, should hovering over the warning text spell out the reasons for the warnings more clearly (for simplicity, just display exactly the same tooltips as ones for the badges)?

kahuna/app/lib/KahunaConfig.scala
```diff
@@ -60,5 +60,12 @@ class KahunaConfig(resources: GridConfigResources) extends CommonConfig(resource
     .getOrElse("This image can be used, but has warnings:")
   val unusableTextHeader: String = configuration.getOptional[String]("warningText.unusableHeader")
     .getOrElse("Unusable image")
+  val enableWarningFlags: Boolean = configuration.getOptional[Map[String, String]]("warningText.imagePreviewFlag").isDefined
```
phillipbarron, line 63: Would be useful to add these new configuration items in the [the config docs](https://github.com/guardian/grid/blob/main/docs/06-objects-of-interest/02-config.md)

NickyPhelanBBC:
hi @paperboyo, Re point 8 - we think the UX here is ok.

1. user browses images
2. user decides upon an image _(at this point taking in the description)_
3. user selects image for downloading _(can no longer see the description or tool tip if there is a restriction/lease/warning overlay)_
4. user sees warning / lease / restriction information _(this is paramount to us as it will save us money/fines which is why it is so prominent)_
5. user can deselect the image to see description and tool tip if desired

What's key here is that **only** images with a restriction/warning or lease have an overlay.

paperboyo:
Thanks, @NickyPhelanBBC, this resolves parts of pt. 8 related to pt. 7 (**extra** tooltip info).
I would still like to know if there is a simple way of not loosing **existing** tooltips info (because why loose it if we don’t have to?). If there isn’t, and you are fine with loosing it – no problem.

Pts. 3a1 (most important) and 3b and 6 still need a decision. I would like to sort 3a1 here, if possible, but given the importance of this work for you, you may decide to just register all three of them as outstanding issues and tackle later?

louisegoodspeed:
Apologies for delay I wanted to chat to Ornel/Honor/Nicky first. @paperboyo on 3a1 - we decided the warning should be red not blue for this particular case, as the image is no rights.
3b - we are happy to ignore the inconsistency and have a red warning appear (if there is a deny lease) once selected, but maybe look at this later to improve
6. agree with you a and b should be fixed

paperboyo:
Thanks, @louisegoodspeed! So 3b and 6 for later, makes sense (if you could kindly open issue(s) for them at any point that would be great, otherwise we will forget).

So let’s fix 3a1 in this PR, then there is just this [comment](https://github.com/guardian/grid/pull/3929#discussion_r1029075202) from Phill about the docs and we can do final review speeding things up for ya.

What about the (existing) tooltips for when warnings are up?

NickyPhelanBBC:
Hiya @paperboyo just chatted to @louisegoodspeed about the tooltips. How about we look at this as an iteration when we know a bit more about how the messaging is being received?

kahuna/app/lib/KahunaConfig.scala
ornel-lloyd-edano: Hi @phillipbarron this PR updates the config docs https://github.com/guardian/grid/pull/3965

ornel-lloyd-edano:
> Thanks, @louisegoodspeed! So 3b and 6 for later, makes sense (if you could kindly open issue(s) for them at any point that would be great, otherwise we will forget).
> 
> So let’s fix 3a1 in this PR, then there is just this [comment](https://github.com/guardian/grid/pull/3929#discussion_r1029075202) from Phill about the docs and we can do final review speeding things up for ya.
> 
> What about the (existing) tooltips for when warnings are up?

Hi @paperboyo tooltips and issue 3a1 should be fixed by recent push

jonathonherbert (approved):
Changes look good, @ornel-lloyd-edano 👍 

Just to reiterate some non-blocking comments about the overlay –

_> The overlay is problematic – it reduces the contrast of text that is already hard to read in some cases. If it's still desirable, would it be better to alter the background of the preview component? Alternatively, using [background-blend-mode](https://developer.mozilla.org/en-US/docs/Web/CSS/background-blend-mode) might be one way of applying a hue whilst preserving contract.

ornel-lloyd-edano:
Hi @jonathonherbert 

> The overlay is problematic – it reduces the contrast of text that is already hard to read in some cases.

I just pushed a fix to make the text contrast better. It seems that z-index: 2 of the text was not working before. Now it appears above the overlay hence should look better.

jonathonherbert (approved):

PR #3931, opened 2022-11-07
reduce usage reindex capi query to MVQ

andrew-nowak:
minimal viable query

## What does this change?

We don't need all fields, elements, etc. Grid usage reingestion only needs media atoms, image elements, and a few fields. This makes the request much lighter for CAPI to process, allowing us to reprocess at a much higher rate!

Deploy to TEST, build and publish an article with an image (and optionally a media atom with a thumbnail selected). Delete the usages out of Grid, then run a reindex request. Do the usages still get readded?

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/usage/app/controllers/UsageApi.scala
+++ b/usage/app/controllers/UsageApi.scala
@@ -105,8 +105,9 @@ class UsageApi(
     )
 
     val query = ItemQuery(contentId)
-      .showFields("all")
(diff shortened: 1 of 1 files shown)


rhystmills (approved):
Tested in TEST and images usage is being reindexed as expected 👍

PR #3940, opened 2022-11-08
Add documentation for configuration options

rhystmills:
## What does this change?
To support new organisations in using the Grid, it would be helpful to make documentation of configurable options be as accurate and accessible as possible.

This PR represents [our existing doc in Google](https://docs.google.com/document/d/1CSERbLwbu6nT_ggzzYxdUt9IHpfGUJLIYPnU_9MZbpc/edit#heading=h.3vceb61ngzf) within the repo itself. I've used html tables rather than markdown tables as html proved much easier to read in document form. It's easier to read [in rendered form](https://github.com/guardian/grid/blob/rm/document-configuration/docs/06-objects-of-interest/02-config.md) than in the html here.

Can we generate this kind of documentation from the repository itself, rather than manually as I've done here? That would be easier to maintain in the future.


phillipbarron (approved):
LGTM 👍 - We added to this as part of the work we were doing for pinboard

PR #3944, opened 2022-11-10
disable the batch crop button when all images uncroppable

andrew-nowak:
## What does this change?

Previously, the "full frame crop" button on the _search_ page was always enabled - even if the image couldn't be cropped! Now the button will be disabled if no images are croppable, or show a warning triangle if some images can't be cropped (like downloads)

image
image

## Tested? Documented?
- [ ] locally by committer


phillipbarron:
The styling is a little inconsistent with the download button (in which the style seem a little broken too but I an not sure that that is related to this change)
Screenshot 2022-11-11 at 10 44 50

PR #3949, opened 2022-11-14
Add script for reindexing usages

Fweddi:
## What does this change?

Worked on with @andrew-nowak and @rhystmills.

Adds a bash script to reindex usages from a `csv` of CAPI articles (of format `internal-composer-code,path`).

Has the ability to replay from the last reindexed ID if needed.

Also logs failed IDs.

We are testing this on PROD. Once we have finished running this script, we will be confident of merging it into the codebase for future reference.

Usages are successfully reindex for all CAPI articles in PROD.

### DIFF
--- a/scripts/usage-reindex/reindex-usages.sh
+++ b/scripts/usage-reindex/reindex-usages.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+isFastForwarding=true
+
+touch lastQueriedId.txt
+lastQueriedId=$(cat lastQueriedId.txt)
+
+if [[ $lastQueriedId == "" ]]
+  then isFastForwarding=false
+fi
+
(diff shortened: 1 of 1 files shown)


alinaboghiu:
> We are testing this on PROD. Once we have finished running this script, we will be confident of merging it into the codebase for future reference.

Not sure from this comment if this actually needs a review?

Fweddi:
> > We are testing this on PROD. Once we have finished running this script, we will be confident of merging it into the codebase for future reference.
> 
> Not sure from this comment if this actually needs a review?

Yeah, I think all it needs is a nod through once this has been tested on PROD.

andrew-nowak (approved):
I believe we have run this already, so good to merge

PR #3950, opened 2022-11-14
Remove automatic rewrites for Peter Dejong

andrew-nowak:
## What does this change?

We have been rewriting Peter Dejong's bylines to Peter de Jong. There are two photographers of note, of each spelling, so we shouldn't rewrite either to the other. 

Removes the rewrite rule

## Tested? Documented?
- [ ] locally by committer


paperboyo:
Oh, no! So it was the renamer only? 😱 me fault!

This dictionary baby-rewriter is generally tad unsafe. I will add a commit here for another silly rewrite:

image

But I strongly feel, it’s more useful than not!

paperboyo (approved):
Thanks!

PR #3959, opened 2022-11-16
Commit mass-deletion script

andrew-nowak:
## What does this change?

Store the mass deletion script in repo for reuse and an example of how to create and run mass operation scripts. 

Does the README make sense? Is the script easy to understand

### DIFF
--- a/scripts/mass-deletion/mass-deletion.sh
+++ b/scripts/mass-deletion/mass-deletion.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+if [[ -n "$GRIDKEY" || -n "$GRIDDOMAIN" ]]; then
+  echo "make sure to set env vars GRIDKEY and GRIDDOMAIN to run this script"
+  exit 1
+fi
+
+del=todelete.txt
+prog=progress.txt
+com="complete.txt"
+errs=errors.txt
+
+touch $prog
+touch $com
+touch $errs
+
+all="$(wc -l $del | awk '{ print $1 }' )"
+ndone=0
+
+skipping=1
+last="$(cat $prog)"
+
+
+
+while read id; do
+  ndone=$((ndone + 1))
+  if [[ $skipping = 1 && -n "$last" && $id != $last ]]; then
+    continue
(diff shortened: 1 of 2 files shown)


scripts/mass-deletion/mass-deletion.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+if [[ -n "$GRIDKEY" || -n "$GRIDDOMAIN" ]]; then
+  echo "make sure to set env vars GRIDKEY and GRIDDOMAIN to run this script"
+  exit 1
+fi
+
+del=todelete.txt
+prog=progress.txt
+com="complete.txt"
```
bryophyta, line 10: if this value is quoted, does that mean the others should be too?

scripts/mass-deletion/mass-deletion.sh
```diff
@@ -0,0 +1,44 @@
...
+skipping=1
+last="$(cat $prog)"
+
+
+
+while read id; do
+  ndone=$((ndone + 1))
+  if [[ $skipping = 1 && -n "$last" && $id != $last ]]; then
+    continue
+  elif [[ $skipping = 1 && -n "$last" ]]; then
```
bryophyta, line 29: it took me a minute to work out why skipping was initially set to 1. I take it that it's because we're looking through the ids until we find the first one after the 'prog' value (if there's one set) and only then do we start processing them? this seems v sensible but might benefit from a brief comment to explain the pattern?

bryophyta (approved):
I added a couple of small comments; looks good to me! I tested success and error conditions locally, though without the `curl` command

scripts/mass-deletion/mass-deletion.sh
andrew-nowak: there's no spaces in the values so all are optional for quoting. I've removed the quotes from `complete` for consistency

scripts/mass-deletion/mass-deletion.sh
andrew-nowak: thanks, added

PR #3963, opened 2022-11-22
Get S3 Watcher running locally

andrew-nowak:
## What does this change?

Update S3 Watcher development scripts to run locally. Main points were removing the handling for ini config files, and configuring for interacting with localstack.

It's not super pretty (there's definitely some scope for DRYing) but it at least now works, and you can test out s3watcher locally.

Set up and have the grid running locally. Then, in another terminal, open s3watcher and run the setup script and then the run-local script. There should be no errors in the terminal, and the image should be uploaded into your local grid.

Changes and upgrades for s3 watcher can be tested locally

## Tested? Documented?
- [x] locally by committer


rhystmills (approved):
Tested locally and working as expected 👍

twrichards:
@andrew-nowak do you recall why the `innerServiceStatusCheck` was removed from `the_pingler.sh`?

andrew-nowak:
not sure why it's part of this pr but it was _massively_ expensive and responsible for probably 90% of my problems running locally - it would often crash my sbt and be almost impossible to get it to close without rebooting

PR #3964, opened 2022-11-22
Add feature switch panel to kahuna

rhystmills:
## What does this change?

This PR adds a feature switch panel to Kahuna. Switch values are usable in the frontend and the backend (via getters). The kahuna client updates the values by setting a cookie, otherwise they are set to the default value defined in the backend.

Screenshot 2022-11-22 at 15 50 48

The panel can be opened with <kbd>shift-F11</kbd>. The PR includes a trivial example feature restyling the feature switch buttons.

Open the feature switch panel. Change the example switch toggle value. Restart the page - does the value remain the same?

Now with tests!

## Tested? Documented?
- [X] locally by committer

### DIFF
--- a/kahuna/app/controllers/FeatureSwitchController.scala
+++ b/kahuna/app/controllers/FeatureSwitchController.scala
@@ -10,12 +10,9 @@ object ExampleSwitch extends FeatureSwitch(
   default = false
 )
 
-object FeatureSwitches {
+class FeatureSwitchController(featureSwitches: List[FeatureSwitch]){
   // Feature switches are defined here, but updated by setting a cookie following the pattern e.g. "feature-switch-my-key"
   // for a switch called "my-key".
-  lazy val featureSwitches = List(
-    ExampleSwitch
-  )
 
   def getFeatureSwitchCookies(request: Request[AnyContent]): List[(FeatureSwitch, Option[Cookie])] =
     featureSwitches.map(featureSwitch => (featureSwitch, request.cookies.get(s"feature-switch-${featureSwitch.key}")))
@@ -51,3 +48,4 @@ object FeatureSwitches {
     }
   }
 }
+
--- a/kahuna/app/controllers/FeatureSwitchController.scala
+++ b/kahuna/app/controllers/FeatureSwitchController.scala
@@ -19,10 +19,10 @@ class FeatureSwitchController(featureSwitches: List[FeatureSwitch]){
 
   def getClientSwitchValues(featureSwitchesWithCookies: List[(FeatureSwitch, Option[Cookie])]): Map[FeatureSwitch, Boolean] = {
     featureSwitchesWithCookies
-      .map(featureSwitchWithCookie => featureSwitchWithCookie match{
+      .map {
         case (featureSwitch, Some(cookie)) => (featureSwitch, getBoolean(cookie.value))
         case (featureSwitch, None) => (featureSwitch, featureSwitch.default)
-      })
+      }
       .toMap
   }
 
--- a/kahuna/app/controllers/FeatureSwitchController.scala
+++ b/kahuna/app/controllers/FeatureSwitchController.scala
@@ -27,13 +27,13 @@ class FeatureSwitchController(featureSwitches: List[FeatureSwitch]){
   }
 
   def getFeatureSwitchesToStringify(clientSwitchValues: Map[FeatureSwitch, Boolean]): List[Map[String, String]] = {
-    clientSwitchValues.map(clientSwitch => clientSwitch match {
+    clientSwitchValues.map {
       case (featureSwitch, value) => Map(
         "key" -> featureSwitch.key,
         "title" -> featureSwitch.title,
         "value" -> value.toString
       )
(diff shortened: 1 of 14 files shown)


kahuna/app/controllers/FeatureSwitchController.scala
```diff
@@ -0,0 +1,52 @@
...
+
+case class FeatureSwitch(key: String, title: String, default: Boolean)
+
+object ExampleSwitch extends FeatureSwitch(
+  key = "example-switch",
+  title = "An example switch. Use rounded corners for the feature switch toggle",
+  default = false
+)
+
+class FeatureSwitchController(featureSwitches: List[FeatureSwitch]){
```
andrew-nowak, line 13: I'm possibly being a bit pedantic, but this class isn't a controller in the play terminology (a controller's methods would be `Action`s, listed in `conf/routes` to handle http requests, see <https://www.playframework.com/documentation/2.8.x/ScalaActions>). I think it'll be confusing to have this named as a controller and in the controller directory, so maybe we should call it a service (or even `FeatureSwitches`?) and move it to the lib directory instead

rhystmills: That sounds very sensible - I'll do a restructure.

kahuna/public/js/components/gr-feature-switch-panel/gr-feature-switch-panel.tsx
```diff
@@ -0,0 +1,150 @@
...
+    <path d="M22.35 34.3h3.6V22h-3.6ZM24 18.7q.9 0 1.475-.575.575-.575.575-1.425 0-.95-.575-1.525T24 14.6q-.9 0-1.475.575-.575.575-.575 1.525 0 .85.575 1.425.575.575 1.475.575Zm0 26q-4.3 0-8.05-1.625-3.75-1.625-6.575-4.45t-4.45-6.575Q3.3 28.3 3.3 24q0-4.35 1.625-8.1T9.35 9.35q2.8-2.8 6.575-4.45Q19.7 3.25 24 3.25q4.35 0 8.125 1.65 3.775 1.65 6.55 4.425t4.425 6.55Q44.75 19.65 44.75 24q0 4.3-1.65 8.075-1.65 3.775-4.45 6.575-2.8 2.8-6.55 4.425T24 44.7Zm.05-3.95q6.95 0 11.825-4.9 4.875-4.9 4.875-11.9 0-6.95-4.875-11.825Q31 7.25 24 7.25q-6.95 0-11.85 4.875Q7.25 17 7.25 24q0 6.95 4.9 11.85 4.9 4.9 11.9 4.9ZM24 24Z"/>
+  </svg>;
+
+
+const FeatureSwitch = ({ data }: { data: FeatureSwitchData}) => {
+  const [on, setOn] = useState(data.value === "true");
+
+  const handleClick = () => {
+    const newOn = !on;
+    document.cookie = `feature-switch-${data.key}=${newOn.toString()}`;
```
andrew-nowak, line 48: right now the cookies (and the feature switch controller) are only set and only readable for kahuna. Would we ever want the feature switches to be able to control serverside behaviour in the other services?

If yes, there is a potential source of frustration - some services are on subdomains of kahuna (eg. media-api at `api.media.gutools`) but others are not (eg. usages at `media-usage.gutools`) (see <https://github.com/guardian/grid/blob/main/common-lib/src/main/scala/com/gu/mediaservice/lib/config/Services.scala#L23-L35>) which would make sharing cookies harder

rhystmills: This is a valid concern and definitely worth being aware of. Could use cookies for now for sake of simplicity and change the mechanism later if we find we want to use a non `media.gutools` subdomain.

My feeling is that Media API is the most likely service to be used outside of the frontend.

kahuna/app/controllers/FeatureSwitchController.scala
rhystmills: (Now done)

kahuna/public/js/components/gr-feature-switch-panel/gr-feature-switch-panel.module.css
```diff
@@ -0,0 +1,105 @@
...
+  display: inline;
+  fill: rgb(204,204,204);
+  width: 20px;
+}
+
+.infoPane {
+  border-bottom: 1px solid #565656;
+  width: 100%;
+  display: flex;
+  padding: 0px 8px;
```
andrew-nowak, line 99: this infopane overflows for me, creating a horizontal scrollbar (because 100% width with padding). I think you could set `box-sizing: border-box` to fix it

andrew-nowak (approved):
can't overstate how huge this is,

kahuna/public/js/components/gr-feature-switch-panel/gr-feature-switch-panel.module.css
rhystmills: Good spot - now sorted

PR #3966, opened 2022-11-25
tweak the `--use-TEST` local dev mode

twrichards:
when using the `--use-TEST` flag of the `start.sh` script, only attempt to `docker-compose down` if Docker daemon is actually running


phillipbarron (approved):
❤️

PR #3970, opened 2022-11-28
Usages: log when processing usage event fails and fix retrievable content update

andrew-nowak:
## What does this change?

Before, exceptions thrown during processing or deserialization of crier events for processing usages would be swallowed silently. Now, log an error.

Also, fix processing of retrievable content updates, by using the ID of the content instead of the URL, and taking the parameters needed as used in the reindex API.

Add a new image to a very, very long liveblog (either restore one from PROD or DM me, I can link you to one that is long enough to trigger retrievablecontentupdate).

The one remaining (known) hole in Composer usages is filled

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/usage/app/lib/CrierEventProcessor.scala
+++ b/usage/app/lib/CrierEventProcessor.scala
@@ -78,7 +78,8 @@ abstract class CrierEventProcessor(config: UsageConfig, usageGroupOps: UsageGrou
 
     records.asScala.foreach { record =>
       val buffer: Array[Byte] = record.getData.array()
-      val deserialization: Try[Unit] = ThriftDeserializer.deserialize(buffer).map(processEvent)
+      val deserialization: Try[Event] = ThriftDeserializer.deserialize(buffer)
+      deserialization.foreach(processEvent)
       deserialization.failed.foreach { e: Throwable =>
         logger.error("Failed to deserialize crier event", e)
       }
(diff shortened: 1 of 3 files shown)


usage/app/lib/CrierEventProcessor.scala
```diff
@@ -99,42 +104,50 @@ abstract class CrierEventProcessor(config: UsageConfig, usageGroupOps: UsageGrou
...
-        }
-
-      case _ => logger.debug(logMarker, s"Unsupported event type $EventType")
+        case EventType.RetrievableUpdate =>
+
+          event.payload match {
+            case Some(retrievableContent: EventPayload.RetrievableContent) =>
+              val capiUrl = retrievableContent.retrievableContent.capiUrl
+
+              val query = liveCapi.usageQuery(retrievableContent.retrievableContent.id)
```
andrew-nowak, line 128: this line is the bugfix, the rest are logging improvements and a tiny refactor to do this DRYly

rhystmills (approved):
Looks good, usage has been picked up and log has appeared where we want it to 👍

usage/app/lib/CrierEventProcessor.scala
andrew-nowak: the bug being: the capi client expects to query using the capi ID, but we were giving it the capi URL

PR #3971, opened 2022-11-28
Usages: fetch retrievable draft updates from preview capi

andrew-nowak:
depends on #3970 

needs testing. must check preview capi url in prod usage config before deploying!!!!!

## What does this change?

The crier stream emits several types of events. The two we are currently configured to listen to are Update (some content is updated, and the updated content is included in the event) and RetrievableUpdate (some content was updated but the updated content is _not_ included in the event). Retrieval of content has been broken for a while, and #3970 fixes for retrieving live content. However there was no implementation for retrieving from preview CAPI, and any retrievable updates sent on the preview stream would be looked up from live CAPI - which makes little sense.

Implement a CAPI client which can retrieve from preview CAPI (involves adding headers signing the request with the Grid's IAM role, which was already allowlisted for preview CAPI access).

Also adds a few more logs around the CAPI fetches and tweaks log levels.

Add an image to a very long liveblog (ask me for one to test with if required) and check that a "pending publication" usage is added on that image.

## Tested? Documented?
- [ ] locally by committer


rhystmills (approved):
Tested on a long liveblog and found that the 'pending publication' usage show up as intended 👍

PR #3974, opened 2022-11-29
Add telemetry to search and filters (chips)

rhystmills:
## What does this change?

Adds telemetry to filters and searches (only those done via the search bar - not the checkboxes for my uploads and free only).

These are applied when the search is updated, which is debounced after the user stops typing for >500ms.

I'm interested in how we think we can make this more configurable for different organisations, who would have their own telemetry/analytics setups - they would have their own endpoints that would expect the data in a certain shape and therefore would need to reshape the data. How can we make this work for the BBC (and possibly Telex)? How could we make it work if we had 10 organisations using the Grid? 

Run locally or on `TEST`, ensuring that the telemetryUri is available from config, i.e. `links.telemetryUri="https://user-telemetry.code.dev-gutools.co.uk"` in kahuna config locally.

Make a search with filters. Check kibana for the logs ([relevant search here](https://logs.gutools.co.uk/s/editorial-tools/goto/b65d6850-6ffc-11ed-8ffd-9129a5724241)).

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/search/structured-query/structured-query.js
+++ b/kahuna/public/js/search/structured-query/structured-query.js
@@ -9,9 +9,7 @@ import {rxUtil} from '../../util/rx';
 
 import {querySuggestions, filterFields} from './query-suggestions';
 import {renderQuery, structureQuery} from './syntax';
-import { search } from '..';
 import { sendTelemetryEvent } from '../../services/telemetry';
-import { setTags } from '@sentry/browser';
 
 export const grStructuredQuery = angular.module('gr.structuredQuery', [
     rxUtil.name,
@@ -77,15 +75,15 @@ grStructuredQuery.directive('grStructuredQuery', ['subscribe$', function(subscri
                     // filter > {type: 'filter', filterType: 'inclusion', key: 'is', value: 'cool'}
                     const { type } = queryComponent;
                     const formattedType = (type) => {
-                        if (type === 'text') return 'GRID_SEARCH'
-                        if (type === 'filter') return 'GRID_FILTER'
-                        return `GRID_${type.toUpperCase()}`
-                    }
+                        if (type === 'text') { return 'GRID_SEARCH'; }
+                        if (type === 'filter') { return 'GRID_FILTER'; }
+                        return `GRID_${type.toUpperCase()}`;
+                    };
                     if (queryComponent.value){
                         // In case search is empty, as with a search containing only filters
                         sendTelemetryEvent(formattedType(type), queryComponent, 1);
-                    }
-                })
+                    };
+                });
             });
         }
     };
--- a/kahuna/public/js/search/structured-query/structured-query.js
+++ b/kahuna/public/js/search/structured-query/structured-query.js
@@ -9,6 +9,9 @@ import {rxUtil} from '../../util/rx';
 
 import {querySuggestions, filterFields} from './query-suggestions';
 import {renderQuery, structureQuery} from './syntax';
+import { search } from '..';
+import { sendTelemetryEvent } from '../../services/telemetry';
+import { setTags } from '@sentry/browser';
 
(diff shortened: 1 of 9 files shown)


kahuna/public/js/search/structured-query/structured-query.js
```diff
@@ -67,6 +68,22 @@ grStructuredQuery.directive('grStructuredQuery', ['subscribe$', function(subscri
 
             subscribe$(scope, ctrl.newQuery$, query => {
                 ngModelCtrl.$setViewValue(query);
+                const structuredQuery = structureQuery(query);
+                structuredQuery.forEach(queryComponent => {
```
andrew-nowak, line 72: it looks like this sends an event per chip, but only one for the entire text (and I think none for empty searches), which will overemphasise chip usage. I took a look but also couldn't see a handy way of sending multiple bits of data in a single event. Maybe the best is to create a random uuid for each search and add to each event, which I think might let us correlate all of the events when we query? cc @jonathonherbert for any ideas

kahuna/public/js/search/structured-query/structured-query.js
```diff
@@ -67,6 +68,22 @@ grStructuredQuery.directive('grStructuredQuery', ['subscribe$', function(subscri
...
+                    // search > {type: 'text', value: 'my search'}
+                    // filter > {type: 'filter', filterType: 'inclusion', key: 'is', value: 'cool'}
+                    const { type } = queryComponent;
+                    const formattedType = (type) => {
+                        if (type === 'text') { return 'GRID_SEARCH'; }
+                        if (type === 'filter') { return 'GRID_FILTER'; }
+                        return `GRID_${type.toUpperCase()}`;
+                    };
+                    if (queryComponent.value){
+                        // In case search is empty, as with a search containing only filters
```
andrew-nowak, line 83: if i add a search, and then delete it, grid sends an event with value: "undefined" (the string undefined, not the value) - I think this is a bug in structureQuery

andrew-nowak: https://github.com/guardian/grid/blob/main/kahuna/public/js/search/structured-query/syntax.js#L36-L40 `'' + undefined === 'undefined'` 🙃

kahuna/public/js/search/structured-query/structured-query.js
rhystmills: Interesting idea. Makes me think - if we had session ID and user ID as well we'd have the potential to do some decent analysis down the line.

rhystmills: Sending chips individually was the intentional behaviour but I think your suggestion makes more sense.

rhystmills: We now should have all three. Will do some testing to check it's working in the morning

kahuna/public/js/search/structured-query/structured-query.js
rhystmills: Made some changes that may have sorted this - will check in the morning

kahuna/public/js/services/telemetry.ts
```diff
@@ -0,0 +1,39 @@
...
+    } else {
+        const id = v4();
+        storage.setItem(key, id.toString());
+        return id;
+    }
+};
+
+const getBrowserId = () => getStoredId(localStorage, 'browserUuid');
+
+const getSessionId = () => getStoredId(localStorage, 'sessionUuid');
```
andrew-nowak, line 17: should this be set on sessionStorage?

rhystmills: Well caught! Yes, should be `sessionStorage`. Fixed now!

andrew-nowak (approved):

PR #3981, opened 2022-12-02
Spike at multitenancy

andrew-nowak:
## What does this change?

Configure user groups/roles/"tenants" with a list of free suppliers (and exclusions) and allow users to switch between these at will. Switching between tenants means that the "free to use" search (and the consequential ability to crop/download) changes, depending on who supplied the image, which tenant the user has selected, and which suppliers are considered free-to-use for that tenant. Leases can be assigned to some of these tenants and will only affect users who have selected that tenancy.

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-tenant-switcher/gr-tenant-switcher.tsx
+++ b/kahuna/public/js/components/gr-tenant-switcher/gr-tenant-switcher.tsx
@@ -1,8 +1,8 @@
 import * as React from "react";
 import * as angular from "angular";
 import { react2angular } from "react2angular";
-import {ChangeEvent, useMemo} from "react";
-import {getFeatureSwitchActive} from "../gr-feature-switch-panel/gr-feature-switch-panel";
+import { ChangeEvent, useMemo } from "react";
+import { getFeatureSwitchActive } from "../gr-feature-switch-panel/gr-feature-switch-panel";
 
 type Tenant = {
   id: string;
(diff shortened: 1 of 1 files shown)


kahuna/app/lib/FeatureSwitch.scala
```diff
@@ -9,6 +9,11 @@ object ExampleSwitch extends FeatureSwitch(
   title = "An example switch. Use rounded corners for the feature switch toggle",
   default = false
 )
+object MultitenancySwitch extends FeatureSwitch(
+  key = "multitenancy",
+  title = """Multi-tenanct behaviour for the grid. Requires setting up the "tenants" config options!""",
```
andrew-nowak, line 14:
```suggestion
  title = """Multi-tenant behaviour for the grid. Requires setting up the "tenants" config options!""",
```

kahuna/public/js/components/gr-tenant-switcher/gr-tenant-switcher.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const onSelectTenant = (ev: ChangeEvent<HTMLSelectElement>) => {
+    const id = (ev.target as any).value as string;
+
+    if (id === '_default') {
+      tenancy.clear();
+    } else {
+      tenancy.set(id);
+    }
+
+    $window.location.reload();
```
andrew-nowak, line 35: TODO: the ui-router equivalent ($state.refresh?)

PR #3982, opened 2022-12-05
add share with url button

abdelrahmansd:
## Tested? Documented?
- [ ] locally by committer


paperboyo:
Thanks for this, looks good! Some initial comments (before my holidays):
1. Maximum allowed by URL length seems to be 45 images. But button and overlay both suggest one can share >45. We should either dim (disable) the button when selecting more than 45 or make the overlay say it’s not possible or that only first 45 will be shared (probs worse). I would go with disabling the button and having an explanation of the limit in its tooltip?
2. Overlay should be dismissable by ESC.
3. Overlay should be dismissable by clicking anywhere outside it (it’s just informational). As is: it survives page changes and one can interact with whole of Grid but not with the top bar. This makes little sense.
4. [not sure here, but maybe even] Overlay should disappear automagically after x seconds?
5. URL for sharing a single image is currently the same (`ids=`) as sharing multiple: it opens a browser (search page). I think it may be more useful, if it would just link directly to a viewer for this one image?

That’s it for now. I’m interested in @louisegoodspeed and @pradasa opinions. Also, pinging @twrichards as this seems related to sharing for Pinboard (although there, one can share crops and searches too)

pradasa:
Thanks for this! We were reviewing it on CODE and it looks like the link isn't accessible at the moment: 
Screenshot 2023-02-21 at 15 51 22

It shows as blue on default and click and white on hover. 
Given the background is light green, it might be better to have the link in black with an underline. 
cc @drewcochrane @louisegoodspeed @NickyPhelanBBC

drewcochrane:
Hey @pradasa my initial design is a black text with a black underline by default, thats how it should be. You can see the example in the second screenshot at the top of this Github page. 

From looking at the underline spacing between the text and link, we should increase the spacing by a couple of px

andrew-nowak:
when the popup is open the entire top bar becomes non-interactive (because the `<ng-messages>` element overlays it). It looks like you could add `max-width: fit-content` to the `.global-errors` class to fix it

kahuna/public/js/errors/global.html
```diff
@@ -42,6 +42,23 @@
...
+                 title="Close"
+                 ng-click="ctrl.dismiss('clipboard')">close</gr-icon>
+    </div>
+    <div class="global-error error autoHide" id="clipboard-max-warn" ng-message="clipboard-max-warn" tabindex="0" ng-blur="ctrl.dismiss('clipboard-max-warn')">
+        <gr-icon >warning</gr-icon>
+        You can only share a maximum of 44 images, please reduce the number of images in order to share your selection.
+        <gr-icon class="global-error__close"
+                 title="Close"
+                 ng-click="ctrl.dismiss('clipboard-max-warn')">close</gr-icon>
+    </div>
```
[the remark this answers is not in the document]
drewcochrane: Hi Andrew, apologies for the delays as I was facilitating a workshop the past two days. We came up with this full width pop up banner in order to capture our users attention in a much more accessible way and the wider accessibility aspect.  The banner is also consistent with our other internal tools at the BBC

Had conversations with the wider UX team around where it is known that disabled buttons are not being accessible hence for our full width banner design. Disabled buttons are designed in a way that they cannot be read by a screen reader (buttons are not focusable, and hence users can't reach them with a keyboard)

I've just seen the current tooltip for the first time today and it is very small. The gr-tooltip is a slight improvement. We could deploy that and have it as a question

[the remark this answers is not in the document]
andrew-nowak: thanks for the context, it's very helpful. unfortunately there's a lot of users who do large selections of images without aiming to share - the popup will be extremely annoying for them. I'm sure there's a way we can be accessible without being annoying!
I had a quick google and came across this documentation for `aria-disabled` - do you think this would make the disabled buttons usable for screenreader users? https://a11y-101.com/development/aria-disabled
Otherwise, maybe we could leave the button not-disabled and show the error when they click share, rather than when they do the selection?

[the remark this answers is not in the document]
paperboyo: Yep, banner must go for the above reason.

> could leave the button not-disabled

I would vote against as we have other disabled buttons, so it would be inconsistent.

Best would be to find a solution to screenreaders being silly about disabled buttons. Perhaps in the following PR.

kahuna/public/js/search/results.html
```diff
@@ -61,6 +61,17 @@
                     <gr-icon-label gr-icon="clear">Clear selection</gr-icon-label>
                 </a>
 
+                <a id="share-with-url" class="results-toolbar-item results-toolbar-item--right"
+                   ng-class="{'batch-archive__button--disabled': ctrl.selectionCount >= 45}"
+                   ng-click="ctrl.selectionCount < 45 && ctrl.shareSelection()"
+                   ng-if="ctrl.selectionCount > 0"
+                   aria-label="{{ctrl.selectionCount >= 45 ? 'You can't share all these images at once, please reduce to 44 or less' : 'Share Selected Images'}}"
```
paperboyo, line 68:
```suggestion
                   aria-label="{{ctrl.selectionCount >= 45 ? 'You can’t share all these images at once, please reduce to 44 or less' : 'Share selected images'}}"
```

kahuna/public/js/search/results.html
```diff
@@ -61,6 +61,17 @@
...
+                   gr-tooltip="{{ctrl.selectionCount >= 45 ? 'You can't share all these images at once, please reduce to 44 or less' : 'Share Selected Images'}}"
```
paperboyo, line 69:
```suggestion
                   gr-tooltip="{{ctrl.selectionCount >= 45 ? 'You can’t share all these images at once, please reduce to 44 or less' : 'Share selected images'}}"
```

paperboyo: Just a real apostrophe and capitalisation fix for sentence case (as other tooltips).

kahuna/public/js/search/results.js
```diff
@@ -350,6 +352,15 @@ results.controller('SearchResultsCtrl', [
...
         };
 
+        ctrl.shareSelection = async () => {
+            $q.all(ctrl.selectedImages.map(image => { return image.data.id; }).toArray()).
+            then(sharedImagesIds => {
+              const sharedUrl = $window._clientConfig.rootUri + "/search?nonFree=true&ids=" + sharedImagesIds.join(',');
+              navigator.clipboard.writeText(sharedUrl);
+              globalErrors.trigger('clipboard', sharedUrl);
+            });
+        };
```
andrew-nowak, line 362: why is this a promise? I can't see anything async here
```suggestion
        ctrl.shareSelection = () => {
            const sharedImagesIds = ctrl.selectedImages.map(image => image.data.id);
            const sharedUrl = $window._clientConfig.rootUri + "/search?nonFree=true&ids=" + sharedImagesIds.join(',');
            navigator.clipboard.writeText(sharedUrl);
            globalErrors.trigger('clipboard', sharedUrl);
        };
```

kahuna/public/js/image/controller.js
```diff
@@ -200,6 +201,11 @@ image.controller('ImageCtrl', [
       return true;
     };
 
+    ctrl.shareImage = async () => {
```
andrew-nowak, line 204:
```suggestion
    ctrl.shareImage = () => {
```

andrew-nowak (commented):
a couple of unnecessarily async functions, but otherwise lgtm

andrew-nowak (approved):
thanks!

PR #3985, opened 2022-12-06
Update basic image to Focal

andrew-nowak:
## What does this change?

Moves the "basic" instances (ones which don't require imagemagick, elasticsearch, etc.) to Ubuntu Focal. grid-imaging is already on focal, and the others will be moved in follow-up PRs.

Deploy to TEST, sense check that instances continue to function correctly.


phillipbarron (approved):
LGTM

PR #3994, opened 2022-12-14
ensure we always log a proper datetime string for `uploadTime` log marker

twrichards:
@akash1810 (of DevX) reported that there were some log ingestion failures in central ELK, owing to incompatible data types for the `uploadTime` log marker...
> ```error={reason=failed to parse field [uploadTime] of type [date] in document with id 'KkTSCoUB82L7TU1Ec37W'. Preview of field's value: 'unknown', caused_by={reason=failed to parse date field [unknown] with format [strict_date_optional_time||epoch_millis]```

This is because for forbidden uploads we were using the generic fallback of `unkown` for the `uploadTime` log marker, which is incompatible with ELKs field mappings.

Since we already fallback to _now_ for allowed uploads, it made sense to move this up slightly to make it part of the `initialContext` (good suggestion from @andrew-nowak). This should ensure we never send `unknown` for `uploadTime` log marker.

### DIFF
--- a/image-loader/app/controllers/ImageLoaderController.scala
+++ b/image-loader/app/controllers/ImageLoaderController.scala
@@ -62,12 +62,13 @@ class ImageLoaderController(auth: Authentication,
   }
 
   def loadImage(uploadedBy: Option[String], identifiers: Option[String], uploadTime: Option[String], filename: Option[String]): Action[DigestedFile] =  {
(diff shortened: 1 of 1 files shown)


andrew-nowak (approved):

akash1810 (approved):
💎

PR #3995, opened 2022-12-15
Removes params

ParisaTork:
Removes params since bucket ssm look up is set to true by default and to remove any riffraff deprecation warnings (where applicable for the repo)

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/riff-raff.yaml
+++ b/riff-raff.yaml
@@ -128,6 +128,5 @@ deployments:
   media-service-fluentbit:
     type: aws-s3
     parameters:
-      bucket: media-service-dist
       cacheControl: private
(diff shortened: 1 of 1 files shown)


andrew-nowak (changes requested):
the aws-s3 deploy type doesn't support ssm lookup (the config validator tool approves it, but deploy still errors https://riffraff.gutools.co.uk/deployment/view/f8b4180b-5706-4627-97c1-670681759163)

andrew-nowak (approved):
looks like riff-raff added support later that same day 😅 thanks @ParisaTork !

PR #3998, opened 2022-12-19
Parent/Child relationships between images (for derivates and replacements)

twrichards:
To support https://trello.com/c/8bt9DhyB/1013-spike-imaging-requests-for-web-via-pinboard when we upload _derivatives_ (cutouts, composites, retouches etc.) we need a way of recording the the media ID(s) of the image(s) which the new image **came from**. This is quite similar in functionality as the requirement that led to #4498 for 'replacing' an image and so we want to model that too. As such this PR places significance/behaviour on two new identifiers:

- `derivative-of-media-ids` (comma separated if multiple)
- `replaces-media-id`

In terms of  ingestion, fortunately we've long had the concept of 'identifiers', (which back in the day `image-loader`'s `/images` `POST` endpoint supported an `identifiers` query param, parsed as JSON - `Map[String, String]` to be precise), and ultimately ends up written to the user-metadata of the image in the S3 image bucket, e.g.
image.

Of course we use queues these days, so this PR adds support for the same identifiers on the ephemeral objects in the queue bucket (i.e. `image-loader` will _forward_ them on to the final file in the image bucket). We don't _yet_ have a need to support this in the main upload UI in the grid, so we've not updated the `/prepare` endpoint to support more than the current map of `mediaId: filename` (but could easily in the future). 
However to make uploading from imaging tooling (where the cutouts, composites etc. are produced) as simple as possible, we've added a new endpoint `/enqueueDerivativeImage` to `image-loader` which receives a single file as the payload and 
- requires `derivativeOfMediaIds` query param (where the _original_ media ID(s) can be specified, comma separated if multiple)
- requires `filename` query param
- with optional `uploadedBy`, `identifiers` & `uploadTime` query params
... which (as its name suggests) places that file in the queue, with the 

It's also surfaced in the 'Additional Metadata' section of the sidebar, in this PR we improve the display of this identifier by...

- mapping the label such that it displays as `Derivative of Media ID(s)` or `Replaces Media ID` accordingly
- making the value(s) clickable, which does a search for that media ID for `Derivative of Media ID(s)` but for `Replaces Media ID` we link directly to the image.
image image

... we also make `picdarurn` display as `Picdar URN` whilst we're there.

Being able to search for a media ID and see its 'derivatives' is useful, so in this PR we add `identifiers.derivative-of-media-ids` as one of the fields in the 'catch-all' search (it doesn't make sense to add `replaces-media-id` because as explained further down we filter out replaced images from search by default), such that we get back the following for example...
image

**Perhaps most fundamentally,** at the point of ingestion (in `image-loader`) we also create a new type of 'usage' on the parent image(s) based on the presence of these new identifier(s) on the child image, by making an 'inner-service' call to the `uploads` micro-service (it had to be 'inner-service' call, since we can't do the usual 'on behalf of' because we're processing a queue and don't have access to the users original request). The new type of usage `ChildUsage` has two types/statuses `derivative` or `replaced` and stores as a field in its metadata the `childMediaId` to point to the derivative/replacement image. 
The new usages look like this (where the media IDs are clickable to take you to the respective image)...
image

And if there is a usage with status `replaced` then it also presents this banner, linking to the child/replacement image.
image

Due to the introduction of the `childUsageMetadata` **an ElasticSearch mapping update is required** - see https://github.com/guardian/grid/blob/main/scripts/README.md#updatemapping.

Usage statuses are searchable via the `usage@status` chip and therefore as a result of all the above, **we can (and do) exclude the parent of replacement images from search by default** (CC @AndyKilmory) thanks to https://github.com/guardian/grid/pull/4519. 

----

_Follow-up PR(s) will build on this to add improved UI for displaying thumbnails etc of parent/child images and perhaps indicators on search results - but this PR adds enough functionality to effectively model the parent/child relationships between images - and be useful for ourselves & BBC._

### DIFF
--- a/image-loader/app/controllers/ImageLoaderController.scala
+++ b/image-loader/app/controllers/ImageLoaderController.scala
@@ -342,11 +342,13 @@ class ImageLoaderController(auth: Authentication,
     val tempFile = createTempFile("requestBody")(initialContext)
     val bodyParser = DigestBodyParser.create(tempFile)
 
-    AuthenticatedAndAuthorised(bodyParser) { req =>
+    AuthenticatedAndAuthorised.async(bodyParser) { req =>
 
-      val s3UserMetaMap = identifiers.map(Json.parse(_).as[Map[String, String]]).getOrElse(Map.empty) ++ Map(
+      val allIdentifiers = identifiers.map(Json.parse(_).as[Map[String, String]]).getOrElse(Map.empty) ++ Map(
         ImageStorageProps.derivativeOfMediaIdsIdentifierKey -> derivativeOfMediaIds
-      ).map { case (k, v) =>
+      )
+
+      val s3UserMetaMap = allIdentifiers.map { case (k, v) =>
         (s"${ImageStorageProps.identifierMetadataKeyPrefix}$k", v)
       }
 
@@ -355,13 +357,12 @@ class ImageLoaderController(auth: Authentication,
         filename = filename,
         s3Meta = s3UserMetaMap,
         file =  tempFile
-      )
+      ).map {_ =>
+        val newMediaId: String = req.body.digest
 
-      val newMediaId: String = req.body.digest
-
-      // TODO consider returning newMediaId (probably need kahuna to show loading page for image whilst it's queued, probs using status table) OR return uploadStatusUri
-
-      Accepted(newMediaId)
+        // return just the new media ID for now, though perhaps in future we could track these in the upload status table and provide the URI to the relevant entry
+        Accepted(newMediaId)
+      }
     }
   }
 
--- a/image-loader/app/lib/ImageLoaderStore.scala
+++ b/image-loader/app/lib/ImageLoaderStore.scala
@@ -31,7 +31,7 @@ class ImageLoaderStore(config: ImageLoaderConfig) extends lib.ImageIngestOperati
     logger.error(logMarker, s"Attempted to read $key from ingest bucket, but it does not exist.")
   }
 
-  def queueS3Object(uploader: String, filename: String, s3Meta: Map[String, String], file: File)(implicit logMarker: LogMarker): Unit = {
+  def queueS3Object(uploader: String, filename: String, s3Meta: Map[String, String], file: File)(implicit logMarker: LogMarker) = {
     store(
         config.maybeIngestBucket.get,
         s"$uploader/$filename",
--- a/kahuna/public/js/image/controller.js
+++ b/kahuna/public/js/image/controller.js
@@ -160,8 +160,8 @@ image.controller('ImageCtrl', [
     ctrl.lowResImageUri = lowResImageUri;
 
     ctrl.maybeReplacedByMediaIds = image.data.usages.data
-    .filter(usage => usage.data.status === "replaced")
-    .map(usage => usage.data.childUsageMetadata.childMediaId);
+      .filter(usage => usage.data.status === "replaced")
+      .map(usage => usage.data.childUsageMetadata.childMediaId);
 
     ctrl.singleImageList = ctrl.image ? new List([ctrl.image]) : new List([]);
 
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.js
@@ -421,14 +421,14 @@ module.controller('grImageMetadataCtrl', [
 
       ctrl.isUsefulMetadata = isUsefulMetadata;
 
-    function hasLocationInformation() {
-      return ctrl.metadata && (
-        ctrl.metadata.subLocation ||
-        ctrl.metadata.city ||
-        ctrl.metadata.state ||
-        ctrl.metadata.country
-      );
-    }
+      function hasLocationInformation() {
+        return ctrl.metadata && (
+          ctrl.metadata.subLocation ||
+          ctrl.metadata.city ||
+          ctrl.metadata.state ||
+          ctrl.metadata.country
+        );
+      }
 
       ctrl.hasLocationInformation = hasLocationInformation;
 


kahuna/public/js/preview/image.html
```diff
@@ -173,6 +174,14 @@
             </div>
         </span>
 
+        <a ng-if="ctrl.image.data.identifiers['original-media-id']"
+           class="bottom-bar__action preview__has-original-media-id"
+           title="{{'this image has has an Original Media ID\ni.e. has been modified'}}"
```
paperboyo, line 179: The nittiest of nits: I would capitalise the tooltip. And also, Guardian style is weird about them and [requires](https://www.theguardian.com/guardian-observer-style-guide-i) `ie` (no fullstops at all). Not sure how Grid writes them elsewhere, but, for ease and to be nice to our partners in crime, I would just avoid? Maybe just `This image has an Original Media ID: has been modified`? (but I’m no English speaker!)

twrichards: ok, addressed, it now reads `This image has an Original Media ID (it has been modified)`

image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -309,6 +308,44 @@ class ImageLoaderController(auth: Authentication,
...
+    val tempFile = createTempFile("requestBody")(initialContext)
+    val bodyParser = DigestBodyParser.create(tempFile)
+
+    AuthenticatedAndAuthorised(bodyParser) { req =>
+
+      val s3UserMetaMap = identifiers.map(Json.parse(_).as[Map[String, String]]).getOrElse(Map.empty) ++ Map(
+        ImageStorageProps.derivativeOfMediaIdsIdentifierKey -> derivativeOfMediaIds
+      ).map { case (k, v) =>
+        (s"${ImageStorageProps.identifierMetadataKeyPrefix}$k", v)
+      }
```
andrew-nowak, line 353: is this map supposed to apply to the merged map? or just to the derivativeOfMediaIdsIdentifierKey?
I think it's the former, but you'd need to add an extra pair of brackets to make it apply to everything

image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -309,6 +308,44 @@ class ImageLoaderController(auth: Authentication,
...
+
+      store.queueS3Object(
```
andrew-nowak, line 355: this is ignoring failures (see comment on ImageLoaderStore)

image-loader/app/lib/ImageLoaderStore.scala
```diff
@@ -30,6 +31,16 @@ class ImageLoaderStore(config: ImageLoaderConfig) extends lib.ImageIngestOperati
     logger.error(logMarker, s"Attempted to read $key from ingest bucket, but it does not exist.")
   }
 
+  def queueS3Object(uploader: String, filename: String, s3Meta: Map[String, String], file: File)(implicit logMarker: LogMarker): Unit = {
```
andrew-nowak: this returns `Unit`, so ignores the return from `store` which is a future and could fail, which we should probably report to the client

image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -309,6 +308,44 @@ class ImageLoaderController(auth: Authentication,
...
+        uploader = uploadedBy.getOrElse(Authentication.getIdentity(req.user)),
+        filename = filename,
+        s3Meta = s3UserMetaMap,
+        file =  tempFile
+      )
+
+      val newMediaId: String = req.body.digest
+
+      // TODO consider returning newMediaId (probably need kahuna to show loading page for image whilst it's queued, probs using status table) OR return uploadStatusUri
```
andrew-nowak: you're doing this below, so we can delete this TODO?

kahuna/public/js/image/controller.js
```diff
@@ -159,6 +159,10 @@ image.controller('ImageCtrl', [
     ctrl.optimisedImageUri = optimisedImageUri;
     ctrl.lowResImageUri = lowResImageUri;
 
+    ctrl.maybeReplacedByMediaIds = image.data.usages.data
+    .filter(usage => usage.data.status === "replaced")
+    .map(usage => usage.data.childUsageMetadata.childMediaId);
```
andrew-nowak:
```suggestion
      .filter(usage => usage.data.status === "replaced")
      .map(usage => usage.data.childUsageMetadata.childMediaId);
```

image-loader/app/controllers/ImageLoaderController.scala
twrichards: ooo good spot, thanks, rather than adding extra brackets I've pulled out the merging into its own val `allIdentifiers` first, easier to read. 
Done in https://github.com/guardian/grid/pull/3998/commits/77fa6c11263d7ae1067242b8da5efebaa5bbd247

image-loader/app/controllers/ImageLoaderController.scala
twrichards: great spot, thank you - done in https://github.com/guardian/grid/pull/3998/commits/77fa6c11263d7ae1067242b8da5efebaa5bbd247

image-loader/app/lib/ImageLoaderStore.scala
twrichards: great spot, thank you - done in https://github.com/guardian/grid/pull/3998/commits/77fa6c11263d7ae1067242b8da5efebaa5bbd247

image-loader/app/controllers/ImageLoaderController.scala
twrichards: replaced with a comment about potential future improvement, but not a TODO - see https://github.com/guardian/grid/pull/3998/commits/77fa6c11263d7ae1067242b8da5efebaa5bbd247

kahuna/public/js/image/controller.js
twrichards: done in https://github.com/guardian/grid/pull/3998/commits/77fa6c11263d7ae1067242b8da5efebaa5bbd247

image-loader/app/lib/ImageLoaderStore.scala
```diff
@@ -30,6 +31,16 @@ class ImageLoaderStore(config: ImageLoaderConfig) extends lib.ImageIngestOperati
     logger.error(logMarker, s"Attempted to read $key from ingest bucket, but it does not exist.")
   }
 
+  def queueS3Object(uploader: String, filename: String, s3Meta: Map[String, String], file: File)(implicit logMarker: LogMarker) = {
+    store(
+        config.maybeIngestBucket.get,
+        s"$uploader/$filename",
+        file,
+        mimeType = None, // we don't care as this is just the queue bucket
```
andrew-nowak, line 39: Images upload successfully but the enqueue request itself returns an error...

The exception comes from https://github.com/guardian/grid/blob/bd853ddb2a39e6112034e8d2e21c323bab3c37a7/common-lib/src/main/scala/com/gu/mediaservice/model/MimeType.scala#L39, saying that application/octet-stream is an unknown mime type (which it is). It seems like if we don't set the Mime type on an S3 metadata request (https://github.com/guardian/grid/blob/67ec3cdd63ea26d60263fb4ccd787f7efbb4ca80/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala#L109) it'll default to octet-stream when it is read (https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/ObjectMetadata.java#L401), such as at https://github.com/guardian/grid/blob/67ec3cdd63ea26d60263fb4ccd787f7efbb4ca80/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala#L53

So I don't think `getContentType` will ever return `null` in that last link, so we should instead catch `octet-stream` and treat that as the `None` case

twrichards: thanks @andrew-nowak - I've had a go at this in https://github.com/guardian/grid/commit/ad24535143bf96b2a288e177fd8d74f350d22f27

PR #4001, opened 2022-12-20
Add checkbox for org-owned images only

rhystmills:
Co-authored with @andrew-nowak 

## What does this change?

This PR adds a checkbox for organisation-owned images. Clicking the checkbox adds or removes an `is:` filter with a value set in config. The checkbox should not show if the value is not set.

The checkbox is also labelled by a value set in config. E.g. in `kahuna.conf`:
```
orgOwned.label="GNM-owned"
orgOwned.value="GNM-owned"
```
image

Run the Grid locally and try the checkbox with/without the required config values. Does it behave as expected?

## Tested? Documented?
- [X] locally by committer

### DIFF
--- a/kahuna/public/js/search/query.js
+++ b/kahuna/public/js/search/query.js
@@ -172,6 +172,18 @@ query.controller('SearchQueryCtrl', [
 
         ctrl.collectionSearch = ctrl.filter.query ? ctrl.filter.query.indexOf('~') === 0 : false;
 
+        const defaultNonFreeFilter = storage.getJs("defaultNonFreeFilter", true);
+        if (defaultNonFreeFilter && defaultNonFreeFilter.isDefault === true){
+          let newNonFree = defaultNonFreeFilter.isNonFree ? "true" : undefined;
+          if (newNonFree !== filter.nonFree){
+            storage.setJs("isNonFree", newNonFree ? newNonFree : false, true);
+            storage.setJs("isUploadedByMe", false, true);
+            storage.setJs("defaultNonFreeFilter", {isDefault: false, isNonFree: false}, true);
+            ctrl.filter.orgOwned = false;
+          }
+          Object.assign(filter, {nonFree: newNonFree, uploadedByMe: false, uploadedBy: undefined});
+        }
+
         if (filter.orgOwned){
             // If the checkbox is ticked, add the chip to the serach bar
             const structuredQuery = structureQuery(ctrl.filter.query);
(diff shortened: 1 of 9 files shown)


kahuna/public/js/search/query.js
```diff
@@ -228,6 +252,13 @@ query.controller('SearchQueryCtrl', [
         } else {
           ctrl.filter.nonFree = undefined;
         }
+
+        if (orgOwned === null) {
+            storage.setJs("orgOwned",ctrl.filter.orgOwned);
```
Fweddi, line 257: Does `ctrl.filter.orgOwned` need some additional checks here, similar to `ctrl.filter.nonFree` above?

Fweddi:
Tested locally and works as expected, great job!

I notice that when you hit the 'x' button on the far right of the search button, it doesn't remove the GNM-owned chip. 

This is different from other chips (the button would remove the chip), but similar to the other checkboxes (the button does not untick checkboxes). 

On balance the current state is probably the right behaviour, but worth double-checking.

Fweddi (approved):

paperboyo:
If checkbox adds a chip, chip should be removed by x (and checkbox cleared), I think.

rhystmills:
> Tested locally and works as expected, great job!
> 
> I notice that when you hit the 'x' button on the far right of the search button, it doesn't remove the GNM-owned chip.
> 
> This is different from other chips (the button would remove the chip), but similar to the other checkboxes (the button does not untick checkboxes).
> 
> On balance the current state is probably the right behaviour, but worth double-checking.

Interesting observation - it would be good to remove this chip as we do other chips by hitting that `x` on the search bar. I'll have a look into it.

rhystmills:
The chip removal issue is now sorted. I'm now looking into an issue with the checkbox not being ticked on load until you alter the query in some way.

rhystmills:
> The chip removal issue is now sorted. I'm now looking into an issue with the checkbox not being ticked on load until you alter the query in some way.
> 
> Edit: now also sorted.
> 
> I've stuck with the existing precedence of the checkbox rules - the user-set value in local storage overrules whatever URL the user inputs. This feels a bit unusual, e.g. leading to inconsistency if users want to share a query with each other. Perhaps something to revisit.

Mateusz reckons storing GNM-owned locally is not a good idea - we wouldn't want to overwrite the URL query param - I will be making some changes to the PR. It may also need a rebase.

Fweddi (approved):
New changes LGTM! Great job

PR #4008, opened 2023-01-04
make `stripDoubleQuotes` function handle undefined and non-strings rather than throwing

twrichards:
follow up to #4006 - because sometimes this function is called with things other than strings 🙈 

✅  TESTED in TEST


phillipbarron (approved):
LGTM 👍

PR #4019, opened 2023-01-20
add confirm dialog when exiting upload screen

abdelrahmansd:
## Tested? Documented?
- [ ] locally by committer


paperboyo:
Thanks! Will review functionality ASAP. Just a note, that one day we should probably apply something similar to (batch)editing in Browser too. Or just make editing independent on UI state…

paperboyo:
Testing functionality (very useful!):
1. While uploads are ongoing, it shows the dialog on tab close and tab reload in Firefox and Chrome. But on navigating browser history back – in Chrome only. Could it also warn on browser back in Firefox? 
2. [minor] Text of warning dialog differs between Firefox and Chrome:
   Firefox:
   image
    Chrome:
    image
3. I can close tab, navigate back and reload the page while metadata batch operations are ongoing. Is that expected (PR description differs here from what I can understand from the code, which mentions metadata)?

abdelrahmansd:
Thanks, @paperboyo for the review,
pt1: navigating browser back on firefox, working on my side firefox version: 108.0.2 (64-bit)
pt2: it can't set custom messages anymore in most modern browsers, I should have to remove it from the code before pushing 
pt: warning text is misleading, will remove it as mentioned in pt2

paperboyo:
Thanks! Will double-check browser-back button on Firefox. Cool about not being able to set the text.

Do we expect this to warn on navigating away while batch metadata operations are in progress on the upload page or only when actual image uploads are in progress? (would be cool to cover both, but just wanted to check what the PR is about as warning text mentioned metadata too…)

abdelrahmansd:
> Thanks! Will double-check browser-back button on Firefox. Cool about not being able to set the text.
> 
> Do we expect this to warn on navigating away while batch metadata operations are in progress on the upload page or only when actual image uploads are in progress? (would be cool to cover both, but just wanted to check what the PR is about as warning text mentioned metadata too…)

this PR is about showing confirmation when image uploads are in progress (upload Jobs)

andrew-nowak:
hey I don't think this is necessarily the right approach - it's currently not working for me on any navigation, but even if fixed I don't think `onbeforeunload` will fire when following any of the SPA links (eg. clicking the top-left icon to go home, "view all your uploads" etc.) and possibly not the back/forwards navigation either. I think you'd need to look at transition hooks https://ui-router.github.io/guide/transitionhooks#aborting-a-transition while we're still on angular

abdelrahmansd:
> 

Thanks, @andrew-nowak for the review.
you are right SPA links navigation is not working with `onbeforeunload`. this can be fixed using `$locationChangeStart`. but unfortunately, this won't be able to track the refresh/close tab and this is why `onbeforeunload` is needed.
regarding transitions hooks, I kept getting this [error](https://github.com/angular-ui/ui-router/issues/2891) when trying to use it and found out that it has been removed in angular router v0.2.18 and not available in the currently used version in our app (v0.4.3)

andrew-nowak:
thanks, that seems a little better, but it's still largely not working. 

on firefox: real page transitions show the dialog, and clicking the top-left icon. still none of the SPA page transitions (eg. clicking "search" or "view all your uploads")
on chrome: real page transitions (eg. refreshing, clicking a bookmark, typing a url etc.) don't show the dialog at all. SPA page transitions are same as firefox (ie. only clicking the top left icon)

kahuna/public/js/upload/jobs/upload-jobs.html
```diff
@@ -1,6 +1,6 @@
 <div class="flex-container section-heading current-uploads">
     <h2 class="flex-spacer">Your current uploads <em ng-if="ctrl.remaining">({{ctrl.remaining}} remaining)</em></h2>
-    <a ng-controller="SessionCtrl" class="coloured-link" ui-sref="search.results({uploadedBy: user.email, nonFree: true})"
+    <a ng-controller="SessionCtrl" class="coloured-link" ng-href="/search?uploadedBy={{user.email}}&nonFree=true"
```
abdelrahmansd, line 3: Hi @andrew-nowak , regarding this change to use `ng-href` instead of `ui-sref` because `ui-sref` actions are not a part of  `$locationChangeStart` events
because it uses the `$location` provider,
and  it can be handled by `$stateChangeStart` since it happens when the state change
so either keep it as it is or revert it back to `ui-sref` and use `$stateChangeStart`
what do you think?
```
$scope.$on('$stateChangeStart', (event, toState, toParams, fromState, fromParams) => {
      if (fromState.url !== toState.url && fromState.url === "/upload") {
        ctrl.displayWarning(event);
    }
  });
```

andrew-nowak: If you find that `$stateChangeStart` works, I think I'd be a lot more comfortable with that. As it happens, it looks like ui-router catches clicks on `<a href="...">`s  and turns them into state transitions - but this seems to be undocumented and definitely took me by surprise, so I'd like to avoid relying on it if at all possible.

That said, I've tried it a few times and haven't directly found any problems with it, so if we can't get `$stateChangeStart` to work then I guess we can go with this approach.

abdelrahmansd: Hi Andrew, thanks for the feedback, at this moment this PR has been tested on Edge, Firefox, Chrome, and Safari and working as expected
I will double-check and test with the `$stateChangeStart` events handler and back to you

abdelrahmansd: Hi @andrew-nowak, it is working but not stable in some cases, so the current approach is more robust and tested successfully, for all browsers

abdelrahmansd:
Hi @andrew-nowak  also I will add another commit to show the confirmation dialog when the upload job is in the indexing state well as the uploading state

andrew-nowak (approved):

PR #4030, opened 2023-02-03
bring back syndication icon colours

andrew-nowak:
## What does this change?

We used to have colours on the syndication icons, they were useful, @paperboyo noticed they're gone, tracked it down to this (repeated class attributes doesn't work, they need to be space delimited)


paperboyo (approved):
I approve of this effort (blindly)!

PR #4035, opened 2023-02-13
Set syndication filters based on `now` to be `def`

andrew-nowak:
## What does this change?

If they're `val` then `DateTime.now` will only be evaluated once - at instance startup.

If instances survive for a moderate period of time, we will start seeing images that are all set to join the queue appear to go missing, because the check will go out of date


paperboyo (approved):
Thanks for that!

twrichards (commented):
nice. 5 years its been broken 🤦

PR #4059, opened 2023-03-31
expose count of 'org-owned' images within search results

twrichards:
Following on from #4001 - we'd like to surface that within any given search results, there are X images which are 'org-owned' (e.g. taken by staff photographers), as its desirable to use these images over agency images in many cases for a number of reasons.

This PR does exactly that, by adding a count beside the existing count of 'matches'...

... this is clickable, to apply the 'org-owned' chip (which is already synced with the 'org-owned' checkbox added in #4001). It obviously doesn't show if the 'org-owned' filter is applied, as it would be redundant (because counts would always be the same).

We do the same for the 'new images' badge which appears based on the polling which occurs in the background...


twrichards:
alternative would be to have the count on the checkbox added in #4001 ... 
image

rhystmills:
Would be worth updating the local config we pull from s3 to include `filters.shouldDisplayOrgOwnedCountAndFilterCheckbox=true` by default once this is released.

twrichards:
> Would be worth updating the local config we pull from s3 to include `filters.shouldDisplayOrgOwnedCountAndFilterCheckbox=true` by default once this is released.

@rhystmills I'd already added it to the `generate-config` script used in `setup.sh` for local development. Currently the `--use-TEST` flag for `start.sh` only copies `media-api.conf` and `kahuna.conf` from TEST, so I'm not sure copying down `common.conf` too is a good idea - happy to discuss...

twrichards:
## Tested in `TEST` ✅

Fweddi (approved):
Looks smashing! Works smashingly!

rhystmills:
> > Would be worth updating the local config we pull from s3 to include `filters.shouldDisplayOrgOwnedCountAndFilterCheckbox=true` by default once this is released.
> 
> @rhystmills I'd already added it to the `generate-config` script used in `setup.sh` for local development. Currently the `--use-TEST` flag for `start.sh` only copies `media-api.conf` and `kahuna.conf` from TEST, so I'm not sure copying down `common.conf` too is a good idea - happy to discuss...

Ah, didn't realise, that sounds sensible 👍

PR #4069, opened 2023-04-17
'preview migration completion' in `thrall` dashboard (i.e. `media-api` only searches new index)

twrichards:
This introduces a new migration status `CompletionPreview` (with  some endpoints and basic UI in `thrall` dashboard to transition in/out of this state). When in this state `media-api` only searches the new index, to give users the illusion migration is complete, whilst still keeping both indexes up to date (this will be particularly useful when making use of https://github.com/guardian/grid/pull/4066), and allowing us to 'unpreview' migration completion if we discover some images are missing etc. (much harder when migration is actually completed as the old index will stop being updated, so would require manual reconciliation).

### when `InProgress` or `Paused`...
image

### when in new `CompletionPreview`...
image

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/MigrationStatusProvider.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/MigrationStatusProvider.scala
@@ -16,6 +16,7 @@ sealed trait Running extends MigrationStatus {
 }
 case class InProgress(migrationIndexName: String) extends Running
 case class Paused(migrationIndexName: String) extends Running
+case class CompletionPreview(migrationIndexName: String) extends Running
 case class StatusRefreshError(cause: Throwable, preErrorStatus: MigrationStatus) extends MigrationStatus
 object StatusRefreshError {
   // custom constructor to unwrap when previousStatus is also Error - prevents nested Errors!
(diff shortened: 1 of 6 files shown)


Fweddi:
Tested, index switches as expected. Code looks good.

I would maybe add some logic to stop you from 'previewing' the migration when it's still ongoing. I know it's advised, but still, I don't think this would be a good state to end up in?

Also, this new change is a little difficult to parse. `UnPreview` just about works, but it becomes a little hard (but not impossible) to reason about when next to the text and button about completing the migration. Might just be the accumulation of Web 1.0 with no styling...

Maybe one solution would be to explain this in plainer English in the migration docs?

Fweddi:
> > Tested, index switches as expected. Code looks good.
> > I would maybe add some logic to stop you from 'previewing' the migration when it's still ongoing. I know we advise against it, but perhaps we should enforce it?
> > Also, this feature is a little difficult to understand in the thrall dashboard. Might just be the accumulation of Web 1.0 noise with no styling...
> > Maybe one solution would be to explain this in plainer English in the migration docs?
> 
> Thank you for the review @Fweddi
> 
> I'd rather keep the explainer text in the application, so please feel free to make a suggestion of something clearer (it's HTML so needn't be just a sentence e.g. bullet points)...

I'm not sure how I'd phrase it, and I don't have the time to look at this in much more detail, so I'm going to go ahead and approve this on the grounds we want to get this out soon.

Fweddi (approved):

PR #4075, opened 2023-04-25
attempt to fix fluentbit config

twrichards:
## Tested? Documented?
- [ ] locally by committer


twrichards:
this is not the best fix - instead we're going to move the applications to use https://github.com/guardian/devx-logs (via the `cdk-base` Amigo role)

PR #4080, opened 2023-04-26
revert some overzealous refactors in #4059 which was dropping quotes etc.

twrichards:
In https://github.com/guardian/grid/pull/4059/files#diff-7b64a31f35126c904a2d8c522158f0b090e63a4045fca6123f6a583ad83cf2f3L201-L207 I attempted to simplify what looked like verbose logic to filter out the org-owned chip, when in fact the previous logic (added by @rhystmills in #4001) was only executing once when the org-owned chip needed removing after having unticked the org-owned checkbox, whereas my refactor was running on every key press, and due to long standing issues was dropping quotes etc. immediately after being typed.

This PR reinstates @rhystmills' superior logic, albeit with some refinements.

### DIFF
--- a/kahuna/public/js/search/query.js
+++ b/kahuna/public/js/search/query.js
@@ -153,6 +153,7 @@ query.controller('SearchQueryCtrl', [
     });
 
 
+    // eslint-disable-next-line complexity
     $scope.$watchCollection(() => ctrl.filter, onValChange(filter => {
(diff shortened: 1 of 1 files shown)


paperboyo (approved):
👍 I approve of this code-blindly.

jonathonherbert (approved):

PR #4084, opened 2023-05-03
hide 'org-owned' badge if zero

twrichards:
Slight tweak following https://github.com/guardian/grid/pull/4059, @paperboyo thinks it would be nice to hide the `org-owned` badge if the count is `zero`

### DIFF
--- a/kahuna/public/js/search/results.html
+++ b/kahuna/public/js/search/results.html
@@ -30,7 +30,7 @@
                         gr-tooltip-updates>
                     {{ctrl.newImagesCount | toLocaleString}} new
                 </button>
-                <button ng-if="ctrl.orgOwnedCount !== undefined && ctrl.orgOwnedCount !== ctrl.totalResults"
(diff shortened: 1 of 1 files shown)


kahuna/public/js/search/results.html
```diff
@@ -30,7 +30,7 @@
                         gr-tooltip-updates>
                     {{ctrl.newImagesCount | toLocaleString}} new
                 </button>
-                <button ng-if="ctrl.orgOwnedCount !== undefined && ctrl.orgOwnedCount !== ctrl.totalResults"
+                <button ng-if="ctrl.orgOwnedCount !== undefined && ctrl.orgOwnedCount !== ctrl.totalResults && (ctrl.orgOwnedCount + (ctrl.newOrgOwnedCount || 0))"
```
andrew-nowak, line 33: nit: Is the condition implicitly `(ctrl.orgOwnedCount + (ctrl.newOrgOwnedCount || 0)) > 0`? it might be worth writing it out long-form just to avoid confusion
alternatively, this condition's getting fairly long - worth moving to controller and giving it a name?

andrew-nowak (approved):
lgtm!

PR #4096, opened 2023-05-30
bump Scala dependencies

aracho1:
## What does this change?
Bumped a number of Scala dependencies which which have reduced the number of high Snyk vulnerabilities from 13 to 3.

The last remaining 3 require major changes which are probably best addressed as separate PRs.

 

Grid runs as normal locally and on TEST.

Grid works as usual and the number of high vulns are down at the same time.

## Who should look at this?

@guardian/digital-cms 
## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -123,7 +123,8 @@ lazy val commonLib = project("common-lib").settings(
     "com.gu" %% "scanamo" % "1.0.0-M8",
     "com.fasterxml.jackson.core" % "jackson-databind" % "2.9.10.7",
     // Necessary to have a mix of play library versions due to scala-java8-compat incompatibility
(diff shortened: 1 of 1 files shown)


aracho1:
> I don't think we use protobuf in this project directly; unsure how to exercise that. Just to check: why was the protobuf dependency necessary to include?

Protobuf was one of the high vuls introduced via amazon-kinesis-client but bumping the kinesis client to the required version breaks the application so the easiest way around that was specifying the protobuf version. Paired with DevX on this so fairly confident this approach is OK.

jonathonherbert:
@aracho1 – thanks for the explanation, makes perfect sense!

PR #4099, opened 2023-06-02
Bump AngularJS to 1.8.3

jonathonherbert:
## What does this change?

Bumps AngularJS and related dependencies to 1.8.3.

Borrows from https://github.com/guardian/flexible-content/pull/4346. As there are no AngularJS specs in Grid, a few of the problems in that PR don't apply here. We also don't make any JSONP calls that I can find.

I strongly recommend reviewing with [whitespace hidden,](https://github.com/guardian/grid/pull/4099/files?diff=split&w=1) as most changes involve reindenting large amounts of code!

A list of issues encountered during the upgrade:

### Directive not seeing properties added to controller.

**How to spot:** likely to vary, but you may see `ctrl.{propertyName} is undefined`, indicating that properties passed to directives have not been initialised before their controller is instantiated.

This affected most of the directives in this project.

Caused by [this change.](https://docs.angularjs.org/guide/migration#migrate1.5to1.6-ng-services-$compile) Moving initialisation code into `this.$onInit` necessary in directive controllers. f6dc3ed6c29143923eef2782dd6c6e48f2027e34 disables the now deprecated behaviour, and 09d177d29f4aa50fe4e48c75b7eb24d1be52a657 fixes.

### Invalid HTML causes parsing issues

Caused by a [change to jqLite.](https://docs.angularjs.org/guide/migration#migrating-from-1-7-to-1-8) Fixed by removing self-closing tags where appropriate: [ae98bc7](https://github.com/guardian/grid/pull/4099/commits/ae98bc765dae2073dd53895789f551b94d147529)

### How to test

There's no test coverage for the Angular UI that I can see, and this PR necessarily touches a large number of components, so this needs a thorough test in TEST to be confident that users won't see problems. 

Recommend testing with dev tools open, as errors caused by this upgrade are likely to result in red message of the sort mentioned above:  `ctrl.{propertyName} is undefined`.

Tested in TEST:

- [x]  Image list – scrolling, selection, multi-selection, presence of badges, editing single and multi-selected images with every available property
- [x] Collections – browsing collections in lhs pane, creating and removing nested collections, adding collections to images 
- [x] Searching – adding and removing chips, adding and removing filters, adding and removing date ranges, altering sort order, connection between collections and search 
- [x] Image preview – browsing through many images 
- [x] Single image – edit individual data
- [x] Crop – crop image too small, crops at custom sizes, crops at preset sizes 
- [x] Upload – upload images, adjust settings
- [x] Integration – single crop in Composer, multiple crops in Composer

What did I miss? 😅


rhystmills (commented):
Code looks good to me, have tried everything I can think to do in Kahuna and can't find anything that's not working 👍

rhystmills (approved):

PR #4108, opened 2023-06-16
handle update fn returning false in batch edit

andrew-nowak:
## What does this change?

the `batchUpdateMetadataField` function is responsible for managing metadata changes (by delegating to `updateMetadataField`) and propagating those changes back into the model. If you've selected several images and change a field in such a way that one image has no change to be applied, then `updateMetadataField` will return `false` for that image ( this communicates "no change" to angular-xeditable - see this comment https://github.com/guardian/grid/blob/51734b8443d1cafd246818eddde4ea3c36e4a86a/kahuna/public/js/edits/service.js#L257-L269 ). This meant that we now tried to replace the image model with `false`, which is missing all of the expected fields of an image object causing some UI breakage and some very noisy console logs.

we now catch this case, and return the original image unchanged instead of the `false`

cc @paperboyo

### DIFF
--- a/kahuna/public/js/edits/service.js
+++ b/kahuna/public/js/edits/service.js
@@ -328,13 +328,11 @@ service.factory('editsService',
 
 
     function batchUpdateMetadataField(images, field, value, editOption = overwrite.key) {
-        return trackAll($q, $rootScope, field, images, async (image) => {
+        return trackAll($q, $rootScope, field, images, (image) => {
             const newFieldValue = getNewFieldValue(image, field, value, editOption);
-            const updated = await updateMetadataField(image, field, newFieldValue, true);
-            if (!updated) { // updateMetadataField returns false if no change
-                return image;
-            }
-            return updated;
+            return updateMetadataField(image, field, newFieldValue, true).then(
+            	updated => updated || image; // updateMetadataField returns false if no change
+            );
         }, 'images-updated');
     }
 
--- a/kahuna/public/js/edits/service.js
+++ b/kahuna/public/js/edits/service.js
@@ -330,7 +330,11 @@ service.factory('editsService',
     function batchUpdateMetadataField(images, field, value, editOption = overwrite.key) {
         return trackAll($q, $rootScope, field, images, image => {
             const newFieldValue = getNewFieldValue(image, field, value, editOption);
-            return updateMetadataField(image, field, newFieldValue, true);
+            const updated = updateMetadataField(image, field, newFieldValue, true);
+            if (!updated) { // updateMetadataField returns false if no change
+                return image;
+            }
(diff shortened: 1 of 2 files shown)


kahuna/public/js/edits/service.js
```diff
@@ -328,10 +328,14 @@ service.factory('editsService',
...
-        return trackAll($q, $rootScope, field, images, image => {
+        return trackAll($q, $rootScope, field, images, async (image) => {
             const newFieldValue = getNewFieldValue(image, field, value, editOption);
-            return updateMetadataField(image, field, newFieldValue, true);
-        },'images-updated');
+            const updated = await updateMetadataField(image, field, newFieldValue, true);
+            if (!updated) { // updateMetadataField returns false if no change
+                return image;
+            }
+            return updated;
```
twrichards, line 337:
```suggestion
            return updated || image; // updateMetadataField returns false if no change
```

twrichards, line 337: or possibly even (function wouldn't need to be async then)
```suggestion
            return updateMetadataField(image, field, newFieldValue, true).then(
            	updated => updated || image; // updateMetadataField returns false if no change
            );
```

twrichards (approved):
code looks good, and I imagine you've tested also looks like @paperboyo deployed to `TEST` so presume it works.

i've made some suggestions to make the code more succinct

kahuna/public/js/edits/service.js
```diff
@@ -328,10 +328,14 @@ service.factory('editsService',
 
 
     function batchUpdateMetadataField(images, field, value, editOption = overwrite.key) {
-        return trackAll($q, $rootScope, field, images, image => {
+        return trackAll($q, $rootScope, field, images, async (image) => {
```
andrew-nowak, line 331:
```suggestion
        return trackAll($q, $rootScope, field, images, (image) => {
```

kahuna/public/js/edits/service.js
```diff
@@ -328,10 +328,12 @@ service.factory('editsService',
 
 
     function batchUpdateMetadataField(images, field, value, editOption = overwrite.key) {
-        return trackAll($q, $rootScope, field, images, image => {
+        return trackAll($q, $rootScope, field, images, (image) => {
             const newFieldValue = getNewFieldValue(image, field, value, editOption);
-            return updateMetadataField(image, field, newFieldValue, true);
-        },'images-updated');
+            return updateMetadataField(image, field, newFieldValue, true).then(
+            	updated => updated || image; // updateMetadataField returns false if no change
```
andrew-nowak, line 334:
```suggestion
            	updated => updated || image // updateMetadataField returns false if no change
```

PR #4111, opened 2023-06-26
script to purge stuff from S3 following 'reaping migration' etc.

twrichards:
script to purge stuff from S3 following https://github.com/guardian/grid/pull/4066 (plus clearing other detritus)


twrichards:
We're going with an AWS Athena approach as the volumes of inventory information (50+ gzipped CSVs per day, ~50MB each) is too much to deal with locally. 

May open a similar PR just for a script to extract all the IDs from ES (which we can load into Athena).

PR #4114, opened 2023-06-27
script to fetch all ElasticSearch IDs

twrichards:
We need to cross reference the IDs available in the grid UI (i.e. ElasticSearch) before reaping loads of files from S3. There's no direct way of doing this in ElasticSearch, it has to be done with a query, which is most efficient as a 'scan and scroll' (see https://stackoverflow.com/a/30855670) so this adds a script to do just that and write to file - for example a CSV file for upload to AWS Athena (see [#4111 ](https://github.com/guardian/grid/pull/4111#issuecomment-1609279848))

Seems to work nicely for `TEST` (finished in a couple of mins)...
image

### DIFF
--- a/scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
+++ b/scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
@@ -366,13 +366,15 @@ object DownloadAllEsIds extends EsScript {
           prevResponse.result.hits.hits.map(_.id).foreach(writer.println)
 
           // can't be map/foreach if we want tail recursion (for efficiency)
-          if(prevResponse.result.scrollId.isDefined) {
-            processResponseAndStartNext(
-              client.execute(
-                searchScroll(prevResponse.result.scrollId.get)
-                  .keepAlive(10.seconds)
-              ).await
-            )
+          prevResponse.result.scrollId match {
+            case None => () // done
+            case Some(scrollId) =>
+              processResponseAndStartNext(
+                client.execute(
+                  searchScroll(scrollId)
+                    .keepAlive(10.seconds)
+                ).await
+              )
           }
         }
 
@@ -396,7 +398,7 @@ object DownloadAllEsIds extends EsScript {
 
 
   def usageError: Nothing = {
-    System.err.println("Usage: GetAllEsIds <ES_URL>")
+    System.err.println("Usage: GetAllEsIds <ES_URL> <output_file>")
     sys.exit(1)
   }
 }


scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
```diff
@@ -337,6 +339,67 @@ object GetSettings extends EsScript {
...
+            writer.close()
+            client.close()
+            throw new Exception("Failed performing search query")
+          }
+
+          println(s"Adding ${prevResponse.result.hits.hits.length} IDs to ${file.getAbsolutePath}")
+
+          prevResponse.result.hits.hits.map(_.id).foreach(writer.println)
+
+          // can't be map/foreach if we want tail recursion (for efficiency)
```
andrew-nowak, line 368: but could be a match statement :)

scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
```diff
@@ -337,6 +339,67 @@ object GetSettings extends EsScript {
...
+
+    println(s"Getting all ids from $esUrl")
+
+    object Client extends EsClient(esUrl) {
+
+      def writeAllEsIdsToFile(filename: String) = {
+
+        val file = new File(filename)
+
+        val writer = new PrintWriter(file)
```
andrew-nowak, line 354: reading through the class docs, I don't think this will automatically flush the buffer, so I think it's possible you could end up with the report entirely buffered in memory, which might cause an OOM on large indices - though idk if our PROD is big enough to cause that
Might be worth explicitly `.flush()`ing every 100k lines or something, just to prevent huge buildups?

andrew-nowak (approved):
logic lgtm, love a good tailrec

scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
andrew-nowak: alternatively you could construct with a `FileOutputStream`, and then you can set automatic flushing (though that'll flush on each println, which I think might be excessive) https://docs.oracle.com/javase/8/docs/api/java/io/PrintWriter.html#PrintWriter-java.io.OutputStream-boolean-

twrichards: good thought, but it seems to flush fairly promptly (as it does the same in https://github.com/guardian/grid/pull/4115) and it didn't OOM when running against PROD

scripts/src/main/scala/com/gu/mediaservice/scripts/EsScript.scala
twrichards: good suggestion, done in https://github.com/guardian/grid/pull/4114/commits/4fe250a12340b6ef41759080ced09dde6e6f5d9f and re-tested

PR #4115, opened 2023-07-03
add script to batch delete files from S3

twrichards:
Following on from https://github.com/guardian/grid/pull/4114 and lots of follow-up data mashing in Athena we now have a list of S3 paths we can safely delete (not in ES anymore, but that we have a copy in the replica bucket [and delete marker replication is OFF]) - this stands at over 50million in PROD, so we need a way to bulk delete...

**```Usage: BulkDeleteS3Files <bucketName> <inputFile> <auditFile>```**

Input file needs to be a CSV, with a heading row and a single column containing the S3 paths to delete from the specified bucket.

This script groups the input IDs into 1000s so it can use the [bulk delete API](https://docs.aws.amazon.com/AmazonS3/latest/userguide/delete-multiple-objects.html) and reports the success or failure for each S3 path to both the console but also the `auditFile` path provided (CSV output).

Note: [bulk delete API](https://docs.aws.amazon.com/AmazonS3/latest/userguide/delete-multiple-objects.html) reports 'deleted' if the path is not found, so this can be run multiple times without issue.

### TESTED in `TEST` ✅ 
Processed 385k deletions in a few mins.

### DIFF
--- a/scripts/src/main/scala/com/gu/mediaservice/scripts/BulkDeleteS3Files.scala
+++ b/scripts/src/main/scala/com/gu/mediaservice/scripts/BulkDeleteS3Files.scala
@@ -1,0 +1,69 @@
+package com.gu.mediaservice.scripts
+
+import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider
+import software.amazon.awssdk.regions.Region
+import software.amazon.awssdk.services.s3.S3Client
+import software.amazon.awssdk.services.s3.model.{Delete, DeleteObjectsRequest, ObjectIdentifier}
+
+import java.io.{File, PrintWriter}
+import scala.io.Source
+import scala.jdk.CollectionConverters.iterableAsScalaIterableConverter
+
+object BulkDeleteS3Files {
+
+  private val profile = "media-service"
+  private val region = Region.EU_WEST_1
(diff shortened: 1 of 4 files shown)


scripts/src/main/scala/com/gu/mediaservice/scripts/BulkDeleteS3Files.scala
```diff
@@ -0,0 +1,69 @@
...
+        }
+        failedDeletesLogLines = batchResult.errors().asScala.map{ error =>
+          s"${error.key()}, FAILED, ${error.code()}, ${error.message()}"
+        }
+        logLines = successfulDeletesLogLines ++ failedDeletesLogLines
+        unit = logLines.foreach {logLine =>
+          println(logLine)
+          auditFileWriter.println(logLine)
+        }
+      } yield unit
```
andrew-nowak, line 60:
```suggestion
        logLine <- successfulDeletesLogLines ++ failedDeletesLogLines
      } {
        println(logLine)
        auditFileWriter.println(logLine)
      }
```

I think this is equivalent, and a bit more idiomatic/readable

andrew-nowak (approved):
logic lgtm

PR #4122, opened 2023-07-18
Add API filters for print usage info

andrew-nowak:
## What does this change?

Adds some media API search filters to allow drilling down into images which have print usages which meet several criteria:

- printUsageIssueDate
- printUsageSectionCode
- printUsagePageNumber
- printUsageEdition
- printUsageOrderedBy

printUsageIssueDate is required to query on print usages like this (there's no technical reason, but it doesn't really make sense to query without it)

These use filters rather than query syntax / chips because that I originally thought that gave us an extra level of control over the querying of nested fields (I thought that `+usage@issueDate:2023-01-01 +usage@sectionCode:1jo` would give us images where at least one usage was on the date, and at least one usage was in the section, but not necessarily the same usage. I'm less convinced of that now.). But I still think it's nicer for our purposes to use filters than chips (where grid UI helps phrase the chip correctly, but in API you're on your own). Lmk if you disagree and I'll double check viability of using chips instead.

Make some print usages, try to slice them with queries.

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/media-api/app/lib/elasticsearch/filters.scala
+++ b/media-api/app/lib/elasticsearch/filters.scala
@@ -22,6 +22,25 @@ object filters {
 
   def boolTerm(field: String, value: Boolean): TermQuery = termQuery(field, value)
 
+  /**
+   * Range query based on dates
+   * Note that `from` is _inclusive_ in this function
+   * @param field Field name to query
+   * @param from Lower bound for date (inclusive)
+   * @param to Upper bound for date (exclusive)
+   * @return Suitable query
+   */
+  def date(field: String, from: DateTime, to: DateTime): Query =
+    rangeQuery(field).gte(printDateTime(from)).lt(printDateTime(to))
+
+  /**
+   * Range query based on dates - handles optional to and from
+   * Note that `from` is _exclusive_ in this function
+   * @param field Field name to query
+   * @param from Lower bound for date (exclusive)
+   * @param to Upper bound for date (exclusive)
+   * @return Suitable query if at least one of `from` and `to` is defined; otherwise nothing
+   */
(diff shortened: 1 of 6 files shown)


media-api/app/lib/elasticsearch/filters.scala
```diff
@@ -22,6 +22,25 @@ object filters {
...
+    rangeQuery(field).gte(printDateTime(from)).lt(printDateTime(to))
+
+  /**
+   * Range query based on dates - handles optional to and from
+   * Note that `from` is _exclusive_ in this function
+   * @param field Field name to query
+   * @param from Lower bound for date (exclusive)
+   * @param to Upper bound for date (exclusive)
+   * @return Suitable query if at least one of `from` and `to` is defined; otherwise nothing
+   */
```
andrew-nowak, line 43: not a fan that I've made the two overloads of `date` have different semantics on the `for` attribute :( 
how would feel about making the old/optional/bottom one also use `gte` (ie. an inclusive lower bound) ? I don't _think_ this should have any major impact - its usages are either for inexact fields (date taken, upload time etc.) or are used when filtering based on `now`

twrichards (approved):
logic LGTM and I've tested in `TEST` with `/images?printUsageIssueDate=2023-07-18&printUsagePageNumber=5` for example - works nicely. 
Will reply separately on your comment about `date` function overload

media-api/app/lib/elasticsearch/filters.scala
twrichards: based on your assessment of impact, perhaps go for that.
an alternative, with no risk would be to change your new function to be exclusive on the `from` too, then increment the day at the call site instead.

PR #4125, opened 2023-07-31
support 'projecting' from replica bucket as a way of restoring images incorrectly hard deleted

twrichards:
Following on from the excellent https://github.com/guardian/grid/pull/4116. Adds a fallback to the 'replica bucket' (see https://github.com/guardian/editorial-tools-platform/pull/696 for more about replica buckets) when the file cannot be found in the main image bucket.

This should help us restore images quickly which have been hard deleted by mistake (perhaps by reaper) - so long as we have the ID of course.

**Requires new (optional) bit of image loader config `s3.image.replicaBucket`.**

### DIFF
--- a/image-loader/app/model/Projector.scala
+++ b/image-loader/app/model/Projector.scala
@@ -32,7 +32,7 @@ object Projector {
   import Uploader.toImageUploadOpsCfg
 
   def apply(config: ImageLoaderConfig, imageOps: ImageOperations, processor: ImageProcessor, auth: Authentication)(implicit ec: ExecutionContext): Projector
-  = new Projector(toImageUploadOpsCfg(config), S3Ops.buildS3Client(config), imageOps, processor, auth)
+  = new Projector(toImageUploadOpsCfg(config), S3Ops.buildS3Client(config), S3Ops.buildS3Client(config, region = "us-west-1"), imageOps, processor, auth)
 }
 
 case class S3FileExtractedMetadata(
@@ -83,6 +83,7 @@ object S3FileExtractedMetadata {
 
 class Projector(config: ImageUploadOpsCfg,
                 s3: AmazonS3,
+                replicaS3: AmazonS3,
                 imageOps: ImageOperations,
                 processor: ImageProcessor,
                 auth: Authentication) extends GridLogging {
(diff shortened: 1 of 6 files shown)


image-loader/app/model/Projector.scala
```diff
@@ -95,11 +95,23 @@ class Projector(config: ImageUploadOpsCfg,
...
+          config.originalFileBucket
+      }
+
+      if (config.maybeReplicaBucket.contains(bucket) && !s3.doesObjectExist(bucket, s3Key)) {
+        logger.error(
+          logMarker,
+          s"Image with id $imageId does not exist at key $s3Key in S3 bucket ${config.originalFileBucket}${config.maybeReplicaBucket.map(rb => s" NOR replica bucket $rb")}"
+        )
+        throw new NoSuchImageExistsInS3(bucket, s3Key)
+      }
```
andrew-nowak, line 111: this logic is a bit tough to follow - I think you can be a little adventurous with the match statement and get something nicer:

```suggestion
      val bucket = config.maybeReplicaBucket match {
        case _ if s3.doesObjectExist(config.originalFileBucket, s3Key) =>
          config.originalFileBucket
        case Some(replicaBucket) if s3.doesObjectExist(replicaBucket, s3Key) =>
          replicaBucket
        case _ =>
          logger.error(
            logMarker,
            s"Image with id $imageId does not exist at key $s3Key in S3 bucket ${config.originalFileBucket}${config.maybeReplicaBucket.map(rb => s" NOR replica bucket $rb")}"
          )
          throw new NoSuchImageExistsInS3(config.originalFileBucket, s3Key)
      }
```

not entirely happy with the top case being a `case _` and being dependent on originalFileBucket though - hard to follow in its own way. Possibly it might read a bit better matching on the tuple (config.originalFileBucket, config.maybeReplicaBucket), even if the original is non-Option so a superfluous match?

this block might also work better as a little standalone function?

andrew-nowak (approved):
code seems good to me, though I think the main logic can be made easier to follow - I haven't tested in TEST myself but trust you have :)

twrichards:
This approach doesn't reinstate the image file in the image bucket (nor does it recreate thumbs or optimised image, which will have been deleted during reaping), so closing this PR in favour of https://github.com/guardian/grid/pull/4128 which achieves the intentions of this PR but via `image-loader` so it can regenerate all it needs but preserving original `uploadTime`, `uploadedBy` and original filename.

PR #4127, opened 2023-08-01
add youtube as a valid `frame-src` in CSP

twrichards:
... so pinboard can render video atoms in the grid

Avoids this...
image


andrew-nowak (approved):

twrichards:
Yep works in `TEST`...
image

PR #4131, opened 2023-08-21
restart syncchecker on failure with backoff

andrew-nowak:
## What does this change?

Sometimes Elasticsearch is unavailable, and queries fail with 502. Akka streams by default exit on failures, which is reasonable. We can wrap the source in a RestartSource https://doc.akka.io/docs/akka/current/stream/operators/RestartSource/onFailuresWithBackoff.html to restart when the stream exits with failure. 

(We could consider doing the same for the Thrall stream, but that seems more resilient to these transient errors, and is monitored by the healthcheck endpoint, causing a redeploy if down. Regardless, for more consideration and a future change)

### DIFF
--- a/thrall/app/lib/SyncChecker.scala
+++ b/thrall/app/lib/SyncChecker.scala
@@ -2,8 +2,8 @@ package lib
 
 import akka.Done
 import akka.actor.ActorSystem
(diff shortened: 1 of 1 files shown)


twrichards (approved):
awesome!

Issue #403, opened 2015-01-28
Uploading portrait images from my camera renders them landscape


steppenwells:
While testing I uploaded some of my holiday snaps. The portrait images get uploaded as landscape so it looks like the embedded orientation metadata is not being respected.

The image renders portrait in finder on osx so it's not a problem with the picture per-se.

This might need looking at if start uploading images directly from photographers etc.

Test image:

![img_3118](https://cloud.githubusercontent.com/assets/145254/5940596/d45f2ade-a703-11e4-83cc-8aa90bb83e34.JPG)

in finder

![screen shot 2015-01-28 at 15 40 11](https://cloud.githubusercontent.com/assets/145254/5940635/05c5335c-a704-11e4-940f-6dbcbc39aa5a.png)

paperboyo:
> Then again I've never seen a single image that had this issue in PROD, so it seems low priority to me.

See e.g.: https://media.gutools.co.uk/search?query=chitra%20ramaswamy&nonFree=true

I would say it’s unlikely to happen to agency-provided images, but very likely to happen to straight-from-camera uploads directly from photographers. Expecting uploaders to fix it before upload is unrealistic and can lead to lowering image quality (lossy JPG resave). Also see arguments in https://github.com/guardian/grid/issues/1691.

paperboyo:
https://github.com/drewnoakes/metadata-extractor/issues/119

paperboyo:
Also GraphicsMagick as of 1.3.26 (July 4, 2017) contains support for EXIF orientation auto-rotate and tag clearing.

paperboyo:
[This link](https://zpl.fi/exif-orientation-in-different-formats/) may come in handy.

PR #4132, opened 2023-08-22
prevent autorotation on image display and crop page

andrew-nowak:
## What does this change?

Some images with rotation metadata get autorotated if the requested image means imgops doesn't have to transform.
Browsers see the metadata and do the rotation.
There is a CSS property to control this, but to take effect it needs the `crossorigin` property set on the `` - even if setting to the default `anonymous`

There is another bug in Firefox <= 113 which will prevent this having an effect.

Check with an image with rotation metadata (@paperboyo knows one in guardian grid)

Grid chooses to explicitly ignore rotation metadata (until we also support it in the backend! then we'll make good use of it!)

### DIFF
--- a/kahuna/public/js/crop/view.html
+++ b/kahuna/public/js/crop/view.html
@@ -105,6 +105,7 @@
     <div class="easel__canvas">
         <div class="easel__image-container" oncontextmenu="return false;">
             <img class="easel__image easel__image--cropper"
+                 crossorigin="anonymous"
                  oncontextmenu="return false;"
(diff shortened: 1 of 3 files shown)


paperboyo (approved):
👏  Great debugging, thank you!

This needs to be (probably) removed after we (finally) fix https://github.com/guardian/grid/issues/403.

andrew-nowak:
turns out it doesn't work on cropper page, (needs more investigation), but it doesn't make anything worse so not reverting :)

PR #4133, opened 2023-08-22
cropperjs: ignore orientation data

andrew-nowak:
## What does this change?

in #4132 I attempted to set kahuna to ignore orientation data - unfortunately I missed that cropperjs also [tries to follow orientation data itself](https://github.com/fengyuanchen/cropperjs#checkorientation). Disable that functionality.

same as #4132, but remembering to check crop page!

same as #4132 - ignore orientation data until we support it in the backend

## Tested? Documented?
- [x] locally by committer


paperboyo (approved):
👍 Thanks again!

PR #4128, opened 2023-08-01
add a `restoreFromReplica` endpoint to `image-loader`

twrichards:
**REQUIRES https://github.com/guardian/grid/pull/4130**

https://trello.com/c/ffAUNID4/1621-lazarus-mechanism-to-restore-deleted-grid-images-from-replica-bucket

Related https://github.com/guardian/grid/pull/4125 (closed, since unlike this PR it didn't reinstate the image file in the image bucket, nor did it recreate thumbs or optimised image, which will have been deleted during reaping). 

Following on from the excellent https://github.com/guardian/grid/pull/4116. Adds a similar looking mechanism, but to reload the file from the 'replica bucket' (see https://github.com/guardian/editorial-tools-platform/pull/696 for more about replica buckets). It then layers on top user edits etc. from Dynamo (thanks to https://github.com/guardian/grid/pull/4130).

**The approach in this PR preserves the original `uploadTime`, `uploadedBy` and original filename.** 

This should help us restore images quickly which have been hard deleted by mistake (perhaps by reaper) - so long as we have the ID of course.

#### Requires new (optional) bit of image loader config `s3.image.replicaBucket`. Added in...

- [x] `TEST`
- [ ] `PROD`

PR #4134, opened 2023-08-31
[reaper] add `batchSoftDelete` and `batchHardDelete` endpoints (batches of up to 1000)

twrichards:
The 'reaper' has been turned off for many years and its implementation was never the most efficient. Thanks to the work in previous PRs (notably https://github.com/guardian/grid/pull/3926) we can easily filter for images which meet the criteria for 'reaping'. 

The new approach will **first 'soft delete' reapable images for two weeks,** so they're not visible in the grid UI (unless the `is:deleted` filter is applied, note all queries have `-is:deleted` by default) and so there's a place to view the candidates for hard deletion for a period of time before they're actually hard deleted.

This PR introduces, four new `media-api` endpoints...

- `/images/nextIdsToBeSoftReaped` and `/images/nextIdsToBeHardReaped` to retrieve next X ids of reapable images (to facilitate the two phases of reaping) where X is the 'size' query param (max. 1000)`
- `/images/batchSoftDelete` and `/images/batchHardDelete` which take a JSON array of IDs in the body, then...
   1. deletes from S3 (main image, thumbs and optimised PNG [if applicable]) in bulk using the [S3 bulk delete API](https://docs.aws.amazon.com/AmazonS3/latest/userguide/delete-multiple-objects.html) (max. 1000)
   2. processes just the ES mutations in `thrall` (via mixin of a new `BatchExternalThrallMessage` message type) making use of the 'update-by-query' and 'delete-by-query' ES operations respectively

... these will be primarily utilised by a re-write of the scheduled 'reaper' lambda (see https://github.com/guardian/grid/pull/4135).

_Note, thanks to #4128 we have a way to restore images even once they've been hard deleted 🪄_

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/ThrallMessage.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/ThrallMessage.scala
@@ -57,7 +57,6 @@ sealed trait ExternalThrallMessage extends ThrallMessage {
     )
   }
 }
-
 object ExternalThrallMessage{
   implicit val yourJodaDateReads = JodaReads.DefaultJodaDateTimeReads.map(d => d.withZone(DateTimeZone.UTC))
   implicit val yourJodaDateWrites = JodaWrites.JodaDateTimeWrites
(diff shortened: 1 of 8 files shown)


twrichards:
Closing in favour of a consolidated approach entirely within `thrall` - see https://github.com/guardian/grid/pull/4145

PR #4135, opened 2023-09-04
[reaper] re-write lambda

twrichards:
Re-write the existing lambda, to simply coordinate (and log) calls to the endpoints added in https://github.com/guardian/grid/pull/4134. Run on a schedule.


.github/workflows/ci.yml
```diff
@@ -34,11 +34,12 @@ jobs:
           yarn install
           yarn test
           yarn synth
-      #- name: Reaper
-      #  working-directory: ./reaper
-      #  run: |
-      #    npm install
-      #    npm run riffraff-artefact
+      - name: Reaper
```
alinaboghiu, line 37: 👏❤️👏❤️👏❤️

cdk/bin/constants.ts
```diff
@@ -0,0 +1,7 @@
+export const AWS_REGION = 'eu-west-1';
+export const STACK = 'media-service';
+export const REAPER_APP_NAME = 'grid-reaper';
+export const ENV_VAR_BATCH_SIZE = 'BATCH_SIZE';
```
alinaboghiu, line 4: this is just a tiny bit confusing because I am assuming the variable name is ENV_VAR_BATCH_SIZE and the value is a number (string but number) but it sounds like the variable name is "BATCH_SIZE"? I could be reading wrong. But maybe clearer if we just use the string in place?

twrichards: this is a variable representing the name of the environment variable, and it's in the `constants.ts` so it can be referred to when it's **set** in the CDK stack, and then when its **read** in the lambda code. Having a constant for this avoids there ever being a discrepancy when using that value.

alinaboghiu: ok, I think I have a mild preference for using the string in place but I will leave it up to you, it's really not important :)

reaper/src/index.ts
```diff
@@ -0,0 +1,103 @@
...
+
+		const hardDeletedStatuses = await fetch(
+			`${mediaApiImagesBaseUrl}/batchHardDelete`,
+			{
+				...fetchOptions,
+				method: 'DELETE',
+				body: JSON.stringify(IDsToHardDelete),
+			},
+		).then(parseResponseAs<HardDeletedStatuses>());
+		// FIXME check for false values in the response
```
alinaboghiu, line 91: I think we still want to do this, or else change the 'Hard delete succeeded.' message?

reaper/src/util.ts
```diff
@@ -0,0 +1,69 @@
...
+	credentials: IS_RUNNING_LOCALLY
+		? fromIni({ profile: LOCAL_PROFILE })
+		: fromNodeProviderChain(),
+};
+
+const ssm = new SSM(standardAwsConfig);
+
+const paramStorePromiseGetter =
+	(WithDecryption: boolean) => (nameSuffix: string) => {
+		const Name = `/${stage}/${STACK as string}/${
```
alinaboghiu, line 39:
```suggestion
		const name = `/${stage}/${STACK as string}/${
```

reaper/src/util.ts
```diff
@@ -0,0 +1,69 @@
...
+		}/${nameSuffix}`;
+		return ssm
+			.getParameter({
+				Name,
+				WithDecryption,
+			})
+			.then((result) => {
+				const value = result.Parameter?.Value;
+				if (!value) {
+					throw Error(`Could not retrieve parameter value for '${Name}'`);
```
alinaboghiu, line 50: is there a useful message in `result` we could log here?

alinaboghiu (approved):

reaper/src/util.ts
twrichards: this naming is so it can be used shorthand for the input to the AWS SSM operation (which annoyingly takes params with capitalised first letter)

twrichards:
Closing in favour of a consolidated approach entirely within `thrall` - see https://github.com/guardian/grid/pull/4145

PR #4137, opened 2023-09-08
Bump exiftool dependency

Georges-GNM:
## What does this change?

Small one to remove a high vulnerability. 

## How to test and measure success? ##

So interestingly, this is part of a particular script which, as far as I can tell, exists as a safety net if an image is missing its metadata, and would only get run manually by following the instructions in the [readme](https://github.com/guardian/grid/tree/main/scripts/inadvisable-metadata-backup). 

I tried running the tool with this command in my terminal: `AWS_PROFILE=$aws_profile ./run.sh $image_bucket $id`, filling in the variables with a random image from a test bucket. Running the command itself didn't completely succeed due to an issue with the aws-sdk version we're using (which it seems should be [bumped to v3](https://aws.amazon.com/blogs/developer/why-and-how-you-should-use-aws-sdk-for-javascript-v3-on-node-js-18/), but that feels just out of scope for this work of resolving a snyk vulnerability. Importantly, one of the commands that does seem to succeed in the script relates to the exiftool, which feels like sufficient proof that the dependency bump has worked. 

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/scripts/inadvisable-metadata-backup/package.json
+++ b/scripts/inadvisable-metadata-backup/package.json
@@ -6,7 +6,7 @@
   "dependencies": {
     "@types/node": "^14.14.37",
     "aws-sdk": "^2.876.0",
-    "exiftool-vendored": "^14.3.0",
+    "exiftool-vendored": "^22.2.0",
     "shell-escape": "^0.2.0",
     "ts-node": "^9.1.1",
     "typescript": "^4.2.3"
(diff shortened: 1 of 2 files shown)


twrichards (approved):
this tool possibly should be archived, but as you said in the PR description making it work (or indeed archiving) is beyond the scope of addressing the vuln dependency, so please go ahead and merge

PR #4138, opened 2023-09-08
Bump play and play-logback

Georges-GNM:
## What does this change?

Resolves a snyk vulnerability. 

Check the project still builds and runs normally (using sbt, running clean, compile and test should not bring up any errors). I've done this on code as well. 

As far as I could tell, the only code changes required were in `MigrationSourceWithSender`, which I understand would only get triggered if we ran [a migration](docs/05-migration/01-about.md), and I've not seen how this can really be manually tested otherwise. There is a `ThrallStreamProcessorTest` though, and that seems to be [passing correctly.](https://teamcity.gutools.co.uk/viewLog.html?buildId=876270&buildTypeId=EditorialTools_GridNew_UnifiedGridBuild&tab=buildLog#_state=44,5977:~:text=ThrallStreamProcessorTest) 

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -111,7 +111,7 @@ lazy val commonLib = project("common-lib").settings(
     "org.im4java" % "im4java" % "1.4.0",
     "com.gu" % "kinesis-logback-appender" % "1.4.2",
     "net.logstash.logback" % "logstash-logback-encoder" % "5.0",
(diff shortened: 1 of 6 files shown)


andrew-nowak:
to fully bump play you'll also need to update the version in project/plugins.sbt

twrichards (approved):
since you've tested on CODE already, this looks good to go - nice work!

PR #4144, opened 2023-09-20
move `SoftDeletedMetadataTable` to `common-lib`

twrichards:
https://trello.com/c/DrGAH8Y0/893-turn-on-the-reaper

Following on from https://github.com/guardian/grid/pull/4143 (which moved 'reapable' eligibility logic into common) this PR moves the `SoftDeletedMetadataTable` to common (so that we can use it from both `media-api` and `thrall` in [upcoming PR](https://github.com/guardian/grid/pull/4145)). It's in its own PR to make it easier to review.

Also removes usage of the global execution context, in favour of an implicit param for execution context.

## IMPORTANT
This involves moving the `dynamo.table.softDelete.metadata` property from `media-api.conf` up to `common.conf`. Since this is a lazy val and only referenced in media-api, AFAICT nothing should break... but since it's going to be relied upon in https://github.com/guardian/grid/pull/4145 very soon and does no harm in `common.conf` it ought to be moved before this PR is merged.

#### Guardian checklist

- [x] `TEST`
- [x] `PROD`

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/SoftDeletedMetadataTable.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/SoftDeletedMetadataTable.scala
@@ -1,13 +1,14 @@
-package lib
+package com.gu.mediaservice.lib.metadata
 
 import com.gu.mediaservice.lib.aws.DynamoDB
+import com.gu.mediaservice.lib.config.CommonConfig
 import com.gu.mediaservice.model.ImageStatusRecord
 import com.gu.scanamo._
 import com.gu.scanamo.syntax._
 
 import scala.concurrent.ExecutionContext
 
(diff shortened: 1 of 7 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/SoftDeletedMetadataTable.scala
```diff
@@ -1,26 +1,25 @@
...
+package com.gu.mediaservice.lib.metadata
 
 import com.gu.mediaservice.lib.aws.DynamoDB
+import com.gu.mediaservice.lib.config.CommonConfig
 import com.gu.mediaservice.model.ImageStatusRecord
-import com.gu.scanamo.error.DynamoReadError
 import com.gu.scanamo._
 import com.gu.scanamo.syntax._
 
-import scala.concurrent.ExecutionContext.Implicits.global
```
andrew-nowak, line 9: 🎉

andrew-nowak (approved):

PR #4145, opened 2023-09-20
rebuild `reaper` functionality in `thrall` AND remove old `reaper` lambda

twrichards:
https://trello.com/c/DrGAH8Y0/893-turn-on-the-reaper

Once upon a time, there was a process (in the form of a lambda) called 'the reaper' which deleted images (on a regular schedule) accordingly to a list of criteria, but was turned off out of caution after a significant chunk of images were permanently lost some years ago. This PR rebuilds 'the reaper', this time all within `thrall`.

#### Pre-requisite PRs:

- https://github.com/guardian/grid/pull/4143
- https://github.com/guardian/grid/pull/4144
- https://github.com/guardian/editorial-tools-platform/pull/706

### What's changed

- Delete the old 'reaper' lambda (and any traces from CI scripts etc.)
- Add new optional config property to `ThrallConfig` (**`s3.reaper.bucket` in `thrall.conf`**) to specify the bucket name where the permanent records of what was soft & hard deleted via the reaper will be stored (see https://github.com/guardian/editorial-tools-platform/pull/706 for Guardian) - **defining this property is required for the reaper to operate**
- Two new endpoints to `thrall` both taking `count` query param (for the batch size, max 1000) ...
  - `doBatchSoftReap` which  **'soft deletes'** (with deletedBy being `reaper`) the oldest batch of [`is:reapable` images](https://github.com/guardian/grid/pull/4143)  which are not already-soft deleted
  - `doBatchHardReap` which **'hard deletes'** the oldest batch of [`is:reapable` images](https://github.com/guardian/grid/pull/4143) which have been in 'soft deleted' state for at least two weeks
- The new `ReaperController` which defines contains the above endpoints also has a 'schedule' (every 15mins) which [IF the  `s3.reaper.bucket` config property is defined, otherwise doesn't run]...
  - queries the number of images uploaded in the last 7 days, then divides that to get the number of images ingested per 15mins
  - calls the `doBatchSoftReap` and `doBatchHardReap` with the count as number of images ingested per 15mins - this ensures we delete at same rate we ingest for a given environment (at the Guardian, our [`TEST` environment ingests roughly 1% of what `PROD` ingests](https://github.com/guardian/grid/pull/4033))
- we report the counts of images soft and hard reaped via new CloudWatch metrics
  image
- The new reaper process can be 'paused' by the presence of a file named `PAUSED` at the root of the new bucket `s3.reaper.bucket`. This is checked on each execution of the schedule, and exists early (with log message) if paused.
- Lastly `ReaperController` provides a couple more endpoints
   - `POST` endpoint for pausing (creating that `PAUSED` file as described above)
   - `POST` endpoint for resuming from paused state (deleting that `PAUSED` file as described above)
   - `GET` endpoint for reading a record file from the bucket
   - finally, an HTML view of 
      - whether Reaper is paused or not (with a button to flip that using the above) 
      - a list of the last day's worth of record files
   image

NOTE: we have the endpoints exposed (in addition to being called by the schedule) so that they can be manually called to for example clear a backlog if the reaper hasn't been running for whatever reason (either historically or because it was paused using the functionality above)


thrall/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -280,10 +288,75 @@ class ElasticSearch(
...
+      // unfortunately 'deleteByQuery' doesn't return the affected IDs so can't do this whole thing in one operation - https://github.com/elastic/elasticsearch/issues/45460
+      _ <- migrationAwareUpdater(
+        requestFromIndexName = indexName =>
+          deleteByQuery(
+            indexName,
+            idsQuery(ids),
+          ),
+        logMessageFromIndexName = indexName => s"ES7 hard delete $count images in $indexName"
+      )
+      // TODO check that the number of hard deleted images matches the number of ids
```
andrew-nowak, line 356: ditto

thrall/app/controllers/ReaperController.scala
```diff
@@ -0,0 +1,196 @@
...
+)(implicit val ec: ExecutionContext) extends BaseControllerWithLoginRedirects with GridLogging {
+
+  private val CONTROL_FILE_NAME = "PAUSED"
+
+  private val INTERVAL = 15 minutes // based on max of 1000 per reap, this interval will max out at 96,000 images per day
+  private val REAPS_PER_WEEK = 7.days / INTERVAL
+
+  implicit val logMarker: MarkerMap = MarkerMap()
+
+  private def getIsReapable = new ReapableEligibility {
```
andrew-nowak, line 43: I suppose it doesn't make a huge difference, but why `def` over `val` here?

thrall/app/controllers/ReaperController.scala
```diff
@@ -0,0 +1,196 @@
...
+
+  private def persistedBatchDeleteOperation(deleteType: String)(doBatchDelete: => Future[JsValue]) = config.maybeReaperBucket match {
+    case None => Future.failed(new Exception("Reaper bucket not configured"))
+    case Some(reaperBucket) => doBatchDelete.map { json =>
+      val now = DateTime.now(DateTimeZone.UTC)
+      val key = s"$deleteType/${now.toString("YYYY-MM-dd")}/$deleteType-${now.toString()}.json"
+      store.client.putObject(reaperBucket, key, json.toString())
+      json
+    }
+  }
```
andrew-nowak, line 101: I think you could merge these two wrappers? which should make the code easier to follow

thrall/app/controllers/ReaperController.scala
```diff
@@ -0,0 +1,196 @@
...
+
+  def doBatchSoftReap(count: Int): Action[AnyContent] = batchDeleteWrapper(count)(doBatchSoftReap)
+
+  def doBatchSoftReap(count: Int, deletedBy: String, isReapable: ReapableEligibility): Future[JsValue] = persistedBatchDeleteOperation("soft"){
+
```
andrew-nowak, line 106: optional for this PR, but I'd suggest building up a logMarker with the request ID (example https://github.com/guardian/grid/blob/main/cropper/app/controllers/CropperController.scala#L55) and passing it around and feeding it into the log statements; may make life easier tracking what happened in a given request. (in normal operation there won't be much ambiguity because calls are 15 mins apart, but it would allow you to quickly filter away any other thrall logs, or useful if we make a bunch of calls quickly to manually clear a backlog)

thrall/app/controllers/ReaperController.scala
twrichards: Because it need to be instantiated each time, since the 2 weeks old bit of it needs to be relative to the time it runs, each time

thrall/app/controllers/ReaperController.scala
twrichards: They're separate as one is for http endpoints, the other is for both http and when it's run be the schedule

thrall/app/controllers/ReaperController.scala
davidfurey: Where is the 2 weeks bit? 

Also, can we make it more clear to future developers why this needs to be a `def`?

thrall/app/controllers/ReaperController.scala
```diff
@@ -0,0 +1,195 @@
...
+          isDeleted = true
+        )
+      ))
+    } yield idsSoftDeletedInES.map { id =>
+      val detail = Map(
+        "ES" -> true,
+        "dynamo.softDelete.metadata" -> !idsNotProcessedInDynamo.contains(id)
+      )
+      logger.info(s"Soft deleted image $id : $detail")
+      metrics.softReaped.increment().run
```
andrew-nowak, line 126: it looks like you can pass a number to increment, so you should be able to pass the number reaped in a batch in a single datapoint, which should be cheaper and faster (PutMetricData calls are $0.01 per 1000, so (for both soft & hard) $20 per million... not really expensive but not cheap either)

common-lib/src/main/scala/com/gu/mediaservice/lib/ImageIngestOperations.scala
```diff
@@ -42,9 +46,29 @@ class ImageIngestOperations(imageBucket: String, thumbnailBucket: String, config
...
+      )
+      keys.map { key =>
+        key -> true
+      }.toMap
+    } catch {
+      case partialFailure: MultiObjectDeleteException =>
+        // FIXME log partial failure
+        keys.map { key =>
+          key -> partialFailure.getErrors.asScala.map(_.getKey).toList.contains(key)
+        }.toMap
```
andrew-nowak, line 62: This looks like it returns key -> true for each key which appears in getErrors (so failed to delete) - but the success branch above returns key -> true for each key which is successfully deleted
```suggestion
        val errorKeys = partialFailure.getErrors.asScala.map(_.getKey).toSet
        keys.map { key =>
          key -> !errorKeys.contains(key)
        }.toMap
```

thrall/app/controllers/ReaperController.scala
twrichards: nice - thanks

thrall/app/controllers/ReaperController.scala
```diff
@@ -0,0 +1,196 @@
...
+
+  config.maybeReaperBucket.foreach { reaperBucket =>
+    scheduler.scheduleAtFixedRate(
+      initialDelay = 0 seconds,
+      interval = INTERVAL,
+    ){ () =>
+      if(store.client.doesObjectExist(reaperBucket, CONTROL_FILE_NAME)) {
+        logger.info("Reaper is paused")
+      } else {
+        es.countImagesIngestedInLast(7 days)(DateTime.now(DateTimeZone.UTC)).flatMap { imagesIngestedInLast7Days =>
```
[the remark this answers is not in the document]
twrichards: done - thanks

andrew-nowak (approved):
there's 2 open TODOs, and a suggestion for improved logging, but otherwise all looks good to me - amazing work 👏 👏 👏

PR #4146, opened 2023-09-29
use 'GitHub Actions' for CI (exclusively)

twrichards:
This migrates (from TeamCity) and consolidates all CI into a single GHA workflow called `CI`.

### Detail

- combines the `ScalaBuild` and `JSBuild` jobs of the GHA into the single job, as in order to package kahuna client-side needs to be built first, so we can't benefit from the parallelisation of jobs (but this was negligble)
- replaces the sbt and node riff-raff plugins, with the best-practice https://github.com/guardian/actions-riff-raff 
- ~removes the `push-pr.sh` script for building PRs from forks, since we can now run the GHA~ 
_alas this wasn't possible, see https://github.com/guardian/grid/pull/4146#discussion_r1346416478_
- consolidates `image-counter-lambda` into the main deployment, leaving a single `riff-raff.yaml`
- removes the `scripts/ci` directory since the steps are now all in the `ci.yaml` GHA workflow file, which provides better visibility in the GHA UI (whilst obviously still being version controlled)
- consolidates the separate 'sbt' action into this CI workflow to avoid complexity/indirection


.github/workflows/ci.yml
```diff
@@ -57,12 +31,75 @@ jobs:
...
+          npm run dist
+      - name: SBT
+        uses: ./.github/actions/sbt
+        env:
+          USE_DOCKER_FOR_TESTS: false
+          ES6_TEST_URL: http://elasticsearch:9200
+          LOCALSTACK_ENDPOINT: http://localstack:4566
+        with:
+          args: clean compile test:compile test debian:packageBin
+      - uses: aws-actions/configure-aws-credentials@v1-node16
```
samanthagottlieb, line 54: There's a later version of this action, [v4](https://github.com/aws-actions/configure-aws-credentials/releases), which I'd hope would be compatible with Node 16.

.github/workflows/ci.yml
```diff
@@ -57,12 +31,75 @@ jobs:
...
-    - uses: actions/checkout@v3.5.0
-    - name: SBT
-      uses: ./.github/actions/sbt
-      env:
-        USE_DOCKER_FOR_TESTS: false
-        ES6_TEST_URL: http://elasticsearch:9200
-        LOCALSTACK_ENDPOINT: http://localstack:4566
-      with:
-        args: clean compile test:compile test
+      - uses: actions/checkout@v3
```
samanthagottlieb, line 34: There's also a later version of this - [v4](https://github.com/actions/checkout/releases), unless there's a specific reason to be using v3?

.github/workflows/ci.yml
```diff
@@ -57,12 +31,75 @@ jobs:
...
+          npm test
+          npm run dist
+      - name: SBT
+        uses: ./.github/actions/sbt
+        env:
+          USE_DOCKER_FOR_TESTS: false
+          ES6_TEST_URL: http://elasticsearch:9200
+          LOCALSTACK_ENDPOINT: http://localstack:4566
+        with:
+          args: clean compile test:compile test debian:packageBin
```
samanthagottlieb, line 53: Would it help to use [semicolons](https://www.scala-sbt.org/1.x/docs/Howto-Running-Commands.html#Provide+multiple+commands+to+run+consecutively) here to ensure the commands execute sequentially? And/or to use [batch mode](https://www.scala-sbt.org/1.x/docs/offline/Running.html#Batch+mode) to ensure all the commands are treated as a single command? As done [here](https://github.com/guardian/editorial-preferences/blob/main/.github/workflows/ci.yml#L35). Not sure if this is necessary/beneficial when using the `sbt` action.

.github/workflows/ci.yml
twrichards: done, thanks for the heads up

.github/workflows/ci.yml
twrichards: Thanks for those links @samanthagottlieb - TIL what they actually mean, as was only vaguely aware before.

With the commands above, I believe they're already in the order matching the sbt 'lifecycle' so I'm fairly sure they execute sequentially, and personally at least I think they read better without the semicolons, and since there's no spaces in each command I don't think batch mode is needed either - will keep in mind for other things though - thank you

davidfurey: I suspect the `;` syntax might be in use elsewhere because the newer syntax for packageBin has a space:
```
[warn] sbt 0.13 shell syntax is deprecated; use slash syntax instead: Debian / packageBin
```

twrichards: I can't seem to get it forward the args via the sbt action (essentially a docker file to provide `graphicsmagick` etc.) when using quotes and semicolons, so I'm going to leave it as is, since it works and is readable (albeit with the warnings you highlighted @davidfurey ).

andrew-nowak: spaces aren't needed for the new-style sbt commands, this is how you'd run newstyle args on the command line (and here)
 
```suggestion
          args: clean compile test:compile test Debian/packageBin
```

batch mode is the name given to running sbt without entering the sbt shell - as these arguments get passed to sbt command line in the custom dockerfile, it looks like this will get executed in batch mode which will run arguments in order. you'd only need semicolons when running inside the sbt shell, and you only need quotes on the command line if you need to pass arguments to a specific command (eg. testOnly)

andrew-nowak: we probably can remove that custom sbt action? the default github image has sbt; the useful bit of that dockerfile is apt-get installing graphicsmagick and exiftool, which we could also do in an earlier step https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/customizing-github-hosted-runners

samanthagottlieb: Thanks all for the clarification on semicolons and batch mode!

twrichards: > we probably can remove that custom sbt action? 

thanks for this suggestion @andrew-nowak - I've gone ahead and consolidated that into the main CI job and its simpler now I feel https://github.com/guardian/grid/pull/4146/commits/7b32ea23478b732b846e393e592c6e528dadabef

.github/workflows/ci.yml
```diff
@@ -57,12 +33,82 @@ jobs:
...
+        run: |
+          sudo apt-get -y update
+          sudo apt-get -y install graphicsmagick
+          sudo apt-get -y install graphicsmagick-imagemagick-compat
+          sudo apt-get -y install exiftool
+      - name: Setup Java
+        uses: actions/setup-java@v3
+        with:
+          java-version: '8'
+          distribution: 'corretto'
```
andrew-nowak, line 58:
```suggestion
          distribution: 'corretto'
          cache: 'sbt'
```
to make sure GHA caches downloaded libs (likely to be important for performance)

andrew-nowak (approved):
tested on test, all services come up - lgtm!

.github/workflows/ci.yml
```diff
@@ -1,46 +1,22 @@
 name: Grid CI
-on: [pull_request]
+on:
+  workflow_dispatch: # Manual invocation.
```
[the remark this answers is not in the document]
twrichards: You're right actually, I'd need to add the `pull_request` trigger and then I was going to change the settings to require approval for external contributors (see https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#controlling-changes-from-forks-to-workflows-in-public-repositories and https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks) but you're also right that it won't have access to the secrets it needs which is a shame, so regrettably will need to re-instate the `push-pr.sh` script 😞

PR #4147, opened 2023-10-02
add to nginx trouble-shooting doc

dblatcher:
## What does this change?

Adds a section to the nginx trouble-shooting doc describing a nginx configuration problem I had running locally for the first time.

If I understand it right, this isn't an issue that is specific to this project, but since it involves configuring several subdomains to run locally, it seems not unlikely that setting it up will suface the problem.

n/a - docs only.

Hopefully useful to any future devs getting the same problem.

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [x] locally by committer


andrew-nowak:
this does also exist [here](https://github.com/guardian/grid/blob/main/docs/01-setup/03-running-setup.md?plain=1#L22); but more discoverability is always better, maybe worth making one the source of truth and the other linking back to it?

PR #4148, opened 2023-10-05
run CI GHA on `pull_request` so that external contributors benefit from CI

twrichards:
Following/tweaking #4146 - we do need to trigger the unified `CI` GHA based on `pull_request`* and not just 'all branches' since this doesn't include branches on forks (understandably) and I imagine our external contributors benefit from the CI checks (compilation, tests etc.) on their PRs. 

To complement the above, we limit `push` based trigger to `main` branch only to avoid duplicate workflow runs (wasteful) as well as branches starting with pr (i.e. those created by `dev/script/push-pr.sh`).

\* this does have the unfortunate downside of building the hidden `pull/XXX/merge` branch, which is actually a preview of what the merge would look like rather than the branch the PR has been raised from - whilst isn't normally an issue but can on occasion cause pernicious problems, so would normally avoid it - but alas its necessary here. You can see an example from this PR's workflow run...
image


Georges-GNM (approved):
Looks good to me!

PR #4151, opened 2023-10-06
[reaper] use bulk API for soft/hard deleting batches from ES, so we don't need follow-up query to confirm success in ES

twrichards:
Small but important follow-up to https://github.com/guardian/grid/pull/4145 - where it looks like the `delete_by_query` was being processed asynchronously and so the guard added in https://github.com/guardian/grid/pull/4145/commits/d3c8b0a9edf5a87a3cce807eaeded5487b633e61 was firing, even though the ES record was disappearing - this in turn meant the S3 files will have been orphaned off.

In this PR we...
> -  ~explicitly 'wait for completion' of the update_by_query and delete_by_query~
> -  ~explicitly verify the IDs actually updated / deleted with subsequent query~
> -  ~don't throw, but log any discrepancy, since we now continue to process the ones which were successfully updated / deleted~

... use the bulk API instead because it waits on completion of the actions automatically and returns the success/error of each - meaning we don't need a follow-up query to verify.

### DIFF
--- a/thrall/app/controllers/ReaperController.scala
+++ b/thrall/app/controllers/ReaperController.scala
@@ -119,9 +119,6 @@ class ReaperController(
         )
       ))
     } yield {
-      if(esIds.isEmpty){
-        logger.info(s"Although $count images were requested to be soft deleted, none were found to be soft deletable.")
-      }
       metrics.softReaped.increment(n = esIdsActuallySoftDeleted.size).run
       esIds.map { id =>
         val wasSoftDeletedInES = esIdsActuallySoftDeleted.contains(id)
@@ -150,9 +147,6 @@ class ReaperController(
       pngsS3Deletions <- store.deletePNGs(esIdsActuallyDeleted)
       idsNotProcessedInDynamo <- softDeletedMetadataTable.clearStatuses(esIdsActuallyDeleted)
     } yield {
-      if (esIds.isEmpty) {
-        logger.info(s"Although $count images were requested to be hard deleted, none were found to be hard deletable.")
-      }
       metrics.hardReaped.increment(n = esIdsActuallyDeleted.size).run
       esIds.map { id =>
         val wasHardDeletedFromES = esIdsActuallyDeleted.contains(id)
--- a/thrall/app/lib/elasticsearch/ElasticSearch.scala
+++ b/thrall/app/lib/elasticsearch/ElasticSearch.scala
@@ -330,6 +330,9 @@ class ElasticSearch(
         logMessageFromIndexName = indexName => s"ES7 soft delete ${ids.size} images in $indexName by ${softDeletedMetadata.deletedBy}"
       ).map(_.result.items)
     } yield {
+      if (ids.isEmpty) {
+        logger.info(s"Although $count images were requested to be soft deleted, none were found to be soft deletable.")
+      }
       esResults.filter(_.error.isDefined).foreach(item =>
         logger.error(logMarker, s"ES7 failed to soft delete image ${item.id} : ${item.error.get}")
       )
@@ -357,6 +360,9 @@ class ElasticSearch(
         logMessageFromIndexName = indexName => s"ES7 hard delete ${ids.size} images in $indexName"
       ).map(_.result.items)
     } yield {
+      if (ids.isEmpty) {
+        logger.info(s"Although $count images were requested to be hard deleted, none were found to be hard deletable.")
+      }
       esResults.filter(_.error.isDefined).foreach(item =>
         logger.error(logMarker, s"ES7 failed to hard delete image ${item.id} : ${item.error.get}")
       )


dblatcher (approved):
Not too familiar with the project yet, but LGTM

andrew-nowak:
well the ES docs are super unclear on the point but I'm 99.9% certain the default is already wait_for_completion=true - try on TEST cerebro on the empty index with a match all query and no update step, and try with and without `?wait_for_completion=false` (and `true`) - you have a completely different response body if wait_for_completion is false (just a task id which you need to feed into the tasks api)

twrichards:
> well the ES docs are super unclear on the point but I'm 99.9% certain the default is already wait_for_completion=true - try on TEST cerebro on the empty index with a match all query and no update step, and try with and without `?wait_for_completion=false` (and `true`) - you have a completely different response body if wait_for_completion is false (just a task id which you need to feed into the tasks api) so the response here would fail to parse [`c4b4ebe`/thrall/app/lib/elasticsearch/ElasticSearch.scala#L331C13-L331C29](https://github.com/guardian/grid/blob/c4b4ebe0aec38419035a6402b2221298cf162e73/thrall/app/lib/elasticsearch/ElasticSearch.scala#L331C13-L331C29)

it might be the default for the update_by_query, but I added it to be explicit that it should be a synchronous call. Curiously the return type of delete_by_query in elastic4s is an `Either[DeleteByQueryResponse, CreateTaskResponse]` and mapping the left projection suggests that its `CreateTaskResponse` so I'm hoping this PR will ensure its a Left and not a Right.

twrichards:
> > well the ES docs are super unclear on the point but I'm 99.9% certain the default is already wait_for_completion=true - try on TEST cerebro on the empty index with a match all query and no update step, and try with and without `?wait_for_completion=false` (and `true`) - you have a completely different response body if wait_for_completion is false (just a task id which you need to feed into the tasks api) so the response here would fail to parse [`c4b4ebe`/thrall/app/lib/elasticsearch/ElasticSearch.scala#L331C13-L331C29](https://github.com/guardian/grid/blob/c4b4ebe0aec38419035a6402b2221298cf162e73/thrall/app/lib/elasticsearch/ElasticSearch.scala#L331C13-L331C29)
> 
> it might be the default for the update_by_query, but I added it to be explicit that it should be a synchronous call and the soft delete endpoint is working fine, consider also that the return type is just `UpdateByQueryResponse` in elastic4s. Curiously the return type of delete_by_query in elastic4s is an `Either[DeleteByQueryResponse, CreateTaskResponse]` and the previous implementation which maps left projection suggests that its `CreateTaskResponse` so I'm hoping this PR will ensure its a Left (i.e. DeleteByQueryResponse) and not a Right.

actually the type situation is baffling 
image

twrichards:
tested on `TEST` ✅

thrall/app/controllers/ReaperController.scala
```diff
@@ -119,14 +119,15 @@ class ReaperController(
         )
       ))
     } yield {
-      if(idsSoftDeletedInES.isEmpty){
+      if(esIds.isEmpty){
```
davidfurey, line 122: Not a big thing, but I was a bit surprised that we log here if no images were found to be deleted, but we don't log if images were found but failed to be deleted. Until I got to `softDeleteNextBatchOfImages` and found that it logs failures. I wonder whether these two bits of logging should be closer together, probably in `softDeleteNextBatchOfImages`?

davidfurey (commented):
Looks fine to me, but would like to understand why the key name changed before approving

thrall/app/controllers/ReaperController.scala
```diff
@@ -143,23 +144,24 @@ class ReaperController(
...
+      esIds.map { id =>
+        val wasHardDeletedFromES = esIdsActuallyDeleted.contains(id)
         val detail = Map(
-          "ES" -> Some(true), // since this is list of IDs deleted from ES
+          "ES" -> Some(wasHardDeletedFromES),
           "mainImage" -> mainImagesS3Deletions.get(ImageIngestOperations.fileKeyFromId(id)),
           "thumb" -> thumbsS3Deletions.get(ImageIngestOperations.fileKeyFromId(id)),
           "optimisedPng" -> pngsS3Deletions.get(ImageIngestOperations.optimisedPngKeyFromId(id)),
-          "dynamo" -> Some(!idsNotProcessedInDynamo.contains(id)),
+          "dynamo.table.softDelete.metadata" -> (if(wasHardDeletedFromES) Some(!idsNotProcessedInDynamo.contains(id)) else None)
```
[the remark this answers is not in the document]
twrichards: oh just to make it clearer that all we're deleting is the record from the soft delete metadata table, rather than deleting all traces of the image from Dynamo (which we don't do anymore since https://github.com/guardian/grid/pull/4130, which in part is to facilitate [restoring from replica](https://github.com/guardian/grid/pull/4128)).

We do need to delete the soft deleted records though otherwise that table will get silly big over time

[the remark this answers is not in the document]
davidfurey: Sorry, completely missed that this is just for logging / json response

davidfurey (approved):

thrall/app/controllers/ReaperController.scala
twrichards: done - see https://github.com/guardian/grid/pull/4151/commits/9bf496b29e33779b3f61b8d2e663c9e617809ccd

davidfurey: Thanks!

PR #4162, opened 2023-10-18
[reaper] fix display of 'recent records' and display last 48 hours of records instead

twrichards:
https://trello.com/c/AxiYosIs/1791-fix-grid-reaper-recent-records-display

Noticed when unpausing the reaper (after https://trello.com/c/oNxloIHp/1763-commence-soft-clear-down-of-reaper-backlog-65m-images finished) that the recent records list was not displaying the most recent records, but instead an arbitrary set (because we now have over 1000 reaper record files).

### Tested in `TEST` ✅ 
image

### DIFF
--- a/thrall/app/controllers/ReaperController.scala
+++ b/thrall/app/controllers/ReaperController.scala
@@ -90,11 +90,13 @@ class ReaperController(
     }
   }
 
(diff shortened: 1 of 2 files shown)


andrew-nowak (approved):

dblatcher (approved):
LGTM

PR #4168, opened 2023-10-19
[reaper] count all soft and hard reapable images and publish metrics…

twrichards:
…so we can visualise how the rate we're reaping compares.

https://github.com/guardian/grid/pull/4145 introduced the `HardReaped` and `SoftReaped` metrics which show the amount of images hard and soft reaped respectively - which have been very useful. The reaping rate is proportional to the 7day rolling average of ingestion, so these two new metrics...
- `SoftReapable`
- `HardReapable` 

... will be useful to know whether the reaping rate is sufficient... or if it needs a rethink.


dblatcher (approved):

twrichards:
TESTED on `TEST` ✅

PR #4169, opened 2023-10-23
[reaper] reap at configured rate (rather than dynamic based on ingestion rate)

twrichards:
Thanks to #4168, we can see that although we do catch-up slightly each night, the amount of reapable images is slowing getting away from us...
image

I had become a little fixated on avoiding config and tried to be too clever (trying to base the reaping rate on the 7day rolling average of ingestion in #4145) which isn't keeping up and is harder to reason about. 

Should've heeded the suggestion from prophet @andrew-nowak that reaping rate should be fixed/configured.

... this PR does exactly that; **[scheduled] reaping is now based on `reaper.countPerRun` config property**.

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
@@ -223,8 +223,7 @@ abstract class CommonConfig(resources: GridConfigResources) extends AwsClientBui
 
   final def stringOpt(key: String): Option[String] = configuration.getOptional[String](key)
 
-  final def int(key: String): Int =
-    configuration.getOptional[Int](key) getOrElse missing(key, "integer")
+  final def intOpt(key: String): Option[Int] =  configuration.getOptional[Int](key)
 
   final def intDefault(key: String, default: Int): Int =
     configuration.getOptional[Int](key) getOrElse default
--- a/thrall/app/controllers/ReaperController.scala
+++ b/thrall/app/controllers/ReaperController.scala
@@ -36,8 +36,6 @@ class ReaperController(
   private val CONTROL_FILE_NAME = "PAUSED"
 
   private val INTERVAL = 15 minutes // based on max of 1000 per reap, this interval will max out at 96,000 images per day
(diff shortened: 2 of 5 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
```diff
@@ -223,8 +223,7 @@ abstract class CommonConfig(resources: GridConfigResources) extends AwsClientBui
 
   final def stringOpt(key: String): Option[String] = configuration.getOptional[String](key)
 
-  final def int(key: String): Int =
-    configuration.getOptional[Int](key) getOrElse missing(key, "integer")
```
twrichards, line 227: not this was unused, so re-purposed

thrall/app/controllers/ReaperController.scala
```diff
@@ -46,35 +44,25 @@ class ReaperController(
...
           val deletedBy = "reaper"
-
           Future.sequence(Seq(
             doBatchSoftReap(countOfImagesToReap, deletedBy),
             doBatchHardReap(countOfImagesToReap, deletedBy)
           ))
         }
       }
-    }
+    case _ => ()
```
andrew-nowak, line 65: worth logging that reaper won't run due to missing config?

andrew-nowak (approved):

thrall/app/controllers/ReaperController.scala
twrichards: good suggestion - done

twrichards:
Just a lil update; you can see this has been very effective... 
image
You can see in PROD...
- the red `SoftReapable` line begin, when #https://github.com/guardian/grid/pull/4168 was released (which shows it generally trending up, despite fluctuations)
- the point this PR was released, the `SoftReaped` count jump up to 1000 per 15mins (up from the dynamic ~510)
- the total `SoftReapable` count come down more aggressively since this PR
- until it reaches sub-1000, then the `SoftReaped` count becomes more scattered as it soft deletes only the ones available 🎉 

The hard reaping metrics will kick in later this week once the big backlog of millions of images reaches two weeks in soft deleted state.

PR #4173, opened 2023-10-26
remove google analytics code from Kahuna

dblatcher:
## What does this change?

Removes the code for Google Analytics:
- googleTrackingId property from config
- the GA script tags in main.scala.html
- the analytics module in the angular app

The account it used was turned off in July and the data wasn't being used before then,  see https://trello.com/c/MvsOLUwA/1435-remove-ga-from-grid.

Runs locally without any noticable change (there was no googleTrackingId in the DEV config, so the GA scripts were not rendered, but the the line in the config script (visible in view-source:https://media.local.dev-gutools.co.uk/) will be gone.

When deployed to CODE, the should work as before, but with no googleTrackingId loaded into config and no https://www.google-analytics.com/analytics.js script

Still works as before, less redundant code.

## Tested? Documented?
- [x] locally by committer


twrichards (approved):
LGTM
I've added an extra commit to remove GA from the `Content-Security-Policy`, will test on `TEST` and have also check this is fine with the beeb

PR #4174, opened 2023-10-30
blur 'potentially graphic' images

twrichards:
With so many graphic images coming through as a result of the war, understandably there have been numerous requests for some mechanism to limit the effect on staff.

Preference is for blurring rather than filtering to avoid the chance of pictures being missed. Identifying graphic images is very challenging given the inconsistent ways the agencies flag such images... so this PR is to deliver the basic functionality sooner rather than later (to benefit Grid users), but we expect to further refine the criteria/heuristics for flagging such images.

### Changes
Initially the 'potentially graphic image' detection was implemented via an [ES Script Field](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/search-fields.html#script-fields) (as this has access to the entire ES document, including all the original metadata where we hoping agencies would flag such images in a structured/consistent manner - alas, they do not). To aid with rapid iteration of the detection criteria/heuristics/logic the ability to control the ['painless'](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-scripting-painless.html) script was essential. There was a default, which could be overridden from the client (no build/deploy delay) via the value of a new cookie `IS_POTENTIALLY_GRAPHIC_SCRIPT` (on the parent domain, so `media-api` can see it).

We also chose to have this feature 'opt-in' (at least for now), which can be done via a checkbox in the menu in the top right...
image
... which sets a new `SHOULD_BLUR_GRAPHIC_IMAGES` cookie (on the parent domain, so `media-api` can see it).

After lots of experimentation with @paperboyo, we concluded we could actually identify the vast majority of such images by **searching for a list of strings (words/phrases) within the `description`, `title`, `special instructions` and `keywords`** - all of which is available in the `media-api` search response data - which meant this could be done purely client-side (thanks to @andrew-nowak for this nudge, after we struggled to get the OR logic working in the painless script). So I've removed the ability to edit the detection script from the client-side (via the `IS_POTENTIALLY_GRAPHIC_SCRIPT` cookie) but have added logic to the server-side to detect `pur:adultContentWarning` XMP field as we've observed this be used.

The actual blurring is achieved via simple CSS filter with some hover behaviour to reveal and a pseudo `after` element to add the explanation text...
graphic_blur_off_by_default

_**In a follow-up PR, I'll look to make the list of strings configurable.**_


kahuna/public/js/common/user-actions.html
```diff
@@ -16,6 +20,14 @@
                 </a>
             </li>
             <li><a class="drop-menu__items__feedback" ng-href="{{ctrl.feedbackFormLink}}" rel="noopener" target="_blank">Feedback</a></li>
+            <li>
+                <div>
+                    <label>Blur &apos;Potentially Graphic&apos; images <input type="checkbox" ng-model="ctrl.shouldBlurGraphicImages" ng-change="ctrl.syncShouldBlurGraphicImages()"></label>
+                </div>
+                <a ng-if="ctrl.hasCustomPotentiallyGraphicScript || isShiftKeyDown" ng-click="ctrl.editPotentiallyGraphicScript()">
+                    Adjust extra &apos;Potentially Graphic&apos; script
```
paperboyo, line 28:
```suggestion
                    <label>Blur &lsquo;Potentially Graphic&rsquo; images <input type="checkbox" ng-model="ctrl.shouldBlurGraphicImages" ng-change="ctrl.syncShouldBlurGraphicImages()"></label>
                </div>
                <a ng-if="ctrl.hasCustomPotentiallyGraphicScript || isShiftKeyDown" ng-click="ctrl.editPotentiallyGraphicScript()">
                    Adjust extra &lsquo;Potentially Graphic&rsquo; script
```

kahuna/public/js/preview/image.js
```diff
@@ -82,6 +82,25 @@ image.controller('uiPreviewImageCtrl', [
...
+        'depicts death',
+        'dead child',
+        'child casualty',
+        'sensitive material',
+        'injured',
+        'wounded',
+        'killed',
+        'body of',
+        'bodies of',
+        'smout'
```
Georges-GNM, line 96: Is "smout" another word for child? Not sure that needs to be in the list? (maybe you were using that as a test though?

Georges-GNM (approved):
Looks good to me as a first iteration - this seems to be working as intended on test, and the added logic makes sense to me. 

The actual blurring effect looks great!

Georges-GNM (approved):
Looks good to me as a first iteration - this seems to be working as intended on test, and the added logic makes sense to me. 

The actual blurring effect looks great!

PR #4177, opened 2023-10-31
[graphic content] tweak criteria for blurring

twrichards:
tweak following #4174 based on user feedback


kahuna/public/js/preview/image.js
```diff
@@ -88,12 +88,11 @@ image.controller('uiPreviewImageCtrl', [
         'dead child',
         'child casualty',
         'sensitive material',
-        'injured',
-        'wounded',
-        'killed',
```
twrichards, line 93: these are too broad, and we don't want users to miss images and stop using the feature

paperboyo (approved):
Thanks!

PR #4179, opened 2023-11-01
[graphic content] log `user` and `shouldFlagGraphicImages` as markers for every image search

twrichards:
… so we can pull stats on uptake of blurring functionality

You can see these being surfaced in logs like so...
image
image


Georges-GNM (approved):
Looks good to me and can see the dashboard reflecting my test search just now 👍

PR #4181, opened 2023-11-03
fix regression bug when adding labels/people to collection of images

AndyKilmory:
fix regression bug when adding labels/people to collection of images via info panel

## What does this change?

Adds null check for variable added as part of work on upload page but not reuired or used here

Ensure that adding labels to multiple images works from info panel when label exists on 1 image but not the others

## Tested? Documented?
- [ ] locally by committer


twrichards (approved):
nice, thank you. @paperboyo is going to test but change LGTM

PR #4182, opened 2023-11-06
fix the zipping of `image-counter-lambda` as was failing to execute

twrichards:
…since 5th October 🤦‍️ (with `Cannot find module './sourcemap-register.js'`)

Made quite a boo boo in https://github.com/guardian/grid/pull/4146 where I was only ZIPing just `index.js` (rather than the whole dist dir as before by https://github.com/guardian/node-riffraff-artifact) - this PR puts that right.

Annoyingly this was missed as major reaping was going on at the same time, and I confused the alarms going off about image count (as they will have been firing from 'no data' rather than in breach of their threshold).

### DIFF
--- a/image-counter-lambda/package.json
+++ b/image-counter-lambda/package.json
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "test": "jest",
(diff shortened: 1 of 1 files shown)


akash1810 (approved):

twrichards:
image fix worked in `TEST` and I can also see the metrics have resumed in CloudWatch

PR #4185, opened 2023-11-06
add `uploadedInLastFiveMinutes` to the imageCounts 

twrichards:
Now that the reaper is back in action (see https://github.com/guardian/grid/pull/4145) we can no longer rely on the total count in ES for the alarms which hang-off the metrics produced by `image-counter-lambda` (see https://github.com/guardian/editorial-tools-platform/blob/main/cloudformation/media-service-account/image-counter-lambda.yaml). These alarms exist for two reasons... 

1. to alarm when we have an ingestion issue (i.e. total count wasn't rising)
2. to alarm when we lose lots of images (as has happened in the past 😬)

**This PR facilitates a new alarm to capture reason 1.**

_2 is much harder since we're reaping at a variable rate (effectively based on the ingestion rate from the moment 20 days prior)._


phillipbarron (approved):
LGTM 👍

PR #4187, opened 2023-11-07
configurable default on/off for 'graphic image' blurring (with tooltip/explainer which must be acknowledged)

twrichards:
Although we've proved out the value of #4174 (opt-in), usage remains fairly low (around 5%) - so this PR makes the default configurable.

If configured (`defaultShouldBlurGraphicImages = true` in `common.conf`), for users who've not yet used the feature (i.e. no `SHOULD_BLUR_GRAPHIC_IMAGES` cookie) then we now present a modal/overlay explaining the feature and the fact its on by default - which must be acknowledged (sets the cookie value to `"true"`) OR can be turned off via the tickbox (sets the cookie value to `"false"`)...

image

Also includes refactor of all 'graphic image' blurring logic into its own Angular 'service'.

### DIFF
--- a/media-api/app/controllers/MediaApi.scala
+++ b/media-api/app/controllers/MediaApi.scala
@@ -413,7 +413,8 @@ class MediaApi(
   def imageSearch() = auth.async { request =>
     implicit val r = request
 
-    val shouldFlagGraphicImages = request.cookies.get("SHOULD_BLUR_GRAPHIC_IMAGES").exists(_.value == "true")
+    // defaults to true if not specified
+    val shouldFlagGraphicImages = request.cookies.get("SHOULD_BLUR_GRAPHIC_IMAGES").map(_.value).getOrElse("true") == "true"
 
     implicit val logMarker = MarkerMap(
       "shouldFlagGraphicImages" -> shouldFlagGraphicImages,
--- a/media-api/app/controllers/MediaApi.scala
+++ b/media-api/app/controllers/MediaApi.scala
@@ -413,8 +413,8 @@ class MediaApi(
   def imageSearch() = auth.async { request =>
     implicit val r = request
 
-    // defaults to true if not specified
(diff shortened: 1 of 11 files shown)


bryophyta:
I haven't run the Grid locally before so not been able to test this yet, but I noticed that the checkbox is outside the modal in the screenshot, so just wanted to check whether this feature has been tested for keyboard-only navigation?

twrichards:
> This looks like a great feature!
> 
> I haven't run the Grid locally before so not been able to test this yet, but I noticed that the checkbox is outside the modal in the screenshot, so just wanted to check whether this feature has been tested for keyboard-only navigation?

Great point @bryophyta - it is indeed keyboard navigable, but your prompt made me check and the keyboard navigation still goes to the inputs beneath the modal/overlay first - I'll see if I can sort that out now

bryophyta:
> Great point @bryophyta - it is indeed keyboard navigable, but your prompt made me check and the keyboard navigation still goes to the inputs beneath the modal/overlay first - I'll see if I can sort that out now

Thanks for taking a look! Yeah, modals can v fiddly; hope you can find a solution without too much hassle!

media-api/app/controllers/MediaApi.scala
```diff
@@ -413,7 +413,9 @@ class MediaApi(
   def imageSearch() = auth.async { request =>
     implicit val r = request
 
-    val shouldFlagGraphicImages = request.cookies.get("SHOULD_BLUR_GRAPHIC_IMAGES").exists(_.value == "true")
+    // defaults to true if not specified
```
davidfurey, line 416: Does it not default to `defaultShouldBlurGraphicImages` if not specified?

twrichards: got spot, thanks. this was an old comment from before it was made configurable - I've removed the comment now as the code is self-explanatory with `.getOrElse(config.defaultShouldBlurGraphicImages`

davidfurey (approved):

PR #4188, opened 2023-11-08
FIX `acceptDefaultOfBlurringGraphicImages`

twrichards:
tiny fix to #4187 so that when navigating round the SPA (e.g. clicking on an image) user is not re-prompted with pop-up

### DIFF
--- a/kahuna/public/js/services/graphic-image-blur.js
+++ b/kahuna/public/js/services/graphic-image-blur.js
@@ -19,8 +19,10 @@ export const graphicImageBlurService = angular.module("kahuna.services.graphicIm
         $cookies.put(COOKIE_SHOULD_BLUR_GRAPHIC_IMAGES, newCookieValue, cookieOptions);
         window.location.reload();
       },
-      acceptDefaultOfBlurringGraphicImages: () =>
-        $cookies.put(COOKIE_SHOULD_BLUR_GRAPHIC_IMAGES, "true", cookieOptions),
+      acceptDefaultOfBlurringGraphicImages: () => {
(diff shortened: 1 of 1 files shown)


Georges-GNM (approved):
Looks good to me! (Could have sworn I'd seen that in the original PR?)

twrichards:
> Looks good to me! (Could have sworn I'd seen that in the original PR?)

there was already a reload for when changing the cookie to a different value, but I had originally wanted to avoid the reload when accepting the default, since it was already applying blurring. using a reload to fix this is quite blunt, just wanna get a fix out then see if I can refine after

Georges-GNM:
> > Looks good to me! (Could have sworn I'd seen that in the original PR?)
> 
> there was already a reload for when changing the cookie to a different value, but I had originally wanted to avoid the reload when accepting the default, since it was already applying blurring. using a reload to fix this is quite blunt, just wanna get a fix out then see if I can refine after

Ah yeah, just had a look to check where the reload was set in the first one. Maybe blunt sure, but it'll do for now!

twrichards:
> > > Looks good to me! (Could have sworn I'd seen that in the original PR?)
> > 
> > 
> > there was already a reload for when changing the cookie to a different value, but I had originally wanted to avoid the reload when accepting the default, since it was already applying blurring. using a reload to fix this is quite blunt, just wanna get a fix out then see if I can refine after
> 
> Ah yeah, just had a look to check where the reload was set in the first one. Maybe blunt sure, but it'll do for now!

Given [angular services don't support two way binding](https://stackoverflow.com/a/46402351) without adding heaps of extra watch logic which is brittle - I'm going to stick with this approach as its not frequently used functionality.

PR #4189, opened 2023-11-09
use 'persistent' cookie for 'graphic content' blurring (rather than session cookie)

twrichards:
Turns out that Angular's `$cookies` library defaults to session cookies (which I didn't spot, since the extension I was using to view the cookies whilst developing showed an expiry time). 

This PR adds an `expires` property to the cookie when setting it (set to today plus 399 days - the max expiry enforced by some browsers) and if the cookie exists it resets it every page load to effectively extend the expiry indefinitely (that also cleans-up/replaces the old 'session' cookies).


Georges-GNM (approved):
🍪  👍

PR #4196, opened 2023-11-21
make the 'graphic content blurring' explainer per-organisation

twrichards:
We need different organisation specific explainers for the 'graphic content' blurring, since it's 'opt-out' (see https://github.com/guardian/grid/pull/4187).

So based on the existing config property `branding.staffPhotographerOrganisation`, it loads the relevant HTML file from the new `public/js/common/blurring` directory (e.g. staffPhotographerOrganisation is `GNM` so loads `GNM-explainer.html` from that directory. This is rendered on top of the translucent overlay and the GNM implementation references the `graphic-image-blur-explainer__tooltip` class as before but allows for completely different styling/layout if needs be).

From a GNM perspective this is a no-op (which can be verified in `TEST` by deleting the `SHOULD_BLUR_GRAPHIC_IMAGES` cookie and reloading to reveal the explainer).

### DIFF
--- a/kahuna/public/js/common/blurring/GNM-explainer.html
+++ b/kahuna/public/js/common/blurring/GNM-explainer.html
@@ -1,0 +1,20 @@
+<div class="graphic-image-blur-explainer__tooltip">
+    <strong>Potentially upsetting/graphic images are now blurred in the Grid with a new setting that is switched on by default.</strong>
+    <p>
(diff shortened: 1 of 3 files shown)


andrew-nowak (approved):
lgtm

AndyKilmory (approved):
Tested with BBC specific settings and all working as expected

dblatcher (approved):

PR #4208, opened 2023-12-12
[graphic-content blurring] refine the blurring based on SMOUT (Sensitive Material Out)

twrichards:
Now only blur images where _smout_ in when it appears capitalised in special instructions or as an entire keyword (case in-sensitive). Crucially, no longer matching on title & description too (like the other search phrases) since this was causing many false positives e.g. when description includes the word `Portsmouth`.

Based on all the [horrific] images I've seen when searching for 'smout' which aren't Portsmouth football games, they consistently have `SMOUT` in special instructions.


twrichards:
tested locally (but with `--use-TEST` i.e. over a million pics) - works nicely

jonathonherbert (approved):

PR #4209, opened 2023-12-12
[graphic-content blurring] fix ORing logic broken in #4208

twrichards:
https://github.com/guardian/grid/pull/4208 changed the blurring logic w.r.t. `SMOUT` but outside the () so was applying blurring regardless of the overall user preference. Ooops.


paperboyo (approved):
This is a blind approval. Based on trust. Where would we be without trust? OTOH…

PR #4210, opened 2023-12-12
Prevent scheduler failures

andrew-nowak:
## What does this change?

If the Runnable in a scheduler throws, the scheduler will stop and no longer run, even if future executions would succeed.

eg. If a quota report fails to parse, sending through a fixed report will still leave no quotas available until media-api is redeployed/restarted.

I checked the codebase for other schedulers and spotted that Reaper would be susceptible to the same problem, so put a try/catch in there too.

A bit hard to trigger an error like this. Possibly we can drop a malformed report into the mail bucket locally or on TEST? give me a shout if you'd like to pair on trying it.

Intermittent failures do not cause consistent failure 👍

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/BaseStore.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/BaseStore.scala
@@ -1,14 +1,13 @@
 package com.gu.mediaservice.lib
 
-import java.io.InputStream
 import akka.actor.{Cancellable, Scheduler}
 import com.gu.Box
 import com.gu.mediaservice.lib.aws.S3
 import com.gu.mediaservice.lib.config.CommonConfig
 import com.gu.mediaservice.lib.logging.GridLogging
 import org.joda.time.DateTime
-import org.slf4j.LoggerFactory
 
+import java.io.InputStream
 import scala.collection.JavaConverters._
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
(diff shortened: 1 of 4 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/BaseStore.scala
```diff
@@ -43,8 +42,16 @@ abstract class BaseStore[TStoreKey, TStoreVal](bucket: String, config: CommonCon
 
   private var cancellable: Option[Cancellable] = None
 
-  def scheduleUpdates(scheduler: Scheduler) {
-    cancellable = Some(scheduler.scheduleAtFixedRate(0.seconds, 10.minutes)(() => update()))
+  def scheduleUpdates(scheduler: Scheduler): Unit = {
+    cancellable = Some(scheduler.scheduleAtFixedRate(0.seconds, 10.minutes)(() => {
+      try {
+        update()
+        lastUpdated.send(DateTime.now())
```
andrew-nowak, line 49: previously 2 out of the 3 subclassing stores did this before trying the update (and the 3rd never sent an updated time). move it here from all of them and do at a sensible time (ie. _after_ the update has returned successfully) so that it makes sense

media-api/app/lib/UsageStore.scala
```diff
@@ -141,18 +141,18 @@ class UsageStore(
...
 
-  def update() {
-    lastUpdated.send(_ => DateTime.now())
-    fetchUsage.foreach { usage => store.send(usage) }
+  def update(): Unit = {
+    store.send(fetchUsage)
   }
 
-  private def fetchUsage: Future[Map[String, UsageStatus]] = {
+  private def fetchUsage: Map[String, UsageStatus] = {
```
andrew-nowak, line 148: this didn't need to be a future at all - the only futures it called were both wrappers around `Future.successful` - so unwrap the futures and just use synchronous code :)

twrichards (approved):
LGTM - nice catch, thank you @andrew-nowak

PR #4219, opened 2023-12-19
FIX pattern matching in the extraction of S3 key from SQS message

twrichards:
FIX to some of the pattern matching added towards the end of #4201 since parsing was failing - added a test to be sure (which also nicely provides an example of an S3 SQS message)

Co-authored-by: @dblatcher


davidfurey (approved):
lgtm

PR #4221, opened 2023-12-20
URI encode the filename in transit through the queue-based mechanism to ensure it handles diacritics etc.

twrichards:
Small tweak following #4201 to ensure we can handle diacritics in filenames uploaded from UI (FTP is unaffected AFAICT) as they pass through S3 metadata.

One could argue instead we use the filename in S3 for filename, and a new metadata header for media id to avoid this 🤔 EDIT: we did go this way in the end in https://github.com/guardian/grid/pull/4222


twrichards:
Closing in favour of this approach https://github.com/guardian/grid/pull/4222 which is generally better

PR #4224, opened 2023-12-21
recover from file not existing in the S3 IngestQueueBucket

twrichards:
Co-authored-by: @dblatcher 

Despite the improved permissions in https://github.com/guardian/editorial-tools-platform/pull/744 our logic for handling when the S3 object referenced in the SQS message doesn't exist (albeit rare) is still problematic, the message is retried every 30mins indefinitely because we don't recover from the specific scenario BEFORE deleting the SQS message... this PR along with https://github.com/guardian/editorial-tools-platform/pull/746 puts that right!


dblatcher (approved):

twrichards:
Tested in TEST ✅  ... by adding a delay to the queue [manually via console] then deleting the file from the ingest bucket before image-loader was able to pick it up, and successfully observed... 
image

PR #4232, opened 2024-01-03
send image loader count metrics both with and WITHOUT the dimension(s) since CloudWatch can't add 

twrichards:
#4230 added a nice dimension for breaking down uploads (successful and failed) by `UploadedBy` but unfortunately CloudWatch (and therefore also Grafana) can't join/sum these back together 🤦‍ - so we need to send both with and WITHOUT the dimension, which it looks like s3watcher used to do too - see... https://github.com/guardian/grid/blob/5b57ac4dd87b0843a35e074af9841cde78a66875/s3watcher/lambda/lib/Metrics.ts#L19-L31

Also added a further dimension for `IsUiUpload` (as per suggestion from @dblatcher ) to make it easier to filter out if needs be 🤞


dblatcher (approved):

twrichards:
Tested in `TEST` ✅

PR #4233, opened 2024-01-03
only attempt to update upload status table for UI uploads

twrichards:
We were attempting to update the upload status table for every single image coming via the queue, which resulted in vast amounts of `ConditionNotMet` errors in the logs - this PR rectifies that.

PLUS, some better logging of `mediaId` as a log marker.


dblatcher (approved):

twrichards:
Tested on `TEST` - will merge tomorrow morning.

PR #4238, opened 2024-01-19
Revert to standard GHA runners, since they're now more powerful (or rather powerful enough)

twrichards:
Reverts guardian/grid#4158 following the announcement (https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/) that standard-runners are now double in size and should be sufficient for grid (since guardian/grid#4158 stated we were likely just on the borderline before).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ on:
       - 'pr**'
 jobs:
   CI:
-    runs-on: 4core-ubuntu-latest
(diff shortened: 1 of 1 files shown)


twrichards:
Thanks for sharing [the](https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/) @akash1810

akash1810 (approved):

PR #4239, opened 2024-01-19
run CI on `push` to ALL branches too (but only actually run if this is a pull_request [incl. from forks] OR push to a branch without a PR)

twrichards:
Prior to https://github.com/guardian/grid/pull/4238 we were running CI on custom runners which cost money, so we were only building once there was a PR, but now we're back using standard runners (i.e. free, since public repo) its useful to have a build for all pushes - however PRs from forks still need to be built and we don't want to build twice for pull_request and push for our internal branches, so added an `if` condition to the overall CI job to hopefully achieve the best of both worlds.

Looks to manifest correctly on this branch... 
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,11 +3,10 @@ on:
   workflow_dispatch: # Manual invocation.
   pull_request:
   push:
-    branches:
-      - main
-      - 'pr**'
 jobs:
   CI:
+    # only actually run if this is a pull_request (incl. from forks) OR push to a branch without a PR
+    if: github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch' || needs.pr-check.outputs.number == null
     runs-on: ubuntu-latest
     permissions:
       # required by aws-actions/configure-aws-credentials


.github/workflows/ci.yml
```diff
@@ -3,11 +3,10 @@ on:
   workflow_dispatch: # Manual invocation.
   pull_request:
   push:
-    branches:
-      - main
-      - 'pr**'
 jobs:
   CI:
+    # only actually run if this is a pull_request from a fork OR push to an internal branch
+    if: github.event.pull_request.head.repo.fork || github.event_name != 'pull_request'
```
andrew-nowak, line 9: what happens when this isn't a PR and you do `.pull_request.head`? are dots in GHA like `?.`s in JS, or will they throw some sort of "cannot get property of undefined" error? (if you swapped the order I think you wouldn't be able to get into that situation)

twrichards: I actually don't know, but I _think_ it behaves like optional chaining, but your suggestion is a good one, so I've updated

twrichards:
Closing this since there's no way to avoid running duplicate builds when there is an active pull request - without introducing an additional 'job' to use the `gh` CLI and passing an output to the existing `CI` job which isn't worth the time/complexity when compared to having to raise a Draft PR when you want to deploy to `TEST`.

PR #4240, opened 2024-01-22
replace `dev/script/push-pr.sh` with an equivalent GHA 

twrichards:
...via `workflow_dispatch` i.e. manual invocation for better developer experience

Tested (with a temporary `push` trigger since `workflow_dispatch` needs to be merged to main before it appears in UI)...
image


.github/workflows/push-pr.yml
```diff
@@ -0,0 +1,26 @@
+name: push-pr
+on:
+  workflow_dispatch:
+    inputs:
+      prNumber:
+        description: 'PR number'
+        required: true
+        type: string
```
andrew-nowak, line 8:
```suggestion
        type: number
```

andrew-nowak (approved):

PR #4241, opened 2024-01-26
use testcontainers directly instead of docker-testkit

andrew-nowak:
## What does this change?

@Conalb97 raised that docker-testkit has some sort of incompatibility with M1 macs on certain JDKs. docker-testkit isn't receiving maintenance updates, and neither is the underlying docker client, so this is a good opportunity to replace it with a direct usage of [testcontainers](https://testcontainers.com/) (mainly taking inspiration from Ophan https://github.com/guardian/ophan/blob/2ecb80da40915ce097e4a1de0e29b352b396323e/shared-lib/src/test/scala/app/BaseElasticTest.scala#L153 though we share the container between the tests, instead of a container per test, for a mild performance boost)

### DIFF
--- a/media-api/test/lib/elasticsearch/ElasticSearchTestBase.scala
+++ b/media-api/test/lib/elasticsearch/ElasticSearchTestBase.scala
@@ -1,61 +1,22 @@
 package lib.elasticsearch
 
-import java.util.UUID
 import com.gu.mediaservice.model._
+import com.gu.mediaservice.testlib.ElasticSearchDockerBase
 import org.joda.time.DateTime
 import org.scalatest.concurrent.PatienceConfiguration.{Interval, Timeout}
 import org.scalatest.concurrent.ScalaFutures
 import org.scalatest.funspec.AnyFunSpec
-import org.scalatest.time.{Milliseconds, Seconds, Span}
-import org.scalatest.BeforeAndAfterAll
 import org.scalatest.matchers.should.Matchers
-import org.testcontainers.containers.wait.strategy.Wait
-import org.testcontainers.elasticsearch.ElasticsearchContainer
+import org.scalatest.time.{Milliseconds, Seconds, Span}
 import play.api.libs.json.JsString
 
-import java.util.concurrent.atomic.AtomicReference
-import scala.util.Properties
-import scala.concurrent.duration._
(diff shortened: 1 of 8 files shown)


AndyKilmory:
I can see nothing that would present a problem for BBC devs. Thanks

media-api/test/lib/elasticsearch/ElasticSearchTestBase.scala
```diff
@@ -1,38 +1,61 @@
...
+  } else None
+
+  val esPort = esContainer.map(_.getMappedPort(9200)).getOrElse(9200)
+  val es6TestUrl = Properties.envOrElse("ES6_TEST_URL", s"http://localhost:$esPort")
+
+  override protected def afterAll(): Unit = {
+    super.afterAll()
+
+    esContainer foreach { _.stop() }
+  }
```
twrichards, line 58: any chance we could DRY this out and have this and thrall extend something in common? (appreciate it's not rule of 3, but seems like a lot of code with various config options therein which presents scope for divergence).

twrichards (approved):
LGTM - nice one @andrew-nowak 🙌 

(note, I rebased https://github.com/guardian/grid/pull/4287 on top to get it to build 🎉  - thank you)

andrew-nowak:
> I had a similar issue with path-manager, which I fixed in [guardian/path-manager#61](https://github.com/guardian/path-manager/pull/61) by adding an explicit dependency on a newer version of jnr-unixsocket

Thanks that's a good find! However the docker lib is still unmaintained and these sorts of problems are gonna continue ramping up - so I think it's still worth replacing with the fresher/slimmer lib

twrichards (approved):
nice, thanks for DRYing

davidfurey:
> > I had a similar issue with path-manager, which I fixed in [guardian/path-manager#61](https://github.com/guardian/path-manager/pull/61) by adding an explicit dependency on a newer version of jnr-unixsocket
> 
> Thanks that's a good find! However the docker lib is still unmaintained and these sorts of problems are gonna continue ramping up - so I think it's still worth replacing with the fresher/slimmer lib

Seems reasonable. If you were feeling keen and wanted to apply the same approach to path-manager, I'd be very grateful.

PR #4247, opened 2024-02-20
fix endless retry for ingestion queue

twrichards:
We observed some files retrying indefinitely (rather than max of two attempts) even though we have logic to check for `ApproximateReceiveCount`... turns out you need to request attributes explicitly when 'receiving' the message... so this PR, requests the `ApproximateReceiveCount` SQS message attribute when receiving message from ingestion queue (otherwise things will retry until the eventually expire - after 4 days currently).


phillipbarron (approved):
LGTM 👍

PR #4248, opened 2024-02-20
don't retry `UnsupportedMimeTypeException` just move straight to fail bucket

twrichards:
following on from #4247 and as per suggestion from @paperboyo let's just move items which fail to ingest with `UnsupportedMimeTypeException` straight to fail bucket, rather than allowing a retry.


Georges-GNM (approved):
Makes sense 👍

PR #4252, opened 2024-03-01
[feature] Hold Alt to exclude, Shift to include when clicking search links

andrew-nowak:
co-authored by: @andrew-nowak (proper code), @paperboyo (copypasta+logorrhea)

## A (sobbing) story [_(skip to content)_](#a-content)
Using metadata for searching is often an exercise in embarking on a journey of discovery too. You search for some text, you find an image, you click on one of it its metadata fields, you note the number of hits and if this number is large and you want to focus the search more, you try to exclude more and more things you don’t want. Or you want to add other metadata conditions to the current one(s). A proper filtering UI helps with the above immediately by displaying different values for multiple metadata fields dynamically in a reverse-popularity sort order, so it’s easy to discover the breakdown of the **specific** corpus you are looking at. It’s then easy to tell immediately how `subjects` for `Victor Hugo` are distributed and separate the writer from the MMA fighter. And the footballer. You can even tell when were they popular, more or less.
Lacking UI like that, my searches often involve opening multiple windows (to note the no. of hits for specific searches) and then copying and pasting parts of the query back to main search as exclusions (to find the ones I haven’t seen yet). This is very boring and manual work. Filters UI should make it a breeze.
But functionality in this PR will make it a bit easier now. It’s not at all discoverable, so is very much a power user level feature, which is sad, but at least it doesn’t interfere with anything else. It has zero visible UI. It uses keyboard shortcuts familiar from other applications.

## What does this change? <a href="#a-content" id="a-content"/>

This adds functionality for more focussed searches while clicking on metadata links and pills in the Info Panel and pills in image thumbnails: holding Alt will exclude, holding Shift will include (add to). In a browser, it adds and excludes from the **current running search**. In the Viewer, excluding just launches a new exclude search (one day, when the Viewer will just be a zoomed-in browser, we will fix that). It applies to:
- metadata values
- Additional metadata values
- Labels (also on thumbnails)
- Collections (also on thumbnails)
- Subjects, People, Keywords

modifier search1

Also, while at it and to make it work, this PR fixes a longstanding confusion of `collection:` and `suppliersCollection:` searches. Now, both work correctly.

Search for something, select an image, expand Info panel, click on any of the above holding Shift or Alt. Do the same from pills on thumbnails and from an image page (in the Viewer).

Some tedious trawling of the corpus is a tad less so.

## Who should look at this?
@guardian/digital-cms 

## Tested? Documented?
- [x] locally by committer

TODO:
- [x] Add this to pills (subjects, people, labels, keywords) 
- [x] Add this to Additional metadata
- [x] Add this to Collections
- [x] learn how to remove all the code repetition (was told it’s fine, will await review later)
- [x] Domain metadata, shockingly, lacks links completely, so no support needed (sad)
- [x] figure out why excluding collections doesn’t work from viewer (image page): `collection:` does not enhance to `~:` like in other instances… ~possibly some suppliersCollection≠collection/~ malarkey~ (not that… but I bet something in `query-syntax.js`, weirdly an extra history step is created too…)

### DIFF
--- a/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
+++ b/kahuna/public/js/components/gr-image-metadata/gr-image-metadata.html
@@ -324,7 +324,6 @@
                                 <a ui-sref="search.results({query: (ctrl.metadata.byline | queryFilter:'by'), nonFree: ctrl.srefNonfree()})"
                                    ng-click="ctrl.searchWithModifiers($event, 'by', ctrl.metadata.byline)"
                                    aria-label="Search images by {{ctrl.metadata.byline}} byline"
-                                   title="Hold alt to exclude or hold shift to include {{ctrl.metadata.byline}}">{{ctrl.metadata.byline}}</a>
                             </span>
 
                             <span class="editable-empty" ng-if="!ctrl.metadata.byline && ctrl.userCanEdit">Unknown (click ✎ to add)</span>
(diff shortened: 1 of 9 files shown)


paperboyo (approved):
This is amazing. Thanks, Andrew for all your help.

PR #4254, opened 2024-03-19
Conal/imagedam 1502 send to capture handle user input

Conalb97:
## What does this change?

As an archivist user of BBC Images, I want to be able to send certain images from BBC Images to BBC Photo Sales, through the Grid UI. This PR introduces this functionality, allowing a user to select a number of images and trigger the outbound photo sales process by clicking on the new 'Send to Photo Sales' button. By doing this, a syndication usage (with a 'pending' usage status) is added to each of the selected image, which is the trigger for an AWS step function that carries out the majority of the outbound logic.  

This PR is only concerned with the ability of a user to select images and add a syndication usage through the frontend, future PRs will handle additional features such as confirmation dialogs, restrictions on what images can be sent, success/failure notifications and changes to the metadata panel.

As this is BBC specific functionality, it is disabled by default via the `showSendToPhotoSales` feature flag.

Screenshot 2024-03-20 at 16 56 07

Note: this PR has been developed on top of the changes introduced in https://github.com/guardian/grid/pull/4244, which is still to be merged. As such there are additional files inlcuded in the PR that are outside its scope. The only files that need to be reviewed as part of this PR are:

- kahuna/app/lib/KahunaConfig.scala
- kahuna/app/views/main.scala.html
- kahuna/public/js/search/results.html
- kahuna/public/js/search/results.js
- kahuna/public/js/services/api/media-api.js
- media-api/app/controllers/MediaApi.scala
- media-api/conf/routes

As a user with elevated permissions and with the `showSendToPhotoSales` flag set to true:

- Select a number of image thumbnails from the main grid view
- Click the 'Send to Photo Sales Button' from the menu that appears

- Confirm that a Pending Publication usage is visible in the usages section of the metadata panel, for the selected images

## Who should look at this?
@guardian/digital-cms

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/kahuna/public/js/search/results.js
+++ b/kahuna/public/js/search/results.js
@@ -387,7 +387,7 @@ results.controller('SearchResultsCtrl', [
       ctrl.sendToCapture = () => {
         ctrl.selectedImages.map(image => {
           console.log("map image: ", image.data.id)
-          mediaApi.sendToCapture(image.data.id)
+          mediaApi.syndicateImage(image.data.id, "Capture", "true")
           // image.data.uploadTime = moment(image.data.uploadTime).format('YYYY-MM-DDTHH:mm:ss.SSSZ');
           // return image;
         });
--- a/kahuna/public/js/search/results.js
+++ b/kahuna/public/js/search/results.js
@@ -96,8 +96,8 @@ results.controller('SearchResultsCtrl', [
         const ctrl = this;
 
         ctrl.$onInit = () => {
-          ctrl.showSendToPhotoSales = () => $window._clientConfig.showSendToPhotoSales
-        }
+          ctrl.showSendToPhotoSales = () => $window._clientConfig.showSendToPhotoSales;
+        };
 
         // Panel control
         ctrl.metadataPanel    = panels.metadataPanel;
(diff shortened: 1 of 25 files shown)


AndyKilmory:
Could amend dev/script/generate-config/service-config.js to provide value for the show/hide flag. Other than the js syntax errors flagged up by CI checks all looks okay.

kahuna/public/js/search/results.js
```diff
@@ -95,6 +95,10 @@ results.controller('SearchResultsCtrl', [
 
         const ctrl = this;
 
+        ctrl.$onInit = () => {
+          ctrl.showSendToPhotoSales = () => $window._clientConfig.showSendToPhotoSales;
```
AndyKilmory, line 99: showSendToPhotoSales could perhaps be a var/const rather than a function (but if makes not functional difference and probably minimal performance impact)

AndyKilmory:
Minor comment added - see above - but all looks okay to me.

paperboyo:
Just for future travellers: closed in favour of https://github.com/guardian/grid/pull/4267.

PR #4255, opened 2024-03-25
replace some hard coded clauses of the Content-Security-Policy with config

twrichards:
@AndyKilmory noticed that there were a few things hardcoded into our CSP which we decided ought to be coming from config.

For example youtube.com (introduced in https://github.com/guardian/grid/pull/4127) is specific to a guardian add-on (https://github.com/guardian/pinboard) which is added via the `scriptsToLoad` config, so this PR extends the `ScriptToLoad` cases class to represent `additionalFrameSourcesForCSP` and `additionalImageSourcesForCSP` (to replace another hard coded pinboard thing, which makes avatars work).

In the process I've removed `https://accounts.google.com` from the `frame-src` portion (added in https://github.com/guardian/grid/commit/bcdcb2e7125ce8a20fd292a2d58a2bc6b3e4b2ae) since I can't see how it would be used (although @andrew-nowak might have an idea).


andrew-nowak:
> In the process I've removed https://accounts.google.com from the frame-src portion (added in https://github.com/guardian/grid/commit/bcdcb2e7125ce8a20fd292a2d58a2bc6b3e4b2ae) since I can't see how it would be used (although @andrew-nowak might have an idea).

this is necessary at least for the guardian for [panda-session](https://github.com/guardian/panda-session) to create the iframe to relog users in after their panda cookie expires, as the reauth will include a redirect through accounts.google.com in that iframe

twrichards:
re-tested on `TEST` after some tweaks, ready for re-review 🙏 

NOTE: 
I've moved the `https://accounts.google.com` into the `security.frameSources` config property in `kahuna.conf` in response to...
> > In the process I've removed [accounts.google.com](https://accounts.google.com) from the frame-src portion (added in [bcdcb2e](https://github.com/guardian/grid/commit/bcdcb2e7125ce8a20fd292a2d58a2bc6b3e4b2ae)) since I can't see how it would be used (although @andrew-nowak might have an idea).
> 
> this is necessary at least for the guardian for [panda-session](https://github.com/guardian/panda-session) to create the iframe to relog users in after their panda cookie expires, as the reauth will include a redirect through accounts.google.com in that iframe

PR #4256, opened 2024-04-02
new react sort control

AndyKilmory:
## What does this change?

The new sort control replaces the existing sort control found in the top bar. The current control displays as;

Screenshot 2024-04-03 at 10 02 39

The new sort control appears as;

Screenshot 2024-04-03 at 12 23 28

It currently offers the same range of sort options as are currently available but has the potential to be extended to offer a wider range of sort options.

This should be tested by ensuring the selected sort option is reflected in the images url (old to new should show as 'oldest'). Testershould also ensure that the control responds correctly under refresh and when the logo is slected.

## Tested? Documented?
- [x] locally by committer


kahuna/public/js/components/gr-sort-control/gr-sort-control.tsx
```diff
@@ -0,0 +1,182 @@
...
+      node = node.parentNode as Element;
+      if (node.classList.contains(className)) {
+        return true;
+      }
+    }
+
+    return false;
+  };
+
+  const autoHideListener = (event: any) => {
```
Conalb97, line 71: from github actions: unexpected any, specify a different type

kahuna/public/js/components/gr-sort-control/gr-sort-control.tsx
```diff
@@ -0,0 +1,182 @@
...
+    } else if (event.key === 'Enter') {
+      if (!isOpen) {
+        setIsOpen(true);
+      } else {
+        handleOptionClick(options[currentIndex]);
+      }
+    }
+  };
+
+  const handleQueryChange = (e: any) => {
```
Conalb97, line 96: from github actions: unexpected any, specify a different type

kahuna/public/js/components/gr-sort-control/gr-sort-control.tsx
```diff
@@ -0,0 +1,182 @@
...
+    const newQuery = e.detail.query ? (" " + e.detail.query) : "";
+    setHasCollection(checkForCollection(newQuery));
+  };
+
+  const handleLogoClick = (e: any) => {
```
Conalb97, line 101: from github actions: unexpected any, specify a different type and 'e' is defined but never used

Conalb97 (commented):
Couple of things flagged by Github actions, but apart from that looks good to me.

kahuna/public/js/components/gr-sort-control/gr-sort-control.tsx
AndyKilmory: understood and accepted

paperboyo:
Heya. Kewl!

Just some testing, will test more later:
1. Not that our naive “responsiveness” from circa 2015 works floawlessly, but maybe we can improve those at least slightly?
   Untitled
   a) I would drop second `Sort by:` when button collapses and turns into an icon with `Sort order` label. Maybe that label should then be `Sort by:`?
   b) this shouldn’t happen:
   image
   c) when newly fixed (see 1a) `Sort by:` button is clicked, the dropdown menu should open such that the button stays put (ie. it can be opened and closed without moving the mouse pointer)
2. ~Not wanting to get into chevron debate (I was in too many), but we have two now working differently:~ [no nevermind, one can argue one is a dropdown and another is… something else ;-)]

AndyKilmory:
Hi Mateusz,
Sorry hadn't spotted those issues - it was working okay and it looks like I've introduced these issues when adding the keyboard based control - will look into fixing

paperboyo:
Some more testing, only pt. 2 serious.

2. `Back to search` button and browser Back button make URL and dropdown inconsistent. Steps:
   a) click Home to reset everything (thanks, works well now!)
   b) click on any Collection, dropdown (correctly) changes to `Added to  Collection (new to old)` and URL (correctly) changes to `orderBy=dateAddedToCollection`
   c) click on any image
   d) click Back to search button or click browser Back button
   Actual result: for both back actions URL is correct, but dropdown incorrectly shows `Upload date (new to old)`
   Expected result: URL and dropdown are always in sync and always correct
3. Keyboard navigation has some issues, but I hope 1a will sort it
4. [very, very minor, to ignore, but wanted to mention] Chosen style of sort UI (no separate order device, spelling out of both options) means we may get into situations (very long names etc). Already `Added to  Collection (new to old)` isn’t really correct: it’s not `new`, it’s `recently added` (but that’s waaay too long!)
5. [minor] Could we add a tiny bit of padding to text inside the dropdown (maybe as much as calendar icon has?):
    image
6. I think we should add `Added to  Collection (old to new)` for consistency, but can be subsequent PR and it’s a feature request as it’s not there currently

AndyKilmory:
Changes made to correct issues 1, 2, 3 and 5. 4 left for UX review and comment. 6 to be addressed in future PR.

paperboyo:
Thanks for the fixes, look much betterer! Playing some more…

7. Keyboard operation still not fully there, I think?
   a) Control won’t open by pressing Space bar after tabbing to it (I think it should? but I’m not accessibility guru, just going by how other stuff works; pressing Space bar in the recording):
   Untitled
   b) When opening dropdown via Enter and immediately hitting Enter again `Select option` shows up. It shouldn’t. Instead, whenever dropdown opens, an already selected option should be highlighted:
   Untitled2
   c) [related to 3b?] there seems to be an intermediary step when pressing up/down arrows going through the dropdown menu options – this third state (a cause of `Select option`?)
   d) when pressing up/down arrows going through the dropdown menu options, images navigation via those same up/down keys should be disabled
9. selecting text of labels should be disabled (see how it is for anytime below), hopefully will fix mouse cursor automagically too:
    Untitled3

AndyKilmory:
PR has been updated to address issues 7 and 8

paperboyo:
Great! Thank you for all the work and fixes, @AndyKilmory! This is ready for a formal code review now.

rhystmills (approved):
Looks good to me 👍

PR #4259, opened 2024-04-04
T1628 usage rights summary

AndyKilmory:
## What does this change?

The aim of this chage is to provide a framework for adapting the information presented in the 'Rights and restrictions' section of the metadata panel when an image (or multupe images) are selected. The actual text and icon displayed is controlled by a separate component - gr-usagerights-bbc.tsx - this controls the range of usage rights subject to the additional information - the logic dictating each set of rights and the icon and text to be displayed.

For example;

Screenshot 2024-04-09 at 09 51 18

In the case of multiple images selected it would look like;

Screenshot 2024-04-09 at 10 11 32

For usage rights not included in the list of options the display is effectively unchanged;

Screenshot 2024-04-09 at 09 52 31

Ensure that when an image matches one fo the criteria in collection in the usage rights tsx file that the usage rights section displays the correct additional text and icon and that when the usage rights of the image is not in that collection that the display remains unchanged.

Ensure that when multiple images are selected the 'info panel' displays usage rights correctly and consistently with current behaviour.

Access to this functionality is controlled by a config variable in the Kahuna config - set usageRightsSummary = true to gain access to functionality (setting to false turns it off) and if the parameeter is not present it defaults to false.

## Tested? Documented?
- [x] locally by committer


KaliedaRik (approved):
We've tested this PR both locally and in our CODE environment. No changes introduced here appear to break existing functionality.

Happy to approve. Though I strongly suggest that we wait until Mon 13 May before merging into production (just in case issues arise).

AndyKilmory:
Many thanks

PR #4260, opened 2024-04-08
trigger CI for `pr*` branches

twrichards:
The `push-pr` workflow (added in https://github.com/guardian/grid/pull/4240) hasn't been triggering CI as it was created to do, this should fix it...


paperboyo (approved):
I believe, I can trust myself with reviewing a two char change… Although…

PR #4262, opened 2024-04-11
improve 'run name' for `push-pr` GHA workflow runs

twrichards:
tiny, but would be good to know which push-pr run is for which PR number

✅ TESTED...
image


paperboyo (approved):
👍

Issue #3860, opened 2022-09-15
When both Allow and Deny leases are applied, Deny should win…


paperboyo:
## Short description
…instead Allow wins, which is not safe.

## Steps to reproduce
Ask someone to apply two active cropping leases: Allow and Deny. Try cropping image yourslef while not having edit_metadata permission. Cropping should be disabled.

## Actual results
Cropping is allowed.

## Expected results
Cropping should be denied.

## OS and browser details
All.

honorcb:
would you ever want to filter which group of users the allow or deny applies to?

for example do you ever want to allow Observer user to crop, but not the Guardian user?  the division in BBC is between News teams  and the Program production teams ,    News users are licenced to Crop agency images,  and Program user not unless they pay.

paperboyo:
Yes, we were thinking about this as this sound helpful. But that’s about it – just thinking. I guess we could think about it as part of multi-tenancy discussions…

PR #4263, opened 2024-04-11
T1629 - interim filters

AndyKilmory:
## What does this change?

This change introduces a 2nd top-bar at the top of the Grid UI to accommodate increased complexity in the filtering, sorting and permissions controls. It introduces a new permissions filters which acts to provide an easy way to define combinations of chips and the chargeable setting when searching images.

Screenshot 2024-04-30 at 12 13 57

Screen shot show the new layout including the new Interim Filters control (written in React), the new Sort Control moved to the 2nd tier (already merged in pervious PR) and a new My Uploads control to improve flexibility over styling.

The contents of the Interim Filter drop down are controlled through configuration;

interimFilterOptions = [
 {
   id:"allPermissions",
   label:"All permissions",
   mapping:"",
   payable:"none"
 },
 {
   id:"usableForAll",
   label:"Usable for all",
   mapping:"is:BBC-owned,-has:restrictions",
   payable:"false"
 },
 {
   id:"usableForNews",
   label:"Agency (Usable for News)",
   mapping:"category:agency",
   payable:"false"
 },
 {
   id:"bbcOwned",
   label:"BBC owned programmes",
   mapping:"category:programmes-organisation-owned",
   payable:"none"
 },
 {
   id:"independent",
   label:"Independent programmes",
   mapping:"category:programmes-independents",
   payable:"none"
 }
]

The use of Interim Filters is controlled use 'usePermissionsFilter' config variable in kahuna (true or false).

Ensure all controls work as described, work for an accessibility perspective and adapt correctly to screen width changes

## Tested? Documented?
- [x] locally by committer


paperboyo:
Thank you! For now, minor points regarding config and naming:
1. Could we (can defo be later!) document all the extensive config [here](https://github.com/guardian/grid/blob/main/docs/06-objects-of-interest/02-config.md) and in PR description (doesn’t mention `kahuna.conf`, `usePermissionsFilter` and `permissionsDefault`?
2. I think we should go back and rename `interimFilterOptions` to `permissionsOptions` (or `permissionsFilterOptions`?). I think the fact this is “interim” is not that helpful and having everything refer to `permissions` or `permissionsFilter` would make it clear this is all related? (so maybe `permissionsFilterDefault` too)?

paperboyo:
_TL;DR: we will give it a code review and can merge if you are happy (only pt. 6 may need fixing as otherwise we have a hard exclusionary behaviour between disparate config options; @AndyKilmory is aware). The below is for your consideration…_

I have tested it a bit more. In general, I think this is a distinctly BBC-only functionality for several reasons, so it may be that the following problems are squarely in your court: we don’t wanna be a blocker with something we are not likely to use. On the other hand, this may be the first piece of functionality which cannot simply be configged-in: some of the following would need to be fixed before that would be possible (ie. before we, or any non-BBC user, could switch this on)…

Some of the reasons we can’t use it, forgetting the fundamental one: we can’t think of useful “permissions” filter for the Guardian as-is, so no real need from us to fix it, let alone quickly or before merge (but still, hopefully, useful):

3. This, when ON, hard-codes BBC-specific `Show payable images`. At the Guardian, and as is default by-config, we use `Free to use` instead. It differs in some fundamental and some trivial ways too:
    a) ours/default is checked, BBC – unchecked
    b) ours/default responds to `show_paid` permission (it controls initial state of the checkbox). This doesn’t work now in this PR: the default state doesn’t correctly respond to the permission.
    c) [trivial] this uses BBC-specific toggle UI (even though there are still checkbox UIs and it collapses into a checkbox UI itself)
4. [trivial, but consistency matters] this uses new type of checkbox UI (for eg. `My uploads`): they have no hover state unlike default Grid ones (bad), but they have unselectable text in labels (good!)
5. this introduces a third (sic!) dropdown UI. We already have two (one too many), not sure why we need a third one:
    image
6. [as mentioned off-github] this doesn’t work with `filters.shouldDisplayOrgOwnedCountAndFilterCheckbox = true`:
    a) the checkbox (`[org]-owned`) isn’t displayed at all
    b) the app goes into a spin and takes down browser in this scenario:
    permissions bug

Here are some general thoughts unrelated to the above. Consider only if useful for you:

7. Your `usableForAll` config (`is:BBC-owned,-has:restrictions`) would be more useful as `is:BBC-owned,-has:restrictions OR is:BBC-owned,has:restrictions,leases.leases.access:allow-use,leases.leases.active:true`, to take into account Restricted imagery which has been granted an Allow cropping lease which is active, but this isn’t possible yet:
    a) we don’t have OR, which, at first, we could allow in search aliases config (to not have to deal with user-facing UI complexity upfront)
    b) leases may need re-modelling (? to be confirmed by an engineer): it is not possible to search for an **effectively** active lease of certain type (ie. I think `?query=leases.leases.active:true leases.leases.access:allow-use` search will return an image with and active Deny cropping lease and not-active Allow cropping one (not intended). This limitation prevents us from making this permission more useful and from using leases in searching.
    c) also see https://github.com/guardian/grid/issues/3860
8. When default Grid collapses UI, `Search filters` dropdown gathers more and more checkboxes:
    image
    With this ON, `Search filters` only ever contains the Date filter (which shows up too high too). It should behave like in default Grid (because of space, but also label makes no sense otherwise):
    image
9. Permissions filter label (here customised to say All images) first collapses to generic `Permissions` and then to just a chevron:
    image
    a) I think it should always use the configured label
    b) it should collapse under `Search filters` (like default `Free to use` does), so that it doesn’t have to do 9a and it wouldn’t ever look like a stray chevron

AndyKilmory:
Thanks @paperboyo - will look to fix blocking issues asap and raise the others with the team for discussion.

AndyKilmory:
Hi @paperboyo, Have pushed changes that I believe fix all the functional issues raised regard the PR. Please could it be re-reviewed? (I'm on leave next week 20th to 24th May so no great rush). Thanks

dev/script/generate-config/service-config.js
```diff
@@ -94,6 +94,11 @@ function getKahunaConfig(config){
         |  }
         |]`;
 
+    const permissionsConfig = stripMargin`
+      |usePermissionsFilter=true
+      |permissionsDefault="allPermissions"
+      `;
```
dblatcher, line 100: I know this is for the default option on the image permissions filter in the UI, but `permissionsDefault="allPermissions"` in the config looked like a security bug at first glance (sounds like "give the user all the permissions unless told otherwise") - could we add a comment here to clarify what this config controls for future reference?

kahuna/public/js/components/gr-permissions-filter/gr-permissions-filter.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+import "./gr-permissions-filter.css";
+import "./gr-toggle-switch.css";
+
+const SHOW_CHARGEABLE = "Show payable images";
+const SHOW_CHARGEABLE_SHORT = "Payable";
+const SELECT_OPTION = "Select an option";
+const CONTROL_TITLE = "Permissions Selector";
+const SELECTED = "Selected";
+const NOT_SELECTED = " Not Selected";
```
dblatcher, line 15:
```suggestion
const NOT_SELECTED = "Not Selected";
```

dblatcher (approved):
looks good - I'd just suggest adding a comment near the "permissionsDefault" config to avoid misunderstanding about what it does.

kahuna/public/js/components/gr-permissions-filter/gr-permissions-filter.tsx
AndyKilmory: Will include this change in next related PR

PR #4265, opened 2024-04-24
Fix issue that stopped Global Errors and Notification Banner displaying correctly simultaneously

AndyKilmory:
## What does this change?

Fixes the problem identified with the Notification Banner control and its 'interaction' with the Global Errors control. It was found that the presence of the Notification Banner control in the main.scala.html file caused the Global Errors to fail to display - all code ran correctly and there were no console errors but the error messages and other user banners did not appear in the browser.

It was found that if the order of the controls was reversed in main.scala.html then the Error control worked correctly but the Notification Banner failed to display.

This was some complex cross interaction between the two components at an AngularJS level. The code has been simplified to ensure full separation of the two controls and the Notification Banner has been wrapped in its own div - this has resulted in the controls workig correctly together.

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/common-lib/src/main/resources/application.conf
+++ b/common-lib/src/main/resources/application.conf
@@ -122,7 +122,17 @@ usageRightsConfigProvider = {
 #   ...
 # ]
 # -----------------------------------------------------------------
-announcements = []
+announcements = [
+{
+announceId: "notification_banner",
+description: "A New Feature Notification Banner",
+endDate: "2025-03-31",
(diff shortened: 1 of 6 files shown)


phillipbarron (approved):
Tested locally - notifications displaying as expected

AndyKilmory:
Hi Phillip, Thanks for the review and approval - please could you also merge the branch as I don't have write access and so can't complete that step. Thanks Andy

PR #4266, opened 2024-04-30
minor js mods to fix sort control console error

AndyKilmory:
minor js mods to fix sort control console error - checking if element classList is not null prior to checking for content.

## What does this change?

Currently there is an issue in the sort control code when assessing the class list for an element and its parents. This results in a console error (it does not impact the operation of the control). The error is;

Screenshot 2024-04-30 at 16 01 37

This can be corrected by null checking against node.classList

Enssure that when the user scrolls through the images pane a console error is not being generated.

## Tested? Documented?
- [x] locally by committer


Conalb97 (commented):
All looks good to me

paperboyo (approved):
Thanks for your review, Conal. Based on that, merging now…

PR #4270, opened 2024-05-07
LOG Guardian users who access grid without any permission

twrichards:
## What does this change?
Wires `Authorisation` into `AuthenticationProviderResources` so `PandaAuthenticationProvider` has access to it in order to verify all user interactions have some form of grid permission (incl. new `grid_access` permission added in https://github.com/guardian/permissions/pull/186) LOGGING ONLY for now. 

Regardless, this should be no-op for other organisations using Grid (since they'll be [plugging](https://github.com/guardian/grid/pull/3131) their own `AuthorisationProvider` or relying on the default implementation of the key function `hasAtLeastBasicAccess` which returns true).

Assuming https://github.com/guardian/permissions/pull/186 is deployed to CODE then deploy this branch to TEST main (or run locally) and experiment accessing grid with and without some form of grid permission (incl. new `grid_access` permission), observe the logs and you should see something like this...
image

adhering better to our security principles [guardian/recommendations@b4746d4/README.md#security-principles](https://github.com/guardian/recommendations/blob/b4746d4cbf0b53f5237980170dd61b5c6b7d8ee8/README.md#security-principles)

## Who should look at this?
@guardian/digital-cms @guardian/newsroom-resilience @RichardLe @AndyKilmory @Conalb97 

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/auth/app/auth/AuthController.scala
+++ b/auth/app/auth/AuthController.scala
@@ -4,15 +4,14 @@ import com.gu.mediaservice.lib.argo.ArgoHelpers
 import com.gu.mediaservice.lib.argo.model.Link
 import com.gu.mediaservice.lib.auth.Authentication.{InnerServicePrincipal, MachinePrincipal, UserPrincipal}
 import com.gu.mediaservice.lib.auth.Permissions.{DeleteImage, ShowPaid, UploadImages}
-import com.gu.mediaservice.lib.auth.provider.{AuthenticationProviders, LocalAuthenticationProvider}
+import com.gu.mediaservice.lib.auth.provider.AuthenticationProviders
 import com.gu.mediaservice.lib.auth.{Authentication, Authorisation, Internal}
 import com.gu.mediaservice.lib.guardian.auth.PandaAuthenticationProvider
-import com.gu.pandomainauth.service.CookieUtils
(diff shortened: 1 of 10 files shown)


andrew-nowak (approved):
lgtm; though is the intention that any grid permission will be enough to access? I think it'd be best to only allow in with the access permission (even if any other permission is set) - that's how the existing access permissions (composer & facia) work and makes more sense to me (eg. ejecting a user from a tool, or trying to figure out why one user has access and another doesn't despite both missing the access permission)

(sorry if we've already discussed this and I've forgotten)

twrichards:
> lgtm; though is the intention that any grid permission will be enough to access? I think it'd be best to only allow in with the access permission (even if any other permission is set) - that's how the existing access permissions (composer & facia) work and makes more sense to me (eg. ejecting a user from a tool, or trying to figure out why one user has access and another doesn't despite both missing the access permission)
> 
> (sorry if we've already discussed this and I've forgotten)

Thanks @andrew-nowak for calling this out, and after your subsequent idea in planning to log both scenarios I've done just in https://github.com/guardian/grid/pull/4270/commits/42c0be543964e73a3079dcd6327fd2091eca40a1

PR #4271, opened 2024-05-07
actually enforce that Guardian users have at `grid_access` permission

twrichards:
https://github.com/guardian/grid/pull/4270 added logging for users without any permission whilst we can get everybody's permissions in good shape. 

## What does this change?
This makes use of the `hasAnyGridPermission` val we created in #4270 for the result of the `validateUser` function. We also `override` the `showUnauthedMessage` function to return a more useful 403 response...
image

NOTE: as with  #4270 this should be no-op for other organisations using Grid (since they'll be [plugging](https://github.com/guardian/grid/pull/3131) their own `AuthorisationProvider` or relying on the default implementation of the key function `hasBasicAccess` which returns true).

Assuming https://github.com/guardian/permissions/pull/186 is deployed to CODE then deploy this branch to TEST main (or run locally) and experiment accessing grid with and without some form of grid permission (incl. new `grid_access` permission), observe the logs and you should see something like the image above.

adhering better to our security principles [guardian/recommendations@b4746d4/README.md#security-principles](https://github.com/guardian/recommendations/blob/b4746d4cbf0b53f5237980170dd61b5c6b7d8ee8/README.md#security-principles)

## Who should look at this?
@guardian/digital-cms @guardian/newsroom-resilience @RichardLe @AndyKilmory @Conalb97 

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/rest-lib/src/main/scala/com/gu/mediaservice/lib/auth/provider/AuthorisationProvider.scala
+++ b/rest-lib/src/main/scala/com/gu/mediaservice/lib/auth/provider/AuthorisationProvider.scala
@@ -31,6 +31,5 @@ trait AuthorisationProvider extends Provider {
     */
   def hasPermissionTo(permission: SimplePermission, principal: Principal): Boolean
 
-  def hasBasicAccessExplicitly(userEmail: String): Boolean = false // default implementation
-  def hasAtLeastBasicAccess(userEmail: String): Boolean = true // default implementation
+  def hasBasicAccess(userEmail: String): Boolean = true // default implementation
 }
--- a/rest-lib/src/main/scala/com/gu/mediaservice/lib/auth/Authorisation.scala
+++ b/rest-lib/src/main/scala/com/gu/mediaservice/lib/auth/Authorisation.scala
@@ -85,6 +85,5 @@ class Authorisation(provider: AuthorisationProvider, executionContext: Execution
     }
   }
 
-  def hasBasicAccessExplicitly(userEmail: String): Boolean = provider.hasBasicAccessExplicitly(userEmail)
-  def hasAtLeastBasicAccess(userEmail: String): Boolean = provider.hasAtLeastBasicAccess(userEmail)
+  def hasBasicAccess(userEmail: String): Boolean = provider.hasBasicAccess(userEmail)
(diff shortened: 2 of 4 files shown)


twrichards:
In https://github.com/guardian/grid/pull/4270 we logged both whether the user accessing has `grid_access` permission and if not whether they have at least some other level of permission. There was only one user in the last two weeks who had the latter, so this PR actually simplifies to require `grid_access` permission (see https://github.com/guardian/permissions/pull/186)
CC @andrew-nowak given https://github.com/guardian/grid/pull/4270#pullrequestreview-2048062942

rest-lib/src/main/scala/com/gu/mediaservice/lib/auth/provider/AuthorisationProvider.scala
```diff
@@ -31,6 +31,5 @@ trait AuthorisationProvider extends Provider {
     */
   def hasPermissionTo(permission: SimplePermission, principal: Principal): Boolean
 
-  def hasBasicAccessExplicitly(userEmail: String): Boolean = false // default implementation
-  def hasAtLeastBasicAccess(userEmail: String): Boolean = true // default implementation
+  def hasBasicAccess(userEmail: String): Boolean = true // default implementation
```
andrew-nowak, line 34: I think it would make sense to leave this unimplemented with a comment; it should be easy (one-line change) for other (existing or new) organisations to override with `true` in their implementations if they don't want to do authorisation checks while still making it an explicit decision (and less chance of accidentally leaving open unintentionally)

andrew-nowak (approved):

rest-lib/src/main/scala/com/gu/mediaservice/lib/auth/provider/AuthorisationProvider.scala
twrichards: hmm I did think about that, but didn't want the hassle/delay from other orgs implementing to delay this, after talking with you @andrew-nowak in person, I'll make the breaking change (which requires other orgs to implement in their plugged-in auth code) in a follow-up PR so I can merge this one now...

PR #4273, opened 2024-05-08
t1720 - basic configurable labels for sortControl

AndyKilmory:
## What does this change?

This change allows an organisation to customise the labels seen on the sort control (an provides a pattern for general label customisation).

It uses the strings.json to hold the label overrides. The key for the entry in strings.json takes the form "<<control name>>_<<label key>>", e.g. "gr-sort-control_oldest". The entry would then be;

"gr-sort-control_oldest": "Upload - oldest first"

If there isn't a key matching the format required in strings.json then the default label is used.

Ensure that when an appropriate configured entry is added to string.json it appears correctly in the UI for the sort control.

## Tested? Documented?
- [x] locally by committer


Conalb97 (commented):
Looks good to me :)

PR #4267, opened 2024-05-02
IMAGEDAM-1665: Generic React modal dialog and send to Photosales functionality

Conalb97:
### Note, this PR builds on changes introduced in IMAGEDAM-1502 and so will replace that PR.

This PR introduces two new functionalities:
- The ability for a user to add a 'send to photosales' syndication usage to an image, via the 'Send to Photosales button' 
- A generic React modal component, used in this case as a confirmation dialog

As an archivist user of BBC Images, I want to be able to send certain images from BBC Images to BBC Photo Sales, through the Grid UI. This PR introduces this functionality, allowing a user to select a number of images and trigger the outbound photo sales process by clicking on the new 'Send to Photo Sales' button.

Upon doing this the confirmation dialog will pop up, asking the user to confirm their choice. The dialog will have different text and button options depending on if the images selected already exist in photosales or not. Once the user has confirmed their intention to send the image, a syndication usage (with a 'pending' usage status) is added to each of the selected image, which is the trigger for an AWS step function that carries out the majority of the outbound logic.

The Photosales functionality is BBC specific, hence it is disabled by default via the `showSendToPhotoSales` feature flag.

image

## How should a reviewer test this change?

As a user with elevated permissions and with the showSendToPhotoSales flag set to true:

### No selected images already exist in Photosales
- Select a number of image thumbnails from the main grid view, ensure none of these images have already been sent to Photosales 
- Click the 'Send to Photo Sales Button' from the menu that appears
- A confirmation dialog will pop up
- Click 'Yes, send'

### The selected images contain a mixture of images that have and have not been sent to Photosales
- Select a number of image thumbnails from the main grid view, ensure that some of these images have already been sent to Photosales and some have not been sent
- Click the 'Send to Photo Sales Button' from the menu that appears
- A confirmation dialog will pop up
- Click 'Yes, send'

### All the selected images have already been sent to Photosales
- Select a number of image thumbnails from the main grid view, ensure that all of these images ahve already been sent to Photosales 
- Click the 'Send to Photo Sales Button' from the menu that appears
- A confirmation dialog, with one button will pop up
- Click 'Okay'

### No selected images already exist in Photosales
- Confirm that a Pending Publication usage is visible in the usages section of the metadata panel, for the selected images
- Confirm that the confirmation dialog has the following appearance:
Screenshot 2024-05-02 at 10 43 41

### The selected images contain a mixture of images that have and have not been sent to Photosales
[...]

PR #4275, opened 2024-05-13
IMAGEDAM-1666: Photosales icon is visible when syndication usage is present

Conalb97:
## What does this change?

**Note: this PR should be reviewed after #4267 as it builds on the changes included in that PR**

The relevant files for this PR are:
- kahuna/public/js/components/gr-icon/gr-icon.css
- kahuna/public/js/components/gr-icon/gr-icon.js
- kahuna/public/js/components/gr-icon/icons/sent-to-photosales.svg
- kahuna/public/js/preview/image.html
- kahuna/public/js/preview/image.js
- kahuna/public/js/services/image/usages.js

As an archivist user of BBC Images, after I have sent an image to BBC Photo Sales I want to see a visual indication that this has taken place. Therefore, after an image has been selected and sent to Photo Sales a 'sales' icon should appear in the thumbnail of that image.

The Photosales functionality is BBC specific, hence it is disabled by default via the showSendToPhotoSales feature flag.

A reviewer can test this change by selecting an image and indicating that they want to send it to BBC Photo Sales. After this action has completed, a 'SALES' icon should appear in the thumnbail of that image, as demonstrated in the following screenshot:

Screenshot 2024-05-13 at 10 20 21

This testing can be considered successful if after following the above steps, the 'SALES' icon appears on the selected image, as demonstrated in the previous screenshot. Additionally, images without a syndication usage should not have this icon present.  

## Tested? Documented?
- [x] locally by committer


paperboyo:
Just a note that we already do have icons associated with [different syndication statuses](https://github.com/guardian/grid/blob/main/common-lib/src/main/scala/com/gu/mediaservice/model/SyndicationStatus.scala#L5-L11):
- [review](https://media.gutools.co.uk/search?nonFree=true&syndicationStatus=review):
    image
- [queued](https://media.gutools.co.uk/search?nonFree=true&syndicationStatus=queued):
    image
- [blocked](https://media.gutools.co.uk/search?nonFree=true&syndicationStatus=blocked):
   image
- [sent](https://media.gutools.co.uk/search?nonFree=true&syndicationStatus=sent):
   image

Now, I have no idea how much of BBC syndication code and functionality overlaps Guardian’s. Or would it be helpful if it did. Just wanted to mention in case useful.

kahuna/public/js/preview/image.js
```diff
@@ -64,6 +64,8 @@ image.controller('uiPreviewImageCtrl', [
           ctrl.imageAsArray = [newImage];
       });
 
+      ctrl.showSendToPhotoSales = () => $window._clientConfig.showSendToPhotoSales;
+
```
AndyKilmory, line 68: Does this need to be a function call here or can it be accessed directly as a variable? (changing to variable call would impact image.html)

rebecca-thompson (approved):
I have tested this locally and it all works as expected and the code looks good.

Just want to second @paperboyo's very valid comment that we already have an icon slot available for syndication status and it would be great to be able to configure that slot rather than having an increasing amount of Guardian-specific and BBC-specific code.

Perhaps a conversation for another time, but something we should definitely consider when migrating this codebase to React. 

IMO, making icons configurable would be a much more sustainable approach.

Conalb97:
> I have tested this locally and it all works as expected and the code looks good.
> 
> Just want to second @paperboyo's very valid comment that we already have an icon slot available for syndication status and it would be great to be able to configure that slot rather than having an increasing amount of Guardian-specific and BBC-specific code.
> 
> Perhaps a conversation for another time, but something we should definitely consider when migrating this codebase to React.
> 
> IMO, making icons configurable would be a much more sustainable approach.

Thanks for the review Rebecca! Apologies for just getting round to responding to this now, I was on a training course earlier this week and this PR slipped my mind until now!

I agree that making icons configurable would be a good change and certainly something to look into for the future.

Would someone from the Guardian be able to merge this PR now it's approved? :)

PR #4277, opened 2024-05-15
Remove unused Thrall MetadataEditorNotifications.

tonytw1:
## What does this change?

Drops mandatory but unused thrall config element `indexed.image.sns.topic.arn` 

This in turn means you don't need to provision this SNS queue, which is one less piece of infrastructure to build when adopting the grid.

No visible effects because the code does not appear to be pushing messages onto this queue.
Opportunity to remove a piece of infrastructure.

## Tested? Documented?
- [ ] locally by committer


tonytw1:
Closing; a stand alone Grid doesn't need this SNS topic because nothing else in the Grid listens for it BUT their is probably something at the Guardian that IS listening.

PR #4279, opened 2024-05-18
Delete unused authorisedForDeleteImageOrUploader method on MediaAPI

tonytw1:
## What does this change?

Delete 2 unused private methods.

Looks like this is from before the generic auth changes. private methods which do not look like overrides and the application compiles without then.

Complies and tests pass.

Should have zero impact; removes a piece redundant code and makes this class slightly easier to alter.

## Tested? Documented?
- [x] locally by committer


andrew-nowak (approved):
thanks!

tonytw1:
Rebasing

PR #4283, opened 2024-05-31
IMAGEDAM-1667: Metadata panel displays send to photo sales usages

Conalb97:
## What does this change?

Note: this PR should be reviewed after https://github.com/guardian/grid/pull/4267 as it builds on the changes included in that PR

The relevant files for this PR are:

- kahuna/public/js/components/gr-image-usage-photosales/gr-image-usage-photosales.css
- kahuna/public/js/components/gr-image-usage-photosales/gr-image-usage-photosales.css
- kahuna/public/js/components/gr-image-usage/gr-image-usage.html
- kahuna/public/js/components/gr-image-usage/gr-image-usage.js
- kahuna/public/js/util/constants/sendToCapture-config.js
- usage/app/controllers/UsageApi.scala

Previously, after the user had completed the process of adding a photo sales and they navigated to the usages panel the usage was presented as if it was a standard syndication usage. Now a custom photo sales 'usage' is rendered instead. Additionally, as deleting the usage from the image, does not cause the image to be deleted in photo sales, changes have been made to usagesAPI so that the 'Delete All' button does not delete a syndication usage when it has a partnerName = Capture (i.e. is a photo sales usage).

The Photosales functionality is BBC specific, hence it is disabled by default via the showSendToPhotoSales feature flag. 

- Select a number of image thumbnails from the main grid view, ensure none of these images have already been sent to Photosales
- Click the 'Send to Photo Sales Button' from the menu that appears
- A confirmation dialog modal will pop up
- Click 'Yes, send'
- A green notification banner will pop up
- Navigate to the usages panel 
- A new 'Photo Sales' usage has been added
- Then, click the 'delete all button' and wait for the page to refresh

Confirm that the added photo sales usage has the following appearance:

Screenshot 2024-05-31 at 16 07 37

And confirm that after clicking the 'delete all button' the 'Added to Photo Sales' usage remains on the image.

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-image-usage-photosales/gr-image-usage-photosales.tsx
+++ b/kahuna/public/js/components/gr-image-usage-photosales/gr-image-usage-photosales.tsx
@@ -1,0 +1,67 @@
+import * as React from "react";
+import * as angular from "angular";
+import { react2angular } from "react2angular";
+import {useEffect, useState} from "react";
+import * as moment from 'moment';
+import "./gr-image-usage-photosales.css";
+
+export interface Usage {
+  title: string,
+  usageName: string,
+  usageType: string,
+  dateAdded: Date
+}
+
+export interface UsageProps {
+  usages: Usage[];
+}
+
+export interface UsagePropsWrapper {
+  props: UsageProps;
+}
+
+const ImageUsagePhotoSales: React.FC<UsagePropsWrapper> = ({ props }) => {
+  const [usages, setUsages] = useState(props.usages);
+  const [photoSalesUsage, setPhotoSalesUsage] = useState<Usage|null>(null);
+
+  const photoSalesIdentifier = "sendToPhotoSales";
+
+  const formatTimestamp = (timestamp: Date) => {
+    return moment(timestamp).fromNow();
+  };
+
+  useEffect(()=> {
+    const photoSalesUsage = usages.find(usage => usage.usageType === photoSalesIdentifier);
+    if (photoSalesUsage) {
+      setPhotoSalesUsage(photoSalesUsage);
+      setUsages(usages.filter(usage => usage.usageType !== photoSalesIdentifier));
+    } else {
+      setPhotoSalesUsage(null);
+    }
+  }, []);
+
+  return (
+    <div>
+      {photoSalesUsage && (
+        <div className="image-info__group">
+          <span className="usageNameStyle">
+            {photoSalesUsage.usageName}
+          </span>
+          <ul>
+            <li className="usageRecordStyle">
+
+              {photoSalesUsage.title}
+
+              <div className="usageMoreInfoStyle">
+                <span className="usageDateAddedStyle">{formatTimestamp(photoSalesUsage.dateAdded)}</span>
+              </div>
+            </li>
+          </ul>
+        </div>
+      )}
+    </div>
+  );
+};
+
+export const imageUsagePhotoSales = angular.module('gr.imageUsagePhotoSales', [])
+  .component('imageUsagePhotoSales', react2angular(ImageUsagePhotoSales, ["props"]));
--- a/kahuna/public/js/components/gr-image-usage/gr-image-usage.html
+++ b/kahuna/public/js/components/gr-image-usage/gr-image-usage.html
@@ -1,9 +1,18 @@
-<div ng-repeat="(type, usages) in ctrl.usages">
-  <gr-image-usage-list type="ctrl.usageTypeToName(type)" usages="usages"></gr-image-usage-list>
+<div ng-if="!ctrl.showSendToPhotoSales">
+    <div ng-repeat="(type, usages) in ctrl.usages">
+        <gr-image-usage-list type="ctrl.usageTypeToName(type)" usages="usages"></gr-image-usage-list>
+    </div>
+</div>
+<div ng-if="ctrl.showSendToPhotoSales">
+    <image-usage-photo-sales props="{'usages':ctrl.photoSalesUsages}"></image-usage-photo-sales>
+    <div ng-repeat="(type, usages) in ctrl.usages" ng-if="type ==='downloaded'">
+        <gr-image-usage-list type="ctrl.usageTypeToName(type)" usages="usages"></gr-image-usage-list>
+    </div>
 </div>
 
+
 <gr-delete-usages
-  ng-if="ctrl.usagesCount > 0"
+  ng-if="(ctrl.usagesCount > 0 && !(ctrl.showSendToPhotoSales && ctrl.hasSyndicationUsages))"
   gr-image="ctrl.image"
   gr-on-delete="ctrl.onUsagesDeleted()">
 </gr-delete-usages>
--- a/kahuna/public/js/components/gr-image-usage/gr-image-usage.js
+++ b/kahuna/public/js/components/gr-image-usage/gr-image-usage.js
@@ -10,8 +10,12 @@ import './gr-image-usage.css';
 import '../../services/image/usages';
 import {deleteUsages} from '../gr-delete-usages/gr-delete-usages';
 
+import '../gr-image-usage-photosales/gr-image-usage-photosales';
+import {sendToCaptureUsagePanelTxt} from "../../util/constants/sendToCapture-config";
+
 export const module = angular.module('gr.imageUsage', [
   'gr.image-usages.service',
+  'gr.imageUsagePhotoSales',
   'util.rx',
   deleteUsages.name
 ]);
@@ -20,15 +24,17 @@ module.controller('grImageUsageCtrl', [
   '$scope',
   '$state',
   'inject$',
+  '$window',
   'imageUsagesService',
 
-  function ($scope, $state, inject$, imageUsagesService) {
+  function ($scope, $state, inject$, $window, imageUsagesService) {
 
     const ctrl = this;
 
     ctrl.$onInit = () => {
-      const usages = imageUsagesService.getUsages(ctrl.image);
+      ctrl.showSendToPhotoSales = $window._clientConfig.showSendToPhotoSales;
 
+      const usages = imageUsagesService.getUsages(ctrl.image);
       const usages$ = usages.groupedByState$.map((u) => u.toJS());
       const usagesCount$ = usages.count$;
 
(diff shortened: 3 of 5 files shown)


usage/app/controllers/UsageApi.scala
```diff
@@ -321,19 +322,40 @@ class UsageApi(
...
+          val updateMessage = UpdateMessage(subject = DeleteUsages, id = Some(mediaId))
+          notifications.publish(updateMessage)
+        } else {
+          filteredUsages.foreach(usage => {
+            usageTable.deleteRecord(usage)
+            val usageId = Some(usage.grouping + "_" + usage.usageId.toString)
+            val updateMessage = UpdateMessage(subject = DeleteSingleUsage, id = Some(mediaId), usageId = usageId)
+            notifications.publish(updateMessage)
+          })
+        }
```
AndyKilmory, line 349: Does this section of code line 339 - line 349 need to sit within the map command - in the original method the thrall elements sat outside that block and I think that could be the case here too?

kahuna/public/js/components/gr-image-usage/gr-image-usage.html
```diff
@@ -1,6 +1,15 @@
...
+    <div ng-repeat="(type, usages) in ctrl.usages">
+        <gr-image-usage-list type="ctrl.usageTypeToName(type)" usages="usages"></gr-image-usage-list>
+    </div>
 </div>
+<div ng-if="ctrl.showSendToPhotoSales">
+    <image-usage-photo-sales props="{'usages':ctrl.processedReactUsages}"></image-usage-photo-sales>
+    <div ng-repeat="(type, usages) in ctrl.usages" ng-if="type ==='downloaded'">
+        <gr-image-usage-list type="ctrl.usageTypeToName(type)" usages="usages"></gr-image-usage-list>
+    </div>
+</div>
```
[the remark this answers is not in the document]
Conalb97: From a discussion at standup, it was decided that only the 'new' photosales usage should be rendered with React - all other exisitng usages would be handled with the AngularJs. And for the BBC's use case it was confirmed that 'download' is the only usage we currently use that will be rendered with AngularJs. 

Therefore, in the case of additional usage types being handled in the future, these will be processed as React usages (i.e. included in the props sent into the React component). In which case the existing react component and processing logic should be straightforward to extend.

Furthermore, at some point in the future the BBC may decide to convert the 'download' usage to React as well, again this should be relatively straightforward to acheive, but was deemed outside the scope of this ticket.

usage/app/controllers/UsageApi.scala
Conalb97: At first I tried to replicate the original approach, with the Thrall elements sitting outside the query block.

This resulted in an issue where the thrall elements were processed before the query block (despite coming after the query block in terms of lines of code) - this is because the query block invovles Futures.

I think this was already happening in the original code, but it wasn't picked up because the deletion from elastic search and the deletion from dynamoDB could occur independently of each other in this case. However for our case, we rely on logic from the query block to determine what usages should be removed from elastic. Therefore, with the thrall block executing first, it was resulting in all usages being deleted, and the photo sales usage was not retained.

By moving the thrall block inside the query block, this forced correct order of operation and provides the desired result.

Conalb97 (commented):
Added some responses!

usage/app/controllers/UsageApi.scala
AndyKilmory: Okay

kahuna/public/js/components/gr-image-usage/gr-image-usage.html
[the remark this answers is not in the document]
AndyKilmory: OKay

kahuna/public/js/components/gr-image-usage-photosales/gr-image-usage-photosales.tsx
```diff
@@ -0,0 +1,67 @@
...
+  usageName: string,
+  usageType: string,
+  dateAdded: Date
+}
+
+export interface UsageProps {
+  usages: Usage[];
+}
+
+export interface UsagePropsWrapper {
```
rebecca-thompson, line 19: You don't need to wrap the props. You can use `UsageProps` directly, then the corresponding Angular code becomes:

`<image-usage-photo-sales usages="ctrl.processedReactUsages"></image-usage-photo-sales>`

kahuna/public/js/components/gr-image-usage/gr-image-usage.js
```diff
@@ -50,6 +56,23 @@ module.controller('grImageUsageCtrl', [
         }
       };
 
+      ctrl.processedReactUsages = () => {
```
rebecca-thompson, line 59: Can we rename this function to something like `ctrl.photoSalesUsages`?

kahuna/public/js/components/gr-image-usage-photosales/gr-image-usage-photosales.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const [usages, setUsages] = useState(props.usages);
+  const [extractedPhotoSalesUsage, setPhotoSalesUsage] = useState<Usage|null>(null);
+
+  const photoSalesIdentifier = "sendToPhotoSales";
+
+  const formatTimestamp = (timestamp: Date) => {
+    return moment(timestamp).fromNow();
+  };
+
+  useEffect(()=> {
```
rebecca-thompson, line 33: You can remove this `useEffect` block by inlining the initialisation of the state. Ie:

```
const [photoSalesUsage] = useState<Usage | undefined>(usages().find(usage => usage.usageType === "sendToPhotoSales"));
```

This highlighted that the UsageProps interface is actually incorrect here. It should be:

```
export interface UsageProps {
  usages: () => Usage[];
}
```

You are passing in a function, not an array.

kahuna/public/js/components/gr-image-usage/gr-image-usage.js
```diff
@@ -50,6 +56,23 @@ module.controller('grImageUsageCtrl', [
...
+            processedPhotoSalesUsage = {
+              title: usage.data.syndicationUsageMetadata.syndicatedBy,
+              usageName: sendToCaptureUsagePanelTxt,
+              usageType: "sendToPhotoSales",
+              dateAdded: usage.data.dateAdded
+            };
+          }
+        });
+        processedUsages.push(processedPhotoSalesUsage);
+        return processedUsages;
```
rebecca-thompson, line 73: This code could be simplified to:

```
        const processedUsages = [];
        ctrl.image.data.usages.data.forEach((usage)  => {
          if (usage.data.platform === "syndication" && usage.data.syndicationUsageMetadata.partnerName === "Capture") {
            processedUsages.push({
              title: usage.data.syndicationUsageMetadata.syndicatedBy,
              usageName: sendToCaptureUsagePanelTxt,
              usageType: "sendToPhotoSales",
              dateAdded: usage.data.dateAdded
            });
          }
        });
        return processedUsages;
```

usage/app/controllers/UsageApi.scala
```diff
@@ -321,19 +322,40 @@ class UsageApi(
...
-        logger.error(logMarker, "UsageApi returned an error.", error)
-        respondError(InternalServerError, "image-usage-delete-failed", error.getMessage)
-    }
 
-    val updateMessage = UpdateMessage(subject = DeleteUsages, id = Some(mediaId))
-    notifications.publish(updateMessage)
+    var photoSalesUsage = false
+    var filteredUsages = ListBuffer[MediaUsage]()
+
+    usageTable.queryByImageId(mediaId)
```
rebecca-thompson, line 329: We should be able to write this in a functional way, avoiding the need for `var` and mutable data structures. Something like this... ?

```
    usageTable.queryByImageId(mediaId)
      .map(usages => {
        val hasPhotoSalesUsage = (usage: MediaUsage) => usage.syndicationUsageMetadata.exists(_.partnerName == "Capture")
        val usagesWithoutPhotoSales = usages.filter(!hasPhotoSalesUsage(_))

        usagesWithoutPhotoSales.foreach(usage => usageTable.deleteRecord(usage))

        if (usages.exists(hasPhotoSalesUsage)) {
          usagesWithoutPhotoSales foreach { usage =>
            val usageId = Some(usage.grouping + "_" + usage.usageId.toString)
            val updateMessage = UpdateMessage(subject = DeleteSingleUsage, id = Some(mediaId), usageId = usageId)
            notifications.publish(updateMessage)
          }
        } else {
          val updateMessage = UpdateMessage(subject = DeleteUsages, id = Some(mediaId))
          notifications.publish(updateMessage)
        }
      })
```

I'm not sure if it makes the logic flow of the code a little easier to follow

rebecca-thompson (commented):
A few comments, but looking great!

I have one question around changing the behaviour of the 'deleteUsages' endpoint... 

Having spoken with @paperboyo about the purpose of the 'Delete ALL Usages' button, it is only used as a breakglass measure, usually just before deleting an image from the Grid, and is only available to a small handful of super-admin users. 

By excluding PhotoSales syndication from deletion, it means you will effectively have images that can **never be deleted**, unless there is future work done to add 'Remove from PhotoSales' functionality. Just checking that this is ok and intended.

Conalb97:
Thanks for the review @rebecca-thompson, just going to go through your comments now to try and resolve them :)

Conalb97 (commented):
Replied to code review comments. With regards to delete all and not being able to delete an image, this is something we're okay with. The main reason (initially) is to prevent a user from deleting usages from BBC Images, which will not translate to the photosales site. Future work is planned for this functionality, at which point this feature can be added back in.

Thanks!

Conalb97:
Hi @rebecca-thompson :) 

After internal discussion and speaking with @paperboyo, we've removed the BBC specific logic from the delete-all endpoint. 

As a replacement we now hide the 'delete all' button on images that have syndication usages (but only if the photosales feature flag is turned on). That way we retain the protection against orphaned images in BBC Photosales, without impacting the Guardian or the code base. This does mean that we will be unable to delete images. However, we are okay with this for now as future work will introduce the functionality to delete images from photosales, at which point we can 'reintroduce' the delete all button. In the meantime, situations where an image needs to be deleted is seen as unlikely, and manageable manually.

Thanks!

rebecca-thompson (approved):

PR #4286, opened 2024-06-13
Fix test running in IntelliJ UI by resolving ScalaTest 3.1 vs 3.2 conflict

tonytw1:
## What does this change?

Fixes this error when trying to run a single test in the IntelliJ IDE:

Screenshot 2024-06-09 at 18 43 03

Downgrade "org.scalatestplus" %% "mockito-3-4" to 3.1 series to match the 3.1 ScalaTest pulled in by scalatestplus-play.

Resolves a scalac -deprecation snag which breaks running of tests in IntelliJ. Warns about a deprecation on .right.value but then doesn't know about the new method because ScalaTest 3.2 is not available on the IntelliJ classpath.

This commit works by pushing the transitive ScalaTest from "org.scalatestplus" %% "mockito-3-4" back down to 3.1 for compatibility with "scalatestplus-play".

Notably "scalatestplus-play" cannot be leveled to 6.0.0 / ScalaTest 3.2 without moving to Scala 2.13 because it has not been released for 2.12.

Single tests can be run in the IDE. Faster feedback loop. Prevent the urban myth that you can't run tests in the IDE from taking hold.

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -29,7 +29,7 @@ val commonSettings = Seq(
   Test / testOptions ++= Seq(Tests.Argument(TestFrameworks.ScalaTest, "-o"), Tests.Argument(TestFrameworks.ScalaTest, "-u", "logs/test-reports")),
   libraryDependencies ++= Seq(
     "org.scalatestplus.play" %% "scalatestplus-play" % "5.1.0" % Test,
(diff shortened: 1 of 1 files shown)


tonytw1:
Hmmm... CI failure but works reliability on a local i386 machine.

tonytw1:
FYI - CI fails even without my change.

PR #4287, opened 2024-06-16
[reaper] retain all images in collections, if 'persisted collections' not specified

twrichards:
Since the reaper was reinstated in https://github.com/guardian/grid/pull/4145 - there have been a few user misunderstandings and work lost (actually only _almost_ lost, thanks to the additional 14days in soft-delete state 💡 😌) because people intuitively think putting something in a collection will protect it from deletion. In this PR we aim to match people's expectations (to prevent loss of work) whilst also ensuring it's still possibly to limit the collections which cause images to be retained/persisted.

## What does this change?
Replaces the config option `persistence.collections` with `persistence.onlyTheseCollections` to better reflect the new purpose, or rather the implication that if not set, all collections will be persisted (i.e. saved from the wrath of the reaper). The new config option now also supports using proper HOCON list/array (thanks for the suggestion @andrew-nowak). This is then modelled as `Option[Set[String]]` as it's passed around the codebase, to result in a new image persistence reason: `collection` (in addition to the existing `persisted-collection`) and results in the relevant ES query being executed, depending on whether the config option is defined or not.

Before deploying this branch to TEST environment, PAUSE the reaper via the thrall dashboard. Find the oldest images which are surfaced via `+is:reapable` search, add some to a collection. RESUME the reaper. See that those images are retained (but more recent ones do get soft-deleted - findable via `+is:deleted`). 

To be thorough one could test configuring `persistence.onlyTheseCollections` but that works as `persistence.collections` did before the _rename_. 

One could also test specifying an empty list for `persistence.onlyTheseCollections` to not retain images based on collections at all.

✅  all the above have been tested in TEST by @twrichards 

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfigWithElastic.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfigWithElastic.scala
@@ -17,8 +17,6 @@ class CommonConfigWithElastic(resources: GridConfigResources) extends CommonConf
   val persistenceIdentifier = string("persistence.identifier")
   val queriableIdentifiers = Seq(persistenceIdentifier)
 
-  val persistedRootCollections: List[String] = stringOpt("persistence.collections") match {
-    case Some(collections) => collections.split(',').toList
-    case None => List(s"${staffPhotographerOrganisation} Archive")
-  }
+  // note this will match any part of the collection path, e.g. "bar" will match "bar", "foo/bar", "bar/baz"
+  val maybePersistOnlyTheseCollections: Option[Set[String]] = getOptionalStringSet("persistence.onlyTheseCollections")
 }
--- a/media-api/app/lib/ImagePersistenceReasons.scala
+++ b/media-api/app/lib/ImagePersistenceReasons.scala
@@ -5,7 +5,7 @@ import com.gu.mediaservice.model.{CommissionedAgency, Illustrator, Image, ImageM
 import com.sksamuel.elastic4s.requests.searches.queries.Query
 
 
-case class ImagePersistenceReasons(persistedRootCollections: List[String], persistenceIdentifier: String) {
+case class ImagePersistenceReasons(maybePersistOnlyTheseCollections: Option[Set[String]], persistenceIdentifier: String) {
   val allReasons: List[PersistenceReason] =
     List(
       HasPersistenceIdentifier(persistenceIdentifier),
@@ -16,7 +16,7 @@ case class ImagePersistenceReasons(persistedRootCollections: List[String], persi
       IsIllustratorCategory,
       IsAgencyCommissionedCategory,
       HasLeases,
-      IsInPersistedCollection(persistedRootCollections),
+      IsInPersistedCollection(maybePersistOnlyTheseCollections),
       AddedToPhotoshoot,
       HasLabels,
       HasUserEdits
@@ -100,17 +100,20 @@ object HasLeases extends PersistenceReason {
   override val reason: String = "leases"
 }
 
-case class IsInPersistedCollection(persistedCollections: List[String]) extends PersistenceReason {
-  override def shouldPersist(image: Image): Boolean = {
-    // list of the first element of each collection's `path`, i.e all the root collections
-    val collectionPaths: List[String] = image.collections.flatMap(_.path.headOption)
-
-    // is image in at least one persisted collection?
-    (collectionPaths diff persistedCollections).length < collectionPaths.length
+case class IsInPersistedCollection(maybePersistOnlyTheseCollections: Option[Set[String]]) extends PersistenceReason {
+  override def shouldPersist(image: Image): Boolean = maybePersistOnlyTheseCollections match {
+    case None =>
+      image.collections.nonEmpty
+    case Some(persistedCollections) if persistedCollections.nonEmpty =>
+      (image.collections.flatMap(_.path) intersect persistedCollections.toList).nonEmpty
+    case _ => false
   }
 
-  override val query: Query = PersistedQueries.addedGNMArchiveOrPersistedCollections(persistedCollections)
-  override val reason: String = "persisted-collection"
+  override val query: Query = PersistedQueries.isInPersistedCollection(maybePersistOnlyTheseCollections)
+  override val reason: String = maybePersistOnlyTheseCollections match {
+    case None => "collection"
+    case Some(_) => "persisted-collection"
+  }
 }
 
 object AddedToPhotoshoot extends PersistenceReason {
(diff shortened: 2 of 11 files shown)


AndyKilmory:
Hi, Currently the BBC ingest processors add all agency images to the News collection - so swapping out the persisted collections functionality for 'isInCollection' option in the PersistedQueries functionality would prove problematic for the BBC at present as unused agency images would not get reaped - however if we could retain both options (i.e. not remove the option to use the persisted queries collection from config) - then it would be up to the organisation implementation to decide which option to employ in the ReapabilityCriteria. Thanks

twrichards:
> Hi, Currently the BBC ingest processors add all agency images to the News collection - so swapping out the persisted collections functionality for 'isInCollection' option in the PersistedQueries functionality would prove problematic for the BBC at present as unused agency images would not get reaped - however if we could retain both options (i.e. not remove the option to use the persisted queries collection from config) - then it would be up to the organisation implementation to decide which option to employ in the ReapabilityCriteria. Thanks

good to know @AndyKilmory. 
How about I rename the config option to something like `persistence.onlyTheseCollections` -- that way, not specifying would persist all collections?

AndyKilmory:
Hi @twrichards, That sounds fine Tom - so long as we can still work with a set of defined collections and incorporate it into our override reapability criteria. Thanks

common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfigWithElastic.scala
```diff
@@ -17,8 +17,6 @@ class CommonConfigWithElastic(resources: GridConfigResources) extends CommonConf
   val persistenceIdentifier = string("persistence.identifier")
   val queriableIdentifiers = Seq(persistenceIdentifier)
 
-  val persistedRootCollections: List[String] = stringOpt("persistence.collections") match {
-    case Some(collections) => collections.split(',').toList
-    case None => List(s"${staffPhotographerOrganisation} Archive")
-  }
+  val maybePersistOnlyTheseCollections: Option[List[String]] = stringOpt("persistence.onlyTheseCollections")
+    .map(_.split(',').toList)
```
andrew-nowak, line 21: understandable as the previous config modelled this as string with comma separated values too, but maybe we could take this opportunity of renaming the key to model properly as a list of strings instead? if only cause then you can choose to persist a collection that's named with a comma (weird but afaik a legit choice)

media-api/app/lib/ImagePersistenceReasons.scala
```diff
@@ -100,17 +100,22 @@ object HasLeases extends PersistenceReason {
...
-    (collectionPaths diff persistedCollections).length < collectionPaths.length
+case class IsInPersistedCollection(maybePersistOnlyTheseCollections: Option[List[String]]) extends PersistenceReason {
+  override def shouldPersist(image: Image): Boolean = maybePersistOnlyTheseCollections match {
+    case None =>
+      image.collections.nonEmpty
+    case Some(persistedCollections) =>
+      // list of the first element of each collection's `path`, i.e all the root collections
+      val collectionPaths: List[String] = image.collections.flatMap(_.path.headOption)
+      // is image in at least one persisted collection?
+      (collectionPaths diff persistedCollections).length < collectionPaths.length
```
andrew-nowak, line 111: looks like this config still only affects root collections? could we put a note to that effect somewhere, I'd say in our config doc but since it's patchy and not really looked at as is, maybe also/alternatively a comment by the code which parses the config key (ie. in CommonConfigWithElastic)

andrew-nowak, line 111:
```suggestion
      // list of the first element of each collection's `path`, i.e all the root collections
      val collectionPaths: List[String] = image.collections.flatMap(_.path.headOption)
      // is image in at least one persisted collection?
      (collectionPaths intersect persistedCollections).nonEmpty
```
I guess this is the same but I don't really understand the previous condition - this version seems clearer to read to me (but it's always worked fine in the previous condition, so not important!)

media-api/test/lib/ImagePersistenceReasonsTest.scala
```diff
@@ -36,6 +36,7 @@ class ImagePersistenceReasonsTest extends AnyFunSpec with Matchers {
     imgPersistenceReasons.reasons(imgWithLeases) shouldBe List("leases")
     val imgWithPersistedRootCollections = img.copy(collections = List(Collection.build(persistedRootCollections.tail, ActionData("testAuthor", now()))))
     imgPersistenceReasons.reasons(imgWithPersistedRootCollections) shouldBe List("persisted-collection")
+    //TODO consider a test for when all images in collections are persisted
```
andrew-nowak, line 39: sounds a good idea to me, worth doing before merge?

andrew-nowak (approved):
lgtm, a couple ideas for some more clarity but probably not blockers (I think just the decision whether to remodel the config, because that'd be painful to do in a follow up)

common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfigWithElastic.scala
twrichards: great shout, doing that now

twrichards: done, with (existing, albeit refactored) fallback to interpret comma based too

media-api/app/lib/ImagePersistenceReasons.scala
twrichards: great spot, I just tried seeing if we could actually match more than just the root collection, and it looks like the actual ES query is matching any part of the path AFAICT, so I'm tempted to adjust the reasons logic to better reflect the ES query, what do you think?

twrichards: I like this, have implemented thank you

andrew-nowak: oh great catch! Yes they should definitely match each other, either way (but I think your suggestion will probably be easier than making ES only query for the first item of the path)

media-api/test/lib/ImagePersistenceReasonsTest.scala
twrichards: done, thanks for the reminder

media-api/app/lib/ImagePersistenceReasons.scala
twrichards: I did have a look to see if we could match full collection paths (e.g. `foo / bar / baz`) in the ES query but it looked hectic, so I've updated the reasons logic, should be building now...

twrichards: I've also added a comment to this effect in the config class

andrew-nowak (approved):

PR #4291, opened 2024-06-25
[reaper] remove the recently added persistence reason 'collection' as existing 'persisted-collection' reason is fine

twrichards:
[reaper] remove the [recently added] (#4287) persistence reason 'collection' as it would require update to `grid-client` (and various consumers of the lib) for very little gain - existing 'persisted-collection' reason is fine

## Tested? Documented?
- [ ] locally by committer


Georges-GNM (approved):
Makes sense to me. 

Should we also remove the collection case from `kahuna/public/js/services/image-logic.js` [here](https://github.com/guardian/grid/blob/be80ea07749fc9c8c432e8303582595bdd7038b1/kahuna/public/js/services/image-logic.js#L64) ?

andrew-nowak (approved):

PR #4292, opened 2024-06-25
Adapt info panel height based on usePermissionsFilter config setting

AndyKilmory:
The usePermissionsFilter kahuna config setting (true/false default:false) introduces a second tier top bar into the UI to hold the new permissions filter control (along other controls). When the second tier top bar is visible the height of the info-panel was found to be set incorrectly causing the lower section of the panel to be inaccessible (e.g. the keywords portion). This change corrects that fault by setting the height of the info-panel based on the presence of the second tier top bar

Ensure that the whole of the info-panel is visible and accessible when the second tier top bar is displayed and also ensure that the info-panel is vertically aligned correctly when the second tier top bar is not visible

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-panels/gr-panels.css
+++ b/kahuna/public/js/components/gr-panels/gr-panels.css
@@ -58,6 +58,12 @@
     margin-top: 36px;
 }
 
-.gr-panel-height {
-    height: calc(100vh - 86px); /* 50px search bar, 36px toolbar */
+.gr-panel-height--single {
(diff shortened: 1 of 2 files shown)


Conalb97 (commented):
Looks good to me :)

paperboyo:
I can confirm this allows panel to be scrolled correctly. Thank for the fix!

image

twrichards (approved):
LGTM and tested by @paperboyo - thank you for the fix @AndyKilmory

PR #4295, opened 2024-07-03
all work relating to notification banner sensitivity

shannonmcgurk:
## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-notifications-banner/gr-notifications-banner.tsx
+++ b/kahuna/public/js/components/gr-notifications-banner/gr-notifications-banner.tsx
@@ -13,6 +13,7 @@ const NOTIFICATION_COOKIE = "notification_cookie";
 const cookie_age = 31536000;
 const checkNotificationsUri = window._clientConfig.rootUri + "/notifications";
 const checkNotificationsInterval = 60000; // in ms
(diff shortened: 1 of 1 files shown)


Conalb97 (commented):
Looks good to me! :)

KaliedaRik (commented):
Tested in local and can confirm the immediate functionality works as intended:
- transient lifespan banners disappear on user interaction
- session and persistent banners remain in view until user directly interacts with them

Two thoughts for consideration.

1. The Grid includes additional keyboard navigation, specifically: `PgUp`, `PgDown`, `Home` and `End` keys. For a consistent user experience, should these keys also be included alongside the arrow and esc keys to dismiss the banners?

2. During testing in local, I noticed that transient/session banners automatically reappear a short while after being dismissed - see screen capture below; this does not affect persistent banners (due to user interaction with those messages leading to the addition of a cookie to the user's device?). This may be due to testing on my local but, if not, perhaps some additional documentation and - if necessary - fixes for this sub-optimal user experience could be added to the PR?

https://github.com/guardian/grid/assets/5357530/7f242bb2-4593-4117-a6b1-6a2269f6ae56

PR #4297, opened 2024-07-08
Fixing bug where metadata-validity and image-cost-message controls don't update correctly when usage rights category is modified.

AndyKilmory:
## What does this change do?

There is currently a bug such that if a user is viewing an image and updates the usage rights category in the metadata panel and adds some restrictions text to the image on save of the updates to the rights category the warning panels metadata-validity and image-cost-message do not display - i.e. they are not getting updated correctly on changes in usage rights category and restrictions. (refreshing the screen will correct the display). The same problem occurs in reverse - if restrictions are removed from an image the visibility of the restrictions panel and the warning panel above it does not update (ie they remain visibile when they should be hidden). 

This PR corrects this issue by 'listening' on $rootScope for the 'images-updated' message and updating both controls to relfect the new image information.

Initial state:
Screenshot 2024-07-08 at 12 23 26

after restrictions removed and saved...

Screenshot 2024-07-08 at 12 24 15

after refresh...

Screenshot 2024-07-08 at 12 26 44

Ensure that the correct restriction information is displayed in the metadata panel after the restrictions text linked to the image has either been added or removed.

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-image-cost-message/gr-image-cost-message.js
+++ b/kahuna/public/js/components/gr-image-cost-message/gr-image-cost-message.js
@@ -8,19 +8,30 @@ import './gr-image-cost-message.css';
 export const module = angular.module('gr.imageCostMessage', [imageService.name]);
 
 module.controller('grImageCostMessage', [
-  'imageService',
+  '$rootScope', 'imageService',
 
-  function (imageService) {
+  function ($rootScope, imageService) {
     let ctrl = this;
 
     ctrl.$onInit = () => {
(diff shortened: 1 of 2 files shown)


Conalb97 (commented):
Looks good to me

paperboyo:
Tested in our TEST env, this fixes not only Info panel in Browser and Viewer, but also badges on thumbnails! Thank you!

KaliedaRik (approved):
Confirm that I can see the new functionality working as expected when I run the branch locally. Approved.

PR #4304, opened 2024-07-26
Stable content-disposition fallback filename field improves client side image caching

tonytw1:
## What does this change?

Remove unintentional cache buster on the thumbnail and preview image urls (filename ByteArray.toString is different for every instance of ByteArray even if they contain the same value).

Improves the client side caching of preview images.

☠️ Uses Image.id as the fallback Latin1 filename.
Assumes that Image.id is a safe Latin1 string (it's a UUID in recent Grid installs).

No change on thumbnails for Grid installs which use CloudFront as they use a different URL scheme which does not include content-disposition.

Screenshot 2024-08-06 at 13 50 15

Thumbnails and previews continue to render correctly.
File downloads continue to have sensible filenames.

Preview images begin to show client side cache hits for preview images (for up to 100 seconds).

Non CloudFront Grid installs begin to show client side cache hits for thumbnails.

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/common-lib/src/test/scala/com/gu/mediaservice/lib/aws/ContentDispositionTest.scala
+++ b/common-lib/src/test/scala/com/gu/mediaservice/lib/aws/ContentDispositionTest.scala
@@ -1,0 +1,28 @@
+package com.gu.mediaservice.lib.aws
+
+import com.gu.mediaservice.lib.elasticsearch.MappingTest
+import com.gu.mediaservice.model.Thumbnail
+import org.scalatest.funsuite.AnyFunSuiteLike
+import org.scalatest.matchers.should.Matchers.convertToAnyShouldWrapper
+
+class ContentDispositionTest extends AnyFunSuiteLike with ContentDisposition {
+
+  def useShortenDownloadFilename: Boolean = false
+
+  test("build content disposition header based on filename") {
+    val image = MappingTest.testImage
+
+    val header = getContentDisposition(image, Thumbnail)
+
+    header shouldBe """attachment; filename="abcdef1234567890"; filename*=UTF-8''filename%20%28abcdef1234567890%29.jpg"""
+  }
+
+  test("only use id for the latin1 fallback filename field") {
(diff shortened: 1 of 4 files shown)


common-lib/src/test/scala/com/gu/mediaservice/lib/aws/ContentDispositionTest.scala
```diff
@@ -0,0 +1,37 @@
...
+import org.scalatest.matchers.should.Matchers.convertToAnyShouldWrapper
+
+class ContentDispositionTest extends AnyFunSuiteLike with ContentDisposition {
+
+  test("build content disposition header based on filename") {
+    val image = withFilename(MappingTest.testImage, "imagefilename.jpg")
+
+    val header = getContentDisposition(image, Thumbnail, shortenDownloadFilename = false)
+
+    header shouldBe """attachment; filename="abcdef1234567890"; filename*=UTF-8''imagefilename%20%28abcdef1234567890%29.jpg"""
```
andrew-nowak, line 15: nitpicking, but I think it'd make sense if the latin1 fallback included the extension

```suggestion
    header shouldBe """attachment; filename="abcdef1234567890.jpg"; filename*=UTF-8''imagefilename%20%28abcdef1234567890%29.jpg"""
```

andrew-nowak (approved):
LGTM!

common-lib/src/test/scala/com/gu/mediaservice/lib/aws/ContentDispositionTest.scala
tonytw1: Tend to agree. The current implementation doesn't.
Deja vu about the extension not been available near this code but I'll look.

tonytw1: Added a new commit to implement this.

andrew-nowak (approved):
LGTM! thanks!

PR #4309, opened 2024-08-02
Cloudfront rounded expires

tonytw1:
## What does this change?

Round signed CloudFront URLs expires times to allow client side caching of CloudFront assets (ie. thumbnails and preview images).

The 1 second resolution of the `Expires` parameter in the CloudFront URLs has probably been acting as a cache buster.

When stacked on top of #4304 this PR should introduce some client side caching of thumbnails for Guardian users
which may make the UI appear more snappy.

This change does not change the current CloudFront to S3 origin behaviour.

Should cache:

- imgops preview image on single image view
  Immediate reload when revisiting a single image.

- thumbnails
   Should be most noticeable when clicking 'Back to search' after viewing a single image.

On the search view, this is a small incremental gain rather than a game changer.
It reduces the number of network fetches at the very end of the page render but that cluster is only 100 to 200ms in width to begin with.

Screenshot 2024-08-02 at 13 04 04

Screenshot 2024-08-02 at 13 11 51

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/RoundedExpiration.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/RoundedExpiration.scala
@@ -6,7 +6,7 @@ trait RoundedExpiration {
 
   // Round expiration time to try and hit the cache as much as possible
   // TODO: do we really need these expiration tokens? they kill our ability to cache...
-  def defaultExpiration: DateTime = roundDateTime(DateTime.now, Duration.standardMinutes(10)).plusMinutes(20)
+  def cachableExpiration: DateTime = roundDateTime(DateTime.now, Duration.standardMinutes(10)).plusMinutes(20)
 
   private def roundDateTime(t: DateTime, d: Duration): DateTime = t minus (t.getMillis - (t.getMillis.toDouble / d.getMillis).round * d.getMillis)
 
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/RoundedExpiration.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/RoundedExpiration.scala
@@ -1,0 +1,13 @@
+package com.gu.mediaservice.lib.aws
+
+import org.joda.time.{DateTime, Duration}
+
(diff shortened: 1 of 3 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/aws/RoundedExpiration.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import org.joda.time.{DateTime, Duration}
+
+trait RoundedExpiration {
+
+  // Round expiration time to try and hit the cache as much as possible
+  // TODO: do we really need these expiration tokens? they kill our ability to cache...
+  def cachableExpiration: DateTime = roundDateTime(DateTime.now, Duration.standardMinutes(10)).plusMinutes(20)
+
+  private def roundDateTime(t: DateTime, d: Duration): DateTime = t minus (t.getMillis - (t.getMillis.toDouble / d.getMillis).round * d.getMillis)
```
andrew-nowak, line 11: tangential at best to this PR so not blocking, but I stumbled over this for a while and I'm still not sure I fully get how it works, I came up with this alternative which does I think the same but a bit clearer

```
    val now = DateTime.now()
    val hour = now.getHourOfDay
    val minute = (now.getMinuteOfHour % 10) * 10
    val seconds = 0
    val millis = 0
    now.withTime(hour, minute, seconds, millis)
```

not flexible like the previous version, but then nothing else calls this so I'm relaxed about hardcoding 10 mins

andrew-nowak (approved):
lgtm! (reviewed last 3 commits only, I believe the others are shared with #4304 )

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/RoundedExpiration.scala
tonytw1: Yeah agree.
I'm tip toeing carefully around this without think about it too much because it's effective in a non CloudFront Grid install.
I'd need to visualise it properly but I expect it's grouping what would have been Date.now into the same block of length X beginning Y seconds from now.

PR #4311, opened 2024-08-05
adding ability to filter usage rights categories for standard users - so they can only select from a subset when uploading and editing

AndyKilmory:
## What does this change do?

It allows the range of rights categories that are available to stanard users during upload and edit to be filtered according to a list of exclusions added to the application config. The config setting is;

Screenshot 2024-08-05 at 19 21 06

For a standard user this will limit the range of available categories in the drop down during upload and edit;

_Edit_
Screenshot 2024-08-05 at 19 16 05

_Upload_
Screenshot 2024-08-05 at 19 16 33

For admin users these changes will have no impact;
Screenshot 2024-08-05 at 19 17 26

When searching by category via a search chip standard users will still be presented with the full list of available categories.

Also if the user is editing an image that is assigned to a category to which they do not usually have access this category will be added to the available list so that the drop down continues to operate in a consistent fashion.

The reason for this change to to try and restrict the number of images that get mis-classified by standard users during upload.

The list of restricted categories can be empty or can be left out of the config if it is not required and the system will continue to operate unaltered.

Ensure that the correct range of rights categories is visible for admin and standard users in line with the configured values. Ensure that the dropdown lists for rights categories operates as expected under all circumstances including use of multi-select in grid view.

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
@@ -204,6 +204,7 @@ abstract class CommonConfig(resources: GridConfigResources) extends AwsClientBui
    * and the singleton instance added to the list.
    */
   val applicableUsageRights: Seq[UsageRightsSpec] = configuration.get[Seq[UsageRightsSpec]]("usageRights.applicable")
+  val stdUserExcludedUsageRights = getStringSet("usageRights.stdUserExcluded")
 
   private def getKinesisConfigForStream(streamName: String) = KinesisSenderConfig(awsRegion, awsCredentials, awsLocalEndpoint, isDev, streamName)
 
--- a/common-lib/src/main/resources/application.conf
+++ b/common-lib/src/main/resources/application.conf
@@ -83,6 +83,12 @@ usageRights.applicable = [
   "com.gu.mediaservice.model.PublicDomain"
 ]
 
+# ----------------------------------------------------------------------------------------
+# List of usage right categories standard users cannot select whilst uploading or editing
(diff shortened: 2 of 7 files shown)


Conalb97 (commented):
Looks good to me

paperboyo:
Hi. This makes sense to me. Thanks for describing the behaviour in the edge cases – super-useful!

Which config will this be taken from (always worth mentioning)? `common.conf`? Also, could you add this [to here](https://github.com/guardian/grid/blob/main/docs/06-objects-of-interest/02-config.md), so we don’t lose it? (I should probably adjust PR template to mention that!)

Will have a play in our TEST…

AndyKilmory:
Hi, These config changes go in common.conf - we call it application config in our build. I'll update the documentation once we're happy with the final construct/format - have a task already in jira for this.

common-lib/src/main/scala/com/gu/mediaservice/lib/config/CommonConfig.scala
```diff
@@ -204,6 +204,7 @@ abstract class CommonConfig(resources: GridConfigResources) extends AwsClientBui
    * and the singleton instance added to the list.
    */
   val applicableUsageRights: Seq[UsageRightsSpec] = configuration.get[Seq[UsageRightsSpec]]("usageRights.applicable")
+  val stdUserExcludedUsageRights = getStringSet("usageRights.stdUserExcluded")
```
tonytw1, line 207: This new config field can be safely ignored by the Guardian and others because `getStringSet` will default to empty if the config field is non present 👍

paperboyo:
Functionality tested in our TEST environment: 👍 .

KaliedaRik (approved):
Happy to approve

PR #4313, opened 2024-08-07
Initial documenting test for RoundedExpiration

tonytw1:
## What does this change?

Parameterise the input so it can be tested.

Add some high level tests as an initial place to allow tinkering with this.

Shows that the current code generations a TTL in the range of 15 to 25 minutes.
Which feels sensible but could try slightly higher given the assets have a larger max-age and could still be in local caches after this time.

## Tested? Documented?
- [x] locally by committer


andrew-nowak (approved):
lgtm!

PR #4314, opened 2024-08-07
Fixes uploadStatuses/ link advertised by image loader

tonytw1:
## What does this change?

Removing the trailing slash makes the link work. This is generally consistent with other Grid advertised links.

This list all end point looks to be unused; it does not appear in recent access logs and seems to be unreferenced in the client code.

From api.media.(redacted PROD domain)
Follow the 'loader' link
Then follow the 'uploadStatuses' link.
Should get a non 404 JSON end point.

## Tested? Documented?
- [x] locally by committer


andrew-nowak (approved):
thanks!

PR #4315, opened 2024-08-08
T1811 - payable images bug

AndyKilmory:
## What does this change do?

There was a bug such that the 'Show Payable Images' checkbox control could get out of sync with the url nonFree value and the query results when clicking the Logo if the result of clicking the logo left the url in a state where it did not contain an entry for nonFree (true or false) - if that occurred and the user then started to type in the search bar they could in some instances see payable images in the results (and nonFree = true in url) but payable images control remained 'off'.

This PR corrects that bug and tidies up the way that the Logo click is conveyed to the interim filters control

Ensure that the 'payable images' control is always in sync with the query url and search results when clicking the logo.

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-permissions-filter/gr-permissions-filter.tsx
+++ b/kahuna/public/js/components/gr-permissions-filter/gr-permissions-filter.tsx
@@ -84,7 +84,6 @@ const PermissionsFilter: React.FC<PermissionsWrapperProps> = ({ props }) => {
   const [isChargeable, setIsChargeable] = useState(props.chargeable);
   const [selectedOption, setSelection] = useState(defPerms);
   const [currentIndex, setCurrentIndex] = useState(-1);
-  const [logoClick, setLogoClick] = useState<LogoClickProperties>({logoClick: false, showPaid:false});
 
   const autoHideListener = (event: any) => {
     if (event.type === "keydown" && event.key === "Escape") {
@@ -97,15 +96,8 @@ const PermissionsFilter: React.FC<PermissionsWrapperProps> = ({ props }) => {
   };
 
   const handleLogoClick = (event: LogoClickEvent) => {
-    setLogoClick({logoClick: true, showPaid: event.detail.showPaid});
+    setIsChargeable(event.detail.showPaid);
   };
-
-  useEffect(() => {
-    if (logoClick.logoClick) {
-      setIsChargeable(logoClick.showPaid);
-      setLogoClick({logoClick: false, showPaid: logoClick.showPaid});
-    }
-  }, [logoClick]);
 
   const handleQueryChange = (event: QueryChangeEvent ) => {
     const newQuery = event.detail.query ? (" " + event.detail.query) : "";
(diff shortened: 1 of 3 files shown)


Conalb97 (commented):
Looks good to me

kahuna/public/js/components/gr-permissions-filter/gr-permissions-filter.tsx
```diff
@@ -85,15 +96,19 @@ const PermissionsFilter: React.FC<PermissionsWrapperProps> = ({ props }) => {
     }
   };
 
-  const handleQueryChange = (e: any ) => {
-    const newQuery = e.detail.query ? (" " + e.detail.query) : "";
+  const handleLogoClick = (event: LogoClickEvent) => {
+    setLogoClick({logoClick: true, showPaid: event.detail.showPaid});
+  };
```
andrew-nowak, line 101: I don't follow the need for this logoClick state, as far as I can tell it's only used to trigger the useEffect block below, which only sets the isChargeable state. I'm pretty sure you can replace all of that by directly setting the isChargeable state

```suggestion
  const handleLogoClick = (event: LogoClickEvent) => {
    setIsChargeable(event.detail.showPaid);
  };
```

kahuna/public/js/components/gr-permissions-filter/gr-permissions-filter.tsx
```diff
@@ -85,15 +96,19 @@ const PermissionsFilter: React.FC<PermissionsWrapperProps> = ({ props }) => {
...
-    const logoClick = window.sessionStorage.getItem("logoClick") ? window.sessionStorage.getItem("logoClick") : "";
-    if (logoClick.includes("logoClick")) {
-      setIsChargeable(false);
-      window.sessionStorage.setItem("logoClick", "");
+  useEffect(() => {
+    if (logoClick.logoClick) {
+      setIsChargeable(logoClick.showPaid);
+      setLogoClick({logoClick: false, showPaid: logoClick.showPaid});
     }
+  }, [logoClick]);
```
andrew-nowak, line 108:
```suggestion
```

kahuna/public/js/components/gr-permissions-filter/gr-permissions-filter.tsx
```diff
@@ -74,6 +84,7 @@ const PermissionsFilter: React.FC<PermissionsWrapperProps> = ({ props }) => {
   const [isChargeable, setIsChargeable] = useState(props.chargeable);
   const [selectedOption, setSelection] = useState(defPerms);
   const [currentIndex, setCurrentIndex] = useState(-1);
+  const [logoClick, setLogoClick] = useState<LogoClickProperties>({logoClick: false, showPaid:false});
```
andrew-nowak, line 87:
```suggestion
```

AndyKilmory: Will test suggestion and confirm that it meets requirements - Thanks

kahuna/public/js/components/gr-permissions-filter/gr-permissions-filter.tsx
AndyKilmory: Have simplified code inline with suggestion

kahuna/public/js/components/gr-permissions-filter/gr-permissions-filter.tsx
AndyKilmory: Have simplified code inline with suggestion

andrew-nowak (approved):

PR #4317, opened 2024-08-12
IMAEGEDAM-1830: showPaid permission prevents standard users from accessing sendToPhotosales functionality

Conalb97:
## What does this change?

This adds a layer of authorisation on top of the send to photosales functionality. Only users that have enhanced access rights (in the BBC case these would be archivists) can access the send to photosales feature.

- Set the sendToPhotoSales feature flag to true. 
- Log in as a standard user and select an image. The 'send to photosales' button should not be present in the menu that appears.
- Log in as a user with enhanced access rights and select an image. The 'send to photosales' button should now be present in the menu that appears. 

Only users with enhanced rights have access to the send to photosales functionality. 

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/search/results.html
+++ b/kahuna/public/js/search/results.html
@@ -83,7 +83,7 @@
 
                 <div class="results-toolbar-item results-toolbar-item--right"
                     tabindex="0"
-                    ng-if="ctrl.showSendToPhotoSales()">
+                    ng-if="ctrl.showSendToPhotoSales() && ctrl.showPaid">
                     <a id="send-to"
                        style="display: contents;"
                        ng-class="{'batch-archive__button--disabled': ctrl.selectionCount >= 45}"
--- a/kahuna/public/js/search/results.js
+++ b/kahuna/public/js/search/results.js
@@ -418,6 +418,11 @@ results.controller('SearchResultsCtrl', [
         return [validImages, invalidImages];
       };
 
+      ctrl.showPaid = undefined;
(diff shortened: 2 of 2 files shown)


paperboyo:
Not actually testing anything, but did you possibly mean `edit_metadata` and not `show_paid`? The latter is used at the Guardian **only** to control the initial state of Free to use checkbox (to make it unchecked if permission is ON).

AFAIK, at the BBC, you don’t use `show_paid` (at least your version of Free to use – a reversed Show payable images –doesn’t observe `show_paid` permission on our TEST env)

Conalb97:
> Not actually testing anything, but did you possibly mean `edit_metadata` and not `show_paid`? The latter is used at the Guardian **only** to control the initial state of Free to use checkbox (to make it unchecked if permission is ON).
> 
> AFAIK, at the BBC, you don’t use `show_paid` (at least your version of Free to use – a reversed Show payable images –doesn’t observe `show_paid` permission on our TEST env)

Okay so having gone over this with @AndyKilmory we believe using showPaid is the correct approach. We're using showPaid as a universal setting that is independent of image, to map to user type. 

Previously we had a bug in the payable images control initialisation that meant it wasn't mapping correctly to showPaid, which is probably it's not working for you on your test environment. 

Andy has a fix for this bug in this PR: https://github.com/guardian/grid/pull/4315

With this in place you should be able to see the payable images control respond correctly based on user type.

andrew-nowak (approved):

PR #4318, opened 2024-08-13
Fix issues leading to scrollTo not working with interimFilters on. 

AndyKilmory:
## What does this change do?

With the introduction of the interim filters control and the second layer in the top-bar we introduced a bug such that the ability to return to the same scroll location in a list of images had been lost - when the user returns to the grid having opened the image detail view for an image they will be returned to the top of the view/list rather than the scroll location they had come from.

The problem has two elements to it;

1) The inclusion of 'ng-if' in the wrapper element for the interim filters causes the refresh of the $scope when we return to the grid view form the image view - which triggers a $destroy event - this in turn resets the topPosition in the scroll-position service to 0 which means the scroll never gets reset to the correct value.

2) The ambiguity surrounding the nonFree parameter in the context causes the currentContext to evaluate as different from the originalContext aas the interimFilters control includes the 'Payable Images' checkbox which forces a value onto the nonFree parameter. nonFree = undefined and nonFree = false both mean the same thing so new code ensures that they will evaluate as equivalent.

Ensure that when a user scrolls down an image list (including seach options if required) and then selects an image - when they then return to the list view from the image view they should return to the same scroll location and the search parameters are retained.

All the main UI controls should continue to operate unaltered and all other screen transition should not be impacted (the navigation to 'My Uploads' and back to the list view should also retain the scroll location).

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/search/results.js
+++ b/kahuna/public/js/search/results.js
@@ -702,7 +702,10 @@ results.controller('SearchResultsCtrl', [
         });
 
         $scope.$on('$destroy', () => {
-            scrollPosition.save($stateParams);
+            // only save scroll position if we're destroying grid scope (avoids issue regarding ng-if triggering scope refresh)
+            if (0 < $scope.ctrl.images.length) {
+              scrollPosition.save($stateParams);
+            }
             freeUpdatesListener();
             freeImageDeleteListener();
             scopeGone = true;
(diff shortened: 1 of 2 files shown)


paperboyo:
Testing functionality on our TEST (where we have your Interim Filters ON): 👍.

AndyKilmory:
thanks

andrew-nowak (approved):

PR #4321, opened 2024-08-19
fixing myUploads bugs

AndyKilmory:
## What does this change do?

The BBC introduced a new set of controls - permissions filter, show payable images, sort control and my uploads - written in react to display in a second tier toolbar at the top of the page when the kahuna config variable usePermissionsFilter is set yo true. It's been found that problems with the interop between AngularJS and React have introduced some regressions to the system functionality (along with some missed specifications). This PR corrects these regressions whilst ensuring continuing operation when permisisons filter is off.

Areas impacted:
- clicking the logo button should clear the filter and set the 'Show Payable Images' control in-line with user type - on for Archivists and off for standard users
- selecting My Uploads control show turn show payable images on so that any user uploaded images without a rights category are also shown. Unchecking the control should turn show payable images off (this was missed specification and so is new functionality)
- selecting 'View All Your Uploads' from the uploads page should return to the search page and set 'show payable images' to on along with the 'my uploads' control checked
- use of logo or search from uploads page should return to search page with clear search (this is the default url /search)
- use of back button from uploads page of image view should return to search page with search retained.
- when pasting urls into a browser tab all controls should be set to match the values in the url

View of permissions filter layout:
Screenshot 2024-08-29 at 10 08 13

Behaviour of system with permissions filter off should be unaltered. View of permissions filter off;
Screenshot 2024-08-29 at 10 13 56

Ensure that all the impacted functional areas behave as expected and described.

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/dev/nginx-mappings.yml.template
+++ b/dev/nginx-mappings.yml.template
@@ -22,7 +22,7 @@ mappings:
   - prefix: media-usage
     port: 9009
   - prefix: media-collections
-    port: 9010
+    port: 9013
   - prefix: media-auth
     port: 9011
   - prefix: media-leases
--- a/dev/nginx-mappings.yml.template
+++ b/dev/nginx-mappings.yml.template
@@ -22,7 +22,7 @@ mappings:
   - prefix: media-usage
     port: 9009
   - prefix: media-collections
-    port: 9013
+    port: 9010
   - prefix: media-auth
     port: 9011
   - prefix: media-leases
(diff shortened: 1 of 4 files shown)


twrichards (approved):
looks good to me.
tested in our TEST environment, with the extra bar turned off (like in our PROD environment) to check for regressions - can't see any so going ahead and approving

PR #4323, opened 2024-08-29
improves the error flow in updateUploadStatusTable

andrew-nowak:
## What does this change?

Currently we recover from the uploadAttempt before we continue on the happy path, which means that the future is turned from a failure to a success, but holding the exception from the previous failure, which of course fails the pattern match, leaving us with a very confusing stack trace. Instead handle both success and failure cases by using a `transformWith`, which provides the bonus that we can handle the table update failure cases together

## Tested? Documented?
- [ ] locally by committer


bryophyta (approved):
I haven't tested it at runtime, but the code changes look good to me 👍

PR #4326, opened 2024-09-05
make an effort to send all metrics before play app shuts down

andrew-nowak:
## What does this change?

Improvement to #3656 - curretly it (and previous implementation) hold metrics for up to a minute before sending to cloudwatch. if the app were to terminate during that minute, then all the held metrics would be lost. That's a shame, so instead add a shutdown lifecycle hook that will make a best effort to send all the buffered metrics (and then stop receiving new metrics to be sent)

## Tested? Documented?
- [ ] locally by committer


twrichards (approved):
nice - good spot/call

PR #4327, opened 2024-09-10
Fixing cannot download padding

shannonmcgurk:
## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-downloader/gr-downloader.html
+++ b/kahuna/public/js/components/gr-downloader/gr-downloader.html
@@ -43,14 +43,14 @@
             <gr-icon-label gr-icon="file_download">Download selected crop</gr-icon-label>
         </a>
     </button>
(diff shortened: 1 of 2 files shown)


Conalb97 (commented):
Looks good to me!

kahuna/public/js/components/gr-downloader/gr-downloader.html
```diff
@@ -47,9 +47,8 @@
           aria-label="Cannot download">
         <gr-icon-label
             gr-icon="file_download"
-            class="download-icon-label"
-            gr-tooltip="Downloading non-free images requires a lease"
-        >
+            class="cannot-download-icon-label"
```
andrew-nowak, line 50: this works, but did you consider using the `side-padded` class that the other buttons use to add padding? it would mean one less CSS rule to maintain for just this version of the button

PR #4328, opened 2024-09-16
Ensure scroll posotion resets to top when logo clicked

AndyKilmory:
## What does this change?

This PR fixes a regression introduced through a combination of PRs that removed ambiguity regarding the 'nonFree' parameter in the current search context. Previously when the logo was clicked the search returned to the top of the page i.e. positionTop = 0 - the code was recording the position of the scroll when the logo was clicked but the setting of nonFree to undefined ensured that the contexts did not match and thus the scroll was no reset to its previous location.

The introduction of typescript react contros has made the use of nonFree = undefined problematic as typescript won't accept that value for a boolean so we have had to do introduce some code changes to ensure both scrolling and page navigation work correctly by removing the ambiguity of nonFree = false/undefined - in some areas the code relied on these being equivalent and in others it required them to be treated as different - e.g. resetting scroll position.

The recent PR to fix issues around page to page navigation and providing correct nonFree values to typescript controls removed further ambiguity around the nonFree parameter such that during logo click process the contexts evaluate as equal and thus the scroll position was reset to its previous value rather than being left at page-top.

The removal of the ambiguity around nonFree is required for other navigations to operate correctly with the typescript controls and so this PR introduces a new function for the scroll service to allow user to say 'reset to top' - this function is called during the logo click process and ensure that positionTop is set to 0 and thus when the contexts evaluate to equal the reset scroll position is now top.

Ensure that on logo click the page resets to the top of the image list.

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/search/index.js
+++ b/kahuna/public/js/search/index.js
@@ -10,6 +10,7 @@ import './results';
 import '../preview/image';
 import '../lib/data-structure/list-factory';
 import '../lib/data-structure/ordered-set-factory';
+import '../services/scroll-position';
 import '../components/gr-top-bar/gr-top-bar';
 import '../components/gr-info-panel/gr-info-panel';
 import '../components/gr-collections-panel/gr-collections-panel';
@@ -35,6 +36,7 @@ export var search = angular.module('kahuna.search', [
     'kahuna.search.query',
     'kahuna.search.results',
     'kahuna.preview.image',
+    'kahuna.services.scroll-position',
     'data-structure.list-factory',
     'data-structure.ordered-set-factory',
     'gr.topBar',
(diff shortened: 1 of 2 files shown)


twrichards:
building for our test environment now, hopefully review and merge this afternoon...

twrichards (approved):
code LGTM and @paperboyo tested functionality - so approving, but will merge on Monday - thanks for the work @AndyKilmory

paperboyo:
Hi. Testing it in our TEST env, I can still see clicking Home not always resetting the app (not scrolling to top).

Steps:
1. reset via Home
2. scroll down
3. click on image
4. click Home

If you can’t reproduce, try to do it quickly?

AndyKilmory:
The second stated case of the problem probably requires an alternative fix as the navigation enacted via the logo click from the image edit page uses an alternate mechanism to return to the home page - i.e. it is simply a url re-direct rather then use of javascript routine to reset the current page. It again relies on the fact that nonFree can be true, false or undefined and the fact that in some circumstances false and undefined need to be evaluated as true - so the return the same search results and in some case need to evaluate as different - when considering context equivalence when deciding whether to reset the assgined scroll position or not. This situation cannot be supported with the new react controls written in typescript which require a proper boolean value for this parameter. Will look into a separate fixx for image edit and other pages (e.g. upload) next week.

twrichards:
@AndyKilmory I need to re-read all the related code to be sure, but I'm fairly sure we can represent the the three states (true, false or undefined) with typescript, no problem. I don't want to block this fix for now, so I'll merge but I'll try to follow-up with you about how to represent what you want and ensure the 'fix' is comprehensive across all the scenarios discussed.

AndyKilmory:
Hi Tom, thanks for merging - I'm sure we can 'fix' to handle the three state issue for nonFree parameter - I just don't really like the design whereby we end up with two values for the context (one with nonFree = false and the other with nonFree = undefined) that are in some instances effectively equivalent (search reaults) and in others need to be treated as unequal (scroll position) for the system to function as required - just doesn't sit well that you have 2 objects that will be equivalent in some case and different in others

PR #4333, opened 2024-09-24
Ensures that scroll position is reset to top when Logo is clicked when no search criteria are entered

AndyKilmory:
## What does this change?

When the Logo is clicked in the top left of the screen it is expected that all search criteria will be wiped out and you will return to the grid view with the scroll position set to zero. This was not happening in some cases when the user had navigated to the MyUploads or Image Edit pages without entering any search or filtering criteria having scrolled through the image list.

The problem was that the scroll position was getting set correctly when the user navigated away from the grid view but on return the original context and the new context were evaluating to equal which caused the scroll to reset to its previous position - this has come about due to changes in the handling of the nonFree parameter in the context to better support the BBC Specific Payable Images control and requirement for BBC users to be able to toggle the control on and off.

The change has introduced a new function in the scroll position control - resetToTop() - which is called in all instances where the logo is clicked. This ensures the scroll always resets.

In addition this PR also introduces a fix to a BBC specific edge case where the react control has not completely initialised on return to the Grid view and does not respond/consume the 'setPayableImages' event triggered in the AngularJS code. To support this case an alternate approach has been introduced to store the occurence of the payableImages event as a simple message in the sessionStorage which can be consumed when the React control initialisation does get processed. The event launch is retained to deal with the case where the 'race' does finish in the other order and the react control is initialised before the user permissions have been correctly assessed.

Minor changes have also been introduced on the MyUploads page to make the 'Back to Search' behave in the same way it does on the Image Edit page - making all such navigation consistent.

Ensure that whenever the Logo is clicked the search and filtering criteria are reset to the default and the scroll position resett to the top of the screen

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-permissions-filter/gr-permissions-filter.tsx
+++ b/kahuna/public/js/components/gr-permissions-filter/gr-permissions-filter.tsx
@@ -138,10 +138,8 @@ const PermissionsFilter: React.FC<PermissionsWrapperProps> = ({ props }) => {
     if (sessionStorage) {
       const defNonFree = sessionStorage.getItem("defaultIsNonFree");
       const payEvent = sessionStorage.getItem("payableImagesEvent");
-      console.log("SessionStorage values defNonFree = " + defNonFree + " payEvent = " + payEvent);
       if ("true" == payEvent) {
-        console.log("Setting Chargeable by PayEvent");
-        sessionStorage.setItem("payableImagesEvent", "false");
+        sessionStorage.setItem("payableImagesEvent", 'false');
         setIsChargeable("true" == defNonFree);
       }
     }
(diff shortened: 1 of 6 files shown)


Conalb97 (commented):
Looks good to me

andrew-nowak:
one bug remains:

when user has the show_paid permission (ie. initial search results include both free and non-free images, `?nonFree=true` is added to the url etc.), opens an image then clocks the logo button, "free to use only" is unticked but `?nonFree=true` is missing from url and only free images are shown -- pressing the button should restore the same initial state as first page load.

AndyKilmory:
I'm not seeing that behaviour on my local dev copy or on our INT environment where I've deployed the branch - I see the correct ui setting and the correct url - will take another look

AndyKilmory:
Short video showing behaviour on our INT environment with permissions filters off...

https://github.com/user-attachments/assets/95a03c62-4dd8-4d1e-9421-b3d165f192a7

AndyKilmory:
Also works when we have permissions filters switched on. Could you drop in a video showing the behaviour you are seeing? Thanks

andrew-nowak:
[nonfree.webm](https://github.com/user-attachments/assets/a5a5dc37-e5a9-4250-aa14-32ad89005137)

From your recording, it looks like you don't open an image before using the logo

AndyKilmory:
I believe latest commit fixes the reported issue - hopefully now okay in all use cases for Guardian config and BBC config (most of the problems relate to getting consistent behaviour with the react/AngularJS interop and getting all event reponses to align correctly).

PR #4337, opened 2024-09-26
FIX `getFeatureSwitchActive` to fallback to default value (not just cookie)

twrichards:
When working on #4336 I noticed the default wasn't pulling through to the client


andrew-nowak (approved):
untested

PR #4340, opened 2024-09-27
IMAGEDAM-1982: Prevent the resend of capture images and restrict icons

Conalb97:
## What does this change?

This change prevents users from sending images that were received from Photo Sales **back** to Photo Sales - this is enabled through the same modal/notification mechanism that is already applied to images that have an 'Added to Photo Sales' usage. Additionally, the 'sales' icon (for BBC Photo Sales) will now also display for images **received** from Photo Sales, whereas before it was only present for images that were **sent** to Photo Sales. Finally it ensures that only archivist users will be able to see the sales icon, standard users will not be able to. 

### Users can not send images back to BBC Photo Sales:
- Apply the 'uploader' chip and enter 'Capture_AutoIngest'
- The query will be executed and the UI will refresh
- Select an image from the returned images 
- Select the 'Send to Photo Sales' button
- A notification banner appears informing the user that the image cannot be sent as it already exists in BBC Photo Sales

### The sales icon is present for images received from Photo Sales:
- Apply the 'uploader' chip and enter 'Capture_AutoIngest'
- The query will be executed and the UI will refresh
- Confirm that the sales icon is present on the returned images

As a user without elevated permissions and with the showSendToPhotoSales flag set to true:
### The sales icon is not present on any images:
- Apply the 'uploader' chip and enter 'Capture_AutoIngest'
- The query will be executed and the UI will refresh
- Confirm that the sales icon is not present on any images
- Remove the chip
- Apply the 'usage@platform' chip and select 'Added to Photo Sales'
- The query will be executed and the UI will refresh
- Confirm that the sales icon is not present on any images

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/kahuna/public/js/search/results.js
+++ b/kahuna/public/js/search/results.js
@@ -428,26 +428,6 @@ results.controller('SearchResultsCtrl', [
           }
           (syndicationExists === true ? invalidImages : validImages).push(image);
         });
-        //
-        //
-        // images.forEach( (image) => {
-        //   if (image.data.uploadedBy === "Capture_AutoIngest") {
-        //     invalidImages.push(image);
-        //   } else {
-        //     if (image.data.usages.data.length === 0) {
-        //       validImages.push(image);
-        //     } else {
-        //       let syndicationExists = false;
-        //       for (const usage of image.data.usages.data) {
-        //         if (usage.data.platform === "syndication") {
-        //           syndicationExists = true;
-        //           break;
-        //         }
-        //       }
-        //       (syndicationExists === true ? invalidImages : validImages).push(image);
-        //     }
-        //   }
-        // });
         return [validImages, invalidImages];
       };
 
--- a/kahuna/public/js/search/results.js
+++ b/kahuna/public/js/search/results.js
@@ -400,26 +400,54 @@ results.controller('SearchResultsCtrl', [
         };
 
       const validatePhotoSalesSelection = (images) => {
-        let validImages = [];
-        let invalidImages = [];
-        images.forEach( (image) => {
-          if (image.data.uploadedBy === "Capture_AutoIngest") {
-            invalidImages.push(image);
-          } else {
+        const validImages = [];
+        const invalidImages = [];
+
+        const filteredImages = images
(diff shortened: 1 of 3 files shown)


AndyKilmory:
Looks good to me

kahuna/public/js/search/results.js
```diff
@@ -400,21 +400,33 @@ results.controller('SearchResultsCtrl', [
...
+              return false;
+            }
+            return true;
+          });
+        filteredImages.forEach((image) => {
+          let syndicationExists = false;
+          for (const usage of image.data.usages.data) {
+            if (usage.data.platform === "syndication") {
+              syndicationExists = true;
+              break;
```
andrew-nowak, line 426: This function is really complex, and I think can be simplified a lot...

* I'd really caution against performing side effects like `.push()`ing while inside a filter function; it _does work_, but if not reading closely I'd assume that a filter doesn't change anything, and is only testing values against a predicate.
* The initial filter for `validImages` (testing if usages is 0) isn't really necessary; if you skipped that you should get the same result because the loop inside the later `forEach` wouldn't run, so `syndicationExists` would stay as `false` and the image would be added to `validImages`.
* The loop inside the `forEach` could be replaced by using a `.some`, which behaves the same but saves a couple of lines and avoids needing to track the `syndicationExists` variable.

Take a look at this I wrote, which I think would produce the same results, and see if it also looks clearer to you?

```js
const imageHasSyndicationUsage = (image) => {
  return image.data.usages.data.some(usage =>
    usage.data.platform === 'syndication'
  );
};

const validatePhotoSalesSelection = (images) => {
  const validImages = [];
  const invalidImages = [];

  images.forEach(image => {
    if (image.data.uploadedBy === 'Capture_AutoIngest' || imageHasSyndicationUsage(image)) {
      invalidImages.push(image);
    } else {
      validImages.push(image);
    }
  });

  return [validImages, invalidImages];
};

```

Conalb97: Yeah that looks a lot better, cheers!

andrew-nowak (approved):

PR #4343, opened 2024-10-01
WIP add a `shouldShowCropGuttersIfApplicable` query param

twrichards:
… and only show crop gutters if `true` (and other conditions for showing the gutters)

## Tested? Documented?
- [ ] locally by committer


bryophyta (approved):
haven't run this, but I've read through the code and the changes look good to me 👍

andrew-nowak (approved):
looks reasonable to me. I think if this was going to be long lived I'd prefer it be more flexible and configurable like `customRatio` is, so grid doesn't even have to be aware of what "if applicable" is, but for now this works great!

PR #4344, opened 2024-10-03
fix dragging over cropping gutters

twrichards:
set `pointer-events: none` to ensure dragging the crop box isn't prevented when the mouse is over the gutters (introduced in https://github.com/guardian/grid/pull/4331). Thanks @paperboyo for the spot.


paperboyo (approved):
Thank you for the speedy fix! I tested in Firefox and it works.

PR #4345, opened 2024-10-03
Scalaz upgrade & simplifications

andrew-nowak:
## What does this change?

Redo of #3671 - removing outdated scalaz syntaxes.

This time around I've left in the NonEmptyList usages - they are serving a purpose, although with updated scalaz they need some more explicit conversions, and there were a few places where I could simplify their usage.

All tests pass, and grid continues to function as normal when deployed to TEST

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/media-api/app/controllers/MediaApi.scala
+++ b/media-api/app/controllers/MediaApi.scala
@@ -477,8 +477,7 @@ class MediaApi(
     SearchParams.validate(searchParams).fold(
       // TODO: respondErrorCollection?
       errors => Future.successful(respondError(UnprocessableEntity, InvalidUriParams.errorKey,
-        // Annoyingly `NonEmptyList` and `IList` don't have `mkString`
-        errors.map(_.message).list.toList.mkString(", ")) //.reduce(_+ ", " +_), List(searchLink))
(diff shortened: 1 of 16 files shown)


media-api/app/controllers/MediaApi.scala
```diff
@@ -478,7 +478,7 @@ class MediaApi(
       // TODO: respondErrorCollection?
       errors => Future.successful(respondError(UnprocessableEntity, InvalidUriParams.errorKey,
         // Annoyingly `NonEmptyList` and `IList` don't have `mkString`
-        errors.map(_.message).list.reduce(_+ ", " +_), List(searchLink))
+        errors.map(_.message).list.toList.mkString(", ")) //.reduce(_+ ", " +_), List(searchLink))
```
bryophyta: if I'm reading this right, can we delete the comment on ln.480 and remove the trailing comment on ln.481?

bryophyta: ah nvm this was done in the subsequent commit 👍

bryophyta (approved):
We've talked through some of the more involved changes, and the code looks good to me 👍

PR #4346, opened 2024-10-07
Remove the ImageType type

andrew-nowak:
## What does this change?

No-op change. The `ImageType` object contained a couple of hardcoded values for "Photograph" etc. - wrongly implying that these are the only valid values! Instead, remove this type and pass the used values directly.

While here, I noticed that a few of the tests were actually not running their assertions; the `be` function, with its argument on the trailing line was in a position that meant it was being _referenced_ rather than _invoked_. That revealed a couple of failing tests, that needed to be updated to match the current version of the code.

Code review should be sufficient.

## Tested? Documented?
- [ ] locally by committer


twrichards (approved):
Changes LGTM and brilliant spot/catch on those tests

PR #4347, opened 2024-10-07
CropSpec metadata can handle optional non String values

tonytw1:
## What does this change?

CropSpec metadata currently persists Option None as literal string "None" .
This does not look intentional and fails for optional non string values.

ie. The current code round trips CropSpec fields like so:

`Some("Hello")` -> `Some("Hello")`
`Some(1)` -> `Some(1)`
`None` -> `Some("None")`
`None[Int]` -> Fails with cannot parse `"None"` as an Int ❌ 

After this change, the following happens:

`Some("Hello")` -> `Some("Hello")`
`Some(1)` -> `Some(1)`
`None` -> `None`
`None[Int]` -> `None[Int]`

- Pull the serialization method into a new trait for testing.
- Change the code so that Option None round trips to None. More verbose implementation to spell out the steps.

This change is backwards compatible but makes no attempt to clean the existing data.
ie. optional None Strings which have already been persisted as `"None"` will continue to be read as Some("None").

This issue was noticed when trying to use an optional non String value in the CropSpec (ie. an Int field).

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/cropper/app/lib/CropSpecMetadata.scala
+++ b/cropper/app/lib/CropSpecMetadata.scala
@@ -19,17 +19,10 @@ trait CropSpecMetadata {
       "height" -> dimensions.height,
       "aspect-ratio" -> r)
 
-    val nonEmptyMetadata = metadata.filter {
-      case (_, None) => false
-      case _ => true
+    metadata.collect {
+      case (key, Some(value)) => key -> value.toString
+      case (key, value) if value != None => key -> value.toString
     }
-
-    val flattenedMetadata = nonEmptyMetadata.collect {
-      case (key, Some(value)) => key -> value
-      case (key, value) => key -> value
-    }.view.mapValues(_.toString).toMap
-
-    flattenedMetadata
   }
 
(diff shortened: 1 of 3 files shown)


tonytw1:
Rebased.

cropper/app/lib/CropSpecMetadata.scala
```diff
@@ -0,0 +1,57 @@
...
+
+    val nonEmptyMetadata = metadata.filter {
+      case (_, None) => false
+      case _ => true
+    }
+
+    val flattenedMetadata = nonEmptyMetadata.collect {
+      case (key, Some(value)) => key -> value
+      case (key, value) => key -> value
+    }.view.mapValues(_.toString).toMap
```
andrew-nowak, line 30:
```suggestion
    val flattenedMetadata = metadata.collect {
      case (key, Some(value))            => key -> value.toString
      case (key, value) if value != None => key -> value.toString
    }

    flattenedMetadata
```

andrew-nowak (approved):
looks great! I've got just one suggestion to simplify slightly

andrew-nowak:
Thanks @tonytw1 !

PR #4349, opened 2024-10-08
upgrade to scanamo v2.0.0 latest

andrew-nowak:
## What does this change?

Upgrades scanamo to v2.0.0
First component switched to AWS SDK v2

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/SoftDeletedMetadataTable.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/SoftDeletedMetadataTable.scala
@@ -1,46 +1,46 @@
 package com.gu.mediaservice.lib.metadata
 
 import com.amazonaws.services.dynamodbv2.model.BatchWriteItemResult
-import com.gu.mediaservice.lib.aws.DynamoDB
 import com.gu.mediaservice.lib.config.CommonConfig
 import com.gu.mediaservice.model.ImageStatusRecord
-import com.gu.scanamo._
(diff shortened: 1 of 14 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/metadata/SoftDeletedMetadataTable.scala
```diff
@@ -1,46 +1,46 @@
...
   }
 
   def setStatus(imageStatus: ImageStatusRecord)(implicit ex: ExecutionContext) = {
-    ScanamoAsync.exec(client)(softDeletedMetadataTable.put(imageStatus))
+    ScanamoAsync(client).exec(softDeletedMetadataTable.put(imageStatus))
   }
 
-  private def extractUnprocessedIds(results: List[BatchWriteItemResult]): List[String] =
-    results.flatMap(_.getUnprocessedItems.values().asScala.flatMap(_.asScala.map(_.getPutRequest.getItem.get("id").getS)))
-
```
andrew-nowak, line 26: scanamo no longer returns the list of unprocessed items; it looks like it internally will repeat until all items have been processed, so we should no longer have to care about it in our application code?! 
https://github.com/scanamo/scanamo/blob/3a68697525c8dde1e1aee757d2d1f729511106b7/scanamo/src/main/scala/org/scanamo/ScanamoFree.scala#L56-L64

bryophyta (approved):
I'm not really familiar with the parts of the Grid that this touches, but with that caveat the code changes here look good to me 👍

PR #4357, opened 2024-10-28
add defaultCropType param

dblatcher:
## What does this change?

Adds a `defaultCropType` param to kahuna's search, image and crop pages and validates and stores it in the same way as the existing `cropType` param.

The `defaultCropType` is used to set what type of crop to initially set the on the crop window. If the param is not set, this will default to 'landscape', (being defined as `defaultCrop`) as currently

 

run locally, open:
https://media.local.dev-gutools.co.uk/search?nonFree=true&defaultCropType=freeform
choose an image, click the 'crop' button
the crop UI should have "freeform" selected instead of "landscape"

The purpose is to allow application that use a Grid window to specify the "preferred" crop type, without preventing user selecting other crops.

In particular, the request is for Gallery Articles in composer - when adding images, they should default to freeform (which will start as a full frame crop), but users should be able to select other, existing crops if desired.

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/crop/controller.js
+++ b/kahuna/public/js/crop/controller.js
@@ -50,6 +50,7 @@ crop.controller('ImageCropCtrl', [
       const allCropOptions = cropSettings.getCropOptions();
 
       const storageCropType = cropSettings.getCropType();
+      const storageDefaultCropType = cropSettings.getDefaultCropType();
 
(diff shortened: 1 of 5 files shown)


andrew-nowak (approved):
tested locally, works nicely, great work!

PR #4360, opened 2024-10-30
migrate kcl-akka-stream 2.1.0=>4.1.1, kcl 1.8.10=>2.6.0

andrew-nowak:
## What does this change?

We're using an old version of kcl-akka-stream, which was only ever released to bintray and so was committed directly into the repository when bintray was removed (see #3283). The earliest version of kcl-akka-stream that was released to Maven central is 4.1.1, which depends on kinesis-client-lib 2.x. Upgrade both together, to get onto a version which is both a) on Maven central and b) compatible with scala 2.13.

Note however that kcl-akka-stream is no longer maintained, and a Pekko alternative has not been created (see https://github.com/StreetContxt/kcl-akka-stream/issues/191) which will be an annoyance for a Play 3.0 upgrade. (Technically it isn't required that we remove all usages of Akka to upgrade to Play 3.0, but it'd be very very nice if we could). To be dealt with at a later date.

Note also that this doesn't upgrade the version of the kinesis-client-library used in usage, since that uses rxscala (also no longer maintained) instead of Akka, so doesn't make use of kcl-akka-stream.

kcl-akka-stream is only used by Thrall - so try making some changes to data, or upload images etc. and check that the updates flow through to Thrall, which will then apply those updates onto the Elasticsearch index.

## Tested? Documented?
- [x] locally by committer


bryophyta (approved):
this looks good to me 👍

PR #4366, opened 2024-11-04
remove dead code and unused elasticsearch dependency

andrew-nowak:
## What does this change?

This code appears unused, and is the only usage of an outdated elasticsearch dependency (all active elasticsearch queries use elastic4s, which depends on a different and up-to-date elasticsearch dependency)

Does the app compile? If so, all should be fine since this only removes code.

Technically this code might be loaded at runtime? I haven't seen any evidence of this from a (admittedly very brief) look around, I'd be pretty shocked if that was the case. Deploy to TEST to be sure?

## Tested? Documented?
- [ ] locally by committer


bryophyta (approved):
nice!

PR #4374, opened 2024-11-25
Fix usages stream for cartoons

simonbyford:
## What does this change?

Usages of images in cartoon elements were not being correctly recorded by the Grid. 

Usages are processed by the `usage-stream` service, which listens to events from CAPI then queries CAPI with the `'show-elements=image'` and `'show-atoms=media'` query params, extracting and building usage records accordingly. This wasn't happening for cartoons, as the images live inside the `cartoonVariants` field of the cartoon element. 

The fix was to add `'show-elements=cartoon'` to the CAPI query and look inside `cartoonVariants` for the images.

It has already been tested on CODE. We created a cartoon in `composer.code.dev-gutools.co.uk` and verified the usage came through as expected in `media.test.dev-gutools.co.uk`

Screenshot 2024-12-04 at 10 25 08

Image usages are correctly recorded for cartoons.

## Who should look at this?

@guardian/newsroom-resilience?

## Tested? Documented?

- [ ] locally by committer

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -171,8 +171,8 @@ lazy val thrall = playProject("thrall", 9002)
 
 lazy val usage = playProject("usage", 9009).settings(
   libraryDependencies ++= Seq(
-    "com.gu" %% "content-api-client-default" % "19.0.4",
-    "com.gu" %% "content-api-client-aws" % "0.7",
+    "com.gu" %% "content-api-client-default" % "32.0.0",
+    "com.gu" %% "content-api-client-aws" % "0.7.6",
     "io.reactivex" %% "rxscala" % "0.27.0",
     "com.amazonaws" % "amazon-kinesis-client" % "1.8.10",
     "com.google.protobuf" % "protobuf-java" % "3.19.6"
--- a/build.sbt
+++ b/build.sbt
@@ -171,8 +171,8 @@ lazy val thrall = playProject("thrall", 9002)
 
 lazy val usage = playProject("usage", 9009).settings(
   libraryDependencies ++= Seq(
-    "com.gu" %% "content-api-client-default" % "32.0.0",
-    "com.gu" %% "content-api-client-aws" % "0.7.6",
+    "com.gu" %% "content-api-client-default" % "32.0.0" exclude("org.slf4j","slf4j-api"), // slf4j exclusion can be removed when we upgrade to SLFJ v2
+    "com.gu" %% "content-api-client-aws" % "0.7.6" exclude("org.slf4j","slf4j-api"), // slf4j exclusion can be removed when we upgrade to SLFJ v2
     "io.reactivex" %% "rxscala" % "0.27.0",
     "com.amazonaws" % "amazon-kinesis-client" % "1.8.10",
     "com.google.protobuf" % "protobuf-java" % "3.19.6"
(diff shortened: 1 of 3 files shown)


usage/app/model/UsageGroup.scala
```diff
@@ -202,6 +207,24 @@ class UsageGroupOps(config: UsageConfig, mediaWrapperOps: MediaWrapperOps)
...
+                  image.mediaId
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }.toSet
+
```
twrichards, line 227: on talking through the deployment issue with @andrew-nowak , I realised this would be nicer as a for comprehension...
```suggestion
  private def extractCartoonUniqueMediaIds(content: Content): Set[String] = 
    (for {
      elements <- content.elements.toSeq
      cartoonElement <- elements.filter(_.`type` == ElementType.Cartoon)
      asset <- cartoonElement.assets.toSeq
      data <- asset.typeData.toSeq
      cartoonVariants <- data.cartoonVariants.toSeq
      cartoonVariant <- cartoonVariants
      image <- cartoonVariant.images
      mediaId <- image.mediaId
    } yield mediaId).toSet

```

simonbyford: Much nicer, thanks @andrew-nowak

build.sbt
```diff
@@ -40,7 +40,7 @@ val commonSettings = Seq(
     "org.mockito" % "mockito-core" % "2.18.0" % Test,
     "org.scalamock" %% "scalamock" % "5.1.0" % Test,
   ),
-  dependencyOverrides += "com.fasterxml.jackson.module" %% "jackson-module-scala" % "2.14.3",
+  dependencyOverrides += "com.fasterxml.jackson.module" %% "jackson-module-scala" % "2.17.2",
```
twrichards, line 43: @andrew-nowak  - doing this for all modules hasn't seemed to cause any problems AFAICT - but do you think it's safer to only do so for `usage`

andrew-nowak: Nah, if it works for everything then let's apply it everywhere 👍

build.sbt
```diff
@@ -171,8 +171,8 @@ lazy val thrall = playProject("thrall", 9002)
 
 lazy val usage = playProject("usage", 9009).settings(
   libraryDependencies ++= Seq(
-    "com.gu" %% "content-api-client-default" % "19.0.4",
-    "com.gu" %% "content-api-client-aws" % "0.7",
+    "com.gu" %% "content-api-client-default" % "32.0.0" exclude("org.slf4j","slf4j-api"), // slf4j exclusion can be removed when we upgrade to SLFJ v2
+    "com.gu" %% "content-api-client-aws" % "0.7.6" exclude("org.slf4j","slf4j-api"), // slf4j exclusion can be removed when we upgrade to SLFJ v2
```
andrew-nowak, line 175: After https://github.com/guardian/grid/pull/4377 was merged yesterday, the rest of the app is on slf4j v2, so if we rebase this branch then we shouldn't need these exclusions any more 🎉 
Worth giving a quick test after rebasing that the Play upgrade doesn't cause any other annoying conflicts in here too

andrew-nowak (approved):

build.sbt
twrichards: tested in `TEST` after this rebase 👍

PR #4377, opened 2024-11-26
Play 2.9

andrew-nowak:
## What does this change?

Perform upgrade to Play 2.9.

Ideally we want to be on Play 3.0, but that upgrade swaps Play's internals from using Akka to Pekko. We also have direct usages of Akka in our project, and while we could go to Play 3.0 and continue to run an Akka actor system to support those, it'd be nicest if we could swap both simultaneously. Stopping at 2.9 offers a nice stepping point where we can investigate that option, and also a spot to perform the Java 11 -> 17 upgrade too.

While testing locally I discovered that #4360 had quietly broken Thrall's ability to read the message queue while running locally, as the underlying Kinesis client attempts to connect with HTTP/2, but localstack only offers HTTP/1.1. Use some AWS SDK overrides to force it back onto HTTP/1.1 while running locally.

Run locally, deploy to TEST, try some usual Grid functionality.

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -35,7 +35,7 @@ val commonSettings = Seq(
   Test / testOptions ++= Seq(Tests.Argument(TestFrameworks.ScalaTest, "-o"), Tests.Argument(TestFrameworks.ScalaTest, "-u", "logs/test-reports")),
   libraryDependencies ++= Seq(
     "org.scalatest" %% "scalatest" % "3.2.19" % Test,
-    "org.scalatestplus.play" %% "scalatestplus-play" % "5.1.0" % Test,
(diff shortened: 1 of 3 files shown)


bryophyta (approved):
the code changes look good to me and it builds locally

I had a look at the logs from when this branch was deployed last week, and I can see healthy-looking logs from all of the separate Grid apps during that time. There are a couple of errors showing up as well, but they are of kinds that we see fairly commonly in both TEST and PROD, so seem unlikely to be related to this change.

(error messages: `Not all bytes were read from the S3ObjectInputStream, aborting HTTP connection...` and `Cannot update as no upload status for...`)

davidfurey:
Very behind on emails but belatedly 🥳

PR #4380, opened 2024-12-03
upgrade node from 16 to 20

twrichards:
This upgrades to node20 for the whole project, and doesn't seem to cause problems building the client side, but was mainly for the benefit of...

## `image-counter-lambda`

Which required stopping excluding `aws-sdk` from the bundle/zip - since v2 aws-sdk is no longer provided at runtime, v3 is, upgrading to which is beyond the scope of this change). 

The corresponding bump to the lambda runtime is in https://github.com/guardian/editorial-tools-platform/pull/827


twrichards:
Tested on `TEST` in conjunction with https://github.com/guardian/editorial-tools-platform/pull/827
image

Seems to work when run manually 
image

PR #4383, opened 2024-12-04
upgrade v2 kcl

andrew-nowak:
## What does this change?

The kinesis-client-library brings in a series of transitive dependencies with high vulnerability warnings. Upgrade to the latest in the v2 series to remove as many as possible with straightforward upgrades.

Further upgrades will be required in the near future, but we should follow the migration steps here https://docs.aws.amazon.com/streams/latest/dev/kcl-migration-from-2-3.html

Deploy to TEST, and upload or edit an image - check that the updates flow through the update stream to thrall and are persisted into the ES cluster.

## Tested? Documented?
- [ ] locally by committer


twrichards (approved):
LGTM (assuming you've tested in `TEST`)

PR #4387, opened 2024-12-17
file size limit for upload/ingest (optional and configurable)

twrichards:
Recently the image-loader boxes got completely choked up and the queue grew in size. This was down to multiple whopping great images (e.g. 1GB), this was in the sweet spot (or rather not-so-sweet spot) between being small enough to successfully process and so big that it causes OoM and the box falls over and is recycled (still not great, but at least eventually self-healing) whereas the boxes stayed alive (passing healthcheck) but ultimately not completing the processing of the images.

## What does this change?
**This PR introduces an optional limit via a new config option `upload.limit.mb` (in common.conf, as it's used by both `kahuna` and `image-loader`.**

This limit is imposed in two places...

- As it picks an item off the ingest queue in `image-loader` - so as to guard against images from any source (agency, photographer over FTP ~or UI upload~) - which means we retain a copy of the offending image in the fail bucket for closer inspection.

- For UI uploads (`kahuna`), to avoid burning the planet by wastefully sending all those bytes to S3 only to have the server tell us the file is too big, we can alert the user promptly in the client, like so...
image

Try dropping an image over 500MB (e.g. download https://media.gutools.co.uk/images/25287e288b9254437e65209b541aa8e8895d571e) into the ingest queue bucket (in a dir with your name on it) and you should see this in the logs (as it's not a UI upload...
image 
...and the file should end up in the fail bucket.

Try uploading the same file to the UI and you should see a much prompter failure.

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/image-loader/app/controllers/ImageLoaderController.scala
+++ b/image-loader/app/controllers/ImageLoaderController.scala
@@ -150,6 +150,7 @@ class ImageLoaderController(auth: Authentication,
               imageId, UploadStatus(StatusType.Failed, Some(errorMessage))
             )
           }
+          metrics.failedIngestsFromQueue.incrementBothWithAndWithoutDimensions(metricDimensions)
           Future.unit
         }
         else if (approximateReceiveCount > 2) {
--- a/image-loader/app/controllers/ImageLoaderController.scala
+++ b/image-loader/app/controllers/ImageLoaderController.scala
@@ -141,8 +141,8 @@ class ImageLoaderController(auth: Authentication,
 
         val approximateReceiveCount = getApproximateReceiveCount(sqsMessage)
 
-        if(s3IngestObject.contentLength > 500000000){ // 500MB
-          val errorMessage = s"File size exceeds the maximum allowed size (500MB). Moving to fail bucket."
+        if(config.maybeUploadLimitInBytes.exists(_ < s3IngestObject.contentLength)){
+          val errorMessage = s"File size exceeds the maximum allowed size (${config.maybeUploadLimitInBytes.get / 1_000_000}MB). Moving to fail bucket."
           logger.warn(logMarker, errorMessage)
           store.moveObjectToFailedBucket(s3IngestObject.key)
           s3IngestObject.maybeMediaIdFromUiUpload foreach { imageId =>
--- a/image-loader/app/controllers/ImageLoaderController.scala
+++ b/image-loader/app/controllers/ImageLoaderController.scala
@@ -141,7 +141,18 @@ class ImageLoaderController(auth: Authentication,
 
         val approximateReceiveCount = getApproximateReceiveCount(sqsMessage)
 
-        if (approximateReceiveCount > 2) {
+        if(s3IngestObject.contentLength > 500000000){ // 500MB
+          val errorMessage = s"File size exceeds the maximum allowed size (500MB). Moving to fail bucket."
+          logger.warn(logMarker, errorMessage)
+          store.moveObjectToFailedBucket(s3IngestObject.key)
+          s3IngestObject.maybeMediaIdFromUiUpload foreach { imageId =>
+            uploadStatusTable.updateStatus( // fire & forget, since there's nothing else we can do
+              imageId, UploadStatus(StatusType.Failed, Some(errorMessage))
+            )
+          }
+          Future.unit
+        }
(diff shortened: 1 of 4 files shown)


paperboyo:
> No image should legitimately be over 500MB

Hmm, this is not strictly true, as, supporting TIFFs, we can encounter, if rarely, genuinely larger ones, because of all the stuff we don’t care about: layers and such. But it is true that most images like this are a result of user error in creating them and not genuine. And Grid isn’t used as an archive for genuine authored multi-layered ones (we may hit this case if we will want to store editable master files from Photoshop integration, for example).

Secondly, it’s the quality of our imaging pipeline that often makes it worse/bombs: lack of parallelism but also bad habits around TIFFs etc (those should be helped by and also fixed in https://github.com/guardian/grid/pull/4068).

Thirdly, it’s often dimensions, not filesize, and one does not always follow the other.

All that said, this is sensible and practical stop gap and prevents someone uploading, mostly absurd, image repeatedly bringing the whole thing down. Downside is: users don’t know how to reduce size effectively losslessly.

Maybe, I would make it configurable (and not effective if not configured)?

RichardLe:
Hi Folks, Agree that there should be a limit on filesize - would be good for it be configurable.  We also need to have the ability to handle >500MB files - we're likely to get some large scans from archives (either by error or design) which could conceivably go over this size.  Short term happy with picking failures up from the bucket, medium term would be good to get more targeted handling, potentially notifications to a large image topic (or something like that)

twrichards:
@paperboyo @RichardLe - as per your comments its now optional and configurable `upload.limit.mb` in `common.conf` (see https://github.com/guardian/grid/pull/4387/commits/096e76164c6086945f61b29b723a57c2678674fd)

kahuna/public/js/upload/manager.js
```diff
@@ -31,7 +31,21 @@ upload.factory('uploadManager',
         };
     }
 
-    async function createJobItems(files){
+    async function createJobItems(_files){
+
+      const maybeUploadLimitInBytes = window._clientConfig.maybeUploadLimitInBytes;
+      const maybeFilesAboveSizeLimit = maybeUploadLimitInBytes && _files.filter(file => file.size > maybeUploadLimitInBytes);
```
bryophyta, line 37: I think the type of `maybeFilesAboveSizeLimit` ends up being `0 | undefined | File[]`, which should work fine in practice (assuming that we'd never want to actually enforce an upload limit of 0 bytes), but feels a bit awkward. Might be easier to reason about if we explicitly turned `0 | undefined` to `false`..?

```suggestion
      const maybeFilesAboveSizeLimit = !!maybeUploadLimitInBytes && _files.filter(file => file.size > maybeUploadLimitInBytes);
```

bryophyta: not blocking, though

image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -141,7 +141,18 @@ class ImageLoaderController(auth: Authentication,
...
+        if(config.maybeUploadLimitInBytes.exists(_ < s3IngestObject.contentLength)){
+          val errorMessage = s"File size exceeds the maximum allowed size (${config.maybeUploadLimitInBytes.get / 1_000_000}MB). Moving to fail bucket."
+          logger.warn(logMarker, errorMessage)
+          store.moveObjectToFailedBucket(s3IngestObject.key)
+          s3IngestObject.maybeMediaIdFromUiUpload foreach { imageId =>
+            uploadStatusTable.updateStatus( // fire & forget, since there's nothing else we can do
+              imageId, UploadStatus(StatusType.Failed, Some(errorMessage))
+            )
+          }
+          Future.unit
```
bryophyta, line 153: As far as I can tell this code path doesn't set off an increment in the `failedIngestsFromQueue` metric, which is incremented when an upload fails due to the wrong mimetype, for instance. Should we be incrementing the metric for size-related failures too?

twrichards: nice spot, added in https://github.com/guardian/grid/pull/4387/commits/064daab40a8782c6f0dd08e27f31945614d242ff

bryophyta (approved):
tested both UI and S3 uploads in `TEST` and all working as described 👍👍

PR #4388, opened 2024-12-18
tell snyk workflow to run with updated python

andrew-nowak:
## What does this change?

the latest ubuntu image will not use this version of python any more. this version isnt used at runtime so lets update the workflow to a nice modern version, which should last a few more years

## Tested? Documented?
- [ ] locally by committer


davidfurey:
I think you should just remove snyk rather than fixing it

andrew-nowak:
superseded by https://github.com/guardian/grid/pull/4394

PR #4391, opened 2025-01-09
remove snakeyaml dependency

andrew-nowak:
## What does this change?

Removes the snakeyaml dependency

Dependabot has detected a high vulnerability in this dependency, but I can't see that we use it anywhere in the codebase. Let's just remove it from the dependency list.

Does the project still compile and pass CI? That should be sufficient in this case.

-1 high vuln reported.

## Tested? Documented?
- [ ] locally by committer


davidfurey (approved):
Looks like this was added to bump a transitive dependency on snakeyaml which no longer exists 👍

PR #4396, opened 2025-01-13
shutdown metrics actors more gracefully

andrew-nowak:
## What does this change?

On shutdown, Play sends the metrics actors a shutdown message, but as it uses an 'ask', it waits for a response. When no response arrives, it assumes something has gone wrong and shuts down with a more severe error. Sometimes (possibly not always?) this can also interrupt Play's dev mode (recompiling and restarting the app on code changes).

In fact nothing has gone wrong (usually), we simply don't bother to send a response. Let's do that, both to be graceful, and to enable slightly faster shutdowns. The actual message sent is unimportant, as the requester does nothing with it, so send Unit.

Run locally, then stop the app. Does the app finish faster? Do the logs contain fewer shutdown related errors. (Currently some will end with errors that an actor did not respond within 5 seconds).

Faster shutdowns, fewer errors, more uninterrupted dev mode.

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/metrics/CloudWatchMetrics.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/metrics/CloudWatchMetrics.scala
@@ -132,8 +132,11 @@ private class MetricsActor(namespace: String, client: AmazonCloudWatch) extends 
   }
 
   def receive: Receive = {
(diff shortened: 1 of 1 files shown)


twrichards (approved):
haven't had a chance to test locally but I know you have, the PR description is well explained and code changes LGTM

PR #4398, opened 2025-01-14
support for diacritics in the basic search query and field names in quotes

AndyKilmory:
## What does this change?

Currently diacritics are not supported correctly in the search bar - if a user enters diacritics or unicode characters they are removed on page refresh (or entering image view followed by 'back to search'). This makes it hard for users to work with the search bar. 

This PR modifies the syntax.js file - specifically the regEx parser used to analyse/breakdown the query content to display correctly as chips etc. - to provide Unicode support

It has also been extended to handle field names that are contained within quotation marks

Ensure that any diacritics entered into search bar are retained on page refresh

## Tested? Documented?
- [x] locally by committer


paperboyo:
Thanks! It makes things better for Unicode. Fixes https://github.com/guardian/grid/issues/1900).

The quotes part is better, but not perfect.
1. [preexisting] `"quoted"` searches when reloaded lose quotes
2. [better] reloading `"fileMetadata.icc.Profile Description":"sRGB IEC61966-2.1"`nicely turns it into a chip
3. [preexisting] but, because of previous point, if, **after reloading**, you start typing normal text next such a new shiny chip, it breaks

No idea how easly/likely would be to fix pt. 1. I can’t see any regressions here and it makes Unicode situation better, so we may decide to follow up in future PRs if the fix isn’t trivial. cc @andrew-nowak

andrew-nowak (approved):

AndyKilmory:
thanks for approval etc

PR #4402, opened 2025-01-16
Play v3 migration

andrew-nowak:
## What does this change?

Upgrades Grid to Play v3.

The main changes include Grid's move from Akka to Pekko, which involves updating import paths but otherwise behaviour should be the same. To accommodate the move, we also switch from kcl-akka-stream to [kcl-pekko-stream](https://github.com/guardian/kcl-pekko-stream), which likewise involves updated import paths but should otherwise behave the same.

Run locally or on the TEST environment, and see if the app continues to perform

## Tested? Documented?
- [x] locally by committer


bryophyta (approved):
code looks good and runs locally as expected 🥳

PR #4404, opened 2025-01-22
Update the validity warning message colour to change colour when leased

andrew-nowak:
## What does this change?

1. Fixes a bug in the indicator of what % of selected images have been leased (some problem with interpolating in a "style" attribute during webpack production mode; works fine locally but completely broken in production. Couldn't find the root cause, so switch to using `ng-style`)

2. Then, change the colour of the validity warning message which applies when the image has restrictions or is overquota etc. to teal, so that it is clearer that the warnings have been overridden and the image can be used (usually because a lease has been applied).

Run on TEST, where I have configured getty to (almost certainly) be overquota - check the quotas screen to confirm. Ensure you have edit_metadata permission!

Then, open an overquota image (ie. search `+uploader:getty`). The warning message on the right should be coloured amber, like so: 
image

Now add an allow-cropping lease to that image. The message should turn teal (which is the colour used in the search page to indicate "not free, but has been leased and can be used"):

image

Hopefully less confusion about whether a leased image may be used.

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-info-panel/gr-info-panel.html
+++ b/kahuna/public/js/components/gr-info-panel/gr-info-panel.html
@@ -24,21 +24,21 @@
 
                     <div ng-switch-when="pay"
                          class="image-notice image-info__group status cost cost--pay"
-                         style="{{ctrl.stylePercentageLeased(cost)}}"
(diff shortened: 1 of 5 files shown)


bryophyta (approved):
works as described 👍👍

PR #4413, opened 2025-01-28
update image flag colour when leased

andrew-nowak:
## What does this change?

Follow on from #4404 - update the image's flag immediately when it's been leased. This was an oversight when the colouring was first added - we set it when the image is loaded, but don't reset when it's been updated. We now correct this mistake.

Deploy to TEST. From the search page, select an image (without _entering_ it - use the tickbox and open up the info panel on the right). Set it to some usage rights such as social media, so the image is _restricted_ but not blocked. Add a lease. Does the image's flag turn the leased teal colour, without having to reload the page?

image

## Tested? Documented?
- [x] locally by committer


bryophyta (approved):
lgtm 👍

PR #4418, opened 2025-02-15
Cost of `Pay`, because of excluded collection, should trump `Overquota`

paperboyo:
Co-authored by Copilot 🙀🔫🤖.  

## What does this change?
We have a concept of Excluded Collections (based on `source` field) where images are Paid, even if their Supplier is Free. Currently, when such a Supplier runs over quota, users are unable to see those images as Paid, they can only see they are Overquota (Overquota badge instead of a `£` one). Paid is a stronger signal (ie. should be leased less lightly) than Overquota, so they are asking for it to trump it.
This PR makes that change.

- Configure a Free supplier to run over quota
- Upload two images from this Supplier (or edit existing ones), so that one of them is from the Excluded Collection (currently, only Getty)
- notice how badges and Usage Rights notices in Viewer differ with this PR: a pound badge is clearly displayed for an image from an Excluded Collection

Hopefully, users will not lease Paid images as lightly as Overquota ones

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/media-api/app/lib/usagerights/CostCalculator.scala
+++ b/media-api/app/lib/usagerights/CostCalculator.scala
@@ -33,14 +33,16 @@ trait CostCalculator {
       val categoryCost: Option[Cost] = usageRights.defaultCost
       val overQuota: Option[Cost] = getOverQuota(usageRights)
       val supplierCost: Option[Cost] = usageRights match {
-        case u: Agency => getCost(u.supplier, u.suppliersCollection)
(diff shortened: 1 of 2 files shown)


paperboyo:
Sorry ChatNFT, but closing this in favour of your human overlords in https://github.com/guardian/grid/pull/4420. If you would want my advice, learn how to say “I don’t know”.

PR #4420, opened 2025-02-19
Make overquota cost only apply to images that would otherwise be free

andrew-nowak:
ie. excluded supplierCollections still appear as cost=pay, even if the agency is otherwise overquota

## What does this change?

We have a concept of Excluded Collections (based on `source` field) where images are Paid, even if their Supplier is Free. Currently, when such a Supplier runs over quota, users are unable to see those images as Paid, they can only see they are Overquota (Overquota badge instead of a `£` one). Paid is a stronger signal (ie. should be leased less lightly) than Overquota, so they are asking for it to trump it.
This PR makes that change.

- Configure a Free supplier to run over quota
- Upload two images from this Supplier (or edit existing ones), so that one of them is from the Excluded Collection (currently, only Getty)
- notice how badges and Usage Rights notices in Viewer differ with this PR: a pound badge is clearly displayed for an image from an Excluded Collection

Hopefully, users will not lease Paid images as lightly as Overquota ones

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/media-api/app/lib/usagerights/CostCalculator.scala
+++ b/media-api/app/lib/usagerights/CostCalculator.scala
@@ -1,18 +1,27 @@
 package lib.usagerights
 
-import com.gu.mediaservice.lib.config.RuntimeUsageRightsConfig
 import com.gu.mediaservice.model._
 import lib.UsageQuota
 
 trait CostCalculator {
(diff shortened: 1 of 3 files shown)


twrichards (approved):
code changes LGTM

PR #4424, opened 2025-02-27
add circle guideline to 1:1 cropping box for GNM users (plus optional circular mask)

twrichards:
Often 1:1 crops are used in circular containers, so it's useful to always have in mind what would be chopped off
image
... this could be applied easily to other ratios in future (e.g. 5:4) by updating the array, as per the comment - the CSS will centre the circle with it's diameter the shorter of height or width (to mirror the behaviour of the platform's auto-cropping of rectangular images).

PLUS (as per user request) a checkbox option to toggle a circular mask (on 1:1 only) instead of the circular guideline.
mask

NOTE: I've made the mask optional via checkbox, rather than always applied to 1:1, because we don't want to imply that the actual crop will be circular.

NOTE ALSO: the mask will only ever work for 1:1 (unlike the guideline, see above) - on any other ratio it will be an oval 🤮

### DIFF
--- a/kahuna/public/js/crop/controller.js
+++ b/kahuna/public/js/crop/controller.js
@@ -182,6 +182,10 @@ crop.controller('ImageCropCtrl', [
           window._clientConfig.staffPhotographerOrganisation === "GNM"
           && maybeCropRatioIfStandard === "5:3";
 
+        ctrl.shouldShowCircularGuideline =
+          window._clientConfig.staffPhotographerOrganisation === "GNM"
+          && maybeCropRatioIfStandard === "1:1";
+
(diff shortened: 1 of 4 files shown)


twrichards:
Could be `1px` if we really want (I prefer thicker for better contrast)
image

paperboyo (approved):
This was tested on TEST and code looks OK to **me** (emphasis mine).

I’m not too hot on the “Apply circular mask” checkbox because a) it’s not generalised to all crop ratios and b) needs to be clicked every time, without being remembered and having a keyboard shortcut, but I’m not freezing either.

Super-tiny nitpick: checkbox label text shouldn’t be selectable.

PR #4425, opened 2025-03-04
update image type detector to handle https in protocol in field

andrew-nowak:
## What does this change?

We've seen at least one pic come in with https as the prefix to the digitalSourceType value. I'm not sure that that's compliant with the spec, but it's also not explicitly forbidden and _is_ being provided, so let's fix it. Also resolves an open TODO now that Grid is running on Scala 2.13.

Updated unit tests should be sufficient for this.

## Tested? Documented?
- [ ] locally by committer


paperboyo (approved):
Google is doing it wrong and then it’s being [described wrongly](https://www.forbes.com/sites/paulmonckton/2025/02/09/google-announces-much-needed-ai-protection-feature-for-google-photos/). Must be correct.

I approve. Of an imperfect world trying to improve…

PR #4432, opened 2025-03-17
 add 5:4 as `Landscape NEW` and make ratio toggle bar more responsive

twrichards:
5:4 is soon to become the main ratio for Landscape crops, so it should be selectable (leaving 5:3 as the default for now). The toggle button bar for the crop options was already crowded (particularly ugly on smaller viewports) so now we display just the aspect ratio (if defined, note Freeform) below a certain viewport width, plus we already have the nice tooltip...
responsive-crop-options


andrew-nowak (approved):
lgtm, approving to unblock though I think there might be a desire to iterate a little further on this

andrew-nowak (approved):
latest commit also lgtm

PR #4433, opened 2025-03-17
remove notion of `shouldShowCropGuttersIfApplicable` query param and always display gutters when cropping 5:3 (for GNM users)

twrichards:
Back in #4343 we added the `shouldShowCropGuttersIfApplicable` query param to allow tools which iframe the grid (such as our CMS composer) to nudge the grid to show the striped gutters (see https://github.com/guardian/grid/pull/4331) on the sides of the crop rectangle when cropping for 5:3 (to show what could be clipped if that were used in a 5:4 slot). 

As the we progress towards 5:4 becoming the dominant crop ratio, it would be helpful for all users to see these gutters to reduce the likelihood that we need to replace crops which don't work in 5:4. So here we remove all notion of `shouldShowCropGuttersIfApplicable` query param and it displays the gutters on 5:3 for all GNM users.

We also add a handy toggle in the explainer to temporarily hide/show the gutters, should they be obscuring something important on a given crop...
gutters-toggle

### DIFF
--- a/kahuna/public/js/crop/controller.js
+++ b/kahuna/public/js/crop/controller.js
@@ -180,7 +180,6 @@ crop.controller('ImageCropCtrl', [
         const maybeCropRatioIfStandard = cropOptions.find(_ => _.key === ctrl.cropType)?.ratioString;
         ctrl.shouldShowVerticalWarningGutters =
           window._clientConfig.staffPhotographerOrganisation === "GNM"
-          && cropSettings.shouldShowCropGuttersIfApplicable()
           && maybeCropRatioIfStandard === "5:3";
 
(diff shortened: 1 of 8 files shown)


andrew-nowak (approved):
looks good, I wonder if grid should remember the show/hide setting between cropping sessions but we can decide that in response to user feedback

twrichards:
> looks good, I wonder if grid should remember the show/hide setting between cropping sessions but we can decide that in response to user feedback

the temporary/ephemeral nature of the hide/show was intentional because people will just turn it off and leave it off, so would only implement if there was strong feedback, I actually hope the hide function is used sparingly

PR #4436, opened 2025-03-17
use the word `Back` rather than `Cancel` on crop view as its less confusing

twrichards:
Particularly when using the grid in an iframe and when taken directly (by the host tool) to the cropping view for a given image, the word `Back` (`Go back to image` in the tooltip and `aria-label`) is more intuitive for users than the current word `Cancel`. 

This is useful when the current image isn't quite suitable and so the user wants to navigate to the photoshoot etc. via the metadata panel (currently the Cancel word confuses people into going back to the beginning and starting the search fresh).

Screenshot 2025-03-17 at 5 08 52 pm


jonathonherbert (approved):
👍, others may have thoughts 😀

PR #4443, opened 2025-03-31
upgrade commons-compress in scripts module to latest

andrew-nowak:
## What does this change?

Upgrades the commons-compress package, used in our scala scripts module, to the latest version. The previous version had  several security warnings against it, although these were for usages of zip, 7z and tar modules, whereas we only use for bz2.

Successful compilation should be enough, it is unlikely we will run these scripts in the near future and commons-compress is pretty stable, so likelihood of breakage is very low.

## Tested? Documented?
- [x] locally by committer


bryophyta (approved):
compiles as expected 🥳

PR #4445, opened 2025-04-03
force an upgraded dynamodb version to transitively get upgraded netty-handler

andrew-nowak:
## What does this change?

Scanamo has a transitive dependency on an old version of netty-handler which includes a high vulnerability, but has not yet published an update upgrading that dependency. Since the scanamo dependency comes through common-lib, and dependencyOverrides aren't shared to dependencies, it's easiest to declare an explicit dependency on the library version we want.

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -63,6 +63,7 @@ Global / concurrentRestrictions := Seq(
 )
 
 val awsSdkVersion = "1.12.470"
+val awsSdkV2Version = "2.31.12"
 val elastic4sVersion = "8.3.0"
 val okHttpVersion = "3.12.1"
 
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -98,14 +99,14 @@ lazy val commonLib = project("common-lib").settings(
...
-    "com.typesafe.scala-logging" %% "scala-logging" % "3.9.2",
     // needed to parse conditional statements in `logback.xml`
     // i.e. to only log to disk in DEV
     // see: https://logback.qos.ch/setup.html#janino
     "org.codehaus.janino" % "janino" % "3.0.6",
     "org.playframework" %% "play-json-joda" % "3.0.4",
     "org.scanamo" %% "scanamo" % "2.0.0",
-    // Necessary to have a mix of play library versions due to scala-java8-compat incompatibility
+    // declare explicit dependency on desired version of aws sdk v2 dynamo
+    "software.amazon.awssdk" % "dynamodb" % awsSdkV2Version,
```
bryophyta, line 109: the netty dependency was coming via the dynamodb sdk before? (I assume so, but just wanted to check)

bryophyta (approved):
sorry, I meant to approve alongside my comment before! 🤦

PR #4448, opened 2025-04-11
Better understand grace periods, and allow UI page loads to demand refresh of cookies in grace period

andrew-nowak:
## What does this change?

Since #4405 we've been running Grid with apiGracePeriod turned on. In most tools which directly use the Panda Actions implementations, this is handled by marking routes as requiring "API" auth, or "regular" (ie. unspecified) auth - API routes will continue to allow access during the grace period and then return 419, while the other sort will only allow access during the regular validity period, and afterwards send users to Google for session refresh, even if a grace period is set. 

However Grid uses a more complicated and customised panda verification logic, and as grace period had previously been turned off it had not been properly implemented. This meant that Grid would not send users to reauth during their grace period, and simply allow them to continue accessing until that also ran out. This meant that connection to other tools with different grace periods would fail until either a) the grace period ran out, or b) they visited another tool, which correctly sent them for a session refresh.

Update the logic in Grid's panda implementation to correctly spot and handle grace periods; so that controllers providing API responses can continue accepting tokens during the grace period, and controllers providing UI responses (ie. users will generally visit as part of a _page navigation_) will send users to reauth earlier.

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/rest-lib/src/main/scala/com/gu/mediaservice/lib/auth/Authentication.scala
+++ b/rest-lib/src/main/scala/com/gu/mediaservice/lib/auth/Authentication.scala
@@ -42,7 +42,8 @@ class Authentication(config: CommonConfig,
     Future.successful(respondError(new Status(419), errorKey = "authentication-expired", errorMessage = "User authentication token has expired", loginLinks))
   }
 
-  def authenticationStatus(requestHeader: RequestHeader): Either[Future[Result], Principal] = {
+  // authenticatedInGracePeriod - if true, then users with valid but recently-expired cookies are considered authenticated, and not required to immediately reauth
+  def authenticationStatus(requestHeader: RequestHeader, authenticatedInGracePeriod: Boolean): Either[Future[Result], Principal] = {
     def flushToken(resultWhenAbsent: Result): Result = {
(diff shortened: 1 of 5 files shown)


joelochlann (approved):
👍 thanks for looking into this! Tested locally and it does indeed issue a new cookie on page load after expiry, while API requests keep working.

This fix unfortunately doesn't fix the Pandular auto-refresh mechanism, which relies on a [419 response from an API request](https://github.com/guardian/pandular/blob/master/src/heal.js#L17), which we now only get after 25 hours.

I'm not 100% sure what this means for Pinboard behaviour because I haven't been able to test this locally. You said that as soon as a user enters the grace period, pinboard won't show up. Does this check only happen on page load, or will pinboard disappear/break after an hour (unless the user manually refreshes the Grid page)?

If the latter (which seems likely), then we're still not back to the functionality we had in Chrome before #4405, where Pandular would refresh the session in an iframe after the first 419 API response (i.e. after an hour), thereby keeping the cookie fresh for Pinboard.

But we are in a better place that we were before! And the Grid is at least more consistent with the behaviour in other apps.

Issue #1925, opened 2016-06-10
Date taken reconciliation logic widely incorrect


paperboyo:
Hello,

Grid’s is not able to acquire `dateTaken` correctly. Here is a bunch of examples converted using latest Lightroom: [PROD]search?query=~"dev blog"&uploadedBy=[me]@guardian.co.uk

All dates are (and, consequently, ordering) are wrongly set at:
`"dateTaken": "2016-06-10T00:00:00Z",`

We already read correct `dateTaken` which is recorded in multiple places inside `fileMetadata`, e.g.:
```
iptc": {
    "Digital Time Created": "15:28:37",
    "Digital Date Created": "20160610",
    "Time Created": "15:28:37",
    "Date Created": "Fri Jun 10 00:00:00 UTC 2016"  
"exifSub": {
"Date/Time Original": "2016:06:10 15:28:37",
    "Date/Time Digitized": "2016:06:10 15:28:37",
```

Recent [updates](https://github.com/drewnoakes/metadata-extractor/releases) to metadata-extractor cite
> Considerable overhaul of date, time and time zone handling across the library #153 #154 #155 #158

Maybe if we would to update it, we wouldn’t have to think too hard about how should we reconcile the dates ourselves?

Regards
Mateusz

paperboyo:
Also, this all-too-common `"dateTaken": "2016-06-14T00:00:00Z"` shows up in the UI differently depending on client's time zone. In my semi-holiday location (PL), it becomes:

"Taken on 14 Jun 2016, 02:00"

Hmm...

paperboyo:
From my very crude tests, it looks like huge majority of all images (~3/4) haven't got time in `dateTaken` read correctly. Please compare:

[PROD]/search?dateField=taken&takenSince=2016-06-12T23:59:59.999Z&takenUntil=2016-06-13T00:00:01.000Z (16 289 matches)

[PROD]/search?dateField=taken&takenSince=2016-06-13T00:00:01.000Z&takenUntil=2016-06-13T23:59:59.999Z (4 721 matches)

paperboyo:
[...]

PR #4455, opened 2025-04-22
Introduction of 'sort by taken date'

AndyKilmory:
## What does this change?

The basic sort b functionality has been extended to allow the user to select to sort matching images by 'Taken Date' - either new to old or old to new.

The sort by control has been extended to include the new options...
Screenshot 2025-04-22 at 14 40 00

Selecting  either taken date option will result in changes to the calls to the api and also the main images url...
Screenshot 2025-04-22 at 14 41 16

Some images do not have a taken date and those images matching the search criteria but without a taken date will be returned at the end of the list of matches. A simple transient notification banner has been included to indicate this...
Screenshot 2025-04-22 at 14 42 53

In addition the summary grid display of image thumbnails has been modified so that when a 'take date' ordering has been selected the displayed date switches from 'Uploaded' to 'Taken'...
Screenshot 2025-04-22 at 14 45 53

The remainder of the system functionality should continue unaltered. Copying of URLs containing the new order by parameters should be honoured. Newly ingested images should slot into the image list in accordance with their order by values etc.

Note: this is the first PR to extend sort by to include taken date a further PR is expected to imporve the functiobality further to include better management of images without taken dates and also re-factoring of the UI to make sort by management more intuitive.

Ensure that the sort by control behaves as expected when the new options are selected and ensure that all other functionality is unaffected by the change and that sort order is retained in the same way for taken date as for upload date

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -157,13 +157,14 @@ lazy val thrall = playProject("thrall", 9002)
     pipelineStages := Seq(digest, gzip),
     libraryDependencies ++= Seq(
       "org.codehaus.groovy" % "groovy-json" % "3.0.7",
-      // amazon-kinesis-client brings in a critical vulnerability warning through apache avro, resolved in versions 1.11.4 and 1.12.0.
-      // updating amazon-kinesis-client? check if the override below can be removed
-      "software.amazon.kinesis" % "amazon-kinesis-client" % "3.0.2",
-      "com.gu" %% "kcl-pekko-stream" % "0.1.1",
+      // TODO upgrading kcl to v3? check if you can remove avro override below
+      "software.amazon.kinesis" % "amazon-kinesis-client" % "2.6.1",
+      "com.gu" %% "kcl-pekko-stream" % "0.1.0",
       "org.testcontainers" % "elasticsearch" % "1.19.2" % Test,
       "com.google.protobuf" % "protobuf-java" % "3.19.6"
     ),
+    // amazon-kinesis-client 2.6.0 brings in a critically vulnerable version of apache avro,
+    // but we cannot upgrade amazon-kinesis-client further without performing the v2->v3 upgrade https://docs.aws.amazon.com/streams/latest/dev/kcl-migration-from-2-3.html
     dependencyOverrides ++= Seq(
       "org.apache.avro" % "avro" % "1.11.4",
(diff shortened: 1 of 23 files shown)


paperboyo:
This is amazing and long-sought for! Thank you.

Testing functionality, everything appears to be working correctly (sort order is reset on Home, correctly changes to Added to collection on Collection, survives adding, removing chips). 

These are some thoughts for consideration:
1. When I click on a collection, order correctly changes from Taken to Added to…. but when I remove collection chip then, it should go back to previous order I think (Taken if that was it)?
2. What is the secondary sort order (for images that have exactly the same, usually partial, Date Taken? I think it should be Upload Time, but it appears to be not that and I can’t tell what it is…
    taken
3. This is something completely unrelated and only to ponder from UX perspective, not for this PR. Some active searches are clear to users: text in searchbox, chips, checkboxes. But for some others – it’s not always clear to users that there is an active search different from default (esp. if it came with a link they clicked). This applies to eg. date filter and sort order. See below for a totally amateur way of trying to make that clearer (would apply only for when options differ from default, could be a dot or something else entirely). The idea is for users to be able to immediately tell they have some options active:
    image

Let us know what you think (only pts. 1–2 are any-relevant to this PR).

AndyKilmory:
Thanks for the feedback Mateusz - I'll look to implement points 1 and 2 and talk with the team regarding point 3.

AndyKilmory:
Have addressed points 1 and 2 with recent revisions. One thing I would say is that the UI doesn't show the full value held for 'taken date' in some cases - it shows to minute only - whilst the sort works on the fully resolved value - so a number of pictures can appear to have the same Taken Date in the UI but still doesn't order according to Upload Date. The sort on 'taken date' is resolving on data (seconds etc) not shown in the UI. Where the Taken Dates are identical then Upload Date will be the secondary sort. Hope that's clear. Thanks

paperboyo:
Thanks for all the fixes, I tested and confirm secondary sort order (uploadTime) works correctly, coming back from teh Collection sort works well and thanks too for more granularity in tooltips.

This is now ready for a formal review.

paperboyo:
Also, this fixes https://github.com/guardian/grid/issues/1925.

andrew-nowak (approved):

PR #4462, opened 2025-05-01
some untested work to make the photographer<->publication link sensitive

andrew-nowak:
to time

## Tested? Documented?
- [ ] locally by committer


common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/MetadataCleaner.scala
```diff
@@ -18,7 +18,6 @@ class MetadataCleaners(resources: ImageProcessorResources)
     BylineCreditReorganise,
     UseCanonicalGuardianCredit,
     ExtractGuardianCreditFromByline,
-    AttributeCreditFromByline.fromPublications(resources.commonConfiguration.usageRightsConfig.allPhotographers),
```
bryophyta, line 21: am I right in thinking that this is being removed as a pass-by refactor because it wasn't being used?

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -52,7 +52,7 @@ case class UsageRightsToMetadataParser(resources: ImageProcessorResources) exten
 class PhotographerParser(photographersConfig: UsageRightsConfigProvider) extends ImageProcessor {
   def apply(image: Image): Image = {
     image.metadata.byline.flatMap { byline =>
-      photographersConfig.getPhotographer(byline).map{
```
bryophyta, line 55: presumably in theory the fallback to `image.uploadTime` could be a lot later than the actual date the photo was taken (e.g. if someone was uploading archival photos)

i'm guessing this has been discussed with relevant people already, though?

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/MetadataCleaner.scala
andrew-nowak: ish - it was being run, but it's a duplicate of this supplier processor https://github.com/guardian/grid/blob/main/common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala#L26 (the difference between a supplier processor and a cleaner isn't particularly well defined, but I'd argue this is more of a "processing" than a "cleaning")

common-lib/src/main/scala/com/gu/mediaservice/lib/guardian/GuardianUsageRightsConfig.scala
```diff
@@ -103,25 +113,25 @@ object GuardianUsageRightsConfig extends UsageRightsConfigProvider {
...
-      "Steve Bell",
+      PublicationPhotographer("Steve Bell"),
     )),
     PublicationPhotographers(ObserverPublication, List(
-      "Chris Riddell",
-      "David Foldvari",
-      "David Simonds",
+      PublicationPhotographer("Chris Riddell"),
+      PublicationPhotographer("David Foldvari"),
+      PublicationPhotographer("David Simonds"),
```
bryophyta, line 134: is the reason why these Obs people don't have `to` dates, whereas the people in `ContractedPhotographers` (mostly) do, because we don't expect any new images to come from these three, whereas the contracted photographers are liable to submit new Guardian images? Or am I misunderstanding something about the logic here?

bryophyta (approved):
I've added a couple of questions, but I think I've followed everything here and it all looks good from what I can see! Going to approve, but if any of my questions suggest I haven't understood something then shall we have a quick synchronous walk-through?

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
andrew-nowak: Yep, while imperfect I think this is fine as new images are more likely to be uploaded than archives, so better to apply the wrong metadata and let uploader find and correct it, than to not apply it all to the images where it would be correct. It's also no more wrong than current behaviour which would do the same.

PR #4463, opened 2025-05-01
also add some horizontal gutters when cropping 5:4s, because it is true that they might be used in 5:3 places

andrew-nowak:
## What does this change?

We've had vertical gutters when cropping 5:3s showing what bits will be cut off when they are forced into 5:4 containers, but now that we're switching over to defaulting to 5:4, for a while there will be the opposite problem: 5:4 crops forced into 5:3 containers. For the duration of the switchover, add a couple more gutters demonstrating what portion of the image may be lost to this.

image

## Tested? Documented?
- [x] locally by committer


bryophyta (approved):
I'm not very familiar with this bit of functionality but the code changes here look good to me!

PR #4464, opened 2025-05-06
explain which crop types are allowed in this context by ratio string …

andrew-nowak:
…instead of alias name

## Tested? Documented?
- [ ] locally by committer


bryophyta (approved):
this looks good to me 👍

PR #4465, opened 2025-05-09
make has and is queries understand quoted values

andrew-nowak:
## What does this change?

Fix the query syntax to allow quoted phrases as input to `+has:` and `+is:` chips, to allow values with spaces

## Tested? Documented?
- [x] locally by committer


paperboyo:
Thanks! This seems to work as expected (I couldn’t break anything) It allows deep searches for fields with spaces in them, like this one:
`…search?query=has:"fileMetadata.iptc.Date Created" -has:"fileMetadata.iptc.Time Created" has:"fileMetadata.iptc.Date Time Created Composite"&nonFree=true`

bryophyta (approved):
lgtm!

PR #4468, opened 2025-05-27
Add a feature switch to use @guardian/cql in the structured search component

jonathonherbert:
## What does this change?

Add a feature switch to toggle using [`@guardian/cql`](https://github.com/guardian/cql) in the structured search component.

This should address most (but maybe not all!) of the issues listed in https://github.com/guardian/grid/issues/1801. It also opens the door to more expressive queries, including boolean operators and groups, which `@guardian/cql` understands, but are disabled here.

## Developer notes

This change also moves a few of the files it directly alters to Typescript, and adds the necessary types where they weren't present. Rather than move the files it consumes but does not touch to typescript, I've added the relevant types as typedoc (e.g. [here](https://github.com/guardian/grid/blob/a77131e542539b526f1216acb2d496bca95ab2fb/kahuna/public/js/services/api/edits-api.js)) to provide a better developer experience; hopefully that paves the way towards better typesafety elsewhere without unnecessarily increasing the surface area of this PR.

I've added tests for the code responsible for transforming the structures CQL produces into the `StructuredQuery` format the Grid uses, but there are no tests that exercise the UI — the layer between the UI the underlying CQL webcomponent is very thin, and `@guardian/cql` itself is well tested at the UI level.

## This all sounds lovely. What's the catch?

This change depends on the https://github.com/guardian/cql package, which has largely (but not entirely!) been worked on by a single developer (me) outside of official project work. That's not a wonderful bus factor as it stands. Bugs and feature requests will need to be released through that package.

A few notes in favour: 
- The package has reasonable test coverage at unit and integration level (~85%)
- It uses a release process that is standard for the dept.

And to bear in mind: 
- Using it will add 80kb (min/gz) to the Grid's bundle size, because:
- It depends on [ProseMirror](https://prosemirror.net/) for most of its editing capabilities. ProseMirror is well-known within the Guardian organisation, but has a learning curve.
- The package carries a reasonable degree of complexity and a fair amount of code, and this increased surface area will come with an associated maintenance cost, plus dependencies (anecdotally, I don't recall any security advisories affecting ProseMirror since I've used it; ~5+ years.)

Running locally or in CODE, have a go at testing queries. Here's a handy checklist drawing from https://github.com/guardian/grid/issues/1801 and mine and @paperboyo's own testing (GIFs are very flashy so they're in details tags)

- Undo/redo should work as expected across the query 

<details><summary>GIF</summary>

undo-redo

</details>

- Copy and paste should work as expected for entire and partial queries. (You can come up with interesting results by copying and pasting part of a chip, which is something we should improve, perhaps by expanding the selection to cover entire chips when serialising for paste.)
<details><summary>GIF</summary>

copy-paste

</details>

- Caret movement and selection should work as expected across the query. `Mod-A` (`Win-A`/`Cmd-A`) selects just the chip value if you are within a chip value, and selects the entire query on the next press. Platform-specific commands to move the caret across words, and to the start/end of the line, should work as expected
<details><summary>GIF</summary>

caret-movement

</details>

- Using quoted terms for chip keys and values should work as expected, e.g. `+"fileMetadata.colourModelInformation.hasAlpha":"true"`
<details><summary>GIF</summary>

quoted-terms

</details>

- Queries with lots of content should scroll once they fill the input
<details><summary>GIF</summary>

large-queries

</details>

A better experience using chips :)

## Tested? Documented?
- [x] locally by committer


kahuna/public/js/window.ts
```diff
@@ -1,18 +1,48 @@
...
+      permissionsDefault?: string | undefined;
+      sessionId: string;
+      feedbackFormLink: string;
+      fieldAliases: FieldAlias[];
+      homeLinkHtml: string;
+      canDownloadCrop: boolean;
+      showDenySyndicationWarning: boolean;
+      showSendToPhotoSales: boolean;
+      domainMetadataSpecs: string[];
+      recordDownloadAsUsage: boolean;
```
lindseydew, line 25: It looks like this is bringing a lot more things into global scope if I'm understanding correctly. Could you explain why this is needed?

kahuna/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "noImplicitAny": true,
     "module": "es6",
```
lindseydew, line 4: Wondering whether it would be worth separating this upgrade into a separate PR and release this independently?

kahuna/public/js/components/gr-cql-input/syntax.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+    structuredQuery: [
+      {
+        filterType: "inclusion",
+        key: "fileMetadata.xmp.xmpMM:OriginalDocumentID",
+        type: "filter",
+        value: "urn:uuid:4d8c1ca2-245d-43d9-82e1-a6be6ec7e059"
+      }
+    ]
+  }
+];
```
lindseydew, line 160: Awesome tests.

lindseydew (approved):
This is looking good to me. I can't comment much on the specifics of the angular code, but as a team we are happy to role this forward with a feature switch. Let's make sure CP are informed, so that there is an easy rollback incase there are unforeseen issues in the production environment

kahuna/public/js/window.ts
jonathonherbert: There's no runtime impact here: this is extending the type definitions of the `_clientConfig` global object, which is added to `window` in [main.scala.html](https://github.com/guardian/grid/blob/6001d42263a04ecfb6b3f116e73885222be1dc95/kahuna/app/views/main.scala.html#L50), to let the compiler understand what we're up to in Typescript when we're interacting with `window._clientConfig`.

The PR needs a fair few of these fields in [query-suggestions.ts](https://github.com/guardian/grid/blob/18610228eaf06b736009162324949a56d62f4873/kahuna/public/js/search/structured-query/query-suggestions.ts) — I think rather than just fill out the fields this PR needed, I did them all, as they'll likely be useful down the line. I could definitely thin these down to just the ones required in the PR if needed.

kahuna/tsconfig.json
jonathonherbert: 👍  https://github.com/guardian/grid/pull/4535

PR #4483, opened 2025-06-30
Eelpie/seperate recent usage signals

tonytw1:
## What does this change?

On the image preview, split `recentUsages` into recent print and recent digital values so that the print and digital usages can be independency highlighted as recent.

Currently it looks like print and digital will both turn red if either of them is qualifies as recent.

Screenshot 2025-06-30 at 22 05 53

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/kahuna/public/js/preview/image.js
+++ b/kahuna/public/js/preview/image.js
@@ -111,13 +111,15 @@ image.controller('uiPreviewImageCtrl', [
       const hasSyndicationUsages$ =
           imageUsagesService.getUsages(ctrl.image).hasSyndicationUsages$;
 
-      const recentUsages$ = imageUsagesService.getUsages(ctrl.image).recentUsages$;
+      const recentPrintUsages$ = imageUsagesService.getUsages(ctrl.image).recentPrintUsages$;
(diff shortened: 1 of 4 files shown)


kahuna/public/js/preview/image.js
```diff
@@ -111,13 +111,15 @@ image.controller('uiPreviewImageCtrl', [
       const hasSyndicationUsages$ =
           imageUsagesService.getUsages(ctrl.image).hasSyndicationUsages$;
 
-      const recentUsages$ = imageUsagesService.getUsages(ctrl.image).recentUsages$;
+      const recentPrintUsages$ = imageUsagesService.getUsages(ctrl.image).recentPrintUsages$;
+      const recentDigitalUsages$ = imageUsagesService.getUsages(ctrl.image).recentDigitalUsages$;
```
tonytw1, line 115: No one else in this file is capturing imageUsagesService.getUsages in a local usages val so presumably there is no performance change from multiple calls to imageUsagesService.getUsages?

andrew-nowak (approved):

andrew-nowak:
thanks Tony!

PR #4494, opened 2025-07-15
resolve deprecation warning in script

andrew-nowak:
also add a compilation step for scripts to CI so we might spot this earlier in the future


bryophyta (approved):
lgtm 👍

PR #4498, opened 2025-07-21
Updating the domainMetadata to allow introduction of 'key' field to allow search link to be included. Along with force soft delete endpoint and additional usage tab info

AndyKilmory:
Updating the domainMetadata to allow introduction of 'key' field to allow search link to be included - so searchable links can be present in domainMetadata. 

Addition of force soft delete endpoint for use with Capture, so images can be soft deleted regardless of usage information etc.

Addition of information on usages tab to indicate linked images are present so that users reviewing image usage information know to look at other images if requried.

## What does this change?

1) Adds in a new media-api endpoint - DELETE:<api-url>/images/<<image-id>>/force-delete - that forces the soft-delete of an image regardless of Usages, Crops etc. Normally attempts to soft-delete an image with Crops will be denied. In the case of Capture ingest we have the need to soft-delete images regardless of Usage history etc. to handle the possibility of logical image re-scans with higher quality physical images. This new endpoint is required to support this.

2) Add in the ability to complete a key/value pair search of the image metadata based on the domain metadata (currently the domain metadata is retained only on a textual basis. This functionality allows the user to add in a key definition which will permit the search of the ElasticSearch index

3) Add messaging to the Usages tab to indicate that the current image replaces or is replaced by and equivalent logical image - so the user knows to look at other images when reviewing usage history. Uses a defined list of metadata fields to identify where the information about the equivalent image information is held.

1) Ensure that an image that would not be soft-deleted via the usual delete endpoint can be deleted via the 'force-delete' endpoint

Screenshot 2025-07-23 at 16 50 14

2) Add in appropriate 'Supersedes'/'Replaces' and 'Superseded By'/'Replaced by' data and ensure the search operation operate correctly.

Screenshot 2025-07-23 at 16 48 10

3) Check the messaging regarding linked images on the Usages tab is correct.

Screenshot 2025-07-23 at 16 49 25

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/common-lib/src/test/scala/com/gu/mediaservice/testlib/ElasticSearchDockerBase.scala
+++ b/common-lib/src/test/scala/com/gu/mediaservice/testlib/ElasticSearchDockerBase.scala
@@ -16,7 +16,7 @@ trait ElasticSearchDockerBase extends BeforeAndAfterAll {
 
   val esContainer: Option[ElasticsearchContainer] = if (useEsDocker) {
     {
-      val container = new ElasticsearchContainer("docker.elastic.co/elasticsearch/elasticsearch:7.16.2")
+      val container = new ElasticsearchContainer("docker.elastic.co/elasticsearch/elasticsearch:8.18.3")
         .withExposedPorts(9200)
         .withAccessToHost(true)
         .withEnv(Map(
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,6 @@
 services:
   elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch:7.16.2
+    image: docker.elastic.co/elasticsearch/elasticsearch:8.18.3
     environment:
       ES_JAVA_OPTS: "-Xms1024m -Xmx1024m"
     volumes:


AndyKilmory:
Hi Mateusz, thanks for your feedback - all very helpful. We're aware that this proposal is a compromise designed to address a very specific need. We are also keen to look in more detail at versioning of images and interconnecting images as part of a much larger piece of work - but we need to address the needs of the Archives team in the short term. I think it would be good if we could convene a group to discuss the broader issues of versions, hidden images etc.

In terms of the use of soft delete of the 'old' images - we need to retain the usage information on the old image so that we have a proper history of what version of the image was used by whom and when - whilst at the same time understanding the concept of a 'newer' version. What is happening is the Archives team are re-scanning the original hard copy to produce a better electronic image and in the Capture/Photosales system they are able to replace the old jpeg with the newer one on the same underlying system record - but when they send this new image to BBC Images it creates a new record and we can't just 'replace' the old image - hence the need for this type of functionality. We can discuss this specific short term requirement in more depth at the next Grid hour. Thanks

twrichards:
Thanks for confirming (via Slack) that https://github.com/guardian/grid/pull/3998 and https://github.com/guardian/grid/pull/4525 fulfil the requirements of this PR (but with a more generic approach) - so I'm closing this PR in favour of those. Thank you.

PR #4516, opened 2025-08-29
Usage and Lease creation datetime displayed in tooltips

paperboyo:
## What does this change?

Our colleagues from Licensing&Syndication would like to see more data around syndication flow right in the UI. This PR adds:
- creation datetime of a Lease in its tooltip (in addition to existing createdBy)
- non-relative datetime when hovering over usage relative one

It also switches Leases tooltips from fancy top standard, because I couldn’t figure out how to make titip accept a line break… (if you know how: shoot!)

Before:
image

After:

image

image

I also fantasised about dollar signs in browser to tooltip when image was sent to syndication (for green ones) and when RCS sent syndication payload (for white ones), but this is too hard.

Find images with Leases and or/Usages and hower over a Lease and over Usage date.

Tooltips should spell out maximum of useful information.

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-image-usage/gr-image-usage-list.html
+++ b/kahuna/public/js/components/gr-image-usage/gr-image-usage-list.html
@@ -11,7 +11,8 @@
             {{usage.title}}
 
             <div class="more-info">
-                <span class="date-added">{{::ctrl.formatTimestamp(usage.dateAdded)}}</span>
+                <span title="{{usage.dateAdded | date:'d MMM yyyy, HH:mm'}}" class="date-added">{{::ctrl.formatTimestamp(usage.dateAdded)}}</span>
+
                 <ul class="reference-list">
                     <li ng-repeat="reference in usage.references">
                         <a ng-if="reference.uri" href="{{reference.uri}}" title="Open in {{reference.type}}" rel="noopener" target="_blank">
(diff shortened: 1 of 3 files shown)


kahuna/public/js/components/gr-tooltip/gr-tooltip.js
```diff
@@ -14,17 +14,33 @@ tooltip.directive('grTooltip', [
...
+
+                 const autoUpdates = angular.isDefined(attrs.grTooltipUpdates);
 
                 if (autoUpdates) {
-                    $scope.$watch(() => attrs.grTooltip, onValChange(newTooltip => {
-                        element.attr('data-title', newTooltip);
-                    }));
+                  $scope.$watch(() => attrs.grTooltip, onValChange(newTooltip => {
+                    if (attrs.grTooltipHtml !== undefined) {
+                      element.find('.titip-content').html(newTooltip);
```
bryophyta: setting content with `.html` opens us to html injection, in principle, and I don't see this being used elsewhere in the codebase (except to clear the html content of a component)

so, although we do trust the content we're putting in, in this case, I think it'd be better to avoid if we can. I wanted to test a possible alternative but am having trouble running Grid locally at the moment; will update when I can

bryophyta (approved):
Code looks good to me, and it works as expected when testing locally 👍

PR #4517, opened 2025-09-01
Set .text instead of .html in tooltips

bryophyta:
## What does this change?

It seems as though if we set tooltip content via `.text` then `\n` will be rendered as a line break, and setting `.text` on an element should be safer than setting `.html`.

image <-- line break is still present

 

## Tested? Documented?
- [ ] locally by committer


paperboyo:
Amazing! Thank you! Also, I didn’t know one can PR into an unmerged branch. Means we can just merge this, right? (would be super-OK to just commit into my branch directly)

PR #4518, opened 2025-09-02
CropController uses GridClient for the get source image call to Media API

tonytw1:
Moving this call to GridClient helps to enclosure all the service to service URL and HTTP client concerns in one place.

This call was probably the only service to service call circa 2021. GridClient seems to be how more recent service to service calls are done.

Get SourceImage requests additional fields via media-api query parameters so add queryStringParameters parameter to makeGetRequestAsync.

Extract the media api uri to image id validation to a function for testing and make it validate image URIs rather than blanket Media API.

## Tested? Documented?
- [ ] locally by committer


cropper/app/controllers/MediaApiUrls.scala
```diff
@@ -0,0 +1,16 @@
+package controllers
+
+trait MediaApiUrls {
```
andrew-nowak, line 3: definitely nitpicking here, but I think I'd have a mild preference for helpers like this without any dependencies to be in an `object` rather than a `trait`

```suggestion
object MediaApiUrls {
```

andrew-nowak (approved):
lgtm, let me know if you disagree with my nitpick and I can merge as is

cropper/app/controllers/MediaApiUrls.scala
tonytw1: Yes. This looks better.
Import `MediaApiUrls._` and lose the `with MediaApiUrls`

Amended and rebased.

PR #4519, opened 2025-09-05
support 'negated' NestedFilters with 🆕 `NegatedNestedFilter`

twrichards:
Co-authored-by: @frederickobrien 

As per https://trello.com/c/6YFv1T5L/955-feature-small-usages-searches-cant-be-used-in-exclusions until now we haven't been able to perform a negated search (a.k.a. NOT - signified by the `-` on the chip, rather than `+`) on `usages@...` (i.e. `NestedFields` because such a 'term' in the query was simply ignored, because there was no logic to deal with it and map/convert it to the ElasticSearch clause... this PR adds that logic 🎉 

This also unblocks #3998 (as there we want to hide `usage@status:replaced` from search by default) which is somewhat urgent (so we don't need to merge #4498).

### DIFF
--- a/media-api/app/lib/elasticsearch/QueryBuilder.scala
+++ b/media-api/app/lib/elasticsearch/QueryBuilder.scala
@@ -88,18 +88,13 @@ class QueryBuilder(matchFields: Seq[String], overQuotaAgencies: () => List[Agenc
     case Nil => matchAllQuery()
     case condList => {
 
-      val (nested: List[Nested], normal: List[Condition]) = (
+      val (nested: List[Nested], negationNested: List[Nested], normal: List[Condition]) = (
         condList collect { case n: Nested => n },
+        condList collect { case NegationNested(n) => n },
         condList collect { case c: Condition => c }
       )
 
-      val query = normal.foldLeft(boolQuery()) {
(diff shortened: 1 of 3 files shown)


media-api/app/lib/elasticsearch/QueryBuilder.scala
```diff
@@ -116,7 +111,15 @@ class QueryBuilder(matchFields: Seq[String], overQuotaAgencies: () => List[Agenc
...
-      nestedQueries.foldLeft(query) { case (q, nestedQ) => q.withMust(nestedQ) }
+      val queryWithNormal = normal.foldLeft(boolQuery()) {
+        case (query, Negation(cond)) => query.withNot(makeQueryBit(cond))
+        case (query, cond@Match(_, _)) => query.withMust(makeQueryBit(cond))
+        case (query, _) => query
+      }
+
+      val queryWithNestedAndNormal = listOfNestedToQueries(nested).foldLeft(queryWithNormal) { case (q, nestedQ) => q.withMust(nestedQ) }
+
+      listOfNestedToQueries(negationNested).foldLeft(queryWithNestedAndNormal) { case (q, negNestedQ) => q.withNot(negNestedQ) }
```
Fweddi, line 122: I can just about make out what's happening here but it could do with some more clarity?

PR #4525, opened 2025-09-12
improve display of parent/child relationships (e.g. thumbnails)

twrichards:
Builds on https://github.com/guardian/grid/pull/3998 adding improved display of parent/child relationships (e.g. thumbnails).

image
 image

### Detail

If any of the following exist on an image...
- Parents
  - `derivative-of-media-ids` identifier (value is comma separated if multiple)
  - `replaces-media-id` identifier 
- Children
  - has usages with replaced status
  - has usages with derivative status

... the image detail endpoint of media-API now looks up each of the referenced media IDs in ElasticSearch (using existing `getImageById` function[^1]) and maps each to the following shape... 
```scala
case class RelationDetail(
  thumbnail: String,
  addedBy: String,
  addedAt: DateTime,
  dimensions: Option[Dimensions]
)
``` 
and are placed in a new object `parentAndChildDetails` on the response, keyed with human readable  titles for the respective lists...
- "Replacement for"
- "Replaced by"
- "Derivative of"
- "Derivatives"

This then powers a relatively generic UI that renders these as thumbnails at the bottom of the left hand sidebar (where original & crops currently display at the top, with some neat sizing such that scrolling is nice when there are lots of crops and lots of parents/children.)

[^1]: I did attempt doing it in a single ES call using the `ids` search param but the logic to piece it back together was rather messy and this is such an infrequent feature I think the extra call(s) are worth it

### DIFF
--- a/media-api/app/controllers/MediaApi.scala
+++ b/media-api/app/controllers/MediaApi.scala
@@ -628,6 +628,8 @@ class MediaApi(
         val deleteImagePermission = authorisation.isUploaderOrHasPermission(request.user, source.instance.uploadedBy, DeleteImagePermission)
         val deleteCropsOrUsagePermission = canUserDeleteCropsOrUsages(request.user)
 
+        import JodaWrites._
+        implicit val jsonDetailsWrites: OWrites[RelationDetail] = Json.writes[RelationDetail]
         val getRelationDetails = elasticSearch.getRelationDetails(id, imageResponse.getSecureThumbUrl)_
         val relationDetails = Map(
           "Replacement for" -> source.instance.identifiers.get(ImageStorageProps.replacesMediaIdIdentifierKey).map(getRelationDetails),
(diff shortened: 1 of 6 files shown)


media-api/app/controllers/MediaApi.scala
```diff
@@ -636,8 +648,12 @@ class MediaApi(
           request.user.accessor.tier
         )
 
-        Some((source.instance, imageData, imageLinks, imageActions))
-
+        Some((
+          source.instance,
+          imageData.asInstanceOf[JsObject] + ("parentAndChildDetails" -> Json.toJson(relationDetails)),
```
andrew-nowak, line 653: you can avoid this cast by changing the return type of `ImageResponse.create` to explicitly return `JsObject`

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -332,6 +332,25 @@ class ElasticSearch(
     }
   }
 
+  def getRelationDetails(mediaIdThisIsFor: String, getSecureThumbUrl: Image => String)(
+    id: String
+  )(implicit ex: ExecutionContext, logMarker:MarkerMap = MarkerMap()): (String, Option[RelationDetail]) = {
+    try {
+      id -> Await.result(getImageById(id), 5.seconds).map{image =>
```
andrew-nowak, line 339: this looks wrong to me - this could potentially be a lot of getImage calls happening synchronously. The data's pretty nested so it isn't trivial moving it up and making it all async and future-based, but I've had a stab at it here https://github.com/guardian/grid/pull/4658

PR #4540, opened 2025-10-09
Fix an issue where typing e.g. -sport{Enter} would not-dechip the input to produce the plain query -sport

jonathonherbert:
## What does this change?

Fix an issue where typing e.g. -sport{Enter} would not-dechip the input to produce the plain query -sport. Caused by an import that was inadvertently removed without error.

|Before|After|
|-|-|
|de-chip-nope|de-chip|

Running locally or in CODE, type `-sport` into the search bar, and then press Enter. The chip should disappear, leaving `-sport`

Fewer confused users wondering why their negations don't work.

## Tested? Documented?
- [x] locally by committer


paperboyo (approved):
I approve of this effort! (blindly, though!)

andrew-nowak (approved):

bryophyta (approved):

PR #4634, opened 2026-02-23
Remove S3 signer override

andrew-nowak:
In the grid, we want to upgrade from AWS Java SDK v1 to v2. Working on the s3 module, we discovered one non-standard usage:

https://github.com/guardian/grid/blob/ad24535143bf96b2a288e177fd8d74f350d22f27/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala#L181-L193

Here we override the SDK defaults which use AWS Signature version 4, to instead use version 2. We do this as the presigned urls created for and sent to "imgops" (the nginx service) are failing to authenticate with the version 4.

It's failing for 2 reasons:

1. It's making requests to the "global", ie. un-regionalized, S3 endpoint, but the version 4 signer is signing against the endpoint which declares the region.
2. It makes requests to the S3 backend by using nginx's "proxy_pass" directive. This receives a request, makes a request (currently the _same_ request) to a different remote URL and then transforms the response. The problem here is that we pass 4 parameters to _imgops_ (width `w`, height `h`, quality `q`, rotation `r`) which are not relevant to the _S3_ endpoint, and so are not included in the signature. But as they're not removed from the request, they are included when the S3 service authenticates the signature, which causes a failure due to a lack of equality. 
 
We need to filter those 4 parameters out, so that only the parameters which are signed are included in the final request. This is not at all straight forward in nginx config, which despite looking like a scripting language (with variables and condition flow statements), is in fact a configuration language, where those constructs regularly do *not* work as you'd expect (read this page for more info if you're interested: <https://github.com/nginxinc/nginx-wiki/blob/master/source/start/topics/depth/ifisevil.rst>).

We've done this for our nginx imops backend already, so we should now be free to remove the signer override, which will free us to perform the S3 SDK upgrade!

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
@@ -177,21 +177,13 @@ object S3Ops {
   val s3Endpoint = "s3.amazonaws.com"
 
   def buildS3Client(config: CommonConfig, forceV2Sigs: Boolean = false, localstackAware: Boolean = true, maybeRegionOverride: Option[String] = None): AmazonS3 = {
-
-    val clientConfig = new ClientConfiguration()
-    // Option to disable v4 signatures (https://github.com/aws/aws-sdk-java/issues/372) which is required by imgops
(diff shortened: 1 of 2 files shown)


lindseydew (approved):
LGTM 👍

PR #4651, opened 2026-03-03
Post PR4475 fixes for Free to use only check box

tonytw1:
After rebasing up to this point, we found the behaviour of the Free to use only check box regressed.
Screenshot 2026-03-03 at 21 36 58

Disclaimer; we are forked quite away from the main Grid so this may not reflect want happens in the main branch.
Provided FYI.

#### Back to search

With Free to use only checked,
clicking into an image then
click Back to search the Free to use only check box has lost it's state:

https://github.com/user-attachments/assets/fd369e15-bdf6-4736-a84a-f83dfaa9adcf

Free to use only should have kept it's checked state

#### Logo click

With Free to use only checked and none of the other check boxes.
clicking on the Grid logo
Free to use only remains checked
and the next click to any of the check boxes is broken.

https://github.com/user-attachments/assets/a60cef9b-909e-4a88-a260-2122a94d1f47

Free to use only should have reset to empty.
The other checkboxes should have worked as normal on the next click.

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/kahuna/public/js/search/query.js
+++ b/kahuna/public/js/search/query.js
@@ -246,8 +246,11 @@ query.controller('SearchQueryCtrl', [
     // eslint-disable-next-line complexity
     function watchSearchChange(newFilter, sender) {
       let showPaid = newFilter.nonFree ? newFilter.nonFree : false;
-      if (sender && sender == "filterChange" && !newFilter.nonFree) {
-        showPaid = ctrl.user.permissions.showPaid;
+      if (ctrl.usePermissionsFilter) {
+        // Fixes Free to use only check-box is cleared on Back to search
+        if (sender && sender === "filterChange" && !newFilter.nonFree) {
+          showPaid = ctrl.user.permissions.showPaid;
+        }
       }
       storage.setJs("isNonFree", showPaid, true);
 
@@ -318,7 +321,7 @@ query.controller('SearchQueryCtrl', [
     }
 
     ctrl.sortProps = {
-      onSortSelect: updateSortChips,
+      onSelect: updateSortChips,
       query: $stateParams.query,
       orderBy: ctrl.ordering ? ctrl.ordering.orderBy : ""
     };


kahuna/public/js/search/query.js
```diff
@@ -246,8 +246,11 @@ query.controller('SearchQueryCtrl', [
     // eslint-disable-next-line complexity
     function watchSearchChange(newFilter, sender) {
       let showPaid = newFilter.nonFree ? newFilter.nonFree : false;
-      if (sender && sender == "filterChange" && !newFilter.nonFree) {
-        showPaid = ctrl.user.permissions.showPaid;
+      if (ctrl.usePermissionsFilter) {
```
tonytw1, line 249: Using the feature switch to bypass this new block restored the Back to search behaviour.

kahuna/public/js/search/query.js
```diff
@@ -318,7 +321,7 @@ query.controller('SearchQueryCtrl', [
     }
 
     ctrl.sortProps = {
-      onSortSelect: updateSortChips,
+      onSelect: updateSortChips,
```
tonytw1, line 324: Using bi-section, 
reverting this line restored the logo click behaviour.

No idea why this is effective.
It could be something todo with the path into `watchSearchChange` been somehow bound to the state of the sort control?

paperboyo:
I can repro Back to search bug when my permissions are set to have Free to use unchecked and I check it first. Revised steps:
1. Check Free to use
2. Click into image
3. Click Back to search
Actual: Free tu use is unchecked
Expected: Free tu use always keeps its state after Back to search. Here it should be checked.

I can't repro the above with default permissions (whe my Free to use is checked). But in that state, to uncheck it, I need to click twice (sic!).

I can repro Logo click with ANY permission.

It would be great to fix both.

AlexAscherson:
Hey folks!  Would it be possible to resolve the merge conflicts for this PR? Looking to do some testing this week if we feel its ready to go?  We are not expecting to update the branch from main I gather from this comment:

Disclaimer; we are forked quite away from the main Grid so this may not reflect want happens in the main branch.
Provided FYI.

Just wanted to check that was right.

paperboyo:
@AlexAscherson My understanding is that this here PR should be closed as dead/research. Real work (also needing main merge) is here: https://github.com/guardian/grid/pull/4730.

And the PR we use for testing the above one (only so that CI is trigerred; to be closed when real one is merged) is here: https://github.com/guardian/grid/pull/4786.

PR #4663, opened 2026-03-14
Getty supplierProcessor cleans more

paperboyo:
Co-authored by Claude Opus 4.6. PR Description written by a human, 48.

## What does this change?

Following from https://github.com/guardian/grid/pull/4615, this attempts cleanup of Getty metadata:
- it removes `(Photo by/Photo credit should read [Byline]/[Credit(s)])` template from Description only when Byline and Credit(s) match in their respective fields (we ignore diacritics in matching)
- it removes variations of `Location1, Location2 - human date: ` from the Description. It only does it when ONE of the locations match (to account for TX/Texas and Turkey/Turkiye). It matches date strings to dateTaken and allows for one day variation within same month/year (to account for timezones, but not when month or year would clash)
- finally (finally!), it EXTRACTS Byline from Description string for when Getty only provides it there and runs usual Byline cleaners on it (because normally they run earlier on the usual Byline fields)

The idea is: better not to clean than clean too greedily. This should first and foremost be safe.

Let's hear from a human if increasingly involved commits do not pass the vibe-craze threshold. I think it's all useful, maybe extractor even most useful (those missing Bylines are a common occurrence on the site).

image

Here are good media ids to test on:
`search?nonFree=true&ids=0d458bc8819e806691f3f5283a29921359accbbb,4760414a441b59b452357fe7f284f239be858b4e,d1b0e4d58b99339d389c24cce68148fec5e0ad45,f2564b9840b462bacc9d13998845903dde3fb247,684a161fc6229aa11423090556cc71fbc56c6e4c,4f18808b8a0b9164e498f8e9cd811461905253e5,3fdbb73ddf1dba0aa9a547c917658c86716e1f09,3d61535686e1b6ed27dbf54eff18760e85d2c965,39c2163e54d5b1563347d78fa8d660670d9322f3,b4869673db4902274e50a6db8d4e1b14bf18b05e,65538756af81c296f0cc29068226975ec7c01580,f06c9abe4495803a68faef8410a91f433cadaeee,6c4fed0921e56dc8ad7ef21ae95b173e6810618a`

Less manual cleanup of Descriptions in Composer.
Finally, no need to manually extract Bylines for certain intermediaries, which almost nobody does anyway (so: less missing photographer names).

## Who should look at this?
You asked for it, robot: assigning.

## Tested? Documented?
- [x] locally by committer

---
  - https://app.asana.com/0/0/1216165727279893


common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -325,15 +327,194 @@ object GettyXmpParser extends ImageProcessor {
...
+        }
+        else description
+      case None => description
+    }
+  }
+
+  // Matches: "LOC1, LOC2 - MONTH DAY:" or "LOC1, LOC2 - MONTH DAY, YEAR:" or "LOC1, LOC2 - MON DAY, YEAR - "
+  private val LocationDatePrefix = """(?i)^([^,]+),\s+(.+?)\s+-\s+([A-Za-z]+)\s+(\d{1,2})(?:,\s*(\d{4}))?\s*(?::\s*|-\s+)""".r
+
+  private val monthNumbers: Map[String, Int] = Map(
```
davidfurey, line 379: Surely this is available in Joda?

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -325,15 +327,194 @@ object GettyXmpParser extends ImageProcessor {
     Agencies
       .getWithCollection("getty", suppliersCollection)
 
+  private val PhotoByPattern = """(?i)\s*\(Photo (?:by|credit should read)\s+(.+?)\s*/\s*(.+?)\)\s*(.*)$""".r
+
+  private def normalise(s: String): String = {
```
davidfurey, line 346: I think this deserves a comment explaining that it strips accent/diacritic marks

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -325,15 +327,194 @@ object GettyXmpParser extends ImageProcessor {
...
+    *
+    * Credit in the description may use "via" as a separator (e.g. "AFP via Getty Images")
+    * while the metadata credit uses "/" (e.g. "AFP/Getty Images"), so we normalise both before comparing.
+    */
+  def cleanDescription(description: String, byline: Option[String], credit: Option[String]): String = {
+    PhotoByPattern.findFirstMatchIn(description) match {
+      case Some(m) =>
+        val descByline = m.group(1)
+        val descCredits = m.group(2)
+        val trailing = m.group(3).trim
```
davidfurey: I wonder whether instead of sometimes matching on `\s*` in the regex and sometimes using `.trim`, you could consistently use the regex when you need to trim strings?

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -325,15 +327,194 @@ object GettyXmpParser extends ImageProcessor {
     Agencies
       .getWithCollection("getty", suppliersCollection)
 
+  private val PhotoByPattern = """(?i)\s*\(Photo (?:by|credit should read)\s+(.+?)\s*/\s*(.+?)\)\s*(.*)$""".r
```
davidfurey:
```suggestion
private val PhotoByPattern = """(?i)^\s*(.*?)\s*\(Photo (?:by|credit should read)\s+(.+?)\s*/\s*(.+?)\)\s*(.*?)\s*$""".r
```

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
davidfurey:
```suggestion
    PhotoByPattern.findFirstMatchIn(description).map(_.subgroups) match {
      case Some(before :: descByline :: descCredits :: trailing :: Nil) =>
```

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -325,15 +327,194 @@ object GettyXmpParser extends ImageProcessor {
...
+
+        // Normalise "via" → "/" so "AFP via Getty Images" matches credit "AFP/Getty Images"
+        val normalisedDescCredits = descCredits.replaceAll("(?i)\\s+via\\s+", "/")
+        val creditMatches = credit.exists(c => normalise(c) == normalise(normalisedDescCredits))
+
+        if (bylineMatches && creditMatches) {
+          val before = description.substring(0, m.start).trim
+          if (trailing.nonEmpty) s"$before $trailing".trim else before
+        }
+        else description
```
davidfurey, line 371:
```suggestion
if (bylineMatches && creditMatches) {
          List(Some(before), Some(trailing)).filter(_.nonEmpty).mkString(" ")
        }
        else description
```

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -325,15 +327,194 @@ object GettyXmpParser extends ImageProcessor {
...
+        else None
+        val loc1Matches = loc1Direct || loc1SuffixField.isDefined
+        val loc2Matches = locationFields.exists(_.equalsIgnoreCase(loc2))
+        val locationMatches = loc1Matches || loc2Matches
+
+        val dateMatches = for {
+          dt <- dateTaken
+          month <- monthNumbers.get(monthStr)
+        } yield {
+          // Allow ±1 day for timezone differences, but only within the same month
```
davidfurey, line 427: Do timezones suddenly not exist at the start/end of the month?

If I take a picture on 2025-12-31T23:00:00Z but am in Sydney, it is entirely reasonable for the year, month and day not to match.

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -325,15 +327,194 @@ object GettyXmpParser extends ImageProcessor {
...
+          val monthMatch = dt.getMonthOfYear == month
+          val dayMatch = Math.abs(dt.getDayOfMonth - day) <= 1
+          val yearMatch = yearOpt.forall(_ == dt.getYear)
+          monthMatch && dayMatch && yearMatch
+        }
+
+        if (locationMatches && dateMatches.getOrElse(false)) {
```
davidfurey:
```suggestion
        if (locationMatches && dateMatches.contains(true)) {
```

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
```diff
@@ -325,15 +327,194 @@ object GettyXmpParser extends ImageProcessor {
...
+          val rest = description.substring(m.end)
```
davidfurey: If you need contents of the string after the match, I think it would be easier to read if you modify the regex to finish with `(.*)$` or similar rather than writing `description.substring(m.end)`

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
paperboyo: If I can even reconstruct my (always) convoluted train of thought, this was to prevent removal of eg. human-supplied date bit from `January 1 2077 – Cyberpunk 8 debuts in the New Year…` and leaving Date Taken which is potentially in the old year from the users’ POV (remember, Grid is stupid about timezones: we flatten than to UTC on arrival which is destructive in maybe exactly similar context and then display in user’s browser tz. That latter part I may be wrong on). Or the other way around. Or something.

Will take another look. And massive thanks!

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
paperboyo: Available in Joda this surely is. But when 🤖 applied it, it was longer and it told me it’s 100% identical functionality. So I decided it’s clearer as silly strings, if only coz we can add another silly string more easily should we find it not being cleaned.

🤖:

```scala

// Joda version — needs two new imports (DateTimeFormat, Locale) and still
// requires a manual entry for "sept". Produces identical output, not shorter:

private val monthNumbers: Map[String, Int] = {
  val fullFormat = DateTimeFormat.forPattern("MMMM").withLocale(Locale.ENGLISH)
  val shortFormat = DateTimeFormat.forPattern("MMM").withLocale(Locale.ENGLISH)
  (1 to 12).flatMap { m =>
    val dt = new DateTime(2000, m, 1, 0, 0)
    List(
      fullFormat.print(dt).toLowerCase -> m,
      shortFormat.print(dt).toLowerCase -> m
    )
  }.toMap + ("sept" -> 9)
}

// vs. the explicit map — no extra imports, self-documenting, same 24 entries:

private val monthNumbers: Map[String, Int] = Map(
  "january" -> 1, "jan" -> 1, "february" -> 2, "feb" -> 2,
  "march" -> 3, "mar" -> 3, "april" -> 4, "apr" -> 4,
  "may" -> 5, "june" -> 6, "jun" -> 6, "july" -> 7, "jul" -> 7,
  "august" -> 8, "aug" -> 8, "september" -> 9, "sep" -> 9, "sept" -> 9,
  "october" -> 10, "oct" -> 10, "november" -> 11, "nov" -> 11,
  "december" -> 12, "dec" -> 12
)
```

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
paperboyo: Mentioned in PR description, but fair point. It’s only used in comparisons, so on rare occasion of one field having more correct data we will still match (seen it happen).

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
paperboyo: 🤖  agrees and addressed. On wider application, it “argued” this, though, about one specific place:

> However, this is a case where applying the reviewer's principle gets awkward — `loc1` is later used for .substring arithmetic (line 425: `loc1.substring(0, loc1.length - f.length).trim)`, so you'd need the trimmed value as a working string regardless. Pushing trimming into the regex here would mean wrapping groups in `\s*(...)\s*` which makes the already-complex regex harder to read.

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
paperboyo: 🤖 likes how this comment goes in unison with the spirit of the other comments too:

Agreed. Two improvements:

1. **Anchors `^...$`** make the regex match the full string rather than searching within it — safer and more intentional.
2. **Capturing `before` as a regex group** instead of extracting it via `description.substring(0, m.start)` keeps all four parts self-contained in the pattern, and enables the subgroups destructuring and simpler join from the follow-up comments.

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
paperboyo: 🤖:

> It's a clear win — you go from opaque numbered groups (m.group(1), m.group(2), etc.) to named bindings in a single line. If someone later adds or reorders a group in the regex, the numbered version silently extracts wrong data; the destructured version fails to compile.

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
paperboyo: 🤖:

> Handles all cases uniformly without branching. Improvement.
One note: the reviewer's actual suggestion used `List(Some(before), Some(trailing))` which would be a bug — `Some("").nonEmpty` is `true` (it's a non-empty `Option`), so empty strings would pass the filter, and `mkString` on `Option` values would produce `"Some(hello)"`. I applied the corrected version with plain strings instead.

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
paperboyo: 🤖:

> It's a minor style preference.


Well, not mine. But neither it is not.

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
paperboyo: 🤖:

> Same reasoning as earlier — self-contained pattern, no reliance on match position offsets

paperboyo:
@davidfurey Thank you so much for the comments. I hope I addressed all. Wait. Not I. Robot addressed (most of) them. I ran tests. Will check TEST too.

BTW, we should allow robots talking to robots here directly, because I felt like a robot having to copy paste only because we castrated them from having github access.

common-lib/src/main/scala/com/gu/mediaservice/lib/cleanup/SupplierProcessors.scala
paperboyo: Obviously, can do Joda if you still think that’s better!

PR #4692, opened 2026-03-31
Add cohere v4 mapping

alexduf:
I'm picking 256 dimensions after we tested the quality of results in the llm-playground and it looks more than sufficiant. According to my (not very scientific) math we should save on RAM by a factor of 6, where we'll be using about 1GB of RAM per node instead of 6GB or RAM per node.

In this PR I'm also implementing the writing of the cohere v4 embed from thrall.

## Tested? Documented?
- [x] locally by committer


common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
```diff
@@ -83,18 +83,31 @@ object Mappings {
...
         indexOptions = Some(DenseVectorIndexOptions(
-          `type`=DenseVectorField.Int8Hnsw,
-          m=Some(16),
-          efConstruction=Some(100)
+          `type` = DenseVectorField.Int8Hnsw,
+          m = Some(16),
+          efConstruction = Some(100)
+        ))
+      ),
+      nonDynamicObjectField("cohereEmbedEnglishV4-1536").copy(properties = Seq(
```
ellenmuller: Not "cohereEmbedEnglishV4"?

ellenmuller (approved):
Nice! Before you merge, we'll have to update the mapping. We should do on TEST first to check it's all okay, then on PROD (potentially with @andrew-nowak on the call as well because it's good to have extra eyes on it. Getting it wrong means a whole reindex)

Happy to pair on it if you want to put some time in my calendar 😄

common-lib/src/main/scala/com/gu/mediaservice/lib/elasticsearch/Mappings.scala
alexduf: I thought putting the number of dimension would allow us to add another field if we decided to change the number of dimensions later

alexduf:
Locally I've applied the mapping using the UpdateMapping script, no error, got
```
{
  "acknowledged" : true
}
```

i've then tested two cases:
- testing with the existing lambda (cohere v3 only) and checked the resulting document: had just the embedding v3, no error
- Tested a modified version of the lambda which published a dummy v4 alongside the v3 embedding: no error, embedding appeared in the document.

Next steps:
- [x] Update the mapping in TEST. It should be backwards compatible and I'll re-verify this
- [x] Deploy this branch in TEST and test pushing an embedding v4 by directly writing to kinesis
- [x] Update the mapping in PROD
- [x] Merge and deploy

alexduf:
I've also re-tested the backfill to ensure it accepts working with the `cohereEmbedV4` field

ellenmuller (approved):
Looks good! I'll update the Thrall message as part of my PR :)

alexduf:
Thanks Ellen!

PR #4693, opened 2026-04-01
Override the version of netty to use fixed vuln

lindseydew:
## What does this change?

This overrides the version of netty to make sure it uses the patched version of that library. According to: https://github.com/guardian/grid/security/dependabot/485

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -34,9 +34,6 @@ lazy val jacksonOverrides = Seq(
   "com.fasterxml.jackson.core" % "jackson-annotations" % jacksonAnnotationsVersion
 )
 
-val nettyOverrides = Seq(
-  "io.netty" % "netty-codec-http2" %  "4.1.132.Final"
-)
 val commonSettings = Seq(
   scalaVersion := "2.13.18",
   description := "grid",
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -54,7 +57,7 @@ val commonSettings = Seq(
     "org.mockito" % "mockito-core" % "2.18.0" % Test,
     "org.scalamock" %% "scalamock" % "5.1.0" % Test,
   ),
-  dependencyOverrides ++= jacksonOverrides,
+  dependencyOverrides ++= jacksonOverrides ++ nettyOverrides,
```
rtyley: Might be better to use `libraryDependencies`! See:

* https://github.com/guardian/maintaining-scala-projects/issues/20

build.sbt
```diff
@@ -101,6 +101,7 @@ lazy val commonLib = project("common-lib").settings(
     "com.amazonaws" % "aws-java-sdk-sts" % awsSdkVersion,
     "com.amazonaws" % "aws-java-sdk-dynamodb" % awsSdkVersion,
     "com.amazonaws" % "aws-java-sdk-kinesis" % awsSdkVersion,
+    "io.netty" % "netty-codec-http2" %  "4.1.132.Final",
```
rtyley: An alternative change - that has been just valid for the past 20 hours (!) - is actually just to update `awsSdkVersion` !

https://github.com/aws/aws-sdk-java-v2/releases/tag/2.42.25 includes this:

* https://github.com/aws/aws-sdk-java-v2/pull/6829 

...so updating AWS SDK from 2.32.33 to 2.42.25 will also fix this vuln!

lindseydew: Ahhh. I thought the latest version was 2.42.24 which did not bring in the latest version. 


But I tested 2.42.25 and you're right, it does pull it in 🥳

andrew-nowak (approved):

PR #4694, opened 2026-04-02
Embed images with Cohere V4, downscaled at 30_000 pixels and with 256 dimensions to ES, 1536 in S3 Vector Store

ellenmuller:
## What does this change?

This PR updates the image embedder to use Cohere V4 embeddings and introduces a new S3 vector bucket using the new CDK construct:

- **New embedding model**: Migrates from Cohere V3 to V4, storing the full 1536-dimensional embeddings in an S3 vector index in eu-west-1
- **Elasticsearch optimisation**: Adds a matryoshka truncation function to reduce embeddings to 256 dimensions for Elasticsearch ingestion via Kinesis
- **Updated message schema**: `updateEmbeddingMessage` now updates with Cohere V4
- **Image downscaling**: Increases downscaling to 30,000 pixels, after some analysis we saw that this doesn't significantly harm performance but allows us a workable throughput for the lambda whilst staying under token limits
- **Cache validation**: We now check the cache checking to see if cached images are not based on the old limits and we automatically re-downscale and re-cache them if so 

**Manual testing**: 
   - Send test images through the embedder and verify V4 embeddings appear in the S3 vector store with 1536 dimensions
   - Confirm truncated 256-dimensional embeddings reach Elasticsearch via Kinesis messages


image-embedder-lambda/src/embedder/constants.ts
```diff
@@ -7,4 +7,6 @@ export const MAX_IMAGE_SIZE_BYTES = 5 * 1024 * 1024;
 // but Cohere v4 downsamples images above this pixel count:
 // https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-embed-v4.html#:~:text=Image%20sizing%3A-,Images%20%3E%202%2C458%2C624,-pixels%20are%20downsampled
 // We use the lower bound of the models we can immediately foresee using.
-export const MAX_PIXELS_COHERE_V4 = 2_458_624;
+export const MAX_PIXELS_COHERE_V4 = 30_000;
```
paperboyo, line 10: I was always bad at maths, so apols if silly q, but are we dropping res from ~1920×1280px to ~210×140px (example for 3:2 ratio)? This seems positively tiny. Less than my Atari 800XL...

Won't `More like this` mix all the actors into one red carpet soup?

alexduf: We were the first to be surprised, but we first eyeballed it and it looks suprisingly good! Still in disbelief, Alice decided to back this up with data, and yes, 30k pixels is all it takes to get decent embedding, and decent enough search result.
That's a 10 fold cost save right there.

cdk/lib/image-embedder-lambda.ts
```diff
@@ -26,6 +27,36 @@ export class ImageEmbedder extends GuStack {
...
+    new CfnIndex(this, 'CohereV4Index', {
+      dataType: 'float32',
+      dimension: 256,
+      distanceMetric: 'cosine',
+      indexName: 'cohere-embed-english-v4',
+      vectorBucketName: `image-embeddings-${this.stage.toLowerCase()}`,
+      tags: [
+        {
+          key: "gu:cdk:version",
+          value: `${this.stage}`,
```
alexduf: Wrong value here, I think the tag itself can be dropped, I think (not sure) this is only use for gu-cdk constructs.

common-lib/src/main/scala/com/gu/mediaservice/model/Embedding.scala
```diff
@@ -11,6 +11,10 @@ case class Embedding(
     cohereEmbedV4: Option[CohereV4Embedding] = None
 )
 
+case class PartialEmbedding(
```
alexduf: Why is it "partial"? is it because it has fewer dimensions or because there's only v4?

In all cases, why not simply removing `cohereEmbedEnglishV3` from `Embedding` ?

image-embedder-lambda/src/embedder/imageEmbedder.ts
```diff
@@ -41,12 +41,15 @@ export async function embedImage(
...
-  const model = 'cohere.embed-english-v3';
+  const model = 'global.cohere.embed-v4:0';
+
   const body = {
-    input_type: 'image',
+    input_type: 'search_document',
     embedding_types: ['float'],
     images: [`data:${imageMimeType};base64,${base64Image}`],
-  };
+    // TODO we get the largest dimension now and then truncate to 256 for ES
```
alexduf: the todo can go

image-embedder-lambda/src/embedder/imageResolver.ts
```diff
@@ -95,20 +96,20 @@ export class CachedImageResolver implements ImageResolver {
 
     const key = this.toPartitionedKey(imageId);
     const bytes = await this.s3Fetcher.fetch(this.downScaledImageBucket, key);
-    if (bytes) {
+    if (bytes && bytes.length <= maxImageSizeBytes) {
```
alexduf: I think this is the wrong check here, we want to check the image size in pixels not in bytes.

This function should probably always return the downscaled version, it's the caller that will check the size of the image in pixels and decide if it's a cache hit or miss

image-embedder-lambda/src/embedder/s3VectorStore.ts
```diff
@@ -19,7 +19,7 @@ export class S3VectorStore {
 
     const input: PutVectorsCommandInput = {
       vectorBucketName: this.vectorBucketName,
-      indexName: 'cohere-embed-english-v3',
+      indexName: 'cohere-embed-english-v4',
```
alexduf: V4 is multi-langual
```suggestion
      indexName: 'cohere-embed-v4',
```

image-embedder-lambda/src/embedder/thrallEventPublisher.ts
```diff
@@ -63,7 +75,7 @@ export class ThrallEventPublisher {
         lastModified: new Date().toISOString(),
         id: v.key,
         embedding: {
-          cohereEmbedEnglishV3: {
+          cohereEmbedEnglishV4: {
```
alexduf: Same as above, this model isn't english specific
```suggestion
          cohereEmbedV4: {
```

alexduf (commented):
A few comments, these are all small things, the logic is solid otherwise

image-embedder-lambda/src/embedder/constants.ts
paperboyo: And faces work just as well too (surprised)? So, maybe it would only matter once we would want to do something like [this](https://github.com/retkowsky/visual-search-cohere-embed-v-4-0?tab=readme-ov-file#visual-search-using-text)?

Savings good! We can always up one day. And we will. 👍

common-lib/src/main/scala/com/gu/mediaservice/model/Embedding.scala
ellenmuller: We can't remove `cohereEmbedEnglishV3` from the `Embedding` class because it's actually the type used for the mapping! We've been using the same type for the mapping and update message because up to now they've been the same, but obviously now we're doing a partial update - That's why I added a new type! But thinking about it, because the fields are both options, we don't technically need a different type so I've decided to still just use `Embedding` :)

image-embedder-lambda/src/embedder/imageResolver.ts
```diff
@@ -96,19 +98,23 @@ export class CachedImageResolver implements ImageResolver {
...
     const bytes = await this.s3Fetcher.fetch(this.downScaledImageBucket, key);
     if (bytes) {
-      console.log(
-        `Cache hit: found downscaled image for ${imageId} (${bytes.length.toLocaleString()} bytes)`,
-      );
-    } else {
-      console.log(`Cache miss: no downscaled image for ${imageId}`);
+      const { width, height } = await sharp(bytes).metadata();
+      const pixels = width * height;
+      if (pixels <= maxImageSizePixels) {
```
alexduf, line 103: I don't think this needs to be fixed within this PR, but the check that the image is the correct size is done twice, once here, and once in `downscaleImageIfNeeded`.
This caching function should only worry about fetching whatever is in the cache, and the caller should be in charge of knowing what dimensions it expects.

alexduf (approved):
This looks good to me, nice work! 🥳 

I've got one comment, which can be a separate PR, once this is merged.

cdk/lib/image-embedder-lambda.ts
```diff
@@ -26,6 +27,18 @@ export class ImageEmbedder extends GuStack {
     const appName = 'image-embedder';
     const downscaledImageBucketName = `${this.stack}-${props.stage.toLowerCase()}-${appName}-downscaled-images`;
 
+    new CfnVectorBucket(this, 'GridEmbeddingsVectorBucket', {
+      vectorBucketName: `image-embeddings-${this.stage.toLowerCase()}`,
+    });
+
+    new CfnIndex(this, 'CohereV4Index', {
```
andrew-nowak, line 34: nit:
```suggestion
    const coherev4Index = new CfnIndex(this, 'CohereV4Index', {
```

cdk/lib/image-embedder-lambda.ts
```diff
@@ -173,7 +186,7 @@ export class ImageEmbedder extends GuStack {
       new PolicyStatement({
         actions: ['s3vectors:PutVectors'],
         resources: [
-          `arn:aws:s3vectors:eu-central-1:${Stack.of(this).account}:bucket/image-embeddings-${props.stage.toLowerCase()}/index/*`,
+          `arn:aws:s3vectors:${Stack.of(this).region}:${Stack.of(this).account}:bucket/image-embeddings-${props.stage.toLowerCase()}/index/*`,
```
andrew-nowak, line 189:
```suggestion
          coherev4Index.attrIndexArn,
```
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_s3vectors.CfnIndex.html#attrindexarn

image-embedder-lambda/src/embedder/thrallEventPublisher.ts
```diff
@@ -37,6 +39,16 @@ export class ThrallEventPublisher {
     this.stage = stage;
   }
 
+  matryoshkaEmbeddingToElasticsearchDimensions(vectors: ValidVector[]): ValidVector[] {
```
andrew-nowak, line 42: In a followup PR please can we add some comments/docs about matryoshka embeddings, how they work etc. for future travellers?

andrew-nowak (approved):
lgtm!

PR #4698, opened 2026-04-07
remove the cloudfront sdk

andrew-nowak:
## What does this change?

Removes the Cloudfront SDK and usage.

We need to migrate from v1 to v2 of all the SDK modules. Our only usage of the Cloudfront SDK is to sign thumbnail URLs. However we don't (and possibly never have) enforced verification of the signatures.

I've had a quick poke and it doesn't appear straightforward to start enforcing this, and since we've not done it in the past and it doesn't appear we're needing to do it soon, I think it might make most sense to just tear it out.

I'll leave a card suggesting reviewing and implementing signatures & verification, with a link to this PR which can serve as a starting point for implementing it properly with the new SDK if we decide to come back to it.

## Tested? Documented?
- [ ] locally by committer


bryophyta (approved):
lgtm!

PR #4700, opened 2026-04-09
on undeletion, mark image as archived so it cannot be immediately rereaped

andrew-nowak:
## What does this change?

Currently, when an image is reaped (because it has not been saved, edited or used in content in 3 weeks since it arrived on the Grid) it is initially "soft-deleted", and then later fully deleted if it hasn't been rescued for a further 2 weeks. To rescue it, a user must press the "undelete" button. This will immediately make it eligible for the initial reaping stage; if a user is unlucky they might press the button a couple of seconds before the reaper runs, meaning that the image is undeleted, and then almost immediately redeleted.

Removing the soft-deleted metadata does not mark the image in any lasting way, so we cannot update the "reapable eligibility criteria" to account for it. I feel like a user manually asking for the image to be undeleted is a sufficient signal that the image should be kept that we should mark the image as "archived" (remember in Grid that "archived" is not "soft-deleted"; it's "this image is important and we should keep it permanently, as part of the "archive"). This would mean that it is no longer eligible for reaping.

So, when we receive a request to undelete an image, the backend will first mark it as archived (doing this first to avoid any other race conditions with the reaper) before removing the deletion marker.

Use the "delete" button on an image, then find it with a "+is:deleted" search (requires edit_metadata permission). Press "undelete", and the image should then be marked as "archived"

image

^ an image which has been marked as archived ^

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-archiver/gr-archiver.js
+++ b/kahuna/public/js/components/gr-archiver/gr-archiver.js
@@ -132,7 +132,8 @@ module.controller('grArchiverCtrl', [
             mediaApi.undelete(imageId)
                 .then(() => pollUndeleted(imageId))
                 .then(() => {
-                    ctrl.canUndelete = ctrl.isDeleted = false;
+                    ctrl.canUndelete = false;
+                    ctrl.isDeleted = false;
                     ctrl.image.softDeletedMetadata = undefined;
                     if (ctrl.image.userMetadata?.data !== undefined) {
                         ctrl.image.userMetadata.data.archived = true;
--- a/kahuna/public/js/components/gr-archiver/gr-archiver.js
+++ b/kahuna/public/js/components/gr-archiver/gr-archiver.js
@@ -27,6 +27,7 @@ module.controller('grArchiverCtrl', [
     'humanJoin',
     'mediaApi',
     'getDeletedState',
+    'apiPoll',
     function ($scope,
               $window,
               archiveService,
@@ -34,7 +35,8 @@ module.controller('grArchiverCtrl', [
               imageLogic,
               humanJoin,
               mediaApi,
-              getDeletedState) {
+              getDeletedState,
+              apiPoll) {
 
         const ctrl = this;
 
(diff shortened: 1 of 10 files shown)


kahuna/public/js/components/gr-archiver/gr-archiver.js
```diff
@@ -115,14 +117,30 @@ module.controller('grArchiverCtrl', [
...
         function undelete() {
             ctrl.undeleting = true;
             const imageId = ctrl.image.data.id;
             mediaApi.undelete(imageId)
-                .then(
-                  ctrl.canUndelete = ctrl.isDeleted = false
-                ).catch(() => {
+                .then(() => pollUndeleted(imageId))
+                .then(() => {
+                    ctrl.canUndelete = ctrl.isDeleted = false;
```
twrichards: should this be 
```suggestion
                    ctrl.canUndelete = ctrl.isDeleted === false;
```
looks like its doing two assignments 🤔 

(appreciate that's how it was before, so would warrant some testing to figure out the original intent of that line)

twrichards (approved):
nice one @andrew-nowak  - I'd always planned to do something like this, but never got round to it I guess (as this was the intent of the soft-deletion phase of reaping).

Do we want to check the reapable eligibility before doing this additional archiving - so that regular undeletes (of non-reaped images) aren't unnecessarily archived.

Also, one notable comment...

kahuna/public/js/components/gr-archiver/gr-archiver.js
andrew-nowak: Yeah, I'm pretty sure this is a rare intentional double assignment (this assignment happens just after we've undeleted the image, and being angular the `isDeleted` and `canUndelete` states aren't derived from other variables, so needs explicitly setting back to `false`. 

I'll split it into two separate assignments to make the intention clearer

andrew-nowak:
> Do we want to check the reapable eligibility before doing this additional archiving - so that regular undeletes (of non-reaped images) aren't unnecessarily archived.

Personally I'm happy to take an "undelete" press as an indication that a user is very interested in keeping this image (as they have gone through the trouble of finding the soft-deleted image, which is not exactly easy to do), and it should be added to the archive so that it is never again a candidate for reaping, regardless of whether that was why it was originally soft-deleted.

twrichards:
> > Do we want to check the reapable eligibility before doing this additional archiving - so that regular undeletes (of non-reaped images) aren't unnecessarily archived.
> 
> Personally I'm happy to take an "undelete" press as an indication that a user is very interested in keeping this image (as they have gone through the trouble of finding the soft-deleted image, which is not exactly easy to do), and it should be added to the archive so that it is never again a candidate for reaping, regardless of whether that was why it was originally soft-deleted.

What about if the user accidentally deletes a regular image, then immediately un-deletes... do we really want it retained forever

PR #4701, opened 2026-04-09
Implement ai search tracking

alexduf:
## What does this change?

 

I'm adding a tag to the search telemetry event such that in our grafana dashboard we can specifically search for events relating to AI search.

## Tested? Documented?
- [ ] locally by committer


alexduf:
When deploying this branch I can see the event being sent, and it looks correct. However, the service that aggregates the events is returning 502. I think this is a fairly low risk change and it probably should be good enough to be merged.

andrew-nowak (approved):

PR #4703, opened 2026-04-10
Only backfill free images

alexduf:
## What does this change?
For context: The Guardian is looking at removing any image that isn't "free" from the grid. I'm using quotes for "free" as this is wider than a matter of cost, there's licensing and more involved, and I quite frankly haven't gone through the necessary effort to fully understand it.

This represents about 2M images that would be removed, about 25% of all images.

In parallel to this, we're currently implementing a semantic search which requires to compute an embedding for every single image we have in our database.

Because this is a significant amount of images, it makes sense to avoid computing the embedding for "non-free" images, saving both time and money.

How do we know if an image is "free" or not?
There's a fairly complex condition based on hard-coded values defined in :
 - [UsageRights.scala](https://github.com/guardian/grid/blob/67ec3cdd63ea26d60263fb4ccd787f7efbb4ca80/common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala)
 - [GuardianUsageRightsConfig.scala](https://github.com/guardian/grid/blob/59f7089a612b51d349ee60aed6c2d2d997cc8b18/common-lib/src/main/scala/com/gu/mediaservice/lib/guardian/GuardianUsageRightsConfig.scala)
 - [SearchFilter.scala](https://github.com/guardian/grid/blob/502cd2b3584320ca90684fba9d766608f0ff82b2/media-api/app/lib/elasticsearch/SearchFilters.scala)

Because these files are in scala it is unfortunately impossible to re-use the logic from the typescript lambda without compromising either runtime speed (via an API call) or engineering speed (by re-writing things). However, we know that the current state we're in (having both "free" and "non-free" images in the same place) is a temporary situation, meaning that whatever logic I end up implementing to avoid picking "non-free" images during backfill _SHOULD_ be a temporary measure.

With that in mind, I've decided to re-implement the business rule by hard-coding the same constants, and re-building the elastic search query directly in typescript.

To speed up the process, and for full transparency, I've used Claude Sonnet 4.6 to assist me, pointing it at the existing rules, the existing constants, as well as the existing elastic search query for enrichment. I then extracted it in a separate file making sure removing this temporary measure was as easy as possible.

## Tested? Documented?
- [ ] locally by committer


alexduf:
I still need to properly test this in the TEST environment, but awaiting for [this](https://github.com/guardian/grid/pull/4694) to be merged as it makes deploying easier.

alexduf:
Tested in TEST, the query works as expected

ellenmuller (approved):
Amazing, thanks!

PR #4705, opened 2026-04-13
Enable Backfill

alexduf:
## What does this change?
In this PR I re-enable the schedule to regularly find images that don't contain the embedding and send them for processing.

I've spent time this afternoon tweaking the various knobs that control how quickly we process the backfill, and I suspect we'll keep tweaking them for the forseable. But at least they're in the right ballpark

I went as high as 40 parallel instances of the embedder lambda, each processing 5 images per execution, and this brought us to 395K tokens per minute. I brought it down to 20 for backfilling only, 5 for production which should be enough.

Depending on how the backfill goes this week I might bring these numbers further down to make sure we have enough headroom.

## Tested? Documented?
- [ ] locally by committer


cdk/lib/image-embedder-lambda.ts
```diff
@@ -12,12 +12,27 @@ import {
...
 import {Architecture} from 'aws-cdk-lib/aws-lambda';
 import {SqsEventSource} from 'aws-cdk-lib/aws-lambda-event-sources';
 import {CfnIndex, CfnVectorBucket} from 'aws-cdk-lib/aws-s3vectors';
 import { Queue } from 'aws-cdk-lib/aws-sqs';
 
+const CONCURRENCY: Record<string, { backfill: number; loader: number; total: number }> = {
+	'PROD': {
+		backfill: 20,
+		loader: 5,
+		total: 35,
```
andrew-nowak: Should this be 25? (Why do you specify this as a separate number, and not backfill + loader?

andrew-nowak (approved):

cdk/lib/image-embedder-lambda.ts
alexduf: Good call, I went through multiple rounds of updating these values and they got our of sync. Thanks!

cdk/lib/image-embedder-lambda.ts
```diff
@@ -12,12 +12,27 @@ import {
...
+	'PROD': {
+		backfill: 20,
+		loader: 5,
+		total: 25,
+	},
+	'TEST': {
+		backfill: 5,
+		loader: 5,
+		total: 10,
+	}
```
ellenmuller, line 32: We probably want to bring the TEST concurrency down to 0 when we merge, just so that they're not competing for Bedrock tokens?

alexduf:
Scaling up the lambda execution revealed a file descriptor leak that took me a long time to understand.

For any future traveler, here's what I've done:
- I traced all the file descriptors that were being used, by [vibecoding some debug traces](https://github.com/guardian/grid/compare/adu-enable-backfill-debug?expand=1#diff-397e4ceb75360b8fbc6da8746a5eab2917d372362646bbfe527c6e5d488a1c34R78-R101)
- This revealed that the file descriptors weren't being used by the image transformation library as I initially thought, but by the AWS sdk, which was surprisingly leaking connections, despite re-using the client across uses!

The solution is simply to limit the number of TCP connections in the connection pool to 5, but I do believe there's a bug in the AWS library.

I haven't investigated deeper into the library, I'm assuming this is probably fixed, or will be shortly, but I'll keep the 5 connections max just to be on the safe side.

cdk/lib/image-embedder-lambda.ts
alexduf: Now that I've finished my tests I've disabled the schedule that pushes new items to the sqs queue.
In TEST it's still the version that pushes items to the backfilling queue, and I'll leave it overnight to make sure that we don't have any more issues.

ellenmuller (approved):

PR #4708, opened 2026-04-20
Use Elasticsearch for semantic search query

ellenmuller:
### Migrate AI image search to Elasticsearch KNN with pagination and caching

Replaces S3 Vectors with Elasticsearch KNN queries for AI search, using the `embedding.cohereEmbedV4.image` dense vector field that is now backfilled in ES!

**ElasticSearch.scala**: Added`knnSearch` using elastic4s `Knn` against the stored image embeddings. Use `offset`/`length`/`countAll` parameters for proper pagination. Total hits are capped to `min(totalHits, k)` to keep result counts predictable across pages.

**MediaApi.scala**: Wired pagination through both AI search paths:
- Text search — embeds the query via Bedrock then runs paginated KNN.
- Image-to-image search (`similar:<imageId>`) — fetches the source image's stored embedding then runs paginated KNN.

**`Bedrock.scala`** / **`Embedder.scala`**: Updated to Cohere Embed v4 (256-dim), with a `createQueryEmbedding` helper for query-time embedding.

## Decoupled KNN k from page size

Added configurable `aiSearchSemanticRetrievalWindow` (default 1000) in `MediaApiConfig`. This `k` value is independent of user-requested page size, allowing efficient retrieval of a bounded candidate set while supporting arbitrary page boundaries. Prevents the bootstrap problem where fetching only 1 result would poison the total count.

## Embedding Cache

Added process-local Guava cache for query embeddings, keyed on normalized (trimmed, lowercased) query text. Avoids repeated Bedrock calls on pagination and deduplicates in-flight requests—concurrent identical queries share a single Bedrock call. Configurable via `ai.search.embeddingCache.maxSize` (default 500) and `ai.search.embeddingCache.ttlSeconds` (default 600). Includes cache hit/miss logging keyed by query hash for metrics.

## Fixed Bootstrap Bug

**results.js**: AI mode now uses a real first-page fetch instead of the lexical `length:1` probe, which was causing "1 matches" display with 2+ results visible.

**results.js**: Skip the 15-second polling loop when `useAISearch` is active to avoid unnecessary Bedrock calls.

 

## Tested? Documented?
- [ ] locally by committer


image-loader/app/ImageLoaderComponents.scala
```diff
@@ -33,11 +33,7 @@ class ImageLoaderComponents(context: Context) extends GridComponents(context, ne
   val maybeEmbedder: Option[Embedder] = config.maybeImageEmbedderQueueUrl
     .filter(_ => config.shouldEmbed)
     .map {queueUrl =>
-      new Embedder(
-        new S3Vectors(config),
-        new Bedrock(config),
-        new SimpleSqsMessageConsumer(queueUrl, config)
-      )
+      new Embedder(new Bedrock(config), new SimpleSqsMessageConsumer(queueUrl, config))(ec)
```
alexduf: any particular reason to explicitly pass the execution context here?

kahuna/public/js/search/results.js
```diff
@@ -282,36 +334,12 @@ results.controller('SearchResultsCtrl', [
...
-
-                    ctrl.imagesAll[position] = image;
-                    imagesPositions.set(imageId, position);
-
-                    results.set(position, image);
-                });
-
-                // images should not contain any 'holes'
-                ctrl.images = compact(ctrl.imagesAll);
+            // Reuse the same merge logic for all subsequent paged fetches.
```
alexduf: nitpick, but this comment only makes sense in the context of this PR, in the context of the code flow it's not immediately clear why applyLoadedImages is re-used

media-api/app/controllers/MediaApi.scala
```diff
@@ -564,55 +575,147 @@ class MediaApi(
...
+      // Normalise key so that "Dogs" and "dogs " share a cache entry.
+      val cacheKey = query.trim.toLowerCase
+      val queryHash = Integer.toHexString(cacheKey.hashCode)
+      val cachedEmbeddingFuture = Option(embeddingCache.getIfPresent(cacheKey))
+
+      cachedEmbeddingFuture.foreach { _ =>
+        logger.info(
+          logMarker,
+          s"AI search embedding cache hit queryHash=$queryHash queryLength=${query.length} cacheSize=${embeddingCache.size()}"
+        )
```
alexduf: So is all this only for logging? 
and why logging the query hash rather than the query itself? If it were to avoid storing the search terms, we already store the content of the query in the same ELK stack for telemetry purposes

And I think we probably want this to be a debug log level right?

media-api/app/controllers/MediaApi.scala
```diff
@@ -564,55 +575,147 @@ class MediaApi(
...
       }
-      else {
-        _searchParams
+
+      // cache.get(key, callable) is atomic: if two requests race on the same key, only one
+      // Callable fires and both callers receive the same Future.
+      val embeddingFuture = embeddingCache.get(cacheKey, new Callable[Future[List[Float]]] {
+        def call(): Future[List[Float]] = {
+          logger.info(
```
alexduf: same here, probably a debug level

media-api/app/controllers/MediaApi.scala
```diff
@@ -564,55 +575,147 @@ class MediaApi(
...
+
+    val searchParams = if(canViewDeletedImages) {
+      _searchParams.copy(uploadedBy = Some(Authentication.getIdentity(request.user)))
     }
+    else {
+      _searchParams
+    }
+
+    SearchParams.validate(searchParams).fold(
+      // TODO: respondErrorCollection?
```
alexduf: what do you want to do with this TODO?

I think you can either make it a card or get rid of it

media-api/app/MediaApiComponents.scala
```diff
@@ -29,7 +29,7 @@ class MediaApiComponents(context: Context) extends GridComponents(context, new M
   val imageResponse = new ImageResponse(config, s3Client, usageQuota)
 
   val softDeletedMetadataTable = new SoftDeletedMetadataTable(config)
-  val embedder = new Embedder(new S3Vectors(config), new Bedrock(config), new SimpleSqsMessageConsumer(config.queueUrl, config))
+  val embedder = new Embedder(new Bedrock(config), new SimpleSqsMessageConsumer(config.queueUrl, config))(ec)
```
alexduf: Same question here about the execution context, any reason you're passing it explicitly here?

alexduf (approved):
The logic and implementation are solid in my view, the only comments I have are pure nitpicking and questions.
Great work!

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/Embedder.scala
```diff
@@ -10,36 +10,19 @@ import java.nio.file.{Files, Path}
...
-class Embedder(s3vectors: S3Vectors, bedrock: Bedrock, sqs: SimpleSqsMessageConsumer)(implicit ec: ExecutionContext) extends GridLogging {
-  def mapCohereResponseToImageIds(response: QueryVectorsResponse): ImageIds = {
-    val results: java.util.List[QueryOutputVector] = response.vectors()
-    ImageIds(results.asScala.map(_.key()).toList)
-  }
+class Embedder(bedrock: Bedrock, sqs: SimpleSqsMessageConsumer)(implicit ec: ExecutionContext) extends GridLogging {
 
-  def createEmbeddingAndSearch(query: String)(implicit logMarker: LogMarker): Future[ImageIds] = {
+  def createQueryEmbedding(query: String)(implicit logMarker: LogMarker): Future[List[Float]] = {
     logger.info(logMarker, s"Searching for image embedding for query: $query")
```
andrew-nowak:
```suggestion
    logger.info(logMarker, s"Creating text embedding for query: $query")
```

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/Bedrock.scala
andrew-nowak:
```suggestion
        s"Successfully extracted text embedding. Vector size: ${embedding.size}"
```

media-api/app/controllers/MediaApi.scala
```diff
@@ -51,6 +53,15 @@ class MediaApi(
   val services: Services = new Services(config.domainRoot, config.serviceHosts, Set.empty)
   val gridClient: GridClient = GridClient(services)(ws)
 
+  // Process-local cache keyed on normalised query text. Stores the Bedrock Future so that
+  // concurrent requests for the same query share a single in-flight Bedrock call, and
+  // subsequent requests within the TTL window skip Bedrock entirely.
+  private val embeddingCache: Cache[String, Future[List[Float]]] =
+    CacheBuilder.newBuilder()
+      .maximumSize(config.aiSearchEmbeddingCacheMaxSize)
+      .expireAfterWrite(config.aiSearchEmbeddingCacheTtlSeconds, TimeUnit.SECONDS)
```
andrew-nowak: Strongly suggest not using Guava cache for async caching. This will probably work, but since it won't understand Scala Futures, if the Future fails to get an embedding out of Bedrock then Guava won't understand that there's a failure, and will store the failed future in the cache. scaffeine behaves similarly to guava's cache, but has modes that understands scala futures much better. I've thrown together this PR which should show how to implement it (though I've not tested!) https://github.com/guardian/grid/pull/4716

andrew-nowak: I'd also consider only using the `maximumSize` parameter, rather than `expireAfterWrite`. The embedding retrieved from Bedrock (if successful) shouldn't become less valid over time, so why not just leave it in the cache for as long as there's space for it?

andrew-nowak (commented):
part way through

image-loader/app/ImageLoaderComponents.scala
ellenmuller: Fair point, not really! Removing :)

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -173,6 +174,43 @@ class ElasticSearch(
       }
   }
 
+  def knnSearch(queryEmbedding: List[Float], k: Int, offset: Int, length: Int, countAll: Boolean, numCandidates: Int)
+               (implicit ex: ExecutionContext, logMarker: LogMarker): Future[SearchResults] = {
+    if (length <= 0 || k <= 0) {
```
andrew-nowak: Is it also valid and worth short-circutiing on `k < offset`?

andrew-nowak (approved):

alexduf (approved):
Still looks good to me. I've noticed the polling call was still being done, but that's not a blocker. We should address it at some point though

PR #4719, opened 2026-05-01
WIP S3 SDK upgrade

andrew-nowak:
## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
@@ -9,7 +9,7 @@ import software.amazon.awssdk.core.ResponseInputStream
 import software.amazon.awssdk.core.sync.RequestBody
 import software.amazon.awssdk.regions.Region
 import software.amazon.awssdk.services.s3.S3Client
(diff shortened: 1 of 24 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
```diff
@@ -46,97 +54,97 @@ object S3Object {
...
   type Key = String
   type UserMetadata = Map[String, String]
 
-  lazy val client: AmazonS3 = S3Ops.buildS3Client(config)
-  // also create a legacy client that uses v2 signatures for URL signing
-  private lazy val legacySigningClient: AmazonS3 = S3Ops.buildS3Client(config, forceV2Sigs = true)
+  lazy val client: S3Client = S3Ops.buildS3Client(config)
+  lazy val presigner = S3Presigner.create()
 
-  def signUrl(bucket: Bucket, url: URI, image: Image, expiration: DateTime = cachableExpiration(), imageType: ImageFileType = Source): String = {
```
andrew-nowak, line 70: turns out nothing's ever caching the presigned urls, so no need to round to make cacheable?
(if we do need this, it's annoying as v2 sdk only allows you to specify an expiry **duration**, not an expiry time, and expiry time is calculated internally against a call to `Instant.now()`, so impossible to guarantee urls match!)

PR #4722, opened 2026-05-05
Stop repeat AI search requests on scroll

ellenmuller:
## What does this change?
This PR stops repeated AI search requests from firing as users scroll through AI search results.

Although the 15-second polling path was already disabled for AI search, the search page was still issuing follow-up range requests via the lazy table as users scrolled. Because the AI search backend returns a single fixed result set, those range requests caused the same AI search to be re-run unnecessarily and led to repeated image reloads in the network panel.

 

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/kahuna/public/js/search/results.js
+++ b/kahuna/public/js/search/results.js
@@ -75,7 +75,6 @@ function compact(array) {
 // rabbit-hole that doesn't seem to have any end. Hence this slightly
 // horrid global state.
 let lastSearchFirstResultTime;
-const aiSearchResultLimit = 200;
 
 results.controller('SearchResultsCtrl', [
     '$rootScope',
@@ -288,7 +287,7 @@ results.controller('SearchResultsCtrl', [
           // TODO: avoid this initial search (two API calls to init!)
           ctrl.searched = search(
             $stateParams.useAISearch
-              ? {offset: 0, length: aiSearchResultLimit}
+              ? {offset: 0, length: $window._clientConfig.aiSearchResultLimit}
               : {length: 1, orderBy: 'newest'}
           ).then(function(images) {
             return initialiseResults(images);
--- a/kahuna/public/js/search/results.js
+++ b/kahuna/public/js/search/results.js
@@ -75,6 +75,7 @@ function compact(array) {
 // rabbit-hole that doesn't seem to have any end. Hence this slightly
 // horrid global state.
 let lastSearchFirstResultTime;
+const aiSearchResultLimit = 200;
 
 results.controller('SearchResultsCtrl', [
     '$rootScope',
@@ -229,11 +230,7 @@ results.controller('SearchResultsCtrl', [
             lastSearchFirstResultTime = undefined;
         }
 
-        // Initial search to find upper `until` boundary of result set
-        // (i.e. the uploadTime of the newest result in the set)
-
-        // TODO: avoid this initial search (two API calls to init!)
-        ctrl.searched = search({length: 1, orderBy: 'newest'}).then(function(images) {
+        function initialiseResults(images) {
             ctrl.totalResults = images.total;
             // FIXME: https://github.com/argo-rest/theseus has forced us to co-opt the actions field for this
             ctrl.tickerCounts = images.$response?.$$state?.value?.actions?.tickerCounts;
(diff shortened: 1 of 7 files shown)


kahuna/public/js/search/results.js
```diff
@@ -274,6 +279,18 @@ results.controller('SearchResultsCtrl', [
...
+          // (i.e. the uploadTime of the newest result in the set)
+
+          // TODO: avoid this initial search (two API calls to init!)
+          ctrl.searched = search(
+            $stateParams.useAISearch
+              ? {offset: 0, length: $window._clientConfig.aiSearchResultLimit}
+              : {length: 1, orderBy: 'newest'}
+          ).then(function(images) {
+            return initialiseResults(images);
         }).catch(error => {
```
alexduf: A nice little syntactic trick
```suggestion
          ).then(initialiseResults).catch(error => {
```

kahuna/public/js/search/results.js
```diff
@@ -274,6 +279,18 @@ results.controller('SearchResultsCtrl', [
             }
 
             return images;
+          }
+
+          // Initial search to find upper `until` boundary of result set
+          // (i.e. the uploadTime of the newest result in the set)
+
+          // TODO: avoid this initial search (two API calls to init!)
+          ctrl.searched = search(
```
alexduf: I think instead of having the code that tests twice for useAISearch, (once at the time the call is being triggered, and once we get the result for this call), it might make sense to have a completely different branch for both the call and the processing of the results

in pseudo code:
```
if (useAiSearch) {
  search({offset: 0, length: $window._clientConfig.aiSearchResultLimit}).then { (images) =>
              initialiseResults(images)
              images.data.forEach((image, index) => {
                ctrl.imagesAll[index] = image;
                imagesPositions.set(image.data.id, index);
                results.set(index, image);
              });
              ctrl.images = images.data.slice(0, totalLength);
  }
} else {
  search({length: 1, orderBy: 'newest'}).then(initialiseResults)
}
```

alexduf (approved):
I think it makes sense but I'm wondering if it could be made easier to follow what's going on? See suggestion.

It's probably not worth spending another day, but maybe a time-boxed period to see if my intuition is right or not.

kahuna/public/js/search/results.js
```diff
@@ -229,11 +229,7 @@ results.controller('SearchResultsCtrl', [
...
 
-        // Initial search to find upper `until` boundary of result set
-        // (i.e. the uploadTime of the newest result in the set)
-
-        // TODO: avoid this initial search (two API calls to init!)
-        ctrl.searched = search({length: 1, orderBy: 'newest'}).then(function(images) {
+        function initialiseResults(images) {
             ctrl.totalResults = images.total;
             // FIXME: https://github.com/argo-rest/theseus has forced us to co-opt the actions field for this
             ctrl.tickerCounts = images.$response?.$$state?.value?.actions?.tickerCounts;
```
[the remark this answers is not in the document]
ellenmuller: In the normal case, images.total will be the same as images.data.length - 200! There are some edge cases tbf, it could be <200 if there were fewer than 200 vectors in the index, or if some hits fail `resolveHit`. But yeah those are such big edge cases I won't change this :)

media-api/app/controllers/MediaApi.scala
```diff
@@ -641,7 +641,7 @@ class MediaApi(
     }
 
     def performAiSearchAndRespond(query: String): Future[Result] = {
-      val k = 200
+      val k = config.aiSearchResultLimit
       val searchResultsFuture = parseAiSearchMode(query) match {
         case SimilarSearch(imageId) => semanticSearchByImage(imageId, k)
         case TextSearch(textQuery) => semanticSearchByText(textQuery, k)
```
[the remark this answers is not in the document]
ellenmuller: Probably also an edge-case, but fair enough. Have changed to `imageHits.length`!

kahuna/public/js/search/results.js
```diff
@@ -259,6 +255,15 @@ results.controller('SearchResultsCtrl', [
...
             imagesPositions = new Map();
 
+            if ($stateParams.useAISearch) {
+              images.data.forEach((image, index) => {
+                ctrl.imagesAll[index] = image;
+                imagesPositions.set(image.data.id, index);
+                results.set(index, image);
+              });
+              ctrl.images = images.data.slice(0, totalLength);
+            }
```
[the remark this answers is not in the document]
ellenmuller: with total being `imageHits.length` this shouldn't be an issue anymore :)

kahuna/public/js/search/results.js
ellenmuller: OOooohh that's nice!! Thanks

kahuna/public/js/search/results.js
ellenmuller: I like that suggestion, I have split out an `initialisePagedResults` and `initialiseAiResults` branch (vibe-coded for speed) here https://github.com/guardian/grid/pull/4722/changes/9352574f8033cbbd5a9dfdb6ce563e0d02f52f78

alexduf (approved):

PR #4737, opened 2026-05-14
upgrade panda to latest; partially switching to s3 sdk v2

andrew-nowak:
## What does this change?

Upgrade pan-domain-auth to latest version, which requires partially switching to the v2 s3 sdk.

## Tested? Documented?
- [ ] locally by committer


bryophyta:
This looks good to me on a code level! When testing in CODE though I found that I wasn't sure how the `/cookieMonster` endpoint is expected to work. It doesn't seem to be working exactly how I would have thought, so perhaps we can take a look together at some point?

bryophyta (approved):
lgtm!

PR #4738, opened 2026-05-14
Hybrid search

ellenmuller:
## What does this change?
This PR switches the AI search from a vector KNN search to a hybrid search that combines semantic similarity with lexical BM25 matching.

It also introduces a `vecWeight` query parameter so we can control the boost parameter between vector relevance and keyword relevance from the client. This will help us experiment with different values and eventually settle on a default score.

**Main changes**
- add a new hybrid Elasticsearch query for AI text search (this is the query developed by the Data Science team, see [here](https://github.com/guardian/grid-embedding-experiments/blob/atve-boost-search/vector_search/src/model/hybrid_search/hybrid_search_es.py)) 
- combine KNN and multi_match results in a single query
- normalise BM25 scores using the query's max BM25 score so lexical and vector signals can be blended more predictably
- thread a new optional `vecWeight` parameter through Kahuna, the Media API, and search param parsing
- default `vecWeight` to 0.8 for AI text search when no value is supplied

Please note that the 'More Like This' search is unchanged :)

I have deployed to TEST, where you can do a semantic search and experiment with adding different vecWeights to the URL, eg `&vecWeight=0.4`.

## Tested? Documented?
- [x] locally by committer


media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -191,6 +195,89 @@ class ElasticSearch(
...
+    vecWeight: Double,
+  )(
+    implicit ex: ExecutionContext,
+    logMarker: LogMarker
+  ): Future[SearchResults] = {
+
+// BM25 scores are unbounded [0,inf] and typically much larger in magnitude
+// than cosine similarity (knn). So we get the max BM25 score for the query and use that to calculate
+// the scaling factor for the lexical part of the query, so that BM25 and knn scores are both between 0-1 scale
+// and can be effectively combined in a hybrid query.
```
alexduf: Tiny detail: I'm not sure why but all your comments (at least in this file) need re-indenting

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -191,6 +195,89 @@ class ElasticSearch(
...
+      val scalingFactor = if (maxScore > 0.0) 1.0 / maxScore else 1.0
+
+//    We want to apply only one boost if we can help it, so we scale the
+//    multi_match boost to be in line with the max_score and the desired
+//    lexical_weight/vec_weight balance
+      val multiMatchBoost = if (vecWeight > 0.0) (lexicalWeight/vecWeight) * scalingFactor else 1.0
+
+      logger.info(logMarker, s"Scaling factor for BM25 score is $scalingFactor, multi-match boost is $multiMatchBoost")
+
+//      TODO make case class for multimatchQuery to avoid repetition
```
alexduf: I'm not sure about making a case class, but at least a share function might help

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -191,6 +195,89 @@ class ElasticSearch(
...
+        boost = Some(multiMatchBoost)
+      )
+
+      val searchRequest = ElasticDsl.search(imagesCurrentAlias)
+        .bool(BoolQuery().should(Seq(multiMatchQuery, knn)))
+        .size(k)
+
+      executeAndLog(withSearchQueryTimeout(searchRequest), "hybrid search").map { r =>
+        val imageHits = r.result.hits.hits.map(resolveHit).toSeq.flatten.map(i => (i.instance.id, i))
+        SearchResults(hits = imageHits, total = imageHits.length, extraCounts = None)
```
alexduf: I think this whole function could do with a little bit more structure. If you define the right functions you may be able to have code that looks like this pseudo code block:

```scala
for {
  maxScore <- fetchMaxBm25Score(query)
  hybridQuery = makeHybridQuery(query, maxScore)
  result <- executeAndLog(hybridQuery)
} yield {
  // build the SearchResults object here
}
```

This should make this function both more readable and more maintainable

media-api/app/lib/elasticsearch/ElasticSearchModel.scala
```diff
@@ -115,6 +116,8 @@ object SearchParams {
 
   // TODO: return descriptive 400 error if invalid
   def parseIntFromQuery(s: String): Option[Int] = Try(s.toInt).toOption
+  def parseDoubleFromQuery(s: String): Option[Double] =
+    Try(s.toDouble).toOption.filter(d => !d.isNaN && !d.isInfinity && d >= 0.0 && d <= 1.0)
```
alexduf, line 120: 👍  this is very defensive, I like it, however the name isn't just `parseDouble`, as we're getting a number between 0 and 1. You can either rename the method to reflect the bounded nature of this check, or only parse the number here, and apply the bounding logic line 181

alexduf (approved):
I've flagged a few things that would be nice to address, either in this PR or in a subsequent one as I think this should work as-is

media-api/app/lib/elasticsearch/ElasticSearch.scala
ellenmuller: Ah that's so nice! Have tried doing this here https://github.com/guardian/grid/pull/4738/commits/44eda6f630f296895b6722dacfaf8b06ceebea22 - haven't tested this refactor on TEST yet but that's one for tomorrow :)

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -191,6 +195,86 @@ class ElasticSearch(
...
+    maxBM25Score(query).flatMap { maxScore =>
+//    KNN results are in [0,1], but BM25 scores are unbounded and typically much
+//    larger than cosine similarity, so we need to apply a scaling factor to the
+//    BM25 score to bring it to the same range as the cosine similarity
+      val scalingFactor = if (maxScore > 0.0) 1.0 / maxScore else 1.0
+
+//    We want to apply only one boost if we can help it, so we scale the
+//    multi_match boost to be in line with the max_score and the desired
+//    lexical_weight/vec_weight balance
+      val multiMatchBoost = if (vecWeight > 0.0) (lexicalWeight/vecWeight) * scalingFactor else 1.0
```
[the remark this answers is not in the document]
ellenmuller: I have actioned this comment here https://github.com/guardian/grid/pull/4738/changes/8a9f214edc747936f68c0e13940960b32a90daab but what do you think @aliceptve is there more to action here? I don't want to actually change the algorithm without consulting with you!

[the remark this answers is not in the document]
aliceptve: I can see that you've actioned this in a later commit, but just to confirm that indeed you need to account for vec_weight == 0.0 both here and when applying the knn boost (line 155 of my python script)

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -191,6 +195,91 @@ class ElasticSearch(
...
+
+  // BM25 scores are unbounded [0,inf] and typically much larger in magnitude
+  // than cosine similarity (knn). So we get the max BM25 score for the query and use that to calculate
+  // the scaling factor for the lexical part of the query, so that BM25 and knn scores are both between 0-1 scale
+  // and can be effectively combined in a hybrid query.
+  private def fetchMaxBm25Score(query: String)(implicit ex: ExecutionContext, logMarker: LogMarker): Future[Double] = {
+    val maxScoreRequest = ElasticDsl.search(imagesCurrentAlias)
+      .query(BoolQuery().must(
+        createMultiMatchQuery(query)
+      ))
```
andrew-nowak:
```suggestion
      .query(createMultiMatchQuery(query))
```

(A bool query with only only `must` subquery should be identical to making that query directly)

build.sbt
```diff
@@ -78,7 +78,7 @@ Global / concurrentRestrictions := Seq(
 
 val awsSdkVersion = "1.12.470"
 val awsSdkV2Version = "2.42.25"
-val elastic4sVersion = "8.18.2"
+val elastic4sVersion = "8.19.1"
```
joelochlann, line 81: Why has this been upgraded? It looks like we're still on elasticsearch 8.18

https://amigo.gutools.co.uk/recipes/grid-elasticsearch-8
https://github.com/guardian/grid/blob/hybrid-search/docker-compose.yml#L3

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -191,6 +195,89 @@ class ElasticSearch(
...
+    val multiMatchBoost = if (vecWeight > 0.0) (lexicalWeight / vecWeight) * scalingFactor else 1.0
+
+    logger.info(logMarker, s"Scaling factor for BM25 score is $scalingFactor, multi-match boost is $multiMatchBoost")
+
+    val multiMatchQuery = createMultiMatchQuery(query, boost = Some(multiMatchBoost))
+
+    ElasticDsl.search(imagesCurrentAlias)
+      .bool(BoolQuery().should(Seq(multiMatchQuery, knn)))
+      .size(k)
+  }
```
joelochlann, line 257: I'll defer to your taste on this but personally I find it a little easier to read with the 0/1 edge cases extracted

```suggestion
  private def combineMultiMatchAndKnn(
    multiMatchQuery: MultiMatchQuery,
    knn: Knn,
    vecWeight: Double,
    maxScore: Double
  )(implicit logMarker: LogMarker): BoolQuery = {
    val lexicalWeight = 1.0 - vecWeight

    // KNN results are in [0,1], but BM25 scores are unbounded and typically much
    // larger than cosine similarity, so we need to apply a scaling factor to the
    // BM25 score to bring it to the same range as the cosine similarity.
    val scalingFactor = 1.0 / maxScore

    // We want to apply only one boost if we can help it, so we scale the
    // multi_match boost to be in line with the max_score and the desired
    // lexical_weight/vec_weight balance
    val multiMatchBoost = (lexicalWeight / vecWeight) * scalingFactor

    logger.info(logMarker, s"Scaling factor for BM25 score is $scalingFactor, multi-match boost is $multiMatchBoost")
    BoolQuery().should(Seq(multiMatchQuery.boost(multiMatchBoost), knn))
  }

  private def makeHybridSearchRequest(
    query: String,
    queryEmbedding: List[Double],
    k: Int,
    numCandidates: Int,
    vecWeight: Double,
    maxScore: Double
  )(implicit logMarker: LogMarker): SearchRequest = {
    val multiMatch = createMultiMatchQuery(query)
    val knn = Knn("embedding.cohereEmbedV4.image")
      .queryVector(queryEmbedding)
      .k(k)
      .numCandidates(numCandidates)

    val q = vecWeight match {
      case 0.0 => multiMatch
      case 1.0 => knn
      case _ => combineMultiMatchAndKnn(multiMatch, knn, vecWeight, maxScore)
    }

    ElasticDsl.search(imagesCurrentAlias)
      .size(k)
      .query(q)
  }
```

joelochlann: Also, a question more for @aliceptve about the data science side.

I notice that in their linear retriever elasticsearch use [min/max normalisation for the BM25](https://www.elastic.co/search-labs/blog/linear-retriever-hybrid-search), and in fact in [this example](https://www.elastic.co/docs/reference/elasticsearch/rest-apis/retrievers/retrievers-examples#retrievers-examples-linear-retriever) they do min/max normalisation to **both** the BM25 and knn sides.

We're just doing max normalisation for BM25. This could definitely change the results in certain cases vs min/max, though I'm not totally clear on which is better or how much it matters.

For instance if you have
doc A: knn 0.9, BM25 9
doc B: knn 0.1, BM25 10

And assuming vecWeight = 0.5, i.e. 50/50 split between BM25 and vector

### max norm
doc A normed BM25 = 9/10 = 0.9
doc B normed BM25 = 10/10 = 1
doc A score = 0.9 + 0.9 = 1.8
doc B score = 0.1 + 1 = 1.1
Ordering is A > B

### min/max norm
If only two docs, normed BM25 values are always 0 and 1
doc A score = 0.9 + 0 = 0.9
doc B score = 0.1 + 1 = 1.1
Ordering is B > A

joelochlann (approved):
I think this looks good - my comments are mainly points for discussion, not blockers - but is the elasticsearch library upgrade intended?

PR #4743, opened 2026-05-19
Api endpoint for image crops

lindseydew:
## What does this change?

This makes the crop options that grid supports available as an API endpoint so that the media atom maker service can use this to ascertain if a specific crop is already supported in grid, so if another one is needed 

## Tested? Documented?
- [ ] locally by committer


media-api/app/controllers/ConfigurationController.scala
```diff
@@ -0,0 +1,10 @@
+package controllers
+
+import com.gu.mediaservice.lib.auth.Authentication
+import lib.crops.CropOption
+import play.api.libs.json.Json
+import play.api.mvc.{BaseController, ControllerComponents}
+
+class ConfigurationController(override val controllerComponents: ControllerComponents) extends BaseController {
+    def cropVariations = Action { _ => Ok(Json.toJson(CropOption.supported)) }
```
andrew-nowak, line 10: I think it's worth a brief comment that this is intentionally without auth, for future travellers who might think it was an oversight and attempt to correct

andrew-nowak (approved):
lgtm! 

In the future we could look to replace the clientside constants https://github.com/guardian/grid/blob/main/kahuna/public/js/util/constants/cropOptions.js with a lookup to this new endpoint, which would allow different deployers of the grid to more easily configure different "supported" ratios

media-api/app/controllers/ConfigurationController.scala
lindseydew: Yeah sounds good 👍

PR #4752, opened 2026-06-02
Add chips and filtering to AI Search

ellenmuller:
## What does this change?

This PR makes AI search respect Grid’s existing chip/filter behaviour and adds structured support for `similar:` queries.

Previously, 'Use AI search' embedded the raw query, including any chips, so structured chips such as file type and other filters did not work as one might expect. This branch changes the frontend and backend flow so AI search separates semantic text from structured filters, applies those filters to Elasticsearch, and handles `similar:` as part of the structured query model rather than via ad hoc string parsing.

**Changes**
- Added similarity 'chip' to the structured query parser.
- Switched AI mode detection to use structured query parsing.
- Split AI queries into:
  - semantic text to embed
  - structured filter conditions to apply to Elasticsearch
- Reused QueryBuilder logic so AI mode inherits the same chip semantics as standard search.
- Wired filter application through hybrid search and similar-image KNN search:
  - added inline KNN filtering to improve recall for filtered similar-image searches.
  - kept top-level query filtering so final hybrid results still obey filters on the lexical branch.
- Added backend guards for invalid AI queries:
  - chips-only AI text queries just say that the user needs to add a query
  - mixed similar: plus positive semantic text also doesn't work
- Added matching frontend guards so unsupported AI queries don't trickle end up in backend. (Copilot helped here!)
- Updated the search UI so filters are visible again in AI mode, except for sorting date. (Copilot helped here!) 

 

## Tested? Documented?
- [ ] locally by committer


joelochlann:
This looks really nice!

A first thought on the UI: now that we have filters & chips, I didn't expect the search to be blank until I enter a query. I thought that I would be able to mess around with filters, seeing how it shapes my dataset and the number of hits etc, before I added in the AI search layer on top.

I feel that this would make it clearer what's actually happening with the pre-filtering, especially if we give the right information in the UI.

Even if we think we want it blank until it enters a query, I think there has to be some information about what's going on with the pre-filtering + knn, otherwise always seeing "200 results" no matter what I do with the filters (unless I get it below 200!) is confusing.

Here's a (fairly clumsy) attempt

**x images matching filters => (of which) showing 200 most similar to AI search**

Screenshot 2026-06-05 at 14 45 46

paperboyo:
🎉🎉🎉 👏  

> I didn't expect the search to be blank until I enter a query

+1 to not blank, yeah.

Agreed signalling why 200 good idea. Changing Sort to Relevance may or (more likely) may not be enough. Maybe just crude text?

image

Whatever we decide, let’s not mimic tickers looks as these are clickable (to filter)?

Unrelatedly, I would also split More like this from the Use AI checkbox completely. Currently, this deadend is possible:

image

joelochlann:
Yeah it's tricky. The problem with "200 relevant" is we don't know that all 200 _are_ relevant! They're just the best 200 we could manage. In fact it's very rare they will all be relevant.

paperboyo:
> Yeah it's tricky. The problem with "200 relevant" is we don't know that all 200 _are_ relevant! They're just the best 200 we could manage. In fact it's very rare they will all be relevant.

Yes, of course. I struggled the come up with a word (just like for “Sort” order: I think they should be the same word). But what matters here more than the fact that the word isn’t perfect is that it’s there at all: it signals a) it’s a bound subset b) of a corpus you can actively shape, c) has to do with this AI thing and d) is about ordering.

Especially b) (`of 2,129`) is particularly revealing about the very UX issue we discussed above related to post-filtering behaving counter-intuitively (versus “normal” filtering).

And `200 AI-similarity-ordered of 2,129` was too long ;-)

[EDIT: sorry deleted doubled comment]

kahuna/public/js/search/structured-query/syntax.ts
```diff
@@ -30,6 +30,65 @@ type StructuredQueryText = {
...
 
+// This is intentionally lower-level than structureQuery(). The UI-oriented
+// structureQuery() merges all text tokens into a single text chip and preserves
+// negation by prefixing the text value with '-', which loses the distinction
+// between positive free text and negated-only text. AI search needs that
+// distinction so the frontend matches the backend rule: only positive unfielded
+// text counts as a valid semantic query.
+// i.e. `fileType:jpeg -hello` = not valid AI query
+// but `cats fileType:jpeg -hello` is!
+export function hasPositiveAiTextQuery(query: string) {
```
joelochlann: @jonathonherbert and @andrew-nowak will know for sure but since `use-cql-chips` is defaulted to `true`, I think we're probably safe to use the CQL abstractions for this, which should mean you don't have to manually parse the string in this way (you can look at structured properties instead)

Of course, as discussed, even simpler would be to simply have these guards only on the server and send a meaningful error back when the conditions aren't met, avoiding having to duplicate the checks clientside - if we can get away with it!

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -178,12 +178,13 @@ class ElasticSearch(
...
   }
 
-  def knnSearch(queryEmbedding: List[Float], k: Int, numCandidates: Int)
+  def knnSearch(queryEmbedding: List[Float], k: Int, numCandidates: Int, filterOpt: Option[Query])
                (implicit ex: ExecutionContext, logMarker: LogMarker): Future[SearchResults] = {
     val knn = Knn("embedding.cohereEmbedV4.image")
         .queryVector(queryEmbedding.map(_.toDouble))
         .k(k)
         .numCandidates(numCandidates)
+        .filter(filterOpt.getOrElse(matchAllQuery()))
```
joelochlann: I *think* this will do the same thing (the case class can accept an `Option` directly)
```suggestion
val knn = Knn("embedding.cohereEmbedV4.image", filter = filterOpt)
.queryVector(queryEmbedding.map(_.toDouble))
.k(k)
.numCandidates(numCandidates)
```

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -227,13 +231,15 @@ class ElasticSearch(
...
-    maxScore: Double
+    maxScore: Double,
+    filterOpt: Option[Query]
   )(implicit logMarker: LogMarker): SearchRequest = {
     val knn = Knn("embedding.cohereEmbedV4.image")
       .queryVector(queryEmbedding)
       .k(k)
       .numCandidates(numCandidates)
       .boost(if (vecWeight > 0.0) 1.0 else 0.0)
+      .filter(filterOpt.getOrElse(matchAllQuery()))
```
joelochlann:
```suggestion
val knn = Knn("embedding.cohereEmbedV4.image", filter = filterOpt)
.queryVector(queryEmbedding)
.k(k)
.numCandidates(numCandidates)
.boost(if (vecWeight > 0.0) 1.0 else 0.0)
```

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -242,17 +248,19 @@ class ElasticSearch(
...
     logger.info(logMarker, s"Scaling factor for BM25 score is $scalingFactor, multi-match boost is $multiMatchBoost")
 
     val multiMatchQuery = createMultiMatchQuery(query, boost = Some(multiMatchBoost))
+    val baseQuery = BoolQuery().should(Seq(multiMatchQuery, knn))
+    val filteredQuery = filterOpt.map(filter => BoolQuery().must(baseQuery).filter(filter)).getOrElse(baseQuery)
 
     ElasticDsl.search(imagesCurrentAlias)
-      .bool(BoolQuery().should(Seq(multiMatchQuery, knn)))
+      .bool(filteredQuery)
       .size(k)
```
joelochlann: I *think* this will work because `filter` accepts an iterable which it calls `toSeq` on. `None` would become an empty `Seq`, which is what `filters` defaults to if not passed at all.


(It does compile, it's just hard to test the functionality locally as I have no data)

Could do something similar for the max BM25 query

```suggestion
ElasticDsl.search(imagesCurrentAlias)
      .bool(BoolQuery().should(Seq(multiMatchQuery, knn)).filter(filterOpt))
.size(k)
```

joelochlann (commented):
Just summarising the decisions from our IRL code review.

These are the rules we distilled about what's currently allowed or not for combining the different types of search:

## Valid queries
- similar-only: image KNN
- text-only: text KNN
- chips/filters + text: chips/filters pre-filter, then text KNN
- chips/filters + similar: chips/filters pre-filter, then image similarity KNN
- negative + text ("keir starmer" -angela): -angela works as a lexical pre-filter, then text KNN
- negative + similar image: : -angela works as a lexical pre-filter, then image similarity KNN

## Invalid queries
- similar + text: bad because how do we combine these rankings?

Anything that does not contain a KNN bit is currently invalid. In AI search mode, we don't currently let the user play around with filters without specifying an image or a text similarity query. Whether this is the functionality we want is up for debate, but given this, the following are disallowed:

- chips/filters only
- negative-only (e.g. -angela)
- negative + chips/filters (e.g. -angela +fileType:jpeg)

## What we need
- [x] Backend tests to prove and explain all the above behaviour
- [x] Meaningful feedback in the UI so that the user can fix the problem in the bad cases listed above

Plan is to try providing this by sending an error from the backend and seeing if we can do away with the clientside guards. If we need clientside guards too, let's try and use the Chips Query Language abstractions to avoid have to re-implement parsing logic.

media-api/app/lib/elasticsearch/ElasticSearchModel.scala
```diff
@@ -23,6 +23,26 @@ case class AiQueryParts(
   similarImageId: Option[String]
 ) {
   def hasSimilarAndSemanticText: Boolean = similarImageId.nonEmpty && semanticQuery.nonEmpty
+
+  // A KNN bit is the only thing AI search can rank by, so a query must contain either
+  // a text query (text KNN) or a similar image (image KNN) to be valid.
+  def hasKnn: Boolean = similarImageId.nonEmpty || semanticQuery.nonEmpty
+
+  // Returns a user-facing error message if the AI query cannot be satisfied, otherwise None.
+  def validationError: Option[String] = {
```
joelochlann: 👌 I like the direction this is going in. FYI an `Either` can be a Scala-idiomatic choice for cases where something can succeed (`Right`) or fail (`Left`), and you want to return some data in the failure case as well as the success case.

ellenmuller: Actioned :)

media-api/app/lib/elasticsearch/ElasticSearchModel.scala
```diff
@@ -17,6 +17,67 @@ case class SearchResults(hits: Seq[(String, SourceWrapper[Image])], total: Long,
...
+        "AI search needs something to rank by. " +
+          "Add a text query or a similar image alongside your filters."
+      )
+    else
+      Right(())
+  }
+}
+
+object AiQueryParts {
+  def from(conditions: List[Condition]): AiQueryParts = {
```
joelochlann, line 49: Ah sorry I wasn't very clear when I mentioned `Either` before. The thing that can succeed or fail or not the `validationError`. It's the whole of `AiQueryParts`

If you do something like this:
```suggestion
  def from(conditions: List[Condition]): Either[String, AiQueryParts] = {
```

and `lazy val aiQueryParts` becomes an `Either`, then it _should_ significantly simplify the calling code and the tests

I don't think this a blocker for this PR but I recommend trying it in a follow up as I think it will be quite satisfying!

joelochlann (approved):
Great work!

PR #4753, opened 2026-06-02
only set Photographer usage rights if no other processor has set usage rights first

andrew-nowak:
## What does this change?

Sometimes staff or contract photographers will also work or send shots through agencies. The Photographer parser is at the end of the list, so overwrites the changes made by the earlier parsers. We've thought about it, and it makes sense for the Photographer parser to only set usage rights and metadata on the image if it's the final parser, and no other parser has detected anything and set any usage rights.

Find an image with a byline of a staff photographer, but also has agency metadata (from Getty, or AP, etc.) Before this change, the image would be assigned Staff photographer rights and be credited to The Guardian incorrectly. Afterwards, it should be assigned Agency rights and the correct credit string given.

## Tested? Documented?
- [x] locally by committer

---
  - https://app.asana.com/0/0/1215293755029218


lindseydew (approved):
LGTM 👍 Thanks for running through it this morning

PR #4758, opened 2026-06-08
T3214 mappings exclude vectors

AndyKilmory:
## What does this change?

This adds a new configuration parameter - es.includeDenseVectorMappings - which can take values 'true' or 'false' (default value if not included = true) and will control whether the DenseVectorField used for adding embedding vector data to the Elasticsearch/Opensearch index will be included or not. False will exclude the mapping structures from the index and so allow continued use with Opensearch (post fork) indexes without modification until a longer term solution can be established.

The config parameter should be added to Thrall and Media-Api services to fully resolve all the issues.

In addition the parameter has been used to add in checks to the knnSearch and hybridSearch to return empty results sets when the embedding data has been excluded from the index.

Ensure that the ES/Opensearch index will operate as expected and the api and thrall services run as expected with both settings for the parameter - true and false.

## Tested? Documented?
- [x] locally by committer

---
  - https://app.asana.com/0/0/1215602241038768


dev/script/generate-config/service-config.js
```diff
@@ -146,7 +146,7 @@ function getKahunaConfig(config){
         |]`;
 
     const permissionsConfig = stripMargin`
-      |usePermissionsFilter=true
+      |usePermissionsFilter=false
```
Conalb97, line 149: should this config change be included as part of this PR?

AndyKilmory: The Guardian had asked if this param could be reverted to false by default - the swap to true had crept in via an earlier PR - it only impacts local running copies so has no real impact on any deployments.

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -180,6 +181,10 @@ class ElasticSearch(
 
   def knnSearch(queryEmbedding: List[Float], k: Int, numCandidates: Int)
                (implicit ex: ExecutionContext, logMarker: LogMarker): Future[SearchResults] = {
+    if (!includeDenseVectorMappings) {
+      logger.warn(logMarker, "knnSearch called but includeDenseVectorMappings=false, returning empty results")
+      return Future.successful(SearchResults(Nil, total = 0, extraCounts = None))
```
Conalb97: I think return would be redundant here since in Scala the last expression in a method is implicitly returned, so:

```
if (!includeDenseVectorMappings) {
  logger.warn(...)
  Future.successful(SearchResults(Nil, total = 0, extraCounts = None))
}
```

Should be fine

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -266,6 +271,10 @@ class ElasticSearch(
     implicit ex: ExecutionContext,
     logMarker: LogMarker
   ): Future[SearchResults] = {
+    if (!includeDenseVectorMappings) {
+      logger.warn(logMarker, "hybridSearch called but includeDenseVectorMappings=false, returning empty results")
+      return Future.successful(SearchResults(Nil, total = 0, extraCounts = None))
```
Conalb97: See above comment on line 186

thrall/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -214,6 +219,11 @@ class ElasticSearch(
   def updateEmbedding(id: String, embedding: Embedding, lastModified: DateTime)
                      (implicit ex: ExecutionContext, logMarker: LogMarker): List[Future[ElasticSearchUpdateResponse]] = {
 
+    if (!includeDenseVectorMappings) {
+      logger.info(s"Skipping embedding update for image $id: dense vector mappings are disabled")
+      return List(Future.successful(ElasticSearchUpdateResponse()))
```
Conalb97: See earlier comments about return statement

media-api/app/lib/elasticsearch/ElasticSearch.scala
AndyKilmory: return removed

thrall/app/lib/elasticsearch/ElasticSearch.scala
AndyKilmory: return removed

media-api/app/lib/elasticsearch/ElasticSearch.scala
AndyKilmory: return removed

media-api/app/lib/elasticsearch/ElasticSearch.scala
andrew-nowak: This is usually true, but in this case this line is not the last expression in the method, so it is not implicitly returned. Either the remaining expressions in the method need to be inside an `else` block, or we need to use the `return` keyword (in general I'd recommend avoiding the `return` keyword as it does not always behave as you might expect, see https://tpolecat.github.io/2014/05/09/return.html. In this case I think it'd technically be fine here, but it's usually better to avoid it out of principle, to avoid needing to think too hard about whether it will definitely do what you want it to)

AndyKilmory:
Have added else clauses to make early return pattern explicit and also resolved conflicts

andrew-nowak (approved):

PR #4763, opened 2026-06-15
Hybrid search - fill and max normalise (Option 1b)

ellenmuller:
## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/media-api/app/lib/elasticsearch/ElasticSearch.scala
+++ b/media-api/app/lib/elasticsearch/ElasticSearch.scala
@@ -208,7 +208,7 @@ class ElasticSearch(
       `type` = Some(BEST_FIELDS),
       fuzziness = Some("AUTO"),
       maxExpansions = Some(50),
-      operator = Some(And),
+      operator = Some(Or),
       prefixLength = Some(1),
       boost = boost
     )
@@ -266,6 +266,48 @@ class ElasticSearch(
       .size(k)
   }
 
+  def fillAndMaxNormalise(
+    query: String,
+    filterOpt: Option[Query],
+    queryEmbedding: List[Double],
+    k: Int,
+    numCandidates: Int,
+    vecWeight: Double
+  )(implicit ex: ExecutionContext, logMarker: LogMarker): Future[Unit] = {
+
+    val knn = Knn("embedding.cohereEmbedV4.image", filter = filterOpt)
+      .queryVector(queryEmbedding)
+      .k(k)
+      .numCandidates(numCandidates)
+
+    val knnRequest = ElasticDsl.search(imagesCurrentAlias)
+      .knn(knn)
+      .size(k)
+
+    val multiMatchQuery = createMultiMatchQuery(query, boost = Some(1.0))
+
+    val multiMatchRequest = ElasticDsl.search(imagesCurrentAlias)
+      .query(multiMatchQuery)
+      .size(k)
+
+    // Kick both queries off before awaiting either, so they execute in parallel.
+    val knnResponseF = executeAndLog(withSearchQueryTimeout(knnRequest), "hybrid search: knn component")
+    val multiMatchResponseF = executeAndLog(withSearchQueryTimeout(multiMatchRequest), "hybrid search: bm25 component")
+
+    for {
+      knnResponse <- knnResponseF
+      multiMatchResponse <- multiMatchResponseF
+    } yield {
+      val knnHits = knnResponse.result.hits.hits
+      val multiMatchHits = multiMatchResponse.result.hits.hits
+
+      logger.info(logMarker, s"knn returned ${knnHits.length} hits: " +
+        knnHits.map(h => s"${h.id}=${h.score}").mkString(", "))
+      logger.info(logMarker, s"bm25 returned ${multiMatchHits.length} hits: " +
+        multiMatchHits.map(h => s"${h.id}=${h.score}").mkString(", "))
+    }
+  }
+
   def hybridSearch(
     query: String,
     queryEmbedding: List[Float],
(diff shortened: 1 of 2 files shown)


media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -266,6 +266,134 @@ class ElasticSearch(
...
+    multiMatchResponseF.onComplete(_ => logger.info(logMarker, s"hybrid search timing: bm25 component took ${multiMatchStopwatch.elapsed.toMillis} ms"))
+
+    // Raw cosine similarity, then mapped onto ES's Cosine `_score` scale of (1 + cos) / 2
+    // so a locally-computed score is directly comparable to the knn scores ES returns.
+    def cosineSim(a: List[Double], b: List[Double]): Double = {
+      val dot = (a zip b).map { case (x, y) => x * y }.sum
+      val magA = math.sqrt(a.map(x => x * x).sum)
+      val magB = math.sqrt(b.map(y => y * y).sum)
+      if (magA == 0.0 || magB == 0.0) 0.0 else dot / (magA * magB)
+    }
```
aliceptve, line 310: If the vectors in ES are unit-normalised, the dot product is equivalent to cos here and you don't need the magnitude & division part - worth checking if they are, to avoid extra steps

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -266,6 +266,134 @@ class ElasticSearch(
...
+        sourceById.get(id).map { source =>
+          // knn score: ES knn score if present, else local cosine mapped to ES Cosine scale.
+          val rawKnnScore = knnScoreById.getOrElse(id,
+            bm25EmbeddingById.get(id).map(emb => (1.0 + cosineSim(queryEmbedding, emb)) / 2.0).getOrElse(0.0)
+          )
+          // bm25 score: bm25 top-k score if present, else exact follow-up score, else 0.
+          val rawBm25Score = bm25ScoreById.getOrElse(id, followUpBm25ScoreById.getOrElse(id, 0.0))
+
+          // HNSW is approximate, so a locally-computed knn score can nudge above maxScore - clamp to 1.
+          val normKnn = if (knnMaxScore > 0.0) math.min(rawKnnScore / knnMaxScore, 1.0) else 0.0
```
aliceptve, line 372: If we genuinely get a KNN score higher than the previous one, it doesn't really matter, so we don't need to do this capping here. If it's higher, it should get placed higher in the ranking

media-api/app/lib/elasticsearch/ElasticSearch.scala
joelochlann: I _think_ that query vectors will be, because we actually request them at 256 dims, but I think image vectors will not, because we've stored the truncated vector that we requested at 1536

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -266,6 +266,134 @@ class ElasticSearch(
...
+    val knn = Knn("embedding.cohereEmbedV4.image", filter = filterOpt)
+      .queryVector(queryEmbedding)
+      .k(k)
+      .numCandidates(numCandidates)
+
+    val knnRequest = ElasticDsl.search(imagesCurrentAlias)
+      .knn(knn)
+      .size(k)
+
+    val multiMatchQuery = createMultiMatchQuery(query, boost = Some(1.0))
```
joelochlann: I don't think you need this `boost`?
```suggestion
    val multiMatchQuery = createMultiMatchQuery(query)
```

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -266,6 +266,134 @@ class ElasticSearch(
...
+          )
+          // bm25 score: bm25 top-k score if present, else exact follow-up score, else 0.
+          val rawBm25Score = bm25ScoreById.getOrElse(id, followUpBm25ScoreById.getOrElse(id, 0.0))
+
+          // HNSW is approximate, so a locally-computed knn score can nudge above maxScore - clamp to 1.
+          val normKnn = if (knnMaxScore > 0.0) math.min(rawKnnScore / knnMaxScore, 1.0) else 0.0
+          val normBm25 = if (bm25MaxScore > 0.0) rawBm25Score / bm25MaxScore else 0.0
+
+          val combinedScore = vecWeight * normKnn + (1.0 - vecWeight) * normBm25
+          (id, source, combinedScore)
```
joelochlann, line 376: could be a bit more readable to create a little case class for this instead of a tuple

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -266,6 +266,134 @@ class ElasticSearch(
...
+      val mergeStopwatch = Stopwatch.start
+
+      // Every candidate id: union of the knn results and the bm25 top-k.
+      val candidateIds = knnScoreById.keySet ++ bm25ScoreById.keySet
+
+      val scored = candidateIds.toSeq.flatMap { id =>
+        sourceById.get(id).map { source =>
+          // knn score: ES knn score if present, else local cosine mapped to ES Cosine scale.
+          val rawKnnScore = knnScoreById.getOrElse(id,
+            bm25EmbeddingById.get(id).map(emb => (1.0 + cosineSim(queryEmbedding, emb)) / 2.0).getOrElse(0.0)
```
joelochlann, line 366: Everything else is beautifully readable but this line is a little dense. Could do with breaking out with variables/functions to make a bit clearer

joelochlann: also you could use `orElse` to avoid the nested `getOrElse` e.g.

```scala
val rawKnnScore = knnScoreById.get(id)
  .orElse(bm25EmbeddingById.get(id).map(localKnnScore))
  .getOrElse(0.0)
```

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -266,6 +266,134 @@ class ElasticSearch(
...
+          )
+          // bm25 score: bm25 top-k score if present, else exact follow-up score, else 0.
+          val rawBm25Score = bm25ScoreById.getOrElse(id, followUpBm25ScoreById.getOrElse(id, 0.0))
```
joelochlann, line 369: again, `orElse` could help you un-nest the `getOrElse`

PR #4764, opened 2026-06-16
Revamp hybrid search algorithm

joelochlann:
As per notes in [this doc](https://docs.google.com/document/d/1Co9FtK4QdcSkEkNy5ZemFNdv_vqviHLPdnGE5lMf1XI/edit?tab=t.0#heading=h.p4rprtn4i73y), we want to implement a new hybrid search algorithm based on the approach in [this paper](https://arxiv.org/pdf/2210.11934). This new algorithm does the following things differently from our current approach:
1. widens the lexical search: BM25 OR (instead of AND)
2. fills in missing scores: get the BM25 scores for KNN-only search results, and cosine similarity for BM25-only results
3. theoretical-min-max normalises both BM25 and KNN scores: that is, move to range [0,1] where 0 is the _theoretical_ min (-1 for cosine similarity, 0 for BM25) and 1 is the actual observed max

From notebooks in https://github.com/guardian/grid-embedding-experiments/pull/18, we've seen dramatically better results, primarily caused by points (1) and (2) above.

For instance, here's "andy burnham happy" before/after:

## before
Zero pictures of andy burnham, because there are zero lexical hits containing andy AND burnham AND happy:

Screenshot 2026-06-19 at 10 42 13

## after
8 out of 8 are of Andy Burnham. The OR is allowing the lexical search to actually find pictures mentioning Andy Burnham, and the score filling approach is then allowing the semantic signal for "happy" to actually do some work in the ranking:
Screenshot 2026-06-19 at 10 42 08

## Implementation notes
This implementation uses `rescore` (thank you @andrew-nowak!) to allow us to do it in just two parallel queries: one to get top k lexical results, and one to get top k semantic results and replace their semantic score with a lexical score. We then just need to calculate semantic scores for all in the Scala code.

This was benchmarked against https://github.com/guardian/grid/pull/4763 which uses a third query to get lexical scores for the semantic hits by id, and found to be slightly quicker. Both implementations were comfortably under a second on TEST.

### other optimisations tried and rejected
- #4775 tried to avoid contention on the parsing of the large result sets (we need the vectors to calculate cosine similarity), but was found to be no faster on TEST
- #4774 tried to reduce the payload of the initial lexical and semantic responses, at a cost of an extra request to hydrate the final chosen k, but also was found to have little overall benefit


common-lib/src/main/scala/com/gu/mediaservice/lib/VectorUtils.scala
```diff
@@ -0,0 +1,15 @@
...
+  def dotProduct(vectorOne: List[Double], vectorTwo: List[Double]): Double = {
+    vectorOne.zip(vectorTwo).map(component => component._1 * component._2).sum
+  }
+
+  def magnitude(vector: List[Double]): Double = {
+    math.sqrt(vector.map(math.pow(_, 2)).sum)
+  }
+
+  def cosineSimilarity(vectorOne: List[Double], vectorTwo: List[Double]): Double = {
+    dotProduct(vectorOne, vectorTwo) / (magnitude(vectorOne) * magnitude(vectorTwo))
```
joelochlann: whoops, doesn't handle the zero magnitude case!

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -259,19 +260,107 @@ class ElasticSearch(
...
+      // Account for the rare case in which KNN doesn't return the true closest vector,
+      // and that true closest vector happens to be among the lexical-only results.
+      val maxLexicalScore = dedupedResults.maxBy(_.lexicalScore).lexicalScore
+      val maxSemanticScore = dedupedResults.maxBy(_.semanticScore).semanticScore
+
+      def combinedScore(result: HybridResult): Double = {
+        val normedLexicalScore = result.lexicalScore / maxLexicalScore
+        // This is theoretical min, i.e. -1, to actual max,
+        // so it effectively does both the max-norming and the ES-score norming.
+        val normedSemanticScore = (result.semanticScore + 1) / (maxSemanticScore + 1)
```
aliceptve: Took me way too long to understand this 😅 maybe add a comment to explain?
```suggestion
        // normedScore = (score - theoretical_min) / (max - theoretical_min)
        // This is theoretical min, i.e. -1, to actual max,
        // so it effectively does both the max-norming and the ES-score norming.
        val normedSemanticScore = (result.semanticScore + 1) / (maxSemanticScore + 1)
```

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -259,19 +260,107 @@ class ElasticSearch(
...
+      }
+
+    def getTopK(results: List[HybridResult], vecWeight: Double, k: Int)(implicit logMarker: LogMarker): List[HybridResult] = {
+      val dedupedResults = results.distinctBy(_.id)
+      logger.info(logMarker, s"${dedupedResults.length} deduped results")
+
+      // Account for the rare case in which KNN doesn't return the true closest vector,
+      // and that true closest vector happens to be among the lexical-only results.
+      val maxLexicalScore = dedupedResults.maxBy(_.lexicalScore).lexicalScore
+      val maxSemanticScore = dedupedResults.maxBy(_.semanticScore).semanticScore
```
aliceptve: Do we need to account for this? If a non-KNN result is the closest vector, it won't be by much, and the only effect will be that its score is a little bit higher than 1 after normalising. It will still be the top one, and the order won't be affected

joelochlann: Well, the code actually comes out slightly more readable doing it this way so we don't lose much (except another pass through the results which will barely be measurable if `k < 1000`). I'm happy either way, I only switched because it actually fell out quite naturally from the way I'd structured things

media-api/app/lib/elasticsearch/ElasticSearch.scala
joelochlann: ah yes! great point, you need the underlying formula to see that the `+ 1` is actually `- (-1)` 😂. thanks!

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -259,19 +260,128 @@ class ElasticSearch(
...
+    queryEmbedding: List[Double],
+    k: Int,
+    numCandidates: Int,
+    vecWeight: Double,
+    filterOpt: Option[Query]
+  )(implicit ex: ExecutionContext, logMarker: LogMarker): Future[SearchResults] = {
+    import HybridResult.{resolveHitAndFillInSemanticScore, combineScoresAndGetTopK}
+    val lexicalQuery = createMultiMatchQuery(query, operator = Or)
+
+    val lexicalSearchRequest = ElasticDsl
```
joelochlann: I'm not applying the filters here!! Fix

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -259,19 +260,128 @@ class ElasticSearch(
...
+
+    def combineScoresAndGetTopK(
+      results: List[HybridResult],
+      vecWeight: Double,
+      k: Int
+    ): List[HybridResult] = {
+      // Account for the rare case in which KNN doesn't return the true closest vector,
+      // and that true closest vector happens to be among the lexical-only results.
+      val maxLexicalScore = results.maxBy(_.lexicalScore).lexicalScore
+      val maxSemanticScore = results.maxBy(_.semanticScore).semanticScore
```
joelochlann: I could use `.hits.maxScore` instead if I want

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -259,19 +260,128 @@ class ElasticSearch(
...
+
+      def combinedScore(result: HybridResult): Double = {
+        val normedLexicalScore = result.lexicalScore / maxLexicalScore
```
joelochlann: hm, these could end up dividing by zero!

media-api/app/lib/elasticsearch/ElasticSearch.scala
joelochlann: fixed

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -179,91 +180,128 @@ class ElasticSearch(
...
-    val filteredQuery = filterOpt.map(filter => boolQuery().must(baseQuery).filter(filter)).getOrElse(baseQuery)
+  private def lexicalRequest(
+    lexicalQuery: MultiMatchQuery,
+    k: Int,
+    filterOpt: Option[Query]
+  ): SearchRequest =
+    ElasticDsl
+      .search(imagesCurrentAlias)
+      .query(boolQuery().must(lexicalQuery).filter(filterOpt))
+      .size(k)
```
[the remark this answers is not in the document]
joelochlann: fixed by https://github.com/guardian/grid/pull/4764/changes/2a969622e4c1890b3150c580297ddef5684a343e

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -281,14 +325,25 @@ class ElasticSearch(
...
       Future.successful(SearchResults(Nil, total = 0, extraCounts = None))
     } else {
-      for {
-        maxScore <- fetchMaxBm25Score(query, filterOpt)
-        searchRequest = makeHybridSearchRequest(query, queryEmbedding, k, numCandidates, vecWeight, maxScore, filterOpt)
-        result <- executeAndLog(withSearchQueryTimeout(searchRequest), "hybrid search")
-      } yield {
-        val imageHits = result.result.hits.hits.map(resolveHit).toSeq.flatten.map(i => (i.instance.id, i))
-        SearchResults(hits = imageHits, total = imageHits.length, extraCounts = None)
+      val stopwatch = Stopwatch.start
```
ellenmuller, line 334: We don't need to time it going forward right?

common-lib/src/main/scala/com/gu/mediaservice/lib/VectorUtils.scala
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  def cosineSimilarity(vectorOne: List[Double], vectorTwo: List[Double]): Option[Double] = {
+    val magnitudeProduct = magnitude(vectorOne) * magnitude(vectorTwo)
+    // Cosine similarity is undefined when either vector has zero magnitude
+    // (the angle to the origin is meaningless), so we return None and let the
+    // caller choose a domain-appropriate fallback.
+    if (magnitudeProduct == 0.0) None
+    else Some(dotProduct(vectorOne, vectorTwo) / magnitudeProduct)
+  }
```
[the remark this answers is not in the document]
joelochlann: fixed

common-lib/src/main/scala/com/gu/mediaservice/lib/VectorUtils.scala
```diff
@@ -0,0 +1,70 @@
...
+  // Builds a vector whose cosine similarity with the first basis vector equals
+  // the requested `similarity` (in [-1, 1], where 1 is identical and 0 is orthogonal).
+  def vectorWithCosineSimilarity(dims: Int, similarity: Double): List[Double] = {
+    require(similarity >= -1.0 && similarity <= 1.0, "Cosine similarity must be between -1 and 1")
+    if (similarity == 0.0) {
+      // A cosine similarity of 0 means orthogonal to the first basis vector.
+      // That vector is one-hot at index 0, so any vector with a zero
+      // first component is orthogonal to it. (0, 1, 0, 0, ...) is the simplest choice.
+      oneHotVector(dims, 1)
+    } else {
```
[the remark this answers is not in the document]
joelochlann: covered

common-lib/src/test/scala/com/gu/mediaservice/lib/VectorUtilsTest.scala
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    it ("should return a magnitude of zero for a zero vector") {
+      magnitude(List(0.0, 0.0, 0.0)) shouldBe 0.0
+    }
+  }
+
+  describe("cosineSimilarity") {
+    it ("should return a cosine similarity of 1 for identical vectors") {
+      cosineSimilarity(List(1.0, 2.0, 3.0), List(1.0, 2.0, 3.0)).value shouldBe 1.0 +- 1e-9
```
ellenmuller: You should use the tolerance defined on line 76 here as well! And other cosine similarity tests :)

common-lib/src/test/scala/com/gu/mediaservice/lib/VectorUtilsTest.scala
```diff
@@ -0,0 +1,140 @@
...
+    it ("should be undefined (None) when both vectors have zero magnitude") {
+      cosineSimilarity(List(0.0, 0.0, 0.0), List(0.0, 0.0, 0.0)) shouldBe None
+    }
+
+    it ("should be undefined (None) when the vectors differ in dimensionality") {
+      cosineSimilarity(List(1.0, 2.0), List(1.0, 2.0, 3.0)) shouldBe None
+    }
+  }
+
+  describe("vectorWithCosineSimilarity") {
```
ellenmuller, line 78: I would add a comment here to explain that this function is just for testing, otherwise future readers may be confused why we'd need to create vectors ourselves

media-api/app/lib/elasticsearch/HybridResult.scala
```diff
@@ -0,0 +1,131 @@
...
+        // *not* the ES-normalised score, but when we normalise
+        // later it will end up in the range 0-1.
+        .flatMap(e => VectorUtils.cosineSimilarity(e.image, queryEmbedding))
+        // cosineSimilarity is None when there's no usable semantic signal (a
+        // missing, mismatched, or zero-magnitude embedding), all of which fall
+        // through to the theoretical min.
+        .getOrElse {
+          logger.warn(logMarker, s"Could not calculate semantic score for image ${hit.id}. Falling back to minimum semantic score")
+          CosineSimilarityTheoreticalMin
+        }
```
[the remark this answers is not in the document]
joelochlann: No, I want a warning because in PROD it's always unexpected. TEST we do have lots of images that aren't embedded, but I'm hoping these warnings might give us a nudge to do a backfill some day

media-api/test/lib/elasticsearch/HybridSearchTest.scala
```diff
@@ -0,0 +1,162 @@
...
+
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+    ES.ensureIndexExistsAndAliasAssigned()
+
+    implicit val logMarker: LogMarker = MarkerMap()
+    purgeTestImages
+    Await.ready(saveImages(testImages), 1.minute)
+    eventually(timeout(fiveSeconds), interval(oneHundredMilliseconds))(totalImages shouldBe testImages.size)
+  }
```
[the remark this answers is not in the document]
joelochlann: true, though also true of lots of other tests... I think a fix for this would require a separate PR

ellenmuller (approved):
Just a few tidy up comments, but looks good otherwise i think :)

media-api/app/lib/elasticsearch/ElasticSearch.scala
joelochlann: we haven't seen how long it takes in PROD yet so worth keeping around for now I think

common-lib/src/test/scala/com/gu/mediaservice/lib/VectorUtilsTest.scala
joelochlann: good spot! fixed in 0b0fcf640b605294d1388d4f66d5ce84f0bb088c

common-lib/src/test/scala/com/gu/mediaservice/lib/VectorUtilsTest.scala
joelochlann: Good idea! Done in f60e38c2761aadc57963762b459d9b884e4277c4

PR #4765, opened 2026-06-17
Missing scores: both options

joelochlann:
## not ready for review yet!
(sorry, I mistakenly thought I needed to take it out of draft to get the build pipeline to run)


joelochlann:
We've benchmarked. Not a lot to choose between them but option 2 (#4764) seems slightly faster. Both are sub-second

PR #4779, opened 2026-06-25
AWS Part 1: Sqs sns upgrade

lindseydew:
## What does this change?
This upgrades the sqs and sns aws sdk 

## Tested? Documented?
- [ ] locally by committer


common-lib/src/main/scala/com/gu/mediaservice/lib/aws/SimpleSqsMessageConsumer.scala
```diff
@@ -1,42 +1,59 @@
...
-   client.getQueueAttributes(queueUrl, List(
-      "ApproximateNumberOfMessagesDelayed",
-      "ApproximateNumberOfMessages",
-      "ApproximateNumberOfMessagesNotVisible"
-    ).asJava).getAttributes.asScala.toMap
+    client.getQueueAttributes(
+      GetQueueAttributesRequest.builder().
+        queueUrl(queueUrl)
+        .attributeNames(
+          QueueAttributeName.APPROXIMATE_NUMBER_OF_MESSAGES,
```
bryophyta, line 44: 👍 👍

bryophyta (dismissed):
code changes look good to me!

bryophyta (approved):
Thought it might not be working but I think I've confirmed that this is due to the fact that the flow was dependent on the Kinesis processing which had stopped in TEST due to a separate change: https://github.com/guardian/grid/pull/4780#issuecomment-4990741936

When I deployed this branch without the Kinesis change it looks like this is working as expected as far as I can see from the logs: https://logs.gutools.co.uk/s/editorial-tools/app/r/s/OW8g7

PR #4780, opened 2026-06-25
AWS Part 2: Kinesis upgrade

lindseydew:
## What does this change?

This upgrades the aws sdk of kinesis to version 2

## Tested? Documented?
- [ ] locally by committer


bryophyta:
Looks like there's a problem with this: I [deployed to TEST](https://riffraff.gutools.co.uk/deployment/view/9a5ebe20-05d7-4438-acc6-5bca027aebda) and I'm seeing lots of `kinesis putRecord failed`. We do seem to get these occasionally anyway, but in this case it looks like all of them have failed since deploy

https://logs.gutools.co.uk/s/editorial-tools/app/r/s/qsta1

**Update:**

The stack trace mentions:

```
software.amazon.awssdk.core.exception.SdkClientException: Unable to unmarshall response (software.amazon.awssdk.thirdparty.jackson.core.JsonParseException: Unexpected character ('¿' (code 191)): expected a valid value (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
```

Nothing came up from googling, but from digging around with Claude it looks like it might be because we aren't pinning the version of the `software.amazon.awssdk:kinesis` explicitly in the `common-lib` package, so it's pulling in a slightly older version of sdk v2 from `kinesis-logback-appender`. I've confirmed that it's resolving an older version in `common-lib` but haven't tried a build/deploy to check.

bryophyta (approved):
this looks good to me now!

PR #4785, opened 2026-06-29
Add GNM-owned, agency picks and filtered matches to AI search

joelochlann:
Screenshot 2026-06-29 at 12 12 10
Screenshot 2026-06-29 at 12 12 55

Unfortunately we need extra queries to achieve this. We cannot simply add an `aggregations` clause to an existing query, primarily because the final result set isn't actually the result of a query, it's the result of a clientside merge. Even if we could do aggregations on the individual queries, we wouldn't know which results contributed to the count and therefore couldn't preserve the information on de-duplication.

For GNM-owned and agency picks, an alternative approach could be to compute them in a script field like `isPotentiallyGraphic` and do all the aggregation in Scala. This _might_ also mean you could get rid of the duplication of logic clientside, e.g. https://github.com/guardian/grid/blob/main/kahuna/public/js/services/image-logic.js#L47-L58. I feel like this would be significant change though and would only be a win if you could consolidate the logic for these in one place.

Meanwhile for the "x matches" in "best k of x matches", aggregations are always computed over the matched documents, e.g. over the `k` for `knn` or over the BM25 matches for the lexical query. This means it would never give you true background filtered count over which your main query was running.

Extra queries shouldn't add too much latency because
- both of them are size=0, we just want the aggregation counts
- one of them is constrained to just `k` ids
- one of them can be run in parallel with existing queries

and indeed they seem to be well under 100ms in TEST

Screenshot 2026-06-29 at 17 31 22

### DIFF
--- a/media-api/app/lib/elasticsearch/ElasticSearch.scala
+++ b/media-api/app/lib/elasticsearch/ElasticSearch.scala
@@ -341,12 +341,45 @@ class ElasticSearch(
         case _ => fusedLexicalAndSemanticSearch(query, queryEmbedding, k, numCandidates, vecWeight, filterOpt)
       }
 
-      searchResults.andThen { case _ =>
+      // Run in parallel: how many images match the filters in total (so we can
+      // show "Best k of N matches"). The ticker count badges are computed over
+      // the top-k results we actually return, not the whole filtered set.
+      val filterTotal = countMatchingFilter(filterOpt)
+
+      (for {
+        results <- searchResults
+        total <- filterTotal
+        extraCounts <- extraCountsForIds(results.hits.map(_._1))
+      } yield results.copy(total = total, extraCounts = Some(extraCounts))).andThen { case _ =>
         val elapsed = stopwatch.elapsed
         logger.info(
           combineMarkers(logMarker, elapsed),
           s"Hybrid AI search completed in ${elapsed.toMillis} ms"
         )
+      }
+    }
+  }
+
+  // How many images match the active filters in total, so we can show
+  // "Best k of N matches".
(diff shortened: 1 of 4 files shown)


media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -351,6 +360,30 @@ class ElasticSearch(
...
+      .size(0)
+
+    executeAndLog(withSearchQueryTimeout(searchRequest), "hybrid AI search filter count").map(_.result.totalHits)
+  }
+
+  // The ticker count badges restricted to the given (top-k) result ids.
+  private def extraCountsForIds(ids: Seq[String])(implicit ex: ExecutionContext, logMarker: LogMarker): Future[ExtraCounts] = {
+    if (ids.isEmpty) Future.successful(ExtraCounts(tickerCounts = Map.empty))
+    else {
+      val searchRequest = prepareSearch(filters.ids(ids.toList))
```
joelochlann: We're not actually using `prepareSearch` in our main query. Either everything AI-search related should be migration-aware, or nothing should be

media-api/app/lib/elasticsearch/ElasticSearch.scala
```diff
@@ -351,6 +360,30 @@ class ElasticSearch(
     }
   }
 
+  // How many images match the active filters in total, so we can show
+  // "Best k of N matches".
+  private def countMatchingFilter(filterOpt: Option[Query])(implicit ex: ExecutionContext, logMarker: LogMarker): Future[Long] = {
```
joelochlann, line 365: Feel like this one could be clearer using the `_count` API? Though there is precedent for this approach - `usageForSupplier` does a size 0 query to get some aggregate numbers

ellenmuller (approved):

PR #4787, opened 2026-07-01
Calculate aggregations against the full filtered set, not top k

joelochlann:
# before
GNM-owned and agency picks were out of the top k, to reflect how many you'd actually see if you scrolled through your 200 results on screen.

Screenshot 2026-07-01 at 12 42 49

However, this meant that if you clicked on the badge to filter, you would see something very surprising: a pre-filter by GNM-owned, suddenly giving you 200 GNM-owned results when you thought you only had 2!

Screenshot 2026-07-01 at 12 42 57

# after
@andrew-nowak suggested we just make the aggregations over the filtered set, not top k. This is great because it lets us delete a whole query.

It also means the journey when clicking is coherent: your "4,696 GNM-"owned becomes the "4,696 matches" on the next page

Screenshot 2026-07-01 at 12 41 39

Screenshot 2026-07-01 at 12 41 49

### DIFF
--- a/media-api/app/lib/elasticsearch/ElasticSearch.scala
+++ b/media-api/app/lib/elasticsearch/ElasticSearch.scala
@@ -342,14 +342,13 @@ class ElasticSearch(
       }
 
       // Run in parallel: how many images match the filters in total (so we can
-      // show "Best k of N matches"). The ticker count badges are computed over
-      // the top-k results we actually return, not the whole filtered set.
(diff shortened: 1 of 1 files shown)


joelochlann:
Suggestion from @ellenmuller: let's try flipping the order of the filter matches and badges, and the best 200. So it reads left to right in the way it filters

andrew-nowak (approved):

PR #4788, opened 2026-07-01
remove use in use AI search

ellenmuller:
## What does this change?

It looks a bit neater for the tickbox to just say 'AI Search' - it takes up less valuable real estate on the search bar and the word 'Use' is superfluous anyway :)

### DIFF
--- a/kahuna/public/js/search/query.html
+++ b/kahuna/public/js/search/query.html
@@ -4,7 +4,7 @@
         <label>
             <input type="checkbox"
                    ng-model="searchQuery.useAISearch"/>
-            Use AI search
+            AI search
         </label>
     </span>
     <span class="search-query">


joelochlann:
Could be worth renaming the variable too while we're here? Elsewhere in the clientside code we have `isAiSearch` (vs `useAISearch`), which is confusing - best to standardise on `isAISearch` I think?

joelochlann (approved):
https://www.youtube.com/watch?v=PEgk2v6KntY

ellenmuller:
> Could be worth renaming the variable too while we're here? Elsewhere in the clientside code we have `isAiSearch` (vs `useAISearch`), which is confusing - best to standardise on `isAISearch` I think?

Welllll.... if we moved to `isAISearch` that would involve changing the URL parameter as well, which would mean any bookmarked/copied URLs would break. I don't think we want to do that right now! I could be in favour of standardising to `useAISearch` though :)

PR #4791, opened 2026-07-02
Display filteredPool when AI search query is absent

ellenmuller:
## What does this change?

When AI search mode has no ranking query, the UI previously showed two different empty states — one for "no query at all" and one for "filters but no query". This unifies them into a single state that tells the user how many images are in the pool a query would rank over.

- **Backend**: The no-query AI case (with or without filters) now returns filterPoolCounts.filteredPool (the size of the pool a query would rank over) instead of erroring or returning an empty response. Consolidated onto the existing countMatchingFilterWithExtraCounts,
- **Frontend**: Combined the two empty states into one message: "N images match your filters. Add a query to see the best 200." The results count below the search bar now also shows the pool size (N matches) in this state.

### DIFF
--- a/kahuna/public/js/search/results.html
+++ b/kahuna/public/js/search/results.html
@@ -222,14 +222,13 @@
         <div ng-if="ctrl.filtersOnlyAiSearch"
              class="image-no-results">
             <span>
-                {{ctrl.filterPoolCounts.filteredPool | toLocaleString}} images match your filters. Add a query to see the best 200.
+                {{ctrl.filterPoolCounts.filteredPool | toLocaleString}} images match your filters. Enter a query to rank them and surface the best 200 matches.
             </span>
         </div>
 
         <div ng-if="! ctrl.filtersOnlyAiSearch && ctrl.totalResults == 0"
              class="image-no-results">
-            <span ng-if="!ctrl.needsQuery">No matches, sorry! Try altering your search.</span>
-            <span ng-if="ctrl.needsQuery">Please enter a query. You can include filters, but must include either some text or a similar:image-id term.</span>
+            <span>No matches, sorry! Try altering your search.</span>
         </div>
     </div>
 </div>
--- a/kahuna/public/js/search/results.html
+++ b/kahuna/public/js/search/results.html
@@ -22,8 +22,10 @@
 
             <div class="results-toolbar-item results-toolbar-item--static
                     image-results-count">
-                <span ng-if="ctrl.isAiSearch && !ctrl.filtersOnlyAiSearch">Best {{ctrl.aiResultsShown | toLocaleString}} of {{ctrl.totalResults | toLocaleString}} matches</span>
-                <span ng-if="ctrl.isAiSearch && ctrl.filtersOnlyAiSearch">{{ctrl.filterPoolCounts.filteredPool | toLocaleString}} matches</span>
+                <span ng-if="ctrl.isAiSearch">
+                    <span ng-if="!ctrl.filtersOnlyAiSearch">Best {{ctrl.aiResultsShown | toLocaleString}} of {{ctrl.totalResults | toLocaleString}} matches</span>
+                    <span ng-if="ctrl.filtersOnlyAiSearch">{{ctrl.filterPoolCounts.filteredPool | toLocaleString}} matches</span>
+                </span>
                 <button ng-if="ctrl.newImagesCount > 0"
                         ng-click="ctrl.revealNewImages()"
                         class="image-results-count__new"
--- a/kahuna/public/js/search/results.html
+++ b/kahuna/public/js/search/results.html
@@ -222,9 +222,7 @@
         <div ng-if="ctrl.filtersOnlyAiSearch"
              class="image-no-results">
             <span>
-                Your filters have narrowed your pool of images down from {{ctrl.filterPoolCounts.totalPool | toLocaleString}}
-                to {{ctrl.filterPoolCounts.filteredPool | toLocaleString}}.
-                Please add a search query to search over these.
+                {{ctrl.filterPoolCounts.filteredPool | toLocaleString}} images match your filters. Add a query to see the best 200.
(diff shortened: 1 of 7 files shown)


kahuna/public/js/search/results.html
```diff
@@ -22,8 +22,8 @@
 
             <div class="results-toolbar-item results-toolbar-item--static
                     image-results-count">
-                <span ng-if="ctrl.isAiSearch">Best {{ctrl.aiResultsShown | toLocaleString}} of {{ctrl.totalResults | toLocaleString}} matches</span>
-                <span ng-if="! ctrl.isAiSearch">{{ctrl.totalResults | toLocaleString}} matches</span>
+                <span ng-if="ctrl.isAiSearch && !ctrl.filtersOnlyAiSearch">Best {{ctrl.aiResultsShown | toLocaleString}} of {{ctrl.totalResults | toLocaleString}} matches</span>
```
joelochlann: Could be clearer to wrap in a `<span ng-if="ctrl.isAiSearch>` since they both contain `ctrl.isAiSearch` condition?

media-api/app/controllers/MediaApi.scala
```diff
@@ -715,11 +741,11 @@ class MediaApi(
...
           performAiSearchAndRespond(_searchParams)
-        // Empty queries do not make sense for AI search as we can
-        // only rank results once we have a meaningful vector to compare with.
-        // So return 0 results if the query was empty.
+        // No query text to rank by, so we can't return ranked results. Instead
+        // return the size of the pool we'd be searching over, so the client can
+        // prompt the user to add a query.
         case _ =>
-          emptyAiSearchResponse
+          filtersOnlyAiSearchResponse(_searchParams)
```
joelochlann: Not a blocker, but I'm pretty sure there's an opportunity to simplify this, especially if you start by implementing this suggestion: https://github.com/guardian/grid/pull/4752#discussion_r3387038464

Basically if you make `AiQueryParts.from` do all the parsing and error handling, it will either give you a `Left` containing the reason why it can't do the search, or a `Right` with the parsed stuff ready to do a query. Then you should be able to call `filtersOnlyAiSearchResponse` in just one place, instead of three.

You could also make the `Left` be a union type (a sealed trait with some case objects) so you could pattern match on it to distinguish the expected cases you want to handle with your filters-only search, from the ones that are genuine error conditions.

media-api/app/controllers/MediaApi.scala
```diff
@@ -694,9 +715,14 @@ class MediaApi(
 
     def performAiSearchAndRespond(params: SearchParams): Future[Result] = {
       params.aiQueryParts.validationError match {
-        case Left(errorMessage) =>
-          logger.info(logMarker, s"Invalid AI search query: $errorMessage")
-          Future.successful(respondError(UnprocessableEntity, "invalid-ai-search", errorMessage))
+        case Left(_) =>
+          if (params.aiQueryParts.hasFilterConditions) {
+            logger.info(logMarker, s"AI search with only filters and no query to rank by; returning filter pool counts: ${params.aiQueryParts.filterConditions}")
+            filtersOnlyAiSearchResponse(params)
```
joelochlann: both these `if/else` branches contain `filtersOnlyAiSearchResponse`, maybe best to pull out

joelochlann (approved):
Nice! Functionality looks great 👌

Only UI thing that surprised me was: I did actually expect the badges (GNM-owned, agency picks) even in the filters-only mode, so that it's literally just the "Best 200 of" popping in when you add a query, i.e.

```
1,178,998 matches [4,705 GNM-owned] [5,960 agency picks]
```
to
```
Best 200 of 1,178,998 matches [4,705 GNM-owned] [5,960 agency picks]
```

But it doesn't matter much so it's only worth changing if it simplifies code or if someone else agrees with me.

On the code side, no blockers to this PR, but I'd definitely suggest having a go at simplifying the logic for distinguishing "normal AI search" from "filters-only AI search" from "an actual error", starting with https://github.com/guardian/grid/pull/4752#discussion_r3387038464 which I think you'll find satisfying and will simplify some boilerplate in the tests. If anything's unclear about that let me know and we could do a quick half an hour's pairing to align

media-api/test/lib/elasticsearch/AiQueryPartsTest.scala
```diff
@@ -19,111 +19,89 @@ class AiQueryPartsTest extends AnyFunSpec with Matchers with ImageFields {
...
-        parts.validationError.isLeft should be(true)
+        aiPartsFor("fileType:jpeg") should be(
+          Left(AiQueryError.NoRankingSignal(jpegFilter :: standardNegations))
+        )
       }
 
       it("rejects a negative-only query") {
-        val parts = aiPartsFor("-angela")
-
-        parts.semanticQuery should be(None)
```
joelochlann, line 112: 🙌 yes! this is the boilerplate cleanup I was hoping for 🧹

joelochlann (approved):
Beautiful!

Double approved! ✅✅

Image

media-api/app/lib/elasticsearch/ElasticSearchModel.scala
```diff
@@ -17,43 +17,43 @@ case class SearchResults(hits: Seq[(String, SourceWrapper[Image])], total: Long,
...
+sealed trait AiQueryError
+object AiQueryError {
+  // The query has no KNN ranking signal (no text and no similar image), so there's
+  // nothing to rank by. We still carry the parsed filter conditions so the caller can
+  // report how big the filtered pool is and prompt the user to add a query.
+  final case class NoRankingSignal(filterConditions: List[Condition]) extends AiQueryError
+
+  // The query combines a similar image with a text query. The two rankings can't be
+  // merged, so this is a genuine error.
+  case object ConflictingRankingSignals extends AiQueryError {
```
joelochlann, line 33: 🤩 this is a great way of modelling it

PR #4794, opened 2026-07-07
Bump @guardian/cql to 1.8.6

jonathonherbert:
## What does this change?

Bumps @guardian/cql from 1.8.4 to 1.8.6, to fix a few behaviour issues spotted by @paperboyo. Changelog is [here.](https://github.com/guardian/cql/releases/) 

Run locally, and deploy to CODE The issues specified as fixed in the changelog should be fixed.

### DIFF
--- a/kahuna/package-lock.json
+++ b/kahuna/package-lock.json
@@ -6,7 +6,7 @@
     "": {
       "dependencies": {
         "@babel/polyfill": "7.8.7",
-        "@guardian/cql": "1.8.5",
+        "@guardian/cql": "1.8.6",
         "@guardian/user-telemetry-client": "1.1.0",
         "@sentry/browser": "6.11.0",
         "@sentry/integrations": "6.11.0",
@@ -2243,9 +2243,9 @@
       "license": "MIT"
     },
     "node_modules/@guardian/cql": {
-      "version": "1.8.5",
-      "resolved": "https://registry.npmjs.org/@guardian/cql/-/cql-1.8.5.tgz",
-      "integrity": "sha512-VU8UPeFeoMUpEc4wMHkXo65nEXGWmxYu5CofKXNrOQvtGSwTE3mYYdwZT9IG22ILpeWhlMF4Wzmo1EnFfwffxA==",
+      "version": "1.8.6",
+      "resolved": "https://registry.npmjs.org/@guardian/cql/-/cql-1.8.6.tgz",
+      "integrity": "sha512-R8UoHaSiUVCOfgbFOCBxfPUQmUNoasd/Rj0o1KF7ikO0g+amsX3Dro2mabuzeuszMjAT4kMD809kuS+w+igvXg==",
       "dependencies": {
         "@floating-ui/dom": "^1.6.11",
         "preact": "^10.25.4",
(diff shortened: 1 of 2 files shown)


paperboyo (approved):
Thank you! Verified all (but “detect polarity changes” as this one would need a change in Grid to be user-visible).

paperboyo:
> this one would need a change in Grid to be user-visible

Because I’m basically a robot slave, I asked them to come up with a fix for something they suggested need fixing. But even after involving multiple human persons in the dance around that ask, I didn’t understand it.

So I asked for an explanation, so I look less like a joke. But I still don’t really get it. Which is fine. Very fine indeed. 

Here is the explanation for when I (won’t ever) get wiser with age. When it says “fixed”, it doesn’t mean real Grid. The code it references is not real Grid.

I really have no idea why I even thought it might be (not) useful to post this comment. But the question of what artifacts should survive the lonely robot sessions, is a good one. I think.

---

🤖 narrative:

What the bug/fix actually is, walked through with is:under-quota:

1. State: User has searched, and the query is is:under-quota (positive — showing only under-quota images). A visible chip in the CQL search box shows this, styled as an "include" chip.

2. User action: They alt-click the same is:under-quota entry somewhere (e.g. the Filters panel) wanting to exclude under-quota images instead.

3. App logic (unaffected by the bug): `upsertFieldTerm(query, "is", "under-quota", negated: true)` runs. `findFieldTerm` parses the AST, finds the existing is:under-quota term, sees it's the opposite polarity to what's wanted, and splices it in place — the query string becomes -is:under-quota. There's never a moment with both chips; it's a single term whose sign changed. This part always worked correctly — the URL updates, and the ES query genuinely excludes under-quota images. The data layer was never the problem.

4. The bug (now fixed): That new string -is:under-quota gets passed down as a prop and the app calls `cqlInputRef.current.setAttribute("value", "-is:under-quota")` on the `<cql-input> web component. Internally, CQL parses that string into a new ProseMirror document and diffs it against the currently displayed document (which still shows the old +is:under-quota chip) to compute a minimal patch — this preserves cursor position and avoids a jarring full re-render. The diff function compared chip nodes by their text content only, not their POLARITY attribute. Since the key (is) and value (under-quota) text are identical between old and new, the diff concluded "nothing changed" and skipped patching. Result: the search results correctly excluded under-quota images, but the chip pill still visually showed the old, positive (green/include) styling — a UI lie: what you see in the box doesn't match what's actually being searched.

5. The workaround (just removed): Force-remount `<CqlSearchInput>` entirely (bump a generation counter used as a React `key` whenever an external polarity flip happens. A brand-new `<cql-input>` element has no prior document to diff against — it just parses -is:under-quota fresh and renders it correctly. Costs ~10ms and a tiny flash on every cell/panel click that flips polarity.

6. The upstream fix (PR #121, now in 1.8.6): The diff function now also compares the POLARITY attribute (via a new sameContentMarkup check), while still ignoring genuinely transient attributes like IS_SELECTED/IS_READ_ONLY (which exist specifically to avoid reintroducing an older cursor-jump bug, #52). So the diff now correctly detects "same key/value, different polarity" and patches just that one attribute in place — `setAttribute("value", "-is:under-quota")` now flips the visible chip's colour/sign immediately, with no remount needed.

So the bug was purely a rendering-sync issue between the app's query string (always correct) and the chip's visual polarity (could go stale) — not a query-construction issue, and not related to two contradictory chips coexisting.

PR #4795, opened 2026-07-07
Dynamo Part 3: Usages upgrade

lindseydew:

- [x] Adding a usage from composer shows in the tab
- [x] Publishing in composer updates the usage tab
- [x] Deleting image from composer + publishing shows the image as taken down in the usage tab
- [x] Delete ALL removes the usages from the dynamo table, but does not affect what is in composer
- [x] Check order is sensible for different usage types

## Tested? Documented?
- [ ] locally by committer
    - [ ] Syndication
        - [ ] getAllImageRightsInPhotoshoot
        - [ ] publishChangedSyndicationRightsForPhotoshoot
    - [ ] Usages
        - [ ] Metadata SQS consumer
    - [ ] ..?


common-lib/src/main/scala/com/gu/mediaservice/lib/usage/ItemToMediaUsage.scala
```diff
@@ -5,42 +5,46 @@ import com.amazonaws.services.dynamodbv2.document.Item
...
-      }.toOption.map(new DateTime(_))
+      UsageId(doc.getString("usage_id")),
+      doc.getString("grouping"),
+      doc.getString("media_id"),
+      UsageType(doc.getString("usage_type")),
+      doc.getString("media_type"),
+      UsageStatus(doc.getString("usage_status")),
+      Option(doc.getMapOfUnknownType("print_metadata"))
+        .map(_.asScala.toMap)
+        .flatMap(buildPrint),
```
andrew-nowak: there's a bug lurking in these somewhere...

When using the usages' `GET /usages/:id` endpoint the usage metadata (`printMetadata`, `digitalMetadata`, etc.) isn't displayed. I think this is because the `buildPrint` and `buildDigital` methods use `Try { ... }.toOption`, which catches and swallows any errors. I can't immediately see what the underlying problem would be, we'd need to add some extra logging or run locally with the debugger to find it

andrew-nowak: I ran it through the debugger locally, the errors are variations of `class software.amazon.awssdk.services.dynamodb.model.AttributeValue cannot be cast to class java.lang.String`

It looks like `getMapOfUnknownType` returns maps of type `Map[String, AttributeValue]`, and you can get the underlying value of `AttributeValue`s with `.s()`, `.n()`, etc. so `buildPrint` and friends no longer need to use type casts

lindseydew: Ah good spot, thank you!

common-lib/src/main/scala/com/gu/mediaservice/lib/usage/ItemToMediaUsage.scala
```diff
@@ -40,6 +43,42 @@ object ItemToMediaUsage {
...
+      UsageId(doc.getString("usage_id")),
+      doc.getString("grouping"),
+      doc.getString("media_id"),
+      UsageType(doc.getString("usage_type")),
+      doc.getString("media_type"),
+      UsageStatus(doc.getString("usage_status")),
+      Option(doc.getMap(
+          "print_metadata",
+          EnhancedType.of(classOf[String]),
+          EnhancedType.of(classOf[AttributeValue])
```
andrew-nowak, line 58: This line is throwing when trying to submit a print usage

To test it locally, I've created a test usage file `print.json` (remember to change `mediaId` to match the image you're testing with)

```json
{
  "printUsageRecords": [{
    "dateAdded" : "2026-07-15T12:13:14.123Z",
    "mediaId": "27bc178cfed8ce27d3fbaaf15501e51fe865acae",
    "printUsageMetadata": {
      "sectionName": "sec",
      "issueDate": "2026-07-16",
      "pageNumber": 5,
      "storyName": "stor",
      "publicationCode": "pubcode",
      "publicationName": "pubname",
      "layoutId": 5,
      "edition": 2,
      "size": {
        "x": 5,
        "y": 10
      },
      "orderedBy": "andrew nowak",
      "sectionCode": "seccode",
      "notes": "hmm nteresting",
      "source": "src"
    },
    "containerId": "cont",
    "usageId": "xyz",
    "usageStatus": "published"
  }]
}
```
and then submitted it with
```sh
curl -i -H 'x-gu-media-key: dev-' -H 'Content-type: application/json' -d @print.json 'https://media-usage.local.dev-gutools.co.uk/usages/print'
```
Usage responds with 202 Accepted, but the usage never arrives on the image, and you can find the failure stack trace in `usage/logs/application.log`

lindseydew: Thank you for the spot! I have made a commit to fix this

lindseydew: update-usages

lindseydew: This is now working for me locally with this change

common-lib/src/main/scala/com/gu/mediaservice/lib/usage/ItemToMediaUsage.scala
```diff
@@ -40,6 +44,42 @@ object ItemToMediaUsage {
...
+        .map(_.asScala.toMap)
+        .flatMap(buildChild),
+      new DateTime(doc.getNumber("last_modified").longValue()),
+      Try(doc.getNumber("date_added").longValue()).toOption.map(new DateTime(_)),
+      Try(doc.getNumber("date_removed").longValue()).toOption.map(new DateTime(_))
+    )
+  }
+
 
   private def buildFront(metadataMap: Map[String, Any]): Option[FrontUsageMetadata] = {
```
andrew-nowak: nit: You can do the same here as you did for `buildDigital` and `buildDownload` etc. - convert the map type to `Map[String, String]` and remove the `.asInstanceOf` casts below

common-lib/src/main/scala/com/gu/mediaservice/lib/usage/ItemToMediaUsage.scala
andrew-nowak: here too

andrew-nowak (approved):

PR #4814, opened 2026-07-20
Reword advanced search tooltip

hk-15:
## What does this change?

It's not very obvious that the info icon next to the search bar in Grid homepage is a clickable button that will display a whole panel of information about searching. This PR changes the tooltip wording to make it clearer that it is clickable and adds a background on hover.

| Before | After|
| ------- | ----- |
|Screenshot 2026-07-20 at 14 10 32 | Screenshot 2026-07-21 at 10 14 58  |

Hover over the info icon.

More users are aware that this is a clickable button and are using it. Especially helpful for newer users.

## Tested? Documented?
- [x] locally by committer


junyuanxue (approved):
Yayy! I'm wondering if giving the button element a background on hover would achieve a more 'button-like' effect? Can always add that in a separate PR though

PR #4818, opened 2026-07-21
Add initial loading state for crop view

junyuanxue:
## What does this change?
Adds a loading state to the crop view so users get immediate feedback instead of a frozen screen during the a few seconds wait for the optimised image to be generated/downloaded.
* Remove `optimisedImageUri` from the blocking resolve — the state transition now only waits on the fast image/imageId resolves, so the crop view renders immediately.
* Inject `imgops` directly into crop controller to fetch the optimised image URI asynchronously after load
* Loading state is then reset when the crop box is ready

* Go to the crop view of an image. You may want to add some throttling when testing locally
loading image

## Tested? Documented?
- [x] locally by committer

---
  - https://app.asana.com/0/0/1216668121615122

### DIFF
--- a/kahuna/public/js/crop/controller.js
+++ b/kahuna/public/js/crop/controller.js
@@ -5,12 +5,14 @@ import {radioList} from '../components/gr-radio-list/gr-radio-list';
 import {cropUtil} from "../util/crop";
 import {cropOptions} from "../util/constants/cropOptions";
 import {storage as storageModule} from "../util/storage";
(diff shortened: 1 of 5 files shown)


paperboyo:
Tested on TEST, works well, thank you 🥳 !
(haven’t noticed speedup, but not sure if I should expect any)

lindseydew (approved):
LGTM 👍

PR #4806, opened 2026-07-15
default AI feature switch to true

ellenmuller:
## What does this change?
This should be merged before morning conference on Wednesday, when we're releasing AI search 🚀

If something goes wrong, we can merge this https://github.com/guardian/grid/pull/4813 to turn the AI search off.

PR #4824, opened 2026-07-22
remove feature switch

ellenmuller:
## What does this change?
It's actually slightly easier to just remove the feature switch, which can be easily re-instated (either with default true or false!).


joelochlann:
This looks good, but let's wait til after Morning Conference to discuss, test, review etc. Unless of course there's an urgent problem with #4806

PR #4832, opened 2026-07-27
Upgrade S3 - Part 2

lindseydew:
## What does this change?

This upgrades the store methods in the s3 class

- [x] Test can make a crop
- [x] Test images are stored

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
@@ -7,11 +7,12 @@ import com.amazonaws.{AmazonServiceException, ClientConfiguration}
 import com.gu.mediaservice.lib.config.CommonConfig
 import com.gu.mediaservice.lib.logging.{GridLogging, LogMarker, Stopwatch}
 import com.gu.mediaservice.model._
-import org.joda.time.{DateTime, Duration}
+import org.joda.time.{DateTime, DateTimeZone, Duration}
 import software.amazon.awssdk.core.ResponseInputStream
+import software.amazon.awssdk.core.sync.RequestBody
 import software.amazon.awssdk.regions.Region
 import software.amazon.awssdk.services.s3.S3Client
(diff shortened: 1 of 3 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
```diff
@@ -7,11 +7,12 @@ import com.amazonaws.{AmazonServiceException, ClientConfiguration}
...
 import com.gu.mediaservice.lib.logging.{GridLogging, LogMarker, Stopwatch}
 import com.gu.mediaservice.model._
-import org.joda.time.{DateTime, Duration}
+import org.joda.time.{DateTime, DateTimeZone, Duration}
 import software.amazon.awssdk.core.ResponseInputStream
+import software.amazon.awssdk.core.sync.RequestBody
 import software.amazon.awssdk.regions.Region
 import software.amazon.awssdk.services.s3.S3Client
-import software.amazon.awssdk.services.s3.model.{GetObjectResponse, NoSuchKeyException, GetObjectRequest => GetObjectRequestV2}
+import software.amazon.awssdk.services.s3.model.{GetObjectResponse, HeadObjectRequest, HeadObjectResponse, NoSuchKeyException, GetObjectRequest => GetObjectRequestV2, PutObjectRequest => PutObjectRequestV2}
```
twrichards, line 15: is there a reason we're aliasing the imports? I can understand if this is temporary to make things more explicit during the transition phase, but after that I think it probably leaves behind confusion as people might be looking at the code further down and be googling the class only to find there's no such things as `PutObjectRequestV2` for example

twrichards (approved):

common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
lindseydew: Yeah, it was just to help me keep track of things whilst this is in flight. I have a branch to deprecate the v1 stuff after so it'll only be temporary

PR #4834, opened 2026-07-27
Upgrade S3 - Part 4

lindseydew:
## What does this change?
This migrates more s3 methods to v2

Reaper
- [x] Test soft deleted ids are still present in the IUI
- [x] Test hard deleted ids are not searchable in the UI
- [x] Tested removed from bucket

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
@@ -194,12 +194,17 @@ class S3(config: CommonConfig) extends GridLogging with ContentDisposition with 
       val listing = clientV2.listObjectsV2(req)
       val s3Objects = listing.contents().asScala.toList
       s3Objects.map(s3Object => {
-        S3Object(bucket, s3Object.key(), size = s3Object.size(), metadata = getMetadata(bucket, s3Object.key()))
+        S3Object(bucket, s3Object.key(), size = s3Object.size(), metadata = getMetadataV2(bucket, s3Object.key()))
       })
     }
 
   def getMetadata(bucket: Bucket, key: Key): S3Metadata = {
     val meta = client.getObjectMetadata(bucket, key)
+    S3Metadata(meta)
+  }
+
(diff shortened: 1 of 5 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/aws/S3.scala
```diff
@@ -223,6 +223,17 @@ class S3(config: CommonConfig) extends GridLogging with ContentDisposition with
...
+  def doesObjectExistV2(bucket: Bucket, key: String) = {
+    try {
+      clientV2.headObject(
+        HeadObjectRequest.builder().key(key).bucket(bucket).build()
+      )
+      true
+    } catch {
+      case _: NoSuchKeyException => false
+    }
+  }
```
twrichards, line 239: shame they got rid of the more readable [`doesObjectExist`](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/AmazonS3.html#doesObjectExist-java.lang.String-java.lang.String-) on the client

thrall/app/controllers/ReaperController.scala
```diff
@@ -109,7 +111,9 @@ class ReaperController(
     case Some(reaperBucket) => doBatchDelete.map { json =>
       val now = DateTime.now(DateTimeZone.UTC)
       val key = s"$deleteType/${s3DirNameFromDate(now)}/$deleteType-${now.toString()}.json"
-      store.client.putObject(reaperBucket, key, json.toString())
+      store.clientV2.putObject(
+        PutObjectRequest.builder().bucket(reaperBucket).key(key).build(),
```
twrichards: is there a reusable function in our own S3 abstraction that can do this?

twrichards (approved):
LGTM - one comment though (that if true applies to a few places in this PR)

thrall/app/controllers/ReaperController.scala
lindseydew: Yeah, added a putMethod in the generic class 👍

PR #4836, opened 2026-07-27
Upgrade S3 - Part 6 

lindseydew:
## What does this change?
This migrates more of the s3 methods in image loader

- [x] Test crops still work
- [x] Test upload a non-image

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/image-loader/app/controllers/ImageLoaderController.scala
+++ b/image-loader/app/controllers/ImageLoaderController.scala
@@ -38,7 +38,7 @@ import software.amazon.awssdk.services.s3.model.{GetObjectRequest, HeadObjectReq
 
 import java.io.{File, FileOutputStream}
 import java.net.URI
-import java.time.Instant
+import java.time.{Duration, Instant}
 import java.util.concurrent.atomic.AtomicReference
 import scala.concurrent.{ExecutionContext, Future}
 import scala.jdk.CollectionConverters.MapHasAsScala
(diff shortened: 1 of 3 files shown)


image-loader/app/controllers/ImageLoaderController.scala
```diff
@@ -241,6 +241,7 @@ class ImageLoaderController(auth: Authentication,
 
   def getPreSignedUploadUrlsAndTrack: Action[AnyContent] = AuthenticatedAndAuthorised.async { request =>
     val expiration = DateTimeUtils.now().plusHours(1)
```
twrichards, line 243: is `expiration` still used at all, given it's no longer used in the `preSignedUrl` bit further down... if not please can it be removed... if it is still in use, please can `signatureDuration` be defined first and then `expiration` calculated by adding `signatureDuration` to `DateTimeUtils.now()`

twrichards (approved):

image-loader/app/controllers/ImageLoaderController.scala
lindseydew: It's used to set the status in the a dynamo table:
https://github.com/guardian/grid/blob/815f1ae637c50428dcbf508c60ab77305c2614dd/image-loader/app/controllers/ImageLoaderController.scala#L259

PR #4837, opened 2026-07-27
Upgrade S3 - Part 7

lindseydew:
## What does this change?
This upgrades more of the paths

- [x] Test the projector functionality 
- should show the diff if the pipeline is working ok
- the `should` check seems not to be working. 
- [x] Test upload pipeline still works

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/image-loader/app/model/Projector.scala
+++ b/image-loader/app/model/Projector.scala
@@ -1,13 +1,14 @@
 package model
 
+import software.amazon.awssdk.core.ResponseInputStream
+import software.amazon.awssdk.services.s3.model.GetObjectResponse
+
 import java.io.{File, FileOutputStream}
-import com.amazonaws.services.s3.AmazonS3
 import com.gu.mediaservice.{GridClient, ImageDataMerger}
 import com.gu.mediaservice.lib.auth.Authentication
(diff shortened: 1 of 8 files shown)


image-loader/app/model/Projector.scala
```diff
@@ -42,12 +42,12 @@ case class S3FileExtractedMetadata(
 )
 
 object S3FileExtractedMetadata {
-  def apply(s3ObjectMetadata: ObjectMetadata): S3FileExtractedMetadata = {
-    val lastModified = new DateTime(s3ObjectMetadata.getLastModified)
-    val userMetadata = s3ObjectMetadata.getUserMetadata.asScala.toMap
+
+  def apply(s3Object: S3Object): S3FileExtractedMetadata = {
+    val lastModified = s3Object.metadata.objectMetadata.lastModified.getOrElse(new DateTime)
```
twrichards: this looks to be subtly different behaviour from before, previously did `getLastModified` return the current time if there was no value??

twrichards (approved):

image-loader/app/model/Projector.scala
lindseydew: I actually don't think we are even using this method, so I have removed it 👍

PR #4838, opened 2026-07-27
Upgrade S3 - Part 8 

lindseydew:
## What does this change?
This updates the presign url 

- [ ] Test the images load in the main page ok
- Merge carefully 

## Tested? Documented?
- [ ] locally by committer


twrichards (approved):
was there pre-signing going on elsewhere in the prior PRs that could be consilidated with this impl. in common-lib?

lindseydew:
> was there pre-signing going on elsewhere in the prior PRs that could be consilidated with this impl. in common-lib?

Yep, we can re-use the same client I believe. The methods are sufficiently that they still need to be separate 👍

PR #4839, opened 2026-07-27
S3 Upgrade - Part 9 

lindseydew:
## What does this change?
This migrates the delete methods

- [ ] Check things are removed from the bucket

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/S3ImageStorage.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/S3ImageStorage.scala
@@ -29,6 +29,7 @@ class S3ImageStorage(config: CommonConfig) extends S3(config) with ImageStorage 
   }
 
   def deleteImage(bucket: String, id: String)(implicit logMarker: LogMarker) = Future {
+    logger.info(logMarker, s"Deleted image $id from bucket $bucket")
     clientV2.deleteObject(
       DeleteObjectRequest.builder().bucket(bucket).key(id).build())
   }
--- a/common-lib/src/main/scala/com/gu/mediaservice/lib/S3ImageStorage.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/lib/S3ImageStorage.scala
@@ -5,6 +5,7 @@ import com.gu.mediaservice.lib.config.CommonConfig
 import com.gu.mediaservice.lib.logging.{GridLogging, LogMarker}
 import com.gu.mediaservice.model.MimeType
 import org.slf4j.LoggerFactory
+import software.amazon.awssdk.services.s3.model.{DeleteObjectRequest, HeadObjectRequest, ListObjectsV2Request}
 
 import java.io.File
 import scala.jdk.CollectionConverters._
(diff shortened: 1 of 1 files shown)


common-lib/src/main/scala/com/gu/mediaservice/lib/S3ImageStorage.scala
```diff
@@ -28,19 +29,23 @@ class S3ImageStorage(config: CommonConfig) extends S3(config) with ImageStorage
   }
 
   def deleteImage(bucket: String, id: String)(implicit logMarker: LogMarker) = Future {
-    client.deleteObject(bucket, id)
-    logger.info(logMarker, s"Deleted image $id from bucket $bucket")
```
twrichards, line 32: are we sure we want to drop the logger line??

common-lib/src/main/scala/com/gu/mediaservice/lib/S3ImageStorage.scala
```diff
@@ -28,19 +29,23 @@ class S3ImageStorage(config: CommonConfig) extends S3(config) with ImageStorage
...
+    clientV2.deleteObject(
+      DeleteObjectRequest.builder().bucket(bucket).key(id).build())
   }
 
   def deleteVersionedImage(bucket: String, id: String)(implicit logMarker: LogMarker) = Future {
```
twrichards, line 37: appreciate this question applies to the existing signature and not your changes, but curious that we have such a function that doesn't takes in a version ID, does the distinction exist purely to log the version id at the point of deletion?

common-lib/src/main/scala/com/gu/mediaservice/lib/S3ImageStorage.scala
```diff
@@ -28,19 +29,23 @@ class S3ImageStorage(config: CommonConfig) extends S3(config) with ImageStorage
   }
 
   def deleteImage(bucket: String, id: String)(implicit logMarker: LogMarker) = Future {
```
twrichards: kinda strange that the argument here is `id` and not `key`, it's not like it calls the function to turn the media id into a key (i.e. using the first few characters of the media id for a sharded prefix e.g. abcde123 -> a/b/c/d/e/abcde123) perhaps the arg should be renamed to key to better convey the intent of the function (same with at least `deleteVersionedImage` and `deleteFolder` below)

common-lib/src/main/scala/com/gu/mediaservice/lib/S3ImageStorage.scala
```diff
@@ -28,19 +29,23 @@ class S3ImageStorage(config: CommonConfig) extends S3(config) with ImageStorage
...
 
   def deleteFolder(bucket: String, id: String)(implicit logMarker: LogMarker) = Future {
-		val files = client.listObjects(bucket, id).getObjectSummaries.asScala
-		files.foreach(file => client.deleteObject(bucket, file.getKey))
+    val files = clientV2.listObjectsV2(
+      ListObjectsV2Request.builder().bucket(bucket).prefix(id).build()
+    ).contents().asScala.toList
+    files.foreach(file => clientV2.deleteObject(
+      DeleteObjectRequest.builder().bucket(bucket).key(file.key()).build()
+    ))
```
twrichards, line 49: observation on the existing code rather than your changes, but this should really use bulk delete rather than iterating through every file in the folder with an API request for each (might encounter rate limiting etc).

twrichards (approved):

common-lib/src/main/scala/com/gu/mediaservice/lib/S3ImageStorage.scala
lindseydew: Oh thanks good spot!

common-lib/src/main/scala/com/gu/mediaservice/lib/S3ImageStorage.scala
lindseydew: Yep, that's a good point 👍 will make a note for a future refactor

PR #4851, opened 2026-07-31
add z-index to .circular-mask-toggle-container

dblatcher:
## What does this change?

Adds a z-index of 1 to the container for the "circular mask" toggle.

A user reported that they could not toggle the option. The reason seems to be that it is masked by the ".easel__canvas"  element. Adding the z-index puts the toggle "above" so it can still be clicked.

Run locally, open an image, select "crop image" (top right of UI), then select 1:1 crop.
 on this branch, the toggle should be clickable.

On main, the toggle is not clickable, but still reachable using the keyboard to tab it into focus

restores functionality

## Tested? Documented?

Just tested by changing the css in the browser (don't have grid set up locally ATM)

- [ ] locally by committer

### DIFF
--- a/kahuna/public/stylesheets/main.css
+++ b/kahuna/public/stylesheets/main.css
@@ -2096,6 +2096,7 @@ FIXME: what to do with touch devices
   right: 0;
   padding: 10px;
   user-select: none;
+  z-index: 1;
 }
 
 


junyuanxue (approved):
Thanks for that!

PR #4852, opened 2026-07-31
remove feature switch

ellenmuller:
## What does this change?
Now that we have the config setting, we can turn AI search on and off by updating the config and re-deploying. 

It's a lot easier for state management to get rid of the feature switch, so let's remove it!


bryophyta (approved):
It looks like there's a TypeScript definition of `Window._clientConfig` which doesn't include the `aiSearchEnabled` field. This was a pre-existing issue I don't think this should be a blocker, but could be a good opportunity to add it in?

Other than that, this looks good to me!

PR #4856, opened 2026-08-06
Script Alamy changes

lindseydew:
## What does this change?
This adds a custom script which will either delete or update the usage rights of selected images in the grid to be applied in PROD. 

## Tested? Documented?
- [ ] locally by committer


scripts/src/main/scala/com/gu/mediaservice/scripts/AlamyCleanUp.scala
```diff
@@ -0,0 +1,58 @@
...
+import play.libs.ws.WSClient
+
+import java.net.URI
+import java.net.http.HttpRequest.BodyPublishers
+import java.net.http.HttpResponse.BodyHandlers
+import java.net.http.{HttpClient, HttpRequest, HttpResponse}
+import scala.io.Source
+
+object AlamyCleanUp extends App {
+  val GRIDDOMAIN = "test.dev-gutools.co.uk"
```
junyuanxue: I'm wondering if we could pass a `stage` argument for TEST or PROD environment to infer the domain from

scripts/src/main/scala/com/gu/mediaservice/scripts/AlamyCleanUp.scala
```diff
@@ -0,0 +1,58 @@
...
+  val GRIDKEY = sys.env.getOrElse("GRIDKEY", throw new RuntimeException("Must set a GRIDKEY env varaible"))
+  val resource = Source.fromResource("alamy-grid-ids.csv")
+  val ids = resource.getLines().toList
+  val requestBody = BodyPublishers.ofString("""{"data":{"restrictions":"No longer available from Alamy","category":"chargeable"}}""")
+  
+  val outcomes = ids.zipWithIndex.map({case (id, index) =>
+    println(s"Running for id $id, $index of ${ids.size}")
```
junyuanxue: Should this be `index + 1`?

lindseydew: Good spot, thanks 😊

scripts/src/main/scala/com/gu/mediaservice/scripts/AlamyCleanUp.scala
```diff
@@ -0,0 +1,67 @@
...
+import play.libs.ws.WSClient
+
+import java.net.URI
+import java.net.http.HttpRequest.BodyPublishers
+import java.net.http.HttpResponse.BodyHandlers
+import java.net.http.{HttpClient, HttpRequest, HttpResponse}
+import scala.io.Source
+
+object AlamyCleanUp extends App {
+  val GRIDDOMAIN = sys.env.getOrElse("GRIDDOMAIN", throw new RuntimeException("Must set a GRIDDOMAIN env varaible"))
```
junyuanxue:
```suggestion
  val GRIDDOMAIN = sys.env.getOrElse("GRIDDOMAIN", throw new RuntimeException("Must set a GRIDDOMAIN env variable"))
```
Same below 😂 
Though I wonder if we need to set both `GRIDDOMAIN` and `STAGE`? If we specify the `STAGE`, we should be able to construct the domain in here, though I could have missed something 😅

lindseydew: oh yeah, that does seem more sensible 👍

junyuanxue (approved):
Looks great! 🎉

PR #4861, opened 2026-08-10
PR and Third Party mapping

lindseydew:
## What does this change?
This introduces the change to map existing categories to PR and Third Party. 

The user experience changes are:

#### Displaying previously selected legacy category

- Any legacy categories will now display as `PR and Third Party`. 

- If `restrictions` or `source` these will be visible in the editable fields. 

- Where a category has a `supplier` or `suppliers` set, this will be mapped to `source`

- Any other fields will not be displayed or edited in the U.I, but can still be accessed at the API level

- Restrictions will now be set to required, and making a modification will enforce a user sets one. 

#### Searching for images in the U.I
- Any legacy categories will display when a user selects `pr-and-third-party` in the drop down search
- The legacy categories will not be available to search in the drop down, but the URL can be changed manually and searching for a specific legacy category will yield the appropriate results

#### Selecting a category
- The legacy categories have been removed from the dropdown and `PR and Third Party` has been added

The main code changes this involved was:

1) changing the logic of the elastic search query to perform a many search across all the legacy category when searching for `pr-and-third-party`

2) Rendering the images from ElasticSearch will perform a transformation to map a legacy category to PR and Third Party

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
@@ -629,11 +629,10 @@ object PRAndThirdParty extends UsageRightsSpec {
 
   override val category: String = "pr-and-third-party"
 
-  override def name(config: CommonConfig): String = "PR & Third Party"
+  override def name(config: CommonConfig): String = "PR & third party"
 
-  override def description(config: CommonConfig): String = "Images received from PRs or as handouts, by default you must explain the restrictions that apply. Also use this category for social media posts, screen grabs, agency commissions and obituraries."
+  override def description(config: CommonConfig): String = "Images received from PRs or as handouts, by default you must explain the restrictions that apply. Also use this category for social media posts, screen grabs, agency commissions and obituaries."
 
-  // TODO - check this
   override val defaultCost: Option[Cost] = Some(Conditional)
 
   implicit val formats: Format[PRAndThirdParty] =
--- a/common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
+++ b/common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
@@ -634,8 +634,26 @@ object PRAndThirdParty extends UsageRightsSpec {
   override def description(config: CommonConfig): String = "Images received from PRs or as handouts, by default you must explain the restrictions that apply. Also use this category for social media posts, screen grabs, agency commissions and obituraries."
 
   // TODO - check this
-  override val defaultCost: Option[Cost] = None
+  override val defaultCost: Option[Cost] = Some(Conditional)
 
   implicit val formats: Format[PRAndThirdParty] =
     UsageRights.subtypeFormat(PRAndThirdParty.category)(Json.format[PRAndThirdParty])
+
+  val legacyCategories: List[String] = List(
+    Agency.category,
+    CommissionedAgency.category,
+    PrImage.category,
+    Handout.category,
+    Screengrab.category,
+    SocialMedia.category,
+    Obituary.category,
+    Pool.category,
+    CrownCopyright.category,
+    CreativeCommons.category,
+    PublicDomain.category,
+    GuardianWitness.category,
+    Composite.category,
+  )
+
+  val allCategories: NonEmptyList[String] = NonEmptyList.fromSeq(category, legacyCategories)
 }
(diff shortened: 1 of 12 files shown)


common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
```diff
@@ -617,3 +619,41 @@ object ProgrammesAcquisitions extends UsageRightsSpec {
...
+  override val defaultCost: Option[Cost] = PRAndThirdParty.defaultCost
+}
+
+object PRAndThirdParty extends UsageRightsSpec {
+
+  override val category: String = "pr-and-third-party"
+
+  override def name(config: CommonConfig): String = "PR & Third Party"
+
+  override def description(config: CommonConfig): String = "Images received from PRs or as handouts, by default you must explain the restrictions that apply. Also use this category for social media posts, screen grabs, agency commissions and obituraries."
```
hk-15: Little typo. Also, do we need to add the second line of the description (pulled from Google sheet)? Maybe not as it seems repetitive and also too long! Thought I'd just check.
`You must explain what restrictions apply to the image as you upload. Where images have been submitted for inclusion in obituaries only please indicate this in the restrictions.`

```suggestion
  override def description(config: CommonConfig): String = "Images received from PRs or as handouts, by default you must explain the restrictions that apply. Also use this category for social media posts, screengrabs, agency commissions and obituaries."
```

media-api/test/lib/elasticsearch/Fixtures.scala
```diff
@@ -1,12 +1,18 @@
 package lib.elasticsearch
 
+import com.gu.mediaservice.lib.config.GridConfigResources
```
hk-15: I think all the added imports in this file are unused?

common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
```diff
@@ -617,3 +619,41 @@ object ProgrammesAcquisitions extends UsageRightsSpec {
...
+final case class PRAndThirdParty(restrictions: Option[String] = None, source: Option[String] = None) extends UsageRights {
+
+  override val defaultCost: Option[Cost] = PRAndThirdParty.defaultCost
+}
+
+object PRAndThirdParty extends UsageRightsSpec {
+
+  override val category: String = "pr-and-third-party"
+
+  override def name(config: CommonConfig): String = "PR & Third Party"
```
hk-15: Might need to double check but I believe this was the preferred casing?
```suggestion
  override def name(config: CommonConfig): String = "PR & third party"
```

common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
```diff
@@ -617,3 +619,41 @@ object ProgrammesAcquisitions extends UsageRightsSpec {
...
+
+  override def description(config: CommonConfig): String = "Images received from PRs or as handouts, by default you must explain the restrictions that apply. Also use this category for social media posts, screen grabs, agency commissions and obituraries."
+
+  // TODO - check this
+  override val defaultCost: Option[Cost] = Some(Conditional)
```
hk-15, line 636: Still need the TODO?

hk-15 (commented):
This is exciting!

media-api/test/lib/elasticsearch/Fixtures.scala
lindseydew: thanks, good spot

common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
lindseydew: yes, I think that's right 👍

common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
lindseydew: Ah no, we did check this!

common-lib/src/main/scala/com/gu/mediaservice/model/UsageRights.scala
lindseydew: Yeah, I'm not sure, let's check this one with David 👍 

I updated the typo in obituaries, but intellij tells me screengrabs is incorrect 🤔

hk-15: Intellij may be right! I suggested that because screengrabs is how we currently style it in the Grid 😊

PR #4863, opened 2026-08-11
Add logic for setting a label

lindseydew:
## What does this change?
This updates the Alamy script to set a label

## Tested? Documented?
- [ ] locally by committer

### DIFF
--- a/scripts/src/main/scala/com/gu/mediaservice/scripts/AlamyCleanUp.scala
+++ b/scripts/src/main/scala/com/gu/mediaservice/scripts/AlamyCleanUp.scala
@@ -55,7 +55,7 @@ object AlamyCleanUp extends App {
             false
           }
         } else {
-          println(s"Unable to set usage rights for $id, recevied response ${response.statusCode()}")
+          println(s"Unable to set usage rights for $id, received response ${response.statusCode()}")
           false
         }
       }
--- a/scripts/src/main/scala/com/gu/mediaservice/scripts/AlamyCleanUp.scala
+++ b/scripts/src/main/scala/com/gu/mediaservice/scripts/AlamyCleanUp.scala
@@ -37,7 +37,7 @@ object AlamyCleanUp extends App {
         true
       }
       case 405 => {
-        println(s"Unable to delete $id, setting usages instead and labels")
+        println(s"Unable to delete $id, setting usages and labels instead")
         val request = HttpRequest.newBuilder(new URI(s"https://media-metadata.$GRIDDOMAIN/metadata/$id/usage-rights"))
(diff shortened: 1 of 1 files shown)


hk-15 (approved):

scripts/src/main/scala/com/gu/mediaservice/scripts/AlamyCleanUp.scala
```diff
@@ -37,13 +37,23 @@ object AlamyCleanUp extends App {
         true
       }
       case 405 => {
-        println(s"Unable to delete $id, setting usages instead")
+        println(s"Unable to delete $id, setting usages instead and labels")
```
junyuanxue:
```suggestion
        println(s"Unable to delete $id, setting usages and labels instead")
```
?

scripts/src/main/scala/com/gu/mediaservice/scripts/AlamyCleanUp.scala
```diff
@@ -37,13 +37,23 @@ object AlamyCleanUp extends App {
...
+            .headers("X-Gu-Media-Key", GRIDKEY, "Content-Type", "application/json").POST(labelBody).build()
+          val response = client.send(request, BodyHandlers.ofString())
+          if(response.statusCode() == 200) {
+            println(s"Successfully set a label for $id")
+            true
+          }
+          else {
+            println(s"Unable to set label for $id, received response ${response.statusCode()}")
+            false
+          }
```
junyuanxue, line 56: This works and not too fussed about readability since it's a script, but if you feel like refactoring I'm wondering if it's worth extracting out things into functions, or I don't know if Scala has a similar pattern like JS to chain functions on success or failure:
```ts
deleteImage
  .then(() => {})
  .catch(
      setUsageRights
          .then(setLabels)
          .catch(() => {})
   )
```

scripts/src/main/scala/com/gu/mediaservice/scripts/AlamyCleanUp.scala
```diff
@@ -37,13 +37,23 @@ object AlamyCleanUp extends App {
...
         } else {
           println(s"Unable to set usage rights for $id, recevied response ${response.statusCode()}")
```
hk-15: Wee typo I just spotted!
```suggestion
          println(s"Unable to set usage rights for $id, received response ${response.statusCode()}")
```

scripts/src/main/scala/com/gu/mediaservice/scripts/AlamyCleanUp.scala
lindseydew: Yeah agree it's got a bit nested, I think for the sake of a one of script it's ok 👍

junyuanxue (approved):
Noice!

PR #4866, opened 2026-08-12
Amazon SDK V1 Clean Up

lindseydew:
## What does this change?

This PR removes the final amazon sdk v1 import and removes any further places the code was importing from this version of the SDK. 

It also removes the credentials version split that was introduced whilst we were migrating the libraries sequentially. 

## Tested? Documented?
- [ ] locally by committer


scripts/src/main/scala/com/gu/mediaservice/scripts/LoadFromS3Bucket.scala
```diff
@@ -18,24 +19,40 @@ object LoadFromS3Bucket {
...
+      val entity = new InputStreamEntity(objStream, length, ContentType.DEFAULT_BINARY)
       postReq.setEntity(entity)
-      httpClient.execute(postReq).close()
+
+      try {
+        httpClient.execute(postReq).close()
+      } finally {
+        objStream.close()
+      }
+
```
junyuanxue, line 55: I don't know if you need to rebase on main - this seems to be updating to v2 rather than clean-up?

lindseydew: Yeah, I only discovered this one during the clean up. It's in the scripts folder and not used by the main app, so I think it should be safe to bundle this into a clean up PR

junyuanxue (approved):
🧹 ✨

PR #4867, opened 2026-08-12
Add feature switch for enforcing upload images permissions

junyuanxue:
## What does this change?

- Looks for a `enforceUploadImagesPermissions` switch from `common.conf` when determining upload images permissions

## Tested? Documented?
- [x] locally by committer

---
  - https://app.asana.com/0/0/1217059235392509


lindseydew (approved):
LGTM 👍 

I'd perhaps suggest that TEST and PROD env are explicitly set when we roll this. It is covered by a default so if this feels tricky do to we can set the env variables later. 

I'm assuming this is isolated from the BBC's implementation? I'd say it's worth getting a double green tick from a dev that would be able to confirm that. Perhaps @andrew-nowak or @twrichards may be able to advise

junyuanxue:
I'm pretty confident that this wouldn't affect BBC's upload permission :)

PR #4869, opened 2026-08-13
Remove min-width from top bar action

junyuanxue:
## What does this change?

Remove min-width from top bar action, so when user doesn't have permission to upload, the other elements in the top bar can stretch across the whole screen.

Screenshot 2026-08-12 at 16 59 12

The `min-width` was put there to address the issue where elements under top bar action wraps awkwardly on a smaller screen size, but this can be fixed by `flex-shrink:0`.

We still have the issue where on a very small screen size, the top bar actions overlaps the filter buttons. But that's the current behaviour so not a regression, and would require a wider design change around placement of top-bar elements on mobile screens.

Screenshot 2026-08-13 at 13 16 20

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-top-bar/gr-top-bar.css
+++ b/kahuna/public/js/components/gr-top-bar/gr-top-bar.css
@@ -16,9 +16,9 @@ gr-top-bar-nav {
 
 gr-top-bar-actions {
     display: flex;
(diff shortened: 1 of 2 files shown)


hk-15 (approved):
Looks good!

PR #4885, opened 2026-08-20
Add searchkey to telemetry and send telemetry when AI search is toggled on

ellenmuller:
## What does this change?

This sorts out some telemetry issues.

We want to be able to group search telemetry by user and by their distinct queries. This PR adds a searchKey to text search telemetry events and tightens up when query telemetry is emitted.

- For text query components, emit a new `searchKey` field of the form `${browserUuid}::${queryValue}`. This lets us filter/aggregate telemetry by user (browser id) and by their distinct search terms. Non-text query components (filters) are unaffected.
- Emit query telemetry when the AI-search toggle is changed, so toggling AI search is now reflected in telemetry.
- Guarded with an `aiSearchInitialised` flag so we only emit on an actual user toggle. The `$watch` fires once on init, and the load-time search event is already emitted exactly once via the `getSession()` -> `watchSearchChange path`, so emitting on init would double-count.

Run locally and look at the telemetry events!

### DIFF
--- a/kahuna/public/js/search/query.js
+++ b/kahuna/public/js/search/query.js
@@ -293,8 +293,8 @@ query.controller('SearchQueryCtrl', [
       manageDefaultNonFree(newFilter);
       manageOrgOwnedSetting(newFilter);
 
-      emitQueryTelemetry()
-      
+      emitQueryTelemetry();
+
       if (ctrl.collectionSearch && !curCollectionSearch) {
         storage.setJs("orderBy", CollectionSortOption.value);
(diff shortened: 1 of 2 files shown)


bryophyta (approved):
Can confirm that the searchKey is populated in GRID_SEARCH events on this branch when running locally, and toggling the AI search checkbox now correctly emits a new GRID_SEARCH event. 🥳 

There's some double-firing of GRID_SEARCH events on initialisation, but these seem to be present on `main` under the same conditions, too. Overall, it doesn't seem like there is a regression here 👍

PR #4890, opened 2026-08-25
Move UsageStore logic to using Grid quota counting instead of RCS emails

junyuanxue:
## What does this change?
This PR moves the logic behind supplier usage status / summary within `UsageStore` over to using the new quota counting logic, instead of relying on parsing emails from RCS.

* As `UsageStore` no longer requires fetching from S3, I've made it no longer inheriting from `BaseStore`. However as it still maintains an in-memory store of over-quota agencies (to be used for the `is:under-quota` search filter), it still needs a scheduler that refreshes the store on an interval.
* I've changed the `exceeded` logic from `>` to `>=`, as we should start surfacing warning when the quota is full, not just exceeded. This is confirmed by the behaviours below. The wording 'exceeded' is a bit misleading, but as it's used quite extensively, I've refrained from not changing it everywhere.

* I have set the Getty quota in TEST to a very low number in S3, so Getty quota would be full in TEST
* Run the Grid locally with `--use-TEST` flag
* On launching the Grid you should see some images with the red square with arrow icon indicating they're from a supplier that's hit their quota:
Screenshot 2026-08-26 at 15 47 13

* Click into one of those images, you should see another warning:
Screenshot 2026-08-26 at 15 47 28

* If you search with the `is:under-quota` filter, the images with the arrow icon should disappear:
Screenshot 2026-08-26 at 16 32 08

* The /quotas page will look like this. Again 4/4 is not exactly 'exceeded', but the alternative is 'quota not full' and I don't think having a 3rd 'quota full' indicator is necessary or particularly useful
Screenshot 2026-08-26 at 15 48 51

## Tested? Documented?
- [x] locally by committer

---
  - https://app.asana.com/0/0/1217597818404838

### DIFF
--- a/media-api/app/lib/UsageStore.scala
+++ b/media-api/app/lib/UsageStore.scala
@@ -1,22 +1,20 @@
 package lib
 
-import java.io.InputStream
-import java.util.Properties
+import java.util.concurrent.atomic.AtomicReference
 
+import org.apache.pekko.actor.{Cancellable, Scheduler}
 import com.gu.mediaservice.lib.BaseStore
-import com.gu.mediaservice.lib.logging.GridLogging
+import com.gu.mediaservice.lib.logging.{GridLogging, LogMarker, MarkerMap}
 import com.gu.mediaservice.model.{Agencies, Agency, UsageRights}
 import com.gu.mediaservice.model.usage.{DigitalUsage, PrintUsage, PublishedUsageStatus, RemovedUsageStatus, UnknownUsageStatus, Usage, UsageStatus, UsageType}
-import javax.mail.Session
-import javax.mail.internet.{MimeBodyPart, MimeMultipart}
-import org.apache.commons.mail.util.MimeMessageUtils
-import org.joda.time.DateTime
+import org.joda.time.{DateTime, Duration}
 import play.api.libs.functional.syntax._
 import play.api.libs.json._
 
-import scala.collection.compat._
 import scala.concurrent.{ExecutionContext, Future}
-import scala.io.Source
+import scala.concurrent.duration._
+import scala.util.control.NonFatal
+import scala.util.{Failure, Success}
 
 case class SupplierUsageQuota(agency: Agency, count: Int)
 object SupplierUsageQuota {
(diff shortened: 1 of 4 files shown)


media-api/app/lib/UsageStore.scala
```diff
@@ -65,143 +62,67 @@ object UsageStore extends GridLogging {
...
+  }
+
+  def stopUpdates(): Unit = cancellable.foreach(_.cancel())
+
+  private def fetchUsage(implicit logMarker: LogMarker): Future[Map[String, SupplierUsageStatus]] = {
+    val supplierQuota = quotaStore.getQuota
+    Future.sequence(
+      Agencies.all.map { case (id, agency) =>
+        getSupplierQuotaCount(id, UsageStore.countPeriodInDays).map { quotaCount =>
+          val quota = supplierQuota.get(id)
```
lindseydew, line 130: Small point, but this feels like the calculations could be done in a standalone method, which could be unit tested independently

media-api/test/lib/UsageStoreTest.scala
```diff
@@ -1,30 +1,153 @@
...
+        val store = makeStore(
+          quotaCounts = Map(gettyId -> 50L),
+          quotaLimits = Map(gettyId -> 100)
+        )
+        val status = updateAndGet(store)
+
+        status.store(gettyAgency.supplier).exceeded shouldBe false
+      }
+
+      it("sets exceeded=true when usage is above the quota limit") {
```
lindseydew: If the calculations were in a different method, some of the test cases here could be reduced

media-api/app/lib/UsageStore.scala
```diff
@@ -65,143 +62,67 @@ object UsageStore extends GridLogging {
...
-  def overQuotaAgencies: List[Agency] = store.get.collect {
+  def overQuotaAgencies: List[Agency] = store.get().collect {
     case (_, status) if status.exceeded => status.usage.agency
   }.toList
 
   def update(): Unit = {
-    store.set(fetchUsage)
+    implicit val logMarker: LogMarker = MarkerMap()
+    logger.info("Updating usage store from ElasticSearch")
+    fetchUsage.foreach { newStore =>
```
lindseydew: Could be worth adding some error handling here if the future was not successful and the store was not able to be updated. 

Wondering also about if the cache was getting stale whether there would need to be a warning to the end user perhaps?

lindseydew (commented):
Looking really good!

Just a question about error handling if the elastic search call fails, and a non-blocking suggestion about abstracting out some of the logic into a smaller method

media-api/app/lib/UsageStore.scala
junyuanxue: Good point re error handling! Re cache getting stale - do you mean towards the end of the 10 mins interval?

lindseydew: I meant if (in the admittedly unlikely) case that the elastic search call continues to fail and the data in the cache gets stale. 

Would it make sense to include the last updated time as part of the response somewhere so it can be surfaced if the data is old? Or write out error logs if it hasn't been updated in a while?

I think it's a pretty unlikely scenario, just thinking about not losing a silent break

junyuanxue: Ah that's a really good point! I think it's a good idea to keep track of last updated time and surface in logs for sure 🙏

lindseydew (approved):
Nice, looks really good ✨

PR #4900, opened 2026-08-28
[e2e-tests] Add upload features

jonathonherbert:
## What does this change?

Adds a set of BDD specifications for the 'uploads' page.

Me and @paperboyo have been through them together.

They should describe everything a user can do on that page. What's 'everything'? Good question! The granularity and detail of these specifications is up for review. I've used [data tables](https://cucumber.io/docs/gherkin/reference/) to parameterise some tests and reduce repetition, which helps us with the surface area for e.g. editable fields.

Read through the steps. Are they comprehensible? Comprehensive? Are there things we are missing?

A comprehensive BDD spec to generate tests from.

## Tested? Documented?
- [x] locally by committer


jonathonherbert:
We probably need a spec for lease addition/removal.

e2e-tests/features/upload.feature.pending
```diff
@@ -0,0 +1,415 @@
...
+  # Evidence: kahuna/public/js/upload/view.html lines 3-6
+
+  Scenario: The current uploads section only appears while an upload is running
+    Given I have an upload in progress
+    When the upload page loads
+    Then I should see my current uploads section
+  # Evidence: kahuna/public/js/upload/view.html lines 14-17
+  # Evidence: kahuna/public/js/upload/controller.js lines 45
+
+  Scenario: Viewing all of my uploads
```
dominickendrick, line 50: Should there be a `Given` Step to set up previous uploads , ie `Given I have already uploaded some images`

e2e-tests/features/upload.feature.pending
```diff
@@ -0,0 +1,415 @@
...
+
+  Scenario: The prompt suggests an example label when no labels are applied
+    Given I have not applied any preset labels
+    When the upload page loads
+    Then I should see a suggested example label to apply to all uploads
+  # Evidence: kahuna/public/js/upload/prompt/prompt.html lines 5-11
+  # Evidence: kahuna/public/js/upload/prompt/prompt.js lines 21
+
+  Scenario: Preset labels are applied to all uploads
+    When I add a preset label in the prompt
```
dominickendrick: Could this also have a `Given` step, as a noob its unclear where / what the `prompt` is in `When I add a preset label in the prompt`

dominickendrick: Sorry I re-read the background step at the begining of the file and saw that there is a default Given step ...

e2e-tests/features/upload.feature.pending
```diff
@@ -0,0 +1,415 @@
...
+    Then I should see an alert that I must drop valid files or URLs
+  # Evidence: kahuna/public/js/upload/dnd-uploader.js lines 222-231
+
+  @todo
+  Scenario: Dragging a Grid image back onto the page is ignored
+    When I drag an image that is already in the Grid over the page
+    Then the dropzone overlay should not appear
+  # Evidence: kahuna/public/js/upload/dnd-uploader.js lines 135-150, 158-160
+
+  # ---------------------------------------------------------------------------
```
dominickendrick, line 180: Given the size of this feature file, I wonder if its worth splitting these scenarios across multiple files ?

e2e-tests/features/upload.feature.pending
```diff
@@ -0,0 +1,415 @@
...
+  Scenario: A missing description defaults to the file name
+    Given an uploaded image has no description
+    When the upload completes
+    Then the description should default to the file name without its extension
+  # Evidence: kahuna/public/js/upload/jobs/upload-jobs.js lines 105-110
+
+  @todo
+  Scenario: Deleting an image removes it from current uploads
+    Given an uploaded image is shown in my current uploads
+    When the image is deleted
```
dominickendrick: Could it be clearer how the image is delete i.e `When the image is deleted using the delete button on the upload form`

e2e-tests/features/upload.feature.pending
```diff
@@ -0,0 +1,415 @@
...
+    Then it should be removed from my current uploads
+  # Evidence: kahuna/public/js/upload/jobs/upload-jobs.js lines 187-195
+
+  @todo
+  Scenario: Uploading a previously deleted image displays it for undeletion
+    Given an image is uploaded
+    And then deleted
```
dominickendrick: Is this the same as the 2 steps above
```
Given an uploaded image is shown in my current uploads
When the image is deleted
```

e2e-tests/features/upload.feature.pending
```diff
@@ -0,0 +1,415 @@
...
+    And I have delete permission
+    Then it should be present in my current uploads
+    And I should be able to undelete it
+  # Evidence: kahuna/public/js/upload/jobs/upload-jobs.js lines 187-195
+
+  # Existing bug: the UI still presents this button even if permission to delete is not present
+  @todo
+  Scenario: Uploading a previously deleted image uploaded by a different user does not display undeletion button without the appropriate permission
+    Given an image is uploaded by a user that is not me
+    And then deleted
```
dominickendrick, line 253: I wonder if the ordering / language would need to change to make the underlying step definitions easier to write, ie by putting the permissions check earlier in the scenario

e2e-tests/features/upload.feature.pending
```diff
@@ -0,0 +1,415 @@
...
+    Then I should see a message that I haven't uploaded anything yet
+  # Evidence: kahuna/public/js/upload/recent/recent-uploads.html lines 4-5
+  # Evidence: kahuna/public/js/upload/recent/recent-uploads.js lines 25-34
+
+  @todo
+  Scenario: My past uploads are listed
+    Given I have uploaded images before
+    When my past uploads load
+    Then I should see each of my past uploaded images
+    And I should be able to delete an image I am permitted to delete
```
dominickendrick, line 406: Is this deletion step already covered in previous deletion scenarios ? Could just remove ?

dominickendrick (approved):
This is looking good. A few small comments

e2e-tests/features/upload.feature.pending
jonathonherbert: I initially imagined just ensuring the query in the URL was enough for this test, e.g. `?uploadedBy=johndoe@example.com` (b/c the actual page behaviour is owned by the search page, which will have a test for actually displaying the images in 'my uploads') — WDYT?

e2e-tests/features/upload.feature.pending
jonathonherbert: Perhaps there's more detail to add here though, e.g. 
```suggestion
    When I add a preset label via the 'apply label to all uploads' button
```

e2e-tests/features/upload.feature.pending
jonathonherbert:
```suggestion
    When the image is deleted using the delete button at the bottom of the job form
```

(job is the vocabulary for the individual thing in this case, so that should be consistent w/ what the tool calls the thing)

e2e-tests/features/upload.feature.pending
jonathonherbert: It's a bit different b/c the current uploads and past uploads lists have slightly different templates. Basically, if you removed or incorrectly specified deletion in one template, it might not impact the other. So it'd be possible to regress.

In another implementation, these behaviours would have one possible representation and so it perhaps would not be possible to regress in this way. 

So I think I'll leave this for now, just b/c we know we want to be able to do this in two places, but it is a bit repetitious.

e2e-tests/features/upload.feature.pending
jonathonherbert: Perhaps permissions gating would be a useful way of organising features 🤔

e2e-tests/features/upload.feature.pending
```diff
@@ -0,0 +1,415 @@
...
+  Scenario: Deleting an image removes it from current uploads
+    Given an uploaded image is shown in my current uploads
+    When the image is deleted using the delete button at the bottom of the job form
+    Then it should be removed from my current uploads
+  # Evidence: kahuna/public/js/upload/jobs/upload-jobs.js lines 187-195
+
+  @todo
+  Scenario: Uploading a previously deleted image displays it for undeletion
+    Given an image is uploaded
+    And then deleted
```
jonathonherbert:
```suggestion
    And then deleted using the delete button at the bottom of the job form
```

e2e-tests/features/upload.feature.pending
jonathonherbert: It's quite large, isn't it! I think this is a good idea.

e2e-tests/features/upload.feature.pending
dominickendrick: For these types of scenario where some setup was required, I opted for being explicit about the setup case ie, `Given I had uploaded 3 images recently`. I guess this scenario could be the default case - so that you always start the stack with some images uploaded by johndoe@example.com, but the fear would be someone changes the default setup, and some tests break. I'm happy to fix forward on this tho

dominickendrick (approved):
Changes look good

PR #4902, opened 2026-08-28
Poll for usage updates on download image

junyuanxue:
## What does this change?

As download usages come from the Grid itself instead of an external platform, it'd be good to have a more 'immediate' update on the usages list when user downloads an image.

When a user downloads an image, `media-api` fires-and-forgets a request to `usage` service, then publishes to Thrall which then updates ElasticSearch, meaning we need to poll for an updated image response to achieve UI feedback.
 
* Ensure that your local confs don't have `image.record.download` set to `false`, or run a `./dev/script/setup.sh` just in case
* Start up Grid `./dev/script/start.sh`
* Go to an image and click on download, you should see the usages list update in a few seconds
download

## Tested? Documented?
- [x] locally by committer

---
  - https://app.asana.com/0/0/1217957220259316

### DIFF
--- a/kahuna/public/js/components/gr-image-usage/gr-image-usage.js
+++ b/kahuna/public/js/components/gr-image-usage/gr-image-usage.js
@@ -22,21 +22,32 @@ export const module = angular.module('gr.imageUsage', [
 
 module.controller('grImageUsageCtrl', [
   '$scope',
+  '$rootScope',
   '$state',
   'inject$',
   '$window',
   'imageUsagesService',
 
-  function ($scope, $state, inject$, $window, imageUsagesService) {
+  function ($scope, $rootScope, $state, inject$, $window, imageUsagesService) {
 
     const ctrl = this;
+
+    const bindUsages = (image) => {
+      const usages = imageUsagesService.getUsages(image);
+      const usages$ = usages.groupedByState$
+        .map((grouped) => grouped
+          .map(list => list.sortBy(usage => usage.get('dateAdded')).reverse())
+          .toJS()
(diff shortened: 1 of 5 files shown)


kahuna/public/js/components/gr-image-usage/gr-image-usage.js
```diff
@@ -72,17 +79,26 @@ module.controller('grImageUsageCtrl', [
...
-      const hasSyndicationUsages$ =
-        imageUsagesService.getUsages(ctrl.image).hasSyndicationUsages$;
+      const freeImagesUpdateListener = $rootScope.$on('images-updated', (e, updatedImages) => {
+        const maybeUpdatedImage = updatedImages.find(u => u.data.id === ctrl.image.data.id);
+        if (maybeUpdatedImage) {
+          ctrl.image = maybeUpdatedImage;
+          const usages = imageUsagesService.getUsages(ctrl.image);
+          const usages$ = usages.groupedByState$
+            .map((grouped) => grouped
+              .map(list => list.sortBy(usage => usage.get('dateAdded')).reverse())
```
lindseydew: Small point, but this block for parsing the usages and injecting is used twice. I wonder if it could be abstracted into its own method?

lindseydew (approved):
LGTM 👍  Nice work ✨

kahuna/public/js/components/gr-image-usage/gr-image-usage.js
junyuanxue: Good call - thanks!

PR #4903, opened 2026-08-31
Allow collapse / expand usages display

junyuanxue:
## What does this change?

When downloads are recorded and displayed as usages, we could easily end up with a longer than usual list of download usages. So I've made a few improvements on usages display:

* Add some sort of order to the statuses displayed. Downloads always come last
* Usages under each list will be sorted by most recent
* Add a toggle to collapse / expand usages. If there are more than 10 items under a usage list, and there is more than 1 usage list, then it'll be collapsed by default

* Go to https://media.test.dev-gutools.co.uk/images/8453f7a7b6fe28e96583f1ece9f834d95a0a69bb, but if the deploy gets overwritten 👇 
* Start up Grid pointing to TEST data `./dev/script/start.sh --use-TEST`
* Go to https://media.local.dev-gutools.co.uk/images/8453f7a7b6fe28e96583f1ece9f834d95a0a69bb, click on 'Usages' tab, you should see 'Downloads' appearing last, collapsed by default as it's got more than 10 usages and has other types of usage lists present:

usage list

## Tested? Documented?
- [x] locally by committer

### DIFF
--- a/kahuna/public/js/components/gr-confirm-delete/gr-confirm-delete.css
+++ b/kahuna/public/js/components/gr-confirm-delete/gr-confirm-delete.css
@@ -8,6 +8,8 @@ gr-confirm-delete.gr-delete-image {
   line-height: inherit;
   width: 100%;
   height: 100%;
+  padding-top: 10px;
+  padding-bottom: 10px;
 }
 
 .gr-confirm-delete:hover {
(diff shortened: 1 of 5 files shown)


hk-15 (approved):
Looks good 🎉

PR #4912, opened 2026-09-03
Do not record download usages from InDesign

junyuanxue:
## What does this change?

Skip usage recording for downloads from InDesign, as print usages are already recorded and they inevitably involve downloads. We mainly care about download usages from actual users.

* Deploy this branch to test
* Create an API key for `InDesign (testing)` in `TEST` key bucket
* In your terminal, run `curl -H "X-Gu-Media-Key: [key]" "https://api.media.test.dev-gutools.co.uk/images/6933e821fc56ee90003c7b0b9b3a8e1d382e055f/download"`
* Go to logs and search for `InDesign (testing)`
*  You should be able to see a log entry with `Skipping usages request to /usages/download for InDesign (testing)`
* And you shouldn't see InDesign downloads under usages on this image: https://media.test.dev-gutools.co.uk/images/6933e821fc56ee90003c7b0b9b3a8e1d382e055f

## Tested? Documented?
- [ ] locally by committer

---
  - https://app.asana.com/0/0/1218150756570430

### DIFF
--- a/media-api/app/controllers/MediaApi.scala
+++ b/media-api/app/controllers/MediaApi.scala
@@ -35,6 +35,14 @@ import scala.concurrent.duration.DurationInt
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.Try
 
+object MediaApi {
+  val InDesignIdentity = "InDesign (testing)"
+
+  def shouldSkipUsageRecording(uri: String, user: String): Boolean =
+    // We don't want to record usages for InDesign downloads for now
+    uri.contains("download") && user == InDesignIdentity
+}
+
 class MediaApi(
                 auth: Authentication,
                 messageSender: ThrallMessageSender,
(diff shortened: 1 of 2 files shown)


media-api/app/controllers/MediaApi.scala
```diff
@@ -35,6 +35,14 @@ import scala.concurrent.duration.DurationInt
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.Try
 
+object MediaApi {
+  val inDesignIdentity = "InDesign (testing)"
```
junyuanxue: Can also stick this in a conf instead of hard-coding here

lindseydew: I think it's fine here. Perhaps use capitals letters to make it clear it's a constant

lindseydew (approved):
LGMT 👍 Nice work ✨

PR #4914, opened 2026-09-04
C2PA detection

junyuanxue:
## What does this change?

We want to be able to show on the Grid whether an image has C2PA metadata available.

The current library we use for parsing image metadata, `metadata-extractor`, does not yet have C2PA support. As the goal is only to show whether C2PA metadata is present, not to parse and display the metadata itself, I've resorted to a 'simple' (relatively-speaking) approach of byte sniffing based on C2PA tech specs, without relying on any external libraries.

* The main C2PA detection logic is in `C2paDetector.scala`, it checks for where C2PA data is stored based on the images mime type according to C2PA tech specs.
* A new `c2pa` field is added on `FileMetadata`:
  * If C2PA metadata is present, it'll be returned as `c2pa: { isAvailable: true }` 
  * If C2PA metadata is absent, it'll be returned as an empty object: `c2pa: {}`
  * The reason for this is: 1) we can extend the `c2pa` object later if we do decide to parse full C2PA metadata one day; 2) this will allow us to use the `has:` search term: `has:fileMetadata.c2pa`

Here are some test images:
[c2pa-absent.jpg](https://github.com/user-attachments/assets/d3bfa9ea-d095-4d2e-98b6-e31f377486a1)
[c2pa-absent.png](https://github.com/user-attachments/assets/292fd51a-1ce1-40bf-9005-360016685663)
[c2pa-absent.tiff](https://github.com/user-attachments/files/31960219/c2pa-absent.tiff)

[c2pa-present.jpg](https://github.com/user-attachments/assets/8aa38bf8-99ad-4aaa-9aa8-b7ae9cda8483)
[c2pa-present.png](https://github.com/user-attachments/assets/34ce0e13-81e8-43e4-a78f-37c764a07559)
Uploading a .tiff image directly in the Grid seems to be failing with a C2PA signature at the moment 😬 but Copilot has suggested: `TIFF is in the C2PA spec (Appendix A.2.5) for completeness, but I'm not aware of any mainstream camera or publishing tool actively signing TIFFs today.`

* In your local Grid `~/.grid/common.conf`, add:
```
field.aliases = [
  {
    elasticsearchPath = "fileMetadata.c2pa.isAvailable"
    alias = "c2paMetadataAvailable"
    label = "C2PA Metadata Available"
    displaySearchHint = true
    searchHintOptions = [ "true", "false" ]
  }
]
```
* Run Grid locally and upload these images
* Go to https://api.media.local.dev-gutools.co.uk/images/[:c2paPresentImageId]/fileMetadata. You should be able to see:
```
    "c2pa": {
      "isAvailable": true
    }
```
* Go to https://api.media.local.dev-gutools.co.uk/images/[:c2paAbsentImageId]/fileMetadata. You should be able to see: `{ "c2pa" : {} }`
* On the image page, you should be able to see that C2PA is available under 'Hidden additional metadata'. This field won't be present is it's not available
Screenshot 2026-09-08 at 15 56 23

* You should be able to search with `has:fileMetadata.c2pa` and `c2paMetadataAvailable:true`
has search

* `c2paMetadataAvailable: false` won't yet be working - will be handled in upstack PR https://github.com/guardian/grid/pull/4920

## Tested? Documented?
- [x] locally by committer

---
  - https://app.asana.com/0/0/1218187532710129


hk-15 (commented):
Very cool! I used the C2PA tool to sign my own images and all working as expected 🎉

I can get the searching to work in the URL and search bar, but clicking on `true` in the metadata sidebar doesn't work as it's populating the url with undefined: https://media.local.dev-gutools.co.uk/search?query=c2paMetadataAvailable:undefined&nonFree=true

Image

common-lib/src/main/scala/com/gu/mediaservice/model/FileMetadata.scala
```diff
@@ -56,7 +60,8 @@ object FileMetadata {
     (__ \ "icc").readNullable[Map[String,String]].map(_ getOrElse Map.empty) ~
     (__ \ "getty").readNullable[Map[String,String]].map(_ getOrElse Map.empty) ~
     (__ \ "colourModel").readNullable[String] ~
-    (__ \ "colourModelInformation").readNullable[Map[String,String]].map(_ getOrElse Map.empty)
+    (__ \ "colourModelInformation").readNullable[Map[String,String]].map(_ getOrElse Map.empty) ~
```
lindseydew, line 63: Small thing and fine for a spike, but I'd suggest making this field a dynamic map like the other image metadata rather than a json object

lindseydew (commented):
Looking great 💯

common-lib/src/main/scala/com/gu/mediaservice/model/FileMetadata.scala
junyuanxue: Ah good catch! Forgot to tidy up 🙏

kahuna/public/js/search/query-filter.js
```diff
@@ -16,7 +16,7 @@ export function maybeQuoted(value) {
 }
 
 export function fieldFilter(field, value) {
-    const cleanValue = stripDoubleQuotes(value);
+    const cleanValue = stripDoubleQuotes(String(value));
```
junyuanxue, line 19: @hk-15 This should fix the `undefined` in the query param:

Image

lindseydew (approved):
Nice looks great 👍

Issue #37, opened 2014-07-30
Allow filtering for images with "free rights"


theefer:
Requested by claire.phipps@guardian.co.uk (and likely many others):

> Is there - or could there be - a way to search pic desk/library only for pictures with a green tick? I appreciate pic editors need to see the full spread, but when liveblogging, for example, I want to be able to search only for pictures I can actually use, and not have to click through pages of red/orange-flagged ones first, which really slows me down. Would be great if possible!

PR #4915, opened 2026-09-04
Add to PhotographerRenamer dictionary #37

paperboyo:
Ronseal.


twrichards (approved):

AlexAscherson (approved):
Looks good!  Assuming the special characters only in the presented name are intended.

PR #4919, opened 2026-09-08
Add Grid ID lookup to AlamyCleanUp script

hk-15:
## What does this change?

The csv provided by Alamy can be pasted into `alamy-refs.csv` and the script will fetch the Grid IDs first then perform the clean up.

Run the script with the stage set to TEST.

Quicker to run clean up.

## Tested? Documented?
- [x] locally by committer


scripts/src/main/resources/alamy-refs.csv
hk-15, line 1: Added this so it's easy to just copy in the csv, rather than adding it to Grid and changing the file name, but perhaps there is an even better way?

junyuanxue: I'm wondering if instead of having a `.csv` file in the repo, we could pass in a filepath to the CSV as an argument of the script?

hk-15: Aha! Yes that is better thank you. Done in [b341e86](https://github.com/guardian/grid/pull/4919/commits/b341e86a713793515e973190280bb3b0115dd812)

scripts/src/main/scala/com/gu/mediaservice/scripts/AlamyCleanUp.scala
```diff
@@ -1,25 +1,62 @@
...
   val GRIDDOMAIN = if(STAGE == "PROD") "gutools.co.uk" else "test.dev-gutools.co.uk"
 
-  val ids = if(STAGE == "PROD") {
-    val resource = Source.fromResource("alamy-grid-ids.csv")
-    resource.getLines().toList
+  println(s"Running for stage $STAGE with domain $GRIDDOMAIN")
+
+  val supplierRefs = if (STAGE == "PROD") {
+    val csvPath = args.headOption.getOrElse {
+      throw new IllegalArgumentException("Usage: AlamyCleanUp <csv-file>")
```
lindseydew: Might be good to have a message to explain that the user needs to pass in a path to a csv file as an arg if no arg is found

lindseydew (approved):
Nice, lovely extension 👍

junyuanxue (approved):

PR #4920, opened 2026-09-08
Support matching via existence in ES for field aliases

junyuanxue:
## What does this change?

Some indexed fields are only ever written when a condition is `true`  — e.g.  `fileMetadata.c2pa.isAvailable` is only present at all when C2PA metadata was detected on an image (for reasoning, see https://github.com/guardian/grid/pull/4914); there's no explicit `false`. Users are given a  `true / false` search-hint dropdown for these fields (e.g.  `c2paMetadataAvailable:true / :false` ), but a literal  `:false` term query could never match anything, since  `false` is never actually indexed — it's only ever implied by the field being absent. That made  `:false` misleadingly return zero results instead of "all images without this field".

To address this:
* Added a  `matchViaExistence`  flag to `FieldAlias`, settable in  `common.conf`. When set:
  * `alias:true`  → rewritten as an exists query on the field's underlying Elasticsearch path
  * `alias:false`  → rewritten as a not-exists query ( bool.must_not(exists) )
* This is handled in `QueryBuilder.makeQueryBit`, which checks `isMatchViaExistence`  and, for boolean-looking phrase/word values on such fields, builds the `exists/not-exists` query instead of the normal literal  `matchQuery`/ `matchPhraseQuery`. Non-boolean values on the same field still fall through to a normal literal match.
* Also extended field lookup ( `resolveFieldPath / isMatchViaExistence` ) so a search can use either the friendly alias (e.g. `c2paMetadataAvailable`) or the raw Elasticsearch path (`fileMetadata.c2pa.isAvailable`) interchangeably — previously only the alias worked.

* In your local `~/.grid/common.conf`, add:
```
field.aliases = [
  {
    elasticsearchPath = "fileMetadata.c2pa.isAvailable"
    alias = "c2paMetadataAvailable"
    label = "C2PA Metadata Available"
    displaySearchHint = true
    searchHintOptions = [ "true", "false" ]
    matchViaExistence = true
  }
]
```
* Start up the Grid locally
* You should now be able to search with: `c2paMetadataAvailable: false` and `+fileMetadata.c2pa.isAvailable:false`

false search
false search 2

## Tested? Documented?
- [x] locally by committer

---
  - https://app.asana.com/0/0/1218249327946580

### DIFF
--- a/media-api/test/lib/elasticsearch/QueryBuilderTest.scala
+++ b/media-api/test/lib/elasticsearch/QueryBuilderTest.scala
@@ -5,7 +5,7 @@ import com.sksamuel.elastic4s.ElasticDsl
 import com.sksamuel.elastic4s.requests.common.Operator
 import com.sksamuel.elastic4s.requests.searches.queries._
 import com.sksamuel.elastic4s.requests.searches.queries.matches.{MatchPhraseQuery, MatchQuery, MultiMatchQuery, MultiMatchQueryBuilderType}
-import lib.querysyntax.Negation
+import lib.querysyntax.{Match, Negation, Phrase, SingleField}
 import com.gu.mediaservice.lib.config.GridConfigResources
 import com.sksamuel.elastic4s.handlers.searches.queries.QueryBuilderFn
 import com.sksamuel.elastic4s.requests.searches.queries.compound.BoolQuery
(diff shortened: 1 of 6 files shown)


media-api/test/lib/elasticsearch/QueryBuilderTest.scala
```diff
@@ -183,6 +183,79 @@ class QueryBuilderTest extends AnyFunSpec with Matchers with ConditionFixtures w
...
+      new ApplicationLifecycle {
+        override def addStopHook(hook: () => Future[_]): Unit = {}
+        override def stop(): Future[_] = Future.successful(())
+      }
+    ))
+    val queryBuilderWithC2paAlias = new QueryBuilder(matchFields, () => Nil, configWithC2paAlias)
+
+    def existsClauseOf(query: Query): ExistsQuery =
+      query.asInstanceOf[BoolQuery].filters.head.asInstanceOf[ExistsQuery]
+
```
lindseydew, line 207: Small pedantic thing, which feel free to ignore, but since the implementation has case insensitive checking could perhaps add a test case for this.

lindseydew (approved):
Nice, looks really good ✨

PR #4930, opened 2026-09-10
Use a date time parser for checking dates

lindseydew:
## What does this change?
This uses date time parsing so that we can check the equivanlence of dates that way, rather than using string and int matches. 
## Tested? Documented?
- [ ] locally by committer


junyuanxue (approved):
Nice, much cleaner! 🧼 ✨  Would be good to have some unit tests for the `PartialDate` method though :)

PR #4934, opened 2026-09-14
Set up takedown page UI

junyuanxue:
## What does this change?

A scaffold of takedown page UI, with some placeholder copy for each step.

* Page /takedown is set up with React, and uses `@guardian/stand` components. I've added some default theme overrides to support the dark background of the Grid.
* I've opted for custom styling localised to elements using `@emotion/react`, in line with `@guardian/stand` style overrides, so we don't have the problem of having multiple ways of styling within one component. (And generally, I think we should move away from CSS stylesheets and towards more localised styles and leaning on thematic abstraction rather than abstraction by custom classnames).
* Some base sketch and UI of hard-coded takedown steps - these will be filled out in upstack PRs.
* Page is behind a feature switch through `kahuna.conf` for now.

* In your local `~/.grid/kahuna.conf`, add `image.takedown.enabled=true`, run the Grid using `--use-TEST` so you have permissions

Screenshot 2026-09-14 at 14 57 45

## Tested? Documented?
- [x] locally by committer

---
  - https://app.asana.com/0/0/1218365541511350


kahuna/public/js/takedown/takedown-page.tsx
```diff
@@ -0,0 +1,149 @@
...
+              status={step.getStatus()}
+              isLastStep={index === STEPS.length - 1}
+            >
+              <step.Component image={image} />
+            </TakedownStep>
+          ))}
+        </Layout.Main>
+      </Layout>
+    </div>
+  );
```
lindseydew, line 141: I like this flow. 

I'm wondering a bit about the status updates. 

We can get to it nearer the time, but I think it would be good to use the data from the API to say whether or not it can be clicked

lindseydew (approved):
LGTM, nice work ✨

kahuna/public/js/takedown/takedown-page.tsx
junyuanxue: Yes, this will be based on data-fetching results inside a context provider: https://github.com/guardian/grid/pull/4935/changes#diff-32c0a79d82d1fd5f0c4e9c8e458d55c64537f51cc7144f77acbaaa3b6daeccb9

PR #4935, opened 2026-09-14
Delete from content step

junyuanxue:
## What does this change?

* Add a `DeleteFromContentStep` component on the image takedown page
* Add a `TakedownContext`, which handles data-fetching for latest live content where the image is used, and derives step status from that

delete from content

* It's a bit annoying to test locally due to `/capiUsages` endpoint calling Preview CAPI, so best way is to test with a deployment in TEST
* Go to Composer and add an image in an article, when you go back to Grid takedown page, you should see the article appear and the 2nd step being disabled
* When you remove the image from the Composer article, and return to Grid takedown page, you should see the article disappear and the 2nd step enabled

## Tested? Documented?
- [x] locally by committer

---
  - https://app.asana.com/0/0/1218460637659940


kahuna/public/js/takedown/takedown-context.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+    const fetchActiveContent = async () => {
+      try {
+        // @TODO: Fetch for CAPI content from Media API rather than usages
+        // Use `.get()` rather than `.getData()` on the usages resource: theseus
+        // Resources cache their response at construction time, so `.getData()`
+        // on the same Resource instance would keep replaying the response from
+        // when `image` was first fetched, never reflecting usages
+        // added/removed since. `.get()` performs a fresh HTTP GET each time.
+        const usagesResource = await image.data.usages.get();
```
junyuanxue: Fetching from usages for now - will switch to fetching for live CAPI content once ready

kahuna/app/lib/KahunaConfig.scala
```diff
@@ -51,8 +51,6 @@ class KahunaConfig(resources: GridConfigResources) extends CommonConfig(resource
   val showSendToPhotoSales: Option[Boolean] = booleanOpt("showSendToPhotoSales")
   val aiSearchResultLimit: Int = intOpt("ai.search.resultLimit").getOrElse(200)
 
-  val imageTakedownEnabled: Boolean = booleanOpt("image.takedown.enabled").getOrElse(false)
```
junyuanxue, line 54: No longer needed as I can use the presence of `capiUsage` link, which is behind a similar conf in Media API, to guard the page

kahuna/public/js/takedown/delete-from-content-step.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+            {activeContent !== null && activeContent.length > 0 && (
+              <>
+                <Typography
+                  element="p"
+                  variant="bodySm"
+                  theme={standThemeOverride.typography.default}
+                >
+                  This image is currently used in the following content. Remove
+                  it from each one through Composer before continuing.
```
hk-15, line 94: nit: this wording sounds a bit more natural to my ear, and the Composer step seems obvious enough, especially given that link is provided and only trained CPers will be doing this 
```suggestion
                  This image is currently used in the following content. Remove
                  it before continuing.
```

kahuna/public/js/takedown/takedown-context.tsx
```diff
@@ -0,0 +1,144 @@
...
+  useEffect(() => {
+    if (!image) {
+      return;
+    }
+
+    let cancelled = false;
+
+    const fetchActiveContent = async (): Promise<CapiUsage[]> => {
+      // Use `.get()` rather than `.getData()` on the followed resource:
+      // theseus Resources cache their response at construction time, so
```
hk-15, line 52: confused by this part of the comment? theseus Resources?

hk-15 (approved):
This works very nicely! I have only small comments

kahuna/public/js/takedown/takedown-context.tsx
junyuanxue: https://github.com/argo-rest/theseus - this is the library they use for API requests

PR #4939, opened 2026-09-15
[e2e-tests] Add drag and drop tests for the upload feature

jonathonherbert:
## What does this change?

Adds e2e tests for the drag-and-drop functionality.

- Read the specs. They should look fairly straightforward.
- Do the folder changes make sense? I feel like aligning on a folder structure for steps across projects would be useful.

More test coverage for uploads.


e2e-tests/steps/healthcheck/kahuna.steps.ts
```diff
@@ -1,4 +1,4 @@
-import { When, Then, expect } from './fixtures.ts';
+import { When, Then, expect } from '../setup.ts';
```
dominickendrick, line 1: I think this new folder structure is great, it helps group similar features etc

e2e-tests/steps/upload/dnd.steps.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/** From kahuna/public/js/strings.json. */
+const DROPZONE_EXPLANATION =
+  'Drop files or GuardianWitness URLs here to upload them instantly to the Grid';
+
+When('I drag files over the upload page', async ({ page }) => {
+  await dragOver(page, { files: filesToUpload });
+});
+
+Then('the dropzone overlay should appear with an explanation', async ({ page }) => {
```
dominickendrick, line 13: I was going to say these might be a bit imperative, but I can't think of a better alternative ;) https://github.com/guardian/newsletters-nx/blob/main/docs/gherkin-style-guide.md#2-declarative-over-imperative

dominickendrick (approved):
Looks great !

e2e-tests/steps/upload/dnd.steps.ts
jonathonherbert: Hm, yeah, that is the behaviour we need to see — it's just a granular enough scenario that there's no getting away from the specificity of 'see the dropzone' 🤔

PR #4941, opened 2026-09-15
Remove metadata logging

jonathonherbert:
## What does this change?

Remove some no-op logging that appears after editing metadata on the upload page. Can't see there's a reason to keep this around, and we accrue a new interval on every save — @andrew-nowak any reason we shouldn't remove?

## Tested? Documented?
- [x] locally by committer


andrew-nowak (approved):
🤷

paperboyo:
It’s starting! Drive-by maintenance ❤️.

owned-dog

jonathonherbert:
I think I am both the dog and the man.

PR #4942, opened 2026-09-15
Delete from Grid step layout

junyuanxue:
## What does this change?

* Add 'delete from grid' step, UI only and not yet wired up with delete API calls (to be handled upstack), to keep the code review easier
* This step will render a list of crops and usages, allow user to select from 1 out of 3 delete methods, then confirm the deletion
* I've opted for a simple, flat list of usages display, with downloads collapsed into 'x downloads'. This is not consistent with 'Usages' tab but as this just serves as a quick confirmation before the user proceeds to nuke the image, I think we should stick to whatever is the simplest
* I needed to add a `security.styleSources` conf so we can load Material Symbols from @guardian/stand, see [icon doc](https://guardian.github.io/stand/?path=/docs/stand-tools-design-system-components-icon--docs) 
* I have restructured the theme overrides a little so we clearer type definitions

* It's a bit annoying to test locally due to /capiUsages endpoint calling Preview CAPI, so best way is to test with a deployment in TEST

delete from grid

## Tested? Documented?
- [x] locally by committer

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

---
  - https://app.asana.com/0/0/1218460637659941


kahuna/public/js/takedown/delete-from-grid-step.tsx
```diff
@@ -0,0 +1,350 @@
...
+                      Cancel
+                    </Button>
+                  </div>
+                </>
+              )}
+            </div>
+          </>
+        );
+      }}
+    </ClassNames>
```
lindseydew, line 348: You might already be thinking about this in a follow up PR, but I'm wondering about how this page will function after a delete action has occured in grid to confim it. 

Eg. when deny lease has already been set, or softdelete has already happened or if the image is 404ing because it has already been deleted

lindseydew (approved):
Looking great 💯

kahuna/public/js/takedown/delete-from-grid-step.tsx
junyuanxue: Yes you're 💯 right - sketching it out upstack at the moment:
- This page won't ever 404 and will indicate whether the image has been soft / hard-deleted or if there's a permanent deny lease
- I'm even thinking, if a deny lease has been applied, and later the user wants to also soft / hard delete it, we can allow them to do that

junyuanxue: > This page won't ever 404

Edit on that - there's a base check for a valid media id, and if it's clearly not valid, the page will 404 :)

PR #4950, opened 2026-09-18
[e2e-tests] Test the current uploads list

jonathonherbert:
## What does this change?

Adds end-to-end coverage (Playwright + `playwright-bdd`) for the upload page's
**current uploads list** (`kahuna/public/js/upload/jobs/*`), turning these
[`e2e-tests/features/upload.feature`](e2e-tests/features/upload.feature) scenarios
from `@todo` into runnable tests:

- the remaining-uploads count while several uploads are in flight
- an uploading job's preview thumbnail, file name, size and status
- the upload-error state, and removing a failed job after confirming
- a completed upload becoming an editable image
- deleting a completed image removing it from the list

Step definitions live in
[`steps/upload/current-uploads.steps.ts`](e2e-tests/steps/upload/current-uploads.steps.ts),
with shared helpers (unique-per-run image, ingest-bucket route stubs, completed-job
locators) added to [`steps/upload/setup.ts`](e2e-tests/steps/upload/setup.ts). No
production code changes — this is test-only.

A couple of things worth a reviewer's attention:

- **Scoping.** The current uploads list and the recent uploads list share markup
  (`li.upload-result`, the `ui-image-editor` metadata region), so every assertion is
  scoped to the "Your current uploads" region — an unscoped locator matches both.
- **Shared-state safety.** Uploads persist on the single e2e user and the Grid
  dedupes by content hash, so the delete scenario uploads a per-run unique image
  (fixture bytes + random trailing bytes) to avoid poisoning other scenarios and
  re-runs.
- **Forcing states.** In-progress and upload-error states are driven by intercepting
  the ingest-bucket `PUT` via `page.route` (hold, or fulfil `403`), not by changing
  app code.
- Also fixes two step imports left pointing at the old `fixtures.ts` after an earlier
  file move, which had broken `bddgen`.

Four scenarios stay `@todo`, each with the reason recorded against it in the feature
file. Three are blocked by app issues surfaced while writing these tests (happy to
raise separately): the unsupported-file-type message is unreachable in
straight-to-bucket mode; the missing-description default is written server-side but
never reflected in the editor; the undelete control is state-sticky and mislabelled
`aria-label="Delete image"`; and the failed-deletion alert never fires because its
`$rootScope.$on` handler reads the wrong argument.

From `e2e-tests/`:

```
npm test
```

This boots (or attaches to) a Grid stack and runs the upload scenarios; `@todo`
scenarios are skipped. Use `npm run test:headed` to watch them, or
`npm run test:report` for the HTML report.

- The current-uploads scenarios pass in CI, giving regression coverage over that part
  of the upload journey.
- The remaining `@todo` scenarios document the backlog still to be covered, and the
  three app bugs called out above.

## Who should look at this?

@guardian/digital-cms

## Tested? Documented?
- [x] locally by committer


e2e-tests/steps/upload/setup.ts
```diff
@@ -36,13 +38,60 @@ export const filesToUpload = [testImages.smaller, testImages.larger];
  */
 export const gridHostedImageUrl = `http://localhost:${KAHUNA_PORT}/assets/images/blocked-cookies.png`;
 
+/**
+ * A JPEG unique to this run. The Grid dedupes by content hash, so a scenario that deletes
+ * its image would otherwise poison the shared fixtures and its own re-runs; random trailing
+ * bytes change the hash without stopping the image decoding.
+ */
+export const uniqueImage = (): TestImage => {
```
dominickendrick: 👏

e2e-tests/steps/upload/setup.ts
```diff
@@ -1,4 +1,6 @@
-import { statSync } from 'node:fs';
+import { statSync, readFileSync, writeFileSync } from 'node:fs';
```
dominickendrick: I wonder if this file could have a different name ? `page-model`? `test-helper`? It just conflicts slightly with the other setup file in the parent dir

build.sbt
```diff
@@ -150,12 +150,18 @@ lazy val collections = playProject("collections", 9010)
 
 lazy val cropper = playProject("cropper", 9006)
 
-lazy val imageLoader = playProject("image-loader", 9003).settings {
+lazy val imageLoader = playProject("image-loader", 9003).settings(
   libraryDependencies ++= Seq(
     "org.apache.tika" % "tika-core" % "3.2.3",
     "com.drewnoakes" % "metadata-extractor" % "2.19.0"
+  ),
+  Universal / mappings ++= Seq(
```
dominickendrick: I presume this needed to stay, even though we changed the build setup ??

dominickendrick (approved):
Looks great!

build.sbt
jonathonherbert: Yes — after all that! Unless we don't do things the `stage`d way, we're stuck with this 😢 (so thanks for the fix)

PR #4953, opened 2026-09-18
[e2e-tests] Add tests for the upload required-metadata editor

jonathonherbert:
## What

Adds Playwright + `playwright-bdd` coverage for the Kahuna upload **required-metadata editor** (phase 6 of the upload journey), enabling these previously `@todo` scenarios in `e2e-tests/features/upload.feature`:

- Editing required metadata (auto-save)
- Description & credit are required
- Description placeholder guidance
- Choosing an image type (from configured types)
- Credit field suggestions
- Metadata only shows when it was already present (embedded IPTC)
- Existing usage instructions are shown with room for more
- Applying a field value to all current uploads in a batch
- Applying / removing a metadata template

To exercise these end-to-end (not just the client), it also:

- **Bakes real server config into the e2e stack** (`e2e-tests/setup/config.ts`): `imageTypes` and `metadata.templates` on `kahuna.conf`, and `usageInstructions` (keyed by usageRights category) on `media-api.conf`.
- **Adds fixtures**: `embedded-metadata.jpg` (IPTC byline/credit/copyright/description/instructions) and `agency-usage.jpg` (credited `AAP`, so supplier processing gives it the `agency` category that drives `usageInstructions`); and derives the `suggestMetadataCredit` completion field into the seeded Elasticsearch fixtures so the real credit suggester works.
- **Small app changes** to make the UI testable via user-facing locators instead of CSS classes (see below).

## Why

These scenarios were specified but unimplemented. Driving them through real Kahuna/media-api config and genuine fixtures (rather than client-side stubs) gives confidence the whole path works, and the coverage documents the editor's behaviour.

## App changes to note (production code)

Both improve accessibility and let the tests use role/text locators per [Playwright's guidance](https://playwright.dev/docs/best-practices#use-locators):

- `kahuna/public/js/forms/datalist.html`: `role="listbox"` / `role="option"` on the suggestions list.
- `kahuna/public/js/upload/dnd-uploader.html`: `role="region"` + `aria-label` ("Dropzone" / "Importing") on the overlays.

## How to test

```bash
cd e2e-tests
GRID_PROXY=true npm run dev:e2e   # boot the stack with the bundled Caddy proxy
npm run test                      # in another shell
```

A fresh boot seeds Elasticsearch automatically. Because this adds media-api config, verifying `usageInstructions` needs a stack (re)boot; against a reused stack, reseed with `GRID_RESEED=true` for the credit-suggestions fixture.

## Notes / follow-ups

- Stacked on `jsh/add-upload-list-steps` (the PR base), not `main`.
- Left `@todo` in phase 6: "Applying an **image-editor** field value in a batch" (Leases/Collections/Labels/Keywords/Photoshoot live in `image-editor.html`), and "Metadata editing is disabled without edit permission" (needs a second user / different permissions, which this single-user stack can't provide).
- `imageType`/`description` are excluded from the batch scenario because their "apply to all" button only renders while the value equals the original.


e2e-tests/features/upload.feature
```diff
@@ -276,90 +276,91 @@ Feature: Uploading images to the Grid
...
     Given an uploaded image that already has usage instructions
     When I view the metadata editor
     Then I should see the existing usage instructions
     And I should be able to add further special instructions
   # Evidence: kahuna/public/js/upload/jobs/required-metadata-editor.html lines 140-167
 
-  @todo
+  # Covers the required-metadata-editor fields whose ⇔ "apply to all" button persists after
+  # editing. imageType/description are excluded — their ⇔ only renders while the value still
+  # equals the original, so it vanishes as you type the value to apply.
```
paperboyo, line 337: I don’t understand that, I may be missing sth. All three: Desc, Leases and Image Type can be batch applied, just like other fields. For some, button indeed disappears when the saving into current image takes place, but then reappears as soon as saving’s done.

